PR #723, opened 2015-07-15
make the split collective shared file pointer operations work

edgargabriel:
- minor code restructering in io/ompio required for that.


edgargabriel:
@hppritcha not sure whether you are interested in reviewing one more commit, namely to make the split collective shared file pointer operations work ( I know, everybody is using these functions :-) ) If you want to run a testcase on that, the filetest in the mpi2basic_tests directory contains something as well. Shared file pointer operations are bit tricky however, either you execute them on a single node on a lustre file system, or on a file system which supports file locking.

hppritcha:
on the cray's lustre usually is configured with file locking supported.  I'll give it a try on the nersc system.

edgargabriel:
@hppritcha , based on some of the comments that I received from Jeff on my other PR, I will close this one and refile a new PR. The reason for that is that I brought in an unrelated fix that should be a separate commit, just to make things cleaner and more clear.

PR #448, opened 2015-03-05
Add a new topo module: Treematch

bosilca:
This PR contains two commits that are combined into a single PR as they are related. First, it adds a new topo module allowing for process reordering based on user-provided weights. Second, it adds a new overlay PML (one that works in combination with another PML), allowing to accurately track the amount of data send between processes.

PR #725, opened 2015-07-16
Add a new topo module: Treematch

bosilca:
This PR is a replacement for the failed #448. It adds a new topo module allowing for process reordering based on user-provided weights

### DIFF
--- a/README
+++ b/README
@@ -1,7 +1,7 @@
 Copyright (c) 2004-2007 The Trustees of Indiana University and Indiana
                         University Research and Technology
                         Corporation.  All rights reserved.
-Copyright (c) 2004-2007 The University of Tennessee and The University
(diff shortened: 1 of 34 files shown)


jsquyres:
@bosilca This generally looks good.  Two minor comments:

1. Do you want to squash these down into 1 commit?  I.e., is there a reason to have the separate commits?
1. Do you need to add anything to README to explain what this new treematch module is?

bosilca:
@jsquyres thanks for the suggestions. I synched with master, updated the README and squashed all commit into a single one.

hppritcha:
Is this ready to merge?

Do we want this pushed back to v2.x?

jsquyres:
Yes, this is ready to go.  Thanks @bosilca!

PR #726, opened 2015-07-20
Pr/sharedfp sm logic

edgargabriel:
@jsquyres would you mind having a brief look at the configure logic for the sharedfp/sm component? The goal is to use named semaphores (sem_open) when available, unnamed semaphores otherwise (sem_init), and disqualify the component if neither of them is available.


edgargabriel:
argh, will do a new pull request, this was based on a branch that is not yet merged into master, so there is waaay too much stuff in it.

lanl-ompi:
Test FAILed.

PR #727, opened 2015-07-20
add the configure logic for detecting whether to use named or unnamed semaphores

edgargabriel:
- restructure the code to use HAVE_SEM_OPEN etc. instead of my own macro

### DIFF
--- a/ompi/mca/sharedfp/sm/configure.m4
+++ b/ompi/mca/sharedfp/sm/configure.m4
@@ -1,0 +1,49 @@
+# -*- shell-script -*-
+#
+# Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+#                         University Research and Technology
+#                         Corporation.  All rights reserved.
+# Copyright (c) 2004-2005 The University of Tennessee and The University
+#                         of Tennessee Research Foundation.  All rights
+#                         reserved.
+# Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
+#                         University of Stuttgart.  All rights reserved.
+# Copyright (c) 2004-2012 The Regents of the University of California.
+#                         All rights reserved.
+# Copyright (c) 2010-2014 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2008-2015 University of Houston. All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# MCA_sharedfp_sm_CONFIG(action-if-can-compile,
+#                        [action-if-cant-compile])
+# ------------------------------------------------
+AC_DEFUN([MCA_ompi_sharedfp_sm_CONFIG],[
+    AC_CONFIG_FILES([ompi/mca/sharedfp/sm/Makefile])
+
+    sharedfp_sm_happy=no
+    sharedfp_sm_open_happy=no
+    sharedfp_sm_init_happy=no
+    AC_CHECK_HEADER([semaphore.h],
+                    [AC_CHECK_FUNCS([sem_open],[sharedfp_sm_open_happy=yes],[])])
+
+
+    if test "$sharedfp_sm_open_happy" = "yes" ; then
+	sharedfp_sm_happy="yes"
+    else
+	AC_CHECK_HEADER([semaphore.h],
+            [AC_CHECK_FUNCS([sem_init],[sharedfp_sm_init_happy=yes],[])])
+	
+	if test "$sharedfp_sm_init_happy" = "yes" ; then
+	    sharedfp_sm_happy="yes"
+	fi
+    fi
+
+    AS_IF([test "$sharedfp_sm_happy" = "yes"],
+          [$1],
+          [$2])
+])dnl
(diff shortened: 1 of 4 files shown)


edgargabriel:
@jsquyres would you mind having a brief look at the configure logic for the sharedfp/sm component? The goal is to use named semaphores (sem_open) when available, unnamed semaphores otherwise (sem_init), and disqualify the component if neither of them is available. This is now the correct pull request

lanl-ompi:
Test FAILed.

ompi/mca/sharedfp/sm/configure.m4
```diff
@@ -0,0 +1,49 @@
...
+#                        [action-if-cant-compile])
+# ------------------------------------------------
+AC_DEFUN([MCA_ompi_sharedfp_sm_CONFIG],[
+    AC_CONFIG_FILES([ompi/mca/sharedfp/sm/Makefile])
+
+    sharedfp_sm_happy=no
+    sharedfp_sm_open_happy=no
+    sharedfp_sm_init_happy=no
+    AC_CHECK_HEADER([semaphore.h],
+                    [AC_CHECK_FUNCS([sem_open],[sharedfp_sm_open_happy=yes],[])])
```
jsquyres: Minor quibble: you don't need the final empty [] clause -- you can just leave it out.

ompi/mca/sharedfp/sm/configure.m4
```diff
@@ -0,0 +1,49 @@
...
+    if test "$sharedfp_sm_open_happy" = "yes" ; then
+	sharedfp_sm_happy="yes"
+    else
+	AC_CHECK_HEADER([semaphore.h],
+            [AC_CHECK_FUNCS([sem_init],[sharedfp_sm_init_happy=yes],[])])
+	
+	if test "$sharedfp_sm_init_happy" = "yes" ; then
+	    sharedfp_sm_happy="yes"
+	fi
+    fi
```
jsquyres: Just curious: is there a reason you used 3 shell variables instead of just 1?

ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
```diff
@@ -135,7 +135,6 @@ int mca_sharedfp_sm_file_open (struct ompi_communicator_t *comm,
         return OMPI_ERROR;
     }
 
-        free(sm_filename);
```
jsquyres: Is this `free()` supposed to be part of this pull request?  It doesn't seem related to the semaphore stuff.

jsquyres:
@edgargabriel General question: is there a reason you have 2 different mutexes?  (i.e., one in `struct sm_offset` and one in `struct mca_sharedfp_sm_data` (which contains a `struct sm_offset*`)?  This leads to lots of `#if` statements that just select whether to `sem_FOO()` on the one in sm_offset vs. mca_sharedfp_sm_data.

jsquyres:
...also, the name `struct sm_offset` doesn't obey the prefix rule.  :smiling_imp:

ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
edgargabriel: you caught it, wow :-) No, technically its not, its a separate fix, somebody added it to the file (probably as part of the covertiy cleanup), but it is completely wrong, it actually makes this component seg fault. I can commit it separately if you want.

ompi/mca/sharedfp/sm/configure.m4
edgargabriel: no, just found the initial logic simpler than with just one variable.

edgargabriel:
Looking into the mutexes, honestly don't remember of the top of my head. But can probably be merged to a single  one. And I can change the name of the structure

ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
jsquyres: :smile: It's probably fine to include it on this PR, but it's probably a good idea to go ahead and make it a separate commit (ping me offline if you wanna see the git magic to split this out into a separate commit).  Thanks!

edgargabriel:
ok, so I spent quite some time looking into that, and it looks like the two mutex can not easily be merged. The reason is, that one function (sem_init) expects the semaphore to be in the shared memory region, while the other (sem_open) does not.  However, I renamed the structure sm_offset to mca_sharedfp_sm_offset 

Furthermore, while debugging the two scenarios I realized that my previous attempt to merge the two mutexes actually broke the non-named semaphore code. I fixed it right now ( so both named and unnamed semaphore passed the test suite), I am just wandering whether I should truly commit it to this branch, or whether you would like me to do it separately. the same question as with the additional free statement

edgargabriel:
After thinking about that, I think the cleanest way will be to i) close this pull request ii) commit the two fixes to master that are independent of this pull request and iii) generate a new pull request with the configure logic and the macro changes only. It should not take me too long, but it will probably be monday before I get to it. Does this sound ok to you?

PR #730, opened 2015-07-21
Fix application finalize race condition in RML

rhc54:
The RML posted recvs are controlled by the async progress thread when in an application process. The call to finalize and close the RML is done from the main thread, and so we need to shift the actual destruct of the posted recv list to the async thread for handling or else we encounter a race condition when accessing the posted recvs.

Thanks to Gilles for providing the required debug info

@ggouaillardet Please see if this solves the problem


ggouaillardet:
:+1:

PR #737, opened 2015-07-22
Add a workaroud for issue in libcuda.so library

rolfv:
We intermittently see that the check for GPU memory returns true when the memory is actually host memory.  This workaround gets us past that until we get a fix in the CUDA library.

@bureddy Can you look at this?  I inadvertently closed #728 so this takes its place.  I made change as requested by @bureddy.  It will not affect anything functionally but I can see how it seems to make sense to put the workaround before the CUDA initialization step.


rolfv:
I made change as requested so I am going to just go ahead and merge now.
Closes #737

hppritcha:
does this need to go in to v2.x?

bureddy:
@rolfv  :+1:

rolfv:
Yes.  Will be submitting for 2.x.  Will soak overnight.

PR #740, opened 2015-07-23
Java Environment Variable Bindings

nrgraham23:
Added the bindings for MPI_GET_VERSION and MPI_GET_LIBRARY_VERSION.

@hppritcha  @jsquyres

### DIFF
--- a/ompi/mpi/java/java/Version.java
+++ b/ompi/mpi/java/java/Version.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ *
+ *
+ * This file is almost a complete re-write for Open MPI compared to the
+ * original mpiJava package. Its license and copyright are listed below.
+ * See <path to ompi/mpi/java/README> for more information.
+ *
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


ompi/mpi/java/java/Version.java
```diff
@@ -0,0 +1,68 @@
...
+
+package mpi;
+
+/**
+ * Version and Subversion for MPI
+ */
+public final class Version
+{
+private final int version;
+private final int subVersion;
```
jsquyres: Minor comment: Seems like weird indenting...?

nrgraham23: It is weird indenting, but I believe it follows the indenting of the rest of the java files.  If you like I can easily change it.

nrgraham23: I could also fix the indenting for all of the other Java files if you
like.  Its very very simple in eclipse.

jsquyres: What's the "Normal" Java way?  To my C-trained eye, this looks weird.  But if it's The Way Java People Do It, it's fine.

If it's not the normal way java people do it, then you might as well keep the indenting here in this PR and make a separate PR for just all the whitespace/indenting changes.  Cool?

nrgraham23: This is not the normal way.  The normal way indents everything inside the class as well.  I will fix the indentation in a separate PR.

hppritcha:
Looks like mlnx jenkins is getting timeouts again on some of its multi-threaded bw tests.

hppritcha:
:+1:

PR #742, opened 2015-07-23
Java bindings for window name get and set

nrgraham23:
Includes bindings for MPI_WIN_GET_NAME and MPI_WIN_SET_NAME. Rather
than call into the C code to set the window name using MPI calls,
we decided it would be better to implement purely in the Java.
The synchronized keyword is used on the get and set methods to ensure
thread safety.

@hppritcha  @jsquyres

### DIFF
--- a/ompi/mpi/java/java/Win.java
+++ b/ompi/mpi/java/java/Win.java
@@ -30,6 +30,7 @@ import java.nio.*;
 public final class Win implements Freeable
 {
 private long handle;
+private String name;
 public static final int WIN_NULL = 0;
 public static final int FLAVOR_PRIVATE = 0;
 public static final int FLAVOR_SHARED = 1;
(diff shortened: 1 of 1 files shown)


jsquyres:
Remind me: are java handles not usable in C or Fortran?  I.e., can you not extract the handle and give it to C?

I ask because if the Java handles can interoperate with the other languages, then you'll need to invoke the underlying C MPI API (or equivalent).

jsquyres:
There's also some MPI semantics about the max length of names, isn't there?

nrgraham23:
There was nothing about a max name length in the area immediately surrounding these methods in the Standard, but I will look around a little more.  However, the handles are used by other languages.  In Java they are stored as longs and get passed into the C to operate on underlying window object.

jsquyres:
Check MPI-3.1 p285 line 39-40:

>  The following functions are used for setting and getting names of windows. The constant MPI_MAX_OBJECT_NAME also applies to these names.

nrgraham23:
Just found that.  I will apply that and go ahead and use the JNI to name the windows.

PR #743, opened 2015-07-24
Java bindings for window name get and set

nrgraham23:
Updated to call to the JNI

Includes bindings for MPI_WIN_GET_NAME and MPI_WIN_SET_NAME.

@jsquyres  @hppritcha


hppritcha:
looks good

PR #745, opened 2015-07-24
Pr/sharedfp sm logic3

edgargabriel:
@jsquyres , I redid the entire branch and pull request and I think I incoroporated all your comments/requests except for the two mutex. TO be more detailed:
 - a separate commit for the free() statement
 - a separate commit for fixing the bug that I found on Wednesday in the unnamed semaphore path
 - renamed the data structure to obey to the Open MPI naming convention
 - redid the configure script to just use one variable.

If you have chance, please have a look at it.

### DIFF
--- a/README
+++ b/README
@@ -921,7 +921,7 @@ NETWORKING SUPPORT / OPTIONS
   if the PSM2 headers and libraries are not in default compiler/linker
   search paths.
 
-  PSM is the support library for Intel Omni-Path network adapters.
+  PSM2 is the support library for Intel Omni-Path network adapters.
 
 --with-psm2-libdir=<directory>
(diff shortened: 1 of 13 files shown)


jsquyres:
@edgargabriel Thanks -- the separate commits made this much easier to review.

FWIW, you could have used the same/existing PR -- just force-push your new/replacement commits to the old PR.  Github will do the Right Thing -- it'll update your PR with the new commits (because the PR is based on the branch, so Github will notice when the branch changes).  Note that this is just about the *only* acceptable case for using a force push.  :-)

edgargabriel:
I pulled in two commits from Jeff, and tested and can confirm that they work correctly. Will merge the pr now.

PR #753, opened 2015-07-27
Standardize verbosity levels

hjelmn:
This is a RFC for defining standard verbosity levels for Open MPI components. The first cut defines the following levels:

```
enum {
    MCA_BASE_VERBOSE_NONE  = 0,
    MCA_BASE_VERBOSE_ERROR = 1,
    MCA_BASE_VERBOSE_WARN  = 10,
    MCA_BASE_VERBOSE_INFO  = 20,
    MCA_BASE_VERBOSE_DEBUG = 40,
    MCA_BASE_VERBOSE_TRACE = 60,
    MCA_BASE_VERBOSE_MAX   = 100,
};
```

Please let me know if I should add/remove any levels or change the mapping between the levels and the corresponding integer value.

I have not made an effort to enforce these levels. Component owners would be expected to modify the verbose output in their components to fit these levels.

For the original RFC please see:

http://www.open-mpi.org/community/lists/devel/2015/05/17468.php

### DIFF
--- a/opal/mca/base/mca_base_var_enum.c
+++ b/opal/mca/base/mca_base_var_enum.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2008-2013 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2012-2014 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2012-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
@@ -23,6 +23,7 @@
 #include "opal_config.h"
 
 #include "opal/mca/base/mca_base_var_enum.h"
+#include "opal/mca/base/base.h"
 
 #include <stdio.h>
 #include <string.h>
@@ -32,6 +33,10 @@ static void mca_base_var_enum_constructor (mca_base_var_enum_t *enumerator);
 static void mca_base_var_enum_destructor (mca_base_var_enum_t *enumerator);
 OBJ_CLASS_INSTANCE(mca_base_var_enum_t, opal_object_t, mca_base_var_enum_constructor,
                    mca_base_var_enum_destructor);
+
+static int enum_dump (mca_base_var_enum_t *self, char **out);
+static int enum_get_count (mca_base_var_enum_t *self, int *count);
+static int enum_get_value (mca_base_var_enum_t *self, int index, int *value, const char **string_value);
 
 static int mca_base_var_enum_bool_get_count (mca_base_var_enum_t *enumerator, int *count)
 {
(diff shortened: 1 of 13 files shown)


hjelmn:
@jsquyres, @rhc54, @ggouaillardet, @goodell 

Anyone have any comments on the suggested standardized levels?

opal/mca/base/mca_base_var_enum.c
```diff
@@ -105,6 +109,126 @@ mca_base_var_enum_t mca_base_var_enum_bool = {
...
+{
+    *count = 5;
+    return OPAL_SUCCESS;
+}
+
+
+static int mca_base_var_enum_verbose_get_value (mca_base_var_enum_t *self, int index,
+                                                int *value, const char **string_value)
+{
+    if (0 > index || 4 < index) {
```
rhc54: Why are these limits hardcoded? I don't think they are even correct, are they? I see 6 valid values in verbose_values, not 5

rhc54:
I find your values a little counter-intuitive - I would have thought that I'd get more verbose as the levels increase. Instead, it goes down with increasing value. Given our opal_output_verbose system, I guess I don't understand that logic.

Or maybe I'm just not understanding their definition? Why do I need a verbosity to be told about an error? Is there some use-case where I wouldn't want to know about an error???

Likewise, I don't really understand how I would use "warn" vs "info". Could you perhaps provide some description as to how you thought these should be used?

ggouaillardet:
@hjelmn 

```
$ ompi_info --all | grep base_verbose
           MCA allocator: parameter "allocator_base_verbose" (current value: "none", data source: default, level: 8 dev/detail, type: int)
```
should this be ```error``` instead of ```none``` ?
or is the ```error``` value somehow inherited and omp_info does not display it ?

some compilers (Fujitsu compiler fwiw) are pretty anal when mixing int and enum....
for example
```
 MCA_BASE_FRAMEWORK_DECLARE(ompi, fcoll, NULL, NULL, NULL, NULL,
                           mca_fcoll_base_static_components, 0);
```
issues a warning, that can be suppressed by doing
```
 MCA_BASE_FRAMEWORK_DECLARE(ompi, fcoll, NULL, NULL, NULL, NULL,
                           mca_fcoll_base_static_components, MCA_BASE_FRAMEWORK_FLAG_DEFAULT);
```
i did not check yet, but this pr might issue extra warnings.


though this is off-topic, i noticed

```
$ ompi_info --all | grep base_verbose
                 MCA mca: parameter "mca_verbose" (current value: "stderr", data source: default, level: 9 dev/all, type: string, deprecated, synonym of: mca_base_verbose)
                 MCA mca: parameter "mca_base_verbose" (current value: "stderr", data source: default, level: 9 dev/all, type: string, synonyms: mca_verbose)
```
1) should we remove (master and v2.x)  the deprecated ```mca_verbose``` param ?
2) ```mca_base_verbose``` definitiion is counter intuitive. i would have expected a verbosity value
(```error```, ```warning```, ...) instead of a location (```stderr```, ```syslog```, ...)

hjelmn:
@rhc54 The levels are in order of increasing verbosity. I would expect the info level would provide basic stuff about things like component selection. At debug there would be more information that would be useful for basic debugging and trace would be everything. An example of what I would expect from trace would be what we get from the openib btl with btl_base_verbose > 0. There is a lot of data about messages sent/received, connections, etc.

@ggouaillardet I was thinking that by default output should be quiet except for critical errors. I agree the definition is counter-intuitive. I will make none = -1 (total silence) and the default (0) be error.

I agree about mca_verbose and mca_base_verbose. The values for those variables is not expected. It might be best to rename them to indicate that they specify an output stream... Speaking of which. Would it be useful to be able to specify an output stream (or file) for a framework? Something like

```
-mca btl_base_verbose_output btl_verbose.%r.log
```

(where %r is replaced by rank)

rhc54:
Okay, that helps - I think of trace, for example, as nothing more than reporting the function that got called. I still wonder why we need an "error" level?

jsquyres:
+1 on everything discussed so far.  I, too, am a bit surprised by the naming / ordering.  Here's my take on what they should be (I'd also put short comments next to exact one describing the intent for each one):

```c
enum {
    // total silence
    MCA_BASE_VERBOSE_NONE  = -1,

    // only errors are output
    MCA_BASE_VERBOSE_ERROR = 0,

    // also emit warnings
    MCA_BASE_VERBOSE_WARN  = 20,

    // also emit general, user-relevant information, such as rationale as to why certain choices
    // or code paths were taken, information gleaned from probing the local system, etc.
    MCA_BASE_VERBOSE_INFO  = 40,

    // also emit relevant tracing information (e.g., which functions were invoked / 
    // call stack entry/exit info)
    MCA_BASE_VERBOSE_TRACE = 60,

    // also emit Open MPI-developer-level (i.e,. highly detailed) information
    MCA_BASE_VERBOSE_DEBUG = 80,

    // also output anything else that might be useful
    MCA_BASE_VERBOSE_MAX   = 100,
};
```

hjelmn:
@jsquyres The levels were based off suggestions on the mailing list. The levels you suggest look better so I will update the PR to use those levels.

hjelmn:
If there are no objections I think this may be ready to pull.

hjelmn:
Rebased this into 3 commits.

gpaulsen:
Why not have these values be a bitmask.  so NONE == 0 , only Errors == 1, etc.  Then if multiple different verbosity are wanted, users could binary or them together... and each message would only need to check for if the appropriate bit is set or not, and not do < or > comparisons.

hjelmn:
@gpaulsen That would be a change from the current verbosity behavior. Currently as the level goes up only additional output is printed. This PR is an attempt to name the levels and give a description of what is expected if the verbosity equals or exceeds that level.

It would be a bit more work to change the verbosity into a bitmask. All component would have to be modified and the opal_output code changed as well. It might be a worthwhile task if there is interest in that sort of change.

hjelmn:
It would also change how the verbosity enumerator is set up. If we used a bitmask it would be great if this worked: ```-mca blah_base_verbose "errors|warn|trace"```.

jsquyres:
@gpaulsen Just to be clear: that sounds like a fine idea.  But I think Nathan's proposed change is much smaller than that.  We have a long-established history of verbosity levels of [0, 100].  Nathan's change basically puts some standardization on what those numbers mean -- nothing more.

I think the idea of moving to a bitmask type of output system would be a bit more, and would likely be worthy of its own discussion.

gpaulsen:
I see.  just wanted to comment before we make all components change, and then change again.

hjelmn:
If we go to a bitmap component shouldn't need to change again since the logic will be down in opal_output.

hjelmn:
I should mention that as part of the PR we request component owners to use the predefined levels. If we go to a bit map in the future it will become a required change.

gpaulsen:
okay, thanks for discussion.

ggouaillardet:
imho, if we decide to move to a bitmap, i'd rather have "everything" in the enum
for example, ERROR=0x1  WARN=0x3 B_ERROR=0x1 B_WARN=0x2 ...
i do not believe end users will use B_ because it is generally way too fine grained,
but giving the possibility of using such fine grain is an interesting idea

Issue #158, opened 2014-10-01
ROMIO:  MPI_MODE_EXCL, Lustre striping


ompiteam:
Arguments passed to MPI_File_open() include
- a path on a Lustre filesystem
- a file mode containing the (MPI_MODE_CREATE | MPI_MODE_EXCL) flags
- striping parameters in the MPI_Info (e.g. "striping_factor" or "striping_unit")

The expected behavior for a non-existant path would be for the file to be created with the provided Lustre striping properties.  Instead, the file is created but MPI_File_open() returns MPI_IO_ERR (35) and does not return a file handle.

The problem stems from ADIOI_LUSTRE_SetInfo()'s pre-creating the file if striping parameters are present in the initial MPI_Info (ca. ad_lustre_hints.c:92) and then closing the descriptor so that ADIOI_LUSTRE_Open() can re-open the file.  However, ADIOI_LUSTRE_SetInfo() does not remove the MPI_MODE_EXCL flag from the ADIO_File's access_mode bit vector, so when ADIOI_LUSTRE_Open() is called it will always fail to open the file since it now exists.

Two possible solutions:

(1) Remove the MPI_MODE_EXCL flag from access_mode on successful file creation in ADIOI_LUSTRE_SetInfo(); once the file is created, all further application of this flag to the ADIO_File will cause failure.

(2) Do not close() the file descriptor open()'ed by ADIOI_LUSTRE_SetInfo(); modify ADIOI_LUSTRE_Open() to check for fd_sys already set and bypass the open() call if so.

roblatham00:
> Has this bug been reported upstream to MPICH/ROMIO?

nope.  Or perhaps, "not at that time" but now it has.  I've opened http://trac.mpich.org/projects/mpich/ticket/226

hjelmn:
I think Ron means http://trac.mpich.org/projects/mpich/ticket/2261

dshrader:
I have run in to similar behavior and Nathan asked me to comment on it here. Basically, the problem is that whenever  MPI_MODE_EXCL is used with a panasas target file, a non-zero return status is returned from MPI_File_open. However, the file is created on the panasas volume but has size zero. If MPI_MODE_EXCL is not used, the file opens just fine. I am using Open MPI 1.8.5.

So, this works: 
ret = MPI_File_open(MPI_COMM_WORLD, file, MPI_MODE_CREATE|MPI_MODE_WRONLY, MPI_INFO_NULL, &fh);

This does not:
[...]

PR #757, opened 2015-07-28
hint processing should not open files

roblatham00:
move opening of files from hint processing and into open routines.

This is MPICH commit 92f1c69f0de8 and 22a77dceda11

more information at https://trac.mpich.org/projects/mpich/ticket/2261
Ref: https://github.com/open-mpi/ompi/issues/158
closes openmpi  #158 
Signed-off-by: Pavan Balaji <balaji@anl.gov>

### DIFF
--- a/ompi/mca/io/romio314/romio/adio/ad_lustre/ad_lustre_hints.c
+++ b/ompi/mca/io/romio314/romio/adio/ad_lustre/ad_lustre_hints.c
@@ -20,8 +20,7 @@ void ADIOI_LUSTRE_SetInfo(ADIO_File fd, MPI_Info users_info, int *error_code)
     char *value;
     int flag;
     ADIO_Offset stripe_val[3], str_factor = -1, str_unit=0, start_iodev=-1;
-    struct lov_user_md lum = { 0 };
-    int err, myrank, fd_sys, perm, amode, old_mask;
+    int err, myrank;
     static char myname[] = "ADIOI_LUSTRE_SETINFO";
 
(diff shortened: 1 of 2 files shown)


goodell:
Hey Rob, thanks for the PR, since I know you're under no obligation to help maintain ROMIO in OMPI.  I'm going to let people who actually deal with MPI-IO in OMPI look at this, but I have one minor nit:

> Ref: #2261

This breadcrumb should probably be removed from the commit before it gets merged, since it refers to an MPICH Trac ticket number, not an OMPI github issue number.  At some point when OMPI hits >2261 PRs/issues, this breadcrumb will potentially become confusing.

roblatham00:
In this case the motivation is more selfish than altruistic: a Darshan user
using OpenMPI couldn't create log files because of this.  Editing the
breadcrumb makes sense (hope github really did edit it...), and I'll try to
remember to namespace these identifiers in the future.

hppritcha:
I'd prefer to merge this PR in after we've fixed some lustre problems in master.  The problem results both in ompi io and romio components failing on all but trivial tests when using lustre.  Once we get lustre working again and can double check this PR, I'm okay with merging it.

goodell:
> (hope github really did edit it...)

Looks like you successfully edited the PR description, but never force-pushed an updated commit with a fixed commit message.

hppritcha:
I tried this out on a lustre file system now that ompi has gotten a visit from the lustre medic.  Works for me.
:+1:

PR #761, opened 2015-07-29
Fix warnings in direct (pml-cm,mtl-ofi) build

jithinjosepkl:
Fix implicit function declaration warnings in direct (pml-cm,mtl-ofi) build

Signed-off-by: Jithin Jose <jithin.jose@intel.com>


rhc54:
:+1:

PR #762, opened 2015-07-30
Fix an invalid memory access in mrecv & imrecv

nysal:
After the call to PML mrecv/imrecv, the message handle
is set to MPI_MESSAGE_NULL. Use the cached communicator
while invoking the error handler.


hjelmn:
This probably should be PRd to 1.8, 1.10 and 2.x.

rhc54:
:+1:  I will PR to 1.8 as I'm planning to release it tomorrow

PR #765, opened 2015-07-31
usnic libfabric 1.1 configury update

jsquyres:
Ensure that we have libfabric >= v1.1

@goodell Assuming https://github.com/ofiwg/libfabric/pull/1187 is merged into libfabric before v1.1, please review.

### DIFF
--- a/opal/mca/btl/usnic/configure.m4
+++ b/opal/mca/btl/usnic/configure.m4
@@ -102,6 +102,26 @@ AC_DEFUN([_OPAL_BTL_USNIC_DO_CONFIG],[
            AC_MSG_RESULT([$opal_btl_usnic_happy])
           ])
 
+    # The usnic BTL requires at least libfabric v1.1 (there was a
+    # critical bug in libfabric v1.0).
+    AS_IF([test "$opal_btl_usnic_happy" = "yes"],
+          [AC_MSG_CHECKING([whether libfabric is >= v1.1])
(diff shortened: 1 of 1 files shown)


jsquyres:
Sweet -- @shefty accepted ofiwg/libfabric#1187.  So this should be good to go.

opal/mca/btl/usnic/configure.m4
```diff
@@ -102,6 +102,26 @@ AC_DEFUN([_OPAL_BTL_USNIC_DO_CONFIG],[
            AC_MSG_RESULT([$opal_btl_usnic_happy])
           ])
 
+    # The usnic BTL requires at least libfabric v1.1 (there was a
+    # critical libfabric v1.0).
```
goodell: this is probably missing the word "bug", but it's not a major problem

goodell:
:+1: 

The direct comparison seems fine to me.  It makes me wonder why we have `FI_VERSION_LT`/`FI_VERSION_GE` at all...

I'm not going to merge this right now in case you want to fix up the typo.

shefty:
> The direct comparison seems fine to me. It makes me wonder why we have
> FI_VERSION_LT/FI_VERSION_GE at all...

The library itself uses these, and they seem useful enough that they are also provided to apps.

lanl-ompi:
Test FAILed.

PR #771, opened 2015-08-04
opal/lifo: add missing opal_atomic_wmb and remove unnecessary opal_atomic_rmb

hjelmn:


hjelmn:
@nysal That should fix the crash on ppc. Let me know if it works.

nysal:
Yes, this works for me. Thanks for the patch Nathan.

Issue #776, opened 2015-08-05
Improve CUDA extension code


rolfv:
Need to make at least two more changes to the extension code.
1. Have them built unconditionally so they are always available.
2. Have them then properly set the MPIX_CUDA_AWARE_MACRO to 0 or 1 depending on whether library has CUDA-aware configured in. Also have the MPIX_Query_cuda_support() function do the right thing.

The problem is now they are only set when we have CUDA-aware so to use them requires a bunch of ifdefs.

rolfv:
bot:milestone:v2.0.0
bot:label:bug
bot:assign: @rolfv

rolfv:
Duplicating bug that I put in wrong place.

rolfv:
This is now in master and v2.0.0.  Closinig.

PR #777, opened 2015-08-05
Build always and fix return value

rolfv:
This change makes the CUDA extensions always build so they are always available whether we configure with CUDA support or not.  That is needed as this extension is being used to determine whether we have CUDA aware support or not.

@jsquyres Can you review?  This is fixing bug #776

### DIFF
--- a/ompi/mpiext/cuda/c/mpiext_cuda.c
+++ b/ompi/mpiext/cuda/c/mpiext_cuda.c
@@ -23,9 +23,9 @@
 #include "opal/constants.h"
 #include "ompi/mpiext/cuda/c/mpiext_cuda_c.h"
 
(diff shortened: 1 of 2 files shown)


ompi/mpiext/cuda/c/mpiext_cuda.c
```diff
@@ -27,5 +27,5 @@
    support.  We may expand on this API to return more features in the future. */
```
jsquyres: @rolfv You need to change this comment, too.

rolfv:
Fixed comment as requested by reviewer.

jsquyres:
Can you squash down to one commit?  :smile:

jsquyres:
Poifect. :+1:

PR #779, opened 2015-08-05
Topic/fcoll fixes

edgargabriel:
@ggouaillardet I included your fix for lockedfile sharedfp component, and provided the other two fixes required. This way, the selection logic for individual remains unchanged. Could you please have a look, it fixes your test case on my machines. Thanks!

@jsquyres thanks for helping with extracting the correct patch from @ggouaillardet branch, and remove the one that we do not need.


edgargabriel:
@ggouaillardet I actually wanted to thank, it is very valuable that somebody else also looks at the code and runs the IO tests. You are also btw. right that I broke something in mpi_test_suite that used to pass, in one of my optimizations that I introduced a couple of days back. I am working on that right now.

PR #782, opened 2015-08-07
btl/openib: fix the detection of the running ofed version.

ggouaillardet:


opal/mca/btl/openib/btl_openib_xrc.c
```diff
@@ -43,22 +43,22 @@ OBJ_CLASS_INSTANCE(ib_address_t,
 bool mca_btl_openib_xrc_check_api()
 {
     void *lib = dlopen(NULL, RTLD_NOW); /* current program */
+    void *sym;
     if (!lib) {
         BTL_ERROR(("XRC error: could not find XRC API version"));
         return false;
     }
 
+    sym = dlsym(lib, OPAL_VERSIONNED_XRC_SYMBOL);
```
jsquyres: You should absolutely not be using dlopen() / dlsym() directly.  Please use opal_dl_open() and friends (see `opal/mca/dl/base/base.h`).

jsquyres:
@ggouaillardet Is this PR now moot / should be closed?

gpaulsen:
Can you give some more details why you think this is moot?
This would be needed to support XRC (any version) correct?

jsquyres:
Were you not just on the call where we discussed this?  :-)

I am pretty sure @ggouaillardet has ripped out the code that this PR is part of.

gpaulsen:
Yeah, I didn't / don't understand.  I'm pretty slow.  :)

I understand not doing dlopen/dlsym directly.  But has XRC support in general been ripped out of openib?

ggouaillardet:
@gpaulsen XRC has never been removed from ompi
the XRC api changed from OFED 3.12 (XRC Domains are used instead of legacy XRC)
the goal of the test was to figure out whether the OFED version at runtime is compatible with the OFED version that was used at compile time, and gently disable the openib btl if not.
by removing the test, the openib btl is always enabled, and it might fail with a more or less obscure message if OFED versions mismatch.
XRC has been disabled for a while with OFED < 3.12 because of an other bug in configury, so i am thinking this test never really worked (fail to detect mismatch and false positive)
i cannot find a robust way to test OFED version compatibility (and that also works with static libs and/or --disable-dlopen) so i gave up and simply removed this test

@jsquyres yes, the PR is moot, so i am closing it

Issue #627, opened 2015-06-05
cuda & MPI_THREAD_MULTIPLE


dribbroc:
This maybe not a real issue, but at least unexpected behaviour.
Testet with cuda 7 and openmpi 1.8.5.

Background: If using hosts with only on gpu equipped, you ever hardly use the cudaSetDevice statement in your code, as all cuda calls default to the only existing device (0).

When allocating cuda device memory in the host thread and calling MPI_Send / Recv from other threads, mpi crashes with CUDA: Error in cuMemcpy.

A "workaround" is, to set the cudaDevice in every thread explitly, which definitly is a good advise at all.

Nevertheless, the cuda documentation states, that every thread's cudaDevice defaults to device 0, if not changed by the user.
This statement holds true, for cuda api calls in different threads, but openmpi seems to do some dark magic under the hood, changing this default device.

rolfv:
Can you provide a test program?  There is some special code in Open MPI for handling when there are threads to make sure we have a context for the threads.

dribbroc:
Testcase & output: https://gist.github.com/dribbroc/86422b72afe9c61b59d1

compiled with
mpic++ -std=c++11 -o testcase testcase.cpp -L/sfw/cuda/7.0/lib64 -lcudart -I/sfw/cuda/7.0/include

with openmpi 1.8.5, cuda 7.0, gcc 5.1

https://gist.github.com/dribbroc/86422b72afe9c61b59d1#file-testcase-cpp-L7
activating Line 7 resolves the issue

rolfv:
I know what the issue is.  This does not exist in CUDA 6.5 as I handle it properly.  I will look at how to fix this.

dribbroc:
Speaking of cuda 6.5 in general and the nvidia jetson k1 board in particular, this little test codes gets even more buggier.
(ompi 1.8.5, gcc 4.8.2, linux4tegra 21.3)

If I make the communication call synchronous (removing all threading stuff), see https://gist.github.com/dribbroc/86422b72afe9c61b59d1#file-testcase2-cpp ,then i get the following error on my jetson board:
https://gist.github.com/dribbroc/86422b72afe9c61b59d1#file-output2

dribbroc:
Further note to make things more clear:
https://github.com/open-mpi/ompi/issues/286 arose only at program exit.
Now the same issue arises during the program execution, when there are plenty of things to do.

rolfv:
[...]

PR #783, opened 2015-08-07
Refs open-mpi/ompi#627. Fix support for multi-threads with CUDA 7.0

rolfv:
The support for threads and CUDA 7.0 was broken.  This fixes it.  We still need to check for the current context and set it if it is not set.

@jsquyres Can you review?  I know you have no expertise here, but just look at the logic.  There are slightly different code paths when we can get multiple attributes and when we cannot.  Prior to this change, we did not check for current context when we had multiple attribute support.  It turns out we have to regardless.

### DIFF
--- a/opal/mca/common/cuda/common_cuda.c
+++ b/opal/mca/common/cuda/common_cuda.c
@@ -1739,19 +1739,19 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t 
     int res;
     CUmemorytype memType = 0;
     CUdeviceptr dbuf = (CUdeviceptr)pUserBuf;
-    CUcontext ctx = NULL;
+    CUcontext ctx = NULL, memCtx = NULL;
 #if OPAL_CUDA_GET_ATTRIBUTES
     uint32_t isManaged = 0;
     /* With CUDA 7.0, we can get multiple attributes with a single call */
     CUpointer_attribute attributes[3] = {CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
                                          CU_POINTER_ATTRIBUTE_CONTEXT,
                                          CU_POINTER_ATTRIBUTE_IS_MANAGED};
-    void *attrdata[] = {(void *)&memType, (void *)&ctx, (void *)&isManaged};
+    void *attrdata[] = {(void *)&memType, (void *)&memCtx, (void *)&isManaged};
 
     res = cuFunc.cuPointerGetAttributes(3, attributes, attrdata, dbuf);
     OPAL_OUTPUT_VERBOSE((101, mca_common_cuda_output,
-                        "dbuf=%p, memType=%d, ctx=%p, isManaged=%d, res=%d",
-                         (void *)dbuf, (int)memType, (void *)ctx, isManaged, res));
+                        "dbuf=%p, memType=%d, memCtx=%p, isManaged=%d, res=%d",
+                         (void *)dbuf, (int)memType, (void *)memCtx, isManaged, res));
 
     /* Mark unified memory buffers with a flag.  This will allow all unified
      * memory to be forced through host buffers.  Note that this memory can
@@ -1787,6 +1787,7 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t 
     }
     /* Must be a device pointer */
     assert(memType == CU_MEMORYTYPE_DEVICE);
+#endif /* OPAL_CUDA_GET_ATTRIBUTES */
 
     /* This piece of code was added in to handle in a case involving
      * OMP threads.  The user had initialized CUDA and then spawned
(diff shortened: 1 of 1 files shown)


opal/mca/common/cuda/common_cuda.c
```diff
@@ -1739,19 +1739,19 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t
...
     /* With CUDA 7.0, we can get multiple attributes with a single call */
     CUpointer_attribute attributes[3] = {CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
                                          CU_POINTER_ATTRIBUTE_CONTEXT,
                                          CU_POINTER_ATTRIBUTE_IS_MANAGED};
-    void *attrdata[] = {(void *)&memType, (void *)&ctx, (void *)&isManaged};
+    void *attrdata[] = {(void *)&memType, (void *)&memCtx, (void *)&isManaged};
 
     res = cuFunc.cuPointerGetAttributes(3, attributes, attrdata, dbuf);
     OPAL_OUTPUT_VERBOSE((101, mca_common_cuda_output,
                         "dbuf=%p, memType=%d, ctx=%p, isManaged=%d, res=%d",
```
jsquyres: Do you want to change this VERBOSE output to say `memCtx` to match the variable name?  I ask because you're making a distinction between `ctx` and `memCtx` in the code, so since this is assumedly a debugging output, it might be good to have the output match the variable name.

opal/mca/common/cuda/common_cuda.c
```diff
@@ -1797,25 +1798,25 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t
      * and set the current context to that.  It is rare that we will not
      * have a context. */
     res = cuFunc.cuCtxGetCurrent(&ctx);
-#endif /* OPAL_CUDA_GET_ATTRIBUTES */
     if (OPAL_UNLIKELY(NULL == ctx)) {
         if (CUDA_SUCCESS == res) {
-            res = cuFunc.cuPointerGetAttribute(&ctx,
+#if !OPAL_CUDA_GET_ATTRIBUTES
+            res = cuFunc.cuPointerGetAttribute(&memCtx,
```
jsquyres: Does this overwrite the value of `memCtx`?  If so, what's the point of filling it in lines 1749 and 1754?  I ask because you specifically changed lines 1749 and 1754 to use the new variable `memCtx`...

More generally, after reading this patch a few times, I'm not sure what the difference is between `ctx` and `memCtx` -- it seems like they're used mutually exclusively, and therefore you don't really need 2 variables.  I don't really care, mind you -- this is your code, and if you want 2 variables just to make the code more readable, that's cool. :smile:  But I figured I'd ask: is there a reason to have 2 variables for the code logic?  Or is it just to make the code more readable?

opal/mca/common/cuda/common_cuda.c
```diff
@@ -1797,25 +1798,25 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t
...
                                                CU_POINTER_ATTRIBUTE_CONTEXT, dbuf);
             if (res != CUDA_SUCCESS) {
                 opal_output(0, "CUDA: error calling cuPointerGetAttribute: "
                             "res=%d, ptr=%p aborting...", res, pUserBuf);
                 return OPAL_ERROR;
+            }
+#endif /* OPAL_CUDA_GET_ATTRIBUTES */
+            res = cuFunc.cuCtxSetCurrent(memCtx);
+            if (res != CUDA_SUCCESS) {
```
jsquyres: Do you want to put an `OPAL_UNLIKELY` around this?  (and all the "if" blocks in this function, for that matter...?)

I ask because I'm guessing this is in the critical performance code path...?

opal/mca/common/cuda/common_cuda.c
```diff
@@ -1797,25 +1798,25 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t
...
-                res = cuFunc.cuCtxSetCurrent(ctx);
-                if (res != CUDA_SUCCESS) {
-                    opal_output(0, "CUDA: error calling cuCtxSetCurrent: "
-                                "res=%d, ptr=%p aborting...", res, pUserBuf);
-                    return OPAL_ERROR;
-                } else {
-                    opal_output_verbose(10, mca_common_cuda_output,
-                                        "CUDA: cuCtxSetCurrent passed: ptr=%p", pUserBuf);
-                }
+                opal_output_verbose(10, mca_common_cuda_output,
```
jsquyres: Also working on my (possibly incorrect!) assumption that this is in the critical code performance path: you might not want to call `opal_output_verbose()` here -- it's not exactly a fast function (even if the verbose level means that it won't get output).

opal/mca/common/cuda/common_cuda.c
rolfv: Yes, I will make that change.  Thanks.

opal/mca/common/cuda/common_cuda.c
rolfv: They are two different code paths based on OPAL_CUDA_GET_ATTRIBUTES.  With OPAL_CUDA_GET_ATTRIBUTES set, we fill in the memCtx variable with a single call that is also filling in two other pieces of information.  Without OPAL_CUDA_GET_ATTRIBUTES, we have to make a specific call for memCtx.  Also, I need the two variables because when I do have OPAL_CUDA_GET_ATTRIBUTES, I have previously saved the memCtx variable.  I agree that when we do not have OPAL_CUDA_GET_ATTRIBUTES, we do not need memCtx.  But, that is what this bug is fixing.  Making things work with OPAL_CUDA_GET_ATTRIBUTES set.

So, the general idea is that we put the memory context into memCtx and the current context into ctx. Does that make sense?

jsquyres: Cool.

rolfv:
Fixed the issues and pushed the changes.  Do you want to look once more Jeff?  If not, then I will just merge when all the automatic tests complete.  Thanks Jeff for the review.

PR #784, opened 2015-08-07
Update the opal_progress_thread API

jsquyres:
@rhc54 Per our discussion this morning, have a look.  The diff might be hard to read; it may be easier to read the final opal_progress_thread.h and .c.


rhc54:
@jsquyres would it make sense to simplify a bit further and say that a NULL passed to opal_progress_thread_<foo> APIs equates to "use the default async progress thread"? Then you can static the name and nobody needs to know it.

jsquyres:
@rhc54 Oh ya, that's better.  Let me adjust.

jsquyres:
@rhc54 New commits pushed: NULL now means "The OPAL-wide async progress thread."

rhc54:
:+1:

PR #786, opened 2015-08-07
Add explicit implementation of Cloneable

nrgraham23:
Added Cloneable to the implemented interface list as per
Coverity suggestion.  The required methods were already
implemented, but it was not explicitly stated.  This is
an intent revealing change.

Coverity pointed out this lapse.

@hppritcha  @jsquyres


hppritcha:
looks good.

PR #787, opened 2015-08-07
remove a erroneous parenthesis which prevents the compilation of the …

edgargabriel:
…lustre adio
@roblatham00 would you mind confirming that this is really correct? Not sure how the parenthesis got in in the first place, but it didn't compile on the Stuttgart Cray originally, but does now.


roblatham00:
I had to massage my patch quite a bit to apply to OpenMPI and apparently missed this brace.  thanks for cleaning up my mess.  There's no bogus brace on the MPICH side of things.

Issue #785, opened 2015-08-07
OFI MTL being auto-selected


jsquyres:
If there are no other available MTLs, the OFI MTL with the sockets provider will be selected over OB1 (and any BTLs).

This is very undesirable...

@yburette Can you have a look?

yburette:
Yes, I bet we will get the same issue with other providers such as MXM, right?

There are two priorities schemes here:
1) between MTLs (e.g. PSM vs OFI MTLs)
2) between PMLs (e.g. OB1 vs CM PMLs)

The OFI MTL doesn't have any control on 2). CM has higher priority over OB1 if any MTL can work.
I guess the only way to fix this is to ignore the sockets OFI provider unless explicitly requested by the user with "--mca mtl_ofi_provider sockets".

@jsquyres Would this work?

jsquyres:
Yes, this is probably a good idea -- I think the sockets OFI provider is really only intended for debugging/testing.  It's not necessarily going to give higher performance than the TCP BTL, right?

yburette:
@jsquyres You're absolutely right, the sockets provider is mostly for testing.

I'll create a PR shortly.

jsquyres:
This should be fixed by #788.

PR #788, opened 2015-08-07
mtl/ofi: Deprioritize some OFI providers.

yburette:
Some OFI providers such as "sockets" are used for debugging
purposes mostly. For these providers, other components usually
offer better performance -- e.g. for sockets, the BTL/TCP would
be a better choice.
Thus, we chose to ignore some providers unless explicitly asked
by the user on the command line:

e.g. --mca mtl_ofi_provider sockets

This is to fix issue #785

@jsquyres @hppritcha @jithinjosepkl Please review

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -107,6 +107,29 @@ ompi_mtl_ofi_progress_no_inline(void)
 	return ompi_mtl_ofi_progress();
 }
 
+static struct fi_info*
+select_ofi_provider(struct fi_info *providers)
+{
+    struct fi_info *prov = providers;
+    /**
+     * Unless explicitly asked by user, ignore the following OFI providers:
+     *     - sockets
+     *     - mxm
+     */
+    if (NULL == ompi_mtl_ofi.provider_name) {
+        while (NULL != prov) {
+            if ((0 == strncmp(prov->fabric_attr->prov_name, "sockets", 7)) ||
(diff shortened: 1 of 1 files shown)


jsquyres:
This is a good step; thanks!

I think I *might* have mentioned this a long time ago... but should there be `mca_mtl_ofi_provider_include`  and `mca_mtl_ofi_provider_exclude` MCA params?  This would make it similar to other MCA params in OMPI.  Specifically:

* a user can set one or the other of these two MCA params -- but it's an error if they set both
* if the user sets `_include`, it's a comma-delimited list of providers that are eligible to be used
* if the user sets `_exclude`, it's a comma-delimited list of providers that are *not* eligible to use (all others are eligible)
  * note that OMPI has utility routines like `opal_argv_split()` to parse comma-delimited lists
* note, too, that OMPI's MCA params can distinguish between user-set values and default values
* hence, you could set the default value of the `_exclude` param to be `mxm,sockets`, which would exclude those two providers by default.  But the user can override this behavior by setting either of the two MCW params with any of the normal MCA mechanisms (CLI, env var, file).

hppritcha:
this looks good to me.

I think @jsquyres suggestion for exclude/include is a good one but could be part of another PR.

yburette:
@jsquyres Yes, you did mention it before. But it requires us to make a lot more changes. I've started working on it. Hopefully, this PR offers a working solution in the meantime.

jithinjosepkl:
:+1:

jsquyres:
Sure -- if that change can come in eventually (the include/exclude solution), and since this one fixes an immediate issue, this PR can come in.

PR #791, opened 2015-08-10
usnic: move cchecker to OPAL-wide progress thread

jsquyres:
There's no longer any need for the usnic BTL to have its own progress thread: it can use the opal_progress_thread() infrastructure.  This commit removes the code to startup/shutdown the usnic-BTL-specific progress thread and instead, just adds its events to the OPAL-wide progress thread.

This necessitated a small change in the finalization step.  Previously, we would stop the progress thread and then tear down the events.  We can no longer stop the progress thread, and if we start tearing down events, this will cause shutdown/hangups to be sent across sockets, potentially firing some of the still-remaining events while some (but not all) of the data structures have been torn down. Chaos ensues.

Instead, queue up an event to tear down all the pending events.  Since the progress thread will only fire one event at a time, having a teardown event means that it can tear down all the pending events "atomically" and not have to worry that one of those events will get fired in the middle of the teardown process.

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic.h
+++ b/opal/mca/btl/usnic/btl_usnic.h
@@ -226,6 +226,9 @@ typedef struct opal_btl_usnic_component_t {
         / API >=v1.1, this is the endpoint.msg_prefix_size (i.e.,
         component.transport_header_len). */
     uint32_t prefix_send_offset;
+
+    /* OPAL async progress event base */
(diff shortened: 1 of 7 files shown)


jsquyres:
Added a second commit that fixes a race condition seen in rare MTT runs: the connectivity checker agent shuts down before one or more clients has a chance to connect.

PR #794, opened 2015-08-10
Java null handle bugfix

nrgraham23:
A helper method in Request.java could cause a crash
if the request array that was passed contained nulls.

@hppritcha @jsquyres


hppritcha:
:+1:
please push to v2.x as well

Issue #793, opened 2015-08-10
Add mca_mtl_provider_include|exclude MCA vars


jsquyres:
Per #788, add `mca_mtl_ofi_provider_include` and `mca_mtl_ofi_provider_exclude` MCA params. This would make it similar to other MCA params in OMPI.  Specifically:

* a user can set one or the other of these two MCA params -- but it's an error if they set both
* if the user sets `_include`, it's a comma-delimited list of providers that are eligible to be used
* if the user sets `_exclude`, it's a comma-delimited list of providers that are *not* eligible to use (all others are eligible)
  * note that OMPI has utility routines like `opal_argv_split()` to parse comma-delimited lists
* note, too, that OMPI's MCA params can distinguish between user-set values and default values
* hence, you could set the default value of the `_exclude` param to be `mxm,sockets`, which would exclude those two providers by default.  But the user can override this behavior by setting either of the two MCW params with any of the normal MCA mechanisms (CLI, env var, file).

@yburette @jithinjosepkl

yburette:
PR #795 is an attempt at fixing this.

jsquyres:
This is now done.

PR #795, opened 2015-08-11
mtl/ofi: add include/exclude list MCA vars.

yburette:
mtl_ofi_provider_include (resp. mtl_ofi_provider_exclude) can be used
to specify which provider(s) the OFI MTL can select (resp. ignore).

e.g. --mca mtl_ofi_provider_include "psm,sockets"

By default, mtl_ofi_provider_exclude is set to "sockets,mxm".

This deprecates the old MCA var named "mtl_ofi_provider".

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -24,6 +24,8 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                             bool enable_mpi_threads);
 
 static int param_priority;
+static char *prov_include;
+static char *prov_exclude;
 
 mca_mtl_ofi_component_t mca_mtl_ofi_component = {
     {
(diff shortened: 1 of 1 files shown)


yburette:
Should fix #793

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -24,6 +24,8 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                             bool enable_mpi_threads);
 
 static int param_priority;
+static char *prov_include;
+static char *prov_exclude;
```
jsquyres: Does the default value of `prov_exclude` need to be set to `sockets,mxm`?

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -107,26 +135,70 @@ ompi_mtl_ofi_progress_no_inline(void)
...
+        }
+    } else if (NULL != prov_exclude) {
+        exclude_list = opal_argv_split(prov_exclude, ',');
+        while ((NULL != prov) &&
+               (is_in_list(exclude_list, prov->fabric_attr->prov_name))) {
+            opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                                "%s:%d: mtl:ofi: \"%s\" in exclude list\n",
+                                __FILE__, __LINE__,
+                                prov->fabric_attr->prov_name);
+            prov = prov->next;
```
jsquyres: If I'm reading the above logic correctly, this uses the *first* provider that it finds in the include list (or *first* provider that is *not* in the exclude list).  Is that correct?

If so, that doesn't agree with the help messages -- it implies that *all* included / non-excluded providers will be used.

jsquyres:
@yburette Other than 2 minor comments, it looks good.  If you can fix those, go ahead and merge.

Thank you!

ompi/mca/mtl/ofi/mtl_ofi_component.c
yburette: This is done below (new line 75) in the same manner as param_priority.
Do you want me to move it to the declaration line?

jsquyres: Nope, I just totally missed it.  :-)

ompi/mca/mtl/ofi/mtl_ofi_component.c
yburette: How about changing the help message for provider_include to the following?

"Comma-delimited list of OFI providers which can be used (e.g. \"psm,sockets\"...."

Is this a more accurate wording?

jsquyres: Ya, something like that.  Perhaps:

> "Comma-delimited list of OFI providers that are considered for use (e.g., \"psm,sockets\"; an empty value means that all providers will be considered). Mutually exclusive with mtl_ofi_provider_exclude."

(note the missing comma after "e.g.," in the original -- sorry, I have a BA in English Literature... :smile: )

yburette: Very good, I'll update it.

(I didn't know about the comma. Thank you for pointing it out. -- As you can guess, English is not my native language :-P)

yburette:
I verified that it worked for several values of mtl_ofi_provider_include and mtl_ofi_provider_exclude with a libfabric that was built with PSM and sockets providers.

:+1:

rhc54:
@yburette Is this coming to 1.10 prior to release?

yburette:
@rhc54 That would be ideal.

PR #796, opened 2015-08-11
Topic/hcoll config

ggouaillardet:


ggouaillardet:
@miked-mellanox could you please review this ?
this was initially discussed on the ompi user mailing list at http://www.open-mpi.org/community/lists/users/2015/08/27418.php

config/ompi_check_libfca.m4
```diff
@@ -25,20 +25,17 @@ AC_DEFUN([OMPI_CHECK_FCA],[
     OPAL_CHECK_WITHDIR([fca], [$with_fca], [lib/libfca.so])
 
     AS_IF([test "$with_fca" != "no"],
-          [AS_IF([test ! -z "$with_fca" && test "$with_fca" != "yes"],
+          [ompi_check_fca_libs=fca
+           AS_IF([test ! -z "$with_fca" && test "$with_fca" != "yes"],
 			  [ompi_check_fca_dir=$with_fca
```
jsquyres: `$with_fca` might equal `no` at this point (if the user specified `--without-fca`).

config/ompi_check_libfca.m4
```diff
@@ -58,7 +55,16 @@ AC_DEFUN([OMPI_CHECK_FCA],[
 			   CPPFLAGS=$CPPFLAGS_save
 			   LDFLAGS=$LDFLAGS_save
 			   LIBS=$LIBS_save],
-			   [ompi_check_fca_happy="no"])
+			  [AC_SUBST([coll_fca_HOME], "/")
```
jsquyres: I don't quite understand the purpose of this check -- is it to catch the `--with-hca` (with no value) case?

If so, it might be better (and more in line with other configure.m4 scripts, I think?) to separate the user's desire to have hcoll support vs. whether or not they supplied a directory (i.e., use two different variables).  And then only have one call to `OPAL_CHECK_PACKAGE`.

config/ompi_check_libhcoll.m4
```diff
@@ -59,7 +52,15 @@ AC_DEFUN([OMPI_CHECK_HCOLL],[
 			   CPPFLAGS=$CPPFLAGS_save
 			   LDFLAGS=$LDFLAGS_save
 			   LIBS=$LIBS_save],
-			   [ompi_check_hcoll_happy="no"])
+			  [OPAL_CHECK_PACKAGE([$1],
```
jsquyres: The same kind of comments that I made in the fca configure.m4 (above) also apply here.

ggouaillardet:
@jsquyres got it, will do

ggouaillardet:
@jsquyres i updated this PR and made 3 for the ompi-release branches

bureddy:
:+1:

config/ompi_check_libfca.m4
```diff
@@ -25,41 +25,37 @@ AC_DEFUN([OMPI_CHECK_FCA],[
     OPAL_CHECK_WITHDIR([fca], [$with_fca], [lib/libfca.so])
```
jsquyres: The help string is inaccurate -- it says it looks for libraries in DIR, but it really looks in DIR/lib (and headers in DIR/include).  Can you fix that up as part of this PR, too?  (sorry!)

jsquyres:
I can't test this, but I *think* it looks ok...? (other than the inaccurate help message)

@bureddy did you test this patch?

bureddy:
@jsquyres , Yes, I tested v1.8 patch with options --with-hcoll, --with-hcoll=yes, --with-hcoll=no,   --with-hcoll=<hcoll_install>.

jsquyres:
Cool.  Can you test v1.10 and v2.x, too?  Feel free the thumbs-up each of the PRs after you test them.  Thanks!

config/ompi_check_libfca.m4
ggouaillardet: i am doing that right now
i also fixed the same inaccurate string in hcoll, mxm (and knem too)

ggouaillardet:
i fixed the messages and did some cleanup,
it took more time than expected, but all should be good now

jsquyres:
One last comment: didn't a user bring this to our attention on the mailing list?  If so, it would be good to cite that user's name in the commit message (as a way of saying "thank you!").  E.g., "Thanks to Joe Smith for bringing this issue to our attention."

PR #806, opened 2015-08-13
Add static definitions where needed and remove one unused definition

rolfv:
Noticed these because I was getting common symbols found.  I am just going to review this one myself.


hppritcha:
should this go in to v2.x as well?

rolfv:
Yes, will do v2.x as well.

PR #808, opened 2015-08-14
patch generated configure to fix an other libtool bug.

ggouaillardet:
Thanks to Eric Schnetter for reporting this issue and providing a fix.


jsquyres:
:+1: 

@ggouaillardet Just curious, did you get any response from upstream Libtool to fix this problem?

ggouaillardet:
not yet
I reported one issue so far, and I am not yet convinced about the second one
(and I cannot reproduce it yet)
I will ping the lib tool folks next week

PR #811, opened 2015-08-14
Fix macro return value when not CUDA-aware

rolfv:
This is to fix things so we set the macro to 0 when no CUDA-aware.  Otherwise, the macro is always set to 1 which is obviously wrong.

@jsquyres can you look at this one?

I still may try and change it so things are undefined, but that is still under discussion and I would like it to at least work as it was originally intended.

### DIFF
--- a/ompi/mpiext/cuda/configure.m4
+++ b/ompi/mpiext/cuda/configure.m4
@@ -17,6 +17,10 @@
 AC_DEFUN([OMPI_MPIEXT_cuda_CONFIG],[
     AC_CONFIG_FILES([ompi/mpiext/cuda/Makefile])
     AC_CONFIG_FILES([ompi/mpiext/cuda/c/Makefile])
+    AC_CONFIG_HEADER([ompi/mpiext/cuda/c/mpiext_cuda_c.h])
+
+    AC_DEFINE_UNQUOTED([MPIX_CUDA_AWARE_SUPPORT],[$CUDA_SUPPORT],
+                       [Macro that is set to 1 when CUDA-aware support is configured in and 0 when it is not])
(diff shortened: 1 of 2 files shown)


ompi/mpiext/cuda/configure.m4
```diff
@@ -17,6 +17,10 @@
 AC_DEFUN([OMPI_MPIEXT_cuda_CONFIG],[
     AC_CONFIG_FILES([ompi/mpiext/cuda/Makefile])
     AC_CONFIG_FILES([ompi/mpiext/cuda/c/Makefile])
+    AC_CONFIG_HEADER([ompi/mpiext/cuda/c/mpiext_cuda_c.h])
+
+    AC_DEFINE_UNQUOTED([MPIX_CUDA_AWARE_SUPPORT],$CUDA_SUPPORT,
```
jsquyres: You should put `$CUDA_SUPPORT` in brackets, i.e., `[$CUDA_SUPPORT]`.

ompi/mpiext/cuda/configure.m4
```diff
@@ -17,6 +17,10 @@
...
+                       [Macro that is set to 1 when CUDA-aware support is configured in])
```
jsquyres: This is a super-picky note: but you might want to update this comment to say that the value is always 0 or 1 (just to distinguish it from being 1 or undefined).

jsquyres:
Other than those 2 minor comments, looks good.

rolfv:
Fixed those two issues.  Will wait for more testing before merging.

rolfv:
bot:retest

Issue #369, opened 2015-02-01
Persistent communication request provokes segfault in Java bindings.


osvegis:
MPI corrupts the memory space of Java.
The following example provokes segfault in Java bindings.
Please, see the comments in the example.

```java
import mpi.*;                                                                         
import java.nio.*;                                                                    

public class CrashTest
{
    private static final int STEPS = 1000000000,
        SIZE  = 4096;

    public static void main(String...args) throws MPIException
    {
        MPI.Init(args);
        int rank = MPI.COMM_WORLD.getRank();
        StringBuilder s = new StringBuilder();

        if(MPI.COMM_WORLD.getSize() != 2)
            throw new MPIException("I need exactly 2 processes.");

        // Only one buffer is needed,                                                 
        // but the test works ok if you only use one.                                 
        ByteBuffer sendBuf = MPI.newByteBuffer(SIZE),
            recvBuf = MPI.newByteBuffer(SIZE);

        Prequest req = MPI.COMM_WORLD.recvInit(recvBuf, SIZE, MPI.BYTE, 0, 0);

        for(int i = 1; i <= STEPS; i++)
            {
                // Allocate memory to provoke GC work and crash.                      
                // If you comment the following line, the test works ok.              
                (s = new StringBuilder(SIZE).append(i)).trimToSize();

                if(rank == 0)
                    {
                        if(i % 100000 == 0)
                            {
                                s.setLength(0);
                                System.out.println(i + s.toString());
                            }

                        MPI.COMM_WORLD.send(sendBuf, SIZE, MPI.BYTE, 1, 0);
                    }
                else
                    {
                        req.start();
                        req.waitFor();
                    }
            }

        MPI.Finalize();
    }

} // CrashTest
```

jsquyres:
@osvegis I'm unable to get this test to fail for me.

Should I be seeing the memory usage of the process increase as the process runs?

I increased SIZE to 1048576, and after running 2 procs of this on a single 128GB server, the memory usage (as reported by "top") still only shows 5.1% memory usage for both java processes (it's been 5.1% the whole time):

```
  PID USER      PR  NI  VIRT  RES  SHR S %CPU %MEM    TIME+  COMMAND                 
19701 jsquyres  20   0 18.7g 3.2g  13m S 100.0  5.1   5:34.79 java                    
19702 jsquyres  20   0 18.7g 3.2g  13m S 100.0  5.1   5:34.47 java
```

jsquyres:
[...]

PR #815, opened 2015-08-18
java/Prequest: keep a pointer to the buffer

ggouaillardet:
this prevents garbage collector from freeing the buffer

Refs open-mpi/ompi#369


hppritcha:
@nrgraham23 we need to discuss a more general solution.

jsquyres:
@ggouaillardet We talked about this on the call today.  @hppritcha does not feel that this is a sufficient solution (as he alluded to above).  He can explain further.

ggouaillardet:
@nrgraham23 is implementing a full solution in #820 so there is no more need to keep this PR open

PR #816, opened 2015-08-18
yalla: fix passing on-demand mapping config to mxm.

yosefe:


jladd-mlnx:
@yosefe Please open a PR on 2.0 also.

PR #817, opened 2015-08-18
opal/asm: remove alpha support

hjelmn:
This commit removes alpha asm support. No current processor
manufacturer makes chips compatible with DEC alpha and no
participating organization has alpha processors. This makes it
difficult to support alpha via assembly.

This doesn't mean Open MPI will no longer build/work on alpha
processors. It should continue to work with gcc's builtin sync
atomics.


hjelmn:
Passes make distcheck on my mac. Merging.

PR #819, opened 2015-08-18
osc-portals4: allow atomic ops on datatypes that are max_fetch_atomic_size bytes in length

tkordenbrock:
Portals4 supports atomic ops on datatypes less than or equal to
max_fetch_atomic_size bytes.  This commit fixes a bug that required
the datatype to be less than max_fetch_atomic_size bytes.

@regrant - Please review.


hppritcha:
Is this going to be pushed back to v2.x?

tkordenbrock:
Yes.  I planned to file a PR for v2.x after it's reviewed.  Do you want me to file that PR now?

hppritcha:
yes it would be great if you could open a pr for v2.x when you get the
chance.

----------

Howard

tkordenbrock:
Haven't had any problem on my end.  Merging.

PR #820, opened 2015-08-18
Java garbage collection bugfix

### LINKED ISSUE #369: Persistent communication request provokes segfault in Java bindings.

nrgraham23:
This pull request adds an arraylist of type Buffer to
the Request class.  Whenever a request object is created
that has associated buffers, the buffers should be added
to this array list so the java garbage collector does
not dispose of the buffers prematurely.

This is a more robust expansion on the idea first proposed by
@ggouaillardet

Fixes #369

@jsquyres @hppritcha @ggouaillardet @osvegis 

If this solution is accepted, I will also also implement this
in the other files, where this is an issue.

### DIFF
--- a/ompi/mpi/java/java/Request.java
+++ b/ompi/mpi/java/java/Request.java
@@ -63,12 +63,16 @@
 
 package mpi;
 
+import java.util.ArrayList;
+import java.nio.Buffer;
+
 /**
  * Request object.
  */
 public class Request implements Freeable
 {
 	protected long handle;
+	protected ArrayList<Buffer> buffers;
 
 	static
 	{
@@ -113,6 +117,22 @@ public class Request implements Freeable
 	private native void cancel(long request) throws MPIException;
 
 	/**
+	 * Adds a buffer to the buffer array list.  This method should
+	 * be called by the internal api whenever a persistent request
+	 * is created and any time a request object, that has an associated
+	 * buffer, is returned from a non-blocking operation to protect 
+	 * the buffers from being garbage collected prematurely.
+	 * @param buf buffer to add to the array list
+	 */
+	protected final void addBufRef(Buffer buf)
+	{
+		if(this.buffers == null)
+			this.buffers = new ArrayList<Buffer>();
+		
+		this.buffers.add(buf);
+	}
+	
+	/**
 	 * Test if request object is null.
 	 * @return true if the request object is null, false otherwise
 	 */
@@ -120,7 +140,7 @@ public class Request implements Freeable
 	{
 		return handle == 0 || handle == MPI.REQUEST_NULL.handle;
 	}
-
+	
 	/**
 	 * Blocks until the operation identified by the request is complete.
 	 * <p>Java binding of the MPI operation {@code MPI_WAIT}.
--- a/ompi/mpi/java/java/Comm.java
+++ b/ompi/mpi/java/java/Comm.java
@@ -641,7 +641,9 @@ public class Comm implements Freeable, Cloneable
 	{
 		MPI.check();
 		assertDirectBuffer(buf);
-		return new Request(iSend(handle, buf, count, type.handle, dest, tag));
+		Request req = new Request(iSend(handle, buf, count, type.handle, dest, tag));
+		req.addBufRef(buf);
+		return req;
 	}
 
 	private native long iSend(
@@ -666,7 +668,9 @@ public class Comm implements Freeable, Cloneable
 	{
 		MPI.check();
 		assertDirectBuffer(buf);
-		return new Request(ibSend(handle, buf, count, type.handle, dest, tag));
+		Request req = new Request(ibSend(handle, buf, count, type.handle, dest, tag)); 
+		req.addBufRef(buf);
+		return req;
 	}
 
 	private native long ibSend(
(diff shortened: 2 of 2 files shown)


hppritcha:
This will need to be pushed back to both v2.x and 1.10 in ompi-release.
I'm good with the general approach.

ompi/mpi/java/java/Request.java
```diff
@@ -113,6 +117,22 @@ public final void cancel() throws MPIException
...
+	 * Adds a buffer to the buffer array list.  This method should
+	 * be called by the internal api whenever a persistent request
+	 * is created and any time a request object, that has an associated
+	 * buffer, is returned from a non-blocking operation to protect 
+	 * the buffers from being garbage collected prematurely.
+	 * @param buf buffer to add to the array list
+	 */
+	protected final void addBufRef(Buffer buf)
+	{
+		if(this.buffers == null)
```
jsquyres: Minor nits on the style here:

* Did we really use 8 space tabs throughout the Java code?  4 spaces would be preferable, since that's what the rest of the OMPI code base uses.
* `if (null == this.buffers)`: put a space after `if`, and put the constant on the left side of the `==` (i.e., defensive programming)
* Put `{}` around all blocks, including 1-line blocks (i.e., even more defensive programming)

See https://github.com/open-mpi/ompi/wiki/CodingStyle.

ompi/mpi/java/java/Request.java
```diff
@@ -120,7 +140,7 @@ public final boolean isNull()
 	{
 		return handle == 0 || handle == MPI.REQUEST_NULL.handle;
 	}
-
+
```
jsquyres: Super-minor nit: looks like extra whitespace got added here.

ompi/mpi/java/java/Request.java
hppritcha: Nathan reformatted all java files to conform more closrly to java
standard.  i think he used eclipse to do the reformatting.

----------

Howard

jsquyres: Oh, I have a vague recollection of this now.

So -- the 8-space tabs is common in the Java world?

I'd still prefer the `{}` around all blocks, though; that's just defensive programming.

hppritcha:
the test in bug #369 now works for me on hopper.  I had modified the test some to explicit call the gc.  Prior to this fix, usually took only a couple of iterations before the gc exploded with a segfault.

ompi/mpi/java/java/Request.java
nrgraham23: The 8 space tabs are now present in all of the java code.  I am not sure what is common in the java world, but I believe it is one of those little issues that people debate back and forth quite a bit.  Based on a quick google, it looks like maybe I cannot make eclipse do 4 space tabs.  I can play with it if you like.

Ill keep what you said about the if statements in mind, but in my updated fix I am currently working on, I have removed the if statement.

osvegis: El 19/08/15 a las 19:20, Nathaniel Graham escribió:
>
> In ompi/mpi/java/java/Request.java 
> <https://github.com/open-mpi/ompi/pull/820#discussion_r37441459>:
>
> > @@ -113,6 +117,22 @@ public final void cancel() throws MPIException
> >  	private native void cancel(long request) throws MPIException;
> >
> >  	/**
> > +	 * Adds a buffer to the buffer array list.  This method should
> > +	 * be called by the internal api whenever a persistent request
> > +	 * is created and any time a request object, that has an associated
> > +	 * buffer, is returned from a non-blocking operation to protect
> > +	 * the buffers from being garbage collected prematurely.
> > +	 * @param buf buffer to add to the array list
> > +	 */
> > +	protected final void addBufRef(Buffer buf)
> > +	{
> > +		if(this.buffers == null)
>
> The 8 space tabs are now present in all of the java code. I am not 
> sure what is common in the java world, but I believe it is one of 
> those little issues that people debate back and forth quite a bit. 
> Based on a quick google, it looks like maybe I cannot make eclipse do 
> 4 space tabs. I can play with it if you like.
>

You can make eclipse do 4 space tabs:

Window / Preferences / General / Editors / Text Editors:
     Displayed tab width: 4
     Insert spaces for tabs: yes

PR #822, opened 2015-08-20
Java garbage collection bugfix

### LINKED ISSUE #369: Persistent communication request provokes segfault in Java bindings.

nrgraham23:
This pull request adds two Buffer references to the 
Request class.  Whenever a request object is created
that has associated buffers, the buffers should be added
to these references so the java garbage collector does
not dispose of the buffers prematurely.

This is a more robust expansion on the idea first proposed by
@ggouaillardet, and further discussed in #369.

Fixes #369

@hppritcha @jsquyres @ggouaillardet @osvegis

### DIFF
--- a/ompi/mpi/java/java/Comm.java
+++ b/ompi/mpi/java/java/Comm.java
@@ -152,7 +152,7 @@ public class Comm implements Freeable, Cloneable
 	}
 
 	protected final native long[] iDup(long comm) throws MPIException;
-	
+
 	/**
 	 * Duplicates this communicator with the info object used in the call.
 	 * <p>Java binding of {@code MPI_COMM_DUP_WITH_INFO}.
@@ -162,12 +162,12 @@ public class Comm implements Freeable, Cloneable
 	 */
 	public Comm dupWithInfo(Info info) throws MPIException
 	{
-	    MPI.check();
-	    return new Comm(dupWithInfo(handle, info.handle));
+		MPI.check();
+		return new Comm(dupWithInfo(handle, info.handle));
 	}
 
 	protected final native long dupWithInfo(long comm, long info) throws MPIException;
-	
+
 	/**
 	 * Returns the associated request to this communicator if it was
 	 * created using {@link #iDup}.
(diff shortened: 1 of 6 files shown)


ompi/mpi/java/java/Comm.java
```diff
@@ -162,12 +162,12 @@ public Comm iDup() throws MPIException
 	 */
 	public Comm dupWithInfo(Info info) throws MPIException
 	{
-	    MPI.check();
-	    return new Comm(dupWithInfo(handle, info.handle));
+		MPI.check();
+		return new Comm(dupWithInfo(handle, info.handle));
```
hppritcha: minor thing, and don't redo for this PR, but normally try to avoid whitespace changes in PRs that are not just whitespace changes.

hppritcha:
:+1:

hppritcha:
@nrgraham23 please push this fix back to v1.10.

PR #823, opened 2015-08-20
Topic/sharedfp fview fix

edgargabriel:


hppritcha:
@edgargabriel  this will NOT go in to v2.x right?

edgargabriel:
actually it it will, except for the patch to the addproc component (which is the reason that it is a separate commit)

PR #824, opened 2015-08-21
opal/mutex: remove unnecessary ()s from OPAL_SCOPED_LOCK macro

hjelmn:
This modification allows the OPAL_SCOPED_LOCK macro to be used in more flexible ways. Example:

```c
OPAL_SCOPED_LOCK(&list_lock,
OPAL_LIST_FOREACH(item, &list, list_type_t) {
    some_action();
}
);
```

### DIFF
--- a/opal/threads/mutex.h
+++ b/opal/threads/mutex.h
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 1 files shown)


hjelmn:
I should note that with gcc the following is currently possible:
```
OPAL_SCOPED_LOCK(&list_lock,{
OPAL_LIST_FOREACH(item, &list, list_type_t) {
    some_action();
}
});
```

But that is non-standard.

jsquyres:
@hjelmn Is there a downside to removing the `()` around the action in the body of the macro?

hjelmn:
there shouldnt be any downside to removing the ()s. the second argument of the macro is expected to be any c statement and this change allows that

PR #826, opened 2015-08-24
btl/openib: add missing lock calls

hjelmn:


hjelmn:
These should be safe to commit once the Mellanox Jenkins is finished.

BTW, I tested the openib btl with multiple thread tests and see 0 failures. It may be time to remove the check that disables the btl in the MPI_THREAD_MULTIPLE case.

Issue #818, opened 2015-08-18
Implement --enable-static-runtime configure option.


gpaulsen:
From #729:

Jeff Squyres recommend this approach to solve this type of issue:

I do like the idea of a `--enable-static-runtime` kind of configure option.  This would seem to solve all the problems, and not require the user to know/do anything at runtime.

The only problem is: I'm not entirely sure how to do that.  

1. With the Intel compiler, there's the `--static-intel` option, but that doesn't make a wholly static `orted`, for example (it just statically links in the Intel libraries -- not *all* libraries).
1. Are there universal options for this for the other compilers?

...after thinking about this for a few minutes, it *may* be sufficient to do something like this:

```shell
if enable_static_runtime was passed
    RUNTIME_LDFLAGS=--static
    if compiler suite is intel
        RUNTIME_LDFLAGS="$RUNTIME_LDFLAGS --static-intel"
    fi
fi
````

Assuming `$(RUNTIME_LDFLAGS)` is used when linking the `orted` (and any other relevant executables?), Libtool *might* automatically translate `--static` into whatever is relevant for the underlying compiler, and we only have one special case for the Intel compiler suite.

jsquyres:
@gpaulsen Your Github wiki-foo is weak.  :-)  I updated the description to be proper wiki notation.

gpaulsen:
:flushed:  I shall aspire to wiki-it-up :heavy_exclamation_mark:

ggouaillardet:
fwiw, except having a statically linked orted (or orted that only links with system libs) i do not see how this can be achieved ...
libopen-rte.so, libopen-pal.so and all plugins (if any) need the intel runtime.
so far, i think the only way to achieve the desired result is to have orted binary link with libopen-rte.a, libopen-pal.a and no plugin at all.

rhc54:
That is correct - Brian Barrett had done this a few years back. The key is that you build all of OMPI as static. Today, that still won't build the orted as fully static, so you have to add some magic dust to get that to include all the dependent libs. I don't remember what he had to do - something in the orted Makefile.am plus some configure stuff. You could send him a note and ask - he might remember.

gpaulsen:
Why must all of OMPI be built static?  Why not just link / archive (ar) each of the OMPI pieces against the static compiler runtime libraries?  seems like you could still build a liborte.so, but just archive in the static C runtime.

rhc54:
[...]

PR #829, opened 2015-08-25
configury: add the --enable-static-runtime option

ggouaillardet:
with this option, static libraries are forced to be built, and
orted is linked with the --static libtool flag.
This is enough so orted does not need the intel compiler runtime
dynamic libraries when compiled with icc.

### DIFF
--- a/config/orte_configure_options.m4
+++ b/config/orte_configure_options.m4
@@ -16,6 +16,8 @@ dnl Copyright (c) 2009      IBM Corporation.  All rights reserved.
 dnl Copyright (c) 2009-2013 Los Alamos National Security, LLC.  All rights
 dnl                         reserved.
 dnl Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
+dnl Copyright (c) 2015      Research Organization for Information Science
+dnl                         and Technology (RIST). All rights reserved.
 dnl
 dnl $COPYRIGHT$
 dnl
@@ -68,4 +70,10 @@ AC_DEFINE_UNQUOTED([ORTE_ENABLE_STATIC_PORTS],
                    [$orte_enable_static_ports],
 		   [Whether we want static ports enabled])
 
+AC_ARG_ENABLE(static-runtime,
+AC_HELP_STRING([--enable-static-runtime],
+               [Link orted with the static compiler runtime]))
+AS_IF([test "x$enable_static_runtime" = "xyes"],
+      [enable_static=yes])
+AM_CONDITIONAL([ORTE_STATIC_RUNTIME], [test "x$enable_static_runtime" = "xyes"])
 ])dnl
--- a/orte/tools/orted/Makefile.am
+++ b/orte/tools/orted/Makefile.am
@@ -11,6 +11,8 @@
 #                         All rights reserved.
 # Copyright (c) 2007-2014 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2008      Sun Microsystems, Inc.  All rights reserved.
+# Copyright (c) 2015      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -43,7 +45,11 @@ orted_SOURCES = orted.c
 #  nothing to -all-static in the Makefile.in
 #  nice for systems that don't have all the shared
 #  libraries on the computes
+if ORTE_STATIC_RUNTIME
+orted_LDFLAGS = -static
+else
 orted_LDFLAGS =
+endif # ORTE_STATIC_RUNTIME
 orted_LDADD = \
 	$(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la \
 	$(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la


ggouaillardet:
Refs #818 
that works for me with intel compiler, note i did not even need to pass the ```-static-intel``` flag when linking ```orted```
note that if ompi if configure'd with ```--disable-static``` ```--enable-static-runtime```, static libs are built and no warning is issued.
@gpaulsen can you please review this ?

jsquyres:
I'm not a big fan of `--enable-static-runtime` silently enabling `--enable-static`.  At a bare minimum, I should think you should print a big warning that this is happening.  My $0.02:

* `--enable-static-runtime` should enable/disable the Libtool `--static` flag when building `orted`
* If `--enable-static-runtime` is set but `--enable-static` is *not* set, emit a large warning saying "This might not do what you want..."  But this should only be a warning, because what most people really want is static dependent library linking (e.g., the Intel runtime static libraries), not necessarily static `libopen-rte.a` and friends.

ggouaillardet:
if we use ```--enable-static-runtime``` and without ```--enable-static```, ```orted``` is linked with ```libopen-pal.so``` and ```libopen-rte.so``` which do depend on the intel runtime, so we do not end up with the expected result.

a better option could be to build ```libopen-pal.a``` and ```libopen-rte.a``` but install them only if ```--enable-static``` was used at configure time.
/* right now, i do not know how that can be easily achieved from an autotool point of view */

fwiw, here is the ldd output of orted when configure'd with ```--enable-static-runtime``` (that implies ```--enable-static``` in this PR)

```
$ ldd bin/orted 
	linux-vdso.so.1 =>  (0x00007fffa19dc000)
	libtorque.so.2 => /usr/lib64/libtorque.so.2 (0x0000003790e00000)
	libxml2.so.2 => /usr/lib64/libxml2.so.2 (0x0000003790a00000)
	libz.so.1 => /lib64/libz.so.1 (0x000000378fa00000)
	libcrypto.so.10 => /usr/lib64/libcrypto.so.10 (0x0000003790200000)
	libssl.so.10 => /usr/lib64/libssl.so.10 (0x0000003790600000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x0000003c62000000)
	librdmacm.so.1 => /usr/lib64/librdmacm.so.1 (0x0000003700a00000)
	libosmcomp.so.3 => /usr/lib64/libosmcomp.so.3 (0x0000003700200000)
	libibverbs.so.1 => /usr/lib64/libibverbs.so.1 (0x00007f568a415000)
	libm.so.6 => /lib64/libm.so.6 (0x0000003c62400000)
	libdl.so.2 => /lib64/libdl.so.2 (0x0000003c61800000)
	librt.so.1 => /lib64/librt.so.1 (0x0000003c62c00000)
	libutil.so.1 => /lib64/libutil.so.1 (0x0000003c71800000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000003c65c00000)
	libc.so.6 => /lib64/libc.so.6 (0x0000003c61c00000)
	libgssapi_krb5.so.2 => /lib64/libgssapi_krb5.so.2 (0x0000003c6ac00000)
	libkrb5.so.3 => /lib64/libkrb5.so.3 (0x0000003c69c00000)
	libcom_err.so.2 => /lib64/libcom_err.so.2 (0x0000003c68c00000)
	libk5crypto.so.3 => /lib64/libk5crypto.so.3 (0x0000003c69400000)
	/lib64/ld-linux-x86-64.so.2 (0x0000003c61400000)
	libibumad.so.3 => /usr/lib64/libibumad.so.3 (0x0000003700600000)
	libwrap.so.0 => /lib64/libwrap.so.0 (0x0000003c6a000000)
	libkrb5support.so.0 => /lib64/libkrb5support.so.0 (0x0000003c6a800000)
	libkeyutils.so.1 => /lib64/libkeyutils.so.1 (0x00007f568a20d000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x0000003c63800000)
	libnsl.so.1 => /lib64/libnsl.so.1 (0x0000003c72000000)
	libselinux.so.1 => /lib64/libselinux.so.1 (0x0000003c63400000)
```
surprisingly, some libs (such as libutil) are dynamic, even if static version are available
```
$ ls -l /usr/lib64/libutil*   
-rw-r--r--  1 root root 13632 Dec  7  2011 /usr/lib64/libutil.a
lrwxrwxrwx. 1 root root    24 Dec  5  2012 /usr/lib64/libutil.so -> ../../lib64/libutil.so.1
```
so passing the libtool ```--static``` flag when linking ```orted``` do not really generate a statically linked binary

ggouaillardet:
in the case of the intel compiler, it seems the runtime might not even be necessary.
it seems the ```-static-intel``` flag is some how intercepted by the libtool wrapper, but it is possible to pass ```-Wc,-static-intel``` instead.
i was able to build an orted not depending on the intel runtime (and __without__ ```--enable-static``` nor ```--disable-dlopen```) by configuring with
```CFLAGS=-Wc,-static-intel LDFLAGS=-Wc,-static-intel```

jsquyres:
I wonder if this is turning into a FAQ item rather than an autotools solution...?

I.e., should we just add @ggouaillardet's trick of `CFLAGS=-Wc,-static-intel LDFLAGS=-Wc,-static-intel` to the FAQ?  (is the CFLAGS really necessary, or just the LDFLAGS?)

ggouaillardet:
I will double check this tomorrow

ggouaillardet:
indeed, only ```LDFLAGS=-Wc,-static-intel``` is enough  to achieve the expected result.
fwiw, only libmpi_usempif08.so requires the intel runtime.

jsquyres:
To close out this issue: we decided to make this an FAQ item and leave autogen/configure alone. The FAQ item has been added, so I'm closing this issue (and open-mpi/ompi#818).

PR #830, opened 2015-08-25
configury: add a --disable-libnl3 configure option

ggouaillardet:
this option is currently part of the reachable/netlink mca,
and prevent usage of libnl3 if available

### DIFF
--- a/opal/mca/reachable/netlink/configure.m4
+++ b/opal/mca/reachable/netlink/configure.m4
@@ -1,6 +1,8 @@
 # -*- shell-script -*-
 #
-# Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -73,16 +75,22 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
 
        save_CPPFLAGS=$CPPFLAGS
        save_LIBS=$LIBS
+       AC_ARG_ENABLE(libnl3,
+                     AC_HELP_STRING([--disable-libnl3],
+                                    [Do not try to use libnl3]))
 
-       $1="-I/usr/include/libnl3"
-       CPPFLAGS="$$1 $CPPFLAGS"
-       AC_MSG_CHECKING([for /usr/include/libnl3])
-       AS_IF([test -d "/usr/include/libnl3"],
-             [AC_MSG_RESULT([present])
-              AC_CHECK_HEADER(
-                [netlink/version.h],
-                [AC_COMPILE_IFELSE(
-                    [AC_LANG_PROGRAM([[
+
+       AS_IF([test "x$enable_libnl3" = "xno"],
+             [HAVE_LIBNL3=0],
+             [$1="-I/usr/include/libnl3"
+              CPPFLAGS="$$1 $CPPFLAGS"
+              AC_MSG_CHECKING([for /usr/include/libnl3])
+              AS_IF([test -d "/usr/include/libnl3"],
+                    [AC_MSG_RESULT([present])
+                     AC_CHECK_HEADER(
+                       [netlink/version.h],
+                       [AC_COMPILE_IFELSE(
+                           [AC_LANG_PROGRAM([[
 #include <netlink/netlink.h>
 #include <netlink/version.h>
 #ifndef LIBNL_VER_MAJ
@@ -92,14 +100,14 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
 #if LIBNL_VER_MAJ < 3
 #error "LIBNL_VER_MAJ < 3, this is very unusual"
 #endif
-                        ]],[[/* empty body */]])],
-                    [HAVE_LIBNL3=1],    dnl our program compiled
-                    [HAVE_LIBNL3=0])],  dnl our program failed to compile
-                [HAVE_LIBNL3=0],  dnl AC_CHECK_HEADER failed
-                [#include <netlink/netlink.h>
-                ])],
-             [AC_MSG_RESULT([missing])
-              HAVE_LIBNL3=0])  dnl "/usr/include/libnl3" does not exist
+                               ]],[[/* empty body */]])],
+                           [HAVE_LIBNL3=1],    dnl our program compiled
+                           [HAVE_LIBNL3=0])],  dnl our program failed to compile
+                       [HAVE_LIBNL3=0],  dnl AC_CHECK_HEADER failed
+                       [#include <netlink/netlink.h>
+                       ])],
+                    [AC_MSG_RESULT([missing])
+                     HAVE_LIBNL3=0])])  dnl "/usr/include/libnl3" does not exist
 
        # nl_recvmsgs_report is a symbol that is only present in v3
        AS_IF([test "$HAVE_LIBNL3" -eq 1],


ggouaillardet:
this is related to the ML discussion starting at http://www.open-mpi.org/community/lists/devel/2015/08/17812.php

hppritcha:
testing the Houston jenkins node - kudos to [Univ. Houston](http://www.uh.edu/)

bot:retest

hppritcha:
okay had to restart jenkins. 
bot:retest

jsquyres:
I'm still uncomfortable with this approach.

I'll mention that we did a lot of hand-wringing over libnl vs. libnl3 in libfabric, and settled on the approach that was copied here (i.e., enable both, but prefer libnl3).  Here's why:

1. If you only only offer the option to *disable* libnl3, it's weird / hard to explain to users.
1. What does `--enable-libnl3` mean if there is no libnl3 available (and it could have fallen back to libnl)?
1. If you offer the ability to enable/disable libnl and libnl3, that's 4 options, and a whole pile of shell logic to write (and check for errors, e.g., if a user specifies both `--enable-libnl` *and* `--enable-libnl3`)
1. The world is moving to libnl3
1. If the user has both `libnl3-devel` and `libnl-devel` installed... WTF?

All that being said, I understand why you're doing this -- you *are* on a system with both `libnl-devel` and `libnl3-devel` installed, and there's something else in the stack that OMPI uses that uses v1, not v3.  But still -- this solution seems... weird (e.g., imagine writing the help text / explanation in README for --disable-libnl3).  :-(

I think the option I hate the least is supporting both enable/disable for both libnl/libnl3.

Thoughts?  Alternatives?

ggouaillardet:
all i can think of is using ```--without-libnl3``` instead of ```--disable-libnl3```
imho, that would make the ```-with-libnl3``` behavior more intuitive

i can easily replace ```--disable-libnl3``` with ```--without-libnl3```, add the extra logic the search for libnl3 in ```dir``` if invoked with ```--with-libnl3=<dir>```, and make configure aborts if ```--with-libnl3=<dir>``` or ```--with-libnl3=yes``` does not work, and maybe also if ```--with-libnl3``` is specified (e.g. no fallback to libnl, but this can be discussed)

as you already understood, on my centos 7 box, my options are currently
1. uninstall the ```libnl3-devel``` rpm
2. configure with ```--without-verbs```
3. configure so the netlink reachable mca is not built
and i like none of them ...

jsquyres:
What if we take a slightly more holistic approach?  Let me throw out an idea that I don't even know if it will work... (btw, let me shout a huge "THANKS!" to the netlink authors for putting us in this situation :frowning:)

What if we have configure-time and run-time detection for libnl-libnl3 conflicts?

## Configure time checks

I can imagine two kinds of checks:

1. Check various `LIBS`-like flags for `-lnl` and `-lnl3`
1. Check linked libraries for libnl and libnl3 symbols

Perhaps we could add new macros like these (off the top of my head):

* `OPAL_LIBNL_DECLARE_LIBNL`: allow anyone to declare that they are linking against libnl
* `OPAL_LIBNL_DECLARE_LIBNL3`: allow anyone to declare that they are linking against libnl3
* `OPAL_LIBNL_CHECK_LIB(lib)`: check if library `lib` contains libnl or libnl3 symbols.  If it contains libnl symbols, invoke `OPAL_LIBNL_DECLARE_LIBNL`.  If it contains libnl3 symbols, invoke `_DECLARE_LIBNL3`.  The cases these would find would be when linking against `libfoo` which implicitly links in libnl or libnl3.
* `OPAL_LIBNL_CHECK_LIBS(libs)`: check `$LIBS` flags specified by `libs` for `-lnl` (and invoke `_DECLARE_LIBNL`) and/or `-lnl3` (and invoke `_DECLARE_LIBNL3`).

If *both* `_DECLARE_LIBNL` *and* `_DECLARE_LIBNL3` end up being invoked, configure can abort with some kind of (hopefully) helpful message.

The use of the four `--enable|disable-libnl|libnl3` flags can also be mixed in with these macros.  For example, if `--enable-libnl` is specified and someone invokes `_DECLARE_LIBNL3`, then configure should abort, because that's a conflict.

## Run-time checks

Add `opal_check_libnl_conflict()` to check for libnl and libnl3 symbols in the current process.  I.e., use `opal_dl_open(NULL)` to open the current process and then `opal_dl_lookup()` for libnl and libnl3 symbols.

If it detects both libnl and libnl3 symbols... well, we can decide what to do (e.g., show_help() and abort?  Or just show_help()?  Or ...?  We can figure that out later).

What do you think?

ggouaillardet:
the configure time checks look good to me

about the runtime, I am not sure that can work as is
in my environment, the crash occurs when dlopen'ing mca_reachable_netlink.so, that is linked with libnl3 and after libnl was already loaded.
since the crash occur in libnl3 constructor, we cannot opal_dl_open it and opal_dl_lookup to see if it links with the wrong libnl.
that being said, we could do some nm or ldd like stuff before so we do not dlopen if we anticipate a libnl conflict
makes sense ?

jsquyres:
configure-time checks: cool

run-time checks: IIRC, the problem is not in a library constructor, but rather the first time you call a netlink function, right?  If that's correct, then calling something like `opal_check_libnl_conflict()` before calling a netlink function (or before calling something that calls a netlink function) should be sufficient, right?

ggouaillardet:
the issue is in the constructor iirc
I will double check this tomorrow and post a stack trace

ggouaillardet:
here is the stack trace
the crash does occur in the libnl3 constructor
```
Program received signal SIGABRT, Aborted.
0x00007ffff61755d7 in raise () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-78.el7.x86_64 libgcc-4.8.3-9.el7.x86_64 libibumad-1.3.10.2.MLNX20150406.966500d-0.1.x86_64 libibverbs-1.1.8mlnx1-OFED.3.0.0.3.2.x86_64 libnl-1.1.4-3.el7.x86_64 libnl3-3.2.21-8.el7.x86_64 libpciaccess-0.13.1-4.1.el7.x86_64 librdmacm-1.0.20mlnx-OFED.3.0.6.g22d4046.x86_64 munge-libs-0.5.11-1.el7.x86_64 numactl-libs-2.0.9-5.el7_1.x86_64 opensm-libs-4.5.0.MLNX20150510.07136f5-0.1.x86_64
(gdb) bt
#0  0x00007ffff61755d7 in raise () from /lib64/libc.so.6
#1  0x00007ffff6176cc8 in abort () from /lib64/libc.so.6
#2  0x00007ffff616e546 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007ffff616e5f2 in __assert_fail () from /lib64/libc.so.6
#4  0x00007ffff26823d9 in rtnl_tc_register () from /lib64/libnl-route-3.so.200
#5  0x00007ffff7dea4e3 in _dl_init_internal () from /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff7deeb04 in dl_open_worker () from /lib64/ld-linux-x86-64.so.2
#7  0x00007ffff7dea2f4 in _dl_catch_error () from /lib64/ld-linux-x86-64.so.2
#8  0x00007ffff7dee21b in _dl_open () from /lib64/ld-linux-x86-64.so.2
#9  0x00007ffff6e2b02b in dlopen_doit () from /lib64/libdl.so.2
#10 0x00007ffff7dea2f4 in _dl_catch_error () from /lib64/ld-linux-x86-64.so.2
#11 0x00007ffff6e2b62d in _dlerror_run () from /lib64/libdl.so.2
#12 0x00007ffff6e2b0c1 in dlopen@@GLIBC_2.2.5 () from /lib64/libdl.so.2
#13 0x00007ffff74d1004 in do_dlopen (fname=0x687af0 "/home/gilles/local/ompi-master/lib/openmpi/mca_reachable_netlink.so", flags=257, handle=0x7fffffffd748, err_msg=0x7fffffffd7f8)
    at ../../../../../../src/ompi-master/opal/mca/dl/dlopen/dl_dlopen_module.c:35
#14 0x00007ffff74d11dc in dlopen_open (fname=0x647580 "/home/gilles/local/ompi-master/lib/openmpi/mca_reachable_netlink", use_ext=true, private_namespace=false, handle=0x63bfb8, err_msg=0x7fffffffd7f8)
    at ../../../../../../src/ompi-master/opal/mca/dl/dlopen/dl_dlopen_module.c:92
#15 0x00007ffff74d0d3a in opal_dl_open (fname=0x647580 "/home/gilles/local/ompi-master/lib/openmpi/mca_reachable_netlink", use_ext=true, private_namespace=false, handle=0x63bfb8, err_msg=0x7fffffffd7f8)
    at ../../../../../src/ompi-master/opal/mca/dl/base/dl_base_fns.c:32
#16 0x00007ffff74aa1b7 in mca_base_component_repository_open (framework=0x7ffff77781c0 <opal_reachable_base_framework>, ri=0x63bf00)
    at ../../../../../src/ompi-master/opal/mca/base/mca_base_component_repository.c:362
#17 0x00007ffff74a8ed2 in find_dyn_components (path=0x0, framework=0x7ffff77781c0 <opal_reachable_base_framework>, names=0x0, include_mode=true)
    at ../../../../../src/ompi-master/opal/mca/base/mca_base_component_find.c:264
#18 0x00007ffff74a8ac4 in mca_base_component_find (directory=0x0, framework=0x7ffff77781c0 <opal_reachable_base_framework>, ignore_requested=true, open_dso_components=true)
    at ../../../../../src/ompi-master/opal/mca/base/mca_base_component_find.c:135
#19 0x00007ffff74b812d in mca_base_framework_components_register (framework=0x7ffff77781c0 <opal_reachable_base_framework>, flags=MCA_BASE_REGISTER_ALL)
    at ../../../../../src/ompi-master/opal/mca/base/mca_base_components_register.c:55
#20 0x00007ffff74b8ae2 in mca_base_framework_register (framework=0x7ffff77781c0 <opal_reachable_base_framework>, flags=MCA_BASE_REGISTER_ALL)
    at ../../../../../src/ompi-master/opal/mca/base/mca_base_framework.c:127
#21 0x00007ffff747b961 in info_register_framework (framework=0x7ffff77781c0 <opal_reachable_base_framework>, component_map=0x7fffffffda80) at ../../../src/ompi-master/opal/runtime/opal_info_support.c:238
#22 0x00007ffff747ba2b in opal_info_register_project_frameworks (project_name=0x7ffff7532d56 "opal", frameworks=0x7ffff7774de0 <opal_frameworks>, component_map=0x7fffffffda80)
    at ../../../src/ompi-master/opal/runtime/opal_info_support.c:259
#23 0x00007ffff747bc64 in opal_info_register_framework_params (component_map=0x7fffffffda80) at ../../../src/ompi-master/opal/runtime/opal_info_support.c:312
#24 0x00007ffff7aba4b6 in ompi_info_register_framework_params (component_map=0x7fffffffda80) at ../../../src/ompi-master/ompi/runtime/ompi_info_support.c:74
#25 0x0000000000401a0f in main (argc=1, argv=0x7fffffffdd28) at ../../../../../src/ompi-master/ompi/tools/ompi_info/ompi_info.c:130
```

jsquyres:
@ggouaillardet Well, that's a major bummer.

I don't really want to fork/exec sub-processes each time we `dlopen()` a DSO; that could lead to some pretty big overheads.  I also don't think it would catch any bad cases -- unless we `dlopen()` `libnl` or `libnl3` directly (which we don't).  Meh.  What to do?

Are the configure-time checks sufficient?

(sidenote: I find it odd that libnl3 is calling `assert()` in a production build :frowning:)

jsquyres:
@ggouaillardet Any further thoughts on this one?

ggouaillardet:
@jsquyres i made PR #1014 
```OPAL_CHECK_PACKAGE``` checks the libnl version that is used (if any)
if there is a conflict, a warning is displayed (i need to revamp that part, issue a clear message or even fail at the end of configure, and do not check this multiple times for a given library)
the ```reachable/netlink``` component uses this information reported by ```opal_check_libnl_version``` and only try ```libnl3``` or ```libnl``` if one of these libs was previously used.
so far, i tested it successfully on CentOS 7 (```libibverbs``` requires ```libnl```, so ```reachable/netlink``` only tries ```libnl```) and OpenSuSE 12.3 (```libibverbs``` requires ```libnl3``` so ```reachable/netlink``` only tries ```libnl3```)

jsquyres:
Ok.  This issue obsoleted by #1014.

PR #831, opened 2015-08-25
fortran: do not build libmpi_mpifh_[p]sizeof.la when fortran bindings…

ggouaillardet:
… are not built

### DIFF
--- a/ompi/mpi/fortran/mpif-h/Makefile.am
+++ b/ompi/mpi/fortran/mpif-h/Makefile.am
@@ -111,7 +111,7 @@ CLEANFILES += sizeof_f.f90
 # out this workaround indefinitely.  So Jeff advised Absoft to fix
 # this bug in Sep 2014.
 noinst_LTLIBRARIES=
-if BUILD_FORTRAN_SIZEOF
+if BUILD_PMPI_FORTRAN_MPIFH_BINDINGS_LAYER
 noinst_LTLIBRARIES += libmpi_mpifh_sizeof.la
 # Do not dist this file; it is generated
 nodist_libmpi_mpifh_sizeof_la_SOURCES = sizeof_f.f90
--- a/ompi/mpi/fortran/mpif-h/profile/Makefile.am
+++ b/ompi/mpi/fortran/mpif-h/profile/Makefile.am
@@ -45,8 +45,6 @@ AM_CPPFLAGS = -DOMPI_PROFILE_LAYER=1 -DOMPI_COMPILING_FORTRAN_WRAPPERS=1
 noinst_LTLIBRARIES =
 if BUILD_PMPI_FORTRAN_MPIFH_BINDINGS_LAYER
 noinst_LTLIBRARIES += libmpi_mpifh_pmpi.la
-else
-noinst_LTLIBRARIES +=
 endif
 
 headers = \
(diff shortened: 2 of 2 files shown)


adrianreber:
With this patch building works again even without gcc-gfortran.ppc64. :+1:

ompi/mpi/fortran/mpif-h/Makefile.am
```diff
@@ -111,7 +111,7 @@ CLEANFILES += sizeof_f.f90
 # out this workaround indefinitely.  So Jeff advised Absoft to fix
 # this bug in Sep 2014.
 noinst_LTLIBRARIES=
-if BUILD_FORTRAN_SIZEOF
+if BUILD_PMPI_FORTRAN_MPIFH_BINDINGS_LAYER
```
jsquyres: A few questions:

1. Shouldn't that be `MPI`, not `PMPI`?
1. Don't you also need the second conditional of `BUILD_FORTRAN_SIZEOF`?
1. Or, instead, should we make the `AM_CONDITIONAL` for `BUILD_FORTRAN_SIZEOF` only ever be true if `BUILD_*MPI_FORTRAN_MPIF_BINDINGS_LAYER` is also true?

ompi/mpi/fortran/mpif-h/profile/Makefile.am
```diff
@@ -45,8 +45,6 @@ AM_CPPFLAGS = -DOMPI_PROFILE_LAYER=1 -DOMPI_COMPILING_FORTRAN_WRAPPERS=1
 noinst_LTLIBRARIES =
 if BUILD_PMPI_FORTRAN_MPIFH_BINDINGS_LAYER
 noinst_LTLIBRARIES += libmpi_mpifh_pmpi.la
-else
-noinst_LTLIBRARIES +=
```
jsquyres: Good call -- these 2 lines can definitely go.

ompi/mpi/fortran/mpif-h/profile/Makefile.am
```diff
@@ -439,12 +437,12 @@ CLEANFILES += psizeof_f.f90
 # Build the MPI_SIZEOF code in a separate convenience library (see
 # lengthy comment in ompi/mpi/fortran/mpif-h/Makefile.am for an
 # explanation why).
-#if BUILD_FORTRAN_SIZEOF
+if BUILD_PMPI_FORTRAN_MPIFH_BINDINGS_LAYER
```
jsquyres: Meh; that `if` shouldn't have been commented out.  Bad Jeff.  :frowning: 

But regardless, the same questions from above apply (i.e., perhaps we should make `BUILD_FORTRAN_SIZEOF` smarter...?).

ompi/mpi/fortran/mpif-h/Makefile.am
ggouaillardet: 1. yes
3. looks better than 2.
I will revamp the fix, but unlikely before Friday

jsquyres: Since we want to get v1.10 out ASAP, I might take a whack at this and do a new PR today.  I'll see what I can do.

jsquyres:
I did a similar version in #832.

PR #832, opened 2015-08-25
fortran sizeof fixes

jsquyres:
This PR replaces #831 (development follows the sun...).


jsquyres:
@adrianreber Can you try this PR quickly on your machine?  I don't actually have a machine without a Fortran compiler...

jsquyres:
Actually, I just forcibly uninstalled gfortran on a machine and this patch seems to work for me.  I'll wait for jenkins to finish before merging to master / filing PRs for release branches.

jsquyres:
FWIW, this is not necessary on v1.10 -- the version of SIZEOF we did on v1.10 is *slightly* different than on master, and is already protected by the "if we're building Fortran..." AM_CONDITIONAL.

PR #835, opened 2015-08-27
Correct the wrong "Name Binding" of functions 

Zhiming-Wang:
There are several spelling mistakes. It is failed when building by IBM XL Fortran. I corrected them.

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/pmpi-f-interfaces-bind.h
+++ b/ompi/mpi/fortran/use-mpi-f08/pmpi-f-interfaces-bind.h
@@ -95,12 +95,11 @@ end subroutine pompi_buffer_attach_f
 ! PMPI_BUFFER_DETACH (i.e., it is different than the mpif.h / mpi
 ! module C implementation function).
 !
-! Note, too, we don't need a "p" version of the C implementation
-! function -- Fortran's interfaces provide MPI_ and PMPI_ names for
-! us; they can just both be bound to the same back-end
+! Note, too, the "p" version of the C implementation
+! function is a 1-line routine. It calls
 ! ompi_buffer_detach_f08 C function.
 subroutine pompi_buffer_detach_f(buffer_addr,size,ierror) &
-   BIND(C, name="ompi_buffer_detach_f08")
+   BIND(C, name="pompi_buffer_detach_f08")
    implicit none
    OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: buffer_addr
    INTEGER, INTENT(OUT) :: size
@@ -1418,7 +1417,7 @@ subroutine pompi_type_set_name_f(type,type_name,ierror,type_name_len) &
 end subroutine pompi_type_set_name_f
 
 subroutine pompi_win_allocate_f(size, disp_unit, info, comm, &
-      baseptr, win, ierror) BIND(C, name="ompi_win_allocate_f")
+      baseptr, win, ierror) BIND(C, name="pompi_win_allocate_f")
   USE, INTRINSIC ::  ISO_C_BINDING, ONLY : C_PTR
   use :: mpi_f08_types, only : MPI_ADDRESS_KIND
   INTEGER(KIND=MPI_ADDRESS_KIND), INTENT(IN) ::  size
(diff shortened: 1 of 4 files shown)


Zhiming-Wang:
Please ignore the file "optimized-power8-xlc".

ompi/mpi/fortran/use-mpi-f08/pmpi-f-interfaces-bind.h
```diff
@@ -98,9 +98,9 @@ end subroutine pompi_buffer_attach_f
 ! Note, too, we don't need a "p" version of the C implementation
 ! function -- Fortran's interfaces provide MPI_ and PMPI_ names for
 ! us; they can just both be bound to the same back-end
-! ompi_buffer_detach_f08 C function.
+! pompi_buffer_detach_f08 C function.
 subroutine pompi_buffer_detach_f(buffer_addr,size,ierror) &
-   BIND(C, name="ompi_buffer_detach_f08")
+   BIND(C, name="pompi_buffer_detach_f08")
```
jsquyres: Are you sure about this one?  See the comment directly above this subroutine for an explanation as to why it's `ompi_`, and not `pompi_`.

Specifically: I do not believe that there is a `pompi_buffer_detach_f08` C function.

jsquyres:
@Zhiming-Wang Sweet; thanks for these fixes.  It looks like Mellanox's Jenkins had some kind of local failure; I wouldn't worry about it.

Can you rebase this PR to remove the 3 commits about optimized-power8-xlc?  If you're new to the Github PR process, know that a Github PR is just about the only acceptable time you can modify public git history (in fact, it's encouraged :smile:).  Like this:

```sh
$ cd YOUR_GIT_TREE
$ git rebase -i origin/master
# This brings up an editor.
# Delete the 3 lines containing the unwanted commits.
# Save the text file and exit the editor.
$ git log
# See that the top of the tree is now just your 2 desired commits.
$ git push YOUR_REMOTE --force --dry-run
# See that the push to YOUR_REMOTE (would have) worked
# Now do the push again, but without --dry-run
$ git push YOUR_REMOTE
```

Replace YOUR_REMOTE with the name / URI of your github remote.

Once you do this, come back and refresh this web page and you'll see that this PR now just contains the 2 commits that you care about.

Make sense?

ompi/mpi/fortran/use-mpi-f08/pmpi-f-interfaces-bind.h
Zhiming-Wang: I don't know whether there is a real "pompi_buffer_detach_f08' function at Open MPI. This is at Fortran 2008 support model. It is rarely used and skipped when building by gfortran. 
The fix is just about bind name which is different from function name before. At ifort, it passed. But there is more strict syntax at xlf.
I am not familiar with GitHub and will rebase this PR later. Thanks very much. @jsquyres 
Error message from xlf:
"pmpi-f-interfaces-bind.h", line 100.1: 1516-194 (S) A binding label must not be the same as the name of a program unit, common block, external procedure, and another binding label, ignoring differences in case.

jsquyres: Ah, I was unaware of the Fortran restriction that we can't use the same binding name across multiple subroutines.  Bummer.

Ok. Can you update the comment in pmpi-f-interfaces-bind.h, and then add an implementation of `pompi_buffer_detatch_f08()`?  You can likely just add a 1-line routine in `ompi/mpi/fortran/use-mpi-f08/buffer_detatch.c` that turns around and calls `pompi_buffer_detatch_f08()` (i.e., just so that we can have 2 different C symbols for Fortran to call, but only have the body of the C routine once).

Zhiming-Wang:
@jsquyres  I created a new changeset and commit and rebase it. I found the "pmpi-f08-interfaces.F90" is needed to be modified, too. I added "pompi_buffer_detatch_f08()" routine at "buffer_detatch.c". Please confirm it.

jsquyres:
Works for me -- thanks! :+1: 

Can you make PR's for this for the v1.10 and v2.x branches?  Check out https://github.com/open-mpi/ompi/wiki/SubmittingPullRequests.

PR #836, opened 2015-08-27
Add config code to check for need of workaround. Add runtime way to turn oiff just in case.

rolfv:
I think the summary above says it all.  I added this workaround, but then the issue was fixed in the CUDA driver so workaround is not needed for later versions.

@jsquyres is a good reviewer as he always finds something.

### DIFF
--- a/opal/mca/common/cuda/common_cuda.c
+++ b/opal/mca/common/cuda/common_cuda.c
@@ -126,6 +126,7 @@ static CUstream ipcStream = NULL;
 static CUstream dtohStream = NULL;
 static CUstream htodStream = NULL;
 static CUstream memcpyStream = NULL;
+static int mca_common_cuda_gpu_mem_check_workaround = (CUDA_VERSION > 7000) ? 0 : 1;
 static opal_mutex_t common_cuda_init_lock;
 static opal_mutex_t common_cuda_htod_lock;
 static opal_mutex_t common_cuda_dtoh_lock;
@@ -300,6 +301,13 @@ void mca_common_cuda_register_mca_variables(void)
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &mca_common_cuda_cumemcpy_timing);
 #endif /* OPAL_ENABLE_DEBUG */
+
+    (void) mca_base_var_register("ompi", "mpi", "common_cuda", "gpu_mem_check_workaround",
+                                 "Set to 0 to disable GPU memory check workaround. A user would rarely have to do this.",
+                                 MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                 OPAL_INFO_LVL_9,
+                                 MCA_BASE_VAR_SCOPE_READONLY,
+                                 &mca_common_cuda_gpu_mem_check_workaround);
 }
 
 /**
@@ -773,6 +781,9 @@ static int mca_common_cuda_stage_three_init(void)
         opal_output_verbose(20, mca_common_cuda_output,
                             "CUDA: cuMemHostRegister OK on test region");
     }
+
+    opal_output_verbose(20, mca_common_cuda_output,
+                        "CUDA: the extra gpu memory check is %s", (mca_common_cuda_gpu_mem_check_workaround == 1) ? "on":"off");
 
     opal_output_verbose(30, mca_common_cuda_output,
                         "CUDA: initialized");
@@ -1832,7 +1843,7 @@ static int mca_common_cuda_is_gpu_buffer(const void *pUserBuf, opal_convertor_t 
      * made it this far, then the assumption at this point is we have GPU memory.
      * Unfotunately, this extra call is costing us another 100 ns almost doubling
      * the cost of this entire function. */
-    {
+    if (OPAL_LIKELY(mca_common_cuda_gpu_mem_check_workaround)) {
         CUdeviceptr pbase;
         size_t psize;
         res = cuFunc.cuMemGetAddressRange(&pbase, &psize, dbuf);


opal/mca/common/cuda/common_cuda.c
```diff
@@ -300,6 +303,15 @@ void mca_common_cuda_register_mca_variables(void)
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &mca_common_cuda_cumemcpy_timing);
 #endif /* OPAL_ENABLE_DEBUG */
+
+#if OPAL_CUDA_GPU_MEM_WAR
+    (void) mca_base_var_register("ompi", "mpi", "common_cuda", "gpu_mem_check_war",
+                                 "Set to 1 to enable GPU memory check war",
```
jsquyres: Is WAR an acronym (e.g., "write-after-read")?  

Do you want to put a little more in the help string to help the user choose which way they should set this (i.e., they should rarely -- if ever -- need to change it, right?).

opal/mca/common/cuda/common_cuda.c
```diff
@@ -300,6 +303,15 @@ void mca_common_cuda_register_mca_variables(void)
...
+                                 MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                 OPAL_INFO_LVL_9,
```
jsquyres: Is this MCA param for a *correctness* issue, or a *performance tuning* issue?

If it's a *correctness* issue, I'd make it level 2 or 3 (per https://github.com/open-mpi/ompi/wiki/MCAParamLevels).
If it's a *tuning* issue, I'd make it level 4 or 5.

opal/mca/common/cuda/common_cuda.c
```diff
@@ -126,6 +126,9 @@ static CUstream ipcStream = NULL;
 static CUstream dtohStream = NULL;
 static CUstream htodStream = NULL;
 static CUstream memcpyStream = NULL;
+#if OPAL_CUDA_GPU_MEM_WAR
+static int mca_common_cuda_gpu_mem_check_war = 1;
```
jsquyres: Shouldn't this *not* be protected by an `#if`, and instead just set the C value to the macro?

```
static int mca_common_cuda_gpu_mem_check_war = OPAL_CUDA_GPU_MEM_WAR;
```

That means that this value is *always* available, but its default value is dependent on the version found in `cuda.h`.

opal/mca/common/cuda/common_cuda.c
```diff
@@ -774,6 +786,11 @@ static int mca_common_cuda_stage_three_init(void)
                             "CUDA: cuMemHostRegister OK on test region");
     }
 
+#if OPAL_CUDA_GPU_MEM_WAR
```
jsquyres: Same logic as above applies to this `#if` (and the `#if` below): if you make it 100% runtime, it might be a little less confusing when one user says "Hey, I have this CUDA MCA param," but another user says "Hey, why don't I have that same CUDA MCA param?  We have the same versions of OMPI, right?"

jsquyres:
Last comment: does CUDA provide an ABI, such that you can compile/link OMPI against CUDA v6, and then change your `LD_LIBRARY_PATH` and run with OMPI run-time linked against CUDA v7?

I.e., is the compile-time detection of the version sufficient, or do you need a run-time detection of which CUDA library version you are running under?  Better yet, can you run-time detect the bad *behavior* (instead of the version)?

opal/mca/common/cuda/common_cuda.c
rolfv: WAR is used internally here.  It is short for workaround.  Updated commit will have better message.

opal/mca/common/cuda/common_cuda.c
rolfv: I would like to leave at 9.  I am basically adding this MCA parameter to give myself an easy way to turn off this workaround. There are performance issues so if folks run into them, I can say, turn of this code and see if that fixes it.  Then I can say, update to CUDA 7.5 and everything will be great again.

opal/mca/common/cuda/common_cuda.c
rolfv: OK, changed it.

opal/mca/common/cuda/common_cuda.c
```diff
@@ -300,6 +303,15 @@ void mca_common_cuda_register_mca_variables(void)
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &mca_common_cuda_cumemcpy_timing);
 #endif /* OPAL_ENABLE_DEBUG */
+
+#if OPAL_CUDA_GPU_MEM_WAR
```
rolfv: Yes, I see.  My thought is eventually, this code will be removed, and by putting the #ifdef around it, it happens naturally.  But I will do as you suggest and then just remove it someday when I can.

opal/mca/common/cuda/common_cuda.c
rolfv: OK.

rolfv:
Updated per review comments.

config/opal_check_cuda.m4
```diff
@@ -143,6 +155,10 @@ AM_CONDITIONAL([OPAL_cuda_gdr_support], [test "x$CUDA_VERSION_60_OR_GREATER" = "
 AC_DEFINE_UNQUOTED([OPAL_CUDA_GDR_SUPPORT],$CUDA_VERSION_60_OR_GREATER,
                    [Whether we have CUDA GDR support available])
 
+# Set a macro to indicate that a workaround is needed for detecting GPU memory
+AC_DEFINE_UNQUOTED([OPAL_CUDA_GPU_MEM_WAR],$CUDA_VERSION_70_OR_LESS,
+                   [Whether we need the WAR for checking GPU memory])
+
```
jsquyres: Is this `AC_COMPILE_IFELSE` and `AC_DEFINE_UNQUOTED` actually necessary?

You could just:

```c
static int mca_common_cuda_gpu_mem_check_workaround = (CUDA_VERSION > 7000) ? 0 : 1;
```

rolfv:
And I guess this is why we have reviews.  Latest submitted.

rolfv:
How about a thumbs up from @jsquyres

jsquyres:
:+1:

PR #838, opened 2015-08-28
mpi conformance: fix prototypes

ggouaillardet:
- MPI_Compare_and_swap
- MPI_Fetch_and_op
- MPI_Raccumulate
- MPI_Win_detach

Thanks to Michael Knobloch and Takahiro Kawashima for bringing this
to our attention

### DIFF
--- a/ompi/include/mpi.h.in
+++ b/ompi/include/mpi.h.in
@@ -17,6 +17,8 @@
  *                         reserved.
  * Copyright (c) 2011-2013 INRIA.  All rights reserved.
  * Copyright (c) 2015      University of Houston. All rights reserved.
+ * Copyright (c) 2015      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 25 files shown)


ggouaillardet:
@hjelmn could you please double check this ?

hjelmn:
Yup. Those are definitely wrong. Probably has to do with the MPI-3 RMA stuff coming in after the const change.

hjelmn:
PR to 1.10 and 2.x once merged.

hjelmn:
BTW, thank you for making the changes in the osc framework. We still need to make the changes for the other MPI interfaces (pt2pt, io, etc).

ggouaillardet:
where are the remaining changes to be made ?
i initially made mimimum changes at the MPI level, and then got a bunch of warnings/errors in the osc framework. now, there is no more warnings

kawashima-fj:
grep -r XXX ompi/mpi/c
and
grep -r TODO ompi/mpi/c

You'll see comments like "do not cast away const".
ompi/mca/osc/osc.h still need to be update for put, accumulate etc.

ggouaillardet:
i see, i am on it now, thanks !

PR #845, opened 2015-08-31
Remove --enable-mpi-profile configure option (i.e., always build PMPI bindings)

ggouaillardet:
This PR does the following:

* Remove the `--enable-mpi-profile` configure option (i.e., the PMPI bindings are now *always* built) and associated configury (because we can now *always* assume that PMPI is built)
  * Full disclosure: deleting this option will lead to longer build times on OS X if you are currently using `--disable-mpi-profile` to shorten your build time.
* Change all calls to `MPI_` functions to be `PMPI_` functions (again, because we can *always* assume that the PMPI functions will be available)
* *Slightly* simplify the `#if` usage in ompi/mpi/*/*.c files


ompi/mpi/fortran/mpif-h/Makefile.am
```diff
@@ -28,15 +28,19 @@ include $(top_srcdir)/Makefile.ompi-rules
...
 # whether "profile/defines.h" is included or not. "profile/defines.h"
 # replaces all MPI_* symbols with PMPI_* symbols. In this directory,
-# we need it to be 0
+# we need it to be 1 when configure'd with --enable-mpi-profile
 #
 
 AM_CPPFLAGS = -DOMPI_PROFILE_LAYER=0 -DOMPI_COMPILING_FORTRAN_WRAPPERS=1
 
+if WANT_MPI_PROFILING
+AM_CPPFLAGS += -DOMPI_PROFILING_DEFINES=1
```
jsquyres: I think you'll need the "else" clause here, too, right?  I.e., since you use `if OMPI_PROFILING_DEFINES`, you need to set `-DOMPI_PROFILING_DEFINES=0` in the false case.

ggouaillardet: you are right, I will update the pr tomorrow when the sun sets in Japan...

jsquyres:
Github shows an abbreviated diff because so many files changed -- is this PR simply:

1. Changes in mpif-h/Makefile.am and mpif-h/profile/Makefile.am
2. Addition of the `#if / #include` block in each mpif-h/*.c file

I.e., I did not check that you edited every single *.c file.

Also, it strikes me: why do we still support `--disable-mpi-profile`?  I think that probably makes the build logic a bit more complicated than it needs to be.  This flag is a relic of an era of when it took a looooong time to build Open MPI on platforms where there were no weak symbols (e.g., on OS X laptops) -- developers would use this flag to reduce their build times.  It's not a useful flag for end-users.  Should we remove `--disable-mpi-profile`?  (as a different PR, of course)

jsquyres:
Actually, thinking about this for another minute: I wonder if we should remove `--disable-mpi-profile` *before* this PR, because then that would remove the need for the `#if` block in every *.c file that you edited, right?

ggouaillardet:
or I can #include "ompi/mpi/c/profile/defines.h" unconditionally and move the #if ... into defines.h for now.
on osx, we still need to build ompi/mpi/{c,c/profile,mpif-h,mpif-h/profile} so I am afraid removing the --disable-mpi-option will not greatly simplify the build logic

ggouaillardet:
on second thought, if we remove the --enable-mpi-profile option, the right way could be to directly invoke the PMPI_* C functions from the *_f.c files instead of #include "ompi/mpi/c/profile/defines.h" and invoke the MPI_* C functions

jsquyres:
Agreed.  And, even more than that, we can replace any `MPI_` in `ompi/mpi/c/*.c` with `PMPI_`.  There's probably not many of them at all, but there might be a few.

ggouaillardet:
I am puzzled
can you give me one example ?
I do not understand how the MPI_* symbols are generated under OSX if we do that ...

jsquyres:
I'm sorry -- I wasn't clear...

I mean in the `mpi/c/*.c` files where one MPI_ function calls another MPI_ function.  I.e., we still leave the MPI functions named `MPI_`, but in any case -- and there are likely only a few (if any? I have a dim recollection that there *are* some, but I don't know what they are offhand) -- where those functions call another MPI function, we should invoke them via their `PMPI_` name.

ggouaillardet:
i found them 
```
$ for i in ompi/mpi/c/profile/.libs/*.o ; do nm $i | grep 'U PMPI_' && echo $i ; done
                 U PMPI_Comm_create_errhandler
ompi/mpi/c/profile/.libs/perrhandler_create.o
                 U PMPI_Comm_get_errhandler
ompi/mpi/c/profile/.libs/perrhandler_get.o
                 U PMPI_Comm_set_errhandler
ompi/mpi/c/profile/.libs/perrhandler_set.o
                 U PMPI_Alloc_mem
                 U PMPI_Free_mem
                 U PMPI_Sendrecv
ompi/mpi/c/profile/.libs/psendrecv_replace.o
                 U PMPI_Type_create_hindexed
ompi/mpi/c/profile/.libs/ptype_hindexed.o
                 U PMPI_Type_create_hvector
ompi/mpi/c/profile/.libs/ptype_hvector.o
                 U PMPI_Type_create_struct
ompi/mpi/c/profile/.libs/ptype_struct.o
```

i will make a RFC shortly to ask whether it is ok to remove the ```--enable-mpi-profile``` option

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ggouaillardet:
:bot:retest

jsquyres:
@ggouaillardet I think this PR is good to go, except that it now has merge conflicts -- I suspect it's  because of d7e6daf75faa30cbe0da6f82d713fb6339519211 and/or eeddb7abc1ec0ad3cba331903ecf7c8dbac433ff.  Should be easy to fix.

jsquyres:
@ggouaillardet More specifically, when you fix the merge conflicts, we should run this by the community and see if people are ok with deleting `--disable-mpi-profiling` (since it will lead to longer build times on OS X, if lazy developers are using this option to shorten their build times...).

lanl-ompi:
Test FAILed.

ggouaillardet:
:bot:retest

lanl-ompi:
Test FAILed.

ggouaillardet:
@jsquyres i rebased and fixed this PR
i already made a RFC one month ago http://www.open-mpi.org/community/lists/devel/2015/09/17960.php
only Ralph replied and he is ok with removing this option

jsquyres:
@ggouaillardet Looks like a false failure at the LANL-Cray XC jenkins.

Per the RFC: ok.  I'll mention it on the call later today anyway, just to make sure everyone really is aware that this is happening...

jsquyres:
Everyone on the call today was receptive.  I'll merge.

jsquyres:
@ggouaillardet Can you PR this over to v2.x?

PR #855, opened 2015-09-02
Fix for openib btl mca command line parameter btl_openib_mtu being

matcabral:
Openib btl command line parameter to set the mtu was being ignored, only the ini file values were being taken into account. With the fix the precedence order is: command line, ini file specifics, ini file defaults.


rhc54:
@hjelmn this looks eerily familiar - IIRC, our prior attempts ran into problems due to MPI_T needing to be able to change the value at some point. Can you please verify that this doesn't cause that problem again?

hjelmn:
:+1:

Issue #857, opened 2015-09-02
Disable mpi_f08 if "USE ... ONLY" does not work


jsquyres:
Per thread starting here: http://www.open-mpi.org/community/lists/devel/2015/08/17867.php, the PGI compiler (at least as of v15.7) does not fully support the "USE ... ONLY" notation.

We decided in the Tuesday teleconf and followup emails with Nvidia to add a configure test to check and see if the compiler supports this behavior.  If it does not, disable the mpi_f08 bindings.

jsquyres:
Tagging @rolfv because he's the nvidia/PGI contact.

PR #858, opened 2015-09-02
fortran configiry: test for USE...ONLY support

jsquyres:
As of v15.7, the PGI Fortran compiler does not properly support how Open MPI uses the `USE ... ONLY` Fortran syntax to include modules with conflicting symbol definitions (interestingly, pgfortran only has a problem with this when compiling with `-g`).

In short, OMPI uses `USE :: module_aaa, ONLY: foo` and `USE :: module_bbb, ONLY: bar` to use modules aaa and bbb, even though they contain conflicting definitions for some symbols.  However, the use of the ONLY clause should preclude the inclusion of the conflicting symbols -- as the word implies, it should direct the compiler to *only* use the symbols identified by the clause (i.e., foo and bar, in this example).

This commit adds a configure test for this capability.  If the compiler fails to build a simple test that mimics this behavior, then disable the mpi_f08 bindings.

Fixes open-mpi/ompi#857

@rolfv please verify
@PHHargrove FYI

### DIFF
--- a/config/ompi_fortran_check_use_only.m4
+++ b/config/ompi_fortran_check_use_only.m4
@@ -1,0 +1,61 @@
+dnl -*- shell-script -*-
+dnl
+dnl Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+dnl                         University Research and Technology
+dnl                         Corporation.  All rights reserved.
+dnl Copyright (c) 2004-2005 The University of Tennessee and The University
+dnl                         of Tennessee Research Foundation.  All rights
+dnl                         reserved.
+dnl Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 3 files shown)


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

rolfv:
I have configured with --enable-debug and saw that the f08 bindings were correctly disabled.  Then I configured without --enable-debug, and I correctly saw the f08 bindings were built.  I tested basic f08 test and it worked.  So, this looks good to go to me.
:+1:

jsquyres:
Sweet; thanks @rolfv

PR #859, opened 2015-09-02
ML: Replace opal ignore with a zero priority

shamisp:
The priority set by default to -1. As a result component open reports
an error and the component is not loaded (no resources allocated).

### DIFF
--- a/ompi/mca/coll/ml/coll_ml_mca.c
+++ b/ompi/mca/coll/ml/coll_ml_mca.c
@@ -205,7 +205,7 @@ int mca_coll_ml_register_params(void)
     /* register openib component parameters */
 
     CHECK(reg_int("priority", NULL, "ML component priority"
-                  "(from 0(low) to 90 (high))", 27, &mca_coll_ml_component.ml_priority, 0));
+                  "(from 0(low) to 90 (high))", 0, &mca_coll_ml_component.ml_priority, 0));
 
     CHECK(reg_int("verbose", NULL, "Output some verbose ML information "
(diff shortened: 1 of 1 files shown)


shamisp:
@jsquyres @bosilca Please review.

rhc54:
forgive me for asking, but i thought i saw two commits already go into the master that did this. am i missing something?

bosilca:
The ml_priority is used in the coll_ml_component.c (line 311) to prevent the component from opening. Above this code there is a comment that states that for 0 the component is disabled. So I guess that setting the priority to zero also qualifies as a solution, while leaving the comment valid. If you want to keep the priority negative please update the comment in the opal function.

rhc54:
I see my confusion - you put it as a branch on master, and then deleted it. sorry for confusion.

shamisp:
@rhc54 - by mistake I pushed my branch to the "origin" instead of my "fork". Obviously, I removed it the same moment. Sorry about this.

lanl-ompi:
Test FAILed.

hppritcha:
Igno the LANL-disable-dlopen-check.  I had to kill off the test job owing to a hang.  The hang is not related to this PR.

shamisp:
What is the policy for merge ? I guess somebody should  :+1: this PR.

jsquyres:
You should probably squash these down into a single commit.

shamisp:
@jsquyres done

jsquyres:
@shamisp Now your commit message doesn't agree with the commit...

shamisp:
@jsquyres - Oh, missed this. I update the commit with a new name.

jsquyres:
:+1:

PR #865, opened 2015-09-03
While debugging user problems, these extra verbosity statements would be helpful

rolfv:
After debugging issues with users, it is clear that these extra verbose messages could be helpful.  I decided to have them in the release build because most of them are off the critical path.

@jsquyres would you take a look?


gpaulsen:
+1

jsquyres:
Verbose = good.  :+1:

PR #868, opened 2015-09-04
Remove OPAL_HAVE_HWLOC qualifier and error out if --without-hwloc is given

rhc54:


hppritcha:
@jsquyres  I think this should be pushed by to v2.x to reduce problems with merge conflicts.  What do you think?

ompi/mca/topo/treematch/configure.m4
```diff
@@ -34,13 +35,6 @@ AC_DEFUN([MCA_ompi_topo_treematch_CONFIG], [
...
     ompi_topo_treematch_happy=0
 
-    AC_MSG_CHECKING([for hwloc support])
-    AS_IF([test $OPAL_HAVE_HWLOC -eq 0],
-          [AC_MSG_RESULT([no (treematch requires hwloc)])
-           ompi_topo_treematch_happy=0],
-          [AC_MSG_RESULT([yes])
-           ompi_topo_treematch_happy=1])
-
     AS_IF([test $ompi_topo_treematch_happy -eq 1],
```
ggouaillardet: i gues this test should be removed since ompi_topo_treematch_happy cannot/should be 1 any more

jsquyres:
@hppritcha Agreed.  Let's let it soak on master for a short while, first.

PR #872, opened 2015-09-08
pmix/s1: fixes for s1 to work with pmix1x

hppritcha:
Discovered while trying to use master on
a truescale/psm system that recent pmix1x
work broke support for s1.  If master was
built with --with-pmi option, mpirun would
segfault almost immediately as it tried to
use pmix/s1 rather than the pmix/pmix1x.

Other fixes to handle kvs query with keys
which had not yet been stored.  Otherwise
when using srun, users see warning messages
which confuse them.

### DIFF
--- a/opal/mca/pmix/s1/pmix_s1.c
+++ b/opal/mca/pmix/s1/pmix_s1.c
@@ -127,9 +127,18 @@ static char* pmix_error(int pmix_err);
 static int kvs_get(const char key[], char value [], int maxvalue)
 {
     int rc;
+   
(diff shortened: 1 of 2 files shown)


rhc54:
:+1:

rhc54:
:-1: 

I didn't realize it right away, but you cannot use ORTE_PROC_IS_APP in the opal layer. I'm pulling the relevant pieces from here and will provide a replacement PR tonight.

hppritcha:
closing this as a different work-around was found.

PR #873, opened 2015-09-08
Add the libs required for PMIx to support static builds (and trim all excess whitespace)

rhc54:


jsquyres:
@rhc54 Do the wrappers really need the `-I`'s for pmix?

rhc54:
@jsquyres When I run the whitespace purge, it hits about a few (small number) files that this PR doesn't touch. Want me to just include everything here?

jsquyres:
I edited the title, and now it says that that is exactly what you are supposed to do.  :smile:

PR #884, opened 2015-09-10
Fix the handling of cpusets so we get the correct cpuset for each loc…

rhc54:
…al peer. Add the ability to indicate that a modex request is "optional" so we don't call the server if we don't find the value. Take advantage of that to allow the MPI layer to decide that the lack of locality info indicates non-local


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

PR #885, opened 2015-09-10
pmix/~pmix1xx: use u32 for OPAL_PMIX_LOCAL_SIZE

hppritcha:
Looks like in ess_pmi_module.c u32 is being used
for retrieving OPAL_PMIX_LOCAL_SIZE, while s1/s2/cray
pmix components were storing as u16.


rhc54:
This looks fine - I'm going to play with s1 tonight, so I'm going to bring this into the master now.

PR #886, opened 2015-09-10
opal/hwloc: fix topology detection when socket is above numa

hjelmn:
The OPAL_PROC_ON_* definitions have been changed from values to
flags. This should not cause any problems as these values were already
used as flags throughout the code base. Note, there will be a
difference between localities produced by the new code and the
old. For example, if a machine does not have a level-3 but two cores
share a level-1 or level-2 cache cache the level-3 bit will not be set
in the locality and OPAL_PROC_ON_LOCAL_L3CACHE will return 0. Before
this change it would have returned 1.

In addition the OPAL_PROC_ON_LOCAL_* macros have been simplified.

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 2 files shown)


rhc54:
@hjelmn did something get left out of this? I don't see where things change when numa and socket are inverted.

hjelmn:
@rhc54 Now that is is a set of flags if NUMA is before SOCKET and two cpu sets share a socket then both the NUMA and SOCKET flags will be set. If SOCKET is before NUMA then only the SOCKET flag will be set.

rhc54:
@hjelmn guess I'm just not seeing it. I don't see where ordering has any impact now - you cycle thru all the levels, doing an "or-equal" for the bit of any level where the two cpusets overlap. So how does the order of the depth get considered?

hjelmn:
@rhc54 It doesn't need to be considered when using flags. On my system the highest level of overlap between cores 0 and 4 is the socket. So as the loop dives down the socket flag will be set when the HWLOC_OBJ_PACKAGE object is hit but nothing else will be set. So, OPAL_PROC_ON_LOCAL_SOCKET will be true but OPAL_PROC_ON_LOCAL_NUMA will be false.

Cores 0 and 1 are on the same numa node so they will get the socket flag when encountering HWLOC_OBJ_PACKAGE and then the numa flag when the overlapping HWLOC_OBJ_NODE object is encountered.

rhc54:
@hjelmn okay, that is what i thought - i misunderstood what you were saying as it implied there was some kind of level detection going on.

:+1:

hjelmn:
@rhc54 Initially, yes. But this is much cleaner.

PR #889, opened 2015-09-11
Use the low instead of the high bit to indicate a proc is a sentinel

hjelmn:
The assumption that the high bit is not in use in pointers on any of our
supported platforms was incorrect. A better assumption is that all
ompi_proc_t pointers will be at least 2-byte aligned. This allows us
to use the low bit. To do this we drop the highest bit of the
opal_process_name_t jobid (hope this is ok) and use the low bit to
indicate the proc is really a sentinel.

### DIFF
--- a/ompi/group/group.c
+++ b/ompi/group/group.c
@@ -562,7 +562,7 @@ bool ompi_group_have_remote_peers (ompi_group_t *group)
 #if OMPI_GROUP_SPARSE
         proc = ompi_group_peer_lookup (group, i);
 #else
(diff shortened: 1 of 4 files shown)


hjelmn:
@bosilca, @ggouaillardet This should fix the problem. I also moved the sentinel check into it's own function to make updating the sentinel easier in the future.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

hjelmn:
Fixing.

PR #893, opened 2015-09-14
Cleanup handle verbose messages

rolfv:
Some updates due to @jsquyres review on release branch.

PR #894, opened 2015-09-14
oshmem/memheap: correct usage of opal_dss functions

hjelmn:
Any buffer given to opal_dss.load becomes the responsibility of the
opal_buffer_t object. It will be freed automatically if either the
opal_buffer_t is released or opal_dss.load is called again on the
buffer. opal_dss.unload will not prevent this unless no unpacking
takes place between the .load and .unload calls.

### DIFF
--- a/oshmem/mca/memheap/base/memheap_base_mkey.c
+++ b/oshmem/mca/memheap/base/memheap_base_mkey.c
@@ -1,9 +1,12 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2013-2015 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2015      Intel, Inc. All rights reserved
+ * Copyright (c) 2015      Intel, Inc. All rights reserved.
+ * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -411,7 +414,6 @@ static int oshmem_mkey_recv_cb(void)
              */
             do_recv(status.MPI_SOURCE, msg);
             OBJ_RELEASE(msg);
-            free(tmp_buf);
         }
 
         rc = MPI_Start(&r->recv_req);
@@ -566,12 +568,12 @@ void mca_memheap_modex_recv_all(void)
     opal_buffer_t *msg = NULL;
     void *send_buffer = NULL;
     char *rcv_buffer = NULL;
-    void *dummy_buffer = NULL;
-    int size, dummy_size;
+    int size;
     int *rcv_size = NULL;
     int *rcv_n_transports = NULL;
     int *rcv_offsets = NULL;
     int rc = OSHMEM_SUCCESS;
+    size_t buffer_size;
 
     if (!mca_memheap_base_key_exchange) {
         oshmem_shmem_barrier();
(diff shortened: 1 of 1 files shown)


hjelmn:
@miked-mellanox This fixes the opal_dss usage problem that came up in #893.

jsquyres:
Adding @rolfv, too, since it affects the (probably-false) jenkins failure on #893.

miked-mellanox:
@hjelmn - was there change in behavior for opal_dss.*load() functions?
Why it started from Sept12?

hjelmn:
@miked-mellanox Yes. The behavior changed in open-mpi/ompi@22d7c0081a58c39408adad3c72fa73549c530083. @rhc54 Was this behavior change intended.

miked-mellanox:
thanks.
This PR: #892 which broke master, has failed on exact same failure w/ jenkins (http://bgate.mellanox.com:8888/jenkins/job/gh-ompi-master-pr/1070/) , but was merged anyway.

hjelmn:
Ah, so jenkins caught it but the failure was ignored?

hjelmn:
Looks like the new behavior does not match the documentation:

```
/**
 * Unload the data payload from a buffer.
 *
 * The unload function provides the caller with a pointer to the data
 * payload within the buffer and the size of that payload. This allows
 * the user to directly access the payload - typically used in the RML
 * to unload the payload from the buffer for transmission.
 *
 * @note This is a destructive operation. While the payload is
 * undisturbed, the function will clear the buffer's pointers to the
 * payload. Thus, the buffer and the payload are completely separated,
 * leaving the caller free to OBJ_RELEASE the buffer.
 *
 * @param buffer A pointer to the buffer whose payload is to be
 * unloaded.
 *
 * @param payload The address to a void* pointer that is to be loaded
 * with the address of the data payload in the buffer.
 *
 * @param size The size (in bytes) of the data payload in the buffer.
 *
 * @retval OPAL_SUCCESS The request was succesfully completed.
 *
 * @retval OPAL_ERROR(s) An appropriate error code indicating the
 * problem will be returned. This should be handled appropriately by
 * the caller.
 *
 * @code
 * opal_buffer_t *buffer;
 * uint8_t *bytes;
 * int32_t size;
 *
 * status_code = opal_dss.unload(buffer, (void**)(&bytes), &size);
 * OBJ_RELEASE(buffer);
 * @endcode
 */
```

hjelmn:
This PR is obviously not the correct fix unless we want to change the documentation of opal_dss.unload. Lets keep it open for discussion.

hjelmn:
Though, part of this patch is correct. I will keep that part.

rhc54:
I disagree - the patch is correct and necessary. The problem is that people aren't using it correctly and never did. Let's get on the phone and talk this thru before doing something wrong

hjelmn:
After talking with @rhc54 I think the behavior of load/unload is correct. We need to update the documentation at some point so the behavior is clear.

I pushed an updated version of the fix. @miked-mellanox if you have no objections I will merge once the checks have passed.

hjelmn:
Misc failure.

:bot:retest:

hjelmn:
Mellanox jenkins did not rerun. @miked-mellanox can you see what is going on?

hjelmn:
:bot:retest:

miked-mellanox:
failed on slurm "srun" launcher we added yesterday. will disable it for now.
The patch seems to fix the issue.

miked-mellanox:
bot:retest

miked-mellanox:
shall it be ported to v2.0 as well?

hjelmn:
The patch should probably be moved to 2.x since the patch that changed the behavior of opal_dss will probably be moved to 2.x.

rhc54:
@hjelmn @miked-mellanox Yeah, I'll add it to my PR since it was changed there. Thanks!

PR #895, opened 2015-09-14
Search for lstat in sys/stat.h

wthrowe:
According to the Linux man page, both sys/stat.h and unistd.h are
supposed to be included before using lstat, and, at least on some
machines, sys/stat.h is actually the necessary one.

This fixes a strange linking failure on one of my systems for reasons that I'm not completely clear on.

### DIFF
--- a/ompi/mca/io/romio314/romio/configure.ac
+++ b/ompi/mca/io/romio314/romio/configure.ac
@@ -1720,7 +1720,8 @@ fi
 AC_CHECK_FUNCS(lstat)
 if test "$ac_cv_func_lstat" = "yes" ; then
     # Do we need to declare lstat?
-    PAC_FUNC_NEEDS_DECL([#include <unistd.h>],lstat)
+    PAC_FUNC_NEEDS_DECL([#include <unistd.h>
+                         #include <sys/stat.h>],lstat)
 fi
 AC_CHECK_FUNCS(readlink)
 if test "$ac_cv_func_readlink" = "yes" ; then


hjelmn:
Has this been fixed upstream?

wthrowe:
Probably not.  Where is upstream?

hjelmn:
mpich is the upstream source. https://www.mpich.org/

In general we pull romio changes directly from mpich.

wthrowe:
I will go report it there, then.  I don't consider this to be a high priority patch, as I seem to be the only one this causes problems for, so I'm just going to close this and let the fix get pulled into OpenMPI by whatever your usual merging procedure is.

hjelmn:
Thanks. Once they have fixed it up stream please let us know and we will pull the appropriate commit and PR it to the affected release branches.

wthrowe:
Upstream commit: http://git.mpich.org/mpich.git/commitdiff/7b8bd055dfdeb

jsquyres:
@hjelmn You want to pull that from ROMIO's upstream?

hjelmn:
Will merge the upstream commit once jenkins finishes.

jsquyres:
@hjelmn Thanks for committing. However, I see that @wthrowe was not cited in the commit message -- can you please be sure that he is cited in NEWS?  His name appears to be William Throwe. Thanks.

hjelmn:
The commit appears with his name and email address. Is that not sufficient?

jsquyres:
Doh, I totally missed that.  (we'll still cite him in NEWS, too)

hjelmn:
I agree on the NEWS.

BTW, to get the upstream patch I downloaded the email version from mpich then changed all the paths to point at where romio is in Open MPI and updated the commit message to indicate the original commit. This patch then was applied using git am. This preserved not only the original author but also the signed off by line.

PR #896, opened 2015-09-15
ompi/comm: fix bug in ompi_comm_set

hjelmn:
This commit updates the behavior of ompi_comm_set to explicitly take
either local/remote group(s) OR local/remote array(s). If array(s) are
in use the sizes will be taken from the appropriate group(s).


hjelmn:
@rhc54 Please verify that this fixes the dynamics issue.

rhc54:
:+1:

PR #901, opened 2015-09-16
ompi/comm: fix comm_[i]dup on intracommunicators

hjelmn:
The behavior of ompi_comm_set was changed to get the remote size from
the remote group. This broke how ompi_comm_[i]dup were using
ompi_comm_set. In order to adapt to the new behavior these functions
now pass NULL for the remote group if the communicator is not an
inter-communicator.


hjelmn:
@edgargabriel This fixes the issue you discovered this morning.

rhc54:
@hjelmn This almost completely fixes master - I see only one remaining fail in IBM test suite, and one timeout in Intel test suite. Well done!

:+1:

PR #904, opened 2015-09-17
hwloc: do not count not allowed cores in df_search_cores

ggouaillardet:


rhc54:
:+1:

rhc54:
@ggouaillardet this needs to go into 1.10 and 2.x - could you please log a PR against each of those?

Thanks again for chasing it down.

ggouaillardet:
will do tomorrow

PR #907, opened 2015-09-18
odls: close only used file descriptors at fork/exec

plesn:
On some machines the time to fork/exec the MPI ranks is really really
slow in the loop closing file descriptors before exec. This loop can
call `close()` for a million of fds in each child creation, but all
this sequentially.

I propose to close only the open file descriptors, it seems better in
all cases. I get those fds from `/proc/$pid/fd/`.

For example on a haswell node with 20 cores at 2,3 Mhz with rhel7 I
get:

    $ time mpirun -n 20 --bind-to core /bin/true | grep real
    real	0m2.211s  # before
    real	0m0.158s  # after

With mannycore it can get worse.

### DIFF
--- a/orte/mca/odls/default/odls_default_module.c
+++ b/orte/mca/odls/default/odls_default_module.c
@@ -101,6 +101,9 @@
 #ifdef HAVE_SYS_SELECT_H
 #include <sys/select.h>
 #endif
+#ifdef HAVE_DIRENT_H
+#include <dirent.h>
+#endif
 
 #include "opal/mca/hwloc/hwloc.h"
 #include "opal/mca/hwloc/base/base.h"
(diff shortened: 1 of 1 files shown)


jsquyres:
In principle, this looks like a nice optimization -- thanks!

Minor OMPI style nits:

* OMPI uses 4-space tabs (always spaces; never the TAB character)
* OMPI always puts {} around blocks, even if they're only 1 line

Can you update your PR for these style nits?  Thank you!

hppritcha:
i'd like to sanity check this on my mac before it gets checked in.  The error handling looks correct but still...

rhc54:
Are we sure this is truly portable to all the various Linux flavors? What happens on things like NetBSD?

rhc54:
I guess as long as we fall back to the old method, we should be no worse than today, so unless the error conditions do something weird, I think this makes sense.

Thanks!

hppritcha:
works on my mac

rhc54:
Kewl then   :-1:

rhc54:
Egad, it reversed direction on me!! I meant :+1:

hppritcha:
@plesn I'm using your changes in the odls/alps and notice when I use --enable-picky that there are
compiler warnings.

Could you configure with --enable-picky option and make sure that you cleanup any compiler warnings you get?

PR #914, opened 2015-09-21
BUILD: add support for platform specific patches

miked-mellanox:


miked-mellanox:
bot:retest

miked-mellanox:
@jsquyres - could you please review? Thanks

ggouaillardet:
@miked-mellanox i guess if you have platform patches and you run ```configure``` twice, bad things will happen on the second ```configure```. best case scenario is ```patch``` will complain the patches have already been applied, worst case scenario is ```patch``` will fail and leave the sources in an undefined state.
also, i do not think that can work if configur'ing with ```VPATH```
imho, the minimum requirement is to
- handle ```VPATH```
- apply the patches once only (for example, ```touch $top_builddir/.platform_paches``` after the patches are applied, and do not re-apply the patches if this file exists. of course, this does not handle the case in which you modified your patches and run ```configure``` again)

rhc54:
I'm still scratching my head trying to understand why we need this change? If you want to change the platform file, why not just change it? We have no restrictions on doing so.

miked-mellanox:
@ggouaillardet , @rhc54 - thanks for comments.

@ggouaillardet - the patch command used in the PR has following flags which should cover most of the cases you describe:

``` shell
  -t  --batch  Ask no questions; skip bad-Prereq patches; assume reversed.
  -s  --quiet  --silent  Work silently unless an error occurs.
```

but having sentinel indicator as you suggested is a better way.

@rhc54 - the use case is as following:

* we (or any vendor) may have a set of patches which cannot be submitted or accepted. We would like to distribute them as "contrib" quality patches and have a way to apply them on site. The alternative is to maintain vendor repo with such patches which is comfortable.

* we have couple of patches which are workarounds for mxm limitations and needed to change in ompi_finalize path. This patch will not be accepted but for mxm-only build it is a real help.

miked-mellanox:
@ggouaillardet, @jsquyres , @rhc54 - could you please review. thanks

config/opal_load_platform.m4
```diff
@@ -97,6 +97,27 @@ AC_DEFUN([OPAL_LOAD_PLATFORM], [
...
+            if [ test ! -f "${patch_done}" ]; then
+                OPAL_LOG_MSG([Applying patches from ${patch_dir}])
+                for one_patch in $patch_dir/*.patch ; do
+                    OPAL_LOG_MSG([Applying patch: $one_patch])
+                    patch -d ${srcdir} -p1 -t -s < ${one_patch}
+                    if [ test "$?" != "0" ]; then
+                        rm -f ${patch_done}
+                        AC_MSG_ERROR([Failed to apply patch ${one_patch}])
+                    fi
+                    touch ${patch_done}
```
jsquyres: Shouldn't this "touch" be outside the "for" loop?  (i.e., after the "done")

jsquyres:
I second @ggouaillardet's concern about patching anything involving configury.  Do you want to apply patches in autogen.pl, instead?  (where we already apply other patches / it's safe to patch configury -- *before* the GNU Autotools are run)

config/opal_load_platform.m4
miked-mellanox: if no *.patch files (line 105) exist - "touch" will be done anyway than.

"touch" should take place only if "for" loop kicked-in.

miked-mellanox:
right -  "--with-platform" should not contain patches touching configury but code. I think this limitation can be ok?

if it will be done in autogen, do you see it as autogen.pl flag to specify "--with-platform" like flag pointing to the contrib/... location?

``` shell
$ ./autogen.pl --with-platform=...
$ ./configure --with-platform=...
```

config/opal_load_platform.m4
jsquyres: Mmm.  I see what you did there.

But it also gets touched if *one* patch is done, and others fail (even though you remove it if a failure occurs).

Perhaps a slightly cleaner way (that wouldn't be subject to Ctrl-C false-failures) could be:

```sh
patched=0
for one_patch ...
   OPAL_LOG_MSG(...)
   patch -d ...
   if test $? -ne 0; then
      AC_MSG_ERROR([...])
   fi
   patched=1
fi
if test patched -eq 1; then
    touch ${patch_done}
fi
```

config/opal_load_platform.m4
```diff
@@ -97,6 +97,27 @@ AC_DEFUN([OPAL_LOAD_PLATFORM], [
             AC_SUBST(OPAL_PARAM_FROM_PLATFORM, "no")
         fi
 
+        patch_dir="${with_platform}.patches"
+        patch_done="${srcdir}/.platform_patches"
+        if [ test -d "${patch_dir}" ]; then
```
jsquyres: Isn't "[ test" redundant?  I.e., don't you just need "if [ ...]" or "if test ..."?

jsquyres:
@miked-mellanox Regarding autogen: how about `--with-patchdir=DIR`?  It doesn't have to be tied to a platform file.  That way, if DIR exists, you can apply any `*.patch` file found in it.

Does that work for you?

config/opal_load_platform.m4
miked-mellanox: if one patch succeeded (touch happened) and next failed - line 109 will remove marker before failing.
Using flag as you suggested maybe cleaner way.

config/opal_load_platform.m4
miked-mellanox: right, should work too.

miked-mellanox:
maybe we can have both options? Here is a rationale:

* ompi.src.rpm or ompi.src.tar.gz downloaded from the open-mpi.org can be configured up to vendor required patch-level with "./configure --with-platform=xxx" (which will apply vendor patches as well)

* having only autogen based command-line "--with-patchdir", customer will need to go through developer routine and install specific tools versions for "autogen", while with "configure" based flag it is not required.

* The limitation posed by "configure" provided patch flag is not important (IMO) and rarely used.

jsquyres:
@miked-mellanox Good point about customer builds -- yes, we definitely don't want them to have to run autogen / the GNU Autotools.   I think my only real concern is that if we tie patches to --with-platform, someday, someone might accidentally put in a configury-changing patch.  Chaos will ensue, but potentially not on the first build -- making it difficult to spot such changes.  Is there a way we can check to see if any of the applied patches will cause a re-invocation of the GNU autotools?  I.e., could we have configure AC_MSG_ERROR if a patch touches something that requires re-running autogen?

jsquyres:
Hmm. The more I think about this, the more questions I have...

Do you plan for customers to build with Mellanox-specific patches from tar balls downloaded from open-MPI.org?  I had thought you were talking about tar balls downloaded from Mellanox.com. 

I'm also still concerned about patching configury - it has happened before, and it will happen again (especially if there is a general infrastructure mechanism that allows it). For a configure-based solution, we need a way to detect that a bad patch has gone in and abort the process.

miked-mellanox:
yes, from ompi tarball.

it is probably doable detect which files are affected by patch and exit with error if configury files are affected.

miked-mellanox:
@jsquyres - could you please review

jsquyres:
Code looks good; thanks.

I'm generally ok with this concept (infrastructure for applying patches).  This can be useful for organizations to create their own Open MPI distributions.

I do have some meta concerns with including vendor-specific patches in community source distributions, however.

This basically makes Open MPI community tarballs be able to be arbitrary vendor tarballs.  ...but how can users tell the difference?  E.g., today, when someone has a problem with (VENDOR) Open MPI, we can tell because they tell us that they downloaded it from VENDOR's web site.  

Additionally, we strongly encourage vendors to use the `--with-ident-string='VENDOR BLAH'` configure CLI option so that users can see right in the first lines of `ompi_info` (e.g., `--with-ident-string='Cisco Open MPI'` results in "Cisco Open MPI" in the first few lines of `ompi_info` output) to distinguish this installation from community Open MPI.

This patch mechanism, now that it is now actively changing the code base, is now actively diverging from community Open MPI.  A few concerns:

1. In some ways, including vendor patches in the community distribution tarballs is an implicit community endorsement of those vendor's patches.
1. Is there a reason the patches can't be upstreamed?  (this is somewhat rhetorical -- I recognize that there are cases -- e.g., in the v1.8 series -- where patches won't be accepted upstream)
1. If we do allow this patching mechanism *and* include vendor patches in the community official distribution tarballs / SRPM, it needs to clearly identify that this is no longer community Open MPI, and is instead (VENDOR) Open MPI.

miked-mellanox:
@jsquyres - thanks for comments, sounds reasonable to request vendor ident string for vendor patches. (did not know about --with-ident-string flag, sounds cool!)

can add it automatically if platform patches detected, what to you think?

miked-mellanox:
@jsquyres - could you please review. Now it adds custom brand if applied vendor patches.

regrant:
While I think this patching infrastructure is a great idea, I share @jsquyres concerns about including the vendor patches in the community tarball. It seems to me that this is something that users will find useful though, so having them download the patches they want from another website seems reasonable. Otherwise I can see some potential confusion about OMPI versions among users.

I think this could also be useful for patches that prototype new proposals for MPI Forum but are not yet part of the standard. So I'm in favour of the infrastructure overall as I think it would be useful to a larger group than just the vendors.

miked-mellanox:
yep, now this PR adds vendor branding if vendor patch was detected and applied.

miked-mellanox:
all comments addressed. good to go

jsquyres:
@miked-mellanox Can you please add some instructions on the wiki somewhere about how people can use this infrastructure?

I should note, too, that there was a bunch of pushback on the call yesterday about using this mechanism to include organization-specific patches in the community distribution from www.open-mpi.org.  I.e., the feeling was that this infrastructure was good, but the community does not want to include non-upstream patches in tarballs and source RPMs distributed from www.open-mpi.org.  That seems to send a conflicting message.  Such distributions are more appropriate to individual organization distribution mechanisms.

jladd-mlnx:
@gpaulsen would this be something useful for IBM? Some of our customers (I would assume some of yours as well) insist on vanilla distros as opposed to vendor modified distros obtained vendor website. We have patches that simply are not ever going to go upstream. This provides a mechanism to allows vendor specific patches to piggyback on community code. Patches would be contained in the already extant vendor contrib/platform/vendor_foo directory. Please see the additional PR on the release branch for more info https://github.com/open-mpi/ompi-release/pull/621 (I can't reference this in Markdown since it points to the ompi-release repo.) We are very interested to hear IBM's thoughts on this.

rhc54:
@jladd-mlnx I looked at the referenced PR, and I confess I'm puzzled as to why you think this wouldn't come upstream? In fact, we already have a mechanism in the master and 2.x code that does what this PR is trying to accomplish - it provides a means by which a given BTL/MTL can indicate that an extra barrier at finalize is required, and then we execute it. It looks like this particular patch was simply written either before that was enabled, or in ignorance of it. So I agree that the patch as written wouldn't come into the master, but the associated functionality certainly could.

As to the general question: I wasn't at the meeting, but I believe the concern is over how we on the mailing list know what version of OMPI the user is asking about. We already often have to ask for a version string, but now the version string won't be enough - we also have to figure out what patches might have been applied. This gets awfully hard very quickly.

When users get the code from a vendor site, they know (a) it is different from the main distro, and (b) they have to get help from the vendor, not the community. If we now distribute these changes from the community, then the expectation will follow that the community will be supporting them. If the code truly cannot be upstreamed, then how can the community support it?

I think that's what we are wrestling with...

jladd-mlnx:
@rhc54 Understood regarding the mechanism in 2.x, but we need this patch in 1.10.x, and beyond this particular patch, we require this mechanism for future business activities. I think Jeff's suggestion to use the `--with-ident-string='Mellanox Open MPI'` solves this - does it not? Seems to me it's more a philosophical objection, because from a technical standpoint I think all of the hooks are in place to keep track of who is the responsible party.

rhc54:
@jladd-mlnx I think ident-string should help - like I said, though, I wasn't at the meeting and so I don't know the full extent of the concerns that were raised. I suspect it will come up at next week's telecon again, so perhaps it can be addressed there?

jladd-mlnx:
@rhc54 We are happy to discuss this further. I would also be interested in learning more about why Sandia objects to this change. @regrant perhaps you could elucidate your organization's standpoint on this topic. Are these concerns your own opinion or are you speaking on behalf of your organization's production folks?

jsquyres:
It was more than Sandia objecting -- there were several voices on the call that objected to this idea (per my prior comments on this issue, I am one of them).  I guess it boils down to two issues:

1. The community distribution is defined as the code that is upstream.  Patches that are only applied in *some* cases are not really upstream.
1. You said: "Some of our customers ... insist on vanilla distributions."  But if you add vendor/organization-specific patches, then isn't it no longer a vanilla distribution?

Bottom line: it just seems weird to include vendor/organization-specific patches in a community distribution.

To be clear: I'm fine with the infrastructure to apply patches.  I'm not convinced that including vendor/organization patches in the community distro is a good idea.

But by the same token, if you need specific functionality in v1.10, why not create a PR for it?  Remember, we don't have the rigid rules in v1.10 that we did in the v1.8 series.  The patches you have in open-mpi/ompi-release#621 -- with a bit of work (e.g., not unconditional as they are now) -- could probably be accepted by the community.  Be advised that the clock is ticking; as discussed on the mailing list and the call yesterday, the timeline for 1.10.1 is very short.

jsquyres:
@jladd-mlnx @miked-mellanox Aside from the rest of this conversation: you should be using `--with-ident-string` for any Mellanox-specific distribution of Open MPI to clearly delineate it from the community Open MPI distribution.

miked-mellanox:
@jsquyres , @rhc54 - the patch sets ident string automatically if "custom patches" were detected and applied during configure. please see here:
https://github.com/miked-mellanox/ompi-release/commit/df4d2fb341dbe470fd5e114f3b4d40dcbbe31d75#diff-5b754e82c2dc319542d4665727ab1f95R150

jsquyres:
@miked-mellanox Understood.  But please read the rest of my replies: that's not really the main issue.

miked-mellanox:
@jsquyres - lets re-discuss it next call, will gladly join and participate.

IMO, having vendor specific patches will help on many fronts:

* for admins - it is easier to track changes and better distribution model for multi-vendor sites.
* for members - it is better visibility and great patching mechanism

I agree that not every vendor patch should be placed under platform, only ones which has no chance to get accepted upstream because affecting vendor specific sw/hw only.

You have seen three patches we keep as vendor-specific, they all fixing specific issues in fca and mxm. All other stuff we develop we submit upstream because it is interest of all members (including mlnx).

Due to the time constrain for v1.10 release we can submit these 3 patches as PRs, but IMO - the vendor-specific fixes can be accepted as vendor patches when appropriate precautions were taking (as you mentioned: branding and rationale)

jsquyres:
@miked-mellanox See @hppritcha's comment here: https://github.com/open-mpi/ompi-release/pull/621#issuecomment-144827055

I think that we should move in that direction: work to upstream your 3 patches to solve the immediate problem (i.e., for 1.10.1), and then have a deeper discussion to talk about how to support the scenarios that you + Howard are describing.

Sound good?

miked-mellanox:
yep, sounds good!thanks

jsquyres:
@miked-mellanox Ok, cool. Sorry it took so long for us all to grok each other and come up with a plan.

gpaulsen:
@jladd-mlnx, I think the infrastructure to apply patches would be useful for our internal builds.  I'm hopeful that we won't need it and all bugfixes we do can be applied upstream, but then there is reality.

I don't care for the concept of having separate vendor specific patches.  This makes our lives harder, since we would want to support BOTH with and without the Mellanox patches.
I also worry that the .patches path would not get tested as much as the "main" path, and quality might suffer.

I'd prefer to continue to do changes like this either at configure / make time, or better yet at runtime if at all possible.

PR #916, opened 2015-09-21
Fix warnings due to missing const

rolfv:
Noticed some warnings due to other changes so fixing.  Basically, add const.

@bosilca Can you review this?


bosilca:
:+1:

PR #929, opened 2015-09-23
Update add_procs support

hjelmn:
This PR updates the add_procs support to allow mtl and btl modules to signal at run time that the new dynamic add_procs behavior is not supported. This PR is needed to support Portal4 which under certain conditions requires we call add_procs with all the procs in the job.

@jsquyres, @tkordenbrock, @bosilca please comment.

@tkordenbrock let me know if this helps the portals4 btl and mtl implementations.

### DIFF
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -590,42 +590,24 @@ static int mca_bml_r2_add_procs( size_t nprocs,
 static int mca_bml_r2_del_procs(size_t nprocs,
                                 struct ompi_proc_t** procs)
 {
-    int rc;
-    struct ompi_proc_t** del_procs = (struct ompi_proc_t**)
-        malloc(nprocs * sizeof(struct ompi_proc_t*));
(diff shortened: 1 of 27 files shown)


hjelmn:
@tkordenbrock BTW, you need to set the btl flag in your btl init function at the latest. I think this should be ok.

tkordenbrock:
@hjelmn I have updated the Portals4 components and created a pull request on your add_procs branch.

jsquyres:
With a quick review, :+1:

hjelmn:
@jsquyres Thanks! From what I see in the portals4 changes it looks like the flag approach makes it much easier for btl and mtl authors.

lanl-ompi:
Test FAILed.

jsquyres:
Now that #943 has been merged to the master, re-submit to Jenkins, which should fix the disable-dlopen test.

bot:retest

hjelmn:
@bosilca Any comments on this approach for signaling that dynamic add_procs is not available?

bosilca:
the code looks good.

PR #931, opened 2015-09-23
Plug a few memory leaks identified by valgrind

rhc54:


lanl-ompi:
Test FAILed.

rhc54:
No idea what went wrong with that test - it appears to be a system failure instead of something related to this code.

hppritcha:
looks like there was a time out running hello world on the uh server using
this commit

----------

Howard

PR #948, opened 2015-09-27
Resolve the naming issue when direct-launched by PMIx-enabled RMs usi…

rhc54:
…ng a minimal-impact approach. Detect if we were launched via ORTE - if so, then use our standard methods for computing the jobid. If not, then just hash the nspace to create the jobid, and track the jobid <-> nspace correspondece down in the opal/mca/pmix/pmix1xx component. We then do the translation any time a function that passes process names is invoked.


artpol84:
@rhc thanks! We'll check this tomorrow against the SLURM plugin.

PR #952, opened 2015-09-29
ras: remove orte_ras_proc_t and associated code

ggouaillardet:


ggouaillardet:
@rhc54 can you please have a look at it ?
I incidentally found this class is not used when trying to silence a warning

jsquyres:
:+1:

PR #957, opened 2015-09-29
btl/ugni: adjust exclusivity below sm and vader

hjelmn:


hjelmn:
@hppritcha I noticed on Cielito that ugni was being used instead of vader. This fixes that.

PR #961, opened 2015-10-01
BUILD: log applied patches into config.log

miked-mellanox:


jsquyres:
@miked-mellanox I actually liked the AC_MSG_NOTICE -- it is directly in stdout, not hidden away in config.log.  I.e., the messages therefore have a hope of being seen by the user.

I'd be in favor of revering this PR -- keep the messages on stdout.

jladd-mlnx:
@jsquyres our partners specifically request such information be logged in config.log and NOT emitted to stdout.

jladd-mlnx:
@miked-mellanox :+1:

jsquyres:
When you AC_MSG_NOTICE, it also goes to config.log.

Why wouldn't you want the fact that you are applying patches emitted to stdout?  I think making the user aware that patches are being applied is a Very Good Thing.  Are you trying to hide the fact that you are applying patches?

miked-mellanox:
@jsquyres - I also thought that NOTICE goes to stdout and log, but appeared that it goes to stdout only.
I can add AC_NOTICE and OPAL_LOG both calls for specific strings to have important messages in the log and stdout.

Some orgs keep config.log together with installation and having "patches" lines there makes sense for reproducability.

jsquyres:
@miked-mellanox That would be great (both AC_NOTICE OPAL_LOG) -- thanks!

hppritcha:
That's what customer wants, both on stdout and in config.log.  Sorry for confusion about this.

PR #968, opened 2015-10-02
simplify use of sockaddr* structs to work around buffer overflow warning

JohnWestlund:
The current method of moving around the sockaddr data causes compile time buffer overflow warnings:
> [  197s] /usr/include/bits/string3.h:84:3: warning: call to __builtin___memset_chk will always overflow destination buffer [enabled by default]
> [  197s]    return __builtin___memset_chk (__dest, __ch, __len, __bos0 (__dest));

by doing the following:
create a temp `sockaddr` struct pointer that is used in calling parse_uri
`memcpy` 0's over the struct using `sizeof `a `sockaddr_in` or `sockaddr_in6` (which is larger and causes the buffer overflow warning)
then upon return
create a `mca_oob_tcp_addr_t `object
`memcpy` the original `sockaddr` struct over `mca_oob_tcp_addr_t->addr` (which is a `sockaddr_storage` struct)

Instead this patch simplifies the process by creating the `mca_oob_tcp_addr` object first, then passes the pointer to its `sockaddr_storage` struct so the `parse_uri` copies in the data directly.

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp.c
+++ b/orte/mca/oob/tcp/oob_tcp.c
@@ -221,7 +221,7 @@ static void accept_connection(const int accepted_fd,
 static int parse_uri(const uint16_t af_family,
                      const char* host,
                      const char *port,
-                     struct sockaddr* inaddr)
+                     struct sockaddr_storage* inaddr)
 {
     struct sockaddr_in *in;
 
@@ -260,7 +260,6 @@ static int parse_uri(const uint16_t af_family,
 static void process_set_peer(int fd, short args, void *cbdata)
 {
     mca_oob_tcp_peer_op_t *pop = (mca_oob_tcp_peer_op_t*)cbdata;
-    struct sockaddr inaddr;
     mca_oob_tcp_peer_t *peer;
     int rc=ORTE_SUCCESS;
     uint64_t *ui64 = (uint64_t*)(&pop->peer);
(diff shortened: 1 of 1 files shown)


orte/mca/oob/tcp/oob_tcp.c
```diff
@@ -301,8 +301,7 @@ static void process_set_peer(int fd, short args, void *cbdata)
                         ORTE_NAME_PRINT(&pop->peer),
                         (NULL == pop->net) ? "NULL" : pop->net,
                         (NULL == pop->port) ? "NULL" : pop->port);
-    maddr = OBJ_NEW(mca_oob_tcp_addr_t);
-    memcpy(&maddr->addr, &inaddr, sizeof(inaddr));
+    /* memcpy(&maddr->addr, &inaddr, sizeof(inaddr)); */
```
rhc54: Why leave this line commented out here? If it needs to be deleted, then let's delete it.

orte/mca/oob/tcp/oob_tcp.c
```diff
@@ -290,7 +289,8 @@ static void process_set_peer(int fd, short args, void *cbdata)
         }
     }
 
-    if ((rc = parse_uri(pop->af_family, pop->net, pop->port, (struct sockaddr*) &inaddr)) != ORTE_SUCCESS) {
+    maddr = OBJ_NEW(mca_oob_tcp_addr_t);
+    if ((rc = parse_uri(pop->af_family, pop->net, pop->port, (struct sockaddr_storage*) &(maddr->addr))) != ORTE_SUCCESS) {
```
rhc54: If you get the error case, where do you release the maddr memory? Also, please put the constant first in the != check.

hjelmn:
Is this related to the error coverity is reporting? See CID 710596

JohnWestlund:
@hjelmn That looks like an OPAL defect, this is in ORTE.

rhc54:
@hjelmn It's a similar problem, but we are only addressing it in the OOB - someone else would have to look at the problem in the TCP BTL.

rhc54:
@JohnWestlund once this clears the automated checks, I think it is good to go.

JohnWestlund:
@rhc54 Great!

PR #972, opened 2015-10-04
Add in-finalize indicator, fca fall back to prev barrier if in-finalize

miked-mellanox:


miked-mellanox:
will add PR for v2.x shortly

PR #973, opened 2015-10-04
MXM: avoid disconnecting from a process which is still doing communication

miked-mellanox:
master way to address this:
https://github.com/open-mpi/ompi-release/pull/632


miked-mellanox:
@yosefe , @alex-mikheev - please review

miked-mellanox:
actually, this patch is not needed on master, this code covers it:

https://github.com/open-mpi/ompi/blob/master/ompi/runtime/ompi_mpi_finalize.c#L230

PR #975, opened 2015-10-05
Fix typo that caused compile error

rolfv:
Subject says it all.  Only see this with older versions of CUDA.


rolfv:
One character change. I reviewed it myself.
:+1:

PR #977, opened 2015-10-05
win: free windows in ompi_win_finalize

hjelmn:
This commit frees any outstanding windows at ompi_win_finalize. If
ompi_debug_show_handle_leaks is set a warning message is printed out
indicating that a window is still allocated.

### DIFF
--- a/ompi/win/win.c
+++ b/ompi/win/win.c
@@ -35,6 +35,7 @@
 #include "ompi/mca/osc/base/base.h"
 #include "ompi/mca/osc/osc.h"
 
+#include "ompi/runtime/params.h"
 
 /*
  * Table for Fortran <-> C communicator handle conversion.  Note that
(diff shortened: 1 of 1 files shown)


hjelmn:
@jsquyres Please take a look. This provides similar support to what we already had from MPI_Comm objects. It fixes an issue when using osc/rdma identified by MTT. Was seeing this message in c_create_no_free:

```
[3:32459] Attempt to free memory that is still in use by an ongoing MPI communication (buffer 0x2aab6de3f000, size 135168).  MPI job will now abort.
```

Which is totally meaningless to the user.

jsquyres:
Is there always a f2c index?  Or is that allocated lazily when the Window is used in Fortran?  (I honestly don't remember)

You might want to call it "Fortran Window handle" instead of "f2c" index...?

Is it useful to show the name of the communicator from which the window was created?

hjelmn:
The dump is mostly the same as the one we provide for MPI_Comm. In this case there is always an ftoc index because all windows are stored in the opal_pointer_array. I will update the message to say "Fortran Window handle" instead of ftoc.

As for printing the communicator name. That seems like good additional information. I will add it to the dump.

hjelmn:
The communicator name is not easy to get. The original communicator may have been freed and we do not store a reference to it.

PR #983, opened 2015-10-06
odls/alps: close the directory

hppritcha:
Close the /proc/self/fd dir after checking for open fds.


rhc54:
@hppritcha is this also missing from odls/default?

hppritcha:
no I think it was you who fixed it in default a couple of weeks ago.

PR #990, opened 2015-10-07
configury: do not build usempif08 is PMPI is not built

ggouaillardet:
this is a revamp of commit open-mpi/ompi@2d1a8fa920325b785553abce0b3408010d1f5b6b
Thanks Jeff for the review !

### DIFF
--- a/config/ompi_setup_mpi_fortran.m4
+++ b/config/ompi_setup_mpi_fortran.m4
@@ -367,18 +367,21 @@ AC_DEFUN([OMPI_SETUP_MPI_FORTRAN],[
 
     # We need to have ignore TKR functionality to build the mpi_f08
     # module
-    AS_IF([test $OMPI_TRY_FORTRAN_BINDINGS -ge $OMPI_FORTRAN_USEMPIF08_BINDINGS &&
+    AS_IF([test $OMPI_TRY_FORTRAN_BINDINGS -ge $OMPI_FORTRAN_USEMPIF08_BINDINGS && \
            test $OMPI_FORTRAN_HAVE_IGNORE_TKR -eq 1],
           [OMPI_BUILD_FORTRAN_BINDINGS=$OMPI_FORTRAN_USEMPIF08_BINDINGS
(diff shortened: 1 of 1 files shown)


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

config/ompi_setup_mpi_fortran.m4
```diff
@@ -367,18 +367,21 @@ AC_DEFUN([OMPI_SETUP_MPI_FORTRAN],[
...
-    # F08 bindings require the usempi PMPI profiling bindings
-    AS_IF([test "$WANT_MPI_PROFILING" -eq 0],
-          [OMPI_TRY_FORTRAN_BINDINGS=$OMPI_FORTRAN_USEMPI_BINDINGS
-           OMPI_BUILD_FORTRAN_BINDINGS=$OMPI_FORTRAN_USEMPI_BINDINGS
-           AC_MSG_WARN([PMPI is not built, cannot build usempif08 bindings])])
+    AS_IF([test $OMPI_TRY_FORTRAN_BINDINGS -ge $OMPI_FORTRAN_USEMPIF08_BINDINGS && \
+           test $OMPI_BUILD_FORTRAN_BINDINGS -ge $OMPI_FORTRAN_USEMPIF08_BINDINGS],
+          [ # If we don't have PMPI, we won't build mpi_f08 at all
+           AC_MSG_CHECKING([whether PMPI is enabled])
+           AS_IF([test "$WANT_MPI_PROFILING" -eq 1],
```
jsquyres: Since you're using -eq here, then $WANT_MPI_PROFILING *must* be a number.  So don't use "" around $WANT_MPI_PROFILING (which will promote the value to a string).

ggouaillardet: will do.
that being said, I am a bit surprised...
if test is not a builtin functions, "" should be just a sh thing to handle the case when the variable is not defined.
but anyway, this is inconsistent with the rest of the code so I will fix it

jsquyres:
:+1:

Issue #988, opened 2015-10-06
Need to add volatile to some globals to workaround gcc bug


rolfv:
All the details are contained in this thread from the User's list. Looks like we planned on adding the workaround, but it never happened. I just had a second user run into this issue.
http://www.open-mpi.org/community/lists/users/2015/06/27039.php

I would like to make this a blocker for Open MPI 1.10.1.

jsquyres:
Oy.  Things really crawl out of the walls when we announce we want to release in the Very Near Future!

Can you make a PR for this ASAP (e.g., by tomorrow)?

jsquyres:
BTW, I assume you're going to read the history about this on #625.  Some solutions were proposed on there, IIRC.

rolfv:
Yes, I read all the notes and both you and Dave Goodall already gave a thumbs up to the user of volatile. And the GCC guy also said to make it volatile. That is 3 yes votes! I will work on making it find its way through the 3 releases.

rolfv:
I misread what you were pointing too. I was reading the email thread, not the thread from #625. I was not quite sure what your "softer" suggestion was.

jsquyres:
See https://github.com/open-mpi/ompi/pull/625#issuecomment-110097661 and https://github.com/open-mpi/ompi/pull/625#issuecomment-110225915.

rolfv:
Are suggesting making the whole structure volatile? Because then you are making things volatile that are unnecessary.  Or am I missing something.

jsquyres:
I think we never resolved which of the two approaches was better.  :smile:

jsquyres:
Whichever you do, please submit a PR *today*

rolfv:
Right, I am just trying to implement your solution to see if it works but I am not exactly sure what you mean. I have shown the volatile fixes it with the 4.9.2 compiler. @jsquyres

jsquyres:
I don't know which way is better -- e.g., are there side effects of making the whole struct volatile that could impact performance, etc.

rolfv:
This is clearly not going to make it and I am mostly against these last minute things anyways, so I will remove the blocker from this.  Maybe 10.0.2......

jsquyres:
@rolfv FWIW, we have no idea if there will ever be a 1.10.2.  So I think it would be ok to put in to 1.10.1.

rolfv:
Well, miracles do happen and this has made it into master, 2.0.0, and 10.0.1.  Closing this bug.

PR #650, opened 2015-06-22
Add new PSM2 MTL.

afriedle-intel:
This new MTL supports PSM2 for Omni Path.  PSM2 is a descendant of PSM
with changes to support more processes and some MPI-3 features like mprobe.

PSM2 will only support Omni Path networks; PSM only supports True Scale.
Likewise, the existing PSM MTL will continue to be maintained for True
Scale, while the PSM2 MTL is developed and maintained for Omni Path.

PR #994, opened 2015-10-07
Add volatile to workaround GCC optimization bug in 4.9.2 and greater.

rolfv:
This pull request is for #988 and is a modified version from #650. I added in some comments about the issue that was being fixed as requested by @jsquyres. However, the fix itself is identical to that outlined in #650. I am hoping to get this through into 1.10.1 but if there are still enough questions about this fix, then I will give that up for now and we can try and work through the issues.

I will add @jsquyres , @goodell , @ggouaillardet to review this.


goodell:
:+1:

Issue #805, opened 2015-08-12
Move opal_progress_thread support to v2.x


jsquyres:
This is an issue to ensure that we don't forget to move all the opal_progress_thread stuff over to v2.x.

Note to whomever does this work: there are several commits that will need to be brought over; make sure you get them all!

jsquyres:
This was closed by open-mpi/ompi-release#670.

PR #995, opened 2015-10-07
opal/progress: added function to call a function as part of opal_progress

hjelmn:
This commit add the opal_progress_run_once function. This function
creates a new event that is put on the opal_sync_event_base that runs
the requested function as part of opal_progress().

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -373,3 +373,42 @@ opal_progress_unregister(opal_progress_callback_t cb)
 
     return ret;
 }
+
+struct opal_progress_event_t {
+    opal_event_t super;
+    void *(*fn)(void *);
+    void *arg;
+    opal_event_t *event;
+};
+
+typedef struct opal_progress_event_t opal_progress_event_t;
+
+static void *opal_progress_run_once_cb (int fd, int flags, void *context)
+{
+    opal_progress_event_t *event = (opal_progress_event_t *) context;
+    void *ret;
+
+    ret = event->fn (event->arg);
+    opal_event_del (&event->super);
+    free (event);
+    return ret;
+}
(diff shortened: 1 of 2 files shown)


hjelmn:
@jsquyres @rhc54 @bosilca This will replace the run_in_main stuff in the openib btl. I can move this function down there if there if this doesn't seem like good functionality to have up in opal/runtime.

hjelmn:
If this goes into master it should be rolled up with the rest of the commits associated with #805.

jsquyres:
I'd have to check, but I have a dim recollection that the run_in_main stuff was only because of the lack of thread safety with mpool... if that's correct, then there's no real need for this run_in_main stuff any more.

hjelmn:
I think some of it had to do with the timing in the thread servicing rdmacm events.

hjelmn:
I should add this also helps with the case when opal_using_threads() is 0. If there is an error the error needs to be invoked on the application thread and not the opal progress thread.

jsquyres:
Are you sure?

The thread was just to ensure that RDMACM events could be processed in a timely fashion, but the progress thread couldn't actually *do* anything once the incoming connection request was fully processed (and therefore connected).  So it threw the connection back up to the main thread.

If mpool is made thread safe, the need for "do something in the main thread as a proxy for my progress thread" functionality goes away.  (I think / disclaimer: I haven't looked at the code)

jsquyres:
@hjelmn and I chatted on the phone and figured it all out.

hjelmn:
Sounds like this is not a function worth exposing. Will put it back down into the btl.

bosilca:
:+1: on the last @hjelmn comment. I don't think we want to encourage developers to start using single shot events on the main thread.

PR #996, opened 2015-10-07
btl/openib: remove extra threads

hjelmn:
This commit removes the service and async event threads from the
openib btl. Both threads are replaced by opal progress thread
support. The run_in_main function is now supported by allocating an
event and adding it to the sync event base. This ensures that the
requested function is called as part of opal_progress.


hjelmn:
@miked-mellanox If mellanox cares about the openib btl please take a look. Otherwise I will commit this in the next day or so.

miked-mellanox:
looks good and very elegant w/ opal_progress thread.

:+1:

PR #999, opened 2015-10-08
coll-portals4: add gather and igather implementations that use Portals4 triggered operations

tkordenbrock:
This commit adds implementations of gather and igather using
Portals4 triggered operations.  The default algorithm is linear,
but binomial can be selected using an MCA parameter -
coll_portals4_use_binomial_gather_algorithm.


regrant:
:+1: Merged

Issue #965, opened 2015-10-02
darray type incorrectly created or interpreted in mpi_alltoallw


dmt4:
Instead of the respective pieces corresponding to the specific darray type, all processes receive the same darray. I've tried to specialise a general routine to a small example working with 2 processes only. Here is a test code: 
https://gist.github.com/anonymous/60de0ae4f52f15b89dae

bosilca:
This test successfully pass with open-mpi/ompi@1cac060 using the basic collective module. Can you check to see what collective function is used in your run.

However, from a datatype perspective the 2 datatypes created by the darray calls look correct.

Datatype 0x7fad99d3faf0[Dup ] id 73 size 4 align 4 opal_id 0 length 4 used 1
true_lb 0 true_ub 4 (true_extent 4) lb 0 ub 8 (extent 8)
nbElems 1 loops 0 flags 114 (committed contiguous )-cC----GD--[---][---]
   contain OPAL_FLOAT4
--C---P-D--[---][---]    OPAL_FLOAT4 count 1 disp 0x0 (0) extent 4 (size 4)
-------G---[---][---]  OPAL_END_LOOP prev 1 elements first elem displacement 0 size of data 4
Optimized description
-cC---P-DB-[---][---]    OPAL_FLOAT4 count 1 disp 0x0 (0) extent 4 (size 4)
-------G---[---][---]  OPAL_END_LOOP prev 1 elements first elem displacement 0 size of data 4

Datatype 0x7fad99d3fd60[Dup ] id 73 size 4 align 4 opal_id 0 length 4 used 1
true_lb 4 true_ub 8 (true_extent 4) lb 4 ub 12 (extent 8)
nbElems 1 loops 0 flags 114 (committed contiguous )-cC----GD--[---][---]
   contain OPAL_FLOAT4
--C---P-D--[---][---]    OPAL_FLOAT4 count 1 disp 0x0 (0) extent 4 (size 4)
-------G---[---][---]  OPAL_END_LOOP prev 1 elements first elem displacement 0 size of data 4
Optimized description
-cC---P-DB-[---][---]    OPAL_FLOAT4 count 1 disp 0x0 (0) extent 4 (size 4)
-------G---[---][---]  OPAL_END_LOOP prev 1 elements first elem displacement 0 size of data 4

dmt4:
[...]

PR #1003, opened 2015-10-09
sync subarray and darray behaviour

gpapaure:
Basic usage works again for both subarray and darray by using create_hindexed instead of create_resized
Test case cited in http://lists.mpich.org/pipermail/discuss/2015-January/003608.html still works

### DIFF
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -13,6 +13,7 @@
  * Copyright (c) 2009      Sun Microsystems, Inc. All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2015      Bull SAS.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


jsquyres:
@bosilca can you comment?  Thanks.

ggouaillardet:
is this related to #965 ?

gpapaure:
I haven't seen this issue before posting the pull request, thanks.
Yes i think it is related.
As you said in your last comment create_resized doesn't produce the right datatype layout.
The solution in this patch is to use create_hindexed with only one element and the same offset than previous create_resized.
I've tested darray-alltoallw-test and darray-alltoallw-test2 and both of them works now.
The fix may apply on release branch.

ggouaillardet:
Thanks Guillaume,

bottom line, is MPI_Type_create_resized broken or not ?

in my test case, both mpich and ompi have the same behavior (well, ompi also issues a warning), and this is not the behavior I expected (could be a bug, or my incorrect expectation)

jsquyres:
Add @bosilca

hjelmn:
We might want to make this issue a blocker on 2.0.0. I remember fixing an issue with subarray but didn't realize darray was also broken.

ggouaillardet:
IMHO #965 is the blocker and the milestone should be v1.10.1

bosilca:
Due to the emergency of the next release, after talking with Jeff we propose to get the patch proposed here in the 1.10, and create an issue for the trunk. This will give me a little bit more time to dig into the internals of the type_resize and understand what is not going as expected there.

bosilca:
The corresponding issue for the devel version is #1063.

jsquyres:
@gpapaure Is this PR obsolete due to #1063?

jsquyres:
I assume this issue is now closed (due to #1063 and #1072).  Closing the ticket.

PR #1007, opened 2015-10-11
init/finalize: extensions

jsquyres:
Proposed extensions for Open MPI:

* If MPI_INITLIZED is invoked and MPI is only partially initialized, wait until MPI is fully initialized before returning.
* If MPI_FINALIZED is invoked and MPI is only partially finalized, wait until MPI is fully finalized before returning.
* If the ompi_mpix_allow_multi_init MCA param is true, allow MPI_INIT and MPI_INIT_THREAD to be invoked multiple times without error (MPI will be safely initialized only the first time it is invoked).

@hppritcha @bosilca This is a bit more than we talked about on the phone last week.  I ended up using a mutex instead of atomics because I have to check multiple values, so it made more sense to put all the checks within a mutex lock.

### DIFF
--- a/ompi/mpi/c/finalized.c
+++ b/ompi/mpi/c/finalized.c
@@ -11,6 +11,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -41,6 +42,14 @@ int MPI_Finalized(int *flag)
 
     OPAL_CR_NOOP_PROGRESS();
 
+    /* We must obtain the lock to guarnatee consistent values of
+       ompi_mpi_initialized and ompi_mpi_finalized.  Note, too, that
+       this lock is held for the bulk of the duration of
+       ompi_mpi_init() and ompi_mpi_finalize(), so when we get the
+       lock, we are guaranteed that some other thread is not part way
+       through initialization or finalization. */
+    opal_mutex_lock(&ompi_mpi_bootstrap_mutex);
+
     if (MPI_PARAM_CHECK) {
         if (NULL == flag) {
 
@@ -49,17 +58,19 @@ int MPI_Finalized(int *flag)
                MPI_Finalize) or not */
 
             if (ompi_mpi_initialized && !ompi_mpi_finalized) {
+                opal_mutex_unlock(&ompi_mpi_bootstrap_mutex);
                 return OMPI_ERRHANDLER_INVOKE(MPI_COMM_WORLD, MPI_ERR_ARG,
                                               FUNC_NAME);
             } else {
+                opal_mutex_unlock(&ompi_mpi_bootstrap_mutex);
                 return OMPI_ERRHANDLER_INVOKE(null, MPI_ERR_ARG,
                                               FUNC_NAME);
             }
         }
     }
 
-    /* Pretty simple */
-
     *flag = ompi_mpi_finalized;
+    opal_mutex_unlock(&ompi_mpi_bootstrap_mutex);
+
     return MPI_SUCCESS;
 }
--- a/ompi/mpi/c/initialized.c
+++ b/ompi/mpi/c/initialized.c
@@ -11,6 +11,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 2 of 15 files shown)


ompi/mpi/c/finalized.c
```diff
@@ -48,17 +51,28 @@ int MPI_Finalized(int *flag)
...
 
-    /* Pretty simple */
+    /* If some process has *started* FINALIZE but not yet completed
+       it, then be a little social and wait here until it is
+       completed. */
+    while (ompi_mpi_finalize_started && !ompi_mpi_finalized) {
+        opal_mutex_unlock(&ompi_mpi_bootstrap_mutex);
+        usleep(1);
+        opal_mutex_lock(&ompi_mpi_bootstrap_mutex);
+    }
```
bosilca: The result of this loop is basically a continuous loop doing a "lock + unlock"every 1us. As you are using a mutex, why don't you rely on the trigger of the mutex by the thread that succeeded to ensure the sequentiality of this code?

ompi/mpi/c/initialized.c
```diff
@@ -48,17 +51,27 @@ int MPI_Initialized(int *flag)
...
     }
 
-    /* Pretty simple */
+    /* If some process has *started* INIT but not yet completed it,
+       then be a little social and wait here until it is completed. */
+    while (ompi_mpi_init_started && !ompi_mpi_initialized) {
+        opal_mutex_unlock(&ompi_mpi_bootstrap_mutex);
+        usleep(1);
+        opal_mutex_lock(&ompi_mpi_bootstrap_mutex);
+    }
```
bosilca: Same comment as above.

ompi/mpi/c/finalized.c
jsquyres: I specifically didn't hold the lock for the duration of ompi_mpi_init() / ompi_mpi_finalize().  Hence, I can't have the thread here just block waiting for the lock (e.g., which would naturally wait until the thread in ompi_mpi_init() completed the function).

Instead, I'm only using the lock to access the variables.  Hence, I have to loop around checking it, unlocking, delaying, and locking again.

Are you advocating that I should hold the lock through ompi_mpi_init() / ompi_mpi_finalize()?  I could probably do that (and then get rid of this loop, and the similar one in INITIALIZED)

jsquyres:
@bosilca Code changed to hold the lock for the bulk of the duration of ompi_mpi_init() and ompi_mpi_finalize(), thereby removing the need to loop checking the values of variables.

bosilca:
Thanks Jeff. Also, adding the blurb of text to the man pages is a great idea. The code looks good :+1:

jsquyres:
@hppritcha Does this look good to you?

jsquyres:
The more I think about this, the less I think it makes sense to allow MPI_INIT[_THREAD] to be invoked multiple times safely, but only allow MPI_FINALIZE to be invoked once.

Should we give up on the idea of invoking MPI_INIT[_THREAD] multiple times, and just have this PR essentially be the bootstrap_mutex stuff?

ompi/runtime/ompi_mpi_init.c
```diff
@@ -124,11 +124,12 @@ const char ompi_version_string[] = OMPI_IDENT_STRING;
  * Global variables and symbols for the MPI layer
  */
 
-bool ompi_mpi_init_started = false;
-bool ompi_mpi_initialized = false;
-bool ompi_mpi_finalized = false;
-bool ompi_rte_initialized = false;
-int32_t ompi_mpi_finalize_started = false;
+opal_mutex_t ompi_mpi_bootstrap_mutex;
```
jsquyres: @hjelmn @bosilca What is the Right way to initialize this mutex?  I notice that it's a class -- not a plain pthread mutex.  Should I use a simple pthread mutex instead, and then statically initialize it with PTHREAD_MUTEX_INITIALIZER?

jsquyres: Looks like the Right way is #1026.

hppritcha:
Looks excellent.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

jsquyres:
I removed the MPIX MCA param and the ability to call MPI_INIT[_THREAD] multiple times.  It's a can of worms.  If someone else wants to tackle that issue, feel free to do so.

This PR is now just:

* If MPI_INITLIZED is invoked and MPI is only partially initialized, wait until MPI is fully initialized before returning.
* If MPI_FINALIZED is invoked and MPI is only partially finalized, wait until MPI is fully finalized before returning.

PR #1014, opened 2015-10-13
configury: check libnl version and warn in case of conflict

ggouaillardet:
libnl and libnl-3 are known to conflict with each other, so warn
if these two libs are used by different third party libs

### DIFF
--- a/config/opal_check_libnl.m4
+++ b/config/opal_check_libnl.m4
@@ -1,0 +1,280 @@
+dnl -*- shell-script -*-
+dnl
+dnl Copyright (c) 2015-2016 Research Organization for Information Science
+dnl                         and Technology (RIST). All rights reserved.
+dnl $COPYRIGHT$
+dnl
+dnl Additional copyrights may follow
+dnl
+dnl $HEADER$
+dnl
+
+dnl
+dnl More libnl v1/v3 sadness: the two versions are not compatible
+dnl and will not work correctly if simultaneously linked into the
+dnl same applications.  Unfortunately, they *will* link into the
+dnl same image!  On platforms like CentOS 7, libibverbs depends on
+dnl libnl-3.so.200 and friends, so if libnl3-devel packages are not
+dnl installed, but libnl-devel are, Open MPI should not try to use libnl.
+dnl
+dnl GROSS: libnl wants us to either use pkg-config (which we
+dnl can't assume is always present) or we need to look in a
+dnl particular directory for the right libnl3 include files.  For
+dnl now, just hard code the special path into this logic.
+dnl
+dnl _OPAL_CHECK_PACKAGE_LIB() invokes OPAL_LIBNL_SANITY_CHECK() in order
+dnl to keep track of which libraries depend on libnl and which libraries
+dnl depend on libnl3.
+dnl Open MPI will not be able to build a component vs a given version of libnl
+dnl if the other libnl version is required by some third party components.
+dnl At the end of configure, display a summary of who is using what, and aborts
+dnl if both libnl versions are required.
+
+dnl OPAL_LIBNL_SANITY_INIT()
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_LIBNL_SANITY_INIT], [
+    opal_libnl_version=0
+    opal_libnlv1_libs=
+    opal_libnlv3_libs=
+    AC_ARG_WITH([libnl],
+                [AC_HELP_STRING([--with-libnl(=DIR)],
+                                [Directory prefix for libnl (typically only necessary if libnl is installed in a location that the compiler/linker will not search by default)])])
+
+    # The --with options carry two pieces of information: 1) do
+    # you want a specific version of libnl, and 2) where that
+    # version of libnl lives.  For simplicity, let's separate
+    # those two pieces of information.
+    case "$with_libnl" in
+        no)
+            # Nope, don't want it
+            opal_want_libnl=no
+            ;;
+        "")
+            # Just try to build with libnl
+            opal_want_libnl=try
+            opal_libnl_location=
+            ;;
+        yes)
+            # Yes, definitely want it
+            opal_want_libnl=yes
+            opal_libnl_location=
+            ;;
+        *)
+            # Yes, definitely want it -- at a specific location
+            opal_want_libnl=yes
+            opal_libnl_location=$with_libnl
+            ;;
+    esac
+])
+
+dnl OPAL_LIBNL_SANITY_CHECK(lib, function, LIBS)
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_LIBNL_SANITY_CHECK], [
+    case $host in
+        *linux*)
+            OPAL_VAR_SCOPE_PUSH([ldd_output libnl_version])
+            AC_LANG_PUSH(C)
+            cat > conftest_c.$ac_ext << EOF
+extern void $2 (void);
+int main(int argc, char *argv[[]]) {
+    $2 ();
+    return 0;
+}
+EOF
+            OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1 $LIBS $3],
+                             [ldd_output=`ldd conftest`
+                              libnl_version=0
+                              AS_IF([echo $ldd_output | grep -q libnl.so],
+                                    [AS_IF([test $opal_libnl_version -eq 3],
+                                           [AC_MSG_WARN([lib nl version conflict: $opal_libnlv3_libs requires libnl-3 whereas $1 requires libnl])],
+                                           [opal_libnlv1_libs="$opal_libnlv1_libs $1"
+                                            OPAL_UNIQ([opal_libnlv1_libs])
+                                            opal_libnl_version=1])
+                                     libnl_version=1])
+                              AS_IF([echo $ldd_output | grep -q libnl-3.so],
+                                    [AS_IF([test $libnl_version -eq 1],
+                                           [AC_MSG_WARN([lib $1 requires both libnl v1 and libnl v3 -- yoinks!])
+                                            AC_MSG_WARN([This is a configuration that is known to cause run-time crashes])
+                                            AC_MSG_ERROR([Cannot continue])])
+                                     AS_IF([test $opal_libnl_version -eq 1],
+                                           [AC_MSG_WARN([lib nl version conflict: $opal_libnlv1_libs requires libnl whereas $1 requires libnl-3])],
+                                           [opal_libnlv3_libs="$opal_libnlv3_libs $1"
+                                            OPAL_UNIQ([opal_libnlv3_libs])
+                                            opal_libnl_version=3])])
+                              rm -f conftest conftest_c.$ac_ext],
+                             [AC_MSG_WARN([Could not link a simple program with lib $1])])
+            AC_LANG_POP(C)
+            OPAL_VAR_SCOPE_POP([ldd_output libnl_version])
+            ;;
+        esac
+])
+
+dnl
+dnl Check for libnl-3.
+dnl
+dnl Inputs:
+dnl
+dnl $1: prefix where to look for libnl-3
+dnl $2: var name prefix of _CPPFLAGS and _LDFLAGS and _LIBS
+dnl
+dnl Outputs:
+dnl
+dnl - Set $2_CPPFLAGS necessary to compile with libnl-3
+dnl - Set $2_LDFLAGS necessary to link with libnl-3
+dnl - Set $2_LIBS necessary to link with libnl-3
+dnl - Set OPAL_HAVE_LIBNL3 1 if libnl-3 will be used
+dnl
+AC_DEFUN([OPAL_CHECK_LIBNL_V3],[
+    OPAL_VAR_SCOPE_PUSH([CPPFLAGS_save opal_tmp_CPPFLAGS LIBS_save LDFLAGS_save])
+    AC_MSG_NOTICE([checking for libnl v3])
+
+    AS_IF([test "$opal_want_libnl" != "no"],
+          [AS_IF([test -z "$opal_libnl_location"],
+                 [AC_MSG_CHECKING([for /usr/include/libnl3])
+                  AS_IF([test -d "/usr/include/libnl3"],
+                        [opal_tmp_CPPFLAGS=-I/usr/include/libnl3
+                         opal_libnlv3_happy=1
+                         AC_MSG_RESULT([found])],
+                        [AC_MSG_RESULT([not found])
+                         AC_MSG_CHECKING([for /usr/local/include/libnl3])
+                         AS_IF([test -d "/usr/local/include/libnl3"],
+                               [opal_tmp_CPPFLAGS=-I/usr/local/include/netlink3
+                                opal_libnlv3_happy=1
+                                AC_MSG_RESULT([found])],
+                               [opal_libnlv3_happy=0
+                                AC_MSG_RESULT([not found])])])],
+                 [AC_MSG_CHECKING([for $1/include/libnl3])
+                  AS_IF([test -d "$1/include/libnl3"],
+                        [opal_tmp_CPPFLAGS="-I$1/include/libnl3"
+                         opal_libnlv3_happy=1
+                         AC_MSG_RESULT([found])],
+                        [opal_libnlv3_happy=0
+                         AC_MSG_RESULT([not found])])])
+           CPPFLAGS_save=$CPPFLAGS
+           CPPFLAGS="$opal_tmp_CPPFLAGS $CPPFLAGS"
+
+           # Random note: netlink/version.h is only in libnl v3 - it is not in libnl v1.
+           # Also, nl_recvmsgs_report is only in libnl v3.
+           AS_IF([test $opal_libnlv3_happy -eq 1],
+                 [OPAL_CHECK_PACKAGE([$2],
+                                     [netlink/version.h],
+                                     [nl-3],
+                                     [nl_recvmsgs_report],
+                                     [],
+                                     [$1],
+                                     [],
+                                     [],
+                                     [opal_libnlv3_happy=0])
+
+                  # Note that OPAL_CHECK_PACKAGE is going to add
+                  # -I$dir/include into $2_CPPFLAGS.  But because libnl v3
+                  # puts the headers in $dir/include/libnl3, we need to
+                  # overwrite $2_CPPFLAGS with -I$dir/include/libnl3.  We can do
+                  # this unconditionally; we don't have to check for
+                  # success (checking for success occurs below).
+                  $2_CPPFLAGS=$opal_tmp_CPPFLAGS])
+
+           # If we found libnl-3, we *also* need libnl-route-3
+           LIBS_save=$LIBS
+           LDFLAGS_save=$LDFLAGS
+           AS_IF([test -n "$$2_LDFLAGS"],
+                 [LDFLAGS="$$2_LDFLAGS $LDFLAGS"])
+           AS_IF([test $opal_libnlv3_happy -eq 1],
+                 [AC_SEARCH_LIBS([nl_rtgen_request],
+                                 [nl-route-3],
+                                 [],
+                                 [opal_libnlv3_happy=0])])
+           LIBS=$LIBS_save
+           LDFLAGS=$LDFLAGS_save
+
+           # Just because libnl* is evil, double check that the
+           # netlink/version.h we found was for libnl v3.  As far as we
+           # know, netlink/version.h only first appeared in version
+           # 3... but let's really be sure.
+           AS_IF([test $opal_libnlv3_happy -eq 1],
+                 [AC_MSG_CHECKING([to ensure these really are libnl v3 headers])
+                  AS_IF([test -n "$$2_CPPFLAGS"],
+                        [CPPFLAGS="$$2_CPPFLAGS $CPPFLAGS"])
+                  AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#include <netlink/netlink.h>
+#include <netlink/version.h>
+#ifndef LIBNL_VER_MAJ
+#error "LIBNL_VER_MAJ not defined!"
+#endif
+/* to the best of our knowledge, version.h only exists in libnl v3 */
+#if LIBNL_VER_MAJ != 3
+#error "LIBNL_VER_MAJ != 3, I am sad"
+#endif
+                                     ]])],
+                                    [AC_MSG_RESULT([yes])],
+                                    [AC_MSG_RESULT([no])
+                                     opal_libnlv3_happy=0])])
+
+           CPPFLAGS=$CPPFLAGS_save],
+
+          [opal_libnlv3_happy=0])
+
+    # If we found everything
+    AS_IF([test $opal_libnlv3_happy -eq 1],
+          [$2_LIBS="-lnl-3 -lnl-route-3"
+           OPAL_HAVE_LIBNL3=1])
+
+   OPAL_VAR_SCOPE_POP
+])
+
+dnl
+dnl Check for libnl.
+dnl
+dnl Inputs:
+dnl
+dnl $1: prefix where to look for libnl
+dnl $2: var name prefix of _CPPFLAGS and _LDFLAGS and _LIBS
+dnl
+dnl Outputs:
+dnl
+dnl - Set $2_CPPFLAGS necessary to compile with libnl
+dnl - Set $2_LDFLAGS necessary to link with libnl
+dnl - Set $2_LIBS necessary to link with libnl
+dnl - Set OPAL_HAVE_LIBNL3 0 if libnl will be used
+dnl
+AC_DEFUN([OPAL_CHECK_LIBNL_V1],[
+    AC_MSG_NOTICE([checking for libnl v1])
+
+    AS_IF([test "$opal_want_libnl" != "no"],
+          [OPAL_CHECK_PACKAGE([$2],
+                              [netlink/netlink.h],
+                              [nl],
+                              [nl_connect],
+                              [-lm],
+                              [$1],
+                              [],
+                              [opal_libnlv1_happy=1],
+                              [opal_libnlv1_happy=0])],
+          [opal_libnlv1_happy=0])
+
+    AS_IF([test $opal_libnlv1_happy -eq 1],
+          [$2_LIBS="-lnl -lm"
+           OPAL_HAVE_LIBNL3=0])
+])
+
+dnl
+dnl Summarize libnl and libnl3 usage,
+dnl and abort if conflict is found
+dnl
+dnl Print the list of libraries that use libnl, 
+dnl the list of libraries that use libnl3,
+dnl and aborts if both libnl and libnl3 are used.
+dnl
+AC_DEFUN([OPAL_CHECK_LIBNL_SUMMARY],[
+    AC_MSG_CHECKING([for libraries that use libnl v1])
+    AS_IF([test -n "$opal_libnlv1_libs"],
+          [AC_MSG_RESULT([$opal_libnlv1_libs])],
+          [AC_MSG_RESULT([(none)])])
+    AC_MSG_CHECKING([for libraries that use libnl v3])
+    AS_IF([test -n "$opal_libnlv3_libs"],
+          [AC_MSG_RESULT([$opal_libnlv3_libs])],
+          [AC_MSG_RESULT([(none)])])
+    AS_IF([test -n "$opal_libnlv1_libs" && test -n "$opal_libnlv3_libs"],
+          [AC_MSG_WARN([libnl v1 and libnl v3 have been found as dependent libraries])
+           AC_ERROR([This is a configuration that is known to cause run-time crashes])])
+])
--- a/config/opal_check_package.m4
+++ b/config/opal_check_package.m4
@@ -138,6 +138,10 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_LIB], [
                   test "$ac_cv_search_$3" != "none required"],
                  [$1_LIBS="$ac_cv_search_$3 $4"],
                  [$1_LIBS="$4"])
+            # libnl v1 and libnl3 are known *not* to coexist
+            # for each library, figure out whether it depends on libnl or libnl3 or none
+            # so conflicts can be reported and/or prevented
+            OPAL_LIBNL_SANITY_CHECK([$2], [$3], [$$1_LIBS])
            $7],
           [$8])
 
--- a/configure.ac
+++ b/configure.ac
@@ -1449,6 +1449,8 @@ m4_ifdef([project_orte], [ORTE_CONFIG_FILES])
 m4_ifdef([project_ompi], [OMPI_CONFIG_FILES])
 m4_ifdef([project_oshmem], [OSHMEM_CONFIG_FILES])
 
+OPAL_CHECK_LIBNL_SUMMARY
+
 AC_OUTPUT
 
 OPAL_SUMMARY_PRINT
--- a/opal/mca/reachable/netlink/configure.m4
+++ b/opal/mca/reachable/netlink/configure.m4
@@ -1,6 +1,8 @@
 # -*- shell-script -*-
 #
-# Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015-2016 Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -40,7 +42,6 @@ dnl LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 dnl ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 dnl POSSIBILITY OF SUCH DAMAGE.
 
-
 dnl Check for libnl; prefer version 3 instead of version 1.  Abort (i.e.,
 dnl AC_MSG_ERROR) if neither libnl v1 or v3 can be found.
 dnl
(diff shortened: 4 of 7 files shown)


config/opal_check_libnl_version.m4
```diff
@@ -0,0 +1,50 @@
...
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_CHECK_LIBNL_VERSION], [
+     AC_LANG_PUSH(C)
+     rm -f conftest_c.$ac_ext
+      cat > conftest_c.$ac_ext << EOF
+int main(int argc, char *argv[]) {
+    return 0;
+}
+EOF
+     OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1],
```
jsquyres: It would be good to avoid OPAL_LOG_COMMAND with explicitly compiling a source and then running ldd on it (e.g., ldd doesn't exist on OS X.  I know libnl isn't on OS X, but we should probably still avoid OS-specific constructs in configure when possible).

It might be better to AC_LINK_IFELSE with symbols that are known to be in libnl vs. libnl3 to determine which libnl you have.  That would seem to be more portable and resilient if the output of ldd varies, or the exact names of the libraries varies between different Linux distros.

config/opal_check_package.m4
```diff
@@ -134,6 +134,8 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_LIB], [
                   test "$ac_cv_search_$3" != "none required"],
                  [$1_LIBS="$ac_cv_search_$3 $4"],
                  [$1_LIBS="$4"])
+            # Check libnl version if any
```
jsquyres: I think you need a better / more explanitive comment here: this is a somewhat subtle issue.  Future maintainers might (will) look at this and ask "why do we have a check for libnl in the generic package library check?"

It might also be worthwhile to rename this macro -- perhaps OPAL_LIBNL_SANITY_CHECK, or somesuch?

opal/mca/reachable/netlink/configure.m4
```diff
@@ -74,15 +76,17 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
...
-       $1="-I/usr/include/libnl3"
-       CPPFLAGS="$$1 $CPPFLAGS"
-       AC_MSG_CHECKING([for /usr/include/libnl3])
-       AS_IF([test -d "/usr/include/libnl3"],
-             [AC_MSG_RESULT([present])
-              AC_CHECK_HEADER(
-                [netlink/version.h],
-                [AC_COMPILE_IFELSE(
-                    [AC_LANG_PROGRAM([[
+       AS_IF([test $opal_libnl_version -eq 1],
```
jsquyres: I think that much of the body of this macro can move to a new macro -- such as OPAL_LIBNL_CHECK_VERSION (i.e., these are the kinds of tests we want to use to figure out what version of libnl is available) -- and then both this macro and OPAL_CHECK_LIBNL -- simply calls OPAL_LIBNL_CHECK_VERSION to figure out what version of libnl is found.

ggouaillardet:
ok, will do

config/opal_check_libnl_version.m4
ggouaillardet: got that, thanks !
is there any risk we might hit the similar issue melannox folks reported ?
(libmpi had to be explicitly linked with libopen-pal e.g. linking with libopen-rte only was not enough,
and btw/fwiw, I have never been able to reproduce this)
also, that is a bit slower since linking must be performed twice to reach a conclusion
(e.g. failure to link with libnl specific symbol, and then with libnl3 specific symbol is required to conclude libXYZ do nor require libnl nor libnl3)

config/opal_check_package.m4
```diff
@@ -134,6 +134,8 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_LIB], [
...
+            OPAL_CHECK_LIBNL_VERSION([$2])
```
jsquyres: Don't you also need to pass $4 here, too?

jsquyres: ...and also $3?  I.e., if you don't link in a symbol from the library in the LIBNL test, then it might not actually pull in the $2 library...

config/opal_check_libnl_version.m4
jsquyres: It's possible, but I don't think it's likely -- this is a slightly different situation: we're linking in libraries which have already successfully been linked.  If it happens, though, we can adjust.

As for twice as long for this individual test, yes, this is true, but I'm not worry about it.  configure already takes a long time on a fast machine.  Also, 2 links instead of 1 is still a very short amount of time.

ggouaillardet:
@jsquyres i am stuck right now :-(

i was naively hoping the following program would link successfully
```c
int (*ibv_open_device)(void);
int nl_geterror(void);

int main (int argc, char *argv[]) {
    ibv_open_device();
    nl_geterror();
    return 0;
}
```
with
```
gcc conftest.c -libverbs
```

unfortunatly, i get the following error
```
/usr/bin/ld: /tmp/cccPwM0X.o: undefined reference to symbol 'nl_geterror'
/usr/bin/ld: note: 'nl_geterror' is defined in DSO /lib64/libnl.so.1 so try adding it to the linker command line
/lib64/libnl.so.1: could not read symbols: Invalid operation
```
note if i explicitly add ```-lnl```, it links successfully

bottom line, i am back to the ldd option, which is known not to work at least on OSX

any thoughts ?

jsquyres:
@ggouaillardet Oy.  Hmm.  No immediate thoughts... but I'll think about this today.  Perhaps I'll get a good idea by the time you get to work tomorrow...

jsquyres:
According to http://www.infradead.org/~tgr/libnl/, libnl is Linux-specific.  So we should probably disable all this checking on non-Linux systems.

That being said, if we can't rely on libraries' implicit dependencies to resolve libnl symbols, then I, too, can't think of a better option than ldd.  :frowning:  I'm worry about the fragility of output from ldd (e.g., if it changes over time/versions).  Make your check be robust enough to handle if it can't understand the output from ldd, I guess.

opal/mca/reachable/netlink/configure.m4
ggouaillardet: i am not sure about what you mean.
this macro is used to find out what we can build (e.g. lib plus devel headers are required)
whereas other macro is used to find out what third party libs are linked against (no devel header required)
all i can think about is move and rename the macro (and maybe cache the result) to config/opal_check_libnl.m4

ggouaillardet:
i returned to ldd
ldd test is pretty bulletproof ... it is a dumb grep with libnl.so or libnl-3.so.
if no match, then assume no libnl whatsoever is used.
this test does not assume any ldd output format (such as <lib name> => <full path to lib>)
if that looks good to you, i will squeeze this PR and commit it

jsquyres:
@ggouaillardet I'm trying to replicate the constructor segv you saw in #830.  Can you tell me what versions of libnl and libnl3 you have installed?

ggouaillardet:
I will check that on Monday
This is a centos 7 box, and the crash occurs when running ompi_info --all

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,58 @@
...
+dnl
+dnl $HEADER$
+dnl
+
+dnl OPAL_LIBNL_SANITY_INIT()
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_LIBNL_SANITY_INIT], [
+    opal_libnl_version=0
+    opal_libnl_libs=""
+    opal_libnl3_libs=""
```
jsquyres: Minor nit: the "" isn't necessary here.  Just `foo=`.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,58 @@
...
+])
+
+dnl OPAL_LIBNL_SANITY_CHECK(lib, function, LIBS)
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_LIBNL_SANITY_CHECK], [
+    case $host in
+        *linux*)
+            AC_LANG_PUSH(C)
+            libnl_LIBS_save="$LIBS"
```
jsquyres: No need for "" here -- just `libnl_LIBS_save=$LIBS`.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,58 @@
...
+                                  if test $opal_libnl_version -eq 1; then
+                                      AC_MSG_WARN([lib nl version conflict: $opal_libnl_libs requires libnl whereas $1 requires libnl-3])
+                                  else
+                                      opal_libnl3_libs="$opal_libnl3_libs $1"
+                                      opal_libnl_version=3
+                                  fi
+                              fi
+                              rm -f conftest conftest_c.$ac_ext],
+                             [AC_MSG_WARN([Could not link a simple program with lib $1])])
+            LIBS="$libnl_LIBS_save"
```
jsquyres: Ditto here -- no "" needed.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,58 @@
...
+            OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1],
+                             [ldd_output=`ldd conftest`
+                              if echo $ldd_output | grep -q libnl.so; then
+                                  if test $opal_libnl_version -eq 3; then
+                                      AC_MSG_WARN([lib nl version conflict: $opal_libnl3_libs requires libnl-3 whereas $1 requires libnl])
+                                  else
+                                      opal_libnl_libs="$opal_libnl_libs $1"
+                                      opal_libnl_version=1
+                                  fi
+                              elif echo $ldd_output | grep -q libnl-3.so; then
```
jsquyres: Shouldn't we test for *both* libnl and libnl3?  (for the case where a dependent library links against *both* libnl and libnl3)  It would be a very poor dependent library that did this, but we should catch that case, too, right?

opal/mca/reachable/netlink/configure.m4
jsquyres: You are right.  Disregard my above comment.

opal/mca/reachable/netlink/configure.m4
```diff
@@ -74,15 +76,17 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
...
+             [HAVE_LIBNL3=0],
+             [$1="-I/usr/include/libnl3"
```
jsquyres: Is this macro enough?

I'm wondering if we need a more complete check for libnl-vs-libnl3.  E.g., something that calls OPAL_CHECK_PACKAGE (which does very complete checks for dependent libraries).

For example, the libfabric usnic provider uses the OPAL_CHECK_PACKAGE-inspired FI_CHECK_PACKAGE to do a very complete set of libnl-vs-libnl3 checks: see https://github.com/ofiwg/libfabric/blob/master/prov/usnic/configure.m4#L143-L312

It also does some gratuitous checks, like checking to ensure that the netlink.h we found matches the library we found (not because we found a real world case where this happened, but rather because we know that libnl is evil and this is not beyond their scope of evilness...).

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

jsquyres:
@ggouaillardet Where are we on this PR? I see you pushed new commits -- should I re-review?

rhc54:
@ggouaillardet Poke! Any thoughts on moving this forward? I saw it recently resurfaced again.

ggouaillardet:
@jsquyres i rebased my commits, could you please review them again ?

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,223 @@
...
+                              [netlink/version.h],
+                              [nl-3],
+                              [nl_recvmsgs_report],
+                              [],
+                              [$1],
+                              [],
+                              [opal_libnl3_happy=1],
+                              [opal_libnl3_happy=0])
+
+           # Note that FI_CHECK_PACKAGE is going to add
```
jsquyres: Minor copy-n-paste error here: should be OPAL_CHECK_PACKAGE, not FI_CHECK_PACKAGE.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,223 @@
...
+#endif
+                              ]])],
+                             [AC_MSG_RESULT([yes])],
+                             [AC_MSG_RESULT([no])
+                              opal_libnl3_happy=0])])
+
+    CPPFLAGS=$CPPFLAGS_save
+
+    # If we found everything
+   AS_IF([test $opal_libnl3_happy -eq 1],
```
jsquyres: Trivial: indenting oddity here.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,223 @@
...
+    LIBS=$LIBS_save
+    LDFLAGS=$LDFLAGS_save
+
+    # Just because libnl* is evil, double check that the
+    # netlink/version.h we found was for libnl3.  As far as we
+    # know, netlink/version.h only first appeared in version
+    # 3... but let's really be sure.
+    AS_IF([test $opal_libnl3_happy -eq 1],
+          [AC_MSG_CHECKING([to ensure these really are libnl3 headers])
+           CPPFLAGS="$$2_CPPFLAGS $CPPFLAGS"
```
jsquyres: Minor: here, you unconditionally assign `CPPFLAGS`, but above (lines 150-151), you conditionally assign `LDFLAGS` with the same type of contents.  I don't really care which way you go, but these assignment styles should likely be consistent for ease of future maintenance, IMHO.

configure.ac
```diff
@@ -1390,3 +1390,6 @@ m4_ifdef([project_ompi], [OMPI_CONFIG_FILES])
 m4_ifdef([project_oshmem], [OSHMEM_CONFIG_FILES])
 
 AC_OUTPUT
+
+OPAL_LOG_MSG([libnl : $opal_libnl_libs])
+OPAL_LOG_MSG([libnl3: $opal_libnl3_libs])
```
jsquyres: Per the discussion on devel the other day about emitting a report at the end of configure (http://www.open-mpi.org/community/lists/devel/2016/01/18497.php), you might not want to add this to the *very* end of configure -- perhaps a few lines earlier, with the output of the final wrapper compiler flags...?

jsquyres: I don't think my comment was clear; let me try again...

Perhaps in the section right before the AC_OUTPUT, where we show the various wrapper compiler flags, you can have something like this:
```sh
AC_MSG_CHECKING([for libraries that use libnl])
AC_MSG_RESULT([$opal_libnl_libs])
AC_MSG_CHECKING([for libraries that use libnl3])
AC_MSG_RESULT([$opal_libnl3_libs])
```

This would put it in both the stdout and in config.log.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,223 @@
...
+                              libnl_version=0
+                              AS_IF([echo $ldd_output | grep -q libnl.so],
+                                    [AS_IF([test $opal_libnl_version -eq 3],
+                                           [AC_MSG_WARN([lib nl version conflict: $opal_libnl3_libs requires libnl-3 whereas $1 requires libnl])],
+                                           [opal_libnl_libs="$opal_libnl_libs $1"
+                                            opal_libnl_version=1])
+                                     libnl_version=1])
+                              AS_IF([echo $ldd_output | grep -q libnl-3.so],
+                                    [AS_IF([test $libnl_version -eq 1],
+                                           [AC_MSG_ERROR([lib $1 requires both libnl and libnl3\nthis is just insane\nAborting...])])
```
jsquyres: This doesn't output like you would expect it to.  Here's what I get:

```
configure: error: lib fabric requires both libnl and libnl3\nthis is just insane\nAborting...
```

I think you probably want something like this:

```
AC_MSG_WARN([$1 requires both libnl and libnl3 -- yoinks!])
AC_MSG_WARN([This is a configuration that is known to cause run-time crashes])
AC_MSG_ERROR([Cannot continue])
```

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,223 @@
...
+    return 0;
+}
+EOF
+            OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1 $LIBS $3],
+                             [ldd_output=`ldd conftest`
+                              libnl_version=0
+                              AS_IF([echo $ldd_output | grep -q libnl.so],
+                                    [AS_IF([test $opal_libnl_version -eq 3],
+                                           [AC_MSG_WARN([lib nl version conflict: $opal_libnl3_libs requires libnl-3 whereas $1 requires libnl])],
+                                           [opal_libnl_libs="$opal_libnl_libs $1"
```
jsquyres: Might want to run opal_libnl_libs (and opal_libnl3_libs) through OPAL_UNIQ -- I got output like this at the end of a run:

```
libnl3: fabric ibverbs ibverbs ibverbs nl-3 ibverbs ibverbs
```

ggouaillardet:
@jsquyres  i made the requested changes
i also found something went really wrong and this did not work at alll (netlink reachable component was not built at all ...) and fixed that
can you please take an other look at this ? if ok, i will rebase, squash and commit the changes

configure.ac
```diff
@@ -1389,4 +1389,19 @@ m4_ifdef([project_orte], [ORTE_CONFIG_FILES])
...
+AS_IF([test -n "$opal_libnl_libs"],
+      [AC_MSG_RESULT([$opal_libnl_libs])],
+      [AC_MSG_RESULT([(none)])])
+AC_MSG_CHECKING([for libraries that use libnl3])
+AS_IF([test -n "$opal_libnl3_libs"],
+      [AC_MSG_RESULT([$opal_libnl3_libs])],
+      [AC_MSG_RESULT([(none)])])
+AS_IF([test -n "$opal_libnl_libs" && test -n "$opal_libnl3_libs"],
+      [AC_MSG_WARN([Both libnl and libnl3 are required])
+       AC_MSG_WARN([This is a configuration that is known to cause run-time crashes])])
```
jsquyres: I would add an AC_ERROR so that configure aborts.  We know that this will crash -- best to make configure abort and let the user figure it out (vs. a configure warning that scrolls by / is ignored, and then the user is confused later when all MPI applications crash).

config/opal_check_package.m4
```diff
@@ -134,6 +134,10 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_LIB], [
                   test "$ac_cv_search_$3" != "none required"],
                  [$1_LIBS="$ac_cv_search_$3 $4"],
                  [$1_LIBS="$4"])
+            # libnl and libnl3 are known *not* to coexist
```
jsquyres: Minor: update the comment to "libnl v1 and v3 are known to cause a crash if they are both linked into the same process."

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+	esac
+])
+
+dnl OPAL_LIBNL_SANITY_INIT()
+dnl --------------------------------------------------------------------
+AC_DEFUN([OPAL_LIBNL_SANITY_INIT], [
+    opal_libnl_version=0
+    opal_libnl_libs=
+    opal_libnl3_libs=
+    AC_ARG_WITH([libnl],
```
jsquyres: I find reading this patch to be a bit confusing -- because we use the string "libnl" to both mean "any version of the netlink library" and "version 1 of the netlink library."

Perhaps we should use this nomenclature:

- `libnl`: a version-agnostic way of referring to the netlink library
- `libnlv1` (as a variable, or `libnl v1` as prettyprint): refers to version 1 of the netlink library
- `libnlv3` (as a variable, or `libnl v3 as prettyprint): refers to version 3 of the netlink library

This makes it clear that `--with-libnl` refers to either version of the netlink library.  But all macros and variables are clear in whether they are dealing with v1 or v3.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+            cat > conftest_c.$ac_ext << EOF
+extern void $2 (void);
+int main(int argc, char *argv[[]]) {
+    $2 ();
+    return 0;
+}
+EOF
+            OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1 $LIBS $3],
+                             [ldd_output=`ldd conftest`
+                              libnl_version=0
```
jsquyres: Put `ldd_output` and `libnl_version` in `OPAL_VAR_SCOPE_PUSH`, and `OPAL_VAR_SCOPE_POP` at the bottom.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+                              AS_IF([echo $ldd_output | grep -q libnl.so],
+                                    [AS_IF([test $opal_libnl_version -eq 3],
+                                           [OPAL_UNIQ([opal_libnl3_libs])
+                                            AC_MSG_WARN([lib nl version conflict: $opal_libnl3_libs requires libnl-3 whereas $1 requires libnl])],
+                                           [opal_libnl_libs="$opal_libnl_libs $1"
```
jsquyres: You could use OPAL_APPEND_UNIQ here.  This would avoid using OPAL_UNIQ, above (and also in the outer configure.ac).

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+    return 0;
+}
+EOF
+            OPAL_LOG_COMMAND([$CC -o conftest $CFLAGS $CPPFLAGS conftest_c.$ac_ext $LDFLAGS -l$1 $LIBS $3],
+                             [ldd_output=`ldd conftest`
+                              libnl_version=0
+                              AS_IF([echo $ldd_output | grep -q libnl.so],
+                                    [AS_IF([test $opal_libnl_version -eq 3],
+                                           [OPAL_UNIQ([opal_libnl3_libs])
+                                            AC_MSG_WARN([lib nl version conflict: $opal_libnl3_libs requires libnl-3 whereas $1 requires libnl])],
```
jsquyres: Is there a reason we only warn here (and don't error/abort)?  Is the intent to go through and find *all* conflicts so that they can be reported at the end?  If so, it might be worth a comment (or another AC_MSG_WARN?) explaining this.

(same question applies a few lines down where we warn again [vs. abort])

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+dnl
+dnl Outputs:
+dnl
+dnl - Set $2_CPPFLAGS necessary to compile with libnl-3
+dnl - Set $2_LDFLAGS necessary to link with libnl-3
+dnl - Set $2_LIBS necessary to link with libnl-3
+dnl - Set HAVE_LIBNL3 1 if libnl-3 will be used
+dnl
+AC_DEFUN([OPAL_CHECK_LIBNL3],[
+    OPAL_VAR_SCOPE_PUSH([CPPFLAGS_save LIBS_save LDFLAGS_save])
```
jsquyres: Need to add opal_tmp_CPPFLAGS to the list.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,228 @@
...
+                              [nl_recvmsgs_report],
+                              [],
+                              [$1],
+                              [],
+                              [opal_libnl3_happy=1],
+                              [opal_libnl3_happy=0])
+
+           # Note that OPAL_CHECK_PACKAGE is going to add
+           # -I$dir/include into $2_CPPFLAGS.  But because libnl3
+           # puts the headers in $dir/libnl3, we need to
```
jsquyres: $dir/libnl3 is actually a typo (I know this came from the original code -- sorry!).  It should be $dir/include/libnl3.

(same issue on the next line, too)

opal/mca/reachable/netlink/configure.m4
```diff
@@ -71,74 +72,37 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
...
-       AC_SUBST([HAVE_LIBNL3])
-       AC_DEFINE_UNQUOTED([HAVE_LIBNL3],[$HAVE_LIBNL3],
-                          [set to 1 if should use libnl v3, set to 0 for libnl v11])
-
-       LIBS=$save_LIBS
-       AS_UNSET([save_LIBS])
-       CPPFLAGS=$save_CPPFLAGS
-       AS_UNSET([save_CPPFLAGS])
+	# Default to a numeric value (this value gets AC_DEFINEd)
+	HAVE_LIBNL3=0
```
jsquyres: It looks like this AC_DEFINE is specific to the reachable/netlink component.  It should be renamed to follow the prefix rule.

opal/mca/reachable/netlink/configure.m4
```diff
@@ -71,74 +72,37 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
...
+
+	###################################################
+	# NOTE: We *must* check for libnl3 before libnl.
+	###################################################
+
+	AS_IF([test "$opal_want_libnl" != "no" &&
+               test $opal_libnl_version -ne 1],
+	      [OPAL_CHECK_LIBNL3([$opal_libnl_location], [opal_nl])])
```
jsquyres: Can not use the variable `opal_nl` here -- must follow the prefix rule: `opal_reachable_nl`.

opal/mca/reachable/netlink/configure.m4
```diff
@@ -71,74 +72,37 @@ AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
...
+	AS_IF([test "$opal_want_libnl" != "no" &&
+               test $opal_libnl_version -ne 3 &&
+	       test "$opal_nl_LIBS" = ""],
+	      [OPAL_CHECK_LIBNL([$opal_libnl_location], [opal_nl])])
+
```
jsquyres: I'm not sure I understand the logic of the 2 preceding AS_IFs.

- We probably shouldn't even reach here if `$opal_want_libnl = no`.
- You check `$opal_libnl_version -ne 1` in the first one, and `-ne 3` in the second one.  Why is this?  Is this to check to see if some other part of the code base has already determined if we are using libnl v1 or v3?  If so, won't `$opal_nl` (or, `$opal_reachable_netlink` when it's renamed :wink:) be empty, and then the AS_IF in the next line will fail?

config/opal_check_libnl.m4
ggouaillardet: ```libnl_version``` is a "global variable' and should not be added,
but i will add ```ldd_output```

ggouaillardet:
@jsquyres i pushed a checkpoint of the changes you requested.

ideally, ```reachable/netlink``` should say "i can work with both libnl v1 and v3", and at the very end,
choose the "best" one.
currently, when configure'd with ```--without-verbs```, ```reachable/netlink``` will choose libnl v3 (if available) or fall back to libnl v1. if after that an other component finds it works only with libnl v1, then
all we can (currently) do is issue a warning / abort.

ibm-ompi:
Test passed.

ggouaillardet:
@jsquyres this PR should be good to go, do you want to give it a final review before i merge ?

lanl-ompi:
Test FAILed.

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,230 @@
...
+dnl $HEADER$
+dnl
+
+dnl
+dnl Helper function to parse --with-libnl* options
+dnl
+dnl $1: variable name
+dnl $2: --with-<foo> value
+dnl
+AC_DEFUN([OPAL_LIBNL_PARSE_WITH],[
```
jsquyres: I'm not quite sure what the purpose of this macro is -- did we use to call this for both `libnl` and `libnl3` (i.e., that's why there's a substitution on `$1`)?

I ask because I only see this macro being called for `libnl` now, which makes me wonder why this is still a macro with `$1` and `$2` substitution...?  Indeed, there's hard-coded places below where you check the value of `$opal_want_libnl`.

opal/mca/reachable/netlink/configure.m4
```diff
@@ -159,16 +121,7 @@ AC_DEFUN([MCA_opal_reachable_netlink_CONFIG],[
...
-                  [opal_reachable_netlink_LIBNL_CPPFLAGS],
-                  [opal_reachable_netlink_LIBNL_LIBS],
-                  [0])
-          ])
-    AS_IF([test "$opal_reachable_netlink_LIBNL_LIBS" == ""],
-          [opal_reachable_netlink_happy=0])
-
-    AC_SUBST(opal_reachable_netlink_LIBNL_CPPFLAGS)
-    AC_SUBST(opal_reachable_netlink_LIBNL_LIBS)
+          [OPAL_REACHABLE_NETLINK_CHECK_LIBNL_xxx])
```
jsquyres: I realize why you used the `_xxx` suffix (to distinguish that it's not "libnl"), but it's still a bit weird...

opal/mca/reachable/netlink/configure.m4
```diff
@@ -50,12 +51,12 @@ dnl - Set $1 to the CPPFLAGS necessary to compile with libnl
...
-dnl - Set HAVE_LIBNL3 to 1 if libnl3 will be used; 0 if libnl1 will be used
-dnl - AC_SUBST $HAVE_LIBNL3
-dnl - AC_DEFINE HAVE_LIBNL3
+dnl - Set OPAL_HAVE_LIBNL3 to 1 if libnl v3 will be used; 0 if libnl v1 will be used
+dnl - AC_SUBST $OPAL_HAVE_LIBNL3
+dnl - AC_DEFINE OPAL_HAVE_LIBNL3
 dnl
 dnl --------------------------------------------------------
-AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL3],[
+AC_DEFUN([OPAL_REACHABLE_NETLINK_CHECK_LIBNL_xxx],[
```
jsquyres: Should this macro be in `config/opal_check_libnl.m4`?  It seems like we want each component that needs to link against libnl to have a `configure.m4` that makes a trivial macro call to do all the libnl magic (e.g., similar to how `opal/mca/btl/openib` calls `OPAL_CHECK_OPENFABRICS` to do the bulk of its work).

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,230 @@
...
+		opal_want_$1=no
+		;;
+	yes)
+		# Yes, definitely want it
+		opal_want_$1=yes
+		;;
+	default)
+		# Default case -- try and see if we can find it
+		opal_want_$1=default
+		opal_$1_location=/usr
```
jsquyres: Setting a `-I` or `-L` to `/usr` is almost never what you want to do.  Is that what happens with this value?

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,230 @@
...
+dnl
+dnl Copyright (c) 2015-2016 Research Organization for Information Science
+dnl                         and Technology (RIST). All rights reserved.
+dnl $COPYRIGHT$
+dnl
+dnl Additional copyrights may follow
+dnl
+dnl $HEADER$
+dnl
+
```
ggouaillardet: done

config/opal_check_libnl.m4
ggouaillardet: i inlined this macro

config/opal_check_libnl.m4
ggouaillardet: this value is never used this way. it is passed to ```OPAL_CHECK_PACKAGE``` in which ```/usr``` or ```/usr/local``` is kind of removed

configure.ac
```diff
@@ -1449,6 +1449,20 @@ m4_ifdef([project_orte], [ORTE_CONFIG_FILES])
 m4_ifdef([project_ompi], [OMPI_CONFIG_FILES])
 m4_ifdef([project_oshmem], [OSHMEM_CONFIG_FILES])
 
+OPAL_UNIQ([opal_libnlv1_libs])
```
ggouaillardet: i moved the macro and added some comment

opal/mca/reachable/netlink/configure.m4
ggouaillardet: i replaced the name with ```OPAL_REACHABLE_NETLINK_CHECK_LIBNL_Vx```

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,262 @@
...
+dnl
+dnl More libnl v1/v3 sadness: the two versions are not compatible
+dnl and will not work correctly if simultaneously linked into the
+dnl same applications.  Unfortunately, they *will* link into the
+dnl same image!  On platforms like SLES 12, libibverbs depends on
+dnl libnl-3.so.200 and friends, while a naive implementation of
+dnl our configure logic would link libnl.so.1 to libdaplusnic,
+dnl resulting in both versions in the dependency map at the same
+dnl time.
```
jsquyres: @ggouaillardet I think you might want to edit this comment (vs. blind copy-n-paste).  E.g., the references here to DAPL aren't relevant.  😄 

What I'm looking for here is both a description of the problem (which this first paragraph essentially does), and also at least some description of the overall strategy we're doing here in this M4 file to try to avoid the problem (e.g., after each OPAL_CHECK_PACKAGE, check to see if any of the dependent libraries include libnl/libnl3, detect conflicts, ...etc.).

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,262 @@
...
+            opal_want_libnl=no
+            ;;
+        yes)
+            # Yes, definitely want it
+            opal_want_libnl=yes
+            ;;
+        default)
+            # Default case -- try and see if we can find it
+            opal_want_libnl=default
+            opal_libnl_location=/usr
```
jsquyres: If this /usr is just going to be stripped away in OPAL_CHECK_PACKAGE, then we probably shouldn't even list it here (i.e., leave it blank), because every time someone looks at this, they'll say "Oh, you shouldn't `-I/usr` or `-L/usr`, and spend time tracking down just to figure out that it's both useless and harmless.

config/opal_check_package.m4
```diff
@@ -138,6 +138,10 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_LIB], [
                   test "$ac_cv_search_$3" != "none required"],
                  [$1_LIBS="$ac_cv_search_$3 $4"],
                  [$1_LIBS="$4"])
+            # libnl v1 and libnl3 are known *not* to coexist
+            # for each library, figure out whether it depends on libnl or libnl3 or none
+            # so conflicts can be reported and/or prevented
+            OPAL_LIBNL_SANITY_CHECK([$2], [$3], [$$1_LIBS])
```
jsquyres: Does one of these `$`'s need to be m4-escaped?

config/opal_check_libnl.m4
```diff
@@ -0,0 +1,262 @@
...
+    AC_MSG_CHECKING([for libraries that use libnl v1])
+    AS_IF([test -n "$opal_libnlv1_libs"],
+          [AC_MSG_RESULT([$opal_libnlv1_libs])],
+          [AC_MSG_RESULT([(none)])])
+    AC_MSG_CHECKING([for libraries that use libnl v3])
+    AS_IF([test -n "$opal_libnlv3_libs"],
+          [AC_MSG_RESULT([$opal_libnlv3_libs])],
+          [AC_MSG_RESULT([(none)])])
+    AS_IF([test -n "$opal_libnlv1_libs" && test -n "$opal_libnlv3_libs"],
+          [AC_MSG_WARN([Both libnl v1 and libnl v3 are required])
```
jsquyres: The use of the word "required" here is a bit confusing -- a user could read that as "to build Open MPI, it's telling me that both libnl v1 and v3 are required, but then it also tells me that this is an error!"

Maybe s/are required/have been found as dependent libraries/ ...?

opal/mca/reachable/netlink/configure.m4
jsquyres: This comment was not addressed.

config/opal_check_package.m4
ggouaillardet: i think it is ok as it is

ggouaillardet:
@jsquyres i pushed an other commit with the requested changes (plus a bit of revamp)
if it works for you, i will squash, rebase and merge

ggouaillardet:
@hppritcha the URL sets by ```LANL-CI``` Jenkins is invalid, can you please have a look ?

ibm-ompi:

 Gist: https://gist.github.com/dc4747f4f5fd8e7c6c05fa37dbf4c5d1

ibm-ompi:

 Gist: https://gist.github.com/38549f772045a006e10d3ac771bbe9c7

ibm-ompi:

 Gist: https://gist.github.com/642ced56860259e02256422416b925e3

jsquyres:
bot:retest

ibm-ompi:

 Gist: https://gist.github.com/6af01971e176ba666d0c8f8022d1a495

ibm-ompi:

 Gist: https://gist.github.com/236a41a0cec1ff3049103eaac39a9c80

jsquyres:
@ggouaillardet The errors that are coming back seem legit.

ibm-ompi:

 Gist: https://gist.github.com/ba0898d40197da70870ab3c6be59a10a

ggouaillardet:
@jsquyres i will fix that. that seems to happen on systems without any libnl devel packages

ggouaillardet:
@jsquyres i fixed that and updated the PR

jsquyres:
@ggouaillardet This looks good to me now; thanks!

Can you squash it down to one commit?  I think we'll then be good to go.

ggouaillardet:
Done !

jsquyres:
@ggouaillardet Let's let this soak on master for a bit.  Given that v2.1.0 was deleted, there may even still be time to make it into v2.1.0.  If not, v2.2.0 sounds like a fine target.

ggouaillardet:
From a pragmatic point of view, the issue was only reported with mca/reachable, and it is not in v2.x
So there is no rush to PR this

jsquyres:
FWIW: we've seen this issue between the usnic BTL and the openib BTL

* usnic: libfabric -> libnl
* openib: libverbs -> librdmacm -> libnl

I agree: this is not urgent.  But it would be nice to get into the v2.x train at some point.

PR #1017, opened 2015-10-13
dynamics: fix OPAL_CR_EXIT_LIBRARY()

jsquyres:
Noticed that these were wrong will working on a different pull
request.  Submit these fixes indepdent of other changes, just to keep
things separated.

### DIFF
--- a/ompi/mpi/c/comm_accept.c
+++ b/ompi/mpi/c/comm_accept.c
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2006      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2006-2015 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2008      University of Houston, Inc.  All rights reserved.
  * Copyright (c) 2013      Los Alamos National Security, LLC.  All rights
  *                         reserved.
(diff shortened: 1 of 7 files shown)


jsquyres:
Random note: no need to bring this to v2.x, because the CR infrastructure is being removed over there.

hppritcha:
this should not be pushed back to v2.x.   The ```OPAL_CR_EXIT_LIBRARY``` and friends are being removed from that branch.

rhc54:
should it even be in master?? I thought we were going the other direction and cleaning up?

jsquyres:
I thought the CR stuff was staying on master...?

I don't really care -- I am happy to deleted this PR; I just noticed that these were absolutely wrong while I was working on some a different PR.

rhc54:
I honestly don't know - if we are going to restore it, we probably would have to start from scratch anyway as the overall arch has changed so much (e.g., I don't think the CR system has any way of dealing with dynamic add_procs, or fenceless operations today). This is why the last attempted stalled as it was trying to force-fit itself into the prior implementation.

Probably worth asking @bosilca and @jjhursey for their opinions. I'm ambivalent - was just curious as to why this PR existed given that the macro is supporting broken functionality.

bosilca:
Any C/R based approach needs to know when we enter and leave the library, so these calls can be easily used by any future FT layer. Moreover, it took a significant effort to put these ENTER and EXIT macros in, and today they always translate to an empty statement, so they are not bothersome. I will strongly support to keep them inside the source code by now.

rhc54:
Sounds good to me - thanks!

jjhursey:
I think this is fine as well per George's comments. They are useful hooks to have in the API for FT (any possibly other) extensions, and are empty macro expansions for anyone not playing with the feature.

PR #1026, opened 2015-10-14
opal static mutex initializers

hjelmn:


hjelmn:
@bosilca, @jsquyres please take a look.

Some time ago I asked for this support for MPI_T but we chose not to add it because we supported non-posix mutexes. That is no longer the case so we might as well add them.

It might also be worth considering adding an equivalent to pthread_once.

hjelmn:
I may remove the debug initializers since they only set some members to 0. This is already guaranteed by the C standard if the variable being initialized has global or file scope.

jsquyres:
:+1:

hjelmn:
One change. Some platforms define PTHREAD_RECURSIVE_MUTEX_INITIALIZER and others use PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP. Added support for the later.

PR #1028, opened 2015-10-15
pmix1xx configury: invoke sub-configure with CFLAGS and CPPFLAGS on t…

ggouaillardet:
…he command line

if CFLAGS and/or CPPFLAGS are passed to the ompi configure command line, pmix1xx
configure will not use the correct ones previously passed in the environment
see discussion started at http://www.open-mpi.org/community/lists/devel/2015/10/18159.php
Thanks Siegmar Gross for bringing this to our attention

### DIFF
--- a/opal/mca/pmix/pmix1xx/configure.m4
+++ b/opal/mca/pmix/pmix1xx/configure.m4
@@ -46,11 +46,9 @@ AC_DEFUN([MCA_opal_pmix_pmix1xx_CONFIG],[
         CFLAGS="$OPAL_CFLAGS_BEFORE_PICKY $OPAL_VISIBILITY_CFLAGS"
     fi
     CPPFLAGS="-I$OPAL_TOP_SRCDIR -I$OPAL_TOP_BUILDDIR -I$OPAL_TOP_SRCDIR/opal/include -I$OPAL_TOP_BUILDDIR/opal/include $CPPFLAGS"
-    export CFLAGS
-    export CPPFLAGS
 
     OPAL_CONFIG_SUBDIR([$opal_pmix_pmix1xx_basedir/pmix],
-        [$opal_pmix_pmix1xx_args $opal_subdir_args],
+        [$opal_pmix_pmix1xx_args $opal_subdir_args 'CFLAGS=$CFLAGS' 'CPPFLAGS=$CPPFLAGS'],
         [opal_pmix_pmix1xx_happy=1], [opal_pmix_pmix1xx_happy=0])
 
     if test $opal_pmix_pmix1xx_happy -eq 1; then


rhc54:
@jsquyres Could you please take a look at this? I don't understand why the lines that set CPPFLAGs and friends above the one being changed aren't picking up any cmd-line settings. I thought those would have been included in either the opal flags or in the CPPFLAGS envar?

ggouaillardet:
@rhc54 
i guess my explanation was not clear, so let me try again.
if ompi is configure'd with CPPFLAGS=XXX, ```opal/mca/pmix/pmix1xx/configure.m4``` did
```
export CPPFLAGS=-I/.../hwloc/include -I/.../libevent/include ... XXX
opal/mca/pmix/pmix1xx/pmix/configure ... CPPFLAGS=XXX
```
in this case, the ```CPPFLAGS``` environment variable is simply ignored because ```CPPFLAGS``` is set on the sub-configure command line. alas, we do need the value from the environment and not from the command line.

the simplest option, is to append ```CPPFLAGS=...``` to the sub-configure cli invokation (CPPFLAGS= might appear twice, but we do not care).
more elegant alternative could be to remove CPPFLAGS= from ```$opal_subdir_args``` (so the environment variable is used) or replace (instead of append) CPPFLAGS from ```$opal_subdir_args``` (and there is no more need to set the environment variable)

make sense ?

jsquyres:
Tagging @siegmargross, since he reported the issue.

jsquyres:
@ggouaillardet Shouldn't CFLAGS and CPPFLAGS be assigned thusly:

```sh
CPPFLAGS="$CPPFLAGS -I$OPAL_TOP_SRCDIR ..."
```

rather than overwriting them?  Then the export statements will properly propagate them to the sub-configure, and all is fine.

ggouaillardet:
@jsquyres 
currently the previous CPPFLAGS value is append'ed to the new one(e.g. not overwritten) you suggest to prepend. I have no opinion on which is best.

the issue is the export statement does nothing useful

the sub-configure is invoked with $opal_subdir_args that contains CPPFLAGS= from the ompi configure invocation (and it has no reference to helicopter nor lib event headers)
since there is something on the cli, sub-configure simply ignores the environment.

in order to evidence this, you can simply 
configure CPPFLAGS=""
and then check
opal/mca/pmix/pmix1xx/pmix/config.status
to get the sub-configure cli
feel free to rerun the sub-configure with any value for the CPPFLAGS environment variable
(it does not matter since the environment is ignored)

rhc54:
@jsquyres You have any further thoughts on this? We would certainly need it in 2.0, if this is correct.

jsquyres:
Scheesh; I forgot about this PR when I filed #1069 this morning.

@ggouaillardet Your explanations make perfect sense; thanks.  :+1:

PR #1029, opened 2015-10-15
pml/ob1: Fix a memory leak regarding pending FIN control messages.

kawashima-fj:
In the current ob1 PML implementation, once a FIN control message is appended to the pending list,
the ob1 PML attempts to send the FIN again in the `mca_pml_ob1_process_pending_packets` function. But if the PML failed to sent the FIN again, the `mca_pml_ob1_send_fin` function creates a new `mca_pml_ob1_pckt_pending_t` object and the old object is not retured to the free list.

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1.c
+++ b/ompi/mca/pml/ob1/pml_ob1.c
@@ -16,7 +16,8 @@
  * Copyright (c) 2011      Sandia National Laboratories. All rights reserved.
  * Copyright (c) 2011-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
- * Copyright (c) 2012 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


hjelmn:
:+1:

hjelmn:
Please PR to 2.x and 1.10.

bosilca:
Are we sure about this ? It looks like we are dropping a fragment that has not been successfully sent instead of adding it back to the list.

hjelmn:
I think this is correct. In the error path of mca_pml_ob1_send_fin we call MCA_PML_OB1_ADD_FIN_TO_PENDING which allocates a new pending packet and puts it on the pending list. In mca_pml_ob1_process_pending_packets if mca_pml_ob1_send_fin it was just returning and not calling MCA_PML_OB1_PCKT_PENDING_RETURN.

PR #1032, opened 2015-10-15
limit the number of bytes used for the semaphore name depending on pl…

edgargabriel:
…atform (31 bytes for MacOS, 252 for Linux)


hppritcha:
this needs to go in to v2.x right?

edgargabriel:
yes, I will add it to the cummulative pr that I am preparing to bring ompio on v2.x up to date with master

PR #1039, opened 2015-10-19
cm: fix selection priority

### LINKED ISSUE #1035: PML selection issues
We are observing two issues with the way the PML is selected:

1) on Truescale, when mpirun is used without any argument, the openib BTL (ob1 PML) gets selected over the PSM MTL (cm PML). This is a problem because users would like to get the best performance (with the PSM MTL) out-of-the-box. Instead, they have to add --mca pml  cm.

2) currently, whichever of the ob1 or the cm PML gets initialized/loaded first wins! Unfortunately, the components are not always loaded in the same order...

hjelmn:
This patch removes a priority check that disables cm if the previous
pml had higher priority. The check was incorrect as coded and is
unnecessary as we finalize all but one pml anyway.

Fixes #1035

### DIFF
--- a/ompi/mca/pml/cm/pml_cm_component.c
+++ b/ompi/mca/pml/cm/pml_cm_component.c
@@ -143,11 +143,6 @@ mca_pml_cm_component_init(int* priority,
 {
     int ret;
 
-    if((*priority) > ompi_pml_cm.default_priority) {
-        *priority = ompi_pml_cm.default_priority;
-        return NULL;
-    }
-    *priority = ompi_pml_cm.default_priority;
     opal_output_verbose( 10, 0,
                          "in cm pml priority is %d\n", *priority);
     /* find a useable MTL */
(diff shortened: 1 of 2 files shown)


jsquyres:
Per https://github.com/open-mpi/ompi/issues/1035#issuecomment-148560757, should the same check be made to ob1?

hjelmn:
@jsquyres Removed the ob1 check as well. I see no reason to disable any pml component in this manner.

bosilca:
With this approach we don't need the field default_priority in the CM PLM.

jsquyres:
@yburette Does this fix your openib / PSM issue?

hjelmn:
@bosilca I will remove the default priority in CM completely in another commit. I want to have the mtl components set the priority of cm if a different manner. Should have the PR up soon.

hjelmn:
For history's sake here is the commit that added the extra selection checks:
open-mpi/ompi@a94101fa

Since dr is dead I think it is safe to remove them.

hjelmn:
Will handle this all in #1040.

PR #1040, opened 2015-10-19
Change how cm's priority is calculated

hjelmn:
@bosilca Please comment. This is a companion to or replacement for #1039.

### DIFF
--- a/ompi/mca/crcp/base/crcp_base_select.c
+++ b/ompi/mca/crcp/base/crcp_base_select.c
@@ -162,7 +162,7 @@ int ompi_crcp_base_select(void)
     if( OPAL_SUCCESS != mca_base_select("crcp", ompi_crcp_base_framework.framework_output,
                                         &ompi_crcp_base_framework.framework_components,
                                         (mca_base_module_t **) &best_module,
-                                        (mca_base_component_t **) &best_component) ) {
+                                        (mca_base_component_t **) &best_component, NULL) ) {
         /* This will only happen if no component was selected */
         return OMPI_ERROR;
     }
--- a/ompi/mca/mtl/base/base.h
+++ b/ompi/mca/mtl/base/base.h
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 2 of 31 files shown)


bosilca:
Looks good. :+1:

yburette:
@hjelmn With this priority scheme, one needs to specify which PML to use in order to be able to use the OFI MTL. Is there a way around this?

`--mca mtl ofi --mca mtl_ofi_provider_include "psm"` doesn't work; still selects ob1
`--mca mtl ofi --mca mtl_ofi_provider_include "psm" --mca pml cm` works

hjelmn:
I didn't adjust ofi's priority. I didn't think it was considered ready for prime time. You can set using -mca mtl_ofi_priority 100 to force it to run. Should we increase the priority?

yburette:
I don't think we need to increase the priority yet.

I was just hoping there would be a way to automagically select cm when an MTL is specified on the command line :-)

hjelmn:
You need to still force cm (or bump ofi's priority) because it shouldn't be selected before ob1 in production. Right now the component initialization code can not tell the difference between only one component being available and the user selecting a single component.

hjelmn:
I should not it worked before because someone added ofi as one of the components that bumped cm's priority to 30.

yburette:
So if I understand well, we should increase OFI MTL's priority to 30 in order to have what we had before the change, right?

hjelmn:
Sort of. We need to bump psm to 31 first to keep it ahead of ofi. I left ofi's relative priority alone here (10).

bosilca:
My understanding of the code is that the CM and PML will now inherit the priorities of their underlying modules (the CM the max of all available MTL, and OB1 the max of all available BTL), or is getting it's own priority form the corresponding MCA parameter. Thus if we have a purely priority-based solution we can make everything work.

By default we can order the MTL in whatever order we consider adequate, but always higher than any BTL. Thus, as the OB1 will inherit the max priority of the BTLs, if any MTL is ready to work the CM (who will inherit this priority) will be picked up. Selecting the OB1 PML become as simple as increasing it's priority to something larger than any MTL.

Sounds simple to implement, and pretty logical for the users. It does change the default (MTL instead of PML), but the change is sensible and desirable.

hjelmn:
@bosilca Pretty much. The only difference is the BTLs don't have a priority since there may be many transports in use. So right now ob1 uses its own priority. In the future it might be useful to determine the priority for ob1 using the active btls. This way we could select ob1 when running only intra-node since vader is much faster than both mxm and psm.

PR #1044, opened 2015-10-20
opal/keyval: reset buffer pointer/size in finalize

hjelmn:


hjelmn:
@jsquyres Not sure how this was not caught by the MPI_T tests but I am seeing a SEGV because of this with one-sided/c_lock_illegal

jsquyres:
:+1:

Issue #1010, opened 2015-10-12
ibm/c_accumulate (and friends) all fail with master


rolfv:
When I run the ibm/onesided/c_accumulate test over openib I get a SEGV.
```
[rvandevaart@drossetti-ivy4 onesided]$ mpirun --host drossetti-ivy4,drossetti-ivy5 -np 2 -mca btl self,openib c_accumulate
--------------------------------------------------------------------------
mpirun noticed that process rank 0 with PID 0 on node drossetti-ivy4 exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
[rvandevaart@drossetti-ivy4 onesided]$ 
```
```
(gdb) where
#0  0x00007fbdf4bc41d0 in ?? ()
#1  <signal handler called>
#2  0x00007fbdf76fda88 in mca_btl_openib_atomic_internal (btl=0x24aa140, endpoint=0x0, local_address=0x7fbdfd786010, remote_address=42016232, local_handle=0x27ca2a0, remote_handle=0x27ca5a0, 
    opcode=IBV_WR_ATOMIC_CMP_AND_SWP, operand=0, operand2=0, flags=0, order=0, cbfunc=0x7fbdf41a7d94 <ompi_osc_rdma_atomic_complete>, cbcontext=0x7fffa8010ca7, cbdata=0x0)
    at ../../../../../opal/mca/btl/openib/btl_openib_atomic.c:85
#3  0x00007fbdf76fdca9 in mca_btl_openib_atomic_cswap (btl=0x24aa140, endpoint=0x0, local_address=0x7fbdfd786010, remote_address=42016232, local_handle=0x27ca2a0, remote_handle=0x27ca5a0, compare=0, 
    value=9223372036854775808, flags=0, order=0, cbfunc=0x7fbdf41a7d94 <ompi_osc_rdma_atomic_complete>, cbcontext=0x7fffa8010ca7, cbdata=0x0) at ../../../../../opal/mca/btl/openib/btl_openib_atomic.c:138
#4  0x00007fbdf419f5f6 in ompi_osc_rdma_lock_try_acquire_exclusive (module=0x27e3f40, peer=0x280bd80, offset=16) at ../../../../../ompi/mca/osc/rdma/osc_rdma_lock.h:201
[...]

PR #1046, opened 2015-10-20
osc/rdma: bug fixes

hjelmn:
This commit fixes several bugs in the osc/rdma component:

 - Complete aggregated requests immediately. Completion of RMA
   requests indicates local completion anyway. This fixes a hang in
   the c_reqops test.

 - Correctly mark Rget_accumulate requests.

 - Set the local base flag correctly on the local peer.

 - Clear or set the no locks flag on the window if the value is
   changed by MPI_Win_set_info.

 - Actually update the target when using MPI_OP_REPLACE.

Fixes open-mpi/ompi#1010


hjelmn:
@rolfv This fixes the MTT problems you identified.

There is still an outstanding issue with the openib btl that MTT does not test. The connection manager is not completely asynchronous. This can cause issues when using passive-target to a peer that has not been communicated with before. I am working on a fix for that issue.

PR #1047, opened 2015-10-21
Cleanup the temporary memory allocation in collectives

bosilca:
Cleanup the memory handling for temporary buffers in some of the collective modules. Added a new function opal_datatype_span, to compute the memory span of count number of datatype, excluding the gaps in the beginning and at the end. If a memory allocation is made using the returned value, the gap (also returned) should be removed from the allocated pointer.

### DIFF
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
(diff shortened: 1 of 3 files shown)


ggouaillardet:
what about using a new function ?

```
int opal_datatype_alloc(opal_datatype_t * type, int count, void ** buf, void ** buf_free);
```

the rationale is code factorization, plus most of the time, we only need ```true_extent``` for
```
buf = buf_free - true_extent
```

bosilca:
The difference with what is already inside is marginal (yours allocate and returns memory, mine if returning everything you need to do it right yourself). I prefer not to have a memory allocator down in OPAL datatype, but instead let people handle this the way they want/need.

jsquyres:
bot:retest

PR #1053, opened 2015-10-21
udcm: fix bugs

hjelmn:
This commit fixes the following bugs:

 - On send failure release newly allocated message.

 - In the destructor for udcm_message_sent_t always remove the send
   timeout event from the event base. Failure to do this can lead to
   memory corruption since the destructor may be called from an event
   callback.


hjelmn:
@miked-mellanox This fixes a long-outstanding issue and is targeted for 1.10.1 and 2.0.0. Without this fix I see random udcm crashes when I hit the CM with lots of requests.

hjelmn:
I should also add that this is a quick fix for a real issue. I am rewriting the connection part of the openib btl to be truly asynchronous. This is needed to support truly passive one-sided communication.

PR #1054, opened 2015-10-21
add_procs: add threading protection for dynamic add_procs

hjelmn:
This commit add protection to the group, ob1, and bml endpoint lookup
code. For ob1 and the bml a lock has been added. For performance
reasons the lock is only held if a bml or ob1 endpoint does not
exist. ompi_group_dense_lookup no uses opal_atomic_cmpset to ensure
the proc is only retained by the thread that actually updates the
group.

### DIFF
--- a/ompi/mca/bml/base/base.h
+++ b/ompi/mca/bml/base/base.h
@@ -63,10 +63,15 @@ OMPI_DECLSPEC  int mca_bml_base_ft_event(int state);
 OMPI_DECLSPEC extern mca_bml_base_component_t mca_bml_component;
 OMPI_DECLSPEC extern mca_bml_base_module_t mca_bml;
 OMPI_DECLSPEC extern mca_base_framework_t ompi_bml_base_framework;
+OMPI_DECLSPEC extern opal_mutex_t mca_bml_lock;
 
 static inline struct mca_bml_base_endpoint_t *mca_bml_base_get_endpoint (struct ompi_proc_t *proc) {
     if (OPAL_UNLIKELY(NULL == proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML])) {
-        mca_bml.bml_add_proc (proc);
+        OPAL_THREAD_LOCK(&mca_bml_lock);
+        if (NULL == proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML]) {
+            mca_bml.bml_add_proc (proc);
+        }
+        OPAL_THREAD_UNLOCK(&mca_bml_lock);
     }
 
     return (struct mca_bml_base_endpoint_t *) proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML];
--- a/ompi/mca/pml/ob1/pml_ob1_comm.h
+++ b/ompi/mca/pml/ob1/pml_ob1_comm.h
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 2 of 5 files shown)


hjelmn:
@jsquyres @bosilca This will need to go to 2.x after it is in master. Please take a look and let me know if I should change anything or if there are any additional areas that need threading protection.

Note I use the OPAL_THREAD versions of the mutex functions but there as the protected code is not in the critical path there is no reason not to use the opal_mutex functions.

hjelmn:
@hppritcha This doesn't fix the hang @bosilca ran into. The email chain reminded me to double-check that the appropriate thread protections were in place for dynamic add_procs.

lanl-ompi:
Test FAILed.

hjelmn:
@hppritcha Weird failure seems unrelated to the PR.

hppritcha:
bot:retest

ompi/mca/pml/ob1/pml_ob1_comm.h
```diff
@@ -71,9 +71,14 @@ static inline mca_pml_ob1_comm_proc_t *mca_pml_ob1_peer_lookup (struct ompi_comm
     mca_pml_ob1_comm_t *pml_comm = (mca_pml_ob1_comm_t *)comm->c_pml_comm;
 
     if (OPAL_UNLIKELY(NULL == pml_comm->procs[rank])) {
-        pml_comm->procs[rank] = OBJ_NEW(mca_pml_ob1_comm_proc_t);
-        pml_comm->procs[rank]->ompi_proc = ompi_comm_peer_lookup (comm, rank);
-        OBJ_RETAIN(pml_comm->procs[rank]->ompi_proc);
+        OPAL_THREAD_LOCK(&pml_comm->proc_lock);
+        if (NULL == pml_comm->procs[rank]) {
```
jsquyres: Didn't you just do this same comparison on line 73?

hjelmn: Yup. One without the lock (to avoid locking when not necessary) and another after the lock to ensure that only one thread creates the proc.

ompi/mca/bml/base/base.h
```diff
@@ -63,10 +63,15 @@ OMPI_DECLSPEC  int mca_bml_base_ft_event(int state);
...
+OMPI_DECLSPEC extern opal_mutex_t mca_bml_lock;
 
 static inline struct mca_bml_base_endpoint_t *mca_bml_base_get_endpoint (struct ompi_proc_t *proc) {
     if (OPAL_UNLIKELY(NULL == proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML])) {
-        mca_bml.bml_add_proc (proc);
+        OPAL_THREAD_LOCK(&mca_bml_lock);
+        if (NULL == proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML]) {
+            mca_bml.bml_add_proc (proc);
+        }
+        OPAL_THREAD_UNLOCK(&mca_bml_lock);
```
jsquyres: I don't know the implications of this lock.  Is there any possibility of a deadlock here?  I.e., is there any conceivable way that the `btl.add_procs()` could somehow end up calling `mca_bml_base_get_endpoint()`?

hjelmn: its possible a btl *could* progress incoming messages in add_procs but none do so now.  that would be the only way we could reenter here. if we want to allow add_procs to be reentrant we can make the mutex recursive.

jsquyres: Should we assert() fail here if it re-enters, or somesuch?

hjelmn: We will assert fail in the mutex lock in debug builds already. It will say something along the lines of deadlock avoided.

jsquyres:
:+1:

Issue #1012, opened 2015-10-12
CUDA-aware one sided put does not work on master


rolfv:
With the new osc/rdma component, the CUDA-aware support does not work anymore on onesided put operations. Interestingly, it seems to work on get. I need to investigate but I want to record this bug so that I do not forget it. Also, it would be nice for this to work prior to the osc/rdma coming over to v2.x which I assume it will. Is that true @hjelmn ?

hjelmn:
Yes. This is intended for 2.x. Let me know if you can determine why puts are failing. It would be great if the new component can make use of the GPU direct support.

rolfv:
I think they are a few things happening here and maybe we can work our way through them. First, I will try to get osc/pt2pt to continue to work on master. Does that seem OK @hjelm or is it going to go away? It may be nice to keep it for a little while so in case we do not get CUDA-aware working with osc/rdma we will have a fallback.  If so, then any idea why I would run into this issue? I only see this with CUDA-aware.  Although interesting I do not see this particular failure on 2.x.
```
[rvandevaart@drossetti-ivy4 nv]$ mpirun --mca osc pt2pt -host drossetti-ivy4,drossetti-ivy5 -np 2 --mca btl self,smcuda,openib osu_get_bw_cuda
# OSU MPI2 MPI_Get Bandwidth Test (Version 2.2)
# Size        Bandwidth (MillionBytes/s) 
--------------------------------------------------------------------------
mpirun noticed that process rank 0 with PID 0 on node drossetti-ivy4 exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
[rvandevaart@drossetti-ivy4 nv]$ 
```

```
(gdb) where
#0  0x00007f04887951d0 in ?? ()
#1  <signal handler called>
#2  0x00007f0487f87fa1 in ompi_osc_pt2pt_sync_array_peer (rank=1, peers=0xacfef70, nranks=1, peer=0x0)
    at ../../../../../ompi/mca/osc/pt2pt/osc_pt2pt_sync.c:58
#3  0x00007f0487f880cf in ompi_osc_pt2pt_sync_pscw_peer (module=0xad04140, target=1, peer=0x0)
    at ../../../../../ompi/mca/osc/pt2pt/osc_pt2pt_sync.c:89
#4  0x00007f0487f855d2 in osc_pt2pt_incoming_post (module=0xad04140, source=1)
[...]

PR #1056, opened 2015-10-22
osc/pt2pt: reset all_sync sync object before sending complete messages

hjelmn:
This commit fixes a bug that occurs when a post message comes in when
sending complete messages or while waiting for all outgoing messages
to flush. In that case the post message might get incorrecly
associated with the ending sync object.

References open-mpi/ompi#1012


hjelmn:
@rolfv This doesn't prevent osc/rdma from trying to work but it fixes osc/pt2pt. In order to disable osc/rdma I need a way to detect that the user buffer is a cuda buffer. Is there a good way to do that?

PR #1057, opened 2015-10-22
win: fix erroneous argument check

hjelmn:
When using dynamic memory windows the displacement becomes a
pointer. Since the high bit may be set on valid pointers on some
platforms the check for disp > 0 is invalid. This commit adds the
window flavor to ompi_win_t and disables the displacement check when
operating on dynamic memory windows.

### DIFF
--- a/ompi/mpi/c/accumulate.c
+++ b/ompi/mpi/c/accumulate.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2009      Sun Microsystmes, Inc.  All rights reserved.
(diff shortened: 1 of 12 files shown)


hjelmn:
This commit keeps the displacement check where it is valid but another option would be to remove the check entirely. The underlying osc component will likely return an RMA range error in that case.

hjelmn:
@ggouaillardet You have a platform that could potentially be affected by this commit. Can you see if the aint tests in ompi-test/ibm works with this change?

hjelmn:
Next targets: v1.10.1, v2.0.0

hjelmn:
Going to push and then PR to 1.10 and 2.x. @ggouaillardet please review there.

PR #1058, opened 2015-10-22
mtl/ofi: Increase priority.

yburette:
@hjelmn I think we can increase the priority of the OFI MTL.
This would automatically select the CM PML when mpirun is run with
`--mca mtl ofi`

By default, the OFI MTL won't run over the sockets and MXM providers.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -56,7 +56,7 @@ mca_mtl_ofi_component_t mca_mtl_ofi_component = {
 static int
 ompi_mtl_ofi_component_register(void)
 {
-    param_priority = 10;   /* for now give a lower priority than the psm mtl and ob1 */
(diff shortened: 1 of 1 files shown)


yburette:
bot:milestone:v1.10.1

hjelmn:
Does the ofi component disqualify itself with the verbs provider. I just want to make sure that it doesn't override the openib btl on mlx hardware.

rhc54:
Ditto for portals, unless @regrant and @tkordenbrock are okay with it

yburette:
@hjelmn The OFI MTL won't run with the verbs provider because the verbs provider doesn't support tag messaging.

hjelmn:
@yburette Great. I am ok with merging this and moving it to 1.10.1. I will include it with the 2.x PR for the cm changes.

yburette:
@hjelmn Thanks a lot!

regrant:
Portals can have low priority, we don't have a problem specifying when you want to use it. I don't expect users to use Portals as a default in the near future.

Issue #1065, opened 2015-10-23
Many buffered tests are failing on master


rolfv:
This problem is in master only and I tracked it down to this commit using git bisect.
```
[rvandevaart@drossetti-ivy5 ompi-master-rolfv]$ git bisect good
Bisecting: 0 revisions left to test after this (roughly 1 step)
[720fa860eee7df4f06eb5604058da57453e88b46]   changing the destruct function of list release API to release list items   caused a bug in oshmem application. Fixing the bug with this patch
[rvandevaart@drossetti-ivy5 ompi-master-rolfv]$ 
```
Although I cannot quite figure out what that commit is when I look at the log.  In any event, the errors can be seen on many different buffered send tests.
```
[rvandevaart@drossetti-ivy4 src]$ mpirun -np 2 MPI_Bsend_ator_c
--------------------------------------------------------------------------
MPITEST info  (0): Starting MPI_Bsend_ator: All Bsend TO Root test
MPITEST info  (0): Node spec MPITEST_comm_sizes[6]=2 too large, using 1
MPITEST info  (0): Node spec MPITEST_comm_sizes[22]=2 too large, using 1
MPITEST info  (0): Node spec MPITEST_comm_sizes[32]=2 too large, using 1
--------------------------------------------------------------------------
mpirun noticed that process rank 1 with PID 0 on node drossetti-ivy4 exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
```

```
(gdb) where
#0  0x00007f16a17ef1d0 in ?? ()
#1  <signal handler called>
#2  0x00007f16a81e3abb in opal_memory_ptmalloc2_int_free (av=0x7f16a843f3a0, mem=0xb8e500)
    at ../../../../../opal/mca/memory/linux/malloc.c:4377
#3  0x00007f16a81e2317 in opal_memory_ptmalloc2_free (mem=0xb8e500) at ../../../../../opal/mca/memory/linux/malloc.c:3513
#4  0x00007f16a81e1051 in opal_memory_linux_free_hook (__ptr=0xb8e500, caller=0x7f16a8132687)
    at ../../../../../opal/mca/memory/linux/hooks.c:708
#5  0x00007f16a8132687 in opal_list_destruct (list=0xb8d250) at ../../opal/class/opal_list.c:108
#6  0x00007f16a4e6eeeb in opal_obj_run_destructors (object=0xb8d250) at ../../../../../opal/class/opal_object.h:460
[...]

PR #1066, opened 2015-10-23
Release the allocator list items from the list before destructing

rhc54:


jsquyres:
@rhc54 Does this fix #1065?

jsquyres:
Ah, you said at the bottom of #1065 that this PR is the fix for it.  :-)

Pro tip (if you care): if you put "Fixes open-mpi/ompi#1065" in the commit message, it'll auto-close that issue when this PR is merged.  :smile:

bosilca:
By looking at the proposed patch I have the impression that it does not fix the issue, but instead it hide it. The objects from the seg_list are never released (unlike the subject of this patch suggest), instead they are simply removed from the list (allowing the list destructor to succeed as the list is now empty). Thus, this looks like a potential memory leak to me.

rhc54:
You may well be correct - I wasn't entirely sure why releasing the items was causing a problem, but this seemed to fix it. We hit a similar issue in the oshmem area, but that was clearly a case where the list objects held some memory that was getting double-free'd. Someone should take a look at the underlying objects to understand why releasing them in that location is an issue.

PR #1068, opened 2015-10-24
Make odsl debug message consistent.

marksantcroos:
Ran into this while debugging #1067.


jsquyres:
@rhc54 Looks trivial to accept to me...

Issue #1060, opened 2015-10-22
2fd176ac causes segvs in MTL OFI on usNIC machines


jsquyres:
Git bisect showed me that 2fd176ac7f99b554da1706a0bfe7d93c648c8804 is the culprit for causing segv's on usNIC clusters.

Somehow the MTL OFI is segv'ing in libfabric 1.1.1 in the verbs provider -- ibv_get_device_list() is returning some devices that point to invalid memory.

I'm digging into it (I'm turning off my MTT tonight, because all runs will just segv and fill up TB worth of space with corefiles...).

Ping:L @hjelmn @yburette

yburette:
@jsquyres Which providers are enabled in your build of libfabric?
Is it segv'ing inside fi_getinfo() or after?

jsquyres:
Providers: verbs, tcp, usnic.

I could disable verbs, of course, but that would just mask the real problem.  :-)

Here's the bt of the segv:

```
(gdb) bt
#0  0x00000037a9a07170 in ibv_get_device_list () from /usr/lib64/libibverbs.so.1
#1  0x00002aaaaee1d3ab in fi_ibv_have_device () at prov/verbs/src/fi_verbs.c:850
#2  0x00002aaaaee1d87f in fi_ibv_init_info () at prov/verbs/src/fi_verbs.c:951
#3  0x00002aaaaee1e0aa in fi_ibv_getinfo (version=65536, node=0x0, service=0x0, flags=0, hin\
ts=0x1c3b9b0, info=0x7fffffffbdd0) at prov/verbs/src/fi_verbs.c:1160
#4  0x00002aaaaedf031c in fi_getinfo_ (version=65536, node=0x0, service=0x0, flags=0, hints=\
0x1c3b9b0, info=0x7fffffffbf28) at src/fabric.c:489
#5  0x00002aaaaf93f003 in ompi_mtl_ofi_component_init (enable_progress_threads=false, enable\
_mpi_threads=false) at mtl_ofi_component.c:256
#6  0x00002aaaaab9dd66 in ompi_mtl_base_select (enable_progress_threads=false, enable_mpi_th\
reads=false, priority=0x7fffffffc05c) at base/mtl_base_frame.c:78
#7  0x00002aaaaf525780 in mca_pml_cm_component_init (priority=0x7fffffffc05c, enable_progres\
s_threads=false, enable_mpi_threads=false) at pml_cm_component.c:143
[...]

PR #1070, opened 2015-10-24
common_usnic: move fake IBV provider to libopen-pal

jsquyres:
Move the fake usnic IBV provider out of common/verbs and into a new
common/verbs_usnic component that is always statically linked into
libopen-pal.  The fake provider is registered with libibverbs at run
time, but there is no *un*register IBV API.  Hence, we can't let the
code containing this provider be dlclosed -- which means it needs to
be statically linked into libopen-pal.

Fixes open-mpi/ompi#1060.

### DIFF
--- a/opal/mca/common/verbs/Makefile.am
+++ b/opal/mca/common/verbs/Makefile.am
@@ -17,7 +17,6 @@ headers = \
 sources = \
 	common_verbs_basics.c \
 	common_verbs_devlist.c \
-	common_verbs_fake.c \
 	common_verbs_find_max_inline.c \
 	common_verbs_find_ports.c \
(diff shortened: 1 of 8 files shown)


yosefe:
I get a symbol error, seems like it's related to this PR:
```
mpirun: symbol lookup error: <...>/install/lib/libmca_common_verbs.so.0: undefined symbol: opal_common_verbs_usnic_register_fake_drivers
```

yosefe:
sorry, my bad, build problem.

jsquyres:
@yosefe Did you guys perchance disable (or just not enable) the verbs_usnic component?

I.e., should there be protection in the code base against that case?

PR #1073, opened 2015-10-27
Cleanup the PMIx direct modex support.

rhc54:
Add an MCA parameter pmix_base_async_modex that will cause the async modex to be used when set to 1. Default it to 0 for now to continue current default behavior.

Also add an MCA param pmix_base_collect_data to direct that the blocking fence shall return all data to each process. Obviously, this param has no effect if async_modex is used.


lanl-ompi:
Test FAILed.

opal/mca/pmix/base/pmix_base_frame.c
```diff
@@ -31,12 +31,21 @@
...
 bool opal_pmix_base_allow_delayed_server = false;
 int opal_pmix_verbose_output = -1;
+bool opal_pmix_base_async_modex = false;
 
 static int opal_pmix_base_frame_register(mca_base_register_flag_t flags)
 {
+    opal_pmix_base_async_modex = false;
+    (void) mca_base_var_register("opal", "pmix", "base", "async_modex", "Use asynchronous modex mode",
+                                 MCA_BASE_VAR_TYPE_BOOL, NULL, 0, 0, OPAL_INFO_LVL_9,
+                                 MCA_BASE_VAR_SCOPE_READONLY, &opal_pmix_base_async_modex);
```
artpol84: @rhc54  Why the scope is "read-only"?

PR #1078, opened 2015-10-28
Disable the use of osc rdma when we detect a GPU buffer

### LINKED ISSUE #1012: CUDA-aware one sided put does not work on master

rolfv:
GPU buffers are not supported in the osc rdma component so force it to use pt2pt in those cases.

This forces a failover to the osc pt2pt component. Fixes #1012

Note that this is not needed in 2.x as the osc rdma has not made it over yet.


hjelmn:
Failure unrelated to this PR.

:bot:retest

PR #1079, opened 2015-10-29
Make CUDA 4.1 a requirement for CUDA-aware support

rolfv:
To clean up the code a bit, make it a requirement to have CUDA 4.1 or later. This allows us to also remove all related preprocessor conditionals. No one is using CUDA 4.0 or earlier anymore.


jsquyres:
:+1:

PR #1080, opened 2015-10-29
oob_tcp: fix peer->state wrong check

federeghe:
Hi! I think I have found a small bug in process_ping function, see the commit.

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp.c
+++ b/orte/mca/oob/tcp/oob_tcp.c
@@ -361,7 +361,7 @@ static void process_ping(int fd, short args, void *cbdata)
     }
 
     /* if we are already connecting, there is nothing to do */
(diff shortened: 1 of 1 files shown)


jsquyres:
@bosilca The TCP BTL is your baby -- can you have a look?

jsquyres:
@bosilca Apparently, I can't read.  Disregard my prior comment.  :-)

@rhc54 The OOB TCP is your baby -- can you have a look?

rhc54:
@federeghe I don't think the reported failure has anything to do with this change as the code path being modified isn't one that is used in OMPI today. It only gets executed if we were to "ping" another peer via OOB, which we don't do.

@miked-mellanox Can you confirm what happened with the test? Can we get a line number where it failed?

rhc54:
bot:retest

PR #1082, opened 2015-10-30
Fix a few more places that utilized CUDA 4.1 checks

rolfv:
I missed a few spots in previous commit for removing CUDA 4.1 checks.  This fixes it.

Will bring into 2.x as well.


rolfv:
I am just going to merge this because without it, all my MTT tests fail. I will have Jeff review the 2.x pull request.

PR #1090, opened 2015-11-03
mtl-portals4: test for valid handle before releasing resources

tkordenbrock:
During component finalize, mtl-portals4 would blindly release
resources without testing if the handle was valid.  This was OK,
but resource allocation is now delayed until add_procs().  If
mtl-portals4 is deselected, it will be finalized without
add_procs() ever being called.  This commit ensures that invalid
handles are not released.

@regrant - please review


regrant:
:+1:

PR #1091, opened 2015-11-03
Cleanup the temporary memory allocation in collectives

bosilca:
Cleanup the memory handling for temporary buffers in some of the collective modules. Added a new function opal_datatype_span, to compute the memory span of count number of datatype, excluding the gaps in the beginning and at the end. If a memory allocation is made using the returned value, the gap (also returned) should be removed from the allocated pointer.

This is a second attempt of fixing this issue (follow-up on #1047)

### DIFF
--- a/ompi/mca/coll/base/README.memory_management
+++ b/ompi/mca/coll/base/README.memory_management
@@ -1,0 +1,124 @@
+    /* This comment applies to all collectives (including the basic
+     * module) where we allocate a temporary buffer.  For the next few
+     * lines of code, it's tremendously complicated how we decided that
+     * this was the Right Thing to do.  Sit back and enjoy.  And prepare
+     * to have your mind warped. :-)
+     *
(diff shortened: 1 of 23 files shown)


ggouaillardet:
@bosilca 
one malloc is missing, patch is inline
```patch
index 17ac951..2ee07d0 100644
--- a/ompi/mca/coll/basic/coll_basic_scan.c
+++ b/ompi/mca/coll/basic/coll_basic_scan.c
@@ -75,6 +75,7 @@ mca_coll_basic_scan_intra(const void *sbuf, void *rbuf, int count,
          * receive into, later. */
 
         dsize = opal_datatype_span(&dtype->super, count, &gap);
+        free_buffer = malloc(dsize);
         if (NULL == free_buffer) {
             return OMPI_ERR_OUT_OF_RESOURCE;
         }
```

jsquyres:
@bosilca Looks like a merge went wrong on this PR.  Can you fix?

PR #1095, opened 2015-11-04
Some trivial fixes

kawashima-fj:
Four trivial fixes.
Probably only c09f9f0 will affect typical user programs.
I'll create a new PR for v2.x if this PR is accepted.


bosilca:
Nice, we survived for years with those typos in. Thanks for catching them up. :+1:

PR #1101, opened 2015-11-04
btl/openib: fix access flags

hjelmn:
Per spec for ibv_reg_mr if remote write or remote atomic is requested also
need to specify local write.


hjelmn:
Also headed for 2.0.0.

PR #1102, opened 2015-11-06
Ensure that we completely register an nspace prior to launching local…

rhc54:
… procs as otherwise we may attempt to send it down before it is registered, leading to data corruption

### DIFF
--- a/orte/orted/pmix/pmix_server_register_fns.c
+++ b/orte/orted/pmix/pmix_server_register_fns.c
@@ -42,15 +42,22 @@
 
 #include "orte/util/name_fns.h"
 #include "orte/runtime/orte_globals.h"
+#include "orte/runtime/orte_wait.h"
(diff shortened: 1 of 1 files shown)


marksantcroos:
Thanks Ralph, one more question though:
Does this imply that for every task there is XML parsing happening? That sounds non-optimal.

rhc54:
Yes and no. What actually happens is:

* the daemon reads its local topology at startup and stores it as that XML string
* each process is given a copy of the string during its startup and parses it to create its own local topology object

The notion is that we don't want every process actually probing the local node to obtain the topology info, so this speeds things up. However, we should delete the string after the object has been created, and we need a more memory-efficient means of passing it to the local server.

The memory problem we hit here is due to the server never releasing the data for a job, and so every job just keeps increasing the memory footprint. I'm working on resolving that one now.

marksantcroos:
Thanks for the explanation.

PR #1107, opened 2015-11-06
Work on cleaning up memory leaks that are causing orte-dvm to eventua…

rhc54:
…lly run out of memory. Still don't have everything plugged, but getting better. Sync to the PMIx master that includes removal of the pmix_common.h.in file that really didn't need to be generated, and update to the PMIx_server_init API.

@marksantcroos You might want to give this a try - won't entirely solve the problem, but it looks like it slows the growth rate down.


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

rhc54:
bot:retest

marksantcroos:
I get this now at orte-dvm startup, is that related?

```
$ orte-dvm 
[comet-ln3.sdsc.edu:12579] [[59063,0],0] ORTE_ERROR_LOG: Not found in file ../../../../../orte/mca/ess/hnp/ess_hnp_module.c at line 632
Open MPI developer):

  opal_pmix_base_select failed
```

rhc54:
bizarre - I've been using orte-dvm all weekend without a problem. Let me take a look.

marksantcroos:
FWIW: I just noticed that I only get this on the SLURM system, not on the Cray.

rhc54:
@marksantcroos I just tested this, and everything works fine under slurm. Are you perhaps pointing to the incorrect library location once you get the allocation?

marksantcroos:
Hmmm, now I become hesitant to verify this pending the embarrassment ... ;-)

marksantcroos:
Luckily it was not that stupid. But I rebuild from scratch and now it works ...
Sorry for the noise, maybe I forgot to autogen.pl or so.

Btw, on that note, is it best to always autogen, rm -rf build, ./configure, make? Or can some shortcuts be taken?

rhc54:
You only need to run autogen if a new component has been added, a component has been removed, or someone "touched" a .m4 file. Otherwise, you shouldn't need to rerun it.

Likewise for configure - if you run autogen, then you need to run configure. Otherwise, you only need to run it if you are changing the configuration.

marksantcroos:
> You only need to run autogen if a new component has been added, a component has been removed, or someone "touched" a .m4 file.

Which happened when you synced PMIx. I guess that explains it.

marksantcroos:
> You might want to give this a try - won't entirely solve the problem, but it looks like it slows the growth rate down.

To speak to the original issue, this is indeed the case. Still grows, but by far not as fast.

rhc54:
Kewl - I'll be able to address the remaining leaks after SC

hppritcha:
@jsquyres I'd like to merge the commits in this PR over to v2.x after PR open-mpi/ompi-release#869.
I've been noticing differences in histories for some of the pmix related files between master and v2.x.

For example

on v2.x:
```
git log --oneline -- orte/orted/pmix/pmix_server_register_fns.c
b8fe6094 Remove OPAL_HAVE_HWLOC qualifier and error out if --without-hwloc is given
0a4e56e3 Ensure that we completely register an nspace prior to launching local procs as otherwise we may attempt to send it down before it is registered, leading to data corruption
78876d8c Integrate PMIx 1.0 with OMPI.
```

while on master:
```
5f446570 Work on cleaning up memory leaks that are causing orte-dvm to eventually run out of memory. Still don't have everything plugged, but getting better. Sync to the PMIx master that includes removal of the pmix_common.h.in file that really didn't need to be generated, and update to the PMIx_server_init API.
ec0cc4bf Ensure that we completely register an nspace prior to launching local procs as otherwise we may attempt to send it down before it is registered, leading to data corruption
dc5796b8 Revert "Revert "Fix the handling of cpusets so we get the correct cpuset for each local peer. Add the ability to indicate that a modex request is "optional" so we don't call the server if we don't find the value. Take advantage of that to allow the MPI layer to decide that the lack of locality info indicates non-local""
6a8fad49 Revert "Fix the handling of cpusets so we get the correct cpuset for each local peer. Add the ability to indicate that a modex request is "optional" so we don't call the server if we don't find the value. Take advantage of that to allow the MPI layer to decide that the lack of locality info indicates non-local"
f94f3cda Fix the handling of cpusets so we get the correct cpuset for each local peer. Add the ability to indicate that a modex request is "optional" so we don't call the server if we don't find the value. Take advantage of that to allow the MPI layer to decide that the lack of locality info indicates non-local
e6add86e Deal with connect/accept between two jobs from different mpirun's. Somewhat optimize connect/accept by using MPI bcast to distribute the participants instead of another PMIx lookup. Cleanup some Coverity issues.
d97bc291 Remove OPAL_HAVE_HWLOC qualifier and error out if --without-hwloc is given
38ba5436 Fix shared memory operations by resolving local peers
0d5814b5 Cleanup Coverity issues
cf6137b5 Integrate PMIx 1.0 with OMPI.
```
if one tries to apply commits like 0d5814b5 one gets conflicts.

jsquyres:
@hppritcha :+1:

rhc54:
:+1:

PR #1109, opened 2015-11-07
Silence reentrant libevent warning except on debug builds

rhc54:


rhc54:
@artpol84 @hppritcha @hjelmn Can one or more of you check to see if this has the desired effect? I have no way of testing it.

opal/mca/event/libevent2022/libevent/event.c
```diff
@@ -1566,8 +1566,11 @@ event_base_loop(struct event_base *base, int flags)
 	EVBASE_ACQUIRE_LOCK(base, th_base_lock);
 
 	if (base->running_loop) {
+/*****   OMPI change   ****/
+#ifdef OPAL_DEBUG
```
jsquyres: @rhc54 How about using `#if OPAL_ENABLE_DEBUG`?

artpol84:
@rhc54 @jsquyres @jladd-mlnx This indeed fixes the problem of "warn" flooding in our installation.

jsquyres:
Please see https://github.com/rhc54/ompi/pull/5.

artpol84:
@rhc54  @jsquyres I'm still seeing this issue in the latest master.

artpol84:
@rhc54 @jsquyres 
It seems that since I'm using master `OPAL_ENABLE_DEBUG=1` is forced by the "Early development override"
https://github.com/open-mpi/ompi/blob/master/config/opal_configure_options.m4#L105

rhc54:
I just realized that the libevent portion of this PR never got merged into the master for some weird reason. I'm fixing that now.

@artpol84 The behavior you are seeing is intentional - the reentrant error still exists and needs to be addressed, so we want that warning to show on developer builds. The intent here is strictly to avoid confusing users with it as (a) there is nothing they can do about it, and (b) the problem eventually works itself out after repeated cycling thru opal_progress (at least, it does that at the moment - we need to fix the problem in case timing changes and it no longer finds a way out of the reentrant loop).

artpol84:
@rhc54 I see, but this error drives my MTT tests crazy. I have to go there and comment it out each time. Not a big deal though.

jsquyres:
@artpol84 Should get fixed by #1168.

rhc54:
@jsquyres Actually, it probably won't if @artpol84 is testing against a git clone instead of a tarball. So what we probably need to do is add some smarts to the pmix configure.m4 to indicate that if --disable-debug is given, then we need to pass that down to the pmix embedded code.

PR #1115, opened 2015-11-09
opal_free_list: fix strange size check

hjelmn:
OPAL free lists can be initialized with a fragment size that differs
from the size of objects from a class. This allows the free list code
to support OPAL objects that have flexible array members.

Unfortunately the free list code will throw out the desired length in
some cases. The code in question was committed in
open-mpi/ompi@90fb58de. The side effects of this are varied and can
cause segmentation faults, assert failures, hangs, etc. This commit
adds a check to ensure the requested size is at least as large as the
class size and makes opal_free_list allocations always honor the
requested fragment size (as long as it is larger than the class
size).

### DIFF
--- a/opal/class/opal_free_list.c
+++ b/opal/class/opal_free_list.c
@@ -121,6 +121,10 @@ int opal_free_list_init (opal_free_list_t *flist, size_t frag_size, size_t frag_
             return OPAL_ERROR;
     }
 
+    if (frag_class && frag_size < frag_class->cls_sizeof) {
+        frag_size = frag_class->cls_sizeof;
+    }
+
     if (frag_size > flist->fl_frag_size) {
         flist->fl_frag_size = frag_size;
     }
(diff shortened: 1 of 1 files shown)


hjelmn:
@rolfv This fixes an issue you saw awhile ago that was fixed by modifying the mca_pml_ob1_send_range_t structure to not use a flexible array member. I was finally able to reproduce then I found the bad check.

hjelmn:
This bug has existed in Open MPI since 2007 but it may not be triggered in versions prior to the 2.x branch.

hjelmn:
Bogus jenkins failure. I want to make some changes to the commit before retesting.

rolfv:
I cannot remember what issue you are referring to. Do you remember more details? But I am glad whatever it was it is fixed.

hjelmn:
I can't remember. I will look back through the archives and see if I can find the exact issue.

PR #1121, opened 2015-11-10
change -0bind-to and -bind-to to --bind-to in the manpages

dmt4:


rhc54:
@dmt4 Thanks for fixing my fat fingers! :+1:

dmt4:
I was reading the manpage and it struct me as odd to have -0bind. Funny enough, I wrote 'bint' twice in the commit message before catching it on second reading :)

PR #1130, opened 2015-11-12
add a verification step looking for the structures that we use in the…

edgargabriel:
… lustre component. Disable the ccomponent if not found.

@rhc54 this should fix the issue that you reported on the mailing list, could you please give it a shot on odin? It should at this point simply disable the lustre component.


rhc54:
Checked it on odin - problem solved!

PR #1131, opened 2015-11-13
osc/rdma: fix some threading bugs

hjelmn:
There were two bugs in osc/rdma when using threads:

 - Deadlock is ompi_osc_rdma_start_atomic. This occurs because
   ompi_osc_rdma_frag_alloc is called with the module lock. To fix the
   issue the module lock is now recursive. In the future I will add a
   new lock to protect just the current rdma fragment.

 - Do not drop the lock in ompi_osc_rdma_frag_alloc when calling
   ompi_osc_rdma_frag_complete. Not only is it not needed but dropping
   the lock at this point can cause a competing thread to mess up the
   state.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -998,7 +998,7 @@ static int ompi_osc_rdma_component_select (struct ompi_win_t *win, void **base, 
     }
 
     /* initialize the objects, so that always free in cleanup */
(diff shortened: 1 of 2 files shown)


hjelmn:
@regrant This should fix the problems you found running your multithreaded benchmarks. I can now run them without issue.

hjelmn:
There is still an issue with put aggregation. I will fix that later. For now set -mca osc_rdma_aggregation_limit 0.

hjelmn:
Mellanox Jenkins fail! Ignoring.

PR #1126, opened 2015-11-11
Topic/ikrit err fix

alex-mikheev:
@miked-mellanox @yosefe  please take a look

PR #1134, opened 2015-11-15
SPML/IKRIT: opal_progress and ud_only fixes

alex-mikheev:
Fixes two problems found during testing of https://github.com/open-mpi/ompi/pull/1126

Some MXM tls such as self, shm can comlete requests immediately.
Make sure that opal_progress() is called before before request
is completed.

fix ud_only logic when hw rdma channel is using ud and main
transport is rc or dc.

https://github.com/open-mpi/ompi/pull/1126

@miked-mellanox @yosefe


miked-mellanox:
:+1:

PR #1143, opened 2015-11-21
Fix "autogen.pl --no-ompi"

rhc54:
This was broken due to inclusion of a conditional (MCA_BUILD_ompi_pml_monitoring_DSO) in the test/monitoring Makefile.am that is only defined if OMPI is built.

@jsquyres @bosilca I suspect this isn't the right way to fix the build, but I've tried everything I can think of to resolve it with no luck. This seems to work, but looks ugly. Could you guys please look at this?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1379,8 +1379,9 @@ AC_CONFIG_FILES([
     test/support/Makefile
     test/threads/Makefile
     test/util/Makefile
-    test/monitoring/Makefile
 ])
+m4_ifdef([project_ompi], [AC_CONFIG_FILES([test/monitoring/Makefile])])
+
 AC_CONFIG_FILES([contrib/dist/mofed/debian/rules],
                 [chmod +x contrib/dist/mofed/debian/rules])
 AC_CONFIG_FILES([contrib/dist/mofed/compile_debian_mlnx_example],
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -11,6 +11,8 @@
 #                         All rights reserved.
 # Copyright (c) 2009      Sun Microsystems, Inc. All rights reserved.
 # Copyright (c) 2010      Cisco Systems, Inc. All rights reserved.
+# Copyright (c) 2015      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -19,5 +21,8 @@
 #
 
 # support needs to be first for dependencies
-SUBDIRS = support asm class threads datatype util monitoring
+SUBDIRS = support asm class threads datatype util
+if PROJECT_OMPI
+SUBDIRS += monitoring
+endif
 DIST_SUBDIRS = event $(SUBDIRS)


ggouaillardet:
my 0.02US$

all test/*/Makefile.am does define check_PROGRAMS  except monitoring that defines noinst_PROGRAMS
check_PROGRAMS is never empty, even when there is no ompi
and I am not sure everything would go fine if that would be empty.

my best bet would be to update test/Makefile.am and append monitoring to SUBDIRS only if PROJECT_OMPI

I will not be able to test that before tuesday...

configure.ac
```diff
@@ -256,7 +256,9 @@ m4_ifdef([project_oshmem], [OSHMEM_CONFIGURE_OPTIONS])
 # Set up project specific AM_CONDITIONALs
 AS_IF([test "$enable_ompi" != "no"], [project_ompi_amc=true], [project_ompi_amc=false])
 m4_ifndef([project_ompi], [project_ompi_amc=false])
-AM_CONDITIONAL([PROJECT_OMPI], [test "$project_ompi_amc" = "true"])
+AM_CONDITIONAL([PROJECT_OMPI], [test "$PROJECT_OMPI_amc" = "true"])
```
jsquyres: This change in case doesn't seem right -- the lower case version is defined on the three lines above.

jsquyres:
A better solution might be to move the test/monitoring directory to be within ompi/mca/pml/monitoring.  That way, it'll already naturally be traversed (or not), depending on whether the monitoring PML is going to be built or not.

configure.ac
rhc54: Argh - that is a typo. I didn't mean to change that line.

jsquyres:
@rhc54 @bosilca Thoughts on moving the `test/monitoring` directory to be in `ompi/mca/pml/monitoring`?  That would make the whole problem much easier to solve.

bosilca:
We always put all our testing together in a single place. Scattering the tests all over the development tree is not only ugly but confusing.

jsquyres:
It also makes the build conditionals a nightmare.

FWIW, the usnic BTL has had its tests in `opal/mca/btl/usnic` for quite a while.  Similar with the debugger tests (in `ompi/debuggers`).

bosilca:
These are exceptions and not the norm. Moreover, the monitoring PML only depends on HWLOC which is now a required block in OMPI, and as a result the monitoring PML is supposedly always compiled. What is the problem with these tests?

jsquyres:
The monitoring PML *can* be compiled out via `--enable-mca-no-build` (which I believe someone did).

bosilca:
Let me check more precisely wht the problem is. The interaction between the PML an the rest of the world is done via MPIT pvars and cvars, and the test should have failed graciously.

jsquyres:
I just closed up my laptop, but IIRC, it's a linking problem.

Sent from my phone. No type good.

On Dec 6, 2015, at 2:10 PM, bosilca <notifications@github.com<mailto:notifications@github.com>> wrote:


Let me check more precisely wht the problem is. The interaction between the PML an the rest of the world is done via MPIT pvars and cvars, and the test should have failed graciously.

-

rhc54:
It isn't a linking problem - the issue is that MCA_BUILD_ompi_pml_monitoring_DSO is only defined if you are (a) building the OMPI layer and (b) didn't exclude that component. So autogen fails because that parameter hasn't been defined and yet is used in the test Makefile.am.

bosilca:
How do we protect the MPI tests today ?

rhc54:
The problem is simple:

```
if PROJECT_OMPI
    noinst_PROGRAMS = monitoring_test
    monitoring_test_SOURCES = monitoring_test.c
    monitoring_test_LDFLAGS = $(WRAPPER_EXTRA_LDFLAGS)
    monitoring_test_LDADD = $(top_builddir)/ompi/libmpi.la $(top_builddir)/opal/libopen-pal.la

if MCA_BUILD_ompi_pml_monitoring_DSO
    lib_LTLIBRARIES = monitoring_prof.la
    monitoring_prof_la_SOURCES = monitoring_prof.c
    monitoring_prof_la_LDFLAGS=-module -avoid-version -shared $(WRAPPER_EXTRA_LDFLAGS)
    monitoring_prof_la_LIBADD = $(top_builddir)/ompi/libmpi.la $(top_builddir)/opal/libopen-pal.la
endif

endif

```
That second if statement is -not- protected by the first one - autogen still requires that the param being tested must be defined, even if the first if is not "true"

ggouaillardet:
fwiw an other way to fix this is to update ```test/Makefile.am``` and append ```monitoring``` to ```SUBDIRS``` only if ```PROJECT_OMPI```

that might be a bit convoluted ...

imho, i agree with @bosilca we should keep the tests in more or less the same location
what we could do is have several directories in test
```
test/opal
test/orte
test/ompi
```
the pros is that would fix such kind of issues in an elegant way
the cons is we would have to split the ```datatype``` directory (for example) into ```opal/datatype``` and ```ompi/datatype```

rhc54:
@ggouaillardet I tried your suggestion, and it does not work - autogen still insists that the parameter *must* be defined:

```
opal/Makefile.am: installing 'config/depcomp'
test/monitoring/Makefile.am:23: error: MCA_BUILD_ompi_pml_monitoring_DSO does not appear in AM_CONDITIONAL
autoreconf: /opt/local/bin/automake failed with exit status: 1
```

bosilca:
Let's comment out the generation of the shared library. This library is a PMPI type of interception layer that can be linked with any applications in order to get automatic support for the monitoring framework. It is not necessary for the testing.

ggouaillardet:
@rhc54 oops, my bad
here is the full (and tested) patch
```patch
diff --git a/configure.ac b/configure.ac
index eec43fc..3e23a3c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1377,8 +1377,9 @@ AC_CONFIG_FILES([
     test/support/Makefile
     test/threads/Makefile
     test/util/Makefile
-    test/monitoring/Makefile
 ])
+m4_ifdef([project_ompi], [AC_CONFIG_FILES([test/monitoring/Makefile])])
+
 AC_CONFIG_FILES([contrib/dist/mofed/debian/rules],
                 [chmod +x contrib/dist/mofed/debian/rules])
 AC_CONFIG_FILES([contrib/dist/mofed/compile_debian_mlnx_example],
diff --git a/test/Makefile.am b/test/Makefile.am
index 5252fd5..5d5323d 100644
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -11,6 +11,8 @@
 #                         All rights reserved.
 # Copyright (c) 2009      Sun Microsystems, Inc. All rights reserved.
 # Copyright (c) 2010      Cisco Systems, Inc. All rights reserved.
+# Copyright (c) 2015      Research Organization for Information Science
 # $COPYRIGHT$
 #
@@ -19,5 +21,8 @@
 #
 
 # support needs to be first for dependencies
-SUBDIRS = support asm class threads datatype util monitoring
+SUBDIRS = support asm class threads datatype util
+if PROJECT_OMPI
+SUBDIRS += monitoring
+endif
 DIST_SUBDIRS = event $(SUBDIRS)
```

rhc54:
@ggouaillardet Looks fine to me, and it seems to work, so let's go with that approach. Thx!

jsquyres:
@rhc54 @ggouaillardet @bosilca No no no!  This does not solve the problem of conditional compilation.

If I `configure --enable-mca-no-build=pml-monitoring`, you will still have a linking problem.

The tests for the monitoring PML need to only be built when the monitoring PML itself is built.  The best way to to do that is to put the tests in the same directory as the monitoring PML itself, so that if the monitoring PML is not built, its directory is skipped and the tests aren't built, either.

The whole GNU `make check` architecture is designed to allow tests to be anywhere in the tree.  It's common and useful to locate tests with the code that they are testing, not randomly somewhere else in the tree.  Yes, we have tests in `tests`, but that's somewhat of a historical artifact -- it's not necessarily the only way that we should do things.

PR #1157, opened 2015-11-28
Adds darwin headers directory for Darwin JDK

gpichot:
On OS X 10.11.1 (El Capitan), configuring with --enable-mpi-java and correct values for --with-jdk-headers and -with-jdk-bindir arguments still fails due to some headers under the darwin include directory ($with_jdk_headers/darwin). I added the following already available for Linux and Solaris SDKs.

I think that the include directory could also be detected automatically using : /usr/libexec/java_home outputs on OS X.

### DIFF
--- a/config/opal_setup_java.m4
+++ b/config/opal_setup_java.m4
@@ -186,6 +186,11 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
                           # too.  Ugh.
                           AS_IF([test -d "$with_jdk_headers/solaris"],
                                 [OPAL_JDK_CPPFLAGS="$OPAL_JDK_CPPFLAGS -I$with_jdk_headers/solaris"])
+                          # Darwin JDK also require -I<blah>/darwin.
+                          # See if that's there, and if so, add a -I for that,
+                          # too.  Ugh.
+                          AS_IF([test -d "$with_jdk_headers/darwin"],
+                                [OPAL_JDK_CPPFLAGS="$OPAL_JDK_CPPFLAGS -I$with_jdk_headers/darwin"])
 
                           CPPFLAGS="$CPPFLAGS $OPAL_JDK_CPPFLAGS"])
                    AC_CHECK_HEADER([jni.h], [],
--- a/config/opal_setup_java.m4
+++ b/config/opal_setup_java.m4
@@ -95,7 +95,9 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
               [ # OS X Snow Leopard and Lion (10.6 and 10.7 -- did not
                 # check prior versions)
                opal_java_found=0
-               opal_java_dir=/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers
+               AS_IF([test -x /usr/libexec/java_home],
+                    [opal_java_dir=`/usr/libexec/java_home`/include],
+                    [opal_java_dir=/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers])
                AC_MSG_CHECKING([OSX locations])
                AS_IF([test -d $opal_java_dir],
                      [AC_MSG_RESULT([found ($opal_java_dir)])


hppritcha:
@gpichot is this intended for helping with non-default java installs on os-x?

gpichot:
I thinks it is for default java installs on OS X : I neither installed my Java JDK nor updated it, I only updated my Macintosh to 10.11.1 prior to downloading, configuring and compiling ompi.

I could check with some of my schoolmates, but in my case, configuring without it lead to a failure due to jni_md.h absent from the include directories (but this header file is actually present under the darwin directory).

jsquyres:
@gpichot Odd.  When I do a default build on OS 10.11.1 (i.e., without `--with-java-headers=DIR`), it picks up the Java headers in `/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers` -- including `jni_md.h`.  Does it not do that for you?

That being said, your `/usr/libexec/java_home` command is cool -- I didn't know about that (I don't know much about Java on OS X).  But it gives a different location for Java than `/System/Library/Frameworks/JavaVM.framework/Versions/Current`, so I don't know how these two Javas are related (i.e., are they actually the same under the covers, or are they actually different installations / versions)?

```sh
$ uname -a
Darwin JSQUYRES-M-H05C 15.0.0 Darwin Kernel Version 15.0.0: Sat Sep 19 15:53:46 PDT 2015; root:xnu-3247.10.11~1/RELEASE_X86_64 x86_64
$ /usr/libexec/java_home
/Library/Java/JavaVirtualMachines/jdk1.7.0_45.jdk/Contents/Home
```

gpichot:
Nope...

     $ ./configure --enable-mpi-java
     ....
     *** Java MPI bindings
     checking if want Java bindings... yes
     configure: WARNING: Java bindings requested but no Java support found
     configure: error: cannot continue

config.log (I am compiling against 1.10.1 release in this case) :

     configure:29430: checking OSX locations
     configure:29439: result: not found
     configure:29451: checking Linux locations
     configure:29629: result: not found
     configure:29638: checking Solaris locations
     configure:29647: result: not found
     configure:29900: checking Java support available
     configure:29903: result: no
     configure:29954: checking if want Java bindings
     configure:29964: result: yes
     configure:29966: WARNING: Java bindings requested but no Java support found
     configure:29968: error: cannot continue

I have no Headers directory under /System/Library/Frameworks/JavaVM.framework/Versions/Current. (I never installed Java by myself). Current points to A in my case :

    $ ls -l /System/Library/Frameworks/JavaVM.framework/Versions/A 
    drwxr-xr-x  47 root  wheel    1598 Sep 17 09:06 Commands
    drwxr-xr-x   4 root  wheel     136 Sep 17 07:30 Frameworks
    drwxr-xr-x   3 root  wheel     102 Aug 23 03:59 JavaPluginCocoa.bundle
    -rwxr-xr-x   1 root  wheel  109488 Sep 17 09:08 JavaVM
    drwxr-xr-x  45 root  wheel    1530 Sep 17 08:57 Resources
    drwxr-xr-x   3 root  wheel     102 Aug 23 01:24 _CodeSignature

My setup :

     $ uname -a
     Darwin gabriel.local 15.0.0 Darwin Kernel Version 15.0.0: Sat Sep 19 15:53:46 PDT 2015; root:xnu-3247.10.11~1/RELEASE_X86_64 x86_64
     $ /usr/libexec/java_home
     /Library/Java/JavaVirtualMachines/jdk1.7.0_79.jdk/Contents/Home

But `/usr/libexec/java_home` does have an include directory :

    -rw-rw-r--   1 root  wheel      3339 Apr 10  2015 COPYRIGHT
    -rw-rw-r--   1 root  wheel        40 Apr 10  2015 LICENSE
    -rw-rw-r--   1 root  wheel       114 Apr 10  2015 README.html
    -rw-rw-r--   1 root  wheel    110114 Mar 16  2015 THIRDPARTYLICENSEREADME-JAVAFX.txt
    -rw-rw-r--   1 root  wheel    173571 Apr 10  2015 THIRDPARTYLICENSEREADME.txt
    drwxrwxr-x  44 root  wheel      1496 Apr 10  2015 bin
    drwxrwxr-x   9 root  wheel       306 Apr 10  2015 db
    drwxrwxr-x   9 root  wheel       306 Apr 10  2015 include
    drwxrwxr-x  10 root  wheel       340 Apr 10  2015 jre
    drwxrwxr-x  14 root  wheel       476 Apr 10  2015 lib
    drwxrwxr-x   5 root  wheel       170 Apr 10  2015 man
    -rw-rw-r--   1 root  wheel       502 Apr 10  2015 release
    -rw-rw-r--   1 root  wheel  20084963 Apr 10  2015 src.zip

If I have some time this week-end I will add a commit that uses `/usr/libexec/java_home` for java location on OS X.

hppritcha:
From the ```java_home``` man page, looks like this showed up with xcode 5.  I like using a mac but I''m not an expert on mac os-x releases, etc.  Would it be problematic to rely on ```java_home``` being present?  I don't like the hardcoded default directory ```/System/Library/...``` in opal_setup_java.m4 and it would be great to replace with ```java_home```, but will the dependence on xcode 5 or higher be a problem for older os-x releases?

jsquyres:
Ah, I see that you have a slightly newer Java than me.  I confess that I don't remember the history of java installs / updates that I have run on my laptop...

I think it would be fine to rely on `/usr/libexec/java_home` (i.e., test if it's there, and then use it if it is).

gpichot:
@jsquyres : I followed your piece of advise on this. I tried on my computer and it works like a charm (no need for extra arguments).

@hppritcha : I do not like the hardcoded default directory too, but I left it as a fallback case (for prior XCode 5 setups).

jsquyres:
@hppritcha I'm cool with this PR.  How about you?

hppritcha:
yes I'm merging.

PR #1158, opened 2015-11-30
coll-portals4: remove unneeded code from gather

tkordenbrock:
This commit removes two pieces of unneeded code from gather.  First
it removes destroy_tree() calls from linear_top(), because the
linear algorithm does not create a tree, so there is no need to
destroy it.  Second it removes unpack_bytes from the gather request
because it was calculated but never used.

@regrant - please review


regrant:
:+1:

PR #1161, opened 2015-11-30
coll-portals4: add scatter and iscatter implementations that use Portals4 triggered operations

tkordenbrock:
This commit adds implementations of scatter and iscatter using
Portals4 triggered operations.  Currently, the only algorithm
is linear.

@regrant - please review


regrant:
:+1: looks good and ready for merge.

PR #1144, opened 2015-11-22
Update PMIx support to the latest production release

rhc54:
Update the PMIx native component to release v1.1.1, with addition of one bug-fix commit beyond the official release

 Rename the pmix1xx component to pmix111 so it reflects the actual release it includes

PR #1163, opened 2015-11-30
ompi/proc: make proc system always thread safe

hjelmn:
This commit changes the OPAL_THREAD_LOCK/OPAL_THREAD_UNLOCK calls in
ompi/proc to opal_mutex_lock/opal_mutex_unlock. This will allow
multi-threaded BTLs the ability to creat ompi_proc_t's without having
to set opal_using_threads. There should be no performance hits as none
of the lock points are in the critical path.

### DIFF
--- a/ompi/proc/proc.c
+++ b/ompi/proc/proc.c
@@ -84,10 +84,10 @@ void ompi_proc_destruct(ompi_proc_t* proc)
     if (NULL != proc->super.proc_hostname) {
         free(proc->super.proc_hostname);
     }
-    OPAL_THREAD_LOCK(&ompi_proc_lock);
+    opal_mutex_lock (&ompi_proc_lock);
     opal_list_remove_item(&ompi_proc_list, (opal_list_item_t*)proc);
     opal_hash_table_remove_value_ptr (&ompi_proc_hash, &proc->super.proc_name, sizeof (proc->super.proc_name));
-    OPAL_THREAD_UNLOCK(&ompi_proc_lock);
+    opal_mutex_unlock (&ompi_proc_lock);
 }
 
 /**
@@ -209,7 +209,7 @@ opal_proc_t *ompi_proc_for_name (const opal_process_name_t proc_name)
         return &proc->super;
     }
 
-    OPAL_THREAD_LOCK(&ompi_proc_lock);
+    opal_mutex_lock (&ompi_proc_lock);
     do {
         /* double-check that another competing thread has not added this proc */
         ret = opal_hash_table_get_value_ptr (&ompi_proc_hash, &proc_name, sizeof (proc_name), (void **) &proc);
@@ -231,7 +231,7 @@ opal_proc_t *ompi_proc_for_name (const opal_process_name_t proc_name)
             break;
         }
     } while (0);
-    OPAL_THREAD_UNLOCK(&ompi_proc_lock);
+    opal_mutex_unlock (&ompi_proc_lock);
 
     return (opal_proc_t *) proc;
 }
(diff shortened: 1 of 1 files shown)


hjelmn:
@ggouaillardet @rhc54 This should address the issue we were seeing with dynamic add_procs and the tcp btl. Let me know if this does indeed fix the issue. Target is 2.0.0.

hjelmn:
@bosilca Let me know if you see any problems with this PR. I don't think it will affect performance but I may be missing something.

bosilca:
I am not sure what issue this patch is solving, but we didn't notice any problem with the multi-threaded TCP BTL. (As a side note the TCP BTL only created ompi_proc_t upon creation of the first TCP endpoint, an operation that is already protected).

Do you mind pointing to the issue solved by this patch?

rhc54:
@hjelmn FWIW: this didn't solve the problem - I'm still seeing a rash of errors when dynamic add_procs is enabled:

```
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_create_info_half 
   Elapsed:       00:00:01 0.01u 0.00s
*** WARNING: Test: c_create_info_half, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_create_no_free 
   Elapsed:       00:00:01 0.01u 0.00s
*** WARNING: Test: c_create_no_free, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_fence_lock 
   Elapsed:       00:00:01 0.00u 0.00s
*** WARNING: Test: c_fence_lock, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_win_attr 
   Elapsed:       00:00:00 0.00u 0.00s
*** WARNING: Test: c_win_attr, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_win_errhandler 
   Elapsed:       00:00:01 0.01u 0.00s
*** WARNING: Test: c_win_errhandler, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_fence_simple 
   Elapsed:       00:00:01 0.01u 0.00s
*** WARNING: Test: c_fence_simple, np=16, variant=1: FAILED
>> Command: mpirun     --hostfile /home/common/hosts -np 16 --prefix /home/common/openmpi/build/foobar/ onesided/c_fence_asserts 
   Elapsed:       00:00:00 0.00u 0.00s
```

hjelmn:
@bosilca It fixes a failure seen in PR #1144. See the reproducer posted by @ggouaillardet.

@rhc54 Those failure may be unrelated to the error fixed by this PR. Do you have any output from one of the failed one-sided tests? I may need to make additional changes to fix those.

hjelmn:
@bosilca Looks like there is more to #1144 but I still think this would be a useful change. With this change BTLs can call opal_proc_for_name for any thread without needing to set opal_using_threads.

bosilca:
My point was that this patch is too early. As long as we look at a single BTL it needs internal locks around the creation of the endpoint (thus proc), so this particular patch will not be needed. However, once we have multiple async BTL, the proposed patch is needed. :+1:

hjelmn:
@bosilca Good point. This probably does not solve an existing problem but it does solve a potential problem. I will merge once the Mellanox Jenkins returns.

hjelmn:
Hmm, mellanox jenkins aborted.

:bot:retest:

hjelmn:
:bot:retest:

hjelmn:
Going to merge this in today and PR for v2.x.

artpol84:
I haven't noticed any regression with this patchset (with full modex
however).
Direct modex has lots of problems, but I can't say right now what are the
reasons.
I'm +1 to merge it.


вторник, 8 декабря 2015 г. пользователь Nathan Hjelm написал:

> Going to merge this in today and PR for v2.x.
>
> —
> <https://github.com/open-mpi/ompi/pull/1163#issuecomment-162905501>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

PR #1173, opened 2015-12-02
ompi/group: #if protect dead code when ompi is not configure'd with -…

ggouaillardet:
…-enable-sparse-groups

### DIFF
--- a/config/ompi_configure_options.m4
+++ b/config/ompi_configure_options.m4
@@ -76,6 +76,7 @@ else
 fi
 AC_DEFINE_UNQUOTED([OMPI_GROUP_SPARSE],$GROUP_SPARSE,
     [Wether we want sparse process groups])
+AM_CONDITIONAL(OMPI_GROUP_SPARSE, test "$GROUP_SPARSE" = "1")
 
 
 #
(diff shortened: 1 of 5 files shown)


ggouaillardet:
This is a proof-of-concept PR that #if out dead code when ompi is not configure'd with --enable-sparse-groups
(e.g. ```OMPI_SPARSE_GROUPS``` macro is ```0```)
note the first two ```#if OMPI_SPARSE_GROUPS``` in ```ompi/group/group.h``` break internal binary compatilbility, so we might not want to keep them when/if this PR gets merged

ggouaillardet:
@hjelmn @jsquyres could you please have a look at this and comment ?

jsquyres:
Adding @edgargabriel

hppritcha:
Is there a downside to enabling sparse groups by default?  Seems to me it will suffer from bit rot if it continues to remain an opt in.

bosilca:
Things are working pretty well the way it is now, adding extra #ifdef just makes the code unreadable.

jsquyres:
I'm actually ok leave the code *not* `#if`'ed out -- it ensures that the code still compiles (at a bare minimum).

hppritcha:
consensus seems to be to close and not accept this PR.

Issue #1174, opened 2015-12-02
MPICH large_type_sendrec hangs with TCP BTL


artpol84:
@bosilca @rhc54 @jsquyres 

I see 100% reproducable hangs of the `large_type_sendrec` test. As far as I can tell, in this test only first and last ranks are actually communicating, all others are just waiting in `MPI_Finalize()`.
This test has an agrument that says the power of 2 for the sen/recv buffer size. I see the following picture:

| argument | time |
| --- | --- |
| 30 | 13 s |
| 31 | 27 s |
| 32 | **hang** |

If I use openib or vader BTL's it works fine for all mentioned buffer sizes.

Backtraces shows the following (I provide only the first and the last ranks):
**rank=0**

``` bash
(gdb) bt
#0  0x00000037feaaccdd in nanosleep () from /lib64/libc.so.6
#1  0x00000037feae1e54 in usleep () from /lib64/libc.so.6
#2  0x00007f4d2700889c in OPAL_PMIX_PMIX1XX_PMIx_Fence (procs=0x0, nprocs=0, info=0x0, ninfo=0) at src/client/pmix_client_fence.c:100
#3  0x00007f4d27238f3c in pmix1_fence (procs=0x0, collect_data=0) at pmix1_client.c:306
#4  0x00007f4d28f1622d in ompi_mpi_finalize () at runtime/ompi_mpi_finalize.c:242
#5  0x00007f4d28f40b2d in PMPI_Finalize () at pfinalize.c:47
#6  0x0000000000401764 in main (argc=2, argv=0x7fff66de9f18) at large_type_sendrec.c:182
```

**rank=16 (last)**

```
(gdb) bt
#0  0x0000003c000cf527 in sched_yield () from /lib64/libc.so.6
#1  0x00007f3adbc60b79 in opal_progress () at runtime/opal_progress.c:199
#2  0x00007f3adc2c7508 in opal_condition_wait (c=0x7f3adc5f3f00, m=0x7f3adc5f3e80) at ../opal/threads/condition.h:76
#3  0x00007f3adc2c75ae in ompi_request_wait_completion (req=0x18d5a00) at ../ompi/request/request.h:383
#4  0x00007f3adc2c75ff in ompi_request_default_wait (req_ptr=0x7fff068dbc68, status=0x7fff068dbc48) at request/req_wait.c:42
[...]

PR #1177, opened 2015-12-03
Topic/large msg

bosilca:
MPICH large_type_sendrec hangs with TCP BTL (#1174)

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -190,7 +190,7 @@ static void mca_pml_ob1_put_completion (mca_pml_ob1_rdma_frag_t *frag, int64_t r
     mca_pml_ob1_recv_request_t* recvreq = (mca_pml_ob1_recv_request_t *) frag->rdma_req;
     mca_bml_base_btl_t *bml_btl = frag->rdma_bml;
 
-    OPAL_THREAD_SUB_SIZE_T(&recvreq->req_pipeline_depth, 1);
+    OPAL_THREAD_ADD32(&recvreq->req_pipeline_depth, -1);
 
     MCA_PML_OB1_RDMA_FRAG_RETURN(frag);
 
@@ -198,7 +198,7 @@ static void mca_pml_ob1_put_completion (mca_pml_ob1_rdma_frag_t *frag, int64_t r
         assert ((uint64_t) rdma_size == frag->rdma_length);
 
         /* check completion status */
-        OPAL_THREAD_ADD_SIZE_T(&recvreq->req_bytes_received, (size_t) rdma_size);
+        OPAL_THREAD_ADD_SIZE_T(&recvreq->req_bytes_received, rdma_size);
         if (recv_request_pml_complete_check(recvreq) == false &&
             recvreq->req_rdma_offset < recvreq->req_send_offset) {
             /* schedule additional rdma operations */
@@ -951,7 +951,7 @@ int mca_pml_ob1_recv_request_schedule_once( mca_pml_ob1_recv_request_t* recvreq,
     }
 
     while(bytes_remaining > 0 &&
-           recvreq->req_pipeline_depth < mca_pml_ob1.recv_pipeline_depth) {
+          recvreq->req_pipeline_depth < mca_pml_ob1.recv_pipeline_depth) {
         mca_pml_ob1_rdma_frag_t *frag = NULL;
         mca_btl_base_module_t *btl;
         int rc, rdma_idx;
@@ -1028,7 +1028,7 @@ int mca_pml_ob1_recv_request_schedule_once( mca_pml_ob1_recv_request_t* recvreq,
         if (OPAL_LIKELY(OMPI_SUCCESS == rc)) {
             /* update request state */
             recvreq->req_rdma_offset += size;
-            OPAL_THREAD_ADD_SIZE_T(&recvreq->req_pipeline_depth, 1);
+            OPAL_THREAD_ADD32(&recvreq->req_pipeline_depth, 1);
             recvreq->req_rdma[rdma_idx].length -= size;
             bytes_remaining -= size;
         } else {
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -983,14 +983,10 @@ int mca_pml_ob1_recv_request_schedule_once( mca_pml_ob1_recv_request_t* recvreq,
         } while(!size);
         btl = bml_btl->btl;
 
-        /* NTH: This conditional used to check if there was a registration in
-         * recvreq->req_rdma[rdma_idx].btl_reg. If once existed it was due to
-         * the btl not needed registration (equivalent to btl->btl_register_mem
-         * != NULL. This new check is equivalent. Note: I feel this protocol
-         * needs work to better improve resource usage when running with a
-         * leave pinned protocol. */
-        if (btl->btl_register_mem && (btl->btl_rdma_pipeline_frag_size != 0) &&
-            (size > btl->btl_rdma_pipeline_frag_size)) {
+         /* NTH: Note: I feel this protocol needs work to better improve resource
+          * usage when running with a leave pinned protocol. */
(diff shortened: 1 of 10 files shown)


lanl-ompi:
Test FAILed.

jsquyres:
bot:retest

ggouaillardet:
now it works for me with ```--mca btl top,self``` but partially because the message is split over 2 or 3 btls
If I explicitly uses one interface ```--mca btl_tcp_if_include eth0``` it hangs.
The reason is writev with a 4GB message size always returns 0, regardless I use lo, eth0 or ib0
This is a rhel6 kernel (I suspect we hit an internal limit that might be different on an other kernel)

bosilca:
We either need a way to detect what are the limits supported by writev/readv or we need to force on all OSes the fragments to be less than 2^31 bytes.

jsquyres:
@bosilca Are you talking about a configure test to figure out the member type of `struct iovec.iov_len`?  I.e., are you concerned about older OS's that don't use a `size_t`?

bosilca:
No, I am concerned about the OS support with regard to the total amount of data that can be sent with a single writev operation. We already have discovered 2 cases, where the outcome is not exactly what we expected (@ggouaillardet RedHat version and OS X).

jsquyres:
@bosilca Ah.  Is there a `configure`-test way to figure this out, perchance?  I dislike hard-coding for "in OS ABC, set the limit to X, in OS DEF, set the limit to Y, ...etc."

bosilca:
Not that I know of. That was what my previous comment was about, either we come up with a test, or we prevent the TCP BTL of handling more that the currently known minimum (2^31-1 bytes) per writev.

jsquyres:
Can we open /dev/null and try to write 2^31+1 bytes?

bosilca:
The following code works on OS X. However, I confirmed before that writev with a sum of iov_len larger than 2^32 return EINVAL.
```
#include <stdio.h>
#include <sys/uio.h>
#include <unistd.h>
#include <fcntl.h>
#include <string.h>
#include <stdlib.h>
#include <errno.h>
#include <math.h>

int main( int argc, char* argv[] )
{
    int fd, err, iovcnt;
    ssize_t rc;
    char filename[] = "/dev/null";
    struct iovec iov[2];

    fd = open(filename, O_WRONLY);
    if( fd < 0 ) {
        printf("Could not open file %s\n", filename);
        exit(-1);
    }

    iovcnt = 2;

    iov[0].iov_len = strlen(filename);
    iov[0].iov_base = filename;

    iov[1].iov_len = (size_t)pow(2, 32);
    iov[1].iov_base = malloc(iov[1].iov_len * sizeof(char));

    rc = writev(fd, iov, iovcnt);
    if( rc < 0 ) {
        err = errno;
        switch(err) {
          case EINVAL:
              printf("Got error EINVAL %s\n", strerror(err));
              break;
          default:
              printf("Got error %d %s\n", err, strerror(err));
        }
    }
    close(fd);
    return 0;
}
```

jsquyres:
Per discussion on the 26 Jan 2016:

* OS X and BSD man pages state that the type is `int32_t`, not `size_t` (i.e., it's not an OS X-specific problem)
* This PR changes our internal type from `size_t` to `int32_t`, but it doesn't check for overflow / loop around `writev(2)` to segment writing the full message

ibm-ompi:
Test FAILED
 
ibm-ompi:

bosilca:
@ggouaillardet, @jsquyres and @hjelmn please review and push. This fixed the long pending issue with large data transfers over TCP. It also fixes an issue that disabled the RDMA protocol for BTL without a registration function.

bosilca:
@ggouaillardet, @hjelmn and @bwbarrett please comment or merge.

bwbarrett:
@jsquyres, mind updating your review (you wanted signed-off-by, George added them)?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
```diff
@@ -983,14 +983,10 @@ int mca_pml_ob1_recv_request_schedule_once( mca_pml_ob1_recv_request_t* recvreq,
         } while(!size);
         btl = bml_btl->btl;
 
-        /* NTH: This conditional used to check if there was a registration in
-         * recvreq->req_rdma[rdma_idx].btl_reg. If once existed it was due to
-         * the btl not needed registration (equivalent to btl->btl_register_mem
-         * != NULL. This new check is equivalent. Note: I feel this protocol
-         * needs work to better improve resource usage when running with a
-         * leave pinned protocol. */
-        if (btl->btl_register_mem && (btl->btl_rdma_pipeline_frag_size != 0) &&
```
bwbarrett: Does anyone know why there used to be a check about register mem here?  I honestly can't figure it out.  @bosilca / @hjelmn?

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -318,7 +318,9 @@ static int mca_btl_tcp_component_register(void)
     mca_btl_tcp_module.super.btl_rndv_eager_limit = 64*1024;
     mca_btl_tcp_module.super.btl_max_send_size = 128*1024;
     mca_btl_tcp_module.super.btl_rdma_pipeline_send_length = 128*1024;
-    mca_btl_tcp_module.super.btl_rdma_pipeline_frag_size = INT_MAX;
+    /* Some OSes have hard coded limits on how many bytes can be manipulated by each writev operation.
+     * Force a reasonable limit, to prevent overflowing a 32-bit integer (limit comes from BSD and OS X) */
+    mca_btl_tcp_module.super.btl_rdma_pipeline_frag_size = ((1UL<<31) - 1024);
```
bwbarrett: This really needs to be a #define and also, I'm not sure I can figure out exactly why the - 1024 from the comment; a bit more context as to why that instead of INT_MAX would be super useful.

ompi/mca/pml/ob1/pml_ob1_sendreq.h
```diff
@@ -45,11 +45,11 @@ struct mca_pml_ob1_send_request_t {
...
     opal_ptr_t req_recv;
-    int32_t req_state;
-    int32_t req_lock;
-    bool req_throttle_sends;
-    size_t req_pipeline_depth;
-    size_t req_bytes_delivered;
+    int32_t  req_state;
+    int32_t  req_lock;
+    bool     req_throttle_sends;
+    int32_t  req_pipeline_depth;
```
bwbarrett: Why is req_pipeline depth different than the other variables that require ADD_SIZE_T atomics?  It seems like either we have a more general problem (which I could believe) or we don't have a problem that needs fixing in this commit...

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: I assumed that someone wanted to prevent pipelining if we have support for RMA. For the problem at hand, if we don't enforce this limit we will need to alter the internals of the btl_tcp_send to cope with very large buffers.

opal/mca/btl/tcp/btl_tcp_component.c
bosilca: (1UL<<31) was found by empirically testing on different systems and then taking the lowest value (OS X). The 1024 was to have some slack before the strict limit imposed by some OSes. This number can certainly be lowered to the largest PML header, but in this particular instance I don't think such accuracy is necessary.

ompi/mca/pml/ob1/pml_ob1_sendreq.h
bosilca: There is no need to have it as a size_t as it counts the number of descriptors inflight (4 by default).

opal/mca/btl/tcp/btl_tcp_component.c
bwbarrett: Can you add that to the comment?  Otherwise, the whole series looks good to me.

bosilca: Comment has been updated to reflect the randomness of choice.

bosilca:
A long-term solution would be to check if the OS support large writes and adapt the code accordingly. A similar issue is discussed in the context of MPI IO in #2399.

bwbarrett:
@bosilca, were you going to merge this version into master?

bosilca:
Yes, this provides a partial fix until we have the configure check.

PR #1179, opened 2015-12-03
Pr/mpi testsome man page update

jsquyres:
Instead of solely relying on the out value definitions in MPI_Waitsome.3, explicitly copy this text here.

Note that the original text in this man page was copied verbatim from the MPI spec; we've now added a bit more text (copied from MPI_Waitsome.3in) that explains the out values so that users don't have to cross-reference to another man page.

Thanks to @eschnett for the suggestion.

Fixes open-mpi/ompi#1153


jsquyres:
@bosilca Got any opinion on this?  It fixes @eschnett's suggestion in #1153.

bosilca:
In general I'm not a fan of repetitions (despite "Repetition is the mother of learning"). Other than this :+1:

PR #623, opened 2015-06-03
Add missing Java and Fortran bindings

ggouaillardet:

PR #1180, opened 2015-12-04
fortran: add missing MPI_xxx_DUP_FN bindings

ggouaillardet:
- MPI_COMM_DUP_FN
- MPI_TYPE_DUP_FN
- MPI_WIN_DUP_FN

### DIFF
--- a/ompi/mpi/fortran/base/attr-fn-int-callback-interfaces.h
+++ b/ompi/mpi/fortran/base/attr-fn-int-callback-interfaces.h
@@ -4,6 +4,8 @@
 ! Copyright (c) 2006-2014 Cisco Systems, Inc.  All rights reserved.
 ! Copyright (c) 2013      Los Alamos National Security, LLC. All rights
 !                         reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@jsquyres can you have a look at this ?
f08 bindings were added in #623 but nothing for mpifh nor usempi.
i cannot remember whether we had a good reason not to add them...

jsquyres:
Are the declarations in `mpif-h/prototypes_mpi.h` and the definitions in `base/attr_fn_f.c` insufficient?

ggouaillardet:
@jsquyres it is enough for ```mpif-h``` but not for ```use-mpi-tkr``` since this ```mpif-h/prototypes_mpi.h``` is not used when building ```mpi.mod```
i amended the commit message to make this crystal clear

jsquyres:
Ok.  Looks like you're right.

Can you also add some MTT tests to make sure we test for DUP_FN in the mpi and mpi_f08 modules?

PR #1184, opened 2015-12-07
Fix intercommunicator split (was triggered by MPICH/icsend test)

artpol84:
@hjelmn 
Nathan, it seems that you've introduced remote_group in `ompi_comm_split()` in a88b24ce. Could you review this fix?
With current master I see failures with `icsend` MPICH test. This test does `MPI_Comm_split` of inter-communicator.
In this case `remote_group` is set to `&ompi_mpi_group_null.group`:
https://github.com/open-mpi/ompi/blob/master/ompi/communicator/comm.c#L511
This leads this condition to trigger since `NULL != &ompi_mpi_group_null.group`:
https://github.com/open-mpi/ompi/blob/master/ompi/communicator/comm.c#L151:L153
and this drops `remote_size` that has useful value and was passed through the function arguments.

I may be missing your point in a88b24ce but it seems that some changes introduced there either:
(a) was later removed; or
(b) seems consistent with this change.

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -180,7 +180,7 @@ int ompi_comm_set_nb ( ompi_communicator_t **ncomm,
     if (0 < remote_size) {
         ompi_communicator_t *old_localcomm;
 
-        if (NULL == remote_group || &ompi_mpi_group_null.group == remote_group) {
+        if (NULL == remote_group) {
             ret = ompi_group_incl(oldcomm->c_remote_group, remote_size,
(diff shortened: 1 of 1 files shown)


hjelmn:
The remote_size was dropped because it provided duplicate information. Looks like something was missed when I changed that. Are we missing a split of intercommunicators in our MTT tests?

Anyway, I think this looks good. It will need to be PR'd to 2.x once it is in master.

artpol84:
@hjelmn Ok.
I was unable to find v2.0? It is not in ompi-release and there is no such branch here. Where can I find it?

PR #1186, opened 2015-12-07
opal/shmem: improve help message

### LINKED ISSUE #1141: corrected/expanded message about shmem temporary directory
This came up on the mail list.  I've noticed it before, and should have reported it.
opal_tmp_directory looks for TMPDIR, as it should on Unix-y systems, not TEMPDIR.
It seems worth mentioning the MCA parameter too.

```diff
diff --git a/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt b/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt
index 67fa8f9..21ce04e 100644
--- a/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt
+++ b/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt
@@ -31,7 +31,8 @@ Open MPI to be much slower than expected.
 
 You may want to check what the typical temporary directory is on your
 node.  Possible sources of the location of this temporary directory
-include the $TEMPDIR, $TEMP, and $TMP environment variables.
+include the $TMPDIR (most likely on Unix), $TEMP, and $TMP environment
+variables.  See also the MCA parameter "orte_tmpdir_base".
 
 Note, too, that system administrators can set a list of filesystems
 where Open MPI is disallowed from creating temporary files by setting
```

hppritcha:
Per suggestions from @loveshack, improve verbiage
in the help message for opal/shmem in the event
problems are hit setting up shared memory segments
for various Open MPI components.

Fixes #1141

@ggouaillardet 

### DIFF
--- a/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt
+++ b/opal/mca/shmem/mmap/help-opal-shmem-mmap.txt
@@ -31,7 +31,8 @@ Open MPI to be much slower than expected.
 
 You may want to check what the typical temporary directory is on your
 node.  Possible sources of the location of this temporary directory
-include the $TEMPDIR, $TEMP, and $TMP environment variables.
+include the $TMPDIR (most likely on Unix), $TEMP, and $TMP environment
+variables.  See also the MCA parameter "orte_tmpdir_base".
 
 Note, too, that system administrators can set a list of filesystems
 where Open MPI is disallowed from creating temporary files by setting


ggouaillardet:
@hppritcha @loveshack
on second thought ...

from ```opal/mca/shmem/mmap/shmem_mmap_component.c```

```c
    opal_shmem_mmap_backing_file_base_dir = "/dev/shm";
    ret = mca_base_component_var_register (&mca_shmem_mmap_component.super.base_version,
                                           "backing_file_base_dir",
                                           "Specifies where backing files will be created when "
                                           "shmem_mmap_relocate_backing_file is in use.", MCA_BASE_VAR_TYPE_STRING,
                                           NULL, 0, MCA_BASE_VAR_FLAG_SETTABLE, OPAL_INFO_LVL_9,
                                           MCA_BASE_VAR_SCOPE_ALL_EQ,
                                           &opal_shmem_mmap_backing_file_base_dir);
```
so basically, this can only go wrong if /dev/shm is a network filesystem (quite unlikely imho) or if ```shmem_mmap_backing_file_base_dir``` was explicitly set to use a network filesystem (bad configuration).

could we revamp the test to something like that
- if ```shmem_mmap_backing_file_base_dir``` has its default value, tell the user this is really strange /dev/shm is NFS mounted, and he/she should explicitly set ```shmem_mmap_backing_file_base_dir``` to point to a non network filesystem
- if ```shmem_mmap_backing_file_base_dir``` was explicitly set, tell the user to use a different value that is not a network filesystem

makes sense ?

loveshack:
Gilles Gouaillardet <notifications@github.com> writes:

> @hppritcha @loveshack
> on second thought ...
>
> from ```opal/mca/shmem/mmap/shmem_mmap_component.c```
>
> ```c
>     opal_shmem_mmap_backing_file_base_dir = "/dev/shm";
>     ret = mca_base_component_var_register (&mca_shmem_mmap_component.super.base_version,
>                                            "backing_file_base_dir",
>                                            "Specifies where backing files will be created when "
>                                            "shmem_mmap_relocate_backing_file is in use.", MCA_BASE_VAR_TYPE_STRING,
>                                            NULL, 0, MCA_BASE_VAR_FLAG_SETTABLE, OPAL_INFO_LVL_9,
>                                            MCA_BASE_VAR_SCOPE_ALL_EQ,
>                                            &opal_shmem_mmap_backing_file_base_dir);
> ```
> so basically, this can only go wrong if /dev/shm is a network filesystem (quite unlikely imho) or if ```shmem_mmap_backing_file_base_dir``` was explicitly set to use a network filesystem (bad configuration).
>
> could we revamp the test to something like that
> - if ```shmem_mmap_backing_file_base_dir``` has its default value, tell the user this is really strange /dev/shm is NFS mounted, and he/she should explicitly set ```shmem_mmap_backing_file_base_dir``` to point to a non network filesystem
> - if ```shmem_mmap_backing_file_base_dir``` was explicitly set, tell the user to use a different value that is not a network filesystem
>
> makes sense ?

No, I don't understand what this has to do with the issue, which was
just to get the environment variable right.

For what it's worth, looking at the code, it seems to be using a
Linux-specific device file without any Linux conditional, which doesn't
look right.  /dev/shm isn't on a Solaris 10 system, for instance.

[I hadn't actually realized /dev/shm was the default.  From experience
I'd say it's not a good one.  Apart from using up the memory, if I
recall correctly, one of the problems is that the memory isn't accounted
properly to the processes using it, which can mess up resource
allocation in a batch system.]

hppritcha:
I think @loveshack just wanted to improve/fix the wording of the help-me text without making it
overly complex for an end user.  I'm pretty sure that details about why ```/dev/shm``` is messed up,
etc. would confuse many end users.  End users do often understand what environment variables
are though.

ggouaillardet:
@hppritcha there are several things here ...
- i have no objection with this change
- using ```/dev/shm``` has pros and cons
  - we could automatically fallback to ```/tmp``` if ```/dev/shm``` does not exist
  - as you pointed, some RM might not correctly account used memory
  - on top of that, these files might not be automatically removed when a job completes
  - using ```/dev/shm``` is more likely to work out of the box (think of a diskless system with ```/tmp``` being a network filesystem and not a ```tmpfs```) than ```/tmp```
- as far as i am concerned, i think the error message is not helpful for an end user.
  we could explicitly direct him/her to set ```shmem_mmap_backing_file_base_dir``` and hint him/her to use an environment variable (```$TMPDIR``` or something else)

Issue #1185, opened 2015-12-07
MPICH p2p failures


artpol84:
I see MTT failures for the following tests:
* **bsend2**
* **bsendalign**
* **bsend5**
* **cancelrecv** (addressed in Issue https://github.com/open-mpi/ompi/issues/1175)
* **icsend** (addressed in PR https://github.com/open-mpi/ompi/pull/1184)

The rest of the tests seems to have same problem: **hang**. It can be observed with **pml/ob1**, **btl/[tcp,openib]**, if I use **pml/yalla** (mxm) I see no problems with those tests. All problems can be reproduced even on one node, but you must make sure that you don't include any sm BTLs:  `-mca btl self,[tcp,openib]`. In all cases rank0 that receives the data remains blocked in `MPI_Recv()` while all senders reach `MPI_Finalize()` and *think* that the data was sent correctly.
Here is some additional information on each of them:
* **bsend2** and **bsendalign** are similar from this problems point of view problem, they can be reproduced with just 2 processes.
* **bsend5** needs 4 processes to hang.

hjelmn:
Weird. I will take a look and see why buffered sends are not working correctly.

hjelmn:
Can't get either bsend2 or bsendalign to hang on my mac. bsend5 does hang however.

hjelmn:
Scratch that. I can reproduce on a different mac with bsend2. Looking at it now.

hjelmn:
Ok, was a problem with my firewall. All those tests pass with -mca btl self,tcp

hjelmn:
Hah, well. I was running with a single change. Can you try with this patch?

```
diff --git a/ompi/mca/pml/ob1/pml_ob1_isend.c b/ompi/mca/pml/ob1/pml_ob1_isend.c
index dba3c22..893e6ce 100644
--- a/ompi/mca/pml/ob1/pml_ob1_isend.c
+++ b/ompi/mca/pml/ob1/pml_ob1_isend.c
@@ -198,7 +198,7 @@ int mca_pml_ob1_send(const void *buf,
             return rc;
         }
 
-        /* free the request and return. don't care if it completes now */
+        ompi_request_wait_completion (brequest);
         ompi_request_free (&brequest);
         return OMPI_SUCCESS;
     }
```

artpol84:
@hjelmn With this patch all those tests works well. Thank you.

hjelmn:
[...]

PR #1187, opened 2015-12-08
pml/ob1: add missing ompi_request_wait_completion for buffered sends

hjelmn:
This commit adds a call to ompi_request_wait_completion for buffered
sends. Without this line it is possible to get into a state where the
data is never sent.

Fixes open-mpi/ompi#1185

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_isend.c
+++ b/ompi/mca/pml/ob1/pml_ob1_isend.c
@@ -198,7 +198,7 @@ int mca_pml_ob1_send(const void *buf,
             return rc;
         }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
Not sure how MTT missed this. Fix will be needed for 1.10.2 and 2.x.

artpol84:
Per https://github.com/open-mpi/ompi/issues/1185. It was tested and fixes issue problem, I think it can be merged if it'll pass standard checks.

hjelmn:
Yup. Will let Jenkins double-check to be safe. Please review the 1.10 and 2.x PRs when they are up. Will be pushed once this is merged.

PR #1188, opened 2015-12-08
Yet one more fix to intercommunicator splitting logic.

artpol84:
@hjelmn :))) I finally got your original idea, Nathan. I've checked this against my icsend and icsplit tests manually and will fire up MTT. Will post the results later, PR'ing now in case you'll have the time to check this today.

Previous commit f2794740 reverts Nathans changes. However it turns out
that I was unable to trace his logic until I started investigation of
icsplit hang. Bug was triggered when splitting Intercom was giving a group
where on side of the communicator was empty (icsplit, intercom create #2).
in this case remote_size == 0 and there is no way to distinguish between
inter- and intra-communicator.
Conclusion: We do need to distinguish between intra- and inter-communicators.
So we should use ompi_mpi_group_null.group.

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -148,7 +148,7 @@ int ompi_comm_set_nb ( ompi_communicator_t **ncomm,
         local_size = ompi_group_size (local_group);
     }
 
-    if (NULL != remote_group) {
+    if ( (NULL != remote_group) && (&ompi_mpi_group_null.group != remote_group) ) {
         remote_size = ompi_group_size (remote_group);
     }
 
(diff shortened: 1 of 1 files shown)


artpol84:
P.S. This fix also fixes **icm**, **probe-intercomm**, **comm_group_rand** tests.
And for *iccreate** test I see 32 errors instead of 222.

artpol84:
P.P.S and I was also observing lots of errors in MPICH/coll tests related to intercomm. They seems to be fixed too. At least **icreduce** works now. I will check further and post here.

artpol84:
According to MTT **icsplit** and **icsend** are working fine now. All mpich/pt2pt tests are passed. I see 2 MPICH/comm tests failing: **icm** and **probe-intercomm** and I see 13 of MPICH/coll tests that failing.

hjelmn:
Ah, yes. I see what was happening. Thanks for tracking this down. Don't forget to PR the patches to 2.x.

artpol84:
Sure!
By the way do we want to retain patch sequence or I can merge them in one
patch?

вторник, 8 декабря 2015 г. пользователь Nathan Hjelm написал:

> Ah, yes. I see what was happening. Thanks for tracking this down. Don't
> forget to PR the patches to 2.x.
>
> —
> <https://github.com/open-mpi/ompi/pull/1188#issuecomment-162891262>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

hjelmn:
Best to retain the patch sequence so we keep a 1-1 mapping of master to release commits. Update the commit messages to include the master commit hash. Take a look at at open PR on ompi-release to see how we usually handle that.

PR #1190, opened 2015-12-08
osc/sm: Fix a bug that `MPI_WIN_TEST` does not update `flag` to 0

kawashima-fj:
`MPI_WIN_TEST` must update the `flag` parameter to 0 when not all
origin processes called `MPI_WIN_COMPLETE`. But sm OSC doesn't.
If the caller initialize the `flag` argument to a non-0 value,
the caller will receive the non-0 `flag` value.

@hjelmn please review

I'll create PRs for v1.10 and v2.0.

### DIFF
--- a/ompi/mca/osc/sm/osc_sm_active_target.c
+++ b/ompi/mca/osc/sm/osc_sm_active_target.c
@@ -315,6 +315,8 @@ ompi_osc_sm_test(struct ompi_win_t *win,
         OBJ_RELEASE(module->post_group);
         module->post_group = NULL;
         *flag = 1;
+    } else {
+        *flag = 0;
     }
 
     OPAL_THREAD_UNLOCK(&module->lock);


hjelmn:
:+1:

kawashima-fj:
@hjelmn Do you have a plan to merge 903762e19456d797bec6d05ff4a56e5f936d14bc to v1.10 and v2.x branches?
Automatic merge on cherry-picking my fix fails because the commit is not merged into v1.10 and v2.x branches.

The bug I found exists in v1.10 and v2.x in another form. Code which set `flag` to 0 exists in a wrong position. The `else` block shoud correspond to `if` block of `complete_count`, not `post_group`.

Current v1.10/v2.x code:

```
    if (NULL != module->post_group) {
        int size = ompi_group_size(module->post_group);

        if (module->my_node_state->complete_count == size) {
            OBJ_RELEASE(module->post_group);
            module->post_group = NULL;
            *flag = 1;
        }
    } else {
        opal_atomic_mb();
        *flag = 0;
    }
```

hjelmn:
Opening PRs now.

PR #1192, opened 2015-12-08
mlx5: need to set comp_mask to get experimental verbs attributes

hjelmn:


hjelmn:
@miked-mellanox Atomics were being disabled by the supported atomic size check. Discovered this missing line while trying to figure out why nvidia MTT is failing.

PR #1193, opened 2015-12-08
btl-portals4: set endpoint rank even if endpoint already exists

tkordenbrock:
If btl-portals4 is configured to use logical mapping of ranks to
physical nodes, then the endpoint must have the rank field set.
This commit fixes a bug that caused the endpoint to have the
nid/pid instead of the rank if the endpoint already exists.

@regrant - please review


hppritcha:
is this a blocker for v2.0.0?

tkordenbrock:
Yes, it's a blocker.  I just added the blocker label.

regrant:
:+1:

PR #1194, opened 2015-12-08
pmix/cray: fix locality bug

hppritcha:
There was a bug with the way the cray pmix component
was setting the locality property for ranks on the
same node, etc.

Improve location/syntax of a comment block.

@hjelmn 


hjelmn:
:+1:

PR #1195, opened 2015-12-08
support Cray nativized slurm environment

hppritcha:
This set of commits enables support for orte based job launches on Cray systems using nativized SLURM as the job launcher.

One of the commits toggles off plm/alps component when the cray wlm detection system reports SLURM is the current active job launcher.

The plm/slurm component has to be slightly modified in order to support ```--disable-dlopen``` configurations of Open MPI on Cray systems, as well as disabling selection of the 
Cray PMIx component when using an orte based job launch.

Note Cray nativized SLURM is only available on CLE 5 and higher systems.

@hjelmn 
@rhc54


hppritcha:
bot:retest

PR #1197, opened 2015-12-09
Fix ompi_comm_create when source communicator is inter-communicator.

artpol84:
@hjelmn Here is one more fix for inter-communicators. It fixes **probe-intercom** and **icm** tests from MPICH test suite.
I expect it to fix MPICH/coll intercom tests. Will check and report later today.


artpol84:
This fix indeed helps to fix MPICH/coll test timeout's. Now I have only 3 timeouting tests versus 13 yesterday and 33 two days ago.

PR #1199, opened 2015-12-10
btl/openib: add check for IBV_EXP_QP_INIT_ATTR_ATOMICS_ARG

hjelmn:
Mofed 2.2 does not have the IBV_EXP_QP_INIT_ATTR_ATOMICS_ARG attribute
flag. Add a check to fix compilation for mofed 2.2. This commit only
fixes complilation with the older mofed. It will not allow an Open MPI
compiled with mofed 2.3 or newer to work on a machine with mofed 2.2.


hjelmn:
@miked-mellanox We found the source of the nvidia MTT failures. They were building on a machine with mofed 2.3 and running on two machines (one with 2.2 and the other with 2.3). The 2.2 machine was failing because the IBV_EXP_QP_INIT_ATTR_ATOMICS_ARG flag is invalid. There is no way to protect against that failure but I am adding this PR to fix compilation with 2.2.

PR #1202, opened 2015-12-10
Fix MPI_Alltoall to support inter-communicators.

artpol84:
@bosilca @jsquyres @miked-mellanox @jladd-mlnx @hjelmn 
Hello, I'm seing hangs of the **icalltoall** MPICH/coll tests.

In **icalltoall** hang occured because in line 46:
http://git.mpich.org/mpich-mellanox.git/blob/master:/test/mpi/coll/icalltoall.c#l46
`Alltoall` is called with **zero** receive buffer. In ompi/master we immidiately exit:
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoall.c#L99
while in v1.10 we would continue:
https://github.com/open-mpi/ompi-release/blob/v1.10/ompi/mpi/c/alltoall.c#L101:L103

As you can see in https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoall.c#L99 we will only exit if receive size is 0, we don't have this option for the send buffer.
As the result left side of the inter-communicator immidiately exits `Alltoall` and continues while the right side waits for the collective to complete.

Other collectives are not so strict about that, i.e. **ialltoall** doesn't have this check at all:
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/ialltoall.c#L100
The same is for **alltoallv**:
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoallv.c#L123

the comment  before deleted piece of code references MPI standard:
```C
-    /* Do we need to do anything? Per MPI standard the (v3.1 page 168 line 48)
-     * the amount of data sent must be equal to the amount of data received.
-     */
-    ompi_datatype_type_size(recvtype, &recvtype_size);
-    if( (0 == recvcount) || (0 == recvtype_size) ) {
-        return MPI_SUCCESS;
-    }
```

Here is the snippet:
```
The type signature associated with sendcount, sendtype, at a process must be equal to the 
type signature associated with recvcount, recvtype at any other process. This implies that 
the amount of data sent must be equal to the amount of data received, pairwise between 
every pair of processes.
```

However I think that comment interprets this only for intra-communicators. In case of inter-communicators we don't know what did processes in other group specified as the sendbuffer/sendcount.
And there is already parameter check for the intra-communicator case in the code:
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoall.c#L85:L92

Other collectives seems to be consistent and pass MPICH/coll tests from this perspective.

### DIFF
--- a/ompi/mpi/c/alltoall.c
+++ b/ompi/mpi/c/alltoall.c
@@ -92,12 +92,10 @@ int MPI_Alltoall(const void *sendbuf, int sendcount, MPI_Datatype sendtype,
         }
     }
 
-    /* Do we need to do anything? Per MPI standard the (v3.1 page 168 line 48)
-     * the amount of data sent must be equal to the amount of data received.
-     */
-    ompi_datatype_type_size(recvtype, &recvtype_size);
-    if( (0 == recvcount) || (0 == recvtype_size) ) {
-        return MPI_SUCCESS;
+    if( !OMPI_COMM_IS_INTER(comm) ){
+        if( (0 == recvcount) || (0 == recvtype_size) ) {
+            return MPI_SUCCESS;
+        }
     }
 
     OPAL_CR_ENTER_LIBRARY();


artpol84:
In case of inter-communicator we have 2 group of processes and "left" group may call
MPI_Alltoall(NULL, 0, MPI_INT, buf, 10, MPI_INT, comm, ...);
and the right one:
MPI_Alltoall(buf,10,MPI_INT, NULL, 0, MPI_INT, comm, ...);

And it would be legal though one of the group will receive 0 bytes from others.

This was triggered by MPICH/coll test called icalltoall.

artpol84:
P.S. this fixes **icalltoall**

jsquyres:
This sounds right to me.  @bosilca?

bosilca:
On an intercommunicator the operation is not symmetric as indicated in the MPI standard v3.3 page 169 line 22:
> When a complete exchange is executed on an intercommunication domain, then the number of data items sent from processes in group A to processes in group B need not equal the number of items sent in the reverse direction. In particular, one can have unidirectional communication by specifying sendcount = 0 in the reverse direction. (End of advice to users.)

Thus, the extra check is not correct for the intercommunicator case. However, it remains valid for the intracommunicator case, fact that is not correctly handled by this patch.

artpol84:
I pointed to the place where it is handled. It only enabled when Params
check is enabled. But it exists.

четверг, 10 декабря 2015 г. пользователь bosilca написал:

> On an intercommunicator the operation is not symmetric as indicated in the
> MPI standard v3.3 page 169 line 22:
>
> When a complete exchange is executed on an intercommunication domain, then
> the number of data items sent from processes in group A to processes in
> group B need not equal the number of items sent in the reverse direction.
> In particular, one can have unidirectional communication by specifying
> sendcount = 0 in the reverse direction. (End of advice to users.)
>
> Thus, the extra check is not correct for the intercommunicator case.
> However, it remains valid for the intracommunicator case, fact that is not
> correctly handled by this patch.
>
> —
> <https://github.com/open-mpi/ompi/pull/1202#issuecomment-163656200>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
@bosilca here:
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoall.c#L85:L92

bosilca:
Your patch remove the shortcut for all types of MPI_Alltoall. I would like to keep it for intracommunicators.

artpol84:
It's not a problem to check this if intracomm, however I still don't fully
understand you:
a) Ithought that the check I've referenced
https://github.com/open-mpi/ompi/blob/master/ompi/mpi/c/alltoall.c#L85:L92
does the same thing. Isn't it? Can you explain what's the difference? Do
you want to check this case even if parameter check is off?
b) I've changed only alltoall.c, all other alltoall*.c files remain
untouched and they didn't had that check at all. Why do we want to do that
for MPI_Alltoall while not doing that for others?


четверг, 10 декабря 2015 г. пользователь bosilca написал:

> Your patch remove the shortcut for all types of MPI_Alltoall. I would like
> to keep it for intracommunicators.
>
> —
> <https://github.com/open-mpi/ompi/pull/1202#issuecomment-163671167>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
Those two checks are indeed different:

```C
ompi_datatype_type_size(recvtype, &recvtype_size);		
if( (0 == recvcount) || (0 == recvtype_size) ) {		
    return MPI_SUCCESS;		
}		
```

```C
if (MPI_IN_PLACE != sendbuf && !OMPI_COMM_IS_INTER(comm)) {
    size_t sendtype_size, recvtype_size;
    ompi_datatype_type_size(sendtype, &sendtype_size);
    ompi_datatype_type_size(recvtype, &recvtype_size);
    if ((sendtype_size*sendcount) != (recvtype_size*recvcount)) {
        return OMPI_ERRHANDLER_INVOKE(comm, MPI_ERR_TRUNCATE, FUNC_NAME);
    }
}
```

But the last one seems closer to what standard wants. According to standart receiving 0 bytes is OK in case you intend to send 0 bytes.

bosilca:
The check you references is a validity check only executed when we check the correctness of the MPI call arguments, and the only outcome of the check is to fail the call. The check you removed is an optimization, it prevent a process from sending zero bytes messages when the collective communications has no tangible outcome.

For most of the other alltoall* functions a similar condition is more complicated to compute, which might explain why they are missing the same type of optimization.

artpol84:
Ok, I'll return it in case the of the inta-rcomm.

2015-12-11 8:12 GMT+06:00 bosilca <notifications@github.com>:

> The check you references is a validity check only executed when we check
> the correctness of the MPI call arguments, and the only outcome of the
> check is to fail the call. The check you removed is an optimization, it
> prevent a process from sending zero bytes messages when the collective
> communications has no tangible outcome.
>
> For most of the other alltoall* functions a similar condition is more
> complicated to compute, which might explain why they are missing the same
> type of optimization.
>
> —
> <https://github.com/open-mpi/ompi/pull/1202#issuecomment-163812143>.
>



-- 
С Уважением, Поляков Артем Юрьевич
Best regards, Artem Y. Polyakov

artpol84:
@bosilca The patch is updated. I need some time for runtime validation (just to be safe). Will post here if all is OK.

bosilca:
Thanks, the patch looks much better.

artpol84:
@bosilca @jsquyres 
I've checked - works as expected. From my perspective PR can be merged.

hjelmn:
Please PR to v2.x and v1.10.

artpol84:
@hjelmn it's not applicable to 1.10:
https://github.com/open-mpi/ompi-release/blob/v1.10/ompi/mpi/c/alltoall.c#L101:L103

artpol84:
@hjelmn It's not applicable to v2.x as well:
https://github.com/open-mpi/ompi-release/blob/v2.x/ompi/mpi/c/alltoall.c#L99:L105

PR #1205, opened 2015-12-10
travis: add config file for travis-ci.org

jsquyres:

PR #1206, opened 2015-12-11
Poc/travis

ggouaillardet:


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ggouaillardet:
@jsquyres @hppritcha 
finally I got Travis working for both Linux and OS X 
I do not fully understand what is going wrong ...
I used --enable-mpirun-prefix-by-default for both build and distcheck and that worked
I guess installing ompi in $HOME was not a good idea, since distcheck might use this instead of _build
anyway, I will continue on Monday
in the mean time, feel free to use my .travis.yml after you removed all the debugging stuff

jsquyres:
@ggouaillardet Cool!  How did you get it working?  I have two main problems left:

1. The GNU Autotools versions are too low.  I had to change the Autoconf requirements in several configure.ac's in the tree to make it work (which is undesirable).  apt-get said that the versions they have at Travis are the latest available.  ...oh, is it because you did `dist: trusty`?
1. After manually adjusting for the autotools (just so that I could run at all), I got several fails in the ddt tests (see https://s3.amazonaws.com/archive.travis-ci.org/jobs/96201631/log.txt).  It looks like you were mucking with this, too; I see some ldd's for datatype tests, etc.

ggouaillardet:
@jsquyres 
to solve the autotools issue, I added
sudo: required
dist: trusty

to solve the distcheck issue, i added
--enable-mpirun-prefix-by-default
DISTCHECK_CONFIGURE_FLAGS global variable

I did not test only these minimal changes, but I am pretty confident this is enough

I am thinking a simpler option might be not to install ompi in $HOME, since $HOME/lib is in LD_LIBRARY_PATH

I did not test that yet

jsquyres:
Cool.

I solved the "install" problem by invoking `make install` followed by `make uninstall` (before invoking `make check`).  I.e., we just want to make sure `install` and `uninstall` work -- we don't need them there to cloud the `make check` when it runs.  More testing happening on #1205...

ggouaillardet:
make distcheck already does make uninstall, so there might not need to test this twice
make distcheck is also very long, so you might want to use -j 4

I suspect things would be easier if ompi prefix was different (and not in the path) than lib fabric prefix (that has to be in the path)

jsquyres:
Ya, good points.

Sent from my phone. No type good.

On Dec 11, 2015, at 7:36 AM, Gilles Gouaillardet <notifications@github.com<mailto:notifications@github.com>> wrote:


make distcheck already does make uninstall, so there might not need to test this twice
make distcheck is also very long, so you might want to use -j 4

I suspect things would be easier if ompi prefix was different (and not in the path) than lib fabric prefix (that has to be in the path)

-

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ggouaillardet:
This was a test pr, #1205 is now good to go, so I am closing this one

PR #1212, opened 2015-12-13
Update the PMIx native component to release v1.1.2

rhc54:
Update the PMIx native component to release v1.1.1, with addition of one bug-fix commit beyond the official release

Rename the pmix1xx component to pmix111 so it reflects the actual release it includes

Resolve the problem of PMIx being passed a bogus --with-platform argument when configuring the PMIx tarball code. There is no reason we should be passing --with-platform arguments to any internal subdirectory, so just leave that out when constructing the opal_subdir_args variable.

Update the PMIx code and continue attempting to debug direct modex

Fix a problem in the ORTE PMIx server - there was an early intent to optimize the direct modex by fetching data for all procs from the target job on the remote node, instead of fetching the data one proc at a time. However, this was never completely implemented, and so we would hang if we had multiple overlapping requests for data from more than one proc on the node.

Update PMIx to v1.1.2


artpol84:
open-mpi/ompi#2072

jladd-mlnx:
@rhc54 @artpol84 Does this need to go into OMPI 2.1?

PR #1221, opened 2015-12-15
btl-portals4: remove unnecessary PtlMDBind result check

tkordenbrock:
When PtlMDBind was removed, the result check was left in which
causes intermittent failures depending on the junk value found in
the 'ret' variable.  The commit removes the result check.

@regrant - please review


regrant:
:+1:

Issue #1220, opened 2015-12-14
Java bindings not loading in OS X 10.11 (El Capitan)


shurickdaryin:
OS X 10.11 does not allow DYLD_LIBRARY_PATH environment variable to be propagated to child processes. As such, it is not possible to find libmpi.dylib:

```
mpirun -np 4 java -jar myapp.jar
Java bindings failed to load libmpi: dlopen(libmpi.dylib, 10): image not found
Java bindings failed to load libmpi: dlopen(libmpi.dylib, 10): image not found
Java bindings failed to load libmpi: dlopen(libmpi.dylib, 10): image not found
Java bindings failed to load libmpi: dlopen(libmpi.dylib, 10): image not found
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[9970,1],3]
  Exit code:    1
--------------------------------------------------------------------------
```

See https://github.com/oracle/node-oracledb/issues/231, http://apple.stackexchange.com/questions/215030/el-capitan-make-check-dyld-library-path

A workaround is to copy libmpi.dylib to the current directory.

Probably a full path name should be used to load libmpi.dylib to avoid using DYLD_LIBRARY_PATH.

ggouaillardet:
@shurickdaryin did you configure with ```--enable-mpirun-prefix-by-default``` ?
iirc, that should fix the issue

shurickdaryin:
@ggouaillardet I reconfigured with --enable-mpirun-prefix-by-default and rebuilt openmpi and my app, but still get the same error.

ggouaillardet:
were you able to test with Yosemite ? and did you get the same error ?
btw, do you run on only one node ?(vs several nodes)
a possible fix is to update ompi/mpi/java/c/mpi_MPI.c and dlopen with the full path if ompi is configured with --enabke-mpirun-prefix-by-default
I am still running yosemite, so I might ask you to test a fix tomorrow

shurickdaryin:
No, it's El Capitan all around me. But I have tried to hard code the path to libmpi.dylib in the indicated file and it works. So the proposed fix should also work.

shurickdaryin:
This has been run on one node.

jsquyres:
[...]

PR #1227, opened 2015-12-16
java: try do dlopen libmpi with the full path

ggouaillardet:
Since OS X 10.11 (aka El Capitan) DYLD_LIBRARY_PATH is no more
propagated to children, so try to dlopen libmpi with the full path
(set at configure time)

Fixes open-mpi/ompi#1220

Thanks Alexander Daryin for reporting this

### DIFF
--- a/config/ompi_setup_java.m4
+++ b/config/ompi_setup_java.m4
@@ -74,6 +74,9 @@ AC_DEFUN([OMPI_SETUP_JAVA_BINDINGS],[
         # header file needs this file, so we need to check for
         # it/include it in our sources when compiling on Mac).
         AC_CHECK_HEADERS([TargetConditionals.h])
+
+        # dladdr and Dl_info are required to build the full path to libmpi on OS X 10.11 aka El Capitan
+        AC_CHECK_TYPES([Dl_info], [], [], [[#include <dlfcn.h>]])
     else
         AC_MSG_RESULT([no])
         WANT_MPI_JAVA_SUPPORT=0
--- a/ompi/mpi/java/c/Makefile.am
+++ b/ompi/mpi/java/c/Makefile.am
@@ -1,9 +1,11 @@
 # -*- makefile -*-
 #
 # Copyright (c) 2011-2013 Cisco Systems, Inc.  All rights reserved.
-# Copyright (c) 2012	Oracle and/or its affiliates.  All rights reserved.
+# Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2015      Los Alamos National Security, LLC. All rights
 #                         reserved.
+# Copyright (c) 2015      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -44,7 +46,7 @@ libmpi_java_la_SOURCES = \
 	mpi_Status.c \
 	mpi_Win.c
 
-libmpi_java_la_LIBADD = $(top_builddir)/ompi/libmpi.la
+libmpi_java_la_LIBADD = -ldl $(top_builddir)/ompi/libmpi.la
 libmpi_java_la_LDFLAGS = -version-info $(libmpi_java_so_version)
 
 endif
--- a/ompi/mpi/java/c/mpi_MPI.c
+++ b/ompi/mpi/java/c/mpi_MPI.c
@@ -14,6 +14,8 @@
  *                         reserved.
  * Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015      Intel, Inc. All rights reserved.
+ * Copyright (c) 2015      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -62,8 +64,13 @@
 #ifdef HAVE_SYS_STAT_H
 #include <sys/stat.h>
 #endif
+#ifdef HAVE_DLFCN_H
 #include <dlfcn.h>
+#endif
 #include <poll.h>
+#ifdef HAVE_LIBGEN_H
+#include <libgen.h>
+#endif
 
 #include "opal/util/output.h"
 #include "opal/datatype/opal_convertor.h"
@@ -126,7 +133,27 @@ jint JNI_OnLoad(JavaVM *vm, void *reserved)
 {
     libmpi = dlopen("libmpi." OPAL_DYN_LIB_SUFFIX, RTLD_NOW | RTLD_GLOBAL);
 
-    if(libmpi == NULL)
+#if defined(HAVE_DL_INFO) && defined(HAVE_LIBGEN_H)
+    /*
+     * OS X El Capitan does not propagate DYLD_LIBRARY_PATH to children any more
+     * so if previous dlopen failed, try to open libmpi in the same directory
+     * than the current libmpi_java
+     */
+    if(NULL == libmpi) {
+        Dl_info info;
+        if(0 != dladdr((void *)JNI_OnLoad, &info)) {
+            char libmpipath[OPAL_PATH_MAX];
+            char *libmpijavapath = strdup(info.dli_fname);
+            if (NULL != libmpijavapath) {
+                snprintf(libmpipath, OPAL_PATH_MAX-1, "%s/libmpi." OPAL_DYN_LIB_SUFFIX, dirname(libmpijavapath));
+                free(libmpijavapath);
+                libmpi = dlopen(libmpipath, RTLD_NOW | RTLD_GLOBAL);
+            }
+        }
+    }
+#endif
+
+    if(NULL == libmpi)
     {
         fprintf(stderr, "Java bindings failed to load libmpi: %s\n",dlerror());
         exit(1);


ompi/mpi/java/c/Makefile.am
```diff
@@ -14,7 +16,7 @@
 if OMPI_WANT_JAVA_BINDINGS
 
 # Get the include files that were generated from the .java source files
-AM_CPPFLAGS = -I$(top_builddir)/ompi/mpi/java/java $(OPAL_JDK_CPPFLAGS) -DOPAL_DYN_LIB_SUFFIX=\"$(OPAL_DYN_LIB_SUFFIX)\"
+AM_CPPFLAGS = -I$(top_builddir)/ompi/mpi/java/java $(OPAL_JDK_CPPFLAGS) -DOPAL_DYN_LIB_SUFFIX=\"$(OPAL_DYN_LIB_SUFFIX)\" -DOMPI_LIBDIR=\"$(OMPI_WRAPPER_LIBDIR)\"
```
jsquyres: I don't think that this is the right way.  You should be using the OPAL installdirs framework to get the libdir.

ggouaillardet: I started with this approach ...
until I figured out opal is not available nor initialized before libmpi is dlopen'ed (!)

that being said, we could use an other environment variable instead of hard-coding the path.

any thoughts ?

jsquyres: Oh, the `JNI_onLoad()` fires before `MPI_INIT`?  Hmm.  Yes, that does create a quandary.  

Hard-coding the answer at compile time seems like it will someday lead to sadness when someone tries to relocate their Open MPI installation, setenv `OPAL_PREFIX` (or one of the others), and is confused as to why the Java bindings don't work.

I'm not sure what the Right answer is.

BTW, mention @shurickdaryin just so that he is aware of this PR.

ggouaillardet: me too...

btw, is there any reason why libmpijava is not linked with libmpi ?
that should make things trivial (thought not relocatable), at least when ompi is configure'd with --enable-mpirun-prefix-by-default.
I can also try to dlopen $OPAL_PREFIX/lib64/libmpi, then $OPAL_PREFIX/lib/libmpi before trying to dlopen the hard coded path. that would make things a bit better on el capitan

ggouaillardet: I just had an other idea...
libmpi is dlopen'ed from libmpi_java, and both libraries are in the same directory.
so if we can find a way to retrieve libmpi_java directory, we can build the path for libmpi.
I ll try to achieve this with dlsym and friends, otherwise I will try to see what the jvm can do for us
(api and/or environment and/or command line)
any thoughts ?

jsquyres: You know, I can't think of a reason why libmpijava is not linked against libmpi (!).  That *may* actually solve the problem described in the comment (that java will dlopen libmpi in a private namespace) -- but I think you'll need to make at least one code path from libmpijava to call into libmpi to make that linkage work.

If that works, when `libmpijava.so` is loaded, it'll pull `libmpi.so` with it, and we won't need the `dlopen()`.

As for the other ideas (if linking against libmpi doesn't work out): 

* trying to dlopen `$OPAL_PREFIX...` is pretty much what installdirs does (you'd also need to check `$OPAL_LIBDIR`, IIRC).  I'm a bit hesitant to basically repeat (some of) that logic up here.  Hmm.
* I think you're right that we should be able to assume that libmpijava is in the same dir as libmpi.  So if you can find some way to discover that directory (even if it's an OS X-specific way), that might be a good way to go.

ggouaillardet: ok, will try tomorrow.
dladdr can be used to find the path of libmpijava, this is a gnu extension on linux, but "standard" in OS X

jsquyres: Cool -- can easily add a configure test for `dladdr()`.

hppritcha:
libmpi_java is linked against libmpi

hppritcha:
there was an explanation around the dlopen code explaining the logic behind the dlopen.  Had to do with the scope of the symbols as imported by the jvm vs what Open MPI likes.

I would want any solution to have a method for linux users to override a hardwired path for libmpi.so/dylib.

hppritcha:
@nrgraham23 you may want to look this PR over

jsquyres:
@hppritcha Yeah, I know why the `dlopen()` is there -- I was just hoping that linking it in manually would have the same effect (i.e., bring in the relevant symbols into the global scope as opposed to a private scope).

jsquyres:
@hppritcha The one difference I was advocating was to create a code path from libmpijava into libmpi -- so that not only does it link against it, but it also calls something in libmpi.  This might make a difference in what the linker actually loads into the global namespace...?  I.e., if we link libmpi to libmpijava today, if there's no libmpi symbols used by libmpijava, the linker might realize it's a no-op.

ggouaillardet:
for the record, is here the first attempt that is using hard coded path if the initial dlopen fails
```patch
commit 1630de64d5760d38f7f315d11ca7bc2f5e1717a1
Date:   Wed Dec 16 15:39:39 2015 +0900

    java: try do dlopen libmpi with the full path
    
    Since OS X 10.11 (aka El Capitan) DYLD_LIBRARY_PATH is no more
    propagated to children, so try to dlopen libmpi with the full path
    (set at configure time)
    
    Fixes open-mpi/ompi#1220
    
    Thanks Alexander Daryin for reporting this

diff --git a/ompi/mpi/java/c/Makefile.am b/ompi/mpi/java/c/Makefile.am
index 95615ea..306c5c9 100644
--- a/ompi/mpi/java/c/Makefile.am
+++ b/ompi/mpi/java/c/Makefile.am
@@ -1,9 +1,11 @@
 # -*- makefile -*-
 #
 # Copyright (c) 2011-2013 Cisco Systems, Inc.  All rights reserved.
-# Copyright (c) 2012   Oracle and/or its affiliates.  All rights reserved.
+# Copyright (c) 2012     Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2015      Los Alamos National Security, LLC. All rights
 #                         reserved.
+# Copyright (c) 2015      Research Organization for Information Science
 # $COPYRIGHT$
 #
@@ -14,7 +16,7 @@
 if OMPI_WANT_JAVA_BINDINGS
 
 # Get the include files that were generated from the .java source files
-AM_CPPFLAGS = -I$(top_builddir)/ompi/mpi/java/java $(OPAL_JDK_CPPFLAGS) -DOPAL_DYN_LIB_SUFFIX=\"$(OPAL_DYN_LIB_SUFFIX)\"
+AM_CPPFLAGS = -I$(top_builddir)/ompi/mpi/java/java $(OPAL_JDK_CPPFLAGS) -DOPAL_DYN_LIB_SUFFIX=\"$(OPAL_DYN_LIB_SUFFIX)\" -DOMPI_LIBDIR=\"$(OMPI_WRAPPER_LIBDIR)\"
 
 headers = \
         mpiJava.h
diff --git a/ompi/mpi/java/c/mpi_MPI.c b/ompi/mpi/java/c/mpi_MPI.c
index 5b3a39e..55941ab 100644
--- a/ompi/mpi/java/c/mpi_MPI.c
+++ b/ompi/mpi/java/c/mpi_MPI.c
@@ -14,6 +14,8 @@
  * Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015      Intel, Inc. All rights reserved.
+ * Copyright (c) 2015      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -124,9 +126,17 @@ OBJ_CLASS_INSTANCE(ompi_java_buffer_t,
  */
 jint JNI_OnLoad(JavaVM *vm, void *reserved)
 {
-    libmpi = dlopen("libmpi." OPAL_DYN_LIB_SUFFIX, RTLD_NOW | RTLD_GLOBAL);
+    libmpi = dlopen("/libmpi." OPAL_DYN_LIB_SUFFIX, RTLD_NOW | RTLD_GLOBAL);
+    if (NULL == libmpi) {
+        /*
+         * OS X El Capitan does not propagate DYLD_LIBRARY_PATH to children any more
+         * so if the previous dlopen failed, try again with the full path
+         * see https://github.com/open-mpi/ompi/issues/1220
+         */
+        libmpi = dlopen(OMPI_LIBDIR "/libmpi." OPAL_DYN_LIB_SUFFIX, RTLD_NOW | RTLD_GLOBAL);
+    }
 
-    if(libmpi == NULL)
+    if(NULL == libmpi)
     {
         fprintf(stderr, "Java bindings failed to load libmpi: %s\n",dlerror());
         exit(1);
```

ggouaillardet:
@jsquyres @hppritcha @nrgraham23 @shurickdaryin
here is a second attempt.
if the initial ```dlopen("libmpi." ...)``` fails, then try to open ```libmpi``` in the same directory than ```libmpi_java```

config/ompi_setup_java.m4
```diff
@@ -74,10 +74,22 @@ AC_DEFUN([OMPI_SETUP_JAVA_BINDINGS],[
...
         # it/include it in our sources when compiling on Mac).
         AC_CHECK_HEADERS([TargetConditionals.h])
+
+        # dladdr is required to build the full path to libmpi on OS X 10.11 aka El Capitan
+        # OPAL_CHECK_PACKAGE([mpi_java], [dlfcn.h], [dl], [dladdr], [], [], [], [OMPI_HAVE_DLADDR=1], [OMPI_HAVE_DLADDR=0])
+        #AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <dlfcn.h> ]],
+                                        #[[Dl_info info; dladdr((void *)0, &info);]])],
+                       #[OMPI_HAVE_DLADDR=1],
+                       #[OMPI_HAVE_DLADDR=0])
+        AC_CHECK_TYPE([Dl_info], [OMPI_HAVE_DLADDR=1], [OMPI_HAVE_DLADDR=0], [[#include <dlfcn.h>]])
```
jsquyres: Can you replace most of the changes in this file with `AC_CHECK_TYPES`?  We do this kind of thing in a bunch of places (even though it will use the GNU convention of `#define` vs. `#undef`).

ompi/mpi/java/c/Makefile.am
```diff
@@ -44,7 +46,7 @@ libmpi_java_la_SOURCES = \
 	mpi_Status.c \
 	mpi_Win.c
 
-libmpi_java_la_LIBADD = $(top_builddir)/ompi/libmpi.la
+libmpi_java_la_LIBADD = -ldl $(top_builddir)/ompi/libmpi.la
```
jsquyres: Does this naked `-ldl` need to be protected?  E.g.:

* is `-ldl` available on every platform?  (I don't think it is...?)
* does configure elsewhere choose specifically not to link against `dl` when we `--disable-dlopen`?
* what happens if you `-ldl` in a static build?

config/ompi_setup_java.m4
ggouaillardet: will do,
btw, I pushed my stuff before I removed the debug stuff

ompi/mpi/java/c/Makefile.am
ggouaillardet: well, libmpi_java calls dlopen, and libmpi_java can only be a dynamic library, so I guess this part of the code is already protected somewhere in an upper level

lanl-ompi:
Test FAILed.

ggouaillardet:
:bot:retest

ompi/mpi/java/c/Makefile.am
jsquyres: @ggouaillardet Ah, right -- duh!  So do we already have protection for only building this Java stuff when we know we can build it as a shared library?  (Forgive me -- it's been eons since I've looked in this part of the code...)  Also, do we know that the library is *always* `-ldl`?  Can it be called other things (e.g., on Cygwin)?

hppritcha:
:+1:
@ggouaillardet I think this is good enough.  Please open a PR against v2.x and v1.10.2 when you get a chance.

ggouaillardet:
@hppritcha i just noticed i mixed ```#ifdef HAVE_xxx_H``` and ```#if HAVE_xxx_H``` so i updated(and rebased) this PR
once it passes the tests, i will merge it and make the PRs for v2.x and v1.10

PR #1229, opened 2015-12-16
Poc/fortran check use only

ggouaillardet:


jsquyres:
@ggouaillardet I'm not sure why Travis elected not to build this PR.  It may have something to do with a problem Travis had with the open-mpi/ompi repo last night (I mailed Travis support last night asking why all of our Travis jobs from yesterday still hadn't started yet -- they fixed something and they all eventually processed), meaning: your PR is the first event after this clearing of the backlog.  I emailed Travis support again asking if there's something else weird going on, but I do note that https://travis-ci.org/open-mpi/ompi/requests says that this PR has a missing `.travis.yml`.  Did it get confused somehow, because this commit *updates* `.travis.yml`...?  We'll see what support says.

.travis.yml
```diff
@@ -27,17 +29,30 @@ env:
...
 
-# Install dependencies for the verbs and usnic providers.  Note that
-# the libfabric verbs provider cannot currently build on Travis, so we
-# disable it with --disable-verbs (see
-# https://github.com/ofiwg/libfabric/issues/1553 for a description of
-# the RAI_FAMILY issue).
+# install dependencies for the verbs and usnic providers
+# libfbric verbs provider cannot current build without RAI_FAMILY
+# see https://github.com/ofiwg/libfabric/issues/1553
+# bla
```
jsquyres: Did you mean to drop the changes I made to this comment?

jsquyres:
@ggouaillardet BTW, did we determine that this is a gfortran 5.x bug?  (requiring multiple source files)  I.e., does it need to be reported upstream?

.travis.yml
ggouaillardet: I thought Travis somehow blacklisted the original pr, so I made a new one
my comment is not related whatsoever to what you wrote above, I simply wanted some change to commit, hopping Travis would start a test.
Travis tested #1228(created a few hours ago) but not this one (!)

I will commit my configure change only tmoor row(I added -I. to FCFLAGS as you pointed), and keep playing with travis in an other pr.
btw, do you want to test 6 environments {linux,osc}*{clang,default gcc,gcc 5} or 4 (e.g. do not test with default gcc since it cannot build f08 bindings)

the error is odd ...
one source file works fine on my mac (yosemite with latest patches, and gcc 5.3 from brew) and centos 7.
but fails on Travis (same gcc 5.3, older xcode, I do not know which osx)
and the test fails with pgi 15.10 with 1 or 3 source files.
so I guess this is not worth reporting, since this issue is solved with recent osx, and I do not know the root cause (gcc ?, OS X ? Xcode ? strange interaction between gcc and old osx/xcode ?)

ggouaillardet:
@jsquyres 
btw, i just found a critical error in my initial ```.travis.yml``` and that did explain why the PR was not built.
i found that by manually checking my config via https://lint.travis-ci.org/
and i could not find this error was reported by travis
(since you have some admin rights, maybe you got something i did not have ?)

jsquyres:
Nope, I didn't see any additional info on Travis about why it didn't build.

Oh well -- I guess some types of .travis.yml errors aren't reported.  So we'll just have to keep that in mind.

jsquyres:
BTW, do you need to PR this .m4 change to v1.10/v2.x?

ggouaillardet:
yep, will do too

PR #1235, opened 2015-12-17
btl/openib: do not mix exp and non exp verbs

ggouaillardet:


ggouaillardet:
@hjelmn can you please have a look at this ?
it fixes warnings reported by the compiler (unreachable case in switch, comparison of two different types), it compiles without warning, but I do not know how to test it.

hjelmn:
Ah, yes. Good catch.

hjelmn:
PR to 2.x

PR #1237, opened 2015-12-17
Fix add_proc deadlock.

artpol84:
`ompi_proc_pack()` calls `ompi_proc_for_name()` with locked `ompi_proc_lock` and `ompi_proc_for_name` tries to acquire this lock too causing a deadlock.

@hjelmn, please review.

### DIFF
--- a/ompi/proc/proc.c
+++ b/ompi/proc/proc.c
@@ -50,6 +50,7 @@ ompi_proc_t* ompi_proc_local_proc = NULL;
 
 static void ompi_proc_construct(ompi_proc_t* proc);
 static void ompi_proc_destruct(ompi_proc_t* proc);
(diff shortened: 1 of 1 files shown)


hjelmn:
Not quite right. If the proc is in the hash table NULL should not be returned.

artpol84:
Ah, yes! a typo - will fix that, thank you.

artpol84:
@hjelmn, fixed. Thank you for the warning.

artpol84:
one slight change to make it look solid.

artpol84:
and small fix to silence the warnings

artpol84:
and one more to silence all warnings

artpol84:
@hjelmn sorry for the mess. This patch looks right to me. Please review.

hjelmn:
Hah, that is one benefit of turning on the locks by default in ompi_proc_t.

:+1:

PR #1238, opened 2015-12-17
Cleanup warnings in opal and orte layers when building optimized on Mac

rhc54:


opal/mca/pmix/pmix112/pmix1_client.c
```diff
@@ -63,7 +63,7 @@ static void myerr(pmix_status_t status,
         iptr = OBJ_NEW(opal_value_t);
         iptr->key = strdup(info[n].key);
         pmix1_value_unload(iptr, &info[n].value);
-        opal_list_append(&plist, &nm->super);
+        opal_list_append(&plist, &iptr->super);
```
jsquyres: @rhc54 This looks like more than a warning stomp -- did you mean to include this in this commit?  (same goes for the change in pmix1_server_south.c)

rhc54: Yes it is - but yes I did. I saw no reason to separate things.

PR #1239, opened 2015-12-17
Cleanup the warnings from the ompi layer when compiling optimized under Mac OSX

rhc54:


rhc54:
@hjelmn Please take a look at this - there are some places where the warning was over uninitialized values. I believe the issue is that you don't check the return code on a function that also returns, e.g., a pointer in one of its params - and then you use that pointer later. I went ahead and initialized the var, but that just means you'd segfault later if the function failed to return the pointer.

ompi/mca/bcol/basesmuma/bcol_basesmuma_buf_mgmt.c
```diff
@@ -99,6 +99,9 @@ int bcol_basesmuma_free_buff( sm_buffer_mgmt * buff_block,
      *   associated with this bank have been freed.
      */
     assert(generation == buff_block->ctl_buffs_mgmt[memory_bank].bank_gen_counter);
+#if !OPAL_ENABLE_DEBUG
```
jsquyres: Just curious -- is there a reason not to put the `assert()` inside `#if OPAL_ENABLE_DEBUG`?  That would obviate the need for the spurious `(void)generation` line.

rhc54: heck if I care - I just used a technique that had been used in a number of other similar places in the code

jsquyres: Yeah, that PR came in recently with the weird `(void)` technique -- I didn't put 2 and 2 together back then to realize that it was because of a variable that only exists in debug builds.

bosilca: Why not simply adding the (void) in all cases and remove the ifdef ?

rhc54: fine with me - just avoiding it when not necessary.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

rhc54:
That doesn't make any sense....try it again

bot:retest

PR #1249, opened 2015-12-21
Do not override any external settings for PMIx component selection

rhc54:


hppritcha:
@rhc54 I'm curious about this one.  Is there a  use case where one might want to use one of the non-pmix components when launching using orte?

rhc54:
No - but you might want to select one of the pmix components, or someone may add something different, so overriding the selection to a forced value of the MCA param isn't a good idea

PR #1251, opened 2015-12-21
odls/alps: changes to accommodate Cray slurm

hppritcha:
It turns out on Cray systems running nativized
SLURM, that the ORTE HNP daemon actually launches
the MPI processes on the first node in the
allocation.  Changes to the odls/alps launcher
to account for this.

@hjelmn 

### DIFF
--- a/orte/mca/odls/alps/odls_alps_component.c
+++ b/orte/mca/odls/alps/odls_alps_component.c
@@ -81,13 +81,14 @@ int orte_odls_alps_component_query(mca_base_module_t **module, int *priority)
     bool flag;
 
     /*
-     * make sure we're in a daemon process
+     * make sure we're in a daemon process.  On systems using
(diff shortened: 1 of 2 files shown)


hppritcha:
Well it turns out that this doesn't work.  The way the HNP daemon is launched (not using srun), there is no way for the MPI ranks on the head node to get the correct RDMA credentials.  Not sure if its worth trying to support using mpirun on Cray systems using nativized slurm.

I will rework this PR to instead cause the job to fail with an info message if a user tries to launch an Open MPI job with mpirun on a Cray system running nativized slurm.  

@hjelmn FYI
@rhc54 (any ideas on how this can be fixed?).  Basically to get the RDMA credentials for the Cray HSN, a process must be a child of a slurmd daemon.

hppritcha:
closing this pull request, this approach won't work.

PR #1255, opened 2015-12-22
Fixup

annu13:
This PR only contains one change in file ess_base_std_orted.c, the rest of the commits are a result of some rebase and incorrect merge operations.

### DIFF
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -300,8 +300,79 @@ int orte_ess_base_orted_setup(char **hosts)
             }
         }
     }
+    /* setup the global job and node arrays */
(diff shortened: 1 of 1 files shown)


artpol84:
@annu13 this is steel incorrect. 
Commit list has to be just one commit. If you need to merge you're doing something wrong. 
In short you need to checkout fresh master, push it as a new branch in your fork and apply you the only one commit. and then do the PR.

rhc54:
@artpol84 I don't understand your concern - there is only one commit in this PR. This looks just fine.

artpol84:
Now it's OK. Note that @annu13 has changed the PR after my comment (according to the timeline). It wasn't OK but now it's fine.

rhc54:
Ah, ok - thx!

PR #1256, opened 2015-12-22
mtl-portals4: initialize endpoint nid/pid when using logical mapping

tkordenbrock:
When mtl-portals4 is configured for logical mapping, coll-portals4
must disqualify because it does not yet support logical mapping.
coll-portals4 looks for the endpoint pid to be zero which tells it
that mtl-portals4 is configured for logical mapping.  This commit
initializes the endpoint nid/pid to zero for logical mapping.


regrant:
:+1:

PR #1257, opened 2015-12-22
plm/alps: only use srun for Native SLURM

hppritcha:
Turns out that the way the SLURM plm works
is not compatible with the way MPI processes
on Cray XC obtain RDMA credentials to use
the high speed network.  Unlike with ALPS,
the mpirun process is on the first compute
node in the job.  With the current PLM launch
system, mpirun (HNP daemon) launches the MPI
ranks on that node rather than relying on
srun.

This will probably require a significant amount
of effort to rework to support Native SLURM
on Cray XC's.  As a short term alternative,
have the alps plm (which gets selected by default
again on Cray systems regardless of the launch system)
check whether or not srun or alps is being used on the
system.  If alps is not being used, print a helpful
message for the user and abort the job launch.

@hjelmn  for your info, perhaps suggestions
@rhc54, @marksantcroos an FYI.

Mark, probably don't want to bother trying to get allocations for your work on
the cori system at NERSC.  The orte-dvm will have the same issues.  If your
use case doesn't require high performance interconnect, you may be able
to get things to work by using self,vader,tcp btls.

### DIFF
--- a/orte/mca/plm/alps/help-plm-alps.txt
+++ b/orte/mca/plm/alps/help-plm-alps.txt
@@ -39,3 +39,7 @@ the map for this application. This can be caused by a lack of
 an allocation, or by an error in the Open MPI code. Please check
 to ensure you have a ALPS allocation. If you do, then please pass
 the error to the Open MPI user's mailing list for assistance.
+#
+[slurm-not-supported]
+mpirun is not a supported launcher on Cray XC using Native SLURM.
+srun must be used to launch jobs on these systems.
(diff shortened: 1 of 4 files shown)


rhc54:
@hppritcha I think this can be resolved. I gather that the problem is that Cray runs mpirun on a node of the allocation, but it should be possible to have mpirun start an orted on mpirun's node. Not something for 2.0, probably, but certainly by 2.1 at the latest.

hppritcha:
thanks!  definitely can wait till at least 2.1.

PR #1260, opened 2015-12-24
Openib proc accounting fix

artpol84:
@hjelmn 
I'm still working on this, however I'd like to know your position regarding this changes.
Right now it works for OSU tests and I am currently testing it with the full MTT test suite.

### DIFF
--- a/opal/mca/btl/openib/btl_openib.c
+++ b/opal/mca/btl/openib/btl_openib.c
@@ -354,8 +354,10 @@ static int create_srq(mca_btl_openib_module_t *openib_btl)
             } else
 #endif
             {
+               opal_mutex_lock(&openib_btl->device->device_lock);
                openib_btl->qps[qp].u.srq_qp.srq =
                    ibv_create_srq(openib_btl->device->ib_pd, &attr);
+               opal_mutex_unlock(&openib_btl->device->device_lock);
             }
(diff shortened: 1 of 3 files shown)


artpol84:
MTT looks fine. No degradation.

artpol84:
Not true - looks like I see some problems with `mpi-test-suite`. Let me check.

artpol84:
@hjelmn I've checked, the problem is out of the scope of this patch.
I'm seeing rear hangs of `mpi_test_suite`, on Gather but this appears with both tcp and openib BTL.

artpol84:
@rhc54 @jsquyres @hppritcha @hjelmn 
Merry Christmas, dear coleagues :).

artpol84:
**Update**: the hang that I was observing with `mpi-test-suite` seems to be related to **vader** BTL. If I use **sm** all runs fine. I recall that @rhc54 has told me that somebody was reporing **vader** hangs. I think that's it.
I'll try to investigate.

artpol84:
`mpi-test-suite` problem addressed in PR https://github.com/open-mpi/ompi/pull/1264.

hjelmn:
PR to v2.x

artpol84:
@hjelmn, to PR this to v2.x we need to merge https://github.com/open-mpi/ompi-release/pull/868 first. This PR depends on it.

PR #1261, opened 2015-12-25
rmaps/round_robin: warn if oversubscribing when manually setting the …

ggouaillardet:
…number of hosts

This is a port of the v1.10 serires one-off open-mpi/ompi-release@8c5ce45ab64ab77eddb798ecf0c3a826021eced5

### DIFF
--- a/orte/mca/rmaps/ppr/rmaps_ppr.c
+++ b/orte/mca/rmaps/ppr/rmaps_ppr.c
@@ -3,6 +3,8 @@
  * Copyright (c) 2011      Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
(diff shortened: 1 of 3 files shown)


ggouaillardet:

that fixes the issue described in http://www.open-mpi.org/community/lists/devel/2015/12/18455.php

with the current machinefile
```
localhost slots=2
```

```mpirun --machinefile machinefile -np 3 ...``` works without any warning without this commit.
with this commit, it aborts because of oversubscription (same behaviour as v1.10)

rhc54:
This looks correct, but don't we also need the changes to the other mappers as well?

ggouaillardet:
I guess so, will do from Monday

ggouaillardet:
@rhc54 i back ported the other changes

rhc54:
@ggouaillardet Thanks!

PR #1270, opened 2015-12-30
Cleanup connection termination

rhc54:
Revert a prior commit that causes mpirun to hang as it doesn't correctly handle TCP connection terminations. Correctly handle some corner cases in the error notification system.


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

PR #1277, opened 2016-01-04
fixes for the plfs file system

edgargabriel:


hppritcha:
@edgargabriel please open PR on v2.x as well.  Does this need to go in to 1.10.x too?

edgargabriel:
I will. I decided however later this afternoon to revert one of the commits in this pr and provide a different solution tomorrow.

edgargabriel:
ok, the plfs_setxattr problem is a mess. I will close this pr without merging, and regenerate a new one where plfs_setxattr is completely removed. The problem is, that there is a discrepancy in the C++ and C interfaces of the plfs_setxattr, in one case having three (C) in the other four (C++ ) arguments. The C binding is defined in the header file, but does not match the implementation in the library (which is C++ based).  I will simply remove this feature right now. 

After further looking into the PLFS github repo, the last commit is one year old, not sure how much development work is going on there.

PR #1278, opened 2016-01-05
osc/pt2pt: use two distinct "namespaces" for tags

ggouaillardet:


hppritcha:
@hjelmn please review this.

Issue #1269, opened 2015-12-30
UDCM hang


artpol84:
@hjelmn I've faced the situation where application hangs because of UDCM. It is **VERY** rare! I saw this only once among 1000's of runs. 
**And I'm still have this hanged app "running" so we can do further postmortem analysis if you'll can think of anything that can be inspected with GDB.**

The application was `mpi_test_suite`. It has hanged at very beginning when application calls `MPI_Comm_dup()` (this seems to be a first communication call in this app).

Backtraces on all ranks looks the same:
```
#0  0x00000030a6ecf527 in sched_yield () from /lib64/libc.so.6
#1  0x00007f57b7aa5c19 in opal_progress () at runtime/opal_progress.c:199
#2  0x00007f57b810c608 in opal_condition_wait (c=0x7f57b8438b40, m=0x7f57b8438ac0) at ../opal/threads/condition.h:76
#3  0x00007f57b810cb99 in ompi_request_default_wait_all (count=2, requests=0x7fffc43044e0, statuses=0x0) at request/req_wait.c:287
#4  0x00007f57b81960ee in ompi_coll_base_allreduce_intra_recursivedoubling (sbuf=0x7fffc43046b8, rbuf=0x7fffc43046bc, count=1, dtype=0x65e4c0, 
    op=0x65fac0, comm=0x660ee0, module=0x2267800) at base/coll_base_allreduce.c:224
#5  0x00007f57b2635bd5 in ompi_coll_tuned_allreduce_intra_dec_fixed (sbuf=0x7fffc43046b8, rbuf=0x7fffc43046bc, count=1, dtype=0x65e4c0, op=0x65fac0, 
    comm=0x660ee0, module=0x2267800) at coll_tuned_decision_fixed.c:66
#6  0x00007f57b80e88ce in ompi_comm_allreduce_intra (inbuf=0x7fffc43046b8, outbuf=0x7fffc43046bc, count=1, op=0x65fac0, comm=0x660ee0, 
    bridgecomm=0x0, local_leader=0x0, remote_leader=0x0, send_first=-1, tag=0x7f57b81c6d2e "nextcid", iter=0) at communicator/comm_cid.c:878
#7  0x00007f57b80e733a in ompi_comm_nextcid (newcomm=0x226db60, comm=0x660ee0, bridgecomm=0x0, local_leader=0x0, remote_leader=0x0, mode=32, 
    send_first=-1) at communicator/comm_cid.c:221
[...]

PR #1283, opened 2016-01-06
Fix race condition in UDCM

artpol84:
where service thread sees that `cm_message_event_active == 1` but main thread has already stopped
processing messages and thus we will have the situation where one
message was left unhandled leading to a hang.

@hjelmn, please check if thes sounds reasonable. I've started my checks but they can take more than one day to reproduce the problem (if still exists). I'm trying to address Issue https://github.com/open-mpi/ompi/issues/1269.


artpol84:
@hjelmn I have 40000 successful runs of ompi-test-suite (2.5 days, 4 independent allocations, 10000 runs in each). Previously it was enough to trigger the bug. I think that this PR indeed fixes Issue https://github.com/open-mpi/ompi/issues/1269.

hjelmn:
PR to v2.0.0

PR #1288, opened 2016-01-07
grpcomm: fix bugs in grpcomm algorithms

hjelmn:
This commit fixes multiple issues in the bruck's and recursive
doubling grpcomm algorithms. The following changes are included:

 - Use the existing bitmap implementation instead of implementing a
   new one. There were bugs in the implementation that caused an
   overrun of the bitmap array.

 - Clean up the algorithms to eliminate errors.

 - Send as little extra data as possible in the bruck's
   algorithm.

The changes were testest with various numbers of ortes varying from 1
to 4096.

### DIFF
--- a/orte/mca/grpcomm/base/grpcomm_base_frame.c
+++ b/orte/mca/grpcomm/base/grpcomm_base_frame.c
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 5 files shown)


hjelmn:
@rhc54 These are the grpcomm changes I was promising you. From what I can tell I can no longer reproduce the SLES hang with these changes. Makes me wonder if the array overrun in both rcd and brks was the cause of that problem.

rhc54:
I think this looks okay. We still have the race condition about when the first message arrives vs when the local procs tell us about a collective vs when we recv/unpack the launch msg, but I'll address that later (already have some related work in a branch).

I'm not sure this is the full cause of the SLES hang, though it does point to a possible cause. The other hang reports on SLES come from the 1.10.2rc3 tarball, and that series doesn't include these components. So there is something else going on over there.

rhc54:
@hjelmn Please ensure this goes over to 2.x

PR #1296, opened 2016-01-12
pmix: fix misc memory leaks involving opal_convert_jobid_to_string

ggouaillardet:


ggouaillardet:
@rhc54 i initially prepared this PR in order to fix misc memory leaks
```opal_convert_jobid_to_string``` allocates the result via ```asprintf``` but it is never released.

then i found that except in ```_cnt``` from ```orte/orted/pmix/pmix_server_dyn.c```, we only use the result of this function as a parameter of ```snprintf```.
bottom line, and though this is clearly not in the critical path, that is pretty unefficient and require extra code to free the allocated memory.

wouldn't it be better to replace ```opal_convert_jobid_to_string``` with ```opal_snprintf_jobid```
and update ```_cnt``` accordingly (e.g. use a ```char jobid[xxx];``` like local variable ) ?

i can make a new PR if you agree this is the way to go

rhc54:
@ggouaillardet I agree with your alternate approach - there are multiple places in the code base where that would be a better. Problem is to figure out a reasonable max size - right now, for OMPI internals, something like 64 or 128 characters would be fine.

ggouaillardet:
@rhc54 i did that in #1302 
now closing this PR

PR #1306, opened 2016-01-15
New MPI_Info handling model

dsolt:
This is my first pull request, please give careful review.   Not sure if this should be on a branch first?   This change is intended for 2.1 and is a new model of handling MPI_Info changes to communicators/windows/files.   It moves much of the info handling to the opal layer, but retains a wrapper for info objects that are exposed to the user.   This allows us to use the info keyvalue concept internally and also implement MPI_Info using this internal concept.   The main driver for this change was to allow different components to see changes to MPI_Info (via MPI_Comm_set_info, etc) and have an opportunity to respond to those changes by either accepting the change and modifying their future behavior or modifying the proposed info change if they do not support it.   There are some scenarios which are difficult for this model.  Particularly if multiple components (called subscribers in the code) use the same key but conflict on the values they support.   Currently the key is presented to each component in turn and each has a chance to modify the value.   The ordering therefore matters.   The key (no pun intended) is to not do this.   Each module should either use its own keys or they should agree on what values they can accept.   This code has received very minimal testing.   I verified that no_locks values could be updated and the new value rejected when the underlying component does not support it.   I can work on further testing, but would like some eyes on this as I do that to make sure the overall direction makes sense and that my approach to moving code around makes sense for the Open MPI architecture.  I would especially like @hjelmn to make sure that this will work for him.

@jsquyres @hjelmn @nysal @jjhursey @gpaulsen


lanl-ompi:
Test FAILed.

dsolt:
Working with @jsquyres -- we're going to close this one and open a new PR.

PR #1307, opened 2016-01-15
Fcoll dynamic gen2

edgargabriel:
will not merge it yet, but want to see the reporst from jenkins

### DIFF
--- a/ompi/mca/fcoll/dynamic_gen2/Makefile.am
+++ b/ompi/mca/fcoll/dynamic_gen2/Makefile.am
@@ -1,0 +1,47 @@
+#
+# Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+#                         University Research and Technology
+#                         Corporation.  All rights reserved.
+# Copyright (c) 2004-2005 The University of Tennessee and The University
+#                         of Tennessee Research Foundation.  All rights
+#                         reserved.
+# Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
+#                         University of Stuttgart.  All rights reserved.
+# Copyright (c) 2004-2005 The Regents of the University of California.
+#                         All rights reserved.
+# Copyright (c) 2008-2015 University of Houston. All rights reserved.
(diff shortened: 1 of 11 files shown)


hppritcha:
@edgargabriel can you point me to a test to run on lustre that will show the problems you were getting on Cray at HLRS?

edgargabriel:
@hppritcha I have an update: this last commit made the hang being exposed again on the cray, this time even with debug built. Digging into that I do *not* see something wrong in the component, but it looks to me like a progression problem. 

I can now even reproduce the hang *without* the new component on my regular cluster as long as I use multiple nodes. It seems that processes are hanging in the poll function of the progress engine in an allreduce operation. I will continue looking into that, but I am 99% sure that the problem is not from my component.

hppritcha:
@edgargabriel on your cluster, which BTL/MTL is being used for inter-node messaging?

edgargabriel:
openib. Everything is also running very slow, but I did not have time to check  what's the reason for that.

edgargabriel:
to refine my previous: I obviously don't know whether the hang on the Cray and on my InfiniBand cluster are related or not. The 'old' components work well on Cray, I see a hang only with the new component. On my IB cluster, we seem to be hanging even with the previous code that used to work without any issues. The two things might not be related.

edgargabriel:
A small update on this item. I have identified the problem with the IB stuff on my network. I (used to) have a flag set for openib in my configuration file, namely 

btl_openib_flags = 305

It looks like removing the flag makes things work fast again. Otherwise, the performance of the openib device drops to ~20MB/s (instead of 2000MB/s). I will check whether this resolves the hangs on my cluster. I also double checked, that the flag does not leave to this performance degradation with the 1.8 series, so its 'new' on master.

edgargabriel:
After resolving the InfiniBand and configure issues on my cluster, I was able to test the module, and everything looks. Tests up to 512 processors finished without any deadlocks etc. Our own as well as the HDF5 testsuite pass all tests for this new module. I will therefore merge the component into trunk.

PR #1309, opened 2016-01-17
ddt: do not explicitly check parameter datatype is not MPI_DATATYPE_N…

ggouaillardet:
…ULL.

Since MPI_DATATYPE_NULL is not a committed datatype, it cannot be
a valid parameter and there is no need for an ad-hoc test.

### DIFF
--- a/ompi/mpi/c/bindings.h
+++ b/ompi/mpi/c/bindings.h
@@ -10,6 +10,8 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@bosilca following the discussion about ```MPI_DATYPE_NULL```, and given the fact this is not a committed datatype in OpenMPI, i found some small room for optimisation when checking parameters.

note if the datatype is not allowed __and__ the count is negative, the behaviour is changed, and OpenMPI will report an error about datatype instead of count.

does that look good to you ?

bosilca:
I'm a believer in "If it ain't broke, don't fix it". I also like the explicit test for the datatype NULL, it is much clear that the obscure fact that NULL datatype is not committed (as you could have noticed from the discussion). I have no strong feeling about this.

ggouaillardet:
ok, i will close this PR then

PR #1311, opened 2016-01-19
revamp the pvfs2 configure logic

edgargabriel:
@jsquyres could you please review this patch?

### DIFF
--- a/config/ompi_check_pvfs2.m4
+++ b/config/ompi_check_pvfs2.m4
@@ -11,7 +11,7 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2006 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2009      Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2008-2012 University of Houston. All rights reserved.
+dnl Copyright (c) 2008-2016 University of Houston. All rights reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl $COPYRIGHT$
@@ -32,10 +32,6 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
     check_pvfs2_LDFLAGS=
     check_pvfs2_LIBS=
 
-    check_pvfs2_save_LIBS="$LIBS"
-    check_pvfs2_save_LDFLAGS="$LDFLAGS"
-    check_pvfs2_save_CPPFLAGS="$CPPFLAGS"
-
     check_pvfs2_configuration="none"
     ompi_check_pvfs2_happy="yes"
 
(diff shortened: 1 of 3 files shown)


config/ompi_check_pvfs2.m4
```diff
@@ -46,70 +46,27 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
...
 
-    AC_ARG_WITH([pvfs2-libs],
-        [AC_HELP_STRING([--with-pvfs2-libs=LIBS],
-                       [Libraries to link with for pvfs2])])
-
-    temp_with_pvfs2="$with_pvfs2"
     AS_IF([test -z "$with_pvfs2"],
-          [with_pvfs2="/usr/local"])
+          [ompi_check_pvfs2_dir="/usr/local"],
+          [ompi_check_pvfs2_dir="$with_pvfs2"])
```
jsquyres: Super-minor quibble (you may care or not): when assigning one string to another in shell script, you don't need quotes.  I.e.:

```sh
ompi_check_pvfs2_dir=$with_pvfs2
```

config/ompi_check_pvfs2.m4
```diff
@@ -46,70 +46,27 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
...
-    # Try to find all the pvfs2 libraries
-    AS_IF([test "$ompi_check_pvfs2_happy" = "yes"],
-	[ AS_IF([test -n "$with_pvfs2_libs"]
-		[for lib in $with_pvfs2_libs ; do
-		    check_pvfs2_LIBS="$check_pvfs2_LIBS -l$lib"
-		    done])
-
-	    $1_LIBS="$check_pvfs2_LIBS"
-	    LIBS="$LIBS $check_pvfs2_LIBS"
+    OPAL_CHECK_PACKAGE([$1], [pvfs2.h], [pvfs2], [PVFS_util_resolve], [-lpthread],
```
jsquyres: I think you need to ensure that `-lpthread` will work before passing it unconditionally down to OPAL_CHECK_PACKAGE.

config/ompi_check_pvfs2.m4
```diff
@@ -46,70 +46,27 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
              [Build Pvfs2 support, optionally adding DIR/include, DIR/lib, and DIR/lib64 to the search path for headers and libraries])])
```
jsquyres: Are the variables above this used at all any more (e.g., the `_save_` variables)?

jsquyres:
I love replacing a lot of customer .m4 with `OPAL_CHECK_PACKAGE`.  :-)

config/ompi_check_pvfs2.m4
edgargabriel: Short question on that: is there a nice way  to extract the required pthreads library from OPAL_CONFIG_POSIX_THREADS? I  added a call to that macro in ompi_check_pvfs2, I can see that it tests succesfully for the -lpthread and friends, I have not found a way on how to extract $PTHREAD_LIBS correctly ( in order to pass it again to OPAL_CHECK_PACKAGE)

jsquyres: Does simply using `$PTHREAD_LIBS` work?

Or has OPAL_CONFIG_POSIX_THREADS already put the relevant token(s) in `$LIBS` and you don't need to specify anything at all...?

edgargabriel: no, unfortunately neither does work. I added an echo$PTHREAD_LIBS, and the variable is empty after calling OPAL_CONFIG_POSIX_THREAD. And using OPAL_CHECK_PACKAGE for the pvfs2 stuff without adding -lpthread fails. I will try to copy some of the pthread library logic from the opal_config_pthreads and see whether I use that.

edgargabriel:
@jsquyres I tried to incorporate your comments, can you please check whenever you have a couple of min?

config/ompi_check_pvfs2.m4
```diff
@@ -46,70 +42,75 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
...
+        ompi_check_pvfs2_libdir="$ompi_check_pvfs2_dir/lib"
+    fi
 
     # Add correct -I and -L flags
-    AS_IF([test -d "$with_pvfs2/include"],
-        [check_pvfs2_CPPFLAGS="-I$with_pvfs2/include"
-            $1_CPPFLAGS="$check_pvfs2_CPPFLAGS"
-            CPPFLAGS="$CPPFLAGS $check_pvfs2_CPPFLAGS"],
-	[ompi_check_pvfs2_happy="no"])
+    OPAL_CHECK_PACKAGE([$1], [pvfs2.h], [pvfs2], [PVFS_util_resolve], [$pthread_lib],
```
jsquyres: @edgargabriel I checked a little further: there's several variables:

* `$PTHREAD_CPPFLAGS`
* `$PTHREAD_CXXCPPFLAGS`
* `$PTHREAD_CFLAGS`
* `$PTHREAD_CXXFLAGS`
* `$PTHREAD_FCFLAGS`
* `$PTHREAD_LIBS`

The pthread setup is run before any of the MCA component configury, so these variables should definitely be available to you.  Some/all may be empty, but I'll bet that in your case, at least `$PTHREAD_CFLAGS` should contain something (e.g., `-pthread`?).

edgargabriel:
@jsquyres you are right, I always just checked PTHREAD_LIBS and that was empty, but some of the other ones are set. I revamped the logic to use those variables, it seems to work on my machine. Let me know what you think, if this version works for you, I will revamp the PLFS configure logic to use the same structure.

jsquyres:
I'm not sure I understad

jsquyres:
#@%#@$ Total comedy of errors there -- I didn't mean to submit that last comment (so I'll delete it :smile:), and I didn't mean to close the PR.  Re-opening...

config/ompi_check_pvfs2.m4
```diff
@@ -46,70 +42,27 @@ AC_DEFUN([OMPI_CHECK_PVFS2],[
...
+        ompi_check_pvfs2_libdir="$ompi_check_pvfs2_dir/lib"
+    fi
 
     # Add correct -I and -L flags
-    AS_IF([test -d "$with_pvfs2/include"],
-        [check_pvfs2_CPPFLAGS="-I$with_pvfs2/include"
-            $1_CPPFLAGS="$check_pvfs2_CPPFLAGS"
-            CPPFLAGS="$CPPFLAGS $check_pvfs2_CPPFLAGS"],
-	[ompi_check_pvfs2_happy="no"])
+    OPAL_CHECK_PACKAGE([$1], [pvfs2.h], [pvfs2], [PVFS_util_resolve], [$PTHEAD_LIBS],
```
jsquyres: Per https://github.com/open-mpi/ompi/pull/1311#issuecomment-174556176, I thought you said `$PTHREAD_LIBS` was empty...?  Shouldn't be also be sending in `$PTHREAD_CFLAGS` and/or others?  

Also, you misspelled `$PTHREAD_LIBS` here.

edgargabriel: I'll fix the typo. PTHREAD_LIBS is empty on my platform, but might not be empty on another. HOw do I pass additional CFLAGS into the PACKAGE? Just concatenate them to  CFLAGS (and copy a copy of the original version for restoring it later)?

jsquyres: I was thinking you'd just pass `[$PTHREAD_CFLAGS $PFLAGS_LIBS]` in that argument.

It's a little mis-use of that argument, but I think it would be "good enough" (`$PTHREAD_CFLAGS` is usually the one that contains `-pthread`, right?).

edgargabriel: I just added an echo $CFLAGS into the section, it looks like PTHREAD_CFLAGS has already been added CFLAGS, since $CFLAGS already contain -pthread. I can still go ahead with your suggestion, would be fine with me.

edgargabriel: Or should leave the additional library section just completely blank, if the PTHREAD stuff is added by default to CFLAGS, LIBS etc. anyway? It should work.

jsquyres: Mmm... good point; we may have filled in the pthreads flags into the main LIBS, CFLAGS, etc. already.  In which case, you're right: you should just pass `[]` for that argument.

Can you take a quick pass at verifying that we do actually do that?  E.g., in opal_config_pthread.m4, right before line 651, do something like:

```sh
PTHREAD_CPPFLAGS="-DEDGAR_IS_AWESOME_CPPFLAGS $PTHREAD_CPPFLAGS"
PTHREAD_LIBS="-DEDGAR_IS_AWESOME_LIBS $PTHREAD_LIBS"
...etc.
```

Then you can see if those (harmless) -D's make it through to CFLAGS, LIBS, etc. by the time we're down in your component's configure.m4.

edgargabriel: Can confirm, the new flags appear when doing an echo on $CPPFLAGS etc. in ompi_check_pvfs2. So I will leave that section of OPAL_CHECK_PACKAGE empty.

jsquyres:
Sweet; it looks good.  Can I ask for one last thing...?  :smile: Can you squash these 3 commits down into 1?  No need to show our-back-n-forth with multiple commits once this hits master.

edgargabriel:
Yep, will do.

edgargabriel:
not sure this is right, why does it should the osc/rdma commit as well?

edgargabriel:
I'll redo the pr, might be easier.

PR #1314, opened 2016-01-20
OSC portals4 : do not generate an EVENT_SEND to avoid to filter it

francois-wellenreiter:


francois-wellenreiter:
@regrant could you please review this enhancement

jsquyres:
@regrant Can you have a look?  Thanks.

regrant:
@francois-wellenreiter while there's no explicit use of the PTL_EVENT_SEND event, it's the only way that you can catch initiator side errors since send events are not counted (CT) either. This could lead to infinite polling on your CTWait call if you're waiting for something that will never complete. (See my comment on https://github.com/open-mpi/ompi-release/pull/903) @jsquyres

regrant:
We've finished looking into the potential issues surrounding this pull request and everything checks out. Merging..

PR #1315, opened 2016-01-21
OSHMEM/UCX: implements atomic support

alex-mikheev:
@miked-mellanox @yosefe 
ucx atomic component has a real code now.
fixes bug in spml ucx arr_procs
removes redundant parameter checks from atomic components.


miked-mellanox:
:+1:

miked-mellanox:
please PR to v2.x and v1.10

PR #1316, opened 2016-01-21
ompi/datatype: make datatype pack thread safe

hjelmn:
This commit makes ompi_datatype_get_pack_description thread safe. The
call is used by osc/pt2pt to send the packed description to remote
peers. Before this commit if MPI_THREAD_MULTIPLE is enabled and the
user uses MPI_Put, MPI_Get, etc we could hit a race where multiple
threads attempt to store the packed description on the datatype. Since
the code in question is not performance-critical the threading fix
uses opal_atomic_* calls instead of bothering with OPAL_THREAD_*.

### DIFF
--- a/ompi/datatype/ompi_datatype_args.c
+++ b/ompi/datatype/ompi_datatype_args.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2006 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
- * Copyright (c) 2013      Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2013-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca Please take a look. This is targeted for 2.0.0.

bosilca:
Make sense, the patch is correct but sub-optimal (the packed description will be generated by all threads and then most of them will drop it).

hjelmn:
@bosilca Was thinking that would be ok in this case. Using atomics avoids adding a lock to the datatype structure. I will see if I can optimize the atomics version a bit more to avoid the the duplicate description generation.

hjelmn:
@regrant This is one found by the rma threading benchmarks.

hjelmn:
@bosilca Ok, optimized it a bit better. Now only one thread will create the datatype description.

bosilca:
@hjelmn this looks good. Thanks for the optimization. :+1:

lanl-ompi:
Test FAILed.

PR #1318, opened 2016-01-21
osc/rdma: disable put aggregation when using threads

hjelmn:
Optimizing put aggregation in the presence of threads will require a
redesign of the code. For now just ensure that put aggregation is
turned off when MPI_THREAD_MULTIPLE is enabled.


hjelmn:
@regrant With this commit I think osc/rdma should run clean multithreaded. Will continue to test.

PR #1334, opened 2016-01-31
Refactor the ORTE DVM code

rhc54:
Refactor the ORTE DVM code so that external codes can submit multiple jobs using only a single connection to the HNP.

* Clean up the DVM so it continues to run even when applications error out and we would ordinarily abort the daemons.
* Create a new errmgr component for the DVM to handle the differences.
* Cleanup the DVM state component.
* Add ORTE bindings directory and brief README
* Pass a local tool index around to match jobs.
* Pass the jobid on job completion.
* Fix initialization logic.
* Add framework for python wrapper.
* Fix terminate-with-non-zero-exit behavior so it properly terminates only the indicated procs, notifies orte-submit, and orte-dvm continues executing.
* Add some missing options to orte-dvm
* Fix a bug in -host processing that caused us to ignore the #slots designator. Add a new attribute to indicate "do not expand the DVM" when submitting job spawn requests.
* It actually makes no sense that we treat the termination of all children differently than terminating the children of a specific job - it only creates confusion over the difference in behavior. So terminate children the same way regardless.

### DIFF
--- a/opal/util/cmd_line.c
+++ b/opal/util/cmd_line.c
@@ -14,6 +14,7 @@
  * Copyright (c) 2012-2015 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      Intel, Inc. All rights reserved
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -146,7 +147,7 @@ static int qsort_callback(const void *a, const void *b);
 int opal_cmd_line_create(opal_cmd_line_t *cmd,
                          opal_cmd_line_init_t *table)
 {
-    int i, ret = OPAL_SUCCESS;
+    int ret = OPAL_SUCCESS;
 
     /* Check bozo case */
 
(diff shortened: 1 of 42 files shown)


rhc54:
@marksantcroos I think this captures what we have done. I adjusted the orte_submit_job signature to emphasize the difference from orte_submit_init. Let me know what you think.

marksantcroos:
@rhc Would it be a good idea to also make orterun.c use this structure? Or at least have only one instance of the supporting functions like parse_locals, parse_globals, etc?

rhc54:
@marksantcroos I am indeed hoping to modify orterun to use this structure. The difference in cmd line options requires a bit of thought as not all of orterun's options make sense for submit. So I'm still pondering the solution.

Either way, factoring out the parse functions makes sense. Do you want to take a crack at it, or shall I?

ompiteam:
@marksantcroos I think I figured out how to handle the cmd line issues, so I'll take a crack at it later today.

rhc54:
Sorry about that - forgot to logout of the admin account after doing something. The above was from me.

rhc54:
Rebased

lanl-ompi:
Test FAILed.

Issue #1299, opened 2016-01-13
osc/pt2pt hang in master


ggouaillardet:
@hjelmn can you please have a look at this ?

here is a reproducer

```c
#include <mpi.h>

int main(int argc, char* argv[])
{
    MPI_Win win;
    double * d, a;

    MPI_Init(&argc, &argv);

    a = 0;
    MPI_Win_allocate(sizeof(double), 1, MPI_INFO_NULL, MPI_COMM_SELF, (void *)&d, &win);
    MPI_Win_lock_all(MPI_MODE_NOCHECK, win);
    *d = 0.;

    MPI_Accumulate(&a, 1, MPI_DOUBLE, 0, 0, 1, MPI_DOUBLE, MPI_SUM, win);

    MPI_Win_flush_all(win);
    MPI_Win_unlock_all(win);
    MPI_Win_free(&win);

    MPI_Finalize();

    return 0;
}
```

this can be ran with only one MPI task.

it works fine with ```--mca osc sm``` on both ```v1.10``` and ```master```
but with ```--mca osc pt2pt```, it works fine on ```v1.10``` but it hangs on ```master```

i ran this under the debugger, and ended up writing this patch so ```master``` mimic ```v1.10```.
that being said, i have no idea whether this is correct or not ...

```patch
diff --git a/ompi/mca/osc/pt2pt/osc_pt2pt_sync.c b/ompi/mca/osc/pt2pt/osc_pt2pt_sync.c
index 7e28914..34df3ab 100644
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_sync.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_sync.c
@@ -2,6 +2,8 @@
 /*
  * Copyright (c) 2015      Los Alamos National Security, LLC.  All rights
  *                         reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -17,6 +19,7 @@ static void ompi_osc_pt2pt_sync_constructor (ompi_osc_pt2pt_sync_t *sync)
     sync->type = OMPI_OSC_PT2PT_SYNC_TYPE_NONE;
     sync->eager_send_active = false;
     sync->epoch_active = false;
+    sync->sync.pscw.group = NULL;
     OBJ_CONSTRUCT(&sync->lock, opal_mutex_t);
[...]

PR #1339, opened 2016-02-02
osc/pt2pt: various threading fixes

hjelmn:
This commit fixes several bugs identified by a new multi-threaded RMA
benchmarking suite. The following bugs have been identified and fixed:

 - The code that signaled the actual start of an access epoch changed
   the eager_send_active flag on a synchronization object without
   holding the object's lock. This could cause another thread waiting
   on eager sends to block indefinitely because the entirety of
   ompi_osc_pt2pt_sync_expected could exectute between the check of
   eager_send_active and the conditon wait of
   ompi_osc_pt2pt_sync_wait.

 - The bookkeeping of fragments could get screwed up when performing
   long put/accumulate operations from different threads. This was
   caused by the fragment flush code at the end of both put and
   accumulate. This code was put in place to avoid sending a large
   number of unexpected messages to a peer. To fix the bookkeeping
   issue we now 1) wait for eager sends to be active before stating
   any large isend's, and 2) keep track of the number of large isends
   associated with a fragment. If the number of large isends reaches
   32 the active fragment is flushed.

 - Use atomics to update the large receive/send tag counters. This
   prevents duplicate tags from being used. The tag space has also
   been updated to use the entire 16-bits of the tag space.

These changes should also fix open-mpi/ompi#1299.

### DIFF
--- a/ompi/mca/osc/pt2pt/osc_pt2pt.h
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt.h
@@ -8,7 +8,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2007-2015 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2007-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012-2013 Sandia National Laboratories.  All rights reserved.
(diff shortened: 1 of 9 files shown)


hjelmn:
@ggouaillardet Can you verify #1299 is fixed by this commit?

hjelmn:
Missed one change in this relevant to #1299. Will push update in a bit.

hjelmn:
@jsquyres Mellanox failure shows usnic issue.

jsquyres:
I see the problem.  Fixing...

jsquyres:
@hjelmn master has been fixed.  Rebase and you should be ok.  Sorry about that.

hjelmn:
:bot:retest:

PR #1345, opened 2016-02-06
Topic/sentinel proc name conversion

ggouaillardet:
Fixes open-mpi/ompi#1350

### DIFF
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -369,6 +369,7 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
     return (intptr_t) proc & 0x1;
 }
 
+#if OPAL_SIZEOF_PROCESS_NAME_T == SIZEOF_VOID_P
 /*
  * we assume an ompi_proc_t is at least aligned on two bytes,
  * so if the LSB of a pointer to an ompi_proc_t is 1, we have to handle
@@ -407,6 +408,27 @@ static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel
   name.vpid = vpid;
   return name;
 }
+#elif 4 == SIZEOF_VOID_P
+/*
+ * currently, a sentinel is only made from the current jobid aka OMPI_PROC_MY_NAME->jobid
+ * so we only store the first 31 bits of the vpid
+ */
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+{
+    assert(OMPI_PROC_MY_NAME->jobid == name.jobid);
+    return (uintptr_t)((name.vpid <<1) | 0x1);
+}
+
+static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
+{
+  opal_process_name_t name;
+  name.jobid = OMPI_PROC_MY_NAME->jobid;
+  name.vpid = sentinel >> 1;
+  return name;
+}
+#else
+#error unsupported pointer size
+#endif
 
 END_C_DECLS
 
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -13,7 +13,7 @@
  * Copyright (c) 2007-2012 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2014 Intel, Inc. All rights reserved
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -369,15 +369,14 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
     return (intptr_t) proc & 0x1;
 }
 
-static inline intptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
 {
-  return (*((intptr_t *) &name) << 1) | 0x1;
+  return (*((uintptr_t *) &name) << 1) | 0x1;
 }
 
-static inline opal_process_name_t ompi_proc_sentinel_to_name (intptr_t sentinel)
+static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
 {
   sentinel >>= 1;
-  sentinel &= 0x7FFFFFFFFFFFFFFF;
   return *((opal_process_name_t *) &sentinel);
 }
 
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -369,15 +369,43 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
     return (intptr_t) proc & 0x1;
 }
 
+/*
+ * we assume an ompi_proc_t is at least aligned on two bytes,
+ * so if the LSB of a pointer to an ompi_proc_t is 1, we have to handle
+ * this pointer as a sentinel instead of a pointer.
+ * a sentinel can be seen as an uint64_t with the following format :
+ * - bit  0     : 1
+ * - bits 1-15  : local jobid
+ * - bits 16-31 : job family
+ * - bits 32-63 : vpid
+ */
 static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
 {
-  return (*((uintptr_t *) &name) << 1) | 0x1;
+    uintptr_t tmp, sentinel = 0;
+    /* local jobid must fit in 15 bits */
+    assert(! (OMPI_LOCAL_JOBID(name.jobid) & 0x8000));
+    sentinel |= 0x1;
+    tmp = (uintptr_t)OMPI_LOCAL_JOBID(name.jobid);
+    sentinel |= ((tmp << 1) & 0xfffe);
+    tmp = (uintptr_t)OMPI_JOB_FAMILY(name.jobid);
+    sentinel |= ((tmp << 16) & 0xffff0000);
+    tmp = (uintptr_t)name.vpid;
+    sentinel |= ((tmp << 32) & 0xffffffff00000000);
+    return sentinel;
 }
 
 static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
 {
-  sentinel >>= 1;
-  return *((opal_process_name_t *) &sentinel);
+  opal_process_name_t name;
+  uint32_t local, family;
+  uint32_t vpid;
+  assert(sentinel & 0x1);
+  local = (sentinel >> 1) & 0x7fff;
+  family = (sentinel >> 16) & 0xffff;
+  vpid = (sentinel >> 32) & 0xffffffff;
+  name.jobid = OMPI_CONSTRUCT_JOBID(family,local);
+  name.vpid = vpid;
+  return name;
 }
 
 END_C_DECLS
--- a/orte/util/name_fns.h
+++ b/orte/util/name_fns.h
@@ -97,6 +97,9 @@ ORTE_DECLSPEC char *orte_pretty_print_timing(int64_t secs, int64_t usecs);
 #define ORTE_CONSTRUCT_LOCAL_JOBID(local, job) \
     ( ((local) & 0xffff0000) | ((job) & 0x0000ffff) )
 
+#define ORTE_CONSTRUCT_JOBID(family, local) \
+    ORTE_CONSTRUCT_LOCAL_JOBID(ORTE_CONSTRUCT_JOB_FAMILY(family), local)
+
 /* a macro for identifying that a proc is a daemon */
 #define ORTE_JOBID_IS_DAEMON(n)  \
     !((n) & 0x0000ffff)
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -18,7 +18,7 @@
  * Copyright (c) 2012      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC.
  *                         All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies. All rights reserved.
@@ -699,7 +699,7 @@ static int ompi_comm_split_type_get_part (ompi_group_t *group, int *results, int
         int include = false;
 
         if (ompi_proc_is_sentinel (proc)) {
-            opal_process_name_t proc_name = ompi_proc_sentinel_to_name ((intptr_t) proc);
+            opal_process_name_t proc_name = ompi_proc_sentinel_to_name ((uintptr_t) proc);
 
             u16ptr = &locality;
 
--- a/ompi/dpm/dpm.c
+++ b/ompi/dpm/dpm.c
@@ -169,7 +169,7 @@ int ompi_dpm_connect_accept(ompi_communicator_t *comm, int root,
         for (i=0; i < size; i++) {
             opal_process_name_t proc_name;
             if (ompi_proc_is_sentinel (proc_list[i])) {
-                proc_name = ompi_proc_sentinel_to_name ((intptr_t) proc_list[i]);
+                proc_name = ompi_proc_sentinel_to_name ((uintptr_t) proc_list[i]);
             } else {
                 proc_name = proc_list[i]->super.proc_name;
             }
@@ -532,7 +532,7 @@ static int construct_peers(ompi_group_t *group, opal_list_t *peers)
             return OMPI_ERR_NOT_FOUND;
         }
         if (ompi_proc_is_sentinel (proct)) {
-            proc_name = ompi_proc_sentinel_to_name ((intptr_t)proct);
+            proc_name = ompi_proc_sentinel_to_name ((uintptr_t)proct);
         } else {
             proc_name = proct->super.proc_name;
         }
--- a/ompi/group/group.h
+++ b/ompi/group/group.h
@@ -16,6 +16,8 @@
  * Copyright (c) 2012      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -352,7 +354,7 @@ static inline struct ompi_proc_t *ompi_group_dense_lookup (ompi_group_t *group, 
 
         /* replace sentinel value with an actual ompi_proc_t */
         ompi_proc_t *real_proc =
-            (ompi_proc_t *) ompi_proc_for_name (ompi_proc_sentinel_to_name ((intptr_t) proc));
+            (ompi_proc_t *) ompi_proc_for_name (ompi_proc_sentinel_to_name ((uintptr_t) proc));
 
         if (opal_atomic_cmpset_ptr (group->grp_proc_pointers + peer_id, proc, real_proc)) {
             OBJ_RETAIN(real_proc);
--- a/ompi/group/group_plist.c
+++ b/ompi/group/group_plist.c
@@ -14,6 +14,8 @@
  * Copyright (c) 2007      Cisco Systems, Inc. All rights reserved.
  * Copyright (c) 2013-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -61,7 +63,7 @@ static struct ompi_proc_t *ompi_group_dense_lookup_raw (ompi_group_t *group, con
 {
     if (OPAL_UNLIKELY(ompi_proc_is_sentinel (group->grp_proc_pointers[peer_id]))) {
         ompi_proc_t *proc =
-            (ompi_proc_t *) ompi_proc_lookup (ompi_proc_sentinel_to_name ((intptr_t) group->grp_proc_pointers[peer_id]));
+            (ompi_proc_t *) ompi_proc_lookup (ompi_proc_sentinel_to_name ((uintptr_t) group->grp_proc_pointers[peer_id]));
         if (NULL != proc) {
             /* replace sentinel value with an actual ompi_proc_t */
             group->grp_proc_pointers[peer_id] = proc;
--- a/ompi/mca/pml/monitoring/pml_monitoring.c
+++ b/ompi/mca/pml/monitoring/pml_monitoring.c
@@ -4,6 +4,8 @@
  *                         reserved.
  * Copyright (c) 2013-2015 Inria.  All rights reserved.
  * Copyright (c) 2015      Bull SAS.  All rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -77,7 +79,7 @@ int mca_pml_monitoring_add_procs(struct ompi_proc_t **procs,
                 my_rank = i;
             /* Extract the peer procname from the procs array */
             if( ompi_proc_is_sentinel(procs[i]) ) {
-                tmp = ompi_proc_sentinel_to_name((intptr_t)procs[i]);
+                tmp = ompi_proc_sentinel_to_name((uintptr_t)procs[i]);
             } else {
                 tmp = procs[i]->super.proc_name;
             }
--- a/ompi/mca/rte/orte/rte_orte.h
+++ b/ompi/mca/rte/orte/rte_orte.h
@@ -3,7 +3,7 @@
  *                         All rights reserved.
  * Copyright (c) 2013-2015 Intel, Inc. All rights reserved
  * Copyright (c) 2014      Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015      Intel, Inc. All rights reserved.
  * $COPYRIGHT$
@@ -59,6 +59,10 @@ typedef orte_ns_cmp_bitmask_t ompi_rte_cmp_bitmask_t;
 #define OMPI_RTE_CMP_JOBID  ORTE_NS_CMP_JOBID
 #define OMPI_RTE_CMP_VPID   ORTE_NS_CMP_VPID
 #define OMPI_RTE_CMP_ALL    ORTE_NS_CMP_ALL
+#define OMPI_LOCAL_JOBID(jobid) ORTE_LOCAL_JOBID(jobid)
+#define OMPI_JOB_FAMILY(jobid)  ORTE_JOB_FAMILY(jobid)
+#define OMPI_CONSTRUCT_JOBID(family,local) ORTE_CONSTRUCT_JOBID(family,local)
+
 /* This is the DSS tag to serialize a proc name */
 #define OMPI_NAME ORTE_NAME
 #define OMPI_PROCESS_NAME_HTON ORTE_PROCESS_NAME_HTON
--- a/ompi/proc/proc.c
+++ b/ompi/proc/proc.c
@@ -14,7 +14,7 @@
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2015 Intel, Inc. All rights reserved
- * Copyright (c) 2014-2015 Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies. All rights reserved.
  *
@@ -642,7 +642,7 @@ ompi_proc_pack(ompi_proc_t **proclist, int proclistsize,
         ompi_proc_t *proc = proclist[i];
 
         if (ompi_proc_is_sentinel (proc)) {
-            proc = ompi_proc_for_name_nolock (ompi_proc_sentinel_to_name ((intptr_t) proc));
+            proc = ompi_proc_for_name_nolock (ompi_proc_sentinel_to_name ((uintptr_t) proc));
         }
 
         /* send proc name */
--- a/opal/dss/dss_types.h
+++ b/opal/dss/dss_types.h
@@ -14,7 +14,7 @@
  * Copyright (c) 2012-2013 Los Alamos National Security, Inc. All rights
  *                         reserved.
  * Copyright (c) 2014      Intel, Inc. All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -47,6 +47,7 @@ typedef struct {
     opal_jobid_t jobid;
     opal_vpid_t vpid;
 } opal_process_name_t;
+#define OPAL_SIZEOF_PROCESS_NAME_T 8
 
 BEGIN_C_DECLS
 


orte/util/name_fns.h
```diff
@@ -97,6 +97,9 @@ ORTE_DECLSPEC char *orte_pretty_print_timing(int64_t secs, int64_t usecs);
 #define ORTE_CONSTRUCT_LOCAL_JOBID(local, job) \
     ( ((local) & 0xffff0000) | ((job) & 0x0000ffff) )
 
+#define ORTE_CONSTRUCT_JOBID(family, local) \
+    ORTE_CONSTRUCT_LOCAL_JOBID(ORTE_CONSTRUCT_JOB_FAMILY(family), local)
+
```
rhc54: Does this really work if family and local are just 16-bit values? My macros assumed they were 32-bit.

ompi/proc/proc.h
```diff
@@ -369,16 +369,42 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
...
-  return (*((intptr_t *) &name) << 1) | 0x1;
+    uintptr_t tmp, sentinel = 0x1;
+    /* local jobid must fit in 15 bits */
+    assert(! (ORTE_LOCAL_JOBID(name.jobid) & 0x8000));
+    tmp = (uintptr_t)ORTE_LOCAL_JOBID(name.jobid);
+    sentinel |= ((tmp << 1) & 0xfffe);
+    tmp = (uintptr_t)ORTE_JOB_FAMILY(name.jobid);
+    sentinel |= ((tmp << 16) & 0xffff0000);
+    tmp = (uintptr_t)name.vpid;
+    sentinel |= ((tmp << 32) & 0xffffffff00000000);
```
rhc54: Doesn't this still assume that uintptr_t is 64-bits?

ompi/proc/proc.h
```diff
@@ -369,16 +369,42 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
...
+ * a sentinel can be seen as a 64 bits array with the following format :
+ * - bit  0     : 1
+ * - bits 1-15  : local jobid
+ * - bits 16-31 : job family
+ * - bits 32-63 : vpid
+ */
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
 {
-  return (*((intptr_t *) &name) << 1) | 0x1;
+    uintptr_t tmp, sentinel = 0x1;
```
rhc54: Just wondering: does every compiler understand that 0x1 means "put a 1 in the left-most bit" as opposed to "set the value of this variable to 1"?

ompi/proc/proc.h
```diff
@@ -369,16 +369,42 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
...
-  sentinel &= 0x7FFFFFFFFFFFFFFF;
-  return *((opal_process_name_t *) &sentinel);
+  opal_process_name_t name;
+  uint16_t local, family;
+  uint32_t vpid;
+  assert(sentinel & 0x1);
+  local = (sentinel >> 1) & 0x7fff;
+  family = (sentinel >> 16) & 0xffff;
+  vpid = (sentinel >> 32) & 0xffffffff;
+  name.jobid = ORTE_CONSTRUCT_JOBID(family,local);
```
rhc54: Technically, to maintain the abstraction, we should call this OMPI_CONSTRUCT_JOBID and equate that to the ORTE macro in the ompi/mca/rte/orte component.

ompi/proc/proc.h
```diff
@@ -369,16 +369,42 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
...
+ * - bits 32-63 : vpid
+ */
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
 {
-  return (*((intptr_t *) &name) << 1) | 0x1;
+    uintptr_t tmp, sentinel = 0x1;
+    /* local jobid must fit in 15 bits */
+    assert(! (ORTE_LOCAL_JOBID(name.jobid) & 0x8000));
+    tmp = (uintptr_t)ORTE_LOCAL_JOBID(name.jobid);
+    sentinel |= ((tmp << 1) & 0xfffe);
```
rhc54: I've been trying to understand this logic, and I'm not sure I believe that the "1" you want to set for the sentinel is winding up in the correct place. Isn't it supposed to be in the MSB position?

ompi/proc/proc.h
```diff
@@ -369,16 +369,42 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
...
-static inline intptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+/*
+ * we assume an ompi_proc_t is at least aligned on two bytes,
+ * so if the LSB of a pointer to an ompi_proc_t is 1, we have to handle
+ * this pointer as a sentinel instead of a pointer.
+ * a sentinel can be seen as a 64 bits array with the following format :
+ * - bit  0     : 1
+ * - bits 1-15  : local jobid
+ * - bits 16-31 : job family
+ * - bits 32-63 : vpid
```
rhc54: FWIW: that memory map isn't quite correct. I'm not sure how you are numbering bits, but assuming that they go from 0 as the LSB to 63 as the MSB, then the layout is:

```
0-31 : vpid
32-47:  local jobid
48-63:  job family
```
So you want to set your sentinel in bit 47, not bit 0. Am I missing something?

ggouaillardet: Sentinel bit is the LSB, aka bit zero.
see my comment in the code above
an ompi_proc_t* can either be a genuine pointer, or a sentinel.
that would require 65 bits.
the trick is pointer are always aligned on two bytes, so the LSB of a pointer is always zero, and we can use this bit as the sentinel bit.
if it is one, it means this is not a pointer but a sentinel.
that still means we have to "pack" the opal_process_name_t into the remaining 63 bits, and we decided to drop the MSB of the local jobid.

note that cannot work "as is" 32 bits.

my best bet is to simply disable cutoff at runtime on 32 bits arch

an other option to keep memory usage low is to use the sentinel bit as an opal_process_name_t* instead of a "packed" opal_process_name_t, but I am not sure it is worth it.

I will likely use jobid and family as uint32_t as you mentionned, and fix the abstraction 

makes more sense now ?

ggouaillardet: fwiw, @hjelmn initially used the MSB as the sentinel bit, assuming the MSB of an ompi_proc_t address was always zero. this assumption was proven wrong on a linux/sparc Fujitsu cluster.

rhc54: I agree with using the MSB of the local jobid. What I'm confused by is that your documented bitmap is incorrect. The local jobid is in the middle of the object, not at the edge as you show.

Nathan mistakenly used the MSB of the entire structure instead of the local jobid. Your bitmap moves it to the LSB of the entire structure, which is the LSB of the vpid, not the local jobid.

ggouaillardet: I am confused too ...
I guess the way we number bits differ

```c
    unitptr_t tmp = (uintptr_t)ORTE_LOCAL_JOBID(name.jobid);
     sentinel |= ((tmp << 1) & 0xfffe);
```
where does the local jobid end into ?
- bits 1:15 as I documented ?
- bits 48:62 ?
and btw, is this endian dependent ?

ompi/proc/proc.h
bosilca: This is mandated by the C standard: prolog of 0x is for a hexadecimal number, prolog of 0 for an octal.

ompi/proc/proc.h
bosilca: Short [code](https://gist.github.com/bosilca/64843961946319497da7) to show the contents of the different fields. Looking at the bit position of different fields we have:
```
Operation              Raw Memory          Little endian (uint64_t)
Alterning bits  0:15   ffff000000000000                ffff
Alterning bits 15:31   0000ffff00000000            ffff0000
Alterning bits 32:63   00000000ffffffff    ffffffff00000000
Setting jobid          ffff000000000000                ffff
Setting family         0000ffff00000000            ffff0000
Setting vpid           00000000ffffffff    ffffffff00000000
```

Because opal_process_name_t is a structure, the location of the 2 fields (jobid, vpid) will never change, their offsets will always be 0 for the jobid and 4 for the vpid. Thus to your question regarding the endianness, the answer is no, you always flip the same logical bit.

rhc54: Agreed. I think the confusion here is over all the bit logic in the proposed code. What I had proposed is that we take the 2^15th bit (the MSB) of the local jobid as the sentinel. Thus, all you have to do is check the sign of the ORTE_LOCAL_JOBID. If it is positive, then it is the sentinel isn't set - if it is negative, then the sentinel is set. There is no reason to be bit-shifting things.

Likewise, when you create the sentinel, you create the process name with the negative of the the local jobid. Then we adjust the orte_create_jobid code so it only uses values up to 2^15 - 1.

Does that help?

ggouaillardet: it helped me understanding where the confusion is coming from :-)
a sentinel is basically an ```opal_process_name_t``` packed/encoded/hidden inside an ```ompi_proc_t *```

here is a snippet of how sentinel are used, from ```ompi/dpm/dpm.c```
```c
        if (ompi_proc_is_sentinel (proct)) {
            proc_name = ompi_proc_sentinel_to_name ((uintptr_t)proct);
        } else {
            proc_name = proct->super.proc_name;
        }
```
```proct``` has type ```ompi_proc_t *```, but it can either be a genuine pointer that can be dereferenced, or a sentinel.
so at first, we must figure out whether this 64 bit thing is a pointer or a sentinel.
since an ```ompi_proc_t``` object is aligned on at least two bytes, it is safe to use the LSB of the 64 bit thing. (if it is zero, then it is a pointer, if it is one, then it cannot be a pointer and it is a sentinel).
if you simply use the MSB of the local jobid, you cannot tell whether the 64 bit thing is a pointer or a sentinel.

using the LSB of the 64 bit thing does require some bit shifting, but we just can't avoid it.

does that make sense now ?

Comment (rhc54) [inline ompi/proc/proc.h]: It sounds like a really fragile idea to me, and unnecessarily breaks 32-bit environments. Why not just have it always be a pointer to at least a process name? In other words, always create the opal_process_name_t portion of the ompi_proc_t, and then just embed that in an ompi_proc_t when you fill it in? I understand it costs us a touch more memory, but at least it wouldn't be so vulnerable to any future changes in opal_process_name_t.

We made the mistake in the past of trying to play bit games with process names, and got burned for it. Seems like we might be heading down that path again.

Anyway, that's my $0.02.

Comment (rhc54) [inline ompi/proc/proc.h]: Let me try to be more concrete, and offer an a less fragile alternative. The issue here is that we are tying this entire logic to the rather arbitrary fact that an opal_process_name_t currently has the same size as a 64-bit pointer. This is definitely _not_ guaranteed to be true in the future. Past proposals have extended that size, and there are even some discussions about decreasing it to 32-bits. So making the assumption that an opal_process_name_t is 64-bits in size is a bad one.

The problem is essentially caused by us putting all the ompi_proc_t's in a single array, regardless of what job they come from. So you can't use the index of the array as representing the vpid of a specific process in a given job. This is easily rectified in a more flexible manner by simply changing the ompi_proc_t array into an array of arrays.

In other words, create an array of ompi_proc_t's for each jobid with which we are connected, and have the "header" for the array include the jobid. The overwhelming majority of applications never cross-connect between jobs or call comm_spawn, and so we will only rarely see more than one such array anyway. Now each location in the array can be uniquely correlated to a vpid as the vpid is just the index of that location.

You no longer require a sentinel in this situation as you can just use NULL to indicate that we haven't yet defined the ompi_proc_t for that vpid. If the location is NULL, then you construct the process name using the jobid in the array's header and the index of that location.

This more cleanly solves the problem, doesn't break 32-bit environments, and future-proofs us. It doesn't hurt us in any way as the communicators all just carry pointers to their participating ompi_proc_t's - there is no requirement that those objects occupy the same array.

Make sense?

Comment (hjelmn) [inline ompi/proc/proc.h]: The sentinel was done the way it is because 1) the size of a process name is right now the same size a pointer (if it is larger than a pointer then there would be an issue), and 2) it was a general way to update the code without changing ompi_group_t. I also do not care about 32-bit systems. We haven't had one here ~ 10 years. Loosing a bit in 64 is not that big a deal but I can see why it causes problems with 32-bit.

It will take a rewrite of the ompi_group_t code to do as you suggest but I do not think that will happen in the 2.0.x time frame. It will take some time to write then it needs to soak in master. I do not want to hold up 2.0.0 for this. I am really surprised the issues @ggouaillardet identified did not come up during testing of the dynamic add_procs code (which is now over 5 months old). If we change ompi_group_t we will have to be very careful how we proceed to avoid increasing the memory cost or increasing the cost of ompi_comm_peer_lookup.

Comment (rhc54) [inline ompi/proc/proc.h]: Just to be clear: I'm not arguing for holding up 2.0.0 for this. I'm saying that the built-in assumption of opal_process_name_t being the size of a pointer is not a good one looking into the future, and basing the logic on that equality isn't something we should be doing.

So the question shouldn't be _if_ we rewrite it, but _when_ do we do it and what the final solution should be.

Comment (hjelmn) [inline ompi/proc/proc.h]: I have a good workaround for 2.0.0. Disable dynamic add_procs on 32-bit platforms.

Comment (hjelmn) [inline ompi/proc/proc.h]: @rhc54 Ok, just making sure. I agree we need to update ompi_group_t. I have some ideas I can start implementing in the next month. Can target 2.1.x or 2.2.x for the changes.

Comment (rhc54) [inline ompi/proc/proc.h]: @hjelmn Thanks!

Comment (bosilca) [inline ompi/proc/proc.h]: If we use the LSB of the uint64_t, and shift the vpid to the correct place (so that we logically use the MSB of the jobid), I dont see why this will be an issue if we decide to make the orte_process_name_t a 32 bits value. Why are we trying to increase the opportunities for cache pollution by moving to a design that will require at least 2 extra indirections (the array of arrays solution).

There is a much simpler solution: dissociate the opal_process_name_t from the orte_process_name_t (in a mechanism similar to the lazy naming resolution for shared libraries). Make opal_process_name_t a simple int (32 to start with) that will be the index to an internal array that we alter every time the process known universe changes. This solution might be a little confusing as the index does not represent a global unique naming, but the ORTE name it points to does, but it has the benefit to let us (OPAL and OMPI) be totally independent on how ORTE wants to name it's processes.

Comment (rhc54) [inline ompi/proc/proc.h]: As I said, I'm not advocating any particular solution - I'm only pointing out that hinging the scheme on a particular size of the process name struct, or on the idea that we can arbitrarily throw away one of its bits without repercussions, is fragile. If you have a better scheme than the one I suggested, that's fine with me.

Comment (bosilca) [inline ompi/proc/proc.h]: It has at least the advantage of making OPAL + OMPI totally independent on how the RTE names processes...

**Comment by ggouaillardet** [comment]
@hjelmn the issue only occurs on big endian platforms (and i am only half surprised cutoff was not tested on this arch)

@bosilca 
```
If we use the LSB of the uint64_t, and shift the vpid to the correct place (so that we logically use the MSB of the jobid), I dont see why this will be an issue if we decide to make the orte_process_name_t a 32 bits value. 
```
the "right" place is endian dependant, that means we would have
```c
struct {
#if WORDS_BIGENDIAN
   opal_jobid_t jobid;
   opal_vpid_t vpid;
#else
   opal_vpid_t vpid;
   opal_jobid_t jobid;
#endif
} opal_process_name_t;
```
and that would require to change the way ```opal_process_name_t``` is converted to support heterogeneous clusters.
/* i remember i suggested that in the past, but we decided this was too much of a hack and dropped that idea */

@rhc54 could you please double check this is PR is good to go ?
(i will squeeze the last two commits and commit it then)


here is a patch to be applied on top of this PR to automatically disable cutoff on non 64 bits arch
```patch
commit 0bee2a570c4ada11cfca9f35268c0a20a5835e74
Date:   Mon Feb 8 10:48:26 2016 +0900

    sentinel: disable mpi_add_procs_cutoff on non 64 bits arch
    - add have_mpi_add_procs_cutoff MCA read-only parameter
    - make mpi_add_procs_cutoff readonly and MAX_UINT32 on non 64 bits arch
    - add an extra check to ensure sizeof(opal_process_name_t) is 8

diff --git a/ompi/proc/proc.h b/ompi/proc/proc.h
index eac16cf..d8fe2f9 100644
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -361,6 +361,7 @@ OMPI_DECLSPEC opal_proc_t *ompi_proc_for_name (const opal_process_name_t proc_na
 
 OMPI_DECLSPEC opal_proc_t *ompi_proc_lookup (const opal_process_name_t proc_name);
 
+#if OPAL_SIZEOF_PROCESS_NAME_T == SIZEOF_VOID_P
 /**
  * Check if an ompi_proc_t is a sentinel
  */
@@ -407,6 +408,23 @@ static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel
   name.vpid = vpid;
   return name;
 }
+#else
+static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
+{
+    return false;
+}
+
+/* these functions will never be called */
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+{
+    return (uintptr_t)-1;
+}
+
+static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
+{
+  return opal_name_invalid;
+}
+#endif
 
 END_C_DECLS
 
diff --git a/ompi/runtime/ompi_mpi_params.c b/ompi/runtime/ompi_mpi_params.c
index 7b32eee..2bf206a 100644
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -17,6 +17,8 @@
  * Copyright (c) 2013-2014 Intel, Inc. All rights reserved
  * Copyright (c) 2015      Mellanox Technologies, Inc.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -64,6 +66,7 @@ int ompi_mpi_event_tick_rate = -1;
 char *ompi_mpi_show_mca_params_string = NULL;
 bool ompi_mpi_have_sparse_group_storage = !!(OMPI_GROUP_SPARSE);
 bool ompi_mpi_preconnect_mpi = false;
+bool ompi_have_add_procs_cutoff = false;
 uint32_t ompi_add_procs_cutoff = 1024;
 bool ompi_mpi_dynamics_enabled = true;
 
@@ -263,13 +266,27 @@ int ompi_mpi_register_params(void)
         ompi_rte_abort(1, NULL);
     }
 
-    ompi_add_procs_cutoff = 1024;
+    /* Procs cutoff support */
+    ompi_have_add_procs_cutoff = (OPAL_SIZEOF_PROCESS_NAME_T == SIZEOF_VOID_P);
+    (void) mca_base_var_register("ompi", "mpi", NULL, "have_add_procs_cutoff",
+                                 "Whether this Open MPI installation supports add_procs_cutoff",
+                                 MCA_BASE_VAR_TYPE_BOOL, NULL, 0,
+                                 MCA_BASE_VAR_FLAG_DEFAULT_ONLY,
+                                 MCA_BASE_VAR_SCOPE_CONSTANT,
+                                 &ompi_have_add_procs_cutoff);
+
+    if (!ompi_have_add_procs_cutoff) {
+        ompi_add_procs_cutoff = UINT32_MAX;
+    }
     (void) mca_base_var_register ("ompi", "mpi", NULL, "add_procs_cutoff",
                                   "communication performance at the cost of memory usage "
-                                  "(default: 1024)", MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL,
-                                  0, 0, OPAL_INFO_LVL_3, MCA_BASE_VAR_SCOPE_LOCAL,
+                                  "(default: 1024)", MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL, 0,
+                                  ompi_have_add_procs_cutoff?0:MCA_BASE_VAR_FLAG_DEFAULT_ONLY,
+                                  OPAL_INFO_LVL_3,
+                                  ompi_have_add_procs_cutoff?MCA_BASE_VAR_SCOPE_LOCAL:MCA_BASE_VAR_SCOPE_CONSTANT,
                                   &ompi_add_procs_cutoff);
 
     ompi_mpi_dynamics_enabled = true;
diff --git a/opal/dss/dss_types.h b/opal/dss/dss_types.h
index 8c1bad9..c261223 100644
--- a/opal/dss/dss_types.h
+++ b/opal/dss/dss_types.h
@@ -14,7 +14,7 @@
  * Copyright (c) 2012-2013 Los Alamos National Security, Inc. All rights
  * Copyright (c) 2014      Intel, Inc. All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -47,6 +47,7 @@ typedef struct {
     opal_jobid_t jobid;
     opal_vpid_t vpid;
 } opal_process_name_t;
+#define OPAL_SIZEOF_PROCESS_NAME_T 8
 
 BEGIN_C_DECLS
 
diff --git a/test/dss/dss_size.c b/test/dss/dss_size.c
index c128ec7..54f0919 100644
--- a/test/dss/dss_size.c
+++ b/test/dss/dss_size.c
@@ -9,6 +9,8 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -42,6 +44,7 @@ static bool test9(void);        /* verify composite (multiple types and element
 static bool test11(void);        /* verify size_t */
 static bool test12(void);        /* verify pid_t */
 static bool test13(void);        /* verify pid_t */
+static bool test14(void);        /* verify opal_process_name_t */
 
 FILE *test_out;
 
@@ -150,6 +153,14 @@ int main (int argc, char* argv[])
         fprintf(test_out, "opal_dss test13 failed\n");
     }
 
+    fprintf(test_out, "executing test14\n");
+    if (test14()) {
+        fprintf(test_out, "Test14 succeeded\n");
+    }
+    else {
+        fprintf(test_out, "opal_dss test14 failed\n");
+    }
+
     fclose(test_out);
 
     opal_finalize();
@@ -624,3 +635,14 @@ static bool test13(void)
     return (true);
 }
 
+static bool test14(void)
+{
+    if (sizeof(opal_process_name_t) != OPAL_SIZEOF_PROCESS_NAME) {
+        fprintf(test_out, "opal_process_name_t does not have expected size (%d != %d)\n",
+                sizeof(opal_process_name_t), OPAL_SIZEOF_PROCESS_NAME);
+        return(false);
+    }
+
+    return (true);
+}
+
```

and here is an other one that does support non 64 bits arch, at the expense of a
```calloc(sizeof(opal_process_name_t), <MPI_COMM_WORLD size>```
(and if the ```calloc``` is acceptable, then that can be used on 64 bits arch too, which
will make the code future proof. note i still assume ```ompi_proc_t``` and opal_process_name_t``` are aligned on at least two bytes, so the LSB of the pointer can be used as the sentinel bit.

```patch
commit 1bcc02c961018e6f0ae82281a7a2caf62de11d78
Date:   Mon Feb 8 10:48:26 2016 +0900

    sentinel: fix non 64 bits arch
    on a non 64 bits arch, a sentinel is an opal_process_name_t * with its LSB forced to one

diff --git a/ompi/communicator/comm_init.c b/ompi/communicator/comm_init.c
index b2200bd..aa30d61 100644
--- a/ompi/communicator/comm_init.c
+++ b/ompi/communicator/comm_init.c
@@ -107,6 +107,8 @@ int ompi_comm_init(void)
     group->grp_proc_pointers = (ompi_proc_t **) calloc (size, sizeof (ompi_proc_t *));
     group->grp_proc_count = size;
 
+    ompi_proc_sentinel_init(size);
+
     for (size_t i = 0 ; i < size ; ++i) {
         opal_process_name_t name = {.vpid = i, .jobid = OMPI_PROC_MY_NAME->jobid};
         /* look for existing ompi_proc_t that matches this name */
@@ -328,6 +330,7 @@ int ompi_comm_finalize(void)
         }
     }
 
+    ompi_proc_sentinel_finalize();
 
     OBJ_DESTRUCT (&ompi_mpi_communicators);
     OBJ_DESTRUCT (&ompi_comm_f_to_c_table);
diff --git a/ompi/proc/proc.h b/ompi/proc/proc.h
index eac16cf..3d3230c 100644
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -361,6 +361,11 @@ OMPI_DECLSPEC opal_proc_t *ompi_proc_for_name (const opal_process_name_t proc_na
 
 OMPI_DECLSPEC opal_proc_t *ompi_proc_lookup (const opal_process_name_t proc_name);
 
+#if OPAL_SIZEOF_PROCESS_NAME_T == SIZEOF_VOID_P
+
+#define ompi_proc_sentinel_init(s)
+#define ompi_proc_sentinel_finalize()
+
 /**
  * Check if an ompi_proc_t is a sentinel
  */
@@ -407,6 +412,37 @@ static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel
   name.vpid = vpid;
   return name;
 }
+#else
+
+static opal_process_name_t * proc_names;
+
+static inline void ompi_proc_sentinel_init(int size) {
+    proc_names = (opal_process_name_t *)calloc(sizeof(opal_process_name_t), size);
+}
+
+static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
+{
+    return (uintptr_t)proc & 0x1;
+}
+
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+{
+    proc_names[name.vpid] = name;
+    return (uintptr_t)(proc_names+name.vpid) | 0x1;
+}
+
+static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
+{
+  opal_process_name_t * name;
+  name = (opal_process_name_t *)((uintptr_t)sentinel & (~0x1));
+  return *name;
+}
+
+static inline void ompi_proc_sentinel_finalize(void)
+{
+  free(proc_names);
+}
+#endif
 
 END_C_DECLS
 
diff --git a/ompi/runtime/ompi_mpi_params.c b/ompi/runtime/ompi_mpi_params.c
index 7b32eee..34abb5f 100644
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -17,6 +17,8 @@
  * Copyright (c) 2013-2014 Intel, Inc. All rights reserved
  * Copyright (c) 2015      Mellanox Technologies, Inc.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -64,6 +66,7 @@ int ompi_mpi_event_tick_rate = -1;
 char *ompi_mpi_show_mca_params_string = NULL;
 bool ompi_mpi_have_sparse_group_storage = !!(OMPI_GROUP_SPARSE);
 bool ompi_mpi_preconnect_mpi = false;
+bool ompi_have_add_procs_cutoff = false;
 uint32_t ompi_add_procs_cutoff = 1024;
 bool ompi_mpi_dynamics_enabled = true;
 
@@ -263,13 +266,14 @@ int ompi_mpi_register_params(void)
         ompi_rte_abort(1, NULL);
     }
 
-    ompi_add_procs_cutoff = 1024;
     (void) mca_base_var_register ("ompi", "mpi", NULL, "add_procs_cutoff",
                                   "communication performance at the cost of memory usage "
-                                  "(default: 1024)", MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL,
-                                  0, 0, OPAL_INFO_LVL_3, MCA_BASE_VAR_SCOPE_LOCAL,
+                                  "(default: 1024)", MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL, 0,
+                                  0,
+                                  OPAL_INFO_LVL_3,
+                                  MCA_BASE_VAR_SCOPE_LOCAL,
                                   &ompi_add_procs_cutoff);
 
     ompi_mpi_dynamics_enabled = true;
diff --git a/opal/dss/dss_types.h b/opal/dss/dss_types.h
index 8c1bad9..c261223 100644
--- a/opal/dss/dss_types.h
+++ b/opal/dss/dss_types.h
@@ -14,7 +14,7 @@
  * Copyright (c) 2012-2013 Los Alamos National Security, Inc. All rights
  * Copyright (c) 2014      Intel, Inc. All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -47,6 +47,7 @@ typedef struct {
     opal_jobid_t jobid;
     opal_vpid_t vpid;
 } opal_process_name_t;
+#define OPAL_SIZEOF_PROCESS_NAME_T 8
 
 BEGIN_C_DECLS
 
diff --git a/test/dss/dss_size.c b/test/dss/dss_size.c
index c128ec7..54f0919 100644
--- a/test/dss/dss_size.c
+++ b/test/dss/dss_size.c
@@ -9,6 +9,8 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -42,6 +44,7 @@ static bool test9(void);        /* verify composite (multiple types and element
 static bool test11(void);        /* verify size_t */
 static bool test12(void);        /* verify pid_t */
 static bool test13(void);        /* verify pid_t */
+static bool test14(void);        /* verify opal_process_name_t */
 
 FILE *test_out;
 
@@ -150,6 +153,14 @@ int main (int argc, char* argv[])
         fprintf(test_out, "opal_dss test13 failed\n");
     }
 
+    fprintf(test_out, "executing test14\n");
+    if (test14()) {
+        fprintf(test_out, "Test14 succeeded\n");
+    }
+    else {
+        fprintf(test_out, "opal_dss test14 failed\n");
+    }
+
     fclose(test_out);
 
     opal_finalize();
@@ -624,3 +635,14 @@ static bool test13(void)
     return (true);
 }
 
+static bool test14(void)
+{
+    if (sizeof(opal_process_name_t) != OPAL_SIZEOF_PROCESS_NAME) {
+        fprintf(test_out, "opal_process_name_t does not have expected size (%d != %d)\n",
+                sizeof(opal_process_name_t), OPAL_SIZEOF_PROCESS_NAME);
+        return(false);
+    }
+
+    return (true);
+}
+
```

rhc54:
Just to clarify: I have no particular solution I'm advocating, but I do have couple of requirements on the solution itself

* it must support 32-bit environments. I know this community is focused on supercomputers, but there are folks trying to use OMPI in embedded environments (e.g., IoT and smart-cars) that are overwhelming based on 32-bit processors. Dynamic add-procs enables some other operating modes of use to them. We shouldn't abandon them until alternative solutions have been exhausted.

* it cannot be based on the peculiarities of the current ORTE/OPAL process name. We are looking at ways of saving memory in those areas as well, and the process name is certainly one target. There is no future guarantee as to the size of the name, or the ability to strip it of bits without breaking things.  Let's not tie our hands.

HTH
Ralph

hjelmn:
I just thought of something we can do for 2.x. Since we always create the procs for dynamic jobs we could 1) add the jobid to the ompi_group_t structure, and 2) make the sentinel (rank << 1) | 1. That would solve the issue.

bosilca:
@hjelmn what if we have processes from multiple jobid on the same group?

ggouaillardet:
This is not possible yet
ompi_proc_name_to_sentinel is only called in ompi_comm_init and with the current (aka MPI_COMM_WORLD) jobid

bosilca:
@ggouaillardet assuming that your comment was an answer to my question, let me clarify that this is not an option. We should be able to have ompi_group_t with processes from multiple jobid.

ggouaillardet:
@bosilca yes, that was an answer.
today, sentinel can only point to a task of mpi_comm_world, and all these tasks have the same jobid.
a group is an array of pointers, and some of them might be sentinels.
so it is possible to have procs with different job ids in the same group, what @hjelmn and I meant is only pointers to procs from mpi_comm_world can be replaced with sentinels.
that might change in the future (e.g. not only mpi_comm_init will create sentinels with ompi_proc_name_to_sentinel), but as of now, this is a valid optimization.

ggouaillardet:
@bosilca let me rephrase my initial reply.
it is possible to have processes from multiple jobid in the same group.
currently, sentinels are only for processes that belongs to mpi_comm_world, and these processes have the same jobid.

bosilca:
So the proposal here is that when we create a group we replace all the "not yet resolved" processes from MPI_COMM_WORLD by their corresponding rank in MPI_COMM_WORLD (which will allow to find them in the global proc array) shifted to the left by one and with the LSB set. This is a nice solution, that should be minimally intrusive on the group operations. 

But then why do you need to have an jobid associated with the ompi_group_t ?

hjelmn:
@bosilca Good point. Was just throwing an idea out there. Don't really need the job id since it can be gotten from OMPI_PROC_MY_NAME.

hjelmn:
I can throw together a patch that does this. Will do that tomorrow.

ggouaillardet:
from ```ompi_mpi_comm_init```

```c
/* Setup MPI_COMM_WORLD */
    OBJ_CONSTRUCT(&ompi_mpi_comm_world, ompi_communicator_t);
    assert(ompi_mpi_comm_world.comm.c_f_to_c_index == 0);
    group = OBJ_NEW(ompi_group_t);

    size = ompi_process_info.num_procs;
    group->grp_proc_pointers = (ompi_proc_t **) calloc (size, sizeof (ompi_proc_t *));
    group->grp_proc_count = size;

    for (size_t i = 0 ; i < size ; ++i) {
        opal_process_name_t name = {.vpid = i, .jobid = OMPI_PROC_MY_NAME->jobid};
        /* look for existing ompi_proc_t that matches this name */
        group->grp_proc_pointers[i] = (ompi_proc_t *) ompi_proc_lookup (name);
        if (NULL == group->grp_proc_pointers[i]) {
            /* set sentinel value */
            group->grp_proc_pointers[i] = (ompi_proc_t *) ompi_proc_name_to_sentinel (name);
        } else {
            OBJ_RETAIN (group->grp_proc_pointers[i]);
        }
    }

    OMPI_GROUP_SET_INTRINSIC (group);
    OMPI_GROUP_SET_DENSE (group);
    ompi_set_group_rank(group, ompi_proc_local());

    ompi_mpi_comm_world.comm.c_contextid    = 0;
    ompi_mpi_comm_world.comm.c_id_start_index = 4;
    ompi_mpi_comm_world.comm.c_id_available = 4;
    ompi_mpi_comm_world.comm.c_my_rank      = group->grp_my_rank;
    ompi_mpi_comm_world.comm.c_local_group  = group;
    ompi_mpi_comm_world.comm.c_remote_group = group;
```

```ompi_proc_name_to_sentinel``` is only invoked here (at least for now)

now, if we ```MPI_Comm_spawn``` and ```MPI_Intercomm_merge``` we will end up creating a group with procs from ```MPI_COMM_WORLD``` (their jobid is ```OMPI_PROC_MY_NAME->jobid``` and some of these pointers might be sentinels) and some procs not from ```MPI_COMM_WORLD``` (their jobid is not ```OMPI_PROC_MY_NAME->jobid``` __and__ they cannot be sentinels)

here is a patch that does that (and i have to fully test it)
```patch
commit f26c7fb2258db5ca51c650bdbd9cb86e3bb090da
Date:   Mon Feb 8 10:48:26 2016 +0900

    sentinel: fix 32 bits arch
    since a sentinel is only made from the current job, only store the
    first 31 bits of the vpid into the sentinel.

diff --git a/ompi/proc/proc.h b/ompi/proc/proc.h
index eac16cf..ecb4346 100644
--- a/ompi/proc/proc.h
+++ b/ompi/proc/proc.h
@@ -369,6 +369,7 @@ static inline bool ompi_proc_is_sentinel (ompi_proc_t *proc)
     return (intptr_t) proc & 0x1;
 }
 
+#if OPAL_SIZEOF_PROCESS_NAME_T == SIZEOF_VOID_P
 /*
  * we assume an ompi_proc_t is at least aligned on two bytes,
  * so if the LSB of a pointer to an ompi_proc_t is 1, we have to handle
@@ -407,6 +408,27 @@ static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel
   name.vpid = vpid;
   return name;
 }
+#elif 4 == SIZEOF_VOID_P
+/*
+ * currently, a sentinel is only made from the current jobid aka OMPI_PROC_MY_NAME->jobid
+ * so we only store the first 31 bits of the vpid
+ */
+static inline uintptr_t ompi_proc_name_to_sentinel (opal_process_name_t name)
+{
+    assert(OMPI_PROC_MY_NAME->jobid == name.jobid);
+    return (uintptr_t)((name.vpid <<1 | 0x1);
+}
+
+static inline opal_process_name_t ompi_proc_sentinel_to_name (uintptr_t sentinel)
+{
+  opal_process_name_t name;
+  name.jobid = OMPI_PROC_MY_NAME->jobid;
+  name.vpid = sentinel >> 1;
+  return name
+}
+#else
+#error unsupported pointer size
+#endif
 
 END_C_DECLS
 
diff --git a/opal/dss/dss_types.h b/opal/dss/dss_types.h
index 8c1bad9..c261223 100644
--- a/opal/dss/dss_types.h
+++ b/opal/dss/dss_types.h
@@ -14,7 +14,7 @@
  * Copyright (c) 2012-2013 Los Alamos National Security, Inc. All rights
  * Copyright (c) 2014      Intel, Inc. All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -47,6 +47,7 @@ typedef struct {
     opal_jobid_t jobid;
     opal_vpid_t vpid;
 } opal_process_name_t;
+#define OPAL_SIZEOF_PROCESS_NAME_T 8
 
 BEGIN_C_DECLS
 
```
```

**Comment by hjelmn** [comment]
Nice. @ggouaillardet beat me to it.

**Comment by ggouaillardet** [comment]
i updated the PR
now sentinels work also on 32bits and big endian archs
- on 64 bits arch, sentinel can be for any jobid (but we are not there yet)
- on 32 bits arch, sentinel can only be for procs of the current jobid.

**Comment by ggouaillardet** [comment]
@bosilca yes, that was a reply
And it us not conflicting with your request
Currently, sentinels are only created from processes in mpi_comm_world and all have the same jobid

It is always possible to have a group with processes from different jobids. My point is only pointers to processes in mpi_comm_world have the potential of being replaced with sentinels, hence the optimization @hjelmn suggested

Ideally, sentinels could be for processes of any job, but we are not there yet.

```c
   group = OBJ_NEW(ompi_group_t);

    size = ompi_process_info.num_procs;
    group->grp_proc_pointers = (ompi_proc_t **) calloc (size, sizeof (ompi_proc_t *));
    group->grp_proc_count = size;

    for (size_t i = 0 ; i < size ; ++i) {
        opal_process_name_t name = {.vpid = i, .jobid = OMPI_PROC_MY_NAME->jobid};
        /* look for existing ompi_proc_t that matches this name */
        group->grp_proc_pointers[i] = (ompi_proc_t *) ompi_proc_lookup (name);
        if (NULL == group->grp_proc_pointers[i]) {
            /* set sentinel value */
            group->grp_proc_pointers[i] = (ompi_proc_t *) ompi_proc_name_to_sentinel (name);
        } else {
            OBJ_RETAIN (group->grp_proc_pointers[i]);
        }
    }
```

hjelmn:
:bot:retest:

hjelmn:
If @bosilca has no objections I think this should be merged once Jenkins finishes. @jsquyres this is a better solution for 2.0.0.

bosilca:
@hjelmn go for it! :+1:

Issue #1348, opened 2016-02-09
Disable dynamic add procs for 32 bit and big endian environments


jsquyres:
Per discussion on the [Feb 9 2016 webex](https://github.com/open-mpi/ompi/wiki/WeeklyTelcon_20160209), we decided that the quickest / easiest solution for v2.0.0 is to simply disable the dynamic add procs behavior for big endian and 32 bit environments.

This will allow discussion to continue on #1345 to come to a better long-term solution (which may take a little time to get consensus, implement, test, ...etc. -- we may also want to discuss this at the [upcoming face to face Dallas meeting](https://github.com/open-mpi/ompi/wiki/Meeting-2016-02)).

It was decided that @hjelmn will do this quick fix for v2.0.0.

FYI/heads up for: @bosilca @rhc54 @ggouaillardet

jsquyres:
@hjelmn Can you refer to https://github.com/open-mpi/ompi/issues/1350 in the comments in your code?

jsquyres:
This issue is now moot -- we do *not* need to disable dynamic add procs for big endian and 32 bit machines.  Closing this issue.

PR #1349, opened 2016-02-09
ompi: temporarily disable dynamic add_procs on big-endian and 32-bit

hjelmn:
This commit temporarily disables the dynamic add_procs support on
big-endian and 32-bit systems. There is an issue with the sentinel
used to indicate when proc needs to be generated that causes issues on
these systems. This will be removed once the underlying issue has been
resolved.

Fixes open-mpi/ompi#1348

### DIFF
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2007-2015 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2007-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013      NVIDIA Corporation.  All rights reserved.
  * Copyright (c) 2013-2014 Intel, Inc. All rights reserved
@@ -65,7 +65,13 @@ char *ompi_mpi_show_mca_params_string = NULL;
 bool ompi_mpi_have_sparse_group_storage = !!(OMPI_GROUP_SPARSE);
 bool ompi_mpi_preconnect_mpi = false;
 
+/* NTH: temporarily disable add_procs on big-endian and 32-bit systems
+ * until we have a better solution for handling proc sentinels. */
+#if SIZEOF_VOID_P == 4 || BYTE_ORDER == BIG_ENDIAN
+#define OMPI_ADD_PROCS_CUTOFF_DEFAULT UINT_MAX
+#else
 #define OMPI_ADD_PROCS_CUTOFF_DEFAULT 0
+#endif
 uint32_t ompi_add_procs_cutoff = OMPI_ADD_PROCS_CUTOFF_DEFAULT;
 bool ompi_mpi_dynamics_enabled = true;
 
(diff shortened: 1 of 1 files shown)


hjelmn:
@ggouaillardet, @bosilca At today's call we decided this was the best way forward to not delay 2.0.0. This gives us time to work on the correct long-term solution.

ggouaillardet:
@hjelmn this does not disable cutoff, it only disables it by default (e.g. it is possible to re-enable it)
for example, on a 32 bits arch
```
$ OMPI_MCA_mpi_add_procs_cutoff=10 ~/local/ompi-master-m32/bin/ompi_info --all | grep add_procs_cutoff
                 MCA mpi: parameter "mpi_add_procs_cutoff" (current value: "10", data source: environment, level: 3 user/all, type: unsigned_int)
```
and running with
```mpirun --mca mpi_add_procs_cutoff 0 ...```
will cause a SIGSEGV

here is a patch that makes the value read-only (and cause mpirun to fail if it is forced)
```patch
diff --git a/ompi/runtime/ompi_mpi_params.c b/ompi/runtime/ompi_mpi_params.c
index c20562c..df55a8c 100644
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -276,7 +276,13 @@ int ompi_mpi_register_params(void)
                                   "communication performance at the cost of memory usage",
-                                  MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL, 0, 0, OPAL_INFO_LVL_3,
+                                  MCA_BASE_VAR_TYPE_UNSIGNED_INT, NULL, 0,
+#if SIZEOF_VOID_P == 4 || BYTE_ORDER == BIG_ENDIAN
+                                  MCA_BASE_VAR_FLAG_DEFAULT_ONLY,
+#else
+                                  0,
+#endif
+                                  OPAL_INFO_LVL_3,
 /* NTH: temporarily disable add_procs on big-endian and 32-bit systems
  * until we have a better solution for handling proc sentinels. */
 #if SIZEOF_VOID_P == 4 || BYTE_ORDER == BIG_ENDIAN
```

a softer option is to force cutoff disabled on 32 bits / big endian archs
```patch
diff --git a/ompi/runtime/ompi_mpi_params.c b/ompi/runtime/ompi_mpi_params.c
index c20562c..7383f6f 100644
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -68,7 +68,7 @@ bool ompi_mpi_preconnect_mpi = false;
 /* NTH: temporarily disable add_procs on big-endian and 32-bit systems
  * until we have a better solution for handling proc sentinels. */
 #if SIZEOF_VOID_P == 4 || BYTE_ORDER == BIG_ENDIAN
-#define OMPI_ADD_PROCS_CUTOFF_DEFAULT UINT_MAX
+#define OMPI_ADD_PROCS_CUTOFF_DEFAULT UINT32_MAX
 #else
 #define OMPI_ADD_PROCS_CUTOFF_DEFAULT 0
 #endif
@@ -285,6 +285,9 @@ int ompi_mpi_register_params(void)
                                   MCA_BASE_VAR_SCOPE_LOCAL,
 #endif
                                   &ompi_add_procs_cutoff);
+#if SIZEOF_VOID_P == 4 || BYTE_ORDER == BIG_ENDIAN
+    ompi_add_procs_cutoff = UINT32_MAX;
+#endif
 
     ompi_mpi_dynamics_enabled = true;
     (void) mca_base_var_register("ompi", "mpi", NULL, "dynamics_enabled",
```

hjelmn:
Ah, yes. Forgot about the default only setting. Will update my PR.

-Nathan

jsquyres:
@hjelmn @bosilca @ggouaillardet @rhc54 please confirm that this PR should be closed, and the solution from #1345 should be used for v2.0.0 instead (per [this comment](https://github.com/open-mpi/ompi/pull/1345#issuecomment-182504087)).

bosilca:
:+1:

jsquyres:
@bosilca Heh -- are you confirming my suggestion (that the solution to #1345 should be used), or that this one should be used?

hjelmn:
I would say this should be closed and @ggouaillardet #1345 PR'd to 2.0.0. It allows us to keep dynamic add_procs on the affected platforms.

jsquyres:
Ok, I'll close this one.  Can you guys open a PR for v2.0.0?

PR #1354, opened 2016-02-11
Add support for Singularity containers

rhc54:
Includes .m4 file for checking if Singularity is available and an orte/schizo component for setting the proper support if a container was given as the executable

### DIFF
--- a/config/opal_check_singularity.m4
+++ b/config/opal_check_singularity.m4
@@ -1,0 +1,60 @@
+# -*- shell-script ; indent-tabs-mode:nil -*-
+#
+# Copyright (c) 2016      Intel, Inc. All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# OPAL_CHECK_SINGULARITY(prefix, [action-if-found], [action-if-not-found])
+# --------------------------------------------------------
+AC_DEFUN([OPAL_CHECK_SINGULARITY],[
+    OPAL_VAR_SCOPE_PUSH([spath have_singularity])
+
+    AC_ARG_WITH([singularity],
+                [AC_HELP_STRING([--with-singularity(=DIR)],
+                                [Build support for the Singularity container, optionally adding DIR to the search path])])
+    spath=
+    AC_MSG_CHECKING([if Singularity support is to be built])
+    AS_IF([test "$with_singularity" = "no"],
+          [AC_MSG_RESULT([no])
+           have_singularity=no],
+          [AC_MSG_RESULT([yes])
+           AS_IF([test -z "$with_singularity" || test "$with_singularity" = "yes"],
+                 [ # look for the singularity command in the default path
+                   AC_CHECK_PROG([SINGULARITY], [singularity], [singularity])
+                   AS_IF([test "$SINGULARITY" != ""],
+                         [have_singularity=yes],
+                         [AS_IF([test "$with_singularity" = "yes"],
+                                [AC_MSG_WARN([Singularity support requested, but required executable])
+                                 AC_MSG_WARN(["singularity" not found in default locations])
+                                 AC_MSG_ERROR([Cannot continue])])
+                          have_singularity=no])],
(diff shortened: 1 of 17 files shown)


config/opal_check_singularity.m4
```diff
@@ -0,0 +1,52 @@
...
+
+# OPAL_CHECK_SINGULARITY(prefix, [action-if-found], [action-if-not-found])
+# --------------------------------------------------------
+AC_DEFUN([OPAL_CHECK_SINGULARITY],[
+    OPAL_VAR_SCOPE_PUSH([spath have_singularity])
+
+    AC_ARG_WITH([singularity],
+                [AC_HELP_STRING([--with-singularity(=DIR)],
+                                [Build support for the Singularity container, optionally adding DIR to the search path (default: no)])],
+                                [], with_singularity=no)
```
jsquyres: It took a little time to puzzle out this statement -- isn't the `[], with_singularity=no)` line mis-indented?  I.e., it's not part of `AC_HELP_STRING` -- they're arguments to `AC_ARG_WITH`.

I don't usually comment on indenting, but please fix this, because m4 is hard enough to read.  :smile:

config/opal_check_singularity.m4
```diff
@@ -0,0 +1,52 @@
...
+    spath=
+    AC_MSG_CHECKING([--with-singularity value])
+    AS_IF([test "$with_singularity" = "no"],
+          [AC_MSG_RESULT([no])
+           have_singularity=no],
+          [AS_IF([test -z "$with_singularity" || test "$with_singularity" = "yes"],
```
jsquyres: You set `with_singularity=no` in the "action if not specified" argument to `AC_ARG_WITH`, so the `-z` case can never happen here (which means you have defaulted to disabling Singularity support -- is that what you intended?).

config/opal_check_singularity.m4
```diff
@@ -0,0 +1,52 @@
...
+                 [ AC_MSG_RESULT([simple ok (unspecified)])
```
jsquyres: This message ("unspecified") does not match the if statement -- even if you `--with-singularity`, it'll say "(unspecified)".

config/opal_check_singularity.m4
```diff
@@ -0,0 +1,52 @@
...
+                   # look for the singularity command in the default location
+                   AC_CHECK_PROG([SINGULARITY], [singularity], [singularity])
+                   AS_IF([test "$SINGULARITY" != ""],
+                         [have_singularity=yes],
+                         [have_singularity=no])],
+                 [ AC_MSG_RESULT([$with_singularity])
+                   # look for the singularity command in the bin subdirectory
+                   save_path=$PATH
+                   PATH=$with_singularity/bin:$PATH
```
jsquyres: I know you're tried if seeing comments on this PR :smile:, but you might want to OPAL_CHECK_WITHDIR on $with_singularity; it'll complain to the user if the path they specified doesn't exist.  We put in these kind of checks because users complained that configure allowed them to `--with-foo=/some/nonexistent/dir` without complaint (e.g., if they fat-fingered the directory name).

config/opal_check_singularity.m4
```diff
@@ -0,0 +1,52 @@
...
+    AS_IF([test "$with_singularity" = "no"],
+          [AC_MSG_RESULT([no])
+           have_singularity=no],
+          [AS_IF([test -z "$with_singularity" || test "$with_singularity" = "yes"],
+                 [ AC_MSG_RESULT([simple ok (unspecified)])
+                   # look for the singularity command in the default location
+                   AC_CHECK_PROG([SINGULARITY], [singularity], [singularity])
+                   AS_IF([test "$SINGULARITY" != ""],
+                         [have_singularity=yes],
+                         [have_singularity=no])],
```
jsquyres: This code path will still allow configure to continue even if a human specifies `--with-singularity` but singularity is not found.

rhc54:
Ease up there, cowboy - this isn't done yet. I appreciate the comments and will correct things, but this is a work-in-progress that will take a few iterations.

jsquyres:
@rhc54 Roger!

Issue #1355, opened 2016-02-11
MPI_GET_ADDRESS fails with MPI_BOTTOM


jsquyres:
As pointed out by @dalcinl, MPI_GET_ADDRESS fails when passed MPI_BOTTOM.

See http://www.open-mpi.org/community/lists/devel/2016/02/18578.php

jsquyres:
@dalcinl also sent a small test reproducer which should be added to our regression tests:

```c
#include <mpi.h>
int main(int argc, char *argv[])
{
 MPI_Aint addr;
 MPI_Init(&argc, &argv);
 MPI_Get_address(MPI_BOTTOM, &addr);
 MPI_Finalize();
 return 0;
}
```

bosilca:
MPI_BOTTOM is defined as a "constant" by the MPI standard. I don't understand what you expect to get as it's address ?

jsquyres:
Actually, you're right -- I read it too quickly this morning and thought it was segv'ing.  It's invoking an MPI exception, which seems to be the correct response.

@dalcinl please feel free to reply if you think that's incorrect.

dalcinl:
On 11 February 2016 at 21:36, Jeff Squyres <notifications@github.com> wrote:

> Actually, you're right -- I read it too quickly this morning and thought
> it was segv'ing. It's invoking an MPI exception, which seems to be the
> correct response.
>
> @dalcinl <https://github.com/dalcinl> please feel free to reply if you
> think that's incorrect.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/open-mpi/ompi/issues/1355#issuecomment-183001129>.
>

​Of course MPI_BOTTOM is a constant, I'm not asking about taking the
"address" of a constant expression. My point is that the following two
lines should succeed (at least on platforms where MPI_BOTTOM is equivalent
to NULL):

​MPI_Get_address(MPI_BOTTOM,&address);
assert(address == 0);

​i.e. this is equivalent to a cast

address = (MPI_Aint)MPI_BOTTOM;

Writing the C code in as slightly different way:

MPI_Aint address;
void *location = MPI_BOTTOM;
​MPI_Get_address(location, &address);
assert(address == 0);

I'm obviously not asking for the address of "location" (otherwise I would
write &location), I'm just asking MPI_Get_address() to "translate" the
address MPI_BOTTOM represents to a signed integral (of type MPI_Aint). Why
would the code above fail?

[...]

PR #1356, opened 2016-02-11
Fix MPI_Get_address (MPI_BOTTOM, ...)

hjelmn:
Nowhere in the standard does it say that it is invalid to pass
MPI_BOTTOM to MPI_Get_address yet we were returning an error. This
commit removes the error check on NULL == location.

Fixes open-mpi/ompi#1355.

### DIFF
--- a/ompi/mpi/c/get_address.c
+++ b/ompi/mpi/c/get_address.c
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2013      Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2013-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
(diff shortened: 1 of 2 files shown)


jsquyres:
I think we will need to update in the Fortran bindings, too.

[Per @dalcinl's comment](https://github.com/open-mpi/ompi/issues/1355#issuecomment-183031915) Is it desirable to return 0 in ADDRESS when you pass MPI_BOTTOM in the Fortran bindings?  I think it is...?

hjelmn:
Ah, good olde fortran. Will take a look at that now.

jsquyres:
I checked: you only need to update ompi/mp/fortran/mpif-h/get_address_f.c.  The mpi and mpi_f08 modules both use this underlying function.

jsquyres:
@hjelmn Did you get a chance to update this PR with the fix for `ompi/mp/fortran/mpif-h/get_address_f.c`?

hjelmn:
Updated.

lanl-ompi:
Test FAILed.

jsquyres:
:+1:

PR #1357, opened 2016-02-12
Protect against a non-privileged port connecting to us when we are running as root

rhc54:
@jsquyres Mind just verifying the logic on this one? Plugs a potential security hole causing angst at the mother ship

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_listener.c
+++ b/orte/mca/oob/tcp/oob_tcp_listener.c
@@ -13,7 +13,7 @@
  *                         All rights reserved.
  * Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
- * Copyright (c) 2013-2014 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2013-2016 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
@@ -384,6 +384,7 @@ static int create_listen(void)
         /* add this port to our connections */
         conn = OBJ_NEW(mca_oob_tcp_listener_t);
         conn->sd = sd;
+        conn->port = ntohs(((struct sockaddr_in*) &inaddr)->sin_port);
         opal_list_append(&mca_oob_tcp_component.listeners, &conn->item);
         /* and to our ports */
         asprintf(&tconn, "%d", ntohs(((struct sockaddr_in*) &inaddr)->sin_port));
@@ -621,7 +622,9 @@ static int create_listen6(void)
 
         /* add this port to our connections */
         conn = OBJ_NEW(mca_oob_tcp_listener_t);
+        conn->tcp6 = true;
         conn->sd = sd;
+        conn->port = ntohs(((struct sockaddr_in6*) &inaddr)->sin6_port);
         opal_list_append(&mca_oob_tcp_component.listeners, &conn->item);
         /* and to our ports */
         asprintf(&tconn, "%d", ntohs(((struct sockaddr_in6*) &inaddr)->sin6_port));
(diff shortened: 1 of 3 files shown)


jsquyres:
@rhc54 It looks like you're checking the file descriptor number, not the port number.

rhc54:
@jsquyres Could you take a quick gander? Turns out you cannot use getpeername as the sockets have to be connected first, and that is what we are trying to avoid. However, I am able to get the port number of the inbound connection once we accept it, and then can break things at that time.

orte/mca/oob/tcp/oob_tcp_listener.c
```diff
@@ -909,7 +936,9 @@ static void connection_event_handler(int incoming_sd, short flags, void* cbdata)
 static void tcp_ev_cons(mca_oob_tcp_listener_t* event)
 {
     event->ev_active = false;
+    event->tcp6 = false;
     event->sd = -1;
+    event->port = -1;
```
jsquyres: The port is a `uint16_t` -- this assignment may well generate a compiler warning...?

jsquyres:
Otherwise :+1:

rhc54:
Good point about the default value for the port number - I changed it to zero. Will let this run thru the Jenkins tests and commit. Thanks!

rhc54:
@bosilca would you mind taking a quick glance at this? The goal is to try and protect against connections from non-privileged ports if we are operating at a privileged level with a privileged port. I think I'm checking the remote port properly, but would welcome another set of eyes to ensure it's doing what I expected.

orte/mca/oob/tcp/oob_tcp_listener.c
```diff
@@ -739,6 +742,30 @@ static void* listen_thread(opal_object_t *obj)
                 pending_connection->fd = accept(sd,
                                                 (struct sockaddr*)&(pending_connection->addr),
                                                 &addrlen);
+
+                /* if we are on a privileged port, we only accept connections
+                 * from other privileged sockets. A privileged port is one
+                 * whose fd is less than 1024 on Linux, so we'll check for that. */
```
ggouaillardet: Two typos
"A privileged socket is one whose port is less than ..."

orte/mca/oob/tcp/oob_tcp_listener.c
```diff
@@ -739,6 +742,30 @@ static void* listen_thread(opal_object_t *obj)
...
+                    }
+                    if (1024 < inport) {
+                        /* someone tried to cross-connect privileges,
+                         * say something */
+                        orte_show_help("help-oob-tcp.txt",
+                                       "privilege failure",
+                                       true, opal_process_info.nodename,
+                                       listener->port, inport);
+                        OBJ_RELEASE(pending_connection);
+                        CLOSE_THE_SOCKET(sd);
```
ggouaillardet: Do you really want to close sd (aka the listening socket) ?
Or did you mean to close pending_connection->fd (aka the non privileged socket) ?

rhc54:
@ggouaillardet Good points - thx!

orte/mca/oob/tcp/oob_tcp_listener.c
```diff
@@ -739,8 +742,31 @@ static void* listen_thread(opal_object_t *obj)
...
+                        inport = ntohs(((struct sockaddr_in*)&pending_connection->addr)->sin_port);
+                    }
+                    if (1024 < inport) {
+                        /* someone tried to cross-connect privileges,
+                         * say something */
+                        orte_show_help("help-oob-tcp.txt",
+                                       "privilege failure",
+                                       true, opal_process_info.nodename,
+                                       listener->port, inport);
+                        OBJ_RELEASE(pending_connection);  // closes the incoming fd
```
ggouaillardet: Do you intentionally __not__ close ```pending_connection->fd``` ?
Or should the destructor (to be created) do that for you ?

rhc54: Sorry - was looking at the wrong destructor. No, it needs to close it.

PR #1358, opened 2016-02-14
Enable the PMIx notification callback system and fix debugger attach

rhc54:
This currently is only supported by the pmix120 component, which is not selected by default. All other components will ignore error registration requests, and thus do not support debugger attach when launched via mpirun. Note that direct launched applications will support such attachment, but may not do so in a scalable fashion.

Fixes ##1225

### DIFF
--- a/ompi/mca/rte/orte/rte_orte.h
+++ b/ompi/mca/rte/orte/rte_orte.h
@@ -91,7 +91,7 @@ OMPI_DECLSPEC void __opal_attribute_noreturn__
 #define OMPI_ERROR_LOG ORTE_ERROR_LOG
 
 /* Init and finalize objects and operations */
-#define ompi_rte_init(a, b) orte_init(a, b, ORTE_PROC_MPI)
+OMPI_DECLSPEC int ompi_rte_init(int *pargc, char ***pargv);
 #define ompi_rte_finalize() orte_finalize()
 OMPI_DECLSPEC void ompi_rte_wait_for_debugger(void);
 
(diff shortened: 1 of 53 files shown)


rhc54:
@jsquyres @ggouaillardet @hjelmn @hppritcha Can someone please verify that this solves the debugger attach at launch problem?

lanl-ompi:
Test FAILed.

jsquyres:
@rhc54 Did this not work?

rhc54:
@jsquyres I think I accidentally deleted the branch and then immediately restored it. It _should_ work, but I was asking that you test it with a debugger

jsquyres:
I actually don't have Totalview.  :-(

I can confirm that this *still* works for DDT, though.

@hppritcha Do you have Totalview?  Or should we ask our friends at TV to test for us?

hppritcha:
I will give this a try on one of the LANL clusters today.

hppritcha:
I did try to run with totalview yesterday - and thought I'd posted here.  I hit a bunch of issues almost certainly not related to this PR.  I'd say go ahead and check this in (after rebase).  @hjelmn and I will check to see whether we have a site specific issue with our totalview install.

hppritcha:
Nathan and I worked more on this today.  Things look good with ddt.  unfortunately, with totalveiw things still don't seem to work.  We get this error when trying to use the mpirun --debug option:
[mu0116.localdomain:30800] CALLING NOTIFY ERROR

hjelmn:
Asked @rhc54 about the message and he indicated it is a stale debug message. Should be cleaned out.

PR #1362, opened 2016-02-15
btl/openib: correctly issue a warning when two btls or more are in th…

### LINKED ISSUE #1352: Misleading openib Warning message for default subnet GID
While testing on a node with 4 HFI (Intel OmniPath devices) using openib btl we saw that a Warning message is shown when the default subent GID is used to warn the user of possible miss behaviors. This is described in detail here: https://www.open-mpi.org/faq/?category=openfabrics#ofa-default-subnet-gid

However, found that the text does not represent what the code is actually testing/finding. The text says that two ports have the default subnet GID. When, instead the code is testing if there is at least one "openib btl" available and the CURRENT subnet ID is the default one and the warning message is not masked, then print the Warning. See:   
"ompi/mca/btl/openib/btl_openib_component.c" line 750 in ompi-release branch v1.10: 

```c
    if(mca_btl_openib_component.ib_num_btls > 0 &&
            IB_DEFAULT_GID_PREFIX == subnet_id &&
            mca_btl_openib_component.warn_default_gid_prefix) {
        opal_show_help("help-mpi-btl-openib.txt", "default subnet prefix",
                true, ompi_process_info.nodename);
    }
```

In my setup, as I mentioned, I have 4 ports and only hfi1_1 with the default subnet GID. Each time that port is "evaluated" the warning message is shown. Playing with  "- x OMPI_MCA_btl_openib_if_include=hfi1_x,hfi1_y "  (being x and y numbers different than 1 and smaller than 4) the warning message is not shown. 
Therefore, there are two ways to fix this: 
   a) update the error message (and the FAQ) entry describing the behavior. 
   b) update the code to compare all subnet GIDs.

ggouaillardet:
…e default subnet gid

Thanks Matias Cabral for reporting this issue.

Fixes #1352

### DIFF
--- a/opal/mca/btl/openib/btl_openib.h
+++ b/opal/mca/btl/openib/btl_openib.h
@@ -18,7 +18,7 @@
  * Copyright (c) 2009-2010 Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2013-2014 NVIDIA Corporation.  All rights reserved.
  * Copyright (c) 2014      Bull SAS.  All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@mcabral did you get any chance to check this pr ?

jsquyres:
@ggouaillardet Do you want to rebase this, since Mellanox just pushed some changes to multi-subnet support?

@matcabral When this is rebased, can you test?

matcabral:
Hi, 
I currently don't have access to this system. I'll ask around to see if is available. 
Regarding the fix, I agree it matches the text description.

ggouaillardet:
@jsquyres  will do
at first glance, these two changes look orthogonal.
but I will rebase and double check that anyway

ggouaillardet:
@alinask @miked-mellanox can you please double check this PR is still valid after #1459 ?

alinask:
:+1:

matcabral:
Tested the patch shared in #1352 and confirmed it works.

PR #1365, opened 2016-02-16
Plug some memory leaks

rhc54:
Release child object when we are recording someone's relatives.
(Thanks to Mark Santcroos!)

Release routing list entries.
(Thanks to Mark Santcroos!)

Address some Coverity concerns


rhc54:
unrelated error

PR #1367, opened 2016-02-16
Fix XRC support

hjelmn:
@miked-mellanox Some of these bugs should have been caught by Jenkins. Are you not testing XRC?


hjelmn:
@hppritcha Put this for 2.0.0 since I would like this to be deployed on our mlx4 cluster. Let me know if you would rather this wait for 2.0.1.

hjelmn:
@artpol84 Does your testing include XRC? I could not test until recently so I didn't catch the regression when the threading fixes to add_procs were committed.

artpol84:
@hjelmn, no i don't think so. Let's discuss that on the OMPI meeting.

PR #1369, opened 2016-02-17
datatype: correctly support MPI_Pack_external when heterogeneous supp…

ggouaillardet:
…ort is enabled

### DIFF
--- a/ompi/mpi/c/pack_external.c
+++ b/ompi/mpi/c/pack_external.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2006      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2013      Los Alamos National Security, LLC.  All rights
  *                         reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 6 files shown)


ggouaillardet:
@bosilca could you please review this ?
I used a logic similar to unpack, so MPI_Pack_external can do the byte swapping if needed.

currently, this works only with --enable-heterogeneous.

In order to get this working without --enable-heterogeneous, I will have to always compile some code that is currently used only with heterogeneous cluster.
this will be a bit more of code which is always executed, but that is only needed for MPI_{Pack,Unpack}_external.
an other option is to duplicate some code without --enable-heterogeneous.

can you please advise ?

ggouaillardet:
@bosilca 
i pushed a second commit to fix openmpi without ```--enable-heterogeneous```
bottom line, i just had to make ```OPAL_CONVERTOR_COMPUTE_REMOTE_SIZE``` the same regardless heterogeneous support, which is a much smaller change than i anticipated.

bosilca:
@ggouaillardet I made a different patch that enables the heterogeneous mode for the OPAL datatype in all cases, and adds support for more complicated datatypes. Take a look at #1394 .

ggouaillardet:
this topic is now handled is #1394, so i am closing this PS

Issue #1363, opened 2016-02-15
PPC64 crash with current master branch


nysal:
I started noticing occasional crashes on PPC64 with master branch. The crash is from the child process soon after mpirun forks it (before it exec's the application).  The stack trace was nearly the same most of the time:

```
#0  0x00003fffb791031c in _int_malloc () from /lib64/power8/libc.so.6
#1  0x00003fffb79134dc in malloc () from /lib64/power8/libc.so.6
#2  0x00003fffb79030ac in vasprintf@@GLIBC_2.17 () from /lib64/power8/libc.so.6
#3  0x00003fffb78e06e4 in asprintf@@GLIBC_2.17 () from /lib64/power8/libc.so.6
#4  0x00003fffb7ef058c in orte_iof_base_setup_child (opts=0x3fffffffd578, env=0x3fffffffd560)
    at /u/src/ompi/orte/mca/iof/base/iof_base_setup.c:186
#5  0x00003fffb5c42cd0 in do_child (context=0x10288060, child=0x10287f00, environ_copy=0x10042600, jobdat=0x10047b00, 
    write_fd=39, opts=...) at /u/src/orte/mca/odls/default/odls_default_module.c:419
#6  0x00003fffb5c43b64 in odls_default_fork_local_proc (context=0x10288060, child=0x10287f00, environ_copy=0x10042600, 
    jobdat=0x10047b00) at /u/src/orte/mca/odls/default/odls_default_module.c:726
#7  0x00003fffb7ef8fcc in orte_odls_base_default_launch_local (fd=-1, sd=4, cbdata=0x10287cb0)
    at /u/src/orte/mca/odls/base/odls_base_default_fns.c:1038
#8  0x00003fffb7da15ec in event_process_active_single_queue (base=0x1008d5d0, activeq=0x1008db50)
    at /u/src/opal/mca/event/libevent2022/libevent/event.c:1370
#9  0x00003fffb7da197c in event_process_active (base=0x1008d5d0)
    at /u/src/opal/mca/event/libevent2022/libevent/event.c:1440
#10 0x00003fffb7da23cc in opal_libevent2022_event_base_loop (base=0x1008d5d0, flags=1)
    at /u/src/opal/mca/event/libevent2022/libevent/event.c:1644
#11 0x0000000010005eb8 in orterun (argc=7, argv=0x3ffffffff068)
    at /u/src/orte/tools/orterun/orterun.c:1077
[...]

PR #1378, opened 2016-02-18
Make UD OOB memory registrations a multiple of page size

nysal:
If ibv_fork_init() has been invoked the pages are marked MADV_DONTFORK.
If we only partially use a page, any data allocated on the remainder of
the page will be inaccessible to the child process.

Fixes open-mpi/ompi#1363


nysal:
:bot:retest

nysal:
Not sure whats wrong with travis. It fails to install gcc 5 packages. Merging since other tests passed.

PR #1380, opened 2016-02-18
btl/openib: XRC fix bug that could cause an invalid SRQ# to be used

hjelmn:
This commit fixes a bug that occurs when attempting a get or put
operation on an endpoint that is not already connected. In this case
the remote_srqn may be set to an invalid value as the rem_srqs array
on the endpoint is not populated. This commit moves the usage of the
rem_srqs array to the internal put/get functions where it is
guaranteed this array is populated.

### DIFF
--- a/opal/mca/btl/openib/btl_openib_atomic.c
+++ b/opal/mca/btl/openib/btl_openib_atomic.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014      Los Alamos National Security, LLC.  All rights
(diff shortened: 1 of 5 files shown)


hjelmn:
@miked-mellanox No need to disable XRC now. Found the problem. Please review this for v2.0.0 when I open that PR.

hjelmn:
This bug is unique to the BTL 3.0 version of openib. In BTL 2.1 get frags were queued before filling them in.

hjelmn:
Also found a deadlock and consistency issue with the ib_addr lock.

hjelmn:
Mellanox failure makes no sense. Merging as this fixes a real bug.

PR #1381, opened 2016-02-18
orterun: allow DDT if options contain :'s

hjelmn:
There is a bug in MPMD detection that disables totalview if a : is
found anywhere on the command line. This includes inside an argument
option or MCA variable value. This commit changes the check to look
for the string " : " instead of the character : which should eliminate
the issue in most cases.

### DIFF
--- a/orte/tools/orterun/orterun.c
+++ b/orte/tools/orterun/orterun.c
@@ -1,5 +1,5 @@
-/* -*- C -*-
- *
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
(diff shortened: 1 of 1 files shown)


rhc54:
:+1:

hjelmn:
Build started before the XRC fix went into master.

hppritcha:
Please open a pr on v2.x and if needed v1.10

hjelmn:
Yup. This should go into 1.10.3 but can wait for 2.0.1 as it affect launching the debugger with mpirun -debug

hjelmn:
Will open the PRs after I track down the remaining XRC issue :).

Issue #1383, opened 2016-02-19
Fix remaining XRC issues


hjelmn:
Just to keep track of the Jenkins failure. I think I know what might be causing the remaining problem. @miked-mellanox I will fix and PR tonight.

hjelmn:
@hppritcha This will fit in with the rest of the XRC fixes which should be reviewed tomorrow.

PR #1384, opened 2016-02-19
btl/openib: XRC save SRQ#s on the loopback endpoint

hjelmn:
This commit fixes a bug that can occur when communicating via XRC to
peers on the same node. UDCM was not saving the SRQ numbers on the
loopback endpoint (which shares its ib_addr info with all local peers)
so any messages to local peers use an invalid SRQ number.

Fixes open-mpi/ompi#1383


hjelmn:
This should be the last of them. I couldn't reproduce the Jenkins issue until I went back to master.

hjelmn:
Got it!

PR #1385, opened 2016-02-19
Topic/hetero spawn

ggouaillardet:


ompi/dpm/dpm.c
```diff
@@ -403,24 +403,21 @@ int ompi_dpm_connect_accept(ompi_communicator_t *comm, int root,
...
-            /* set the locality */
-            new_proc_list[i]->super.proc_flags = OPAL_PROC_NON_LOCAL;
-            /* have to save it for later */
+            proc = cd->p;
+            new_proc_list[i] = proc ;
+            /* ompi_proc_complete_init_single() optionally retrieves OPAL_PMIX_LOCALITY
+             * and OPAL_PMIX_HOSTNAME, and we can live withot them, so initialize
+             * these values here in case they are not retrieved */
+            proc->super.proc_flags = OPAL_PROC_NON_LOCAL;
+            proc->super.proc_hostname = NULL;
```
rhc54: FWIW: ompi_proc_complete_init_single already initializes these fields to these values, so this isn't really necessary. The only time they won't be initialized is when you call that function with your own process name, and that cannot happen here as we are only looping over processes we previously didn't know about.

ggouaillardet: Thanks !
I will update and commit this on Monday

PR #1398, opened 2016-02-24
tests: fix library name

jsquyres:
Use @OPAL_LIB_PREFIX@ as appropriate in the library that we link against.


jsquyres:
False Travis failures (i.e., they look like failures in the travis infrastructure).

PR #1400, opened 2016-02-24
Adjust the pmix external component configure error messages

rhc54:
pmix: Make the error message on external pmix library a little clearer

Make the error message on external pmix library a little clearer by separating out the libevent from the libhwloc checks by separating out the libevent from the libhwloc checks


lanl-ompi:
Test FAILed.

PR #1411, opened 2016-02-27
Fix a number of issues, some of which have lingered for a long time

rhc54:
* provide a more reliable way of determining that a process is a singleton by leveraging the schizo framework. Add new components for slurm, alps, and orte to detect when we are in a managed environment, and if we have been launched by mpirun or a native launcher. Set the correct envars to control ess and pmix selection in each case.

* ignore the old pmix112 component in favor of the newer pmix120 that has error notification support. We retain the old component just in case we run into problems when we test the newer one and need to quickly switch back.

* fix singleton comm-spawn by correctly setting the num_apps field of the orte_job_t created by the daemon - this fixes a segfault in register_nspace on newly created daemons

* ensure orterun doesn't propagate any ess or pmix directives in its environment

### DIFF
--- a/opal/mca/pmix/isolated/pmix_isolated.c
+++ b/opal/mca/pmix/isolated/pmix_isolated.c
@@ -113,26 +113,26 @@ const opal_pmix_base_module_t opal_pmix_isolated_module = {
     .register_jobid = isolated_register_jobid
 };
 
-static int pmix_init_count = 0;
-static opal_process_name_t pmix_pname;
+static int isolated_init_count = 0;
+static opal_process_name_t isolated_pname;
 
 static int isolated_init(void)
 {
     int rc;
     opal_value_t kv;
 
-    ++pmix_init_count;
+    ++isolated_init_count;
 
     /* store our name in the opal_proc_t so that
      * debug messages will make sense - an upper
      * layer will eventually overwrite it, but that
      * won't do any harm */
-    pmix_pname.jobid = 1;
-    pmix_pname.vpid = 0;
-    opal_proc_set_name(&pmix_pname);
+    isolated_pname.jobid = 1;
+    isolated_pname.vpid = 0;
+    opal_proc_set_name(&isolated_pname);
     opal_output_verbose(10, opal_pmix_base_framework.framework_output,
                         "%s pmix:isolated: assigned tmp name %d %d",
-                        OPAL_NAME_PRINT(pmix_pname),pmix_pname.jobid,pmix_pname.vpid);
+                        OPAL_NAME_PRINT(isolated_pname),isolated_pname.jobid,isolated_pname.vpid);
 
     // setup hash table
     opal_pmix_base_hash_init();
@@ -248,11 +248,11 @@ err_exit:
 
 static int isolated_fini(void)
 {
-    if (0 == pmix_init_count) {
+    if (0 == isolated_init_count) {
         return OPAL_SUCCESS;
     }
 
-    if (0 != --pmix_init_count) {
+    if (0 != --isolated_init_count) {
         return OPAL_SUCCESS;
     }
     opal_pmix_base_hash_finalize();
@@ -261,7 +261,7 @@ static int isolated_fini(void)
 
 static int isolated_initialized(void)
 {
-    if (0 < pmix_init_count) {
+    if (0 < isolated_init_count) {
         return 1;
     }
     return 0;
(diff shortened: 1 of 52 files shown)


lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

rhc54:
@ggouaillardet @hjelmn @jsquyres Hey folks - I could use some help on this one. I am continuing to see a problem with things hanging if more than one pmix component is in memory, and I have had zero luck tracking down why. I fixed it for dynamic builds by simply avoiding the situation, but we still hang in static builds.

Looking at gdb and valgrind reveals nothing of any use. I find all the threads sitting peacefully in event loops, but there is no information beyond that - we aren't sitting in some blocking location waiting for an event to show up. So I honestly don't know what is happening.

Can some fresh eyes take a look at this?

ggouaillardet:
@rhc54 can you explain how to reproduce the hang ?
Does it occurs when some processes use pmix112 and other use pmix120 ?

rhc54:
For now, the easiest way to reproduce it is to --enable-static --disable-shared. Then any execution of a simple MPI program will hang during Init.

rhc54:
I should have added: if you want to reproduce it with a dynamic build, just remove the .opal_ignore in the pmix112 component.

ggouaillardet:
is there any kind of mechanism to ensure all processes use the same pmix version ?
iirc, I tried to blacklist pmix112 via the command line but could not do that,
I dimply forgot about it since everything was working for me

rhc54:
hang on - I may have found a race condition due to a missing thread-shift. will continue in the morning.

ggouaillardet:
with your branch, MPI always hang with pmix120, even when pmix120 is alone
with pmix112 alone, I see no hang
with both pmix112 and pmix120, I get an error because somehow pmix112 talks with pmix120
I "fixed" that by adding a dummy PMIX_PADDING in pmix_data_type_t right after PMIX_TIME

rhc54:
I found it - the problem was actually a race condition due to a missing thread shift. Adding components just changed the timing a little and masked the true issue. This now seems to be working even with all components being available.

rhc54:
@hjelmn @hppritcha Can you folks please test this on the Cray? The new schizo functionality should work there, but I cannot test it myself.

ggouaillardet:
@rhc54 it works for me now :-)

here is what happened previously
- on mpirun
  - pmix112 priority is 5
  - pmix120 priority is 5

since no pmix is explicitly required by user, a.out is invoked with
```
OMPI_MCA_pmix=^s1,s2,cray,isolated
```

- on a.out
  - pmix112 priority is 80
  - pmix120 priority is 100

a.out chose pmix120, but for some reasons, mpirun chose pmix112, and these cannot talk to each other because of the difference in the ```pmix_data_type_t``` enumeration


that raises several questions
- pmix component priority is hard coded and cannot be set by a mca parameter, is this intended ?
 (but it is possible to include/exclude components with the ```pmix``` mca parameter)
- should mpirun explicitly sets the pmix component it selected ?
for example ```OMPI_MCA_pmix=pmix112``` instead of ```OMPI_MCA_pmix=^s1,s2,cray,isolated``` so mpirun/orted/a.out all use the same pmix version
- should pmix (or OpenMPI pmix glue) add an extra test in the "handshake" to check both pmix versions are identical, or at least compatible ?
- what should happen when two components have the hightest and same priority ?
iirc, the selected module could depend on how files are enumerated by the system, which means it can be different from one node to an other and leads to site specific crashes
one option could be to issue a warning (two modules could be selected based on their priority, choosing module xxx, do not know what will happen next)
an other option could be to explicitly force the ("randomly"?) selected module
(not this works only in mpirun ... if such a component is open'ed only in a.out, there is no way to enforce all tasks will select the same component)

rhc54:
Answering your questions in order:

* yes - we hardcode the priorities. We have found that people rarely understand the numbering and prefer to just specify which ones to use

* the schizo framework is now setting an adequate level of specification. PMIx is being updated to handle backward compatible communications between versions

* the PMIx library itself is being updated to detect the version on each end of a connection

* yes, the selection if two have the same priority depends on the order in which they are loaded. Not sure how to avoid that situation other than to ensure the priorities are correctly set. In this case, I had missed updating the priority for the server side of the pmix connection - so it was a bug and easily corrected.

HTH

ggouaillardet:
Thanks Ralph !

hppritcha:
i will look at this when i get to chicago later this evening.

----------

Howard

jsquyres:
I sent a tarball of this PR to John Delsignore for testing with Totalview.

hppritcha:
This is working for me with srun and aprun, but fails using mpirun on a Cray system using aprun.  Looks like pmix initialization fails.  Debugging...

hppritcha:
Hmm... looks like alps schizo isn't working right. 

I think to fix this, I need to understand what it is the schizo/alps is suppose to be doing?  The code I see there can't distinguish between whether the process is started via alps or mpirun.

rhc54:
@hppritcha Sorry about that - the schizo/alps component was indeed missing a few lines of code. I've fixed it now, so please let me know if this is now correct.

The purpose of the component is to correctly detect between launch via mpirun, direct launch,  singletons running inside of an allocation, and singletons running outside of an allocation (but in a managed environment). It also allows us to set environment-specific controls when direct launched (e.g., not overriding SLURM binding directives).

hppritcha:
works for me :+1:

PR #1414, opened 2016-02-29
examples: update ompi_info bindings checks

jsquyres:
Use `-q` option to grep/egrep to suppress output (we only need the exit status).  Also, use egrep for the "use mpi" check, because some versions of ompi_info say `bindings:use_mpi:yes` and others say `bindings:use_mpi:"yes` (i.e., with the double quote).  This regexp will work with both versions.

@ggouaillardet Please review.  If acceptable, I'll add this to open-mpi/ompi-release#993 and open-mpi/ompi-release#994.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -84,19 +84,19 @@ all: hello_c ring_c connectivity_c
 # MPI examples
 
 mpi:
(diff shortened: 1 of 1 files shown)


jsquyres:
@ggouaillardet if you have a few seconds, please check out this PR.  Thanks!

ggouaillardet:
:+1:

PR #1416, opened 2016-03-01
Fix a potential race condition in recv cancel

abouteiller:
The access to request->req_match_received is safe, however, when we read it false, we may unsafely assume that the request is not matched and go on and cancel it. As the matching lock is taken late, the request may actually get matched in the meantime, and we end-up cancelling a matched request.


hppritcha:
@abouteiller please open a PR for v2.x once this has been merged in to master.

PR #1417, opened 2016-03-02
RFC: change default build to always be optimized (even for developers)

jsquyres:
A not-uncommon complaint we hear is someone who gets lower-than-expected performance, and it is later discovered that they used a debugging build.  This commit enables ortertun to automatically print a stern warning/admonishion to not use this build for performance tests if it is a debug build.  You can use the `orterun --nowarn-debug-build` CLI option to disable this message, or export `OMPI_MCA_nowarn_debug_build=1`.

The rationale for this option is to make it totally, completely obvious that you are using a debug build.  Yes, this extra output will screw up the output of running benchmarks and other applications.  That's kinda the point -- if you're running benchmarks / parsing stdout to get performance data, you shouldn't be using a debugging build, and we want to smack you in the face with that fact.

That being said, there are many useful scenarios for a debug build.  Hence, you can disable the warning with an mpirun CLI option and/or setting an env var.

Comments?

### DIFF
--- a/config/opal_configure_options.m4
+++ b/config/opal_configure_options.m4
@@ -10,7 +10,7 @@ dnl Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
 dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
-dnl Copyright (c) 2006-2015 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2009      IBM Corporation.  All rights reserved.
 dnl Copyright (c) 2009      Los Alamos National Security, LLC.  All rights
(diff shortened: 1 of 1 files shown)


rhc54:
Why not a different, less annoying approach? The comments in the code explicitly state that defaulting to a debug build when using a repo was done for "early development" purposes. So why not just turn that off? It would be much less annoying to have to remember to enable-debug.

jsquyres:
Yeah, I just replied on the mailing list -- I think we agreed to do this last week in Dallas, anyway (disable debug-by-default builds).  I think these could still be complimentary features though -- if debug builds are not common (because they're not the default), then perhaps a warning like this is less obtrusive, anyway.

I'll add a "don't do a debug build by default" commit to this PR.

jsquyres:
Ok, I added a 2nd commit that disables debug-build-by-default behavior if a .git directory is found.

rhc54:
I believe @gpaulsen was going to do it, but I see you've done it here. Pretty trivial change.

I would advocate to remove the warning and just do your 2nd commit.

gpaulsen:
No worries.  I made a comment on the commit.  Is that easily visible to you Jeff?  I assume it emailed you.
Is this for 2.0?

jsquyres:
@gpaulsen The milestone on this PR is v2.1.0.

Per discussion on the call today, it seems like everyone is comfortable with:

* Change the default to build optimized (i.e., never magically make a debug build -- debug builds now must be specifically requested via `--enable-debug --enable-mem-debug --enable-picky`)
* Remove the warning that was initially added in this PR

I.e., you only get a debug build if you ask for it, and if you asked for it, we don't need to give you a warning.

PR has been updated.

goodell:
> * Change the default to build optimized

I'm very happy to see this change, I've found it to be surprising since basically forever.

ggouaillardet:
:+1: 
i would also have ```mpirun --version``` and the likes clearly state this is a debug build

gpaulsen:
:+1:

jsquyres:
Point made 15 Mar 2016 webex: it would be nice to leave `--enable-picky` enabled by default for developer builds.

Ok, will do.

jsquyres:
PR updated per the above comment.

jsquyres:
bot:retest

jsquyres:
@miked-mellanox Is the Mellanox Jenkins busted at the moment?

PR #1422, opened 2016-03-02
Refactor the request completion

bosilca:
The current code to handle request completions (for both single and multi-requests calls) is complex and highly inefficient. It has a big impact on our message rate for single threaded applications, and prevents  good performance on multi-threaded cases (global lock for completion, wakes all threads upon each request completions, and many more issues).

This patch fixes all cases. First, it refactors the completion code and the link between the completion and opal_progress. It introduces a new synchronization primitive, that is the completion status of the request (COMPLETE, PENDING or a synchronization primitive). As a result the request status is now handled via atomic operations, at least one (to mark the request as completed) and at most 2 per request (one to set the synchronization primitive and one to mark it as complete). Second, refactor all wait* and test* functions to reduce the number of iterations over the array of requests (if any).

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -71,7 +71,6 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
 {
     mca_pml_ob1_recv_request_t* recvreq = *(mca_pml_ob1_recv_request_t**)request;
 
-    OPAL_THREAD_LOCK(&ompi_request_lock);
     if(false == recvreq->req_recv.req_base.req_free_called){
 
         recvreq->req_recv.req_base.req_free_called = true;
@@ -92,7 +91,6 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
         }
 
     }
-    OPAL_THREAD_UNLOCK(&ompi_request_lock);
     *request = MPI_REQUEST_NULL;
     return OMPI_SUCCESS;
 }
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -71,25 +71,28 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
 {
     mca_pml_ob1_recv_request_t* recvreq = *(mca_pml_ob1_recv_request_t**)request;
 
-    assert(false == recvreq->req_recv.req_base.req_free_called);
-    recvreq->req_recv.req_base.req_free_called = true;
+    OPAL_THREAD_LOCK(&ompi_request_lock);
+    if(false == recvreq->req_recv.req_base.req_free_called){
 
-    PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_NOTIFY,
-                             &(recvreq->req_recv.req_base), PERUSE_RECV );
+        recvreq->req_recv.req_base.req_free_called = true;
+        PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_NOTIFY,
+                                 &(recvreq->req_recv.req_base), PERUSE_RECV );
 
-    if( true == recvreq->req_recv.req_base.req_pml_complete ) {
-        /* make buffer defined when the request is compeleted,
-           and before releasing the objects. */
-        MEMCHECKER(
-            memchecker_call(&opal_memchecker_base_mem_defined,
-                            recvreq->req_recv.req_base.req_addr,
-                            recvreq->req_recv.req_base.req_count,
-                            recvreq->req_recv.req_base.req_datatype);
-        );
+        if( true == recvreq->req_recv.req_base.req_pml_complete ) {
+            /* make buffer defined when the request is compeleted,
+               and before releasing the objects. */
+            MEMCHECKER(
+                memchecker_call(&opal_memchecker_base_mem_defined,
+                                recvreq->req_recv.req_base.req_addr,
+                                recvreq->req_recv.req_base.req_count,
+                                recvreq->req_recv.req_base.req_datatype);
+            );
 
-        MCA_PML_OB1_RECV_REQUEST_RETURN( recvreq );
+            MCA_PML_OB1_RECV_REQUEST_RETURN( recvreq );
+        }
+
     }
-
+    OPAL_THREAD_UNLOCK(&ompi_request_lock);
     *request = MPI_REQUEST_NULL;
     return OMPI_SUCCESS;
 }
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -71,9 +71,7 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
 {
     mca_pml_ob1_recv_request_t* recvreq = *(mca_pml_ob1_recv_request_t**)request;
 
-    assert( false == recvreq->req_recv.req_base.req_free_called );
-
-    OPAL_THREAD_LOCK(&ompi_request_lock);
+    assert(false == recvreq->req_recv.req_base.req_free_called);
     recvreq->req_recv.req_base.req_free_called = true;
 
     PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_NOTIFY,
@@ -92,7 +90,6 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
         MCA_PML_OB1_RECV_REQUEST_RETURN( recvreq );
     }
 
-    OPAL_THREAD_UNLOCK(&ompi_request_lock);
     *request = MPI_REQUEST_NULL;
     return OMPI_SUCCESS;
 }
@@ -126,14 +123,12 @@ static int mca_pml_ob1_recv_request_cancel(struct ompi_request_t* ompi_request, 
     request->req_recv.req_base.req_pml_complete = true;
     OB1_MATCHING_UNLOCK(&ob1_comm->matching_lock);
 
-    OPAL_THREAD_LOCK(&ompi_request_lock);
     ompi_request->req_status._cancelled = true;
     /* This macro will set the req_complete to true so the MPI Test/Wait* functions
      * on this request will be able to complete. As the status is marked as
      * cancelled the cancel state will be detected.
      */
     MCA_PML_OB1_RECV_REQUEST_MPI_COMPLETE(request);
-    OPAL_THREAD_UNLOCK(&ompi_request_lock);
     /*
      * Receive request cancelled, make user buffer accessible.
      */
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.h
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.h
@@ -158,47 +158,49 @@ recv_request_pml_complete(mca_pml_ob1_recv_request_t *recvreq)
 {
     size_t i;
 
-    assert(false == recvreq->req_recv.req_base.req_pml_complete);
+    if(false == recvreq->req_recv.req_base.req_pml_complete){
+        OPAL_THREAD_LOCK(&ompi_request_lock);
 
-    if(recvreq->req_recv.req_bytes_packed > 0) {
-        PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_XFER_END,
-                &recvreq->req_recv.req_base, PERUSE_RECV );
-    }
-
-    for(i = 0; i < recvreq->req_rdma_cnt; i++) {
-        struct mca_btl_base_registration_handle_t *handle = recvreq->req_rdma[i].btl_reg;
-        mca_bml_base_btl_t *bml_btl = recvreq->req_rdma[i].bml_btl;
-
-        if (NULL != handle) {
-            mca_bml_base_deregister_mem (bml_btl, handle);
+        if(recvreq->req_recv.req_bytes_packed > 0) {
+            PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_XFER_END,
+                    &recvreq->req_recv.req_base, PERUSE_RECV );
         }
-    }
-    recvreq->req_rdma_cnt = 0;
 
-    if(true == recvreq->req_recv.req_base.req_free_called) {
-        if( MPI_SUCCESS != recvreq->req_recv.req_base.req_ompi.req_status.MPI_ERROR ) {
-            ompi_mpi_abort(&ompi_mpi_comm_world.comm, MPI_ERR_REQUEST);
+        for(i = 0; i < recvreq->req_rdma_cnt; i++) {
+            struct mca_btl_base_registration_handle_t *handle = recvreq->req_rdma[i].btl_reg;
+            mca_bml_base_btl_t *bml_btl = recvreq->req_rdma[i].bml_btl;
+
+            if (NULL != handle) {
+                mca_bml_base_deregister_mem (bml_btl, handle);
+            }
         }
-        MCA_PML_OB1_RECV_REQUEST_RETURN(recvreq);
-    } else {
-        /* initialize request status */
-        recvreq->req_recv.req_base.req_pml_complete = true;
-        recvreq->req_recv.req_base.req_ompi.req_status._ucount =
-            recvreq->req_bytes_received;
-        if (recvreq->req_recv.req_bytes_packed > recvreq->req_bytes_expected) {
+        recvreq->req_rdma_cnt = 0;
+
+
+        if(true == recvreq->req_recv.req_base.req_free_called) {
+            if( MPI_SUCCESS != recvreq->req_recv.req_base.req_ompi.req_status.MPI_ERROR ) {
+                ompi_mpi_abort(&ompi_mpi_comm_world.comm, MPI_ERR_REQUEST);
+            }
+            MCA_PML_OB1_RECV_REQUEST_RETURN(recvreq);
+        } else {
+            /* initialize request status */
+            recvreq->req_recv.req_base.req_pml_complete = true;
             recvreq->req_recv.req_base.req_ompi.req_status._ucount =
-                recvreq->req_recv.req_bytes_packed;
-            recvreq->req_recv.req_base.req_ompi.req_status.MPI_ERROR =
-                MPI_ERR_TRUNCATE;
+                recvreq->req_bytes_received;
+            if (recvreq->req_recv.req_bytes_packed > recvreq->req_bytes_expected) {
+                recvreq->req_recv.req_base.req_ompi.req_status._ucount =
+                    recvreq->req_recv.req_bytes_packed;
+                recvreq->req_recv.req_base.req_ompi.req_status.MPI_ERROR =
+                    MPI_ERR_TRUNCATE;
+            }
+            if (OPAL_UNLIKELY(recvreq->local_handle)) {
+                mca_bml_base_deregister_mem (recvreq->rdma_bml, recvreq->local_handle);
+                recvreq->local_handle = NULL;
+            }
+            MCA_PML_OB1_RECV_REQUEST_MPI_COMPLETE(recvreq);
         }
-        if (OPAL_UNLIKELY(recvreq->local_handle)) {
-            mca_bml_base_deregister_mem (recvreq->rdma_bml, recvreq->local_handle);
-            recvreq->local_handle = NULL;
-        }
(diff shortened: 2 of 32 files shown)


ompi/mca/pml/ob1/pml_ob1_recvreq.c
```diff
@@ -92,7 +90,6 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
         MCA_PML_OB1_RECV_REQUEST_RETURN( recvreq );
     }
```
hppritcha: why are the calls to OPAL_THREAD_LOCK removed?

ompi/mca/pml/ob1/pml_ob1_recvreq.h
```diff
@@ -196,9 +195,11 @@ recv_request_pml_complete(mca_pml_ob1_recv_request_t *recvreq)
             mca_bml_base_deregister_mem (recvreq->rdma_bml, recvreq->local_handle);
             recvreq->local_handle = NULL;
         }
+#if OPAL_ENABLE_MULTI_THREADS
+        opal_atomic_wmb();
+#endif
         MCA_PML_OB1_RECV_REQUEST_MPI_COMPLETE(recvreq);
     }
```
hppritcha: here too, how to do we know its safe to remove the serialization?

hppritcha:
bot:retest

hppritcha:
@jsquyres I think this is too risky for 2.0.0

hppritcha:
@bosilca please have your student rebase down to a few commits.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: Because there is no locking anymore. Everything related to the synchronization is now handled down in the sync_wait, via atomic operations.

ompi/mca/pml/ob1/pml_ob1_recvreq.h
bosilca: Because this serialization was against a global mutex. All requests are now serialized based on their corresponding waiting context (via the sync_wait).

bosilca:
Performance tests show almost 400% message rate increase over vader with 4 threads per process. Increasing the number of threads only increases this performance boost.

jsquyres:
@hppritcha I agree that this is too big for v2.0.0 -- I think v2.0.1 at the earliest.

@bosilca What's the merge commit for in the middle?  That does not seem correct at all.

Also, it would be really good to clean up this patch set a bunch -- there's a whole lot of intermediate commits (e.g., "foo" is added in commit A, but then removed in commit B).  Can this be squashed to a smaller number of logical commits?  (it probably doesn't make sense to squash it down to 1 commit, but something less than 30 would be good...)

jsquyres:
@bosilca tells me that this PR is not yet ready -- they're still working on the code.  For the moment, I'm deferring this to v2.1.0.

jsquyres:
@bosilca Does this PR represent all the commits that went into master?  The merge commit near the bottom of the list seems weird / wrong...

ibm-ompi:
Test passed.

ibm-ompi:
Test passed.

ibm-ompi:
Test passed.

jsquyres:
bot:retest

jsquyres:
@Di0gen @jladd-mlnx UCX builds still seem to be broken in the Mellanox Jenkins.

jsquyres:
Per discussion on the 2015-05-24 webex, we decided:

* This is a blocker performance bug.  We need it for v2.0.0.  We/the OMPI community acknowledge that this will further delay v2.0.0.
* Let's merge this to master and see how it does.

jsquyres:
@bosilca After the webex conversation, however, I see that the Mellanox Jenkins failed an OSHMEM test with this PR.  It looks like a legit failure -- can you please investigate?

bosilca:
@jsquyres looking at the jenkins failure indicates the issue as being in the RTE, more precisely in the OOB. Here is the only usable stack I can get from the Mellanox output.

```
10:09:19 [jenkins01:02266] *** Process received signal ***
10:09:19 [jenkins01:02266] Signal: Segmentation fault (11)
10:09:19 [jenkins01:02266] Signal code:  (0)
10:09:19 [jenkins01:02266] Failing at address: 0xb20000008d9
10:09:19 oshrun: Forwarding signal 18 to job
10:09:19 [jenkins01:02266] [ 0] /lib64/libpthread.so.0[0x3d6980f710]
10:09:19 [jenkins01:02266] [ 1] /lib64/libc.so.6(__select+0x33)[0x3d690e15e3]
10:09:19 [jenkins01:02266] [ 2] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/lib/openmpi/mca_oob_tcp.so(+0xc1ca)[0x7ffff55411ca]
10:09:19 [jenkins01:02266] [ 3] /lib64/libpthread.so.0[0x3d698079d1]
10:09:19 [jenkins01:02266] [ 4] /lib64/libc.so.6(clone+0x6d)[0x3d690e8b6d]
```

jsquyres:
@bosilca Before this merge, the problem with the oshmem test was not occurring.  I know the backtrace shows an OOB function, but it still seems like this PR is somehow involved with making the problem occur.  E.g., look at the Mellanox Jenkins on https://github.com/open-mpi/ompi/pull/1704 -- it's a PR that is only scripting changes, but has a very similar failure signature in the same test.  It's not in the OOB, but it's in an event loop:

```
04:10:00 [jenkins01:25237] *** Process received signal ***
04:10:00 [jenkins01:25237] Signal: Segmentation fault (11)
04:10:00 [jenkins01:25237] Signal code:  (0)
04:10:00 [jenkins01:25237] Failing at address: 0xb200000628b
04:10:00 [jenkins01:25237] [ 0] /lib64/libpthread.so.0[0x3d6980f710]
04:10:00 [jenkins01:25237] [ 1] /lib64/libc.so.6(epoll_wait+0x33)[0x3d690e9163]
04:10:00 [jenkins01:25237] [ 2] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/lib/libopen-pal.so.0(+0x968f6)[0x7ffff7a548f6]
04:10:00 [jenkins01:25237] [ 3] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/lib/libopen-pal.so.0(opal_libevent2022_event_base_loop+0x1f6)[0x7ffff7a5a0b6]
04:10:00 [jenkins01:25237] [ 4] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/bin/oshrun[0x401abb]
04:10:00 [jenkins01:25237] [ 5] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/bin/oshrun[0x401044]
04:10:00 [jenkins01:25237] [ 6] /lib64/libc.so.6(__libc_start_main+0xfd)[0x3d6901ed1d]
04:10:00 [jenkins01:25237] [ 7] /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/bin/oshrun[0x400f69]
04:10:00 [jenkins01:25237] *** End of error message ***
```

@jladd-mlnx Going to need some Mellanox help on this one -- it seems to be particular to MXM and Oshmem.  Can you provide more debugging information?

rhc54:
See if it is during finalize - we've had problems before where oshmem didn't properly let ORTE's progress thread finalize

jladd-mlnx:
@jsquyres I'm getting a little cranky here. Not sure why people are merging when tests are failing. I cannot just drop everything and start debugging on a moment's notice. We have our own deadline to meet this week. The failure is coming from a hello world OSHMEM test.

rhc54:
@jladd-mlnx People are merging because Jenkins has proven to be rather unreliable. It isn't uncommon for the errors to come from something changed or broken on the Mellanox side. As a result, we are all somewhat skeptical of those results, especially if the other tests all come up "good".

I'd suggest we just leave this broken until you have time to help debug it. No real harm is being done. We can discuss how to move forward with 2.0 as a separate question.

jladd-mlnx:
@rhc54 Completely disagree. Not sure how your memory can be so selective. I can get to it in 2 weeks.

rhc54:
I suggest you check your logs, but that's okay. If oshmem isn't that important, then we can just disable it on master for now so it won't disrupt things.

thananon:
@jladd-mlnx I can pursue this if I can get into the node and see what happened there.

jladd-mlnx:
@thananon I will get you a guest account. What's your email address?

jsquyres:
@jladd-mlnx @thananon is just starting at Cisco (summer intern; works for @bosilca at UTK).  He's going to try to build and replicate the oshmem error on our cluster.

hppritcha:
@jladd-mlnx i also would like to reproduce the problem on one of our MLNX systems.  Looking at the command line options, it looks like we need to pick up yoda spml and mxm mtl correct?  No dependence on UCX is apparent given the command line options.  I want to rule out having to use UCX to reproduce the problem.  Is UCX needed do you think to reproduce?

hjelmn:
One problem with oshmem. req_complete field is no longer a boolean. Initialize to REQUEST_PENDING not false.

Mellanox should have been following this PR. Been a long time coming.

jladd-mlnx:
@hppritcha I sent you and @jsquyres a private note. If you send me your public keys, you can have access to the Jenkins server. UCX is not needed.

hjelmn:
Nevermind. oshmem has a different request structure. It looks like the old ompi one. Does oshmem cast between ompi_request_t and oshmem_request_t anywhere? That will not work on master.

jladd-mlnx:
@hjelmn Not sure what you mean here, Nathan. Are you saying there is an action item that we missed regarding this PR? I've reviewed the minutes and I don't see anything regarding a commitment on our end.

hjelmn:
@jladd-mlnx This PR has come up multiple times. Was supposed to be merged some time ago but slipped due to being slipped to 2.1.0. Should keep track of these sorts of things. I certainly do. I checked all my code to make sure it would work.

jladd-mlnx:
@hjelmn This PR is your project, Nathan. Not ours. And we do keep track; with Jenkins. We can't fix your bugs.

hjelmn:
Try changing the req_complete field in your oshmem request structure to a volatile void *. That will make oshmem_request_t match ompi_request_t again.

hjelmn:
@jladd-mlnx Not my PR. And oshmem is not my code. I do all my checking with it disabled. Not my job to keep it working.

hjelmn:
I got tired of tracking issues related to poor coding in oshmem layer.

bosilca:
As long the OSHMEM and OMPI requests are only used in their own context there should be no issue over the fact that they are not identical. A quick grep does not seem to indicate any mismatch.

bosilca:
Let's chill down. If any fault it is ours for not running the OSHMEM tests. In defense I have to say we never imagined that the OSHMEM layer is using the OMPI request. If the Mellanox folks can confirm this we can help to port the OMPI improvements in OSHMEM.

rhc54:
Something is very off here, guys - look at the stack trace. The error is in oshrun, not in an application process. Unless Mellanox has done something very odd, there is no way that a change in the MPI request can cause oshrun to fail.

hjelmn:
@rhc54 The question is whether the application processes have started or not. They certainly haven't produced any output.

hjelmn:
@rhc54 BTW, oshrun crashed because the mellanox jenkins sends a SEGV on timeout.

jsquyres:
Oh, wow -- we've been chasing a SEGV, but the SEGV was coming from Mellanox's scripts itself (note the parameter to the "timeout" command):

```
02:59:18 + taskset -c 16,17 timeout -s SIGSEGV 10m /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/bin/oshrun -np 8 --bind-to core -x SHMEM_SYMMETRIC_HEAP_SIZE=1024M --mca btl_openib_if_include mlx4_0:1 -x MXM_RDMA_PORTS=mlx4_0:1 -x UCX_NET_DEVICES=mlx4_0:1 -x UCX_TLS=rc,cm --mca rmaps_base_dist_hca mlx4_0:1 --mca sshmem_verbs_hca_name mlx4_0:1 --mca spml ikrit -mca pml cm -mca mtl mxm /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/examples/hello_oshmem
03:09:19 [jenkins01:02266] *** Process received signal ***
03:09:19 [jenkins01:02266] Signal: Segmentation fault (11)
```

So it looks like the real issue is a hang.

jladd-mlnx:
@jsquyres #1716

jsquyres:
@jladd-mlnx Yep -- fingers crossed!  Can we change the timeout signal to something other than SEGV?  That misled us for a while today; it might be helpful in the future to not have to remember that the SEGV isn't actually a SEGV.  Thanks!

miked-mellanox:
@jsquyres - the reason we send SEGV is to *make* mpirun dump the stacktrace.
Is there any way to make it happen w/o sending SEGV? maybe some mca param to dump stack after X min?

hjelmn:
SIGABRT?

miked-mellanox:
SIGABRT is not interceptable by mpirun and just will kill it w/o stacktrace.

jsquyres:
@jladd-mlnx Would it be better to grab a stack trace with something like `pstack`?  E.g., here's `pstack` output from the PID of my emacs:

```
$ pstack 28231
Thread 2 (Thread 0x2aaab13aa700 (LWP 28232)):
#0  0x0000003491e0e75d in read () from /lib64/libpthread.so.0
#1  0x0000003492e4110b in ?? () from /lib64/libglib-2.0.so.0
#2  0x0000003492e691b4 in ?? () from /lib64/libglib-2.0.so.0
#3  0x0000003491e079d1 in start_thread () from /lib64/libpthread.so.0
#4  0x00000034916e8b5d in clone () from /lib64/libc.so.6
Thread 1 (Thread 0x2aaaaaee2b60 (LWP 28231)):
#0  0x00000034916e17de in pselect () from /lib64/libc.so.6
#1  0x00000000005b2ce2 in xg_select ()
#2  0x0000000000582d33 in wait_reading_process_output ()
#3  0x00000000004dbd7f in read_decoded_event_from_main_queue.isra.39 ()
#4  0x00000000004dff37 in read_char ()
#5  0x00000000004e0d0f in read_key_sequence.constprop.41 ()
#6  0x00000000004e2900 in command_loop_1 ()
#7  0x0000000000543a1e in internal_condition_case ()
#8  0x00000000004d4f6e in command_loop_2 ()
#9  0x000000000054392b in internal_catch ()
#10 0x00000000004d9497 in recursive_edit_1 ()
#11 0x00000000004d97ad in Frecursive_edit ()
#12 0x00000000004d42a5 in main ()
```

jsquyres:
@jladd-mlnx Oops -- I mean to tag the prior comment to both @jladd-mlnx and @miked-mellanox. I'm apparently still not awake yet.  Sorry!

miked-mellanox:
the problem is how to detect hang. mpirun has progress thread which can dump stack after user-provided time limit. With your proposal - one need to develop watchdog client and integrate it - it is very error prune IMO.

jladd-mlnx:
We currently have no plans to change the script.

jladd-mlnx:
For the public record, there were several inaccurate and misleading comments posted to this ticket. OSHMEM has nothing to do with this failure. A hang in the CM PML was triggered during an OSHMEM hello world test because OSHMEM init uses MPI collectives and, hence, the PML to bootstrap itself once MPI is init-ed.  The root cause of the hang was because of a lack of support in the CM PML for the new request handling. The combination of OSHMEM with CM PML MXM MTL and IKRIT SML just happened to trigger the hang.  A fix for this issue was made in #1716, however, subsequent testing has exposed another bug in the OpenIB BTL that needs to be fixed as well. 

Thanks to Mellanox Jenkins for catching this.

rhc54:
Just a little further clarification: part of the reason for the confusion was the use of an induced segfault in the Mellanox Jenkins script to initiate a timeout shutdown of oshrun. This provides a false positive in a portion of the code totally unrelated to the submitted PR, leaving the impression that the test, and not the code, was the root cause.

It appears we are unable to reach agreement on how best to avoid this going forward. It therefore remains a reasonable possibility that this problem will repeat in the future, pending agreement on a better solution.

jladd-mlnx:
The resolution going forward is the following; any community member is more than welcome to access the Mellanox Jenkins server and is encouraged to do so. All that I need is your public RSA key, and I will be happy to add it to the list of authorized keys. Once added, you will simply need to:

`$ssh hpcguest@bgate.mellanox.com`

And then to the Jenkins server:

`$ssh jenkins01`

This takes Mellanox out of the critical path of unrelated projects.

rhc54:
Good to know, but not likely to be useful, to be honest. Many of us are in the same situation as you and don't have the time to chase apparently unrelated issues, and there are legal concerns for some members regarding accessing Mellanox machines. So I don't think that will provide a practical solution.

Probably best for the community to discuss this offline.

jladd-mlnx:
The offer stands. It's the best we can do.

nysal:
@miked-mellanox Probably using a tool like PADB (http://padb.pittman.org.uk/) to collect the stack trace after waiting for a timeout period is one option. I have used the code from https://code.google.com/archive/p/padb/source and that seems to work.

miked-mellanox:
yep. this tool does not fit jenkins framework.

One need to write complicated logic to detect mpirun hang and fork padb to extract stack and properly communicate status to jenkins.

rhc54:
Yeah, I'm not sure it's worth all that effort. The problem we really need to solve is to tell us that the test hung - getting a stack trace from mpirun won't tell us much about why as mpirun isn't the source of the problem. I think a simple error statement that the test timed out like we do for MTT would be more useful and create less "false positive" confusion.

miked-mellanox:
yep, ``` timeout -s SEGV mpirun" can be replaced w/ SIGKILL and jenkins will reported (in the console) that test killed. The next logical questions from community will be - where it hangs? 
So, having mpirun print stacktrace after some internal timeout (if provided) is probably best automated way to satisfy jenkins report and developers questions.

rhc54:
Yeah, we should think about this a bit. What we really want to know is that we timed out because the applications never terminated, as opposed to we timed out because of a problem in mpirun itself. The problem with just generating an mpirun stack trace is that it won't tell you anything if the problem is in the application. All it can show is that mpirun is merrily waiting for something to happen.

Perhaps the better solution is to provide some way of asking mpirun to output a state report - i.e., what does it think the current state of the job is? Does it think all procs terminated and it is just hung trying to exit? Or does it think procs are still running?

Another thing we could get is a simple "ps" output. Are there procs still running? Or have all the apps exited? This should be simple to get, and would provide more info than a stack trace.

If we can do some of that, it would be helpful to avoid the confusion.

miked-mellanox:
yep,something like that would be very helpful to all.

achieving same effect with jenkins and external watchdog processes will make things worse, because it is hard to orchestrate inside jenkins single process.


``` shell
% mpirun -dump-stack-from-all-ranks-and-exit-with-error-after-time 10m imb.exe
```

rhc54:
Might be doable - I can take a crack at ti.

jsquyres:
@rhc54 Perhaps this is something you could daisy chain off the $MPIEXEC_TIMEOUT functionality...?

rhc54:
😆 that is exactly what I am coding right this second

rhc54:
What do you think of this:

```shell
$ mpirun -n 4 --map-by node -timeout 2 --report-state-on-timeout ./mpi_spin
The user-provided time limit for job execution has been
reached:

  MPIEXEC_TIMEOUT: NULL seconds

The job will now be aborted. Please check your code and/or
adjust/remove the job execution time limit (as specified
by MPIEXEC_TIMEOUT in your environment or -timeout on the cmd line).
DATA FOR JOB: [42410,0]
	Num apps: 1	Num procs: 2	JobState: ALL DAEMONS REPORTED	Abort: False
	Num launched: 0	Num reported: 2	Num terminated: 0

	Procs:
		Rank: 0	Node: bend001	PID: 10484	State: RUNNING	ExitCode 0
		Rank: 1	Node: bend003	PID: 10489	State: RUNNING	ExitCode 0

DATA FOR JOB: [42410,1]
	Num apps: 1	Num procs: 4	JobState: SYNC REGISTERED	Abort: False
	Num launched: 4	Num reported: 4	Num terminated: 0

	Procs:
		Rank: 0	Node: bend001	PID: 10490	State: SYNC REGISTERED	ExitCode 0
		Rank: 1	Node: bend003	PID: 28765	State: SYNC REGISTERED	ExitCode 0
		Rank: 2	Node: bend001	PID: 10491	State: SYNC REGISTERED	ExitCode 0
		Rank: 3	Node: bend003	PID: 28766	State: SYNC REGISTERED	ExitCode 0
```

PR #1423, opened 2016-03-02
Fix registration of error handlers thru the pmix120 component.

rhc54:
A thread-shift operation was hanging on the sync_event_base, which made it dependent on someone calling opal_progress. Unfortunately, a process in "sleep" or spinning outside the MPI library won't do that, and so we never complete errhandler registration.


lanl-ompi:
Test FAILed.

rhc54:
Sorry folks - want this to get into the nightly tarball, and it has been over 3 hours now.

PR #1432, opened 2016-03-07
Fix misc memchecker issues

ggouaillardet:
Fix misc memchecker issues

Thanks Florian Lindner for the report.

### DIFF
--- a/ompi/mpi/c/isend.c
+++ b/ompi/mpi/c/isend.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2006-2007 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2013      Los Alamos National Security, LLC.  All rights
  *                         reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -82,9 +82,13 @@ int MPI_Isend(const void *buf, int count, MPI_Datatype type, int dest,
 
     OPAL_CR_ENTER_LIBRARY();
 
-    MEMCHECKER (
-        memchecker_call(&opal_memchecker_base_mem_noaccess, buf, count, type);
-    );
+    /*
+     * today's MPI standard mandates the send buffer remains accessible during the send operation
+     * hence memchecker cannot mark buf as non accessible, but it might mark buf as read-only in
+     * order to trap end user errors. Unfortunatly valgrind does not support marking buffers as read-only,
+     * so there is pretty much nothing we can do here.
+     */
+     
     rc = MCA_PML_CALL(isend(buf, count, type, dest, tag,
                             MCA_PML_BASE_SEND_STANDARD, comm, request));
     OMPI_ERRHANDLER_RETURN(rc, comm, rc, FUNC_NAME);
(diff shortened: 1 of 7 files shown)


ggouaillardet:
@bosilca can you please have a look at this ?

ompi/mpi/c/isend.c
```diff
@@ -82,11 +82,11 @@ int MPI_Isend(const void *buf, int count, MPI_Datatype type, int dest,
 
     OPAL_CR_ENTER_LIBRARY();
 
+    rc = MCA_PML_CALL(isend(buf, count, type, dest, tag,
+                            MCA_PML_BASE_SEND_STANDARD, comm, request));
     MEMCHECKER (
         memchecker_call(&opal_memchecker_base_mem_noaccess, buf, count, type);
     );
```
bosilca: @ggouaillardet we should remove this code, it is incorrect based on today's MPI standard.

bosilca:
Overall :+1:  but I think we should remove the lines of code where I put a comment (the standard mandate that the sendbuffer remains accessible during the MPI_Send operation).

ggouaillardet:
ok, will do
in a previous mail, you said the send buffer should be marked as read only (so memchecker can trap user errors), but i could not find how to do that with valgrind. is this even possible ?

bosilca:
@ggouaillardet I was under the impression that the valgrind WRITABLE flag was meant for that. However, after your answer I went back and read the valgrind documentation and I reached the same conclusion as you, it is not possible to mark the buffer read-only.

ggouaillardet:
@bosilca i fixed that, added a comment about marking the buffer read-only (maybe one day valgrind will allow that ...) and pushd an other commit so memchecker does nothing when MPI_IN_PLACE is used as the send buffer of a neighborhood collective

ggouaillardet:
@bosilca i noticed the ompi will fail with ```MPI_ERR_ARG``` if ```MPI_IN_PLACE``` is used, but the man page only says
```
The MPI_IN_PLACE option for sendbuf is not meaningful for this function.
```
shall i update the man page and make it more explicit ?
for example
```
MPI_IN_PLACE is not a valid sendbuf for this function.
```

ggouaillardet:
fwiw https://sourceforge.net/p/valgrind/mailman/message/34919133/

jsquyres:
@ggouaillardet What's the status of this PR?

ibm-ompi:
Test passed.

ggouaillardet:
i signed-off the commits, and rebased.
the man pages use the wording from the MPI standard, so i guess there is no need rewording them.

ompi/mpi/c/isend.c
```diff
@@ -82,11 +82,14 @@ int MPI_Isend(const void *buf, int count, MPI_Datatype type, int dest,
...
-    MEMCHECKER (
-        memchecker_call(&opal_memchecker_base_mem_noaccess, buf, count, type);
-    );
     rc = MCA_PML_CALL(isend(buf, count, type, dest, tag,
                             MCA_PML_BASE_SEND_STANDARD, comm, request));
+    /*
+     * today's MPI standard mandates the send buffer remains accessible during the send operation
+     * hence memchecker cannot mark buf as non accessible, but it might mark buf as read-only in
+     * order to trap end user errors
+     */
```
jsquyres: Quick note on this comment: are we actually marking the buffer as read-only somewhere else?

ggouaillardet: no we do not.
fwiw, ```valgrind``` does not know ```read-only```, and when i asked, they had no plan to have this feature. are you aware of any other tool that would correctly handle read-only memory regions ?

jsquyres: Ah, I gotcha.  No, I'm unaware of a tool that does this (but I don't closely follow tools like this).

Could you slightly revise the comment to explain that?  I.e., we leave this comment here because Valgrind can't currently do it, but if it does someday, it would be nice to use here, yadda yadda yadda...  

I say this because, as the comment is written now, it implies that we *could* do this (and may actually do it elsewhere).  Just making the comment a little clearer for our Future Selves would be awesome.  Thanks!

ggouaillardet:
i fixed the comment in ```MPI_ISend()```, now merging into master

PR #1436, opened 2016-03-08
Ensure the process name is positive when using direct launch

### LINKED ISSUE #1425: External PMIx server support is broken
@ggouaillardet @rhc54 @jladd-mlnx 
The following commit:
https://github.com/open-mpi/ompi/commit/b55b9e6aeee96fe76768b0438fad11ffb4e49fe4
Brakes launch through the SLURM plugin the assert:
https://github.com/open-mpi/ompi/blob/master/ompi/proc/proc.h#L399
is triggered because jobid was created from hash function:
https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/pmix112/pmix1_client.c#L122

rhc54:
Fixes #1425


artpol84:
We need to PR this to 2.0

PR #1448, opened 2016-03-10
MCA component alias support (alternative approach)

hjelmn:


hjelmn:
@jsquyres, @bosilca Comments? Which approach is better?

bosilca:
I wonder if it is not more flexible to expose the aliasing mechanism through the MCA framework, and configure it in the Open MPI /etc file? I can see how something like the following could be useful:
```
alias_btl_tcp = BTCP
alias_btl_openib = btl_ib
alias_btl_shmem = btl_sm,btl_vader
BTCP_if_include = eth0
```
We can provide some aliasing rules and the savvy sysadmin or users can provide their own.

hjelmn:
Thats an interesting idea. Let me think about how it could be implemented.

jsquyres:
@hjelmn @bosilca A minor point: I'd like to see a space (" ") instead of an underscore ("_") after "alias":

```ini
alias btl_tcp = BTCP
alias btl_openib = ib
```

This would then still allow an MCA param that just happens to have the prefix `alias_`.  It's also slightly clearer for humans to read (IMHO).

bosilca:
Wouldn't this requires changes in the lexer (we only support x = y syntax)?

hjelmn:
Yeah, wouldn't be too hard to add support for ```alias foo=bar```. The MCA variable system could then expose aliases for selection. I will try coding that up and see what it ends up looking like sometime in the next week or two.

jsquyres:
Yeah, my point was that "x = y" doesn't necessarily imply whether "x" or "y" contain spaces.  IIRC, our current lexer supports spaces in the "x" value...?  If not, it might not be too hard to add it.

bosilca:
Our current lexer does not support spaces in the x value. In fact it is not the lexer issue, but the parser, which only accepts a trivially simple "x = y" or "-x ... = ..." language. I just checked and the lexer correctly tokenize the "alias x = y" line, but the parse_line correctly rejects it.

ibm-ompi:
Test passed.

PR #1454, opened 2016-03-15
Fix memory corruption in orte-ps

nysal:
I was trying to get PADB to work, which uses orte-ps and glibc complained about double free's.


nysal:
@rhc54 Could you please review?

rhc54:
:+1:

PR #1458, opened 2016-03-16
If the server loses connection with a local client, check to see if a…

rhc54:
…ny active collectives were impacted and adjust if so. This will avoid having the daemons hang in a collective that can never complete.

@abouteiller @bosilca I don't have an easy way of testing this, and none of the existing Jenkins tests will exercise this code path. Can you folks "pull" this change and see if it solves the problem?


hjelmn:
bot:retest

lanl-ompi:
Test FAILed.

rhc54:
Closing this out to integrate it with other upstream changes

PR #1465, opened 2016-03-17
configure: re-enable built-in atomic support

hjelmn:
This commit removes an erroneous else statement from the OSX built-in
atomics check. The else branch sets the built-in atomics support to
BUILTIN_NO if either opal_cv_asm_builtin is not BUILTIN_NO or OSX
atomics support is disabled.


hjelmn:
@bosilca Whoops, I found this problem when trying to figure out why gcc 5.3 fails the opal_fifo test. I will have to check if any branches are affected.

bosilca:
Nice catch. :+1:

PR #1473, opened 2016-03-18
Topic/fortran attributes

ggouaillardet:


ggouaillardet:
@dalcinl @jsquyres this is a snapshot of my branch

it works on sparc with/without -i8 fortran equivalent

i still have to update the doc ...

Refers open-mpi/ompi-release#1018

ggouaillardet:
:bot:retest

ompi/mpi/fortran/mpif-h/keyval_create_f.c
```diff
@@ -70,31 +70,34 @@ static const char FUNC_NAME[] = "MPI_keyval_create_f";
...
     /* Set the "F77_OLD" bit to denote that the callbacks should use
        the old MPI-1 INTEGER-parameter functions (as opposed to the
        new MPI-2 INTEGER(KIND=MPI_ADDRESS_KIND)-parameter
        functions). */
 
     ret = ompi_attr_create_keyval_fint(COMM_ATTR, copy_fn, del_fn,
-                                       keyval, *extra_state,
-                                       OMPI_KEYVAL_F77 | OMPI_KEYVAL_F77_MPI1,
+                                       &keyval, *extra_state,
+                                       OMPI_KEYVAL_F77,
```
jsquyres: @ggouaillardet Why did you remove `OMPI_KEYVAL_F77_MPI1`?  This seems to contradict the comment right about this call.

ompi/attribute/attribute.c
```diff
@@ -270,7 +270,7 @@ do { \
         else { \
             MPI_Aint attr_val = translate_to_fortran_mpi2(attribute); \
             (*((keyval_obj->delete_attr_fn).attr_mpi2_fortran_delete_fn)) \
-                (&(((ompi_##type##_t *)object)->attr_##type##_f), \
+                (&attr_##type##_f,                                        \
```
jsquyres: Seems like this hunk should be part of 2dba1ec26fa95c2d63434ffe5499998b219a4a7c

ompi/attribute/attribute.c
```diff
@@ -317,9 +319,11 @@ do { \
         /* MPI-2 Fortran-style */ \
         else { \
             MPI_Aint in, out;                        \
+            MPI_Fint attr_##type##_f;                \
             in = translate_to_fortran_mpi2(in_attr); \
+            attr_##type##_f = OMPI_INT_2_FINT(((ompi_##type##_t *)old_object)->attr_##type##_f); \
             (*((keyval_obj->copy_attr_fn).attr_mpi2_fortran_copy_fn)) \
-                (&(((ompi_##type##_t *)old_object)->attr_##type##_f), \
+                (&attr_##type##_f, \
```
jsquyres: Seems like all these hunks should be part of 2dba1ec26fa95c2d63434ffe5499998b219a4a7c, too.  I think you can either squash 2dba1ec26fa95c2d63434ffe5499998b219a4a7c and this commit together, or separate out all the `attr_##type##_f` changes into its own commit.

ompi/attribute/attribute.c
```diff
@@ -343,16 +347,22 @@ do { \
...
-    OMPI_ATTRIBUTE_FORTRAN_MPI2
+    OMPI_ATTRIBUTE_FORTRAN_INT,
+    OMPI_ATTRIBUTE_FORTRAN_AINT
 } ompi_attribute_translate_t;
 
+#if SIZEOF_INT == OMPI_SIZEOF_FORTRAN_INTEGER
+#define OMPI_ATTRIBUTE_FORTRAN_MPI1 OMPI_ATTRIBUTE_FORTRAN_INT
+#else
+#define OMPI_ATTRIBUTE_FORTRAN_MPI1 OMPI_ATTRIBUTE_FORTRAN_AINT
+#endif
```
jsquyres: I don't understand this block.  It is saying that if `sizeof(int) != sizeof(INTEGER)`, then use the sizeof an Aint.

I understand wanting to rename MPI1/MPI2 to more semantic names, but why not rename `OMPI_ATTRIBUTE_FORTRAN_MPI1` --> `OMPI_ATTRIBUTE_FORTRAN_INTEGER`?  That clearly indicates that we're dealing with Fint's (not ints or Aints).

ompi/attribute/attribute.c
```diff
@@ -381,8 +391,19 @@ static int set_value(ompi_attribute_type_t type, void *object,
...
+#if SIZEOF_INT == OMPI_SIZEOF_FORTRAN_INTEGER
+static int translate_to_fortran_int(attribute_value_t *val);
+static inline MPI_Fint translate_to_fortran_mpi1(attribute_value_t *val) {
+    return translate_to_fortran_int(val);
+}
+#else
+static inline MPI_Fint translate_to_fortran_mpi1(attribute_value_t *val) {
+    return (MPI_Fint)translate_to_fortran_aint(val);
+}
+#endif
```
jsquyres: This `#if` block is confusing to me for the same reason as the block above.  This stuff is so complicated that I think we should try to be as clear as possible, even if it means repeating some code.  Hence, I'd kinda prefer not to be clever and re-use the same subroutines behind the scenes, but rather have just two functions: one for INTEGER and one for Aint.  If there's some repeated code between the two, so be it.

ompi/mpi/fortran/mpif-h/keyval_create_f.c
jsquyres: After reading the other commits, I *think* you removed it because you add it back in in the call to `ompi_attr_create_keyval_fint()` if `sizeof(int) == sizeof(INTEGER)`.  But:

1. These commits are out of order.
2. This particular flag refers to how attribute callbacks are invoked, not how the attributes are stored, right?

jsquyres:
@ggouaillardet With all your changes, do all the tests in `ompi-tests/simple/attr` still pass, for both the i4 and i8 cases?

ggouaillardet:
@jsquyres I will review all this, update the PR and likely squeeze some commits.
the main task yet to be done is update the comments.
iirc, all the tests passed with {little endian, big endian} x { i4, i8}

ggouaillardet:
@jsquyres, bottom line, what we only care is the size of the data that was written in Fortran.
it is not mpi1 vs mpi2, nor ```Fint``` vs ```MPI_Aint```
on top of that, C must write an ```int``` (__not__ a ```MPI_Fint``` !) when creating a window.

one option is to have 12 cases
4 writers (```C```, ```int```, ```Fint``` and ```MPI_Aint```) and 3 readers (```C```, ```Fint```, ```MPI_Aint```)

for now, ```sizeof(Fint) == sizeof(int)``` or ```sizeof(Fint) = sizeof(MPI_Aint)```, so an other option is to use 9 cases (```C```, ```32 bits```, ```64 bits```)^2
and use macros in MPI-1 to use 32 or 64 bits based on ```sizeof(Fint)```

the first option requires more code but might be easier to read and maintain in the long run.

any thoughts ?

jsquyres:
I had to write these out to make sense of it.  According to the original work a few years ago, there's 9 cases:

1. C writes the attribute value, C reads the attribute value (unity)
1. C writes, Fortran MPI-1 reads
1. C writes, Fortran MPI-2 reads
1. Fortran MPI-1 writes, C reads
1. Fortran MPI-1 writes, Fortran MPI-1 reads (unity)
1. Fortran MPI-1 writes, Fortran MPI-2 reads
1. Fortran MPI-2 writes, C reads
1. Fortran MPI-2 writes, Fortran MPI-1 reads
1. Fortran MPI-2 writes, Fortran MPI-2 reads (unity)

I agree that "MPI-1" and "MPI-2'" are not accurate enough labels, so let's be a little more precise:

1. C `void*` is written into the attribute value, and the value is read into a C `void*` (unity)
1. C `void*` is written, Fortran `INTEGER` is read
1. C `void*` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read
1. Fortran `INTEGER` is written, C `void*` is read
1. Fortran `INTEGER` is written, Fortran `INTEGER` is read (unity)
1. Fortran `INTEGER` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, C `void*` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, Fortran `INTEGER` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read (unity)

I guess I'm confused as to why these 9 cases are not sufficient.  Regardless of the size of each of these three types, can't we just do the right conversion at each one?

dalcinl:
@jsquyres Some predefined MPI attibutes are C ``int`` values. When queried from C, you have to return an address to the value, however when queried from Fortran you have to return the value. You can store ``int`` in ``void*`` memory through an union type, but you have to know if what is stored is actually an ``int`` or ``void*`` value. So I think your 12-entry table is missing some cases.

ggouaillardet:
@jsquyres how do you set the ```disp_unit``` attribute (int) of a MPI_Win when sizeof(Fint) > sizeof(int) ?
think about retrieving it in C on a big endian arch ... and you will see the issue.
my conclusion is the 3 types we need are
- reference (aka void* in C)
- 32 bits value (aka C int, or, if sizeof(Fint) == 4 Fint aka MPI1
- 64 bits value (aka Aint, aka MPI2, or, if sizeof(Fint) == 8, Fint aka MPI1

@dalcinl you meant the __9__ entries table is missing some cases, right ? (12 is ok, we do write int, but we never read int, hence no 16 entries table)

dalcinl:
@ggouaillardet Yes, I think you got it right, it is 4 writers x 3 readers = 12 cases.

jsquyres:
@dalcinl @ggouaillardet Ok, I see -- 4 writers, with the 4th being the whacko intrinsic schizophrenic attributes.  Got it.

1. C `void*` is written into the attribute value, and the value is read into a C `void*` (unity)
1. C `void*` is written, Fortran `INTEGER` is read
1. C `void*` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read
1. Fortran `INTEGER` is written, C `void*` is read
1. Fortran `INTEGER` is written, Fortran `INTEGER` is read (unity)
1. Fortran `INTEGER` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, C `void*` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, Fortran `INTEGER` is read
1. Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read (unity)
1. Intrinsic is written, C `void*` is read
1. Intrinsic is written, Fortran `INTEGER` is read
1. Intrinsic is written, Fortran `INTEGER(KIND=MPI_ADDRESS_KIND)` is read

@ggouaillardet as per your question -- remember that the MPI RMA attributes (MPI-3.1 p414) are only *readable* by the user.  They are not *settable* by the user.  Hence, only we have to set the value (in cases 10, 11, 12) and make sure that they are *readable* as an `int*`, `INTEGER`, and `INTEGER(KIND=MPI_ADDRESS_KIND)`.

I also don't think we want to treat them as absolute sizes (i.e., 32 bit and 64 bit values), because that precludes environments where the value is smaller or bigger (e.g., 16 bits for an embedded environment -- don't laugh, this is an actual possibility!).  Just treat them as the types as they were discovered by configure (`void*`, `MPI_Fint`, `MPI_Aint`), and make sure all the conversions work for all 12 cases (for both endians).

BTW: let me say that I'd like to crack the skulls of the MPI RMA WG for making attributes of multiple different types -- which is different than the entire rest of the MPI standard!  :frowning:

....actually, reading MPI-3.1 p414 39-46, it's not actually clear how we're supposed to read these values from Fortran:

> In Fortran, calls to MPI_WIN_GET_ATTR(win, MPI_WIN_BASE, base, flag, ierror), MPI_WIN_GET_ATTR(win, MPI_WIN_SIZE, size, flag, ierror),
MPI_WIN_GET_ATTR(win, MPI_WIN_DISP_UNIT, disp_unit, flag, ierror), MPI_WIN_GET_ATTR(win, MPI_WIN_CREATE_FLAVOR, create_kind, flag, ierror), and MPI_WIN_GET_ATTR(win, MPI_WIN_MODEL, memory_model, flag, ierror) will return in base, size, disp_unit, create_kind, and memory_model the (integer representation of) the base address, the size, the displacement unit of the window win, the kind of routine used to create the window, and the memory model, respectively.

It says "integer representation of" -- does that mean `INTEGER`, or `INTEGER(KIND=MPI_ADDRESS_KIND)`, or ...?

E.g., `MPI_WIN_SIZE` is of type `MPI_Aint` in C, but `MPI_WIN_DISP_UNIT` (and others) is of type `int` in C.

Arrrghhh...

dalcinl:
@jsquyres I think we just have to use common sense

* ``MPI_WIN_BASE``: Use ``INTEGER(KIND=MPI_ADDRESS_KIND)``, I think this is obvious, the returned value should be appropriate for using the Cray pointer extension, right?
* ``MPI_WIN_SIZE``: Use ``INTEGER(KIND=MPI_ADDRESS_KIND)``, as that is what is used when the windows is created.
* ``MPI_WIN_DISP_UNIT``: Use ``INTEGER``, as that is what is used when the windows is created.
* ``MPI_WIN_CREATE_FLAVOR``/``MPI_WIN_MODEL``: Use ``INTEGER``, there is no need to use address-sized integers, and this lines up nicely with the C interface, were the attribute is an 'int' (and the model and flavor are small integer ``#define``d constants that fit in ``int`` and that could alternative be implemented as enumerations).

In general, in case of doubt or ambiguity, use the same Fortran datatype of companion create or set call that specify the attribute value.

jsquyres:
@dalcinl You're probably right, but I posted to the MPI RMA working group asking for clarification, anyway: https://github.com/mpiwg-rma/rma-issues/issues/1

jsquyres:
@RolfRabenseifner pointed out this morning that Fortran datatypes *must* be of type `INTEGER(KIND=MPI_ADDRESS_KIND)` -- that's the only type that the Fortran binding for `MPI_WIN_GET_ATTR` allows.

ggouaillardet:
@jsquyres I pushed my changes based on your review and discussion.
since most things have been changed, could you please review this again ?
I ran a few tests on little and big endian arch, and with 4 and 8 bytes fortran integer, and did not find any errors.

ibm-ompi:
Test passed.

rhc54:
@jsquyres @ggouaillardet Given that this is a continuing test failure on 1.10.3, shall we either move this forward or simply not worry about it for the 1.10 series?

ggouaillardet:
@rhc54 i am on the fence ...

on one hand, this issue was reported via the mailing list, so ideally, that should be fixed in v1.10.3
on the other hand, this is quite a big change that should deserve a good review and proper testing. on top of that, the scope of the issue is limited to big endian arch and/or 8 bytes Fortran integer.

if @jsquyres can review and merge this PR, i already made open-mpi/ompi-relase#1204

ggouaillardet:
i just found the ```onesided/cxx_win_attr``` error detected by MTT on big endian was caused by a bad test. i fixed it in open-mpi/ompi-tests@85502de9d042a31b63adfb3dafaab8b6d30bf7e8

now testing Fortran ...

jsquyres:
bot:retest

ompi/attribute/attribute.c
```diff
@@ -50,6 +52,8 @@
  *   means writing a pointer to an instance of something; changing the
  *   value of that instance will make it visible to anyone who reads
  *   that attribute value).
+ * - C also internally store some int attributes of a MPI_Win by value,
```
jsquyres: "... store**s** ..."

Should probably also list MPI-3.1 Section 11.2.6 (pp414-415) to the list of sections to reference (earlier in this comment).

ompi/attribute/attribute.c
```diff
@@ -50,6 +52,8 @@
...
+ *   and these attributes appear to be read-only (e.g. set once for all)
```
jsquyres: They **are** read-only (they don't just **appear** to be read-only 😄).  And it would be "i.e.," not "e.g."  😈

ompi/attribute/attribute.c
```diff
@@ -109,6 +113,38 @@
  *            CALL MPI_COMM_GET_ATTR(..., ret, ierr)
  *            --> ret will equal &bar
  *
+ * Cases where C writes an int attribute:
```
jsquyres: These 3 cases are not just where C writes an int -- the RMA attributes in question can be `(void*)`, `(MPI_Aint)`, or `(int)`.  Should probably make that clear in the title, the description below, and the text of these 3 cases.

jsquyres:
I see you listed the 12 cases in the commit message -- good.  Can you also cite this PR and https://github.com/mpiwg-rma/rma-issues/issues/1 in the commit message?  It would be good to provide as much cross-reference material as possible for Future Gilles and Future Jeff, when they're trying to remember/figure out what happened back here in 2016.

ompi/attribute/attribute.c
```diff
@@ -313,9 +353,11 @@ do { \
         /* MPI-2 Fortran-style */ \
         else { \
             MPI_Aint in, out;                        \
-            in = translate_to_fortran_mpi2(in_attr); \
-            (*((keyval_obj->copy_attr_fn).attr_mpi2_fortran_copy_fn)) \
-                (&(((ompi_##type##_t *)old_object)->attr_##type##_f), \
+            MPI_Fint attr_##type##_f;                \
```
jsquyres: Shouldn't this be an MPI_Aint?

jsquyres: Oh, this is the keyval -- not the value.  Got it.  Disregard my question, then...

ompi/attribute/attribute.c
```diff
@@ -687,13 +742,44 @@ int ompi_attr_set_c(ompi_attribute_type_t type, void *object,
 
 
 /*
+ * Front-end function internally called by the C API functions to set an
+ * int attribute.
```
jsquyres: By "int" here (and below), do you really mean "intrinsic"?

jsquyres: I guess I'm confused by this "int" label (in this comment / function name, and all its corresponding declarations / uses) because the MPI_Win intrinsic attributes are not just of type `int`.

jsquyres:
It seems https://github.com/mpiwg-rma/rma-issues/issues/1 has come to the conclusion that the Fortran bindings should just cast all of the RMA attribute values to `INTEGER(KIND=MPI_ADDRESS_KIND)` (i.e., `MPI_Aint`).

ompi/attribute/attribute.c
ggouaillardet: DISP_UNIT is an int * in C.
we used to treat it internally as an Fint, which is incorrect with -i8
since it is read only, there is no need for a pointer, and storing the value makes things much easier.

ggouaillardet:
@jsquyres i updated the PR

jjmaldonis:
Hello everyone, I am a user and am using OpenMPI 1.10.2 compiled with Intel.  I am trying to get the `spawn` functionality to work inside a for loop, but continue to get the error `too many retries sending message to <addr>, giving up` somewhere down the line in the for loop, seemingly because the processors are not being fully freed when disconnecting/finishing.  I found the [`orte/test/mpi/loop_spawn.c`](https://github.com/open-mpi/ompi/blob/master/orte/test/mpi/loop_spawn.c) example/test, and it has the exact same problem. I also found [this](https://www.open-mpi.org/community/lists/devel/2016/04/18814.php) mailing list post from ~ a month and a half ago.

Is this PR about the same issue I am having (ie the `loop_spawn` example not working)? If so, do you know if we can downgrade to e.g. 1.10.1 or another version? Or is there another solution to fix this bug?

Below is the output of the `loop_spawn` test on UW Madison's ACI cluster, which I know very little about in terms of architecture but can get information if it's helpful.

Thanks for your time and the excellent library.

```
mpiexec -np 5 loop_spawn
parent*******************************
parent: Launching MPI*
parent*******************************
parent: Launching MPI*
parent*******************************
parent: Launching MPI*
parent*******************************
parent: Launching MPI*
parent*******************************
parent: Launching MPI*
parent: MPI_Comm_spawn #0 return : 0
parent: MPI_Comm_spawn #0 return : 0
parent: MPI_Comm_spawn #0 return : 0
parent: MPI_Comm_spawn #0 return : 0
parent: MPI_Comm_spawn #0 return : 0
Child: launch
Child merged rank = 5, size = 6
parent: MPI_Comm_spawn #0 rank 4, size 6
parent: MPI_Comm_spawn #0 rank 0, size 6
parent: MPI_Comm_spawn #0 rank 2, size 6
parent: MPI_Comm_spawn #0 rank 3, size 6
parent: MPI_Comm_spawn #0 rank 1, size 6
Child 329941: exiting
parent: MPI_Comm_spawn #1 return : 0
parent: MPI_Comm_spawn #1 return : 0
parent: MPI_Comm_spawn #1 return : 0
parent: MPI_Comm_spawn #1 return : 0
parent: MPI_Comm_spawn #1 return : 0
Child: launch
parent: MPI_Comm_spawn #1 rank 0, size 6
parent: MPI_Comm_spawn #1 rank 2, size 6
parent: MPI_Comm_spawn #1 rank 1, size 6
parent: MPI_Comm_spawn #1 rank 3, size 6
Child merged rank = 5, size = 6
parent: MPI_Comm_spawn #1 rank 4, size 6
Child 329945: exiting
parent: MPI_Comm_spawn #2 return : 0
parent: MPI_Comm_spawn #2 return : 0
parent: MPI_Comm_spawn #2 return : 0
parent: MPI_Comm_spawn #2 return : 0
parent: MPI_Comm_spawn #2 return : 0
Child: launch
parent: MPI_Comm_spawn #2 rank 3, size 6
parent: MPI_Comm_spawn #2 rank 0, size 6
parent: MPI_Comm_spawn #2 rank 2, size 6
Child merged rank = 5, size = 6
parent: MPI_Comm_spawn #2 rank 1, size 6
parent: MPI_Comm_spawn #2 rank 4, size 6
Child 329949: exiting
parent: MPI_Comm_spawn #3 return : 0
parent: MPI_Comm_spawn #3 return : 0
parent: MPI_Comm_spawn #3 return : 0
parent: MPI_Comm_spawn #3 return : 0
parent: MPI_Comm_spawn #3 return : 0
Child: launch
[node:port?] too many retries sending message to <addr>, giving up
Child job 5 terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
mpiexec detected that one or more processes exited with non-zero status, thus causing

  Process name: [[...],0]
  Exit code:    255
```

ggouaillardet:
:bot:retest

jsquyres:
@jladd-mlnx @Di0gen @artpol84 http://bgate.mellanox.com/jenkins/job/gh-ompi-master-pr/2740/ Is the same periodic local fail we've seen at Mellanox Jenkins over time.

bot:mellanox:retest

jsquyres:
bot:retest

jjhursey:
Should this PR be closed? It looks like this has already made its way into `master`:
 * https://github.com/open-mpi/ompi/commit/72cfbb665c3555f2db21fc31511389732c3e187f

ggouaillardet:
Closed by open-mpi/ompit@72cfbb665c3555f2db21fc31511389732c3e187f
Thanks @jjhursey for pointing this

PR #1482, opened 2016-03-21
When using btls always consider the self component

hjelmn:
This PR also contains the necessary infrastructure to support this new feature. This PR reflects the reality that btl/self is always needed when using btls for communication.

### DIFF
--- a/opal/mca/base/help-mca-base.txt
+++ b/opal/mca/base/help-mca-base.txt
@@ -59,3 +59,9 @@ all components *except* a and b", while "c,d" specifies the inclusive
 behavior and means "use *only* components c and d."
 
 You cannot mix inclusive and exclusive behavior.
+[framework-param:exclude-always-component]
+A request was made to exclude a component from consideration that
+must always be considered.
+
+Framework: %s
+Component: %s
--- a/opal/mca/base/mca_base_component_find.c
+++ b/opal/mca/base/mca_base_component_find.c
@@ -115,9 +115,21 @@ int mca_base_component_find (const char *directory, mca_base_framework_t *framew
     /* Find all the components that were statically linked in */
     if (static_components) {
         for (int i = 0 ; NULL != static_components[i]; ++i) {
-            if ( (static_components[i]->mca_component_flags & MCA_BASE_COMPONENT_FLAG_ALWAYS_CONSIDER) ||
-                 use_component(include_mode, (const char**)requested_component_names,
-                               static_components[i]->mca_component_name) ) {
+            bool can_use;
+
+            can_use = use_component(include_mode, (const char **) requested_component_names,
+                                    static_components[i]->mca_component_name);
+
+            if (!can_use && static_components[i]->mca_component_flags & MCA_BASE_COMPONENT_FLAG_ALWAYS_CONSIDER) {
+                if (!include_mode) {
+                    opal_show_help ("help-mca-base.txt", "framework-param:exclude-always-component", true,
+                                    framework->framework_name, static_components[i]->mca_component_name);
+                }
+
+                can_use = true;
+            }
+
+            if (can_use) {
                 cli = OBJ_NEW(mca_base_component_list_item_t);
                 if (NULL == cli) {
                     ret = OPAL_ERR_OUT_OF_RESOURCE;
@@ -189,9 +201,17 @@ int mca_base_components_filter (mca_base_framework_t *framework, uint32_t filter
         mca_base_open_only_dummy_component_t *dummy =
             (mca_base_open_only_dummy_component_t *) cli->cli_component;
 
-        can_use = (cli->cli_component->mca_component_flags & MCA_BASE_COMPONENT_FLAG_ALWAYS_CONSIDER) ||
-            use_component (include_mode, (const char **) requested_component_names,
-                           cli->cli_component->mca_component_name);
+        can_use = use_component (include_mode, (const char **) requested_component_names,
+                                 cli->cli_component->mca_component_name);
+
+        if (!can_use && cli->cli_component->mca_component_flags & MCA_BASE_COMPONENT_FLAG_ALWAYS_CONSIDER) {
+            if (!include_mode) {
+                opal_show_help ("help-mca-base.txt", "framework-param:exclude-always-component", true,
+                                framework->framework_name, cli->cli_component->mca_component_name);
+            }
+
+            can_use = true;
+        }
 
         if (!can_use || (filter_flags & dummy->data.param_field) != filter_flags) {
             if (can_use && (filter_flags & MCA_BASE_METADATA_PARAM_CHECKPOINT) &&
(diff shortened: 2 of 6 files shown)


hjelmn:
@jsquyres, @bosilca Comments?

opal/mca/mca.h
```diff
@@ -257,6 +257,15 @@ typedef int (*mca_base_register_component_params_2_0_0_fn_t)(void);
 #define MCA_BASE_MAX_COMPONENT_NAME_LEN 63
 
 /**
+ * Component flags (mca_component_flags field)
+ */
+enum {
+    /** Always consider this component for selection. For this flag to
+     * work properly the component must always be built statically. */
+    MCA_BASE_COMPONENT_FLAG_ALLWAYS_CONSIDER = 1,
```
jsquyres: `ALWAYS` has one `L`, not 2.  :speak_no_evil:

opal/mca/base/mca_base_component_find.c
```diff
@@ -189,8 +189,9 @@ int mca_base_components_filter (mca_base_framework_t *framework, uint32_t filter
         mca_base_open_only_dummy_component_t *dummy =
             (mca_base_open_only_dummy_component_t *) cli->cli_component;
 
-        can_use = use_component (include_mode, (const char **) requested_component_names,
-                                 cli->cli_component->mca_component_name);
+        can_use = (cli->cli_component->mca_component_flags & MCA_BASE_COMPONENT_FLAG_ALLWAYS_CONSIDER) ||
```
jsquyres: Do you need to check here that this component is static?

opal/mca/mca.h
hjelmn: hah, misspelled it once and emacs was happy to complete it. fixing.

opal/mca/base/mca_base_component_find.c
hjelmn: not sure if we want to disable the flag if a component is dynamic. the static requirement allows us to check the flag without dlopening the component.

jjhursey:
Just to clarify, with this commit the user need only:
 ``-mca btl tcp``
to get the current ``tcp,self``. Which would be the same if they say:
 ``-mca btl tcp,self``

What happens if the user does:
 ``-mca btl ^self``

From my reading of the PR it look like this will be silently ignored by the MCA system. Correct?
I think we should throw an error in that last case. Something like: ``You tried to exclude a component that is flagged as always to be considered``

hjelmn:
@jjhursey That is correct and I agree we should probably print an error is they try to exclude a component that can not be excluded. Will update the PR shortly.

hjelmn:
Ok, the show help is in there as a warning:

```
➜  examples git:(always_self) ✗ mpirun -n 2 -mca btl ^self ./ring_c
A request was made to exclude a component from consideration that
must always be considered.

Framework: btl
Component: self
Process 0 sending 10 to 1, tag 201 (2 processes in ring)
Process 0 sent to 1
Process 0 decremented value: 9
Process 0 decremented value: 8
Process 0 decremented value: 7
Process 0 decremented value: 6
Process 0 decremented value: 5
Process 0 decremented value: 4
Process 0 decremented value: 3
Process 0 decremented value: 2
Process 0 decremented value: 1
Process 0 decremented value: 0
Process 0 exiting
Process 1 exiting
[mordor:85215] 3 more processes have sent help message help-mca-base.txt / framework-param:exclude-always-component
[mordor:85215] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

bosilca:
As no other BTL takes care of local communications, how it is possible that the initialization stage was successfully completed ?

hjelmn:
Initialization completed because the self btl is forced. That is the idea behind the PR. Allow components to be forced on.

opal/mca/base/help-mca-base.txt
```diff
@@ -59,3 +59,9 @@ all components *except* a and b", while "c,d" specifies the inclusive
 behavior and means "use *only* components c and d."
 
 You cannot mix inclusive and exclusive behavior.
+[framework-param:exclude-always-component]
+A request was made to exclude a component from consideration that
+must always be considered.
+
```
jjhursey: Let's make this super clear for the user (per George's comment) that this component will be considered anyway.

Maybe change to:
```text
Warning: A request was made to exclude a component from consideration that
must always be considered. This component (noted below) will -not- be excluded
from consideration. The program will continue running with this component being
made available for selection.

Framework: %s
Component: %s
```

What do you think?

jsquyres: My $0.02: if someone tries to disable it and it can't be disabled, we should abort (in line with the philosophy of "a human asked for X, and OMPI can't provide X, so OMPI needs to abort and let a human figure it out").

bosilca:
I understand that having to add self might became an annoyance. However, it is unexpected, that despite the user explicit request not to select self, we forced it's inclusion.

We are limiting the flexibility of our MCA components. With this patch, even if someone implements a different self component, she will not be able to test/run it on an installed Open MPI. If instead of forcing self indiscriminately, we strictly abide by the meaning of the inclusion/exclusion parameters, we do not prevent such developments.

Would it be enough to add "self" to the mca_btl_base_include variable to obtain a similar effect ?

hjelmn:
@bosilca I agree that it might limit the usefulness of the MCA system for developers. The idea is to make it so users do not have to worry about including a component that should always be there. I can add yet another MCA parameter to disable this behavior so developers who know what they are doing can disable forced loading.

jjhursey:
@hjelmn Having a parameter to override the forced loading seems a bit odd to me in terms of semantics. It also causes a problem for my _eventual_ usecase - where I want to force a component to be there always with no means for the user to exclude it. However, I could propose to add that back end when the time comes.

hjelmn:
@jjhursey Good point. Also, right now there is a way to override self if a replacement is being tested. You can simply set btl_self_exclusivity to 0.

bosilca:
I just noticed this is breaking the backward compatibility with the other version by decreasing the length of the component name. Not sure the expected benefit outweights the drawback.

So let me reask my question: Would it be enough to add "self" to the mca_btl_base_include variable (during the MCA parameter registration) to obtain a similar effect (improving our user friendliness), but without breaking backward compatibility nor limiting the scope of the MCA ?

hjelmn:
The component name has not been shortened. I took 4 bytes out of the reserved to handle the flags so this is a backwards compatible change as reserved is always 0. As for adding self to the mca_btl_base_include parameter, there is currently no such parameter. The btl parameter is the only selection parameter and if we set it before registration then it can get overridden by the environment, a file, or the command line. That is what we have now. We can add it after registration which will have more or less the same effect as this PR but moves the responsibility down to the base. I have no objections to handling this functionality in that way but I think this PR provides a cleaner solution. If a component is always required by not also require it to be static.

hjelmn:
Also, if we use the selection parameter it will need to be added in both the framework register an open functions because of MPI_T.

jsquyres:
@hjelmn Does this summarize the current status of this PR:

* @bosilca is worried about the loss of flexibility in the MCA system
* @jjhursey would like to see this because he has a different use case for the same functionality (a licensing/DRM component)
* @jsquyres asked for the system to *error* (not *warn*) when a user asks to exclude a component that cannot be excluded (I think @jjhursey will need this, too)

ibm-ompi:
Test passed.

PR #1483, opened 2016-03-21
RFC: Add support for flag enumerators for MCA variables

hjelmn:
This PR adds support for flag enumerators for MCA variables. A new function and types have been added to ```mca_base_var_enum.h``` to support these enumerators. Use ```mca_base_var_enum_create_flag``` to create a flag enumerator with a given set of flags and use the enumerator with ```mca_base_var_register()/mca_base_pvar_register()```.

Variables using a flag enumerator can take a comma-delimited list of flags. For example, instead of

```
# set put, get, send, and in-place
btl_vader_flags = 0xf
```

a user can can now use

```
# set put, get, send, and in-place
btl_vader_flags = send,put,get,in-place
```

The old value way of setting the flags will continue to work.

### DIFF
--- a/ompi/info/info.c
+++ b/ompi/info/info.c
@@ -275,7 +275,7 @@ int ompi_info_set (ompi_info_t *info, const char *key, const char *value)
 int ompi_info_set_value_enum (ompi_info_t *info, const char *key, int value,
                               mca_base_var_enum_t *var_enum)
 {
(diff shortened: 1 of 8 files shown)


hjelmn:
I should also mention I changed the semantics of the enum string_from_value function. It now must return an string that can be freed with free().

PR #1484, opened 2016-03-22
ompi/comm: clean up includes in comm_request.h

hjelmn:


hjelmn:
@jsquyres This should break the include chain.

Issue #1406, opened 2016-02-25
Fix TCP BTL THREAD_MULTIPLE deadlock


jsquyres:
Discussed at Feb 2016 Dallas meeting.  @bosilca says he knows what the problem is and will fix it.

jsquyres:
@hjelmn @bosilca Did this get fixed yet, perchance?

jsquyres:
@bosilca Says that he has large modifications to the TCP BTL (it's a giant change).  Not clear yet on how we want to handle this...

hppritcha:
@bosilca is there a chance we can have a narrowly focused fix to the existing TCP BTL?

jsquyres:
@bosilca If we can't get a focused fix, how about this:

* Have the old TCP BTL code selected on non-THREAD_MULTIPLE
* Have your new TCP BTL code selected on THREAD_MULTIPLE

This would allow us to have somewhat newer/less-tested code in the THREAD_MULTIPLE path (and therefore make solid advances in the THREAD_MULTIPLE category, modulo bug fixes), but have solid/reliable/time-tested code in the non-THREAD_MULTIPLE case (i.e., the majority of the world).

As the new code gets more testing and we get higher confidence in it, we can remove this division / remove the old code, and just have the new code.

What do you think?

jsquyres:
@bosilca Any comment?

bosilca:
#1491

jsquyres:
Fixed via #1491.

PR #1491, opened 2016-03-23
BTL TCP async progress

bosilca:
This patch fixed known issues with multi-threading support in the TCP BTL and add support for async progress to the same BTL (even when the multi-threading is not enabled). By default the async progress is compiled in (assuming thread support has been discovered during configure) but is disabled. Use the MCA parameter btl_tcp_progress_thread to activate this extra thread and gain async progress.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -108,6 +108,7 @@ sylvain.jeaugey@bull.net        Sylvain Jeaugey             Bull
 terry.dontje@oracle.com         Terry Dontje                Sun, Oracle
 thkorde@sandia.gov              Todd Kordenbrock            SNL
 tmattox@gmail.com               Tim Mattox                  IU, Cisco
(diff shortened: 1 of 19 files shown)


bosilca:
To answer the @jsquyres questions from #1406. The answer is: this patch doesn't modify the default behavior of the TCP BTL, except that it make it thread safe. However, in addition to this, the code add a second execution path that provides support for async progress for the TCP BTL. However, this support has to be enabled via an MCA parameter in order to be activated.

gpaulsen:
Do we need btl_tcp_progress_thread asyc progress for correctness or just for progress / throughput?

bosilca:
This is only needed for async progress. This construct is protected behind am MCA parameter, so in order to get async progress one must explicitly ask for it.

PR #1502, opened 2016-03-28
fix libfabric user intent

jsquyres:
@rhc54 This should fix the problem you found.

Also: minor tweaks to the OPAL summary output.


lanl-ompi:
Test FAILed.

jsquyres:
@hppritcha That looks like some kind of local failure on the LANL-Cray XC jenkins -- can you have a look?

hppritcha:
more /tmp problems on nersc login nodes.  I think its fixed now.
bot:retest

rhc54:
:+1: 

Thanks!

PR #1504, opened 2016-03-28
pml/ob1: detect unreachable errors

### LINKED ISSUE #1501: error message when no btl is available and cutoff is used
Here is the previous behavior
(running two tasks on two nodes with only ```sm``` and ```self``` btl, and ```cutoff=1024```)

obviously, this cannot work, and the error message is explicit

```
$ mpirun --hostn1,n2 -np 2 --mca mpi_add_procs_cutoff 1024 --mca btl  sm,self ./hw
At least one pair of MPI processes are unable to reach each other for
MPI communications.  This means that no Open MPI device has indicated
that it can be used to communicate between these processes.  This is
an error; Open MPI requires that all MPI processes be able to reach
each other.  This error can sometimes be the result of forgetting to
specify the "self" BTL.

  Process 1 ([[39443,1],0]) is on host: n1
  Process 2 ([[39443,1],1]) is on host: n2
  BTLs attempted: self

Your MPI job is now going to abort; sorry.
```

since default ```cutoff``` is now zero, there is no such check at ```MPI_Init()``` time, but such check is missing at runtime.
a cryptic error message is issued followed by a crash

```
$ mpirun --host n1,n2 -np 2 --mca mpi_add_procs_cutoff 0 --mca btl  sm,self ./hw
[n1:03052] mca_bml_base_btl_array_get_next: invalid array size
[n2:01380] mca_bml_base_btl_array_get_next: invalid array size
There are 2 tasks
mpirun noticed that process rank 0 with PID 3052 on node n1 exited on signal 11 (Segmentation fault).
```

@rhc54 @hjelmn FYI

hjelmn:
This commit adds code to detect when procs are unreachable when using
the dynamic add_procs functionality.

Fixes #1501

### DIFF
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -410,6 +410,8 @@ static int mca_bml_r2_add_proc (struct ompi_proc_t *proc)
     }
 
     if (!btl_in_use) {
+        proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_BML] = NULL;
+        OBJ_RELEASE(bml_endpoint);
         /* no btl is available for this proc */
         if (mca_bml_r2.show_unreach_errors) {
             opal_show_help ("help-mca-bml-r2.txt", "unreachable proc", true,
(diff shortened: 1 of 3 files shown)


hjelmn:
@ggouaillardet The code checking for a NULL endpoint was after the first dereference of the endpoint in the send path. This commit fixes that path.

Issue #1487, opened 2016-03-23
Build fail on CentOS 5.11: Absoft (ASM symbols missing)


jsquyres:
MTT link from last night showing the problem: https://mtt.open-mpi.org/index.php?do_redir=2279

```
../../../opal/.libs/libopen-pal.so: undefined reference to `__sync_sub_and_fetch_4'
./../../opal/.libs/libopen-pal.so: undefined reference to `__sync_bool_compare_and_swap_4'
libopen-pal.so: undefined reference to `__sync_add_and_fetch_4'
: undefined reference to `__sync_add_and_fetch_8'
```

@hjelmn Can you have a look?

hjelmn:
Thats weird. I don't have an absoft compiler. They only make a fortran compiler, correct?

hjelmn:
gcc 4.1.2 is the c compiler. It should have the sync builtin atomics (configure found them).

jsquyres:
Correct: Absoft makes a Fortran compiler.  But to test Open MPI with their Fortran compiler, they have to test with whatever C compiler is on the system, too.  In this case, it's not their Fortran compiler that's causing the problem -- it's something else, because the error occurs when trying to link `opal_wrapper`.

jsquyres:
@cagoelz sent us some output from that failed MTT run:

* [absoft-config.log.txt](https://github.com/open-mpi/ompi/files/186186/absoft-config.log.txt)
* [absoft-make-V=1.log.txt](https://github.com/open-mpi/ompi/files/186188/absoft-make-V.1.log.txt)

bosilca:
From the [gcc](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html) documentation of 

> Not all operations are supported by all target processors. If a particular operation cannot be implemented on the target processor, a warning will be generated and a call an external function will be generated. The external function will carry the same name as the builtin, with an additional suffix `_n' where n is the size of the data type.

Based on this documentation, the error message above, aka. missing __sync_add_and_fetch_4, suggests that the compiler does not have intrinsic support for these atomics operations and relies on external functions. Our configure reached a different conclusion because of a bug.

Here is what I gather from the log file:

1. Looking for __sync:
[...]

PR #1508, opened 2016-03-29
configure: only enable sync builtin atomics if they link

hjelmn:
This commit fixes the check for sync builtin atomics.
AC_COMPILE_IFELSE is insufficient to check for the builtins. Need to
use AC_LINK_IFELSE.

Fixes open-mpi/ompi#1487


jsquyres:
@hjelmn Thanks.  Let's get this in as soon as CI finishes so that it's included in tonight's MTT.  Thanks!

hjelmn:
Sure. Not sure why it was an AC_COMPILE_IFELSE in the first place.

PR #1515, opened 2016-04-02
opal_info_support: output component versions

kmroz:
Hopefully a better approach than https://github.com/open-mpi/ompi/pull/1514. Aims at fixing https://github.com/open-mpi/ompi/issues/1396.

Tests:

    % ompi_info --param btl all
                 MCA btl: vader (MCA v2.1.0, API v3.0.0, Component v3.0.0)
                 MCA btl: sm (MCA v2.1.0, API v3.0.0, Component v3.0.0)
                 MCA btl: self (MCA v2.1.0, API v3.0.0, Component v3.0.0)
                 MCA btl: tcp (MCA v2.1.0, API v3.0.0, Component v3.0.0)
                 MCA btl: parameter "btl_tcp_if_include" (current value: "",
                          data source: default, level: 1 user/basic, type:
                          string)
                          Comma-delimited list of devices and/or CIDR
                          notation of networks to use for MPI communication
                          (e.g., "eth0,192.168.0.0/16").  Mutually exclusive
                          with btl_tcp_if_exclude.
                 MCA btl: parameter "btl_tcp_if_exclude" (current value:
                          "127.0.0.1/8,sppp", data source: default, level: 1
                          user/basic, type: string)
                          Comma-delimited list of devices and/or CIDR
                          notation of networks to NOT use for MPI
                          communication -- all devices not matching these
                          specifications will be used (e.g.,
                          "eth0,192.168.0.0/16").  If set to a non-default
                          value, it is mutually exclusive with
                          btl_tcp_if_include.
                 MCA btl: parameter "btl_tcp_progress_thread" (current value:
                          "0", data source: default, level: 1 user/basic,
                          type: int)

    % ompi_info --param btl tcp
                 MCA btl: tcp (MCA v2.1.0, API v3.0.0, Component v3.0.0)
                 MCA btl: parameter "btl_tcp_if_include" (current value: "",
                          data source: default, level: 1 user/basic, type:
                          string)
                          Comma-delimited list of devices and/or CIDR
                          notation of networks to use for MPI communication
                          (e.g., "eth0,192.168.0.0/16").  Mutually exclusive
                          with btl_tcp_if_exclude.
                 MCA btl: parameter "btl_tcp_if_exclude" (current value:
                          "127.0.0.1/8,sppp", data source: default, level: 1
                          user/basic, type: string)
                          Comma-delimited list of devices and/or CIDR
                          notation of networks to NOT use for MPI
                          communication -- all devices not matching these
                          specifications will be used (e.g.,
[...]

PR #1516, opened 2016-04-02
opal_info_support: fix api comments

kmroz:
Nothing exciting, just something that caught my eye during https://github.com/open-mpi/ompi/pull/1515.


hjelmn:
We should probably create a roll-up PR for of your ompi_info patches for ompi-release 2.x

kmroz:
@hjelmn - Agreed. One more that would be good to get reviewed beforehand would be: https://github.com/open-mpi/ompi/pull/1522

Then I can put together a PR with the cherry picked commits and open against ompi-release/v2.x?

PR #1519, opened 2016-04-04
Poc/oob usock

ggouaillardet:


ggouaillardet:
@rhc54 can you please have a look at this ?
the issue can be evidenced with ```loop_spawn``` from the ```ibm``` test suite.

```
mpirun -np 1 ./loop_spawn
```

with current master, a ```SIGPIPE``` occurs when ```mpirun``` tries to write to a peer that has already gone (and hence the socket has been closed)
/* not sure that should occur ... */
the first commit fixes that (plus a typo)

then we run into an other issue, when ```mpirun``` tries to write to a peer that has already gone,
it ends up sending the message to itself (!)
the second commit fixes that by simply discarding these messages instead of sending them.
the second commit might make the first one useless (except for the typo ...)

bottom line, there is a real issue, and i am not convinced i fixed that in a proper way

rhc54:
I think it looks okay - thanks!

rhc54:
BTW: this needs to go over to 2.x as well, when you get a chance

PR #1523, opened 2016-04-04
type_hindexed: fix compiler warnings

kmroz:
Specifically:
    adio/common/utils.c: In function ‘ADIOI_Type_create_hindexed_x’:
    adio/common/utils.c:97:3: warning: passing argument 3 of
    ‘PMPI_Type_hindexed’ discards ‘const’ qualifier from pointer target
    type  [-Wdiscarded-qualifiers]

Signed-off-by: Karol Mroz <mroz.karol@gmail.com>


hjelmn:
As nice as this is I don't think we can take it. The MPI_Type_hindexed call was deprecated and did not get the const keyword. A better fix would be to patch romio to not use the deprecated routine. @jsquyres am I wrong?

bosilca:
s/deprecated/removed/ since MPI 3.0 (source MPI 3.1 page 18 line 17).

kmroz:
Ahhh, I see. Then definitely replacing the calls is the way to go :smile:

PR #1531, opened 2016-04-09
bml: always enable the bml

hjelmn:
This commit ensures the bml is always enabled whether or not it will
be used. This ensures that any available btls communicate their modex
so that they can be used for one-sided communication.


hjelmn:
Didn't catch this earlier. We need the btls open to ensure their modex messages are available if osc/rdma ends up being used. Target 2.0.0 if possible. Caught when using ucx for pt2pt and osc/rdma for one-sided.

PR #1532, opened 2016-04-10
romio,java: cleanup deprecated hindexed call

kmroz:
Following up on https://github.com/open-mpi/ompi/pull/1523.

I've left headers, makefiles and sources related to MPI_Type_hindexed() in place. Since I saw various other APIs marked as deprecated, I have no idea what you guys typically do in such cases?

### DIFF
--- a/ompi/mca/io/romio314/romio/adio/common/ad_coll_build_req_new.c
+++ b/ompi/mca/io/romio314/romio/adio/common/ad_coll_build_req_new.c
@@ -844,7 +844,7 @@ int ADIOI_Build_agg_reqs(ADIO_File fd, int rw_type, int nprocs,
     {
 	if (client_comm_sz_arr[i] > 0)
 	{
-	    MPI_Type_hindexed(client_ol_ct_arr[i], client_blk_arr[i],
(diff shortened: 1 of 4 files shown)


hppritcha:
@nrgraham23 could you check this?

hppritcha:
Java change looks good :+1:

kmroz:
@hppritcha - Thank you for taking a look.

jsquyres:
@hjelmn You were the last one to do the ROMIO update (right?).  We don't care about this kind of update to the ROMIO code base, right?  The changes look small / trivial, remove some compiler warnings, and won't affect a new ROMIO update.  Is that correct?

If so, we might as well take this stuff.

hjelmn:
Yeah, and if they haven't fixed this upstream MPICH should probably make this change as well.

jsquyres:
Good point.  @kmroz Might want to check the latest ROMIO and see if they've fixed this upstream.  If they haven't might want to give them a heads up.

kmroz:
@jsquyres - Thanks. I will look into upstream ROMIO and see if they have something similar.

Issue #1275, opened 2016-01-03
ARMCI-MPI hangs when datatypes are used


jeffhammond:
See https://travis-ci.org/jeffhammond/armci-mpi/builds/99887902 for full details.  ARMCI-MPI passes all the tests with MPICH 3.2.  When I use Open-MPI 1.10, tests pass when I disable user-defined datatypes (e.g. `MPI_Type_subarray`), but otherwise fail.

ggouaillardet:
@jeffhammond 1.10.2rc3 fixes bugs (missing resize) for subarray and darray.
could you give it a try ?

jeffhammond:
@ggouaillardet It does not appear to help: https://travis-ci.org/jeffhammond/armci-mpi/jobs/99913181.  While it has not failed yet, it's almost certainly hung, since successful builds always finish in under 10 minutes.

ARMCI tests (`armci-perf` and `armci-test`) were written by PNNL and take liberties with ISO C and even the ARMCI interface (which is easy to do when there is no proper specification), so it is possible that Open-MPI is not at fault, but MPICH is pretty strict about RMA and it has no problem here.  I guess I will start debugging locally if none of the Open-MPI devs knows a reason why it would hang.

ggouaillardet:
I confirm the hang even with master.
The test is made of two tasks, and they are doing at the same time into the other task
MPI_Accumulate(MPI_REPLACE)
MPI_Get
MPI_Win_flush_all

I always get a hang on one node with --mca btl  tcp,self
It occurs at various places, but when datatype is a subarray of dimension 5 or more.


@hjelmn can you please confirm this should work, even without a fence between accumulate and get ?
Note if i replace MPI_Accumulate with MPI_Put, i got the same hang.

I initially suspected ordering is not correctly enforced, but i cannot make sense of the wrong results i am seeing... so that could indicate something trickier

ggouaillardet:
my bad, there is no MPI_Win_flush_all at all

jeffhammond:
@ggouaillardet Yes, there is.  `ARMCI_AllFence` maps directly to `MPI_Win_flush_all`.

Can you try running with each of the environment variables:
* `ARMCI_NO_FLUSH_LOCAL=1`?  This turns `MPI_Win_flush_local` calls inside of all the ARMCI blocking operations into `MPI_Win_flush`.
[...]

PR #1534, opened 2016-04-12
osc/pt2pt: Fix tag conflicts on parallel RMA communications

kawashima-fj:
This is a reimplementation of 06ecdb6.

As noted in [Gilles' comment](https://github.com/open-mpi/ompi/issues/1275#issuecomment-168931765), his commit 06ecdb6 only fixes the race condition between `MPI_ACCUMULATE` and `MPI_GET_ACCUMULATE`. Similar race conditions between `MPI_PUT` and `MPI_GET`, and others, exist.

My commit 5ac95df has detail description of the cause and the fix.

@hjelmn, @ggouaillardet Could you have take a look?

### DIFF
--- a/ompi/mca/osc/pt2pt/osc_pt2pt.h
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt.h
@@ -636,21 +636,49 @@ static inline void osc_pt2pt_add_pending (ompi_osc_pt2pt_pending_t *pending)
 /**
  * get_tag:
  *
- * @short Get a send/recv tag for large memory operations.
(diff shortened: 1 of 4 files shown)


kawashima-fj:
[My gist](https://gist.github.com/kawashima-fj/815c7a1fd5e6d5e2952a28bf33375b00) can reproduce the bug.

Running this with 2 processes will show `NG put-get (wbuf)` or `NG put-get (rbuf)` without this PR.
It will also show `NG acc-gacc` if Gilles' original commit is reverted.

PR #1537, opened 2016-04-12
datatype: Fix incorrect predefined datatype names and other datatype bugs

kawashima-fj:
See each commit messages for detail.

@bosilca Please review


kawashima-fj:
Rationale.
MPI-3.1 p.284 says:
Named predefined datatypes have the default names of the datatype name. For example, MPI_WCHAR has the default name of MPI_WCHAR .

A program to confirm names of predefined datatypes is on [my gist](https://gist.github.com/kawashima-fj/9da95122681105e7c4aeb4cfc25a54fd). This program outputs names of predefined datatypes that doesn't have correct names defined in MPI standard. This program outputs `MPI_LONG_LONG != MPI_LONG_LONG_INT` even after my fix but this case is not problem because MPI-3.1 p.26 says `MPI_LONG_LONG` is a synonym of `MPI_LONG_LONG_INT`.

jsquyres:
I added the test from the gist into the ompi-tests/ibm/datatype test suite.

bosilca:
:+1:

PR #1539, opened 2016-04-12
config: check for more __sync builtins

hjelmn:
This commit updates the check for __sync builtin atomics to see if the
compiler supports both __sync_bool_compare_and_swap and
__sync_add_and_fetch. If either of these functions are not available
then we can't use the __sync builtins.


hjelmn:
@bosilca Turns out some compiler versions have __sync_synchronize() but not the others. Go figure.

jsquyres:
:+1:

bosilca:
As CAS is the most basic atomic operation if we have it we could implement our own version of all the other atomics. But I am not sure this is worth the effort, especially when we already have atomic implementations for most of the architectures. :+1:

PR #1548, opened 2016-04-15
Roll to PMIx 1.1.4rc1 and remove the PMIx 1.2.0 directory...

rhc54:
…as the community has decided to not do that release version. This incorporates a number of bug fixes that have been identified and repaired in the PMIx and OMPI code bases. Also includes several minor corrections to the PMIx code so it now supports run-thru without hanging on collectives involving a process that exits


jsquyres:
@hppritcha What do you think of putting this in v2.0.1?  That way, we're not holding up v2.0.0 for stability of this new PMIx stuff, but since it won't affect backwards compatibility (and probably solves some bugs), we could put it in v2.0.1.  Let's discuss Monday.

hppritcha:
we can put it in 2.0.1

----------

Howard

PR #1550, opened 2016-04-15
Per request from Jeff, aggregate all help messages during MPI_Init th…

rhc54:
…ru MPI_Finalize as long as the RTE is available

### DIFF
--- a/ompi/errhandler/errhandler_predefined.c
+++ b/ompi/errhandler/errhandler_predefined.c
@@ -148,7 +148,7 @@ void ompi_mpi_errors_return_win_handler(struct ompi_win_t **win,
 
 static void out(char *str, char *arg)
 {
-    if (ompi_mpi_initialized && !ompi_mpi_finalized) {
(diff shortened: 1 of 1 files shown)


ompi/errhandler/errhandler_predefined.c
```diff
@@ -148,7 +148,7 @@ void ompi_mpi_errors_return_win_handler(struct ompi_win_t **win,
 
 static void out(char *str, char *arg)
 {
-    if (ompi_mpi_initialized && !ompi_mpi_finalized) {
+    if (ompi_rte_initialized && !ompi_mpi_finalized) {
```
jsquyres: @rhc54 Will `ompi_rte_initialized` automatically be false once we get far enough in MPI_FINALIZE?  I.e., do we still need to check `!ompi_mpi_finalized`?

rhc54: We set it to false as soon as the rte finalizes, so we should be fine.

rhc54:
going to add this to this weekend's tarball - can always adjust later if needed

Issue #1551, opened 2016-04-15
No standardization on hostname max length


kmroz:
I was tracing a program in examples/ this evening and noticed that we have a myriad of different max hostname lengths ranging from 20 - 1024, and slight variations in handling gethostname():

https://gist.github.com/kmroz/6fb7699712d41abb4cd2ba9553163c82

I thought it might be a good idea to use, if available, what's in limits.h:HOST_NAME_MAX. Something like:

    diff --git a/opal/include/opal_config_bottom.h b/opal/include/opal_config_bottom.hindex f906278..c2f1643 100644
    --- a/opal/include/opal_config_bottom.h
    +++ b/opal/include/opal_config_bottom.h
    @@ -491,7 +491,14 @@ static inline uint16_t ntohs(uint16_t netvar) { return netvar; }      
    #include <hostLib.h>

    #ifndef MAXHOSTNAMELEN
    -#define MAXHOSTNAMELEN 64
    +/* From limits.h. Since we're using gethostname(), we add 1 to the max to allow
    + * for null termination.
    + */
    +#ifdef HOST_NAME_MAX
    +#define MAXHOSTNAMELEN HOST_NAME_MAX + 1
    +#else
    +#define MAXHOSTNAMELEN 64 + 1
    +#endif
    #endif
    #endif

and sift through all callers of gethostname(). I have some concerns though:

1. I'm probably going to be stomping on components that may have an 'upstream' somewhere?
2. Non-POSIX systems might have some issue with this?

Perhaps this is barking up the wrong tree?

rhc54:
we've been shamefully inconsistent on it - no harm in fixing it. any upstream conflicts are easily resolved

PR #1552, opened 2016-04-16
ompi/opal/orte/oshmem/test: max hostname length cleanup

kmroz:
Trying to standardize from where max hostname length is derived and how gethostname() called. This is low priority/cleanup, but touches a fair number of files.

I've left MPI_Get_processor_name() to continue using MPI_MAX_PROCESSOR_NAME, as the variable to which it writes the hostname is passed into the function. If for some reason hostname max lengths ever exceed MPI_MAX_PROCESSOR_NAME, we could segfault.

Also, programs in orte/test/ are not configured to build, and further, they have various build failures. I've added the hostname max len fixups to these files as well, knowing they don't build. Perhaps a future TODO :)

Fixes: https://github.com/open-mpi/ompi/issues/1551

### DIFF
--- a/ompi/info/info.c
+++ b/ompi/info/info.c
@@ -95,7 +95,7 @@ opal_pointer_array_t ompi_info_f_to_c_table = {{0}};
  */
 int ompi_info_init(void)
 {
-    char val[MPI_MAX_INFO_VAL];
+    char val[OPAL_MAXHOSTNAMELEN];
     char *cptr;
 
     /* initialize table */
@@ -134,7 +134,7 @@ int ompi_info_init(void)
     }
 
     /* local host name */
-    gethostname(val, MPI_MAX_INFO_VAL);
+    gethostname(val, sizeof(val));
     ompi_info_set(&ompi_mpi_info_env.info, "host", val);
 
     /* architecture name */
--- a/ompi/runtime/ompi_mpi_abort.c
+++ b/ompi/runtime/ompi_mpi_abort.c
@@ -119,7 +119,7 @@ int
 ompi_mpi_abort(struct ompi_communicator_t* comm,
                int errcode)
 {
-    char *msg, *host, hostname[MAXHOSTNAMELEN];
+    char *msg, *host, hostname[OPAL_MAXHOSTNAMELEN];
     pid_t pid = 0;
 
     /* Protection for recursive invocation */
--- a/opal/mca/event/libevent2022/libevent/evdns.c
+++ b/opal/mca/event/libevent2022/libevent/evdns.c
@@ -48,6 +48,8 @@
  * Version: 0.1b
  */
 
+#include "opal_config.h"
+
 #include <sys/types.h>
 #include "event2/event-config.h"
 
@@ -120,10 +122,6 @@
 #define EVDNS_LOG_DEBUG 0
 #define EVDNS_LOG_WARN 1
 #define EVDNS_LOG_MSG 2
-
-#ifndef HOST_NAME_MAX
-#define HOST_NAME_MAX 255
-#endif
 
 #include <stdio.h>
 
@@ -3108,7 +3106,7 @@ evdns_search_ndots_set(const int ndots) {
 
 static void
 search_set_from_hostname(struct evdns_base *base) {
-	char hostname[HOST_NAME_MAX + 1], *domainname;
+	char hostname[OPAL_MAXHOSTNAMELEN], *domainname;
 
 	ASSERT_LOCKED(base);
 	search_postfix_clear(base);
--- a/opal/mca/pmix/pmix114/pmix/examples/dynamic.c
+++ b/opal/mca/pmix/pmix114/pmix/examples/dynamic.c
@@ -45,12 +45,12 @@ int main(int argc, char **argv)
     uint32_t nprocs;
     char nsp2[PMIX_MAX_NSLEN+1];
     pmix_app_t *app;
-    char hostname[1024], dir[1024];
+    char hostname[PMIX_MAXHOSTNAMELEN], dir[1024];
     pmix_proc_t *peers;
     size_t npeers, ntmp=0;
     char *nodelist;
 
-    gethostname(hostname, 1024);
+    gethostname(hostname, sizeof(hostname));
     getcwd(dir, 1024);
 
     /* init us */
--- a/opal/util/output.c
+++ b/opal/util/output.c
@@ -127,7 +127,7 @@ OBJ_CLASS_INSTANCE(opal_output_stream_t, opal_object_t, construct, NULL);
 bool opal_output_init(void)
 {
     int i;
-    char hostname[32];
+    char hostname[OPAL_MAXHOSTNAMELEN];
     char *str;
 
     if (initialized) {
@@ -177,7 +177,6 @@ bool opal_output_init(void)
         verbose.lds_want_stderr = true;
     }
     gethostname(hostname, sizeof(hostname));
-    hostname[sizeof(hostname)-1] = '\0';
     asprintf(&verbose.lds_prefix, "[%s:%05d] ", hostname, getpid());
 
     for (i = 0; i < OPAL_OUTPUT_MAX_STREAMS; ++i) {
@@ -254,7 +253,7 @@ bool opal_output_switch(int output_id, bool enable)
 void opal_output_reopen_all(void)
 {
     char *str;
-    char hostname[32];
+    char hostname[OPAL_MAXHOSTNAMELEN];
 
     str = getenv("OPAL_OUTPUT_STDERR_FD");
     if (NULL != str) {
--- a/ompi/mca/pml/v/pml_v_output.c
+++ b/ompi/mca/pml/v/pml_v_output.c
@@ -21,7 +21,7 @@
 
 int pml_v_output_open(char *output, int verbosity) {
     opal_output_stream_t lds;
-    char hostname[32] = "NA";
+    char hostname[OPAL_MAXHOSTNAMELEN] = "NA";
 
     OBJ_CONSTRUCT(&lds, opal_output_stream_t);
     if(!output) {
@@ -40,7 +40,7 @@ int pml_v_output_open(char *output, int verbosity) {
             lds.lds_file_suffix = output;
         }
         lds.lds_is_debugging = true;
-        gethostname(hostname, 32);
+        gethostname(hostname, sizeof(hostname));
         asprintf(&lds.lds_prefix, "[%s:%05d] pml_v: ", hostname, getpid());
         lds.lds_verbose_level = verbosity;
         mca_pml_v.output = opal_output_open(&lds);
--- a/ompi/mpi/c/get_processor_name.c
+++ b/ompi/mpi/c/get_processor_name.c
@@ -66,7 +66,7 @@ int MPI_Get_processor_name(char *name, int *resultlen)
        Guard against gethostname() returning a *really long* hostname
        and not null-terminating the string.  The Fortran API version
        will pad to the right if necessary. */
-    gethostname(name, MPI_MAX_PROCESSOR_NAME - 1);
+    gethostname(name, (MPI_MAX_PROCESSOR_NAME - 1));
     name[MPI_MAX_PROCESSOR_NAME - 1] = '\0';
     *resultlen = (int) strlen(name);
 
--- a/ompi/runtime/ompi_mpi_finalize.c
+++ b/ompi/runtime/ompi_mpi_finalize.c
@@ -113,7 +113,7 @@ int ompi_mpi_finalize(void)
         /* Note that if we're not initialized or already finalized, we
            cannot raise an MPI exception.  The best that we can do is
            write something to stderr. */
-        char hostname[MAXHOSTNAMELEN];
+        char hostname[OPAL_MAXHOSTNAMELEN];
         pid_t pid = getpid();
         gethostname(hostname, sizeof(hostname));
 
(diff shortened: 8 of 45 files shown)


hppritcha:
@kmroz would you mind rebasing this down to a smaller set of commits?  it helps in tracking when we/if we want to merge in to other branches.  Squashing down to one isn't necessary, more like 2 or 3 would be fine.

kmroz:
@hppritcha - Will do.

ompi/runtime/ompi_mpi_abort.c
```diff
@@ -133,7 +133,7 @@ ompi_mpi_abort(struct ompi_communicator_t* comm,
     if (ompi_rte_initialized) {
         host = ompi_process_info.nodename;
     } else {
-        gethostname(hostname, sizeof(hostname));
+        gethostname(hostname, MAXHOSTNAMELEN);
```
jsquyres: Is there a reason to make changes like this?  I somewhat like the defensive coding style of using `sizeof(foo)` to indicate the length of an array.  In this case, it's obviously `MAXHOSTNAME`, but using `sizeof` protects you in case the array changes size someday (i.e., there's only 1 place to update instead of 2).

Indeed, I wouldn't mind seeing:

1. Changing the size of all hostname-like arrays to `MAXHOSTNAMELEN` (which you do already in this PR)
1. Changing all references to the length of the array (e.g., passing as the 2nd arg to `gethostname()`) use `sizeof(array_name)`.

opal/mca/event/libevent2022/libevent/evdns.c
```diff
@@ -3108,11 +3106,11 @@ evdns_search_ndots_set(const int ndots) {
 
 static void
 search_set_from_hostname(struct evdns_base *base) {
-	char hostname[HOST_NAME_MAX + 1], *domainname;
+	char hostname[MAXHOSTNAMELEN], *domainname;
```
jsquyres: FWIW, we import libevent from an external package.  We typically modify as little code in this libevent core code as possible (because then we'll just need to remember to make the same updates the next time we upgrade the internal copy of libevent).

Specifically: it's probably ok / harmless to make these changes as part of this PR, but we should probably expect them to drift back the next time we upgrade libevent.

opal/mca/pmix/pmix114/pmix/examples/dynamic.c
```diff
@@ -45,12 +45,12 @@ int main(int argc, char **argv)
     uint32_t nprocs;
     char nsp2[PMIX_MAX_NSLEN+1];
     pmix_app_t *app;
-    char hostname[1024], dir[1024];
+    char hostname[MAXHOSTNAMELEN], dir[1024];
```
jsquyres: Similar to libevent, the pmix component comes from outside Open MPI, too.  Patches for PMIx should likely go to their upstream, too.

opal/util/output.c
```diff
@@ -176,8 +176,7 @@ bool opal_output_init(void)
     } else {
         verbose.lds_want_stderr = true;
     }
-    gethostname(hostname, sizeof(hostname));
-    hostname[sizeof(hostname)-1] = '\0';
+    gethostname(hostname, MAXHOSTNAMELEN);
```
jsquyres: Is this kind of stuff correct?

I ask because, for example, the OS X man page for `gethostname(2)` says:

>      Host names are limited in length to {sysconf(_SC_HOST_NAME_MAX)} characters, not including the trailing null, currently 255.

Since we're using `MAXHOSTNAMELEN` and not `sysconf()`, we can't know that it is big enough to include the terminating character.

jsquyres:
@kmroz Pro tip: if you put "Fixes #1551" in a commit message, it'll auto-close that issue when this is merged.  I don't remember offhand, but you *might* get the same effect if you put that string in the original description of this PR...?

opal/util/output.c
kmroz: From the Linux `gethostname(2)` man page:

> SUSv2 guarantees that "Host names are limited to 255 bytes".  POSIX.1 guarantees that "Host names (not including the terminating null  byte) are  limited  to HOST_NAME_MAX bytes".  On Linux, HOST_NAME_MAX is defined with the value 64, which has been the limit since Linux 1.0 (earlier kernels imposed a limit of 8 bytes).

For POSIX systems, I think defining a MAXHOSTNAMELEN = limits.h:HOST_NAME_MAX + 1 and using it in this fashion is fine. We have space for the null terminator which gethostname() will append provided it doesn't overrun. I imagine your concern is for non-POSIX systems (which I also feared a bit).

Perhaps the alternative would be to do:

    +/* From limits.h. Since we're using gethostname(), we add 1 to the max to allow
    + * for null termination.
    + */
    +#ifdef HOST_NAME_MAX
    +#define MAXHOSTNAMELEN HOST_NAME_MAX + 1
    +#else
    +#define MAXHOSTNAMELEN 255 + 1
    +#endif
    #endif

Or just set it to 255 + 1 across the board and not worry about pulling it via limits.h? BTW - what does OS X limits.h:HOST_NAME_MAX say?

kmroz:
@jsquyres Ah, cool! I didn't know this. We can test the "PR" hypothesis with this PR... provided I can get it into the right shape :)

ompi/info/info.c
```diff
@@ -134,7 +134,7 @@ int ompi_info_init(void)
     }
 
     /* local host name */
-    gethostname(val, MPI_MAX_INFO_VAL);
+    gethostname(val, MAXHOSTNAMELEN);
     ompi_info_set(&ompi_mpi_info_env.info, "host", val);
```
bosilca: This is limited by MPI_MAX_INFO_KEY.

bosilca:
So the plan is to change [where possible] to use sizeof instead of hardcoding the MAXHOSTNAMELEN ?

jsquyres:
@bosilca I was specifically referring to using `char foo[X]` to size the arrays (e.g., X = `MAXHOSTNAMELEN`), but then use `sizeof(foo)` when passing the length of the array to something (e.g., `gethostname()`) -- just seems like good defensive programming.

bosilca:
That's exactly what I was referring to. I was wondering if the plan was to change all the instances where sizeof(array[]) can be used (and thus provide a consistent usage) or just the one where it has been specifically mentioned ?

kmroz:
I will rework this PR to pass sizeof() to all invocations of gethostname().

jsquyres:
This PR was just about `MAXHOSTNAMELEN` -- I don't think there's a larger effort to unify the whole code base on `sizeof(foo)`.  But I agree, that would be a Good Thing.  😄

ompi/info/info.c
kmroz: Thanks! I will look into changing this occurrence.

kmroz:
Yes of course, I meant using sizeof() in the context of this PR :smile:

bosilca:
Much appreciated @kmroz that will put some consistency in the code. @jsquyres I wasn't talking about the entire code base, but at everything touched by this patch.

kmroz:
Rebased on top of current master, and s/MAXHOSTNAMELEN/sizeof()/ when invoking gethostname().

Some additional things:

1. In `opal_config_bottom.h` - if `HOST_NAME_MAX` is not defined in `limits.h`, I've increased `MAXHOSTNAMELEN` to the maximum possible value according to SUSv2 (ie. 255 + 1).
2. @bosilca - `ompi/info/info.c:MPI_MAX_INFO_KEY = OPAL_MAX_INFO_VAL = 256`. So, we shouldn't exceed this with `gethostname()` and `MAXHOSTNAMELEN`. Agreed?
3. `pmix_config_bottom.h` - includes `limit.h`, so we should get a valid `HOST_NAME_MAX` on which to base `MAXHOSTNAMELEN`, otherwise I've bumped it to 255 + 1.
4. I've mostly left `get_processor_name.c:MPI_Get_processor_name()` alone and using `MPI_MAX_PROCESSOR_NAME` for now. It's invoked in a number of places with varying allocations for `*name`. I think this would also benefit some cleanup but from another PR as it would touch fortran code, etc... and I've no idea yet what I'm doing there :smile:
5. I've left `win_compat.h` alone. It defines `MAXHOSTNAMELEN` as `_MAX_PATH`. **Any thoughts?**

@jsquyres - If you feel it would be better to drop libevent/pmix modifications from this PR, I'm happy to rebase things without those changes. I could try to go straight upstream, but would need a pointer where upstream is.

Thanks again for all the reviews!

rhc54:
@kmroz Please do a "git diff opal/mca/pmix" and send me the results - I'll upstream them to the pmix master. You can leave them here as well for now and I'll update them with the next PMIx release.

kmroz:
@rhc54 - The diff you asked for here: https://gist.github.com/kmroz/e786d5ea9e5dfd90ccc342d97792b997 I'd appreciate your feedback on `pmix_config_bottom.h`. The `MAXHOSTNAMELEN` section appears to only be included as part of `#ifdef MCS_VXWORKS` - which means all the pmix code using MAXHOSTNMELEN is actually getting it from elsewhere :grimacing: I assumed from opal_config_bottom.h, but introducing compiler errors there didn't break pmix. Any ideas where it might be picking up the macro?! Thanks!

kmroz:
@rhc54 - I think I narrowed it down. In pmix code, `MAXHOSTNAMELEN` is taken from `/usr/include/asm-generic/param.h`. I've moved the `#ifndef MAXHOSTNAMELEN` higher in `pmix_config_bottom.h` to cover all environments.

Above gist updated with most recent pmix diffset.

rhc54:
Thanks - that indeed looks correct. I'll update this upstream, but let you include it in your commit here to avoid conflicts.

kmroz:
@rhc54 - Great, thanks!

jsquyres:
@kmroz Sorry for the delay.  Some (very minor) nits:

1. I wonder if we should rename `MAXHOSTNAMELEN` to be `OPAL_MAXHOSTNAMELEN`, given that we are adding one to it, and therefore it is *not* the same as the system `MAXHOSTNAMELEN` (if there is one).  This would reinforce the idea that this value is guaranteed to have enough space for the `\0` terminator.
1. In a small number of places, you multiply the length of the string by `sizeof(char)`.  That probably isn't necessary.

kmroz:
@jsquyres - No problem.

I've removed the sizeof(char) instances and updated the PR.
I will investigate s/MAXHOSTNAMELEN/OPAL_MAXHOSTNAMELEN/.

kmroz:
@jsquyres - Hmmm, did you have something like this in mind for both `opal_config_bottom.h` and `pmix_config_bottom.h`? Then, globally `s/hostname[MAXHOSTNAMELEN]/hostname[OPAL_MAXHOSTNAMELEN]/`.

    #ifndef MAXHOSTNAMELEN
    /* From limits.h. Since we're using gethostname(), we add 1 to the max to allow
     * for null termination.
     */
    #ifdef HOST_NAME_MAX
    #define OPAL_MAXHOSTNAMELEN HOST_NAME_MAX + 1
    #else
    /* SUSv2 guarantees that "Host names are limited to  255  bytes". */
    #define OPAL_MAXHOSTNAMELEN 255 + 1
    #else
    #define OPAL_MAXHOSTNAMELEN MAXHOSTNAMELEN + 1
    #endif
    #endif

rhc54:
Yes, that is what he is proposing. I would second it as it will avoid future confusion, if you wouldn't mind.

kmroz:
@jsquyres / @rhc54 - Hi gents!

OK - I think this should do it. Please have a look and let me know if you think I've missed something.

I've also modified `win_compat.h` to include a `#define OPAL_MAXHOSTNAMELEN MAXHOSTNAMELEN + 1`. Lastly, I've moved the `#define` block in `opal_config_bottom.h` out of the ` #ifdef MCS_VXWORKS`.

@rhc54 - I've updated the pmix diff (https://gist.github.com/kmroz/e786d5ea9e5dfd90ccc342d97792b997) with the latest.

jsquyres:
@kmroz Perhaps a little simpler, without the negative logic (pro tip: put a "c" after the triple-single-tick markdown for verbatim, and it syntax highlights for C):

```c
#if defined(MAXHOSTNAMELEN)
#define OPAL_MAXHOSTNAMELEN (MAXHOSTNAMELEN + 1)
#elif defined(HOST_NAME_MAX)
#define OPAL_MAXHOSTNAMELEN (HOST_NAME_MAX + 1)
#else
/* SUSv2 guarantees that "Host names are limited to 255 bytes". */
#define OPAL_MAXHOSTNAMELEN (255 + 1)
#endif
```

Thanks for your patience!  Otherwise, I think it looks great.

kmroz:
@jsquyres - On it... will fix it up shortly.

kmroz:
@jsquyres - PR updated per your suggestion. Way cleaner, btw :smile: And thanks for the syntax highlighting tip!

@rhc54 - Above gist updated with latest pmix changes.

rhc54:
Ummm...not quite right. The changes in the pmix code base should be PMIX_MAXHOSTNAMELEN, not OPAL_...

Sorry to be a pain.

kmroz:
@rhc54 - No pain at all! PR (specfically commit https://github.com/open-mpi/ompi/pull/1552/commits/e1c64e6e59e9e49cab6872bcfe8beae474a73e66) and gist [pmix.diff](https://gist.github.com/kmroz/e786d5ea9e5dfd90ccc342d97792b997) updated.

rhc54:
:+1:

kmroz:
@jsquyres - Is there anything else you see that's needed to improve upon this one?

jsquyres:
Works for me.  👍

jsquyres:
This is probably work PR'ing to the v2.x branch -- probably for v2.0.1.  It's not critical functionality, by any means, but it will help reduce drift between master and the v2.x branch.

kmroz:
@jsquyres - Will do, thanks.

PR #1555, opened 2016-04-18
coll-portals4: check return value from reduce kary tree functions

shawone:
Check error return values (functions reduce_kary_tree_top and reduce_kary_tree_bottom)


jsquyres:
@regrant @tkordenbrock Could you review?

tkordenbrock:
This looks good. :+1: 
Merging.

PR #1557, opened 2016-04-18
memory/patcher: munmap hook could be called from within a malloc() implementation

hjelmn:


jsquyres:
Assumedly this needs to also be added to open-mpi/ompi-release#1079 when CI passes.

Issue #1559, opened 2016-04-19
PSM2 MTL doesn't work for single node jobs


hppritcha:
PSM2 MTL has problems with initialization, in particular, in the call to ```psm2_ep_open``` when runninig on a single node.

The failure signature is
```
srun -N 1 -n 1 a.out          
opal13.0Assertion failure at psm_ep.c:832: ep->epid != 0
 
miniFE.x:52223 terminated with signal 6 at PC=2aaaabd835f7 SP=7fffffffb7d8. Backtrace:
/lib64/libc.so.6(gsignal+0x37)[0x2aaaabd835f7]
/lib64/libc.so.6(abort+0x148)[0x2aaaabd84ce8]
/lib64/libpsm2.so.2(+0x11dea)[0x2aaab97c7dea]
/lib64/libpsm2.so.2(+0x10782)[0x2aaab97c6782]
/lib64/libpsm2.so.2(psm2_ep_open+0x3c5)[0x2aaab97c5165]
/opt/openmpi/1.10/intel/lib/openmpi/mca_mtl_psm2.so(ompi_mtl_psm2_module_init+0x196)[0x2aaab95b25e6]
/opt/openmpi/1.10/intel/lib/openmpi/mca_mtl_psm2.so(+0x2a00)[0x2aaab95b2a00]
/opt/openmpi/1.10/intel/lib/libmpi.so.12(ompi_mtl_base_select+0x9d)[0x2aaaaad6148d]
/opt/openmpi/1.10/intel/lib/openmpi/mca_pml_cm.so(+0x3d9b)[0x2aaab8916d9b]
/opt/openmpi/1.10/intel/lib/libmpi.so.12(mca_pml_base_select+0x433)[0x2aaaaad68643]
/opt/openmpi/1.10/intel/lib/libmpi.so.12(ompi_mpi_init+0x6ac)[0x2aaaaad1a06c]
/opt/openmpi/1.10/intel/lib/libmpi.so.12(MPI_Init+0xe4)[0x2aaaaad3b134]
miniFE.x[0x432da2]
miniFE.x[0x404e1a]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x2aaaabd6fb15]
miniFE.x[0x404c69]
```
This is with v1.10.1 but also shows up in v2.x.

The issue appeared on devel mail list:
https://www.open-mpi.org/community/lists/devel/2016/04/18762.php

hppritcha:
[...]

Issue #1159, opened 2015-11-30
Update shared library version numbers for v2.0.0


jsquyres:
This needs to be done before v2.0.0 is released.

jsquyres:
Closed via open-mpi/ompi-release@3fe21ec8ddf0e8bf4

PR #1578, opened 2016-04-25
mtl/psm2: do not set PSM2_DEVICES env variable

hppritcha:
The PSM2 MTL was setting the PSM2_DEVICES environment
variable to self,shm for single node jobs.  This is
causing single node/single process jobs to fail on
some omnipath systems.  Not setting this environment
variable fixes these issues.

This fix is needed as part of bringup of omnipath
clusters at several customer sites.

Fixes issue #1559 

@matcabral 
@rhc54  (copying you just for fyi)

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -198,10 +198,6 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
 	return NULL;
     }
 
-    if (num_local_procs == num_total_procs) {
-      setenv("PSM2_DEVICES", "self,shm", 0);
-    }
-
     err = psm2_init(&verno_major, &verno_minor);
     if (err) {
       opal_show_help("help-mtl-psm2.txt",


hjelmn:
@hppritcha The psm2 shared memory performance is already pretty bad. Does this make is worse?

hppritcha:
Doubtful.  But intel says if you ask for HFI for single mode jobs probably can't run as many ranks as cores - probably for BWL or the likes.  Iterating with intel on this.  Meanwhile TOSS took patch for now.

matcabral:
Hi @hppritcha,  
When initializing the PSM2_DEVICES=hfi the maximum number ranks on a single node will be defined by the number of HW contexts in the HFI card. The cards have 160 contexts. By default the hfi1.ko driver will load 1 HW context per physical core on the system, but this can be edited using module load time parameter num_user_contexts. Therefore, there is no limitation on having one rank per core as described by the above numbers. Additionally it can support up to 8 way sharing of the hardware contexts to get to ~1280 ranks per node.
In regards to the performance penalty, shm device will continue to be used transparently (in single node jobs). There will be little overheard only at init time to init hfi.

hppritcha:
@matcabral thanks for the detailed answer.  this helps a lot.  Given this though can you explain why its not good to get rid of this PSM2_DEVICES env setting in the psm2 mtl?

matcabral:
@hppritcha,
  I have pending to test the proposed change in our lab (I’m waiting for resources) to evaluate potential impacts. By analytical review, I see the change could fit as an immediate fix for the issues found (root cause research ongoing). However, I am not fully satisfied with the PSM2 MTL having PSM to lock  HW resources (HFI card contexts) that are not used. In the short term, there may not be significant impact *when running under-subscribed* . However, as cores count grow (KNx series) this will become more impactfull.

hppritcha:
closing this.  looks like a better fix is available.

hppritcha:
reopening in case may be useful for @matcabral

matcabral:
@hppritcha , having fixes for both of the problems reported in open-mpi/ompi#1159 (root cause fix) there is no need for this PR. However, would till be convenient to check the environment before setting PSM2_DEVICES. I will send a new PR for that. 

thanks,

ibm-ompi:
Test passed.

hppritcha:
closing this PR,  fixed elsewhere.

PR #1582, opened 2016-04-26
orte/rml: ORTE_RML_SEND_COMPLETE handles messages with both NULL iov …

ggouaillardet:
…and cbfunc.buffer

### DIFF
--- a/orte/mca/rml/base/base.h
+++ b/orte/mca/rml/base/base.h
@@ -13,6 +13,8 @@
  * Copyright (c) 2007-2014 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2014-2016 Intel, Inc. All rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@rhc54 
a ```rml_send_t``` created in ```mca_oob_tcp_recv_handler``` has both NULL ```iov``` and ```cbfunc.buffer``` and that caused a crash when ```ORTE_RML_SEND_COMPLETE``` is invoked.

i am not sure whether this is the right fix, or it fixes a case that should never happen.
fwiw, i was only able to run into this when invoking ```mpirun``` on a ```x86_64``` host, and run the ```ibm/dynamic/spawn``` program on a ```sparcv9``` host (!)

rhc54:
It shouldn't happen, but that is a very unusual use-case, so maybe it somehow gets invoked in a strange way. Regardless, there is no harm in this protection.

ggouaillardet:
I think the root cause is ```mca_oob_{usock,tcp}_send_handler``` do not correctly handle zero size messages.
I think we should test
```if (NULL != msg->msg->iov)```
instead of
```if (NULL != msg->msg->data)```
I will test that from tomorrow

Issue #1593, opened 2016-04-27
--enable-sparse-groups causes assert failures in intel and ibm tests


jsquyres:
v2.x results https://mtt.open-mpi.org/index.php?do_redir=2306 shows that `--enable-sparse-groups` is returning in assertion failures in:

* Intel MPI_Comm_split3_c
* Intel MPI_Comm_split3_f 
* Intel MPI_Comm_split4_c 
* Intel MPI_Comm_split4_f 

And v2.x results https://mtt.open-mpi.org/index.php?do_redir=2309 shows an `--enable-sparse-groups` fail in

* IBM group/compare
* Similar failures occurred in IBM communicator/create_group_* (Fortran) in different MTT results

There's a bunch of similar failures in master results: https://mtt.open-mpi.org/index.php?do_redir=2310

@edgargabriel Can you have a look?

edgargabriel:
Honestly, I suspect that these failures have to do something with the addproc changes, since the sparse group code has not changes since ~2008

rhc54:
Is this on the master, or 2.x? I thought we had fixed such things on the master, and so this might just be something that hasn't come across.

jsquyres:
This is on v2.x.  But I see the same failure signature on master.

jsquyres:
Looks like there's more failures of the same type on master (i.e., more fails on master than there are on v2.x).  I added an MTT link for master fails to the initial description.

edgargabriel:
I can have a look, but it will take me a couple of days to get to it

jsquyres:
@edgargabriel Many thanks.  There's no immediate rush, but given all the progress we've made in terms of scalability for the 2.x series, having sparse groups continue to work is probably a Good Thing.  Doesn't need to be v2.0.0, though.

rhc54:
FWIW: I believe @ggouaillardet was looking at this and (IIRC) had a patch for it that is pending for master

jsquyres:
Also FWIW: I don't see an obvious PR pending about sparse groups.  Shrug.  @hjelmn thinks it may be a missing OBJ_RETAIN or OBJ_RELEASE.

hjelmn:
I couldn't find an existing PR so I opened #1595 to fix this. Tested and the failure is gone and I do not see any procs being leaked.

PR #1595, opened 2016-04-27
ompi/group: fix sparse group proc reference counting

hjelmn:
This commit fixes a bug when sparse groups are in use. Since sparse
group do not actually increment the reference counts of any procs
(they just retain the parent group) it is wrong to decrement the
reference counts of all procs in the group using
ompi_group_decrement_proc_count(). This commit makes the call to
ompi_group_decrement_proc_count() conditional on the group being
dense.

Fixes open-mpi/ompi#1593


hjelmn:
@ggouaillardet When I bring this over to 2.0.1 I will set you as the reviewer.

PR #1600, opened 2016-04-28
pmix/cray: set fence_nb to NULL

### LINKED ISSUE #1527: aprun launched jobs hang in MPI_Finalize
On master aprun launched jobs now hang in `orte_rml_base_close`.  Processes are hung in the
`ORTE_WAIT_FOR_COMPLETION` macro.

### LINKED ISSUE #1597: aprun launched jobs hang on v2.x
Looks like a recent commit broke aprun launched jobs.

hppritcha:
Rather than have a stub function for the pmix fence_nb
operation, just set to NULL.  Causes fewer problems.

Fixes #1597
Fixes #1527

@rhc54 

Signed-off-by: hppritcha <howardp@lanl.gov>

### DIFF
--- a/opal/mca/pmix/cray/pmix_cray.c
+++ b/opal/mca/pmix/cray/pmix_cray.c
@@ -58,8 +58,6 @@ static int cray_resolve_peers(const char *nodename,
 static int cray_resolve_nodes(opal_jobid_t jobid, char **nodelist);
 static int cray_put(opal_pmix_scope_t scope, opal_value_t *kv);
 static int cray_fence(opal_list_t *procs, int collect_data);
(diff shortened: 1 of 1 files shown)


rhc54:
ah, indeed - that would solve this problem! Thanks Howard - sorry I didn't spot that when I made the change. Any idea why it didn't show up in Jenkins?

I'm changing the milestone to 2.0 as this doesn't exist in the 1.10 series.

hppritcha:
jenkins doesn't actually run anything, just makes sure it builds for cray.  I'm going to change that now that NERSC switched to slurm - no need for job reservations for small jobs.

PR #1604, opened 2016-04-29
Improve the transport key print statement to ensure that we don't get…

rhc54:
… zero fields as this can be a problem for PSM

@hppritcha Could you please try this out?

### DIFF
--- a/orte/mca/ess/pmi/ess_pmi_module.c
+++ b/orte/mca/ess/pmi/ess_pmi_module.c
@@ -217,6 +217,9 @@ static int rte_init(void)
             ORTE_ERROR_LOG(ORTE_ERR_OUT_OF_RESOURCE);
             return ORTE_ERR_OUT_OF_RESOURCE;
         }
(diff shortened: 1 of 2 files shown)


orte/mca/ess/pmi/ess_pmi_module.c
```diff
@@ -161,17 +161,7 @@ static int rte_init(void)
         error = "getting univ size";
         goto error;
     }
-    orte_process_info.max_procs = u32;
-
```
jjhursey: It looks like this is reverting the PMIX_UNIV_SIZE commit from earlier, was that intended?

rhc54: crud - no, will fix. Not sure how that happened.

Thx!

hppritcha:
yep, after 2.0.0 dust settles.

hppritcha:
works for me, both 1 process and multi-process on a single node.
:+1:

Issue #1599, opened 2016-04-28
Graph500 failure with ompi-release v2.x


artpol84:
Hello, I'm facing the the error (the output log is below) with graph500 and ompi-release v2.x and `ob1/openib`.
With ompi-release v1.10 this doesn't happen. This also doesn't appear with v2.x and `yalla`

Way to reproduce:
1. `git clone` ompi-release and switch to the v2.x branch
2. build and install ompi-release
3. Use following sources of graph500 http://www.graph500.org/sites/default/files/files/graph500-2.1.4.tar.bz2
4. untar and go to ./mpi directory and run `make`.
5. Run using the command line below adjusted to the actual environment.

**The error output:**
```
mpirun -np 16 -bind-to core -mca btl_openib_warn_default_gid_prefix 0 -report-bindings \
     -display-map -mca btl_openib_if_include mlx4_0:1 -mca pml ob1 -mca btl self,vader,openib \
     -mca opal_pmix_base_async_modex 0 -mca mpi_add_procs_cutoff 100000 \
     -mca scoll_mpi_enable 0 -mca coll_hcoll_enable 0 -mca scoll_fca_enable 0 \
     -mca coll_fca_enable 0 --map-by node ./graph500_mpi_simple 16

 Data for JOB [52131,1] offset 0

 ========================   JOB MAP   ========================

 Data for node: cn31	Num slots: 8	Max slots: 0	Num procs: 4
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 0 Bound: socket 0[core 0[hwt 0]]:[B/././.][./././.]
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 4 Bound: socket 0[core 1[hwt 0]]:[./B/./.][./././.]
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 8 Bound: socket 0[core 2[hwt 0]]:[././B/.][./././.]
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 12 Bound: socket 0[core 3[hwt 0]]:[./././B][./././.]

 Data for node: cn32	Num slots: 8	Max slots: 0	Num procs: 4
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 1 Bound: socket 0[core 0[hwt 0]]:[B/././.][./././.]
 	Process OMPI jobid: [52131,1] App: 0 Process rank: 5 Bound: socket 0[core 1[hwt 0]]:[./B/./.][./././.]
[...]

PR #1605, opened 2016-04-29
osc/rdma: fix global index array calculation

hjelmn:
This commit fixes a bug that occurs when ranks are either not mapped
evenly or by something other than core.

Fixes open-mpi/ompi#1599


hjelmn:
@artpol84 This should fix osc/rdma. Please test.

osc/pt2pt actually appears to be working but very slowly. It looks like it has nothing to do with osc/pt2pt either. Works fine with btl/vader and will test with btl/ugni.

artpol84:
@hjelmn, ok will do today.

artpol84:
@hjelmn 
I've checked. Works fine. Please, PR to the release repo.

PR #1615, opened 2016-05-02
memory/patcher: add #if check for MREMAP_FIXED

hjelmn:
This commit fixes a compile error when the system has mremap but not
MREMAP_FIXED. In this case we do not care about the value of
new_address as the argument does not exist.


jsquyres:
Thanks to @PHHargrove for reporting

hjelmn:
@jladd-mlnx @Di0gen Unrelated Jenkins failure. Will go ahead and merge to get into tonights MTT.

PR #1616, opened 2016-05-02
osc/rdma: clear all_sync object when obtaining a lock

hjelmn:
This commit fixes a bad synchronization detection bug that occurs when
mixing MPI_Win_fence() and MPI_Win_lock(). If no communication has
occurred in the fence epoch it is safe to just clear the all_sync
object (it was set up by fence).


hjelmn:
@ggouaillardet This fixes the bug from https://www.open-mpi.org/community/lists/users/2016/05/29066.php

hjelmn:
@jladd-mlnx @Di0gen Another unrelated failure.

PR #1621, opened 2016-05-03
linux: actually enable libudev based on the result of AC_CHECK_LIB

bgoglin:
instead of doing AC_CHECK_HEADERS+AC_CHECK_LIB and only using the result of the former.

Thanks to Paul Hargrove for reporting the issue (OMPI build with -m32).

(cherry picked from open-mpi/hwloc@9549fd59af04dca2e2340e17f0e685f8c552d818)


jsquyres:
👍

Issue #120, opened 2014-10-01
Add TCP btl bandwidth detection


ompiteam:
Per http://www.open-mpi.org/community/lists/devel/2010/09/8461.php, at least on some flavors of Linux, we can get the available TCP bandwidth from /sys/class/net/ethX/speed.  It seems like this would be a useful way to accurately set the btl_bandwidth value, when possible.

kmroz:
In Linux, we can get this via an ioctl(..., SIOCETHTOOL, ...) + ethtool_cmd_speed(). However, this wouldn't work on the BSDs, etc.

@jsquyres - Does OMPI come with some #defines for specific OSs? We could do the ioctl() if Linux, otherwise default to the current 100 Mbit/sec that's being set?

jsquyres:
No, we tend to stay away from OS/version #defines and instead test for desired behavior in configure.  For example, in this case, we could just check to see if `SIOCETHTOOL` exists.  And if so, it's probably a safe enough assumption that it can be used for this `ioctl()`.

Want to make a PR to do that?

kmroz:
Ahh, I see... like you do with ```#ifdef HAVE_SYS_TYPES_H``` etc.

Yes, I will put something together for this.

PR #1628, opened 2016-05-04
btl/tcp: autodetect bandwidth and latency if unset by the user

kmroz:
A first pass at putting something together for open-mpi/ompi#120.

Small tests:
---------------
#### Check component default bandwidth value
```
% ompi_info --param btl tcp --level 5
[...]
             MCA btl tcp: parameter "btl_tcp_bandwidth" (current value:
                          "100", data source: default, level: 5 tuner/detail,
                          type: unsigned_int)
                          Approximate maximum bandwidth of interconnect (0 =
                          auto-detect value at run-time [not supported in all
                          BTL modules], >= 1 = bandwidth in Mbps)
[...]
```
Note that the default of 100 is used during component register. On btl creation (for each interface/link), this is refined.

#### Obtain bandwidth values for each BTL instance via ioctl(SIOCETHTOOL)
```
% mpirun -d --debug-daemons --mca btl self,tcp  --mca btl_tcp_if_include wlan0,tun0 -np 2 ./hello_c
[...]
[elm][[32725,1],0][btl_tcp_component.c:259:mca_btl_tcp_component_register] 
[elm][[32725,1],0][btl_tcp_component.c:1207:mca_btl_tcp_component_init] 
[elm][[32725,1],1][btl_tcp_component.c:259:mca_btl_tcp_component_register] 
[elm][[32725,1],1][btl_tcp_component.c:1207:mca_btl_tcp_component_init] 
[elm][[32725,1],0][btl_tcp_component.c:714:mca_btl_tcp_component_create_instances] 
[elm][[32725,1],0][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:wlan0
[elm][[32725,1],0][btl_tcp_component.c:580:mca_btl_tcp_create] wlan0: btl_bandwidth: 0
[elm][[32725,1],0][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:tun0
[elm][[32725,1],0][btl_tcp_component.c:580:mca_btl_tcp_create] tun0: btl_bandwidth: 10
[elm][[32725,1],1][btl_tcp_component.c:714:mca_btl_tcp_component_create_instances] 
[elm][[32725,1],1][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:wlan0
[elm][[32725,1],1][btl_tcp_component.c:580:mca_btl_tcp_create] wlan0: btl_bandwidth: 0
[elm][[32725,1],1][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:tun0
[elm][[32725,1],1][btl_tcp_component.c:580:mca_btl_tcp_create] tun0: btl_bandwidth: 10
[...]
```

#### Allow user to override ioctl() obtained bandwidth values
```
% mpirun -d --debug-daemons --mca btl self,tcp --mca btl_tcp_bandwidth_tun0 200 --mca btl_tcp_if_include wlan0,tun0 -np 2 ./hello_c
[...]
[elm][[32513,1],0][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:wlan0
[elm][[32513,1],0][btl_tcp_component.c:580:mca_btl_tcp_create] wlan0: btl_bandwidth: 0
[elm][[32513,1],0][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:tun0
[elm][[32513,1],0][btl_tcp_component.c:580:mca_btl_tcp_create] tun0: btl_bandwidth: 200
[...]
[elm][[32513,1],1][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:wlan0
[elm][[32513,1],1][btl_tcp_component.c:580:mca_btl_tcp_create] wlan0: btl_bandwidth: 0
[elm][[32513,1],1][btl_tcp_component.c:540:mca_btl_tcp_create] if_name:tun0
[elm][[32513,1],1][btl_tcp_component.c:580:mca_btl_tcp_create] tun0: btl_bandwidth: 200
[...]
```

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -85,6 +85,9 @@
 #include "opal/mca/common/cuda/common_cuda.h"
 #endif /* OPAL_CUDA_SUPPORT */
 
+#define MCA_BTL_TCP_BTL_BANDWIDTH 100
+#define MCA_BTL_TCP_BTL_LATENCY 100
+
 /*
  * Local functions
  */
@@ -320,8 +323,8 @@ static int mca_btl_tcp_component_register(void)
                                        MCA_BTL_FLAGS_HETEROGENEOUS_RDMA |
                                        MCA_BTL_FLAGS_SEND;
 
-    mca_btl_tcp_module.super.btl_bandwidth = 100;
-    mca_btl_tcp_module.super.btl_latency = 100;
+    mca_btl_tcp_module.super.btl_bandwidth = MCA_BTL_TCP_BTL_BANDWIDTH;
+    mca_btl_tcp_module.super.btl_latency = MCA_BTL_TCP_BTL_LATENCY;
 
     mca_btl_base_param_register(&mca_btl_tcp_component.super.btl_version,
                                 &mca_btl_tcp_module.super);
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -56,6 +56,7 @@
 #include <limits.h>
 
 #include "opal/mca/event/event.h"
+#include "opal/util/ethtool.h"
 #include "opal/util/if.h"
 #include "opal/util/output.h"
 #include "opal/util/argv.h"
@@ -323,8 +324,11 @@ static int mca_btl_tcp_component_register(void)
                                        MCA_BTL_FLAGS_HETEROGENEOUS_RDMA |
                                        MCA_BTL_FLAGS_SEND;
 
-    mca_btl_tcp_module.super.btl_bandwidth = MCA_BTL_TCP_BTL_BANDWIDTH;
-    mca_btl_tcp_module.super.btl_latency = MCA_BTL_TCP_BTL_LATENCY;
+    /* Bandwidth and latency initially set to 0. May be overridden during
+     * mca_btl_tcp_create().
+     */
+    mca_btl_tcp_module.super.btl_bandwidth = 0;
+    mca_btl_tcp_module.super.btl_latency = 0;
 
     mca_btl_base_param_register(&mca_btl_tcp_component.super.btl_version,
                                 &mca_btl_tcp_module.super);
@@ -516,6 +520,27 @@ static int mca_btl_tcp_create(int if_kindex, const char* if_name)
         /* allow user to override/specify latency ranking */
         sprintf(param, "latency_%s:%d", if_name, i);
         mca_btl_tcp_param_register_uint(param, NULL, btl->super.btl_latency, OPAL_INFO_LVL_5, &btl->super.btl_latency);
+
+        /* Only attempt to auto-detect bandwidth and/or latency if it is 0.
+         *
+         * If detection fails to return anything other than 0, set a default
+         * bandwidth and latency.
+         */
+        if (0 == btl->super.btl_bandwidth) {
+            unsigned int speed = opal_ethtool_get_speed(if_name);
+            btl->super.btl_bandwidth = (speed == 0) ? MCA_BTL_TCP_BTL_BANDWIDTH : speed;
+            if (i > 0) {
+                btl->super.btl_bandwidth >>= 1;
+            }
+        }
+        /* We have no runtime btl latency detection mechanism. Just set a default. */
+        if (0 == btl->super.btl_latency) {
+            btl->super.btl_latency = MCA_BTL_TCP_BTL_LATENCY;
+            if (i > 0) {
+                btl->super.btl_latency <<= 1;
+            }
+        }
+
 #if 0 && OPAL_ENABLE_DEBUG
         BTL_OUTPUT(("interface %s instance %i: bandwidth %d latency %d\n", if_name, i,
                     btl->super.btl_bandwidth, btl->super.btl_latency));
--- a/opal/util/ethtool.c
+++ b/opal/util/ethtool.c
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2016 Karol Mroz.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "opal_config.h"
+
+#include <string.h>
+#include <limits.h>
+
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+#ifdef HAVE_NET_IF_H
+#include <net/if.h>
+#endif
+#ifdef HAVE_LINUX_ETHTOOL_H
+#include <linux/ethtool.h>
+#endif
+#ifdef HAVE_SYS_IOCTL_H
+#include <sys/ioctl.h>
+#endif
+#ifdef HAVE_LINUX_SOCKIOS_H
+#include <linux/sockios.h>
+#endif
+
+#include "opal/util/ethtool.h"
+#include "opal/util/if.h"
+
+#if !defined(HAVE_DECL_ETHTOOL_CMD_SPEED)
+static inline unsigned int
+ethtool_cmd_speed(const struct ethtool_cmd *ep)
+{
+    return (ep->speed_hi << 16) | ep->speed;
+}
+#endif
+
+/*
+ * Obtain an appropriate bandwidth for the interface if_name. On Linux, we
+ * get this via an ioctl(). Elsewhere or in the error case, we return the
+ * speed as 0.
+ */
+unsigned int
+opal_ethtool_get_speed (const char *if_name)
+{
+    unsigned int speed = 0;
+
+#if defined(HAVE_DECL_SIOCETHTOOL) && defined(HAVE_STRUCT_IFREQ) && defined(HAVE_STRUCT_ETHTOOL_CMD)
+    int sockfd;
+    struct ifreq ifr;
+    struct ethtool_cmd edata = {
+        .cmd = ETHTOOL_GSET,
+    };
+
+    sockfd = socket(PF_INET, SOCK_DGRAM, 0);
+    if (sockfd < 0) {
+        goto out;
+    }
+
+    memset(&ifr, 0, sizeof(struct ifreq));
+    strncpy(ifr.ifr_name, if_name, IF_NAMESIZE);
+    ifr.ifr_data = (char *)&edata;
+
+    if (ioctl(sockfd, SIOCETHTOOL, &ifr) < 0) {
+        goto out;
+    }
+
+    speed = ethtool_cmd_speed(&edata);
+    if (UINT_MAX == speed) {
+        speed = 0;
+    }
+
+out:
+    close(sockfd);
+    return speed;
+#else
+    return speed;
+#endif
+}
--- a/opal/util/ethtool.h
+++ b/opal/util/ethtool.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2016 Karol Mroz.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#ifndef OPAL_ETHTOOL_H
+#define OPAL_ETHTOOL_H
+
+/*
+ * Obtain an appropriate bandwidth for the interface if_name. On Linux, we
+ * get this via an ioctl(). Elsewhere or in the error case, we return the
+ * speed as 0.
+ */
+unsigned int opal_ethtool_get_speed(const char *if_name);
+
+#endif
--- a/configure.ac
+++ b/configure.ac
@@ -577,7 +577,8 @@ AC_CHECK_HEADERS([alloca.h aio.h arpa/inet.h dirent.h \
     dlfcn.h execinfo.h err.h fcntl.h grp.h libgen.h \
     libutil.h memory.h netdb.h netinet/in.h netinet/tcp.h \
     poll.h pthread.h pty.h pwd.h sched.h \
-    strings.h stropts.h sys/fcntl.h sys/ipc.h sys/shm.h \
+    strings.h stropts.h linux/ethtool.h linux/sockios.h \
+    net/if.h sys/fcntl.h sys/ipc.h sys/shm.h \
     sys/ioctl.h sys/mman.h sys/param.h sys/queue.h \
     sys/resource.h sys/select.h sys/socket.h sys/sockio.h \
     sys/stat.h sys/statfs.h sys/statvfs.h sys/time.h sys/tree.h \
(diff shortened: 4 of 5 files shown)


kmroz:
@jsquyres - Some comments/questions:

I'm marking this as DNM for now (it has debugs, etc).

1. I'm pretty green when it comes to the AC_* stuff. Is there some better way to ensure that all the necessary headers are in place _along with_ SIOCETHTOOL?
2. The ethool_get_speed() might be happier at a higher level so that other btl's could take advantage?
3. #define MCA_BTL_TCP_BTL_BANDWIDTH/LATENCY - keep, remove?

Thanks!

lanl-ompi:
Test FAILed.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -467,6 +481,48 @@ static int mca_btl_tcp_component_close(void)
...
+{
+#if !HAVE_DECL_SIOCETHTOOL
+    return MCA_BTL_TCP_BTL_BANDWIDTH;
+#else
+    int sockfd;
+    struct ifreq ifr;
+    struct ethtool_cmd edata = {
+        .cmd = ETHTOOL_GSET,
+    };
+    unsigned int speed = 0;
```
jsquyres: Should speed be set to MCA_BTL_TCP_BTL_BANDWIDTH here, just in case you `goto out`?

I ask because in the non-SIOCETHTOOL case, you return MCA_BTL_TCP_BTL_BANDWIDTH.  But in the error cases, you seem to return 0 (i.e., where you `goto out` and if `ethtool_cmd_speed()` returns -1).

If you want to distinguish between returning an error and the returned speed, then we'll probably want to return the error code, and pass in a `uint32 *speed` parameter to return the speed.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -467,6 +481,48 @@ static int mca_btl_tcp_component_close(void)
...
 }
 
+/*
+ * Set an appropriate bandwidth for the btl instance. On Linux, we obtain this
+ * via an ioctl(). All other OSs, we return MCA_BTL_TCP_BTL_BANDWIDTH.
+ */
+static unsigned int
+mca_btl_tcp_ethtool_get_speed (const char *if_name)
+{
+#if !HAVE_DECL_SIOCETHTOOL
```
jsquyres: You might want to change this to:

```c
#if !defined(HAVE_DECL_SIOCETHTOOL) || !defined(HAVE_DECL_STRUCT_IFREQ) || !defined(HAVE_DECL_STRUCT_ETHTOOL_CMD)
```

This is a bit pedantic, I admit, but it will cover us if we end up on a system that defines `SIOCETHTOOL`, but *doesn't* define, for example, `struct ethtool_cmd` (e.g., they pre-emptively defined `SIOCETHTOOL` and just always return `ENOTIMPL` or somesuch).  This case may never happen, but we've seen weird things like this before on esoteric systems.

opal/mca/btl/tcp/btl_tcp_component.c
kmroz: Good point, I will examine this.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -467,6 +481,48 @@ static int mca_btl_tcp_component_close(void)
...
+
+    memset(&ifr, 0, sizeof(struct ifreq));
+    strncpy(ifr.ifr_name, if_name, IF_NAMESIZE); /* TODO: IFNAMSIZ instead? */
+    ifr.ifr_data = (char *)&edata;
+
+    if (ioctl(sockfd, SIOCETHTOOL, &ifr) < 0) {
+        goto out;
+    }
+
+    speed = ethtool_cmd_speed(&edata);
```
jsquyres: Do we need to check for the presence of this `ethtool_cmd_speed` function?  E.g., see http://stackoverflow.com/questions/26031152/linux-ethtool-command

opal/mca/btl/tcp/btl_tcp_component.c
kmroz: Definitely this way, thanks!

jsquyres:
@kmroz Answers:

1. I think your configury looks great (with one addition: do we need to check for `ethtool_get_speed()`?  See my inline comment about that).
1. Good call.  Moving `mca_btl_tcp_ethtool_get_speed()` to `opal/util` might not be a bad idea (e.g., renaming it to `opal_ethtool_get_speed()`.  This would allow it to be used anywhere.
1. I like your `BANDWIDTH/LATENCY` defines -- works nicely for the alternate path when we don't have `SIOCETHTOOL` and friends.

@hppritcha Looks like the LANL-Cray XC jenkins melted down for some reason...?

opal/mca/btl/tcp/btl_tcp_component.c
kmroz: Will look into it. Thanks!

jsquyres:
One last comment: the MCA param help message for the bandwidth is:

> Approximate maximum bandwidth of interconnect (0 = auto-detect value at run-time [not supported in all all modules], >= 1 = bandwidth in Mbps)

This should probably be updated, because the default is 100 -- not 0.  Does this MCA param come from ob1, or the BTL framework?  Or is it created in the TCP BTL itself?

It *may* be desirable to keep the default as 0, and only call get_speed later during the init process if the value is still 0.

kmroz:
I forgot to ask about this one actually. So ```--mca btl_tcp_bandwidth_tun0``` gets generated by, I think:

```C
static int mca_btl_tcp_create(int if_kindex, const char* if_name)
{
...
mca_btl_tcp_param_register_uint(param, NULL, btl->super.btl_bandwidth, OPAL_INFO_LVL_5, &btl->super.btl_bandwidth);
...
}
```
So I presume in the TCP BTL itself, yes?

That help message confused me. I thought it meant that _if_ the user sets ```--mca btl_tcp_bandwidth_tun0 0```, the system will just send. Otherwise, btl_bandidth is set as above.

That's why I put the `mca_btl_tcp_get_ethtool_speed()` call _before_ the user options so that the user could overwrite what ethtool found.

It's very possible my understanding is a *little* off ;)

jsquyres:
You're quite close. 

We have a bunch of MCA Params that we default to some sentinel value, meaning "auto detect". Hence, later in the init sequence, if the value is still that sentinel, we (try to) auto detect. If it's not that sentinel, we know the user set it and we shouldn't change it. 

I'm saying that this bandwidth param should probably fall in that same category.

kmroz:
@jsquyres - I see. Thanks for the review and insight. I'll fix this up when I have some off time.

hppritcha:
bot:retest

kmroz:
@jsquyres - Updated the PR per your suggestions. Things that have changed:

1. Moved ethtool calls one level up to opal/util -> created `ethtool.[ch]`
2. Hopefully took care of potential of non-existent `ethtool_cmd_speed()` kernel API.
  - Added ethtool_cmd_speed() to `AC_CHECK_DECLS`. Would have been nice to add to `AC_CHECK_FUNCS`, but as this is a `static inline` coming from a kernel header, I don't think we can use `AC_CHECK_FUNCS`?
3. `btl_latency` and `btl_bandwidth` start off as 0 (to align with auto-detect help message). If latency/bandwidth is not set by the user _or_ is set to 0 by the user, we perform auto-detection.
  - On all error cases of bandwidth auto-detect, function returns 0 and we set `btl_bandwidth = MCA_BTL_TCP_BTL_BANDWIDTH`.
  - Since we have no latency auto-detection method, we just `btl_latency = MCA_BTL_TCP_BTL_LATENCY`.

Some small questions:
What do we do about copyrights for new files? Do we have an Open MPI org that we use?
The bandwidth and latency bit shifting when dealing with interfaces with multiple links looks a bit odd to me. For example, an interface capable of 100 Mbits/sec and with 2 links, will have the 1st link's btl_bandwidth == 100, the second == 50. Assuming more links, each subsequent link's btl_bandwidth would also be set to 50. Is this accurate? The same kind of calculation happens for btl_latency.

Thanks!

opal/util/ethtool.c
```diff
@@ -0,0 +1,66 @@
+#include "opal_config.h"
```
jsquyres: Please put the usual copyright comment block at the top of this file.

jsquyres: Meaning, something like this:

```c
/*
 * Copyright (c) 2016 Karol Mroz.  All rights reserved.
 * $COPYRIGHT$
 *
 *
 * $HEADER$
 */
```

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -513,6 +520,27 @@ static int mca_btl_tcp_create(int if_kindex, const char* if_name)
...
+        /* Only attempt to auto-detect bandwidth and/or latency if it is 0.
+         *
+         * If detection fails to return anything other than 0, set a default
+         * bandwidth and latency.
+         */
+        if (0 == btl->super.btl_bandwidth) {
+            unsigned int speed = opal_ethtool_get_speed(if_name);
+            btl->super.btl_bandwidth = (speed == 0) ? MCA_BTL_TCP_BTL_BANDWIDTH : speed;
+            if (i > 0) {
+                btl->super.btl_bandwidth >>= 1;
```
jsquyres: I don't quite understand this computation (and the corresponding computation for latency).  What is its purpose?

opal/util/ethtool.c
```diff
@@ -0,0 +1,66 @@
...
+#ifdef HAVE_LINUX_SOCKIOS_H
+#include <linux/sockios.h>
+#endif
+
+#include "opal/util/ethtool.h"
+#include "opal/util/if.h"
+
+/*
+ * Set an appropriate bandwidth for the btl instance. On Linux, we obtain this
+ * via an ioctl(). Elsewhere or in the error case, we return the speed as 0.
```
jsquyres: Minor nit: don't mention the BTL in the comment.  This routine can/will be used elsewhere.

opal/util/ethtool.c
```diff
@@ -0,0 +1,66 @@
...
+ */
+unsigned int
+opal_ethtool_get_speed (const char *if_name)
+{
+    unsigned int speed = 0;
+
+#if !defined(HAVE_DECL_SIOCETHTOOL) || !defined(HAVE_STRUCT_IFREQ) || !defined(HAVE_STRUCT_ETHTOOL_CMD)
```
jsquyres: I think you also need to also check for ethtool_cmd_speed (you check for it in configure).

opal/util/ethtool.h
```diff
@@ -0,0 +1,25 @@
+#ifndef OPAL_ETHTOOL_H
```
jsquyres: Same thing with the "usual" copyright headers here.

opal/util/ethtool.c
jsquyres: Oh -- disregard; I see you define `ethtool_cmd_speed()` if it does not exist.

opal/util/ethtool.h
```diff
@@ -0,0 +1,25 @@
...
+#else
+static inline unsigned int
+ethtool_cmd_speed(const struct ethtool_cmd *ep)
+{
+    return (ep->speed_hi << 16) | ep->speed;
+}
+#endif
+#endif
+
+unsigned int opal_ethtool_get_speed(const char *if_name);
```
jsquyres: Please add at least a 1-line description of this function in a comment above the function declaration.

opal/util/ethtool.h
```diff
@@ -0,0 +1,25 @@
+#ifndef OPAL_ETHTOOL_H
+#define OPAL_ETHTOOL_H
+
+#include "opal_config.h"
+
+#if !defined(HAVE_DECL_ETHTOOL_CMD_SPEED)
```
jsquyres: Do these `ethtool_cmd_speed()` definitions need to be in this .h file?  I think they can probably go in `ethtool.c`, since we don't need anyone else to be able to call `ethtool_cmd_speed()`.

opal/util/ethtool.c
kmroz: Will do, thanks.

opal/mca/btl/tcp/btl_tcp_component.c
kmroz: @jsquyres - I commented about this above (pasting here):

_The bandwidth and latency bit shifting when dealing with interfaces with multiple links looks a bit odd to me. For example, an interface capable of 100 Mbits/sec and with 2 links, will have the 1st link's btl_bandwidth == 100, the second == 50. Assuming more links, each subsequent link's btl_bandwidth would also be set to 50. Is this accurate? The same kind of calculation happens for btl_latency._

Since this logic was in the original code, and I don't have the historical context, I've included it here as well.

@bosilca - Hi George, this bandwidth/latency shifts appear to have come from you. Could you please comment? Thanks!

opal/util/ethtool.c
kmroz: Will remove.

opal/util/ethtool.h
kmroz: Will do.

opal/util/ethtool.h
kmroz: You're likely right... will take a look, thanks.

kmroz:
@jsquyres - Your latest batch of comments addressed. Waiting on some insight regarding:
`btl->super.btl_bandwidth >>= 1; ` etc.

bosilca:
In the context where the multi link has been developed (Playstation 4), due to the hypervisor overhead a single link was unable to use the entire capacity of the network card. Thus, in order to improve the aggregated bandwidth, a second and 3rd links were necessary, each one with a lower bandwidth. Outside that particular context, the only interest in having multiple links is in increasing the size of the kernel buffer (and thus the potential overlapping). In this case, the latency will remain the same but it is difficult to define what happens with the bandwidth per link.

kmroz:
@bosilca - Thanks for the quick response, George. I've added the same type of shift in the auto-detect case in this PR. Are you OK with these changes? Thanks!

bosilca:
👍

opal/util/ethtool.c
```diff
@@ -0,0 +1,93 @@
...
+#ifdef HAVE_LINUX_SOCKIOS_H
+#include <linux/sockios.h>
+#endif
+
+#include "opal/util/ethtool.h"
+#include "opal/util/if.h"
+
+#if !defined(HAVE_DECL_ETHTOOL_CMD_SPEED)
+#if !defined(HAVE_STRUCT_ETHTOOL_CMD)
+struct ethtool_cmd;
```
jsquyres: This will ultimately result in an error on a system that doesn't have `struct ethtool_cmd`, right?  I.e., you forward reference it here and then use it in the declaration of `ethtool_cmd_speed()` here.  But then I don't see it defined anywhere, even though it is used down on line 67.  I think in this conditional block, you need to declare `struct ethtool_cmd` as well.

kmroz: Hmmm...i thought it would not be necessary below as its being used from the #else case of the !defined(HAVE_STRUCT_ETHTOOL_CMD) conditional?

jsquyres: Ah, I see.  Then why define `ethtool_cmd_speed()` at all for this case?

kmroz: @jsquyres - So I think what you're saying is, since `opal_ethtool_get_speed()` is just a simple `return 0` if `... || !defined(HAVE_STRUCT_ETHTOOL_CMD)`, then all we need is:

```C
#if !defined(HAVE_DECL_ETHTOOL_CMD_SPEED)
static inline unsigned int
ethtool_cmd_speed(const struct ethtool_cmd *ep)
{
    return (ep->speed_hi << 16) | ep->speed;
}
#endif
```
Am I parsing you correctly?

kmroz:
@jsquyres - Not sure how to parse the Mellanox test failure, however I think it looks unrelated.

kmroz:
@jsquyres - Hi Jeff, I've simplified the ```#if !defined(HAVE_DECL_ETHTOOL_CMD_SPEED)```. Anything else that you see that might need adjustment? Thanks!

opal/util/ethtool.c
```diff
@@ -0,0 +1,84 @@
...
+ * Obtain an appropriate bandwidth for the interface if_name. On Linux, we
+ * get this via an ioctl(). Elsewhere or in the error case, we return the
+ * speed as 0.
+ */
+unsigned int
+opal_ethtool_get_speed (const char *if_name)
+{
+    unsigned int speed = 0;
+
+#if !defined(HAVE_DECL_SIOCETHTOOL) || !defined(HAVE_STRUCT_IFREQ) || !defined(HAVE_STRUCT_ETHTOOL_CMD)
```
jsquyres: I'm sorry -- one last nit-pick, I swear... 😄 

I tend to prefer positive logic (vs. negative logic) -- I just find it easier to read/parse.  E.g.:

```c
#if defined(HAVE_DECL_SIOCETHTOOL) && defined(HAVE_STRUCT_IFREQ) && defined(HAVE_STRUCT_ETHTOOL_CMD)
    /*...long block with all the code ... */
#else
    return speed;
#endif
```

jsquyres:
@kmroz BTW, yes, the Mellanox error is unrelated.  Will hopefully be fixed by https://github.com/open-mpi/ompi/pull/1675

kmroz:
@jsquyres - Cool, will go for positive logic. Thanks!

kmroz:
@jsquyres - Positive logic addressed.

PR #1640, opened 2016-05-05
debuggers: remove some useless code

jsquyres:
MPIR-1.0 specifies that the following symbols are only relevant in the starter process:

- MPIR_Breakpoint
- MPIR_being_debugged
- MPIR_debug_state
- MPIR_debug_abort_string

I.e., the code filling in values in these various symbols was useless / never used.

MPIR-1.1 will define that MPIR_being_debugged *is* relevant in MPI processes.  That symbol is currently defined in libopen-rte (which is currently causing a duplicate symbol error for static builds -- this commit fixes that error), and is therefore still available for MPI processes.

@bosilca @rhc54 Please review.


bosilca:
👍

rhc54:
👍

Issue #1635, opened 2016-05-05
romio build failure on OpenBSD-5.7


PHHargrove:
Just wanted to make a record of this other than the devel list posting.

As described in more detail in https://www.open-mpi.org/community/lists/devel/2015/05/17449.php OpenBSD-5.7 cannot build the romio version embedded in Open MPI v1.8, v1.10 or v2.0.
The problem, not present in OpenBSD-5.6 and earlier, relates to struct statfs which is not defined as romio expects:

```
/home/phargrov/OMPI/openmpi-1.8.6rc1-openbsd5-amd64/openmpi-1.8.6rc1/ompi/mca/io/romio/romio/adio/common/ad_fstype.c: In function 'ADIO_FileSysType_fncall': 
/home/phargrov/OMPI/openmpi-1.8.6rc1-openbsd5-amd64/openmpi-1.8.6rc1/ompi/mca/io/romio/romio/adio/common/ad_fstype.c:374: error: 'struct statfs' has no member named 'f_type' 
/home/phargrov/OMPI/openmpi-1.8.6rc1-openbsd5-amd64/openmpi-1.8.6rc1/ompi/mca/io/romio/romio/adio/common/ad_fstype.c:398: error: 'struct statfs' has no member named 'f_type' 
/home/phargrov/OMPI/openmpi-1.8.6rc1-openbsd5-amd64/openmpi-1.8.6rc1/ompi/mca/io/romio/romio/adio/common/ad_fstype.c:419: error: 'struct statfs' has no member named 'f_type' 
*** Error 1 in ompi/mca/io/romio/romio (Makefile:3335 'adio/common/ad_fstype.lo') 
```

As of the current mpich-3.2 there does not appear to be any upstream fix.

hppritcha:
closing as we no longer support ROMIO.

PR #1643, opened 2016-05-06
io/romio: fix filesystem type check on OpenBSD

ggouaillardet:
Thanks Paul Hargrove for the report

### DIFF
--- a/ompi/mca/io/romio314/romio/adio/common/ad_fstype.c
+++ b/ompi/mca/io/romio314/romio/adio/common/ad_fstype.c
@@ -346,6 +346,8 @@ static void ADIO_FileSysType_fncall(const char *filename, int *fstype, int *erro
 # endif
 
 
+# ifdef ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE
+
 #ifdef ROMIO_GPFS
     if (fsbuf.f_type == GPFS_SUPER_MAGIC) {
 	*fstype = ADIO_GPFS;
@@ -412,6 +414,8 @@ static void ADIO_FileSysType_fncall(const char *filename, int *fstype, int *erro
 	    return;
     }
 # endif
+
+# endif /*ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE */
 
 # ifdef ROMIO_UFS
     /* if UFS support is enabled, default to that */
--- a/ompi/mca/io/romio314/romio/configure.ac
+++ b/ompi/mca/io/romio314/romio/configure.ac
@@ -1360,11 +1360,53 @@ if test "$pac_cv_have_statfs" = yes ; then
     AC_DEFINE(HAVE_STRUCT_STATFS,1,[Define if struct statfs can be compiled])
 fi
 
+AC_MSG_CHECKING([for f_type member of statfs structure])
+AC_TRY_COMPILE([
+#ifdef HAVE_SYS_VFS_H
+#include <sys/vfs.h>
+#endif
+#ifdef HAVE_SYS_STATVFS_H
+#include <sys/statvfs.h>
+#endif
+#ifdef HAVE_SYS_PARAM_H
+#include <sys/param.h>
+#endif
+#ifdef HAVE_SYS_MOUNT_H
+#include <sys/mount.h>
+#endif
+#ifdef HAVE_STRING_H
+#include <string.h>
+#endif
+    ],[
+    struct statfs f;
+    memset(&f, 0, sizeof(f));
+    f.f_type = 0;
+    ],
+    pac_cv_have_statfs_f_type=yes,
+    pac_cv_have_statfs_f_type=no
+)
+AC_MSG_RESULT($pac_cv_have_statfs_f_type)
+if test $pac_cv_have_statfs_f_type = yes ; then
+    AC_DEFINE(ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE, 1,[Define if statfs has f_type])
+fi
+
 AC_MSG_CHECKING([for f_fstypename member of statfs structure])
 AC_TRY_COMPILE([
+#ifdef HAVE_SYS_VFS_H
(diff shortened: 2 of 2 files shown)


PHHargrove:
@ggouaillardet :+1:
I have successfully built 2.0.0rc2 +pull/1643.patch, with romio enabled, on both OpenBSD-5.9/amd64 and /i386.

Just a reminder that this should probably be pushed upstream to MPICH as well.

ggouaillardet:
@PHHargrove i overwrote the PR with a better one, that is more likely to be accepted in mpich

@roblatham00 here is the patch for mpich (master), could you please review it and apply it to mpich ?

```patch
diff --git a/src/mpi/romio/adio/common/ad_fstype.c b/src/mpi/romio/adio/common/ad_fstype.c
index 51893c0..dbafde3 100644
--- a/src/mpi/romio/adio/common/ad_fstype.c
+++ b/src/mpi/romio/adio/common/ad_fstype.c
@@ -346,6 +346,8 @@ static void ADIO_FileSysType_fncall(const char *filename, int *fstype, int *erro
 # endif
 
 
+# ifdef ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE
+
 #ifdef ROMIO_GPFS
     if (fsbuf.f_type == GPFS_SUPER_MAGIC) {
 	*fstype = ADIO_GPFS;
@@ -413,6 +415,8 @@ static void ADIO_FileSysType_fncall(const char *filename, int *fstype, int *erro
     }
 # endif
 
+# endif /*ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE */
+
 # ifdef ROMIO_UFS
     /* if UFS support is enabled, default to that */
     *fstype = ADIO_UFS;
diff --git a/src/mpi/romio/configure.ac b/src/mpi/romio/configure.ac
index 250391d..6d15dd7 100644
--- a/src/mpi/romio/configure.ac
+++ b/src/mpi/romio/configure.ac
@@ -1178,11 +1178,53 @@ if test "$pac_cv_have_statfs" = yes ; then
     AC_DEFINE(HAVE_STRUCT_STATFS,1,[Define if struct statfs can be compiled])
 fi
 	  
+AC_MSG_CHECKING([for f_type member of statfs structure])
+AC_TRY_COMPILE([
+#ifdef HAVE_SYS_VFS_H
+#include <sys/vfs.h>
+#endif
+#ifdef HAVE_SYS_STATVFS_H
+#include <sys/statvfs.h>
+#endif
+#ifdef HAVE_SYS_PARAM_H
+#include <sys/param.h>
+#endif
+#ifdef HAVE_SYS_MOUNT_H
+#include <sys/mount.h>
+#endif
+#ifdef HAVE_STRING_H
+#include <string.h>
+#endif
+    ],[
+    struct statfs f;
+    memset(&f, 0, sizeof(f));
+    f.f_type = 0;
+    ],
+    pac_cv_have_statfs_f_type=yes,
+    pac_cv_have_statfs_f_type=no
+)
+AC_MSG_RESULT($pac_cv_have_statfs_f_type)
+if test $pac_cv_have_statfs_f_type = yes ; then
+    AC_DEFINE(ROMIO_HAVE_STRUCT_STATFS_WITH_F_TYPE, 1,[Define if statfs has f_type])
+fi
+
 AC_MSG_CHECKING([for f_fstypename member of statfs structure])
 AC_TRY_COMPILE([
+#ifdef HAVE_SYS_VFS_H
+#include <sys/vfs.h>
+#endif
+#ifdef HAVE_SYS_STATVFS_H
+#include <sys/statvfs.h>
+#endif
+#ifdef HAVE_SYS_PARAM_H
 #include <sys/param.h>
+#endif
+#ifdef HAVE_SYS_MOUNT_H
 #include <sys/mount.h>
+#endif
+#ifdef HAVE_STRING_H
 #include <string.h>
+#endif
     ],[
     struct statfs f;
     memset(&f, 0, sizeof(f));
```

PHHargrove:
@ggouaillardet 
The new PR works for me.

ggouaillardet:
Refs #1635 
@roblatham00 to be clear,the issue started with OpenBSD 5.7, and occurs in current OpenBSD 5.9

ibm-ompi:
Test passed.

lanl-ompi:
Test FAILed.

roblatham00:
Am I reading your changes to ad_fstype. c correctly?  The tests for panfs, lustre, pvfs2, and xfs depend on f_type, so on OpenBSD the only auto-detected file system can be UFS? 

No, it's worse than that.  OpenBSD implements f_fstypename , and the only file system we look for in that case is NFS.  

So this patch is a build-time fix, and leaves us in no worse state than before.  Auto-detecting the file system in the unusual case of Lustre or GPFS on an OpenBSD client still remains unsupported.

jsquyres:
@ggouaillardet @roblatham00  I still advise against using `AC_TRY_COMPILE`.  It's both deprecated and `AC_CHECK_MEMBERS` was created for exactly this purpose.

roblatham00:
thanks, @jsquyres...  veering a little off topic for this specifci but, but I'll make that a subsequent patch.   I was surprised to find only 60 instances of obsolete macros in ROMIO's configure.ac  (I also learned about autoconf's --warnings=all).

ggouaillardet:
@roblatham00 yes you are reading the patch correctly

if ```struct statfs``` has no ```f_type``` field (i am only aware this happens in OpenBSD >= 5.7), then romio only considers ```ufs``` and ```nfs```. i am not aware GPFS, lustre, panfs or xfs are available on OpenBSD, and i assumed they are not. i definitely should have documented that.

if needed, i am fine reworking the patch and use ```strncmp``` with ```f_fstypename``` to (try) detecting non ```ufs```/```nfs``` when ```f_type``` is not available.

ggouaillardet:
@roblatham00 i replaced some deprecated macros in open-mpi/ompi-release#1178 (see the full patch in the comment). i hope this helps.

Issue #1623, opened 2016-05-03
mindist mapper broken on master


jsquyres:
A bunch of Mellanox Jenkins runs have been failing on master with this kind of error:

```
18:32:14 + /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace-2/ompi_install1/bin/mpirun -np 8 --map-by dist -mca rmaps_dist_device mlx4_0 -x TEST_CLOSEST_NUMA -x TEST_PHYS_ID_COUNT -x TEST_CORE_ID_COUNT /scrap/jenkins/jenkins/jobs/gh-ompi-master-pr/workspace-2/jenkins_scripts/jenkins/ompi/mindist_test
18:32:16 
18:32:16 Success rank - 4: only one NUMA is scheduled.
18:32:16 
18:32:16 Success rank - 6: only one NUMA is scheduled.
18:32:16 
18:32:16 Success rank - 0: only one NUMA is scheduled.
18:32:16 
18:32:16 Success rank - 2: only one NUMA is scheduled.
18:32:16 
18:32:16 Error rank - 5: scheduled on wrong NUMA node - 1, should be 0
18:32:16 
18:32:16 Error rank - 3: scheduled on wrong NUMA node - 1, should be 0
18:32:16 
18:32:16 Error rank - 1: scheduled on wrong NUMA node - 1, should be 0
18:32:16 
18:32:16 Error rank - 7: scheduled on wrong NUMA node - 1, should be 0
18:32:17 -------------------------------------------------------
18:32:17 Primary job  terminated normally, but 1 process returned
18:32:17 a non-zero exit code. Per user-direction, the job has been aborted.
18:32:17 -------------------------------------------------------
18:32:17 --------------------------------------------------------------------------
18:32:17 mpirun detected that one or more processes exited with non-zero status, thus causing
18:32:17 the job to be terminated. The first process to do so was:
18:32:17 
18:32:17   Process name: [[34935,1],5]
18:32:17   Exit code:    1
18:32:17 --------------------------------------------------------------------------
```

In https://github.com/open-mpi/ompi/pull/1612#issuecomment-216546013, @jladd-mlnx reported that the PMIx external component changes (i.e., the changes from that PR) seem to be what broke this test.

@rhc54 Can you investigate?

jladd-mlnx:
I see it in Brice's unmerged PR too. HWLOC change makes sense, but how has it crept into master? Don't see it on 2.x. Strange...

[...]

PR #1648, opened 2016-05-06
Repair the processing of cmd line options that mapped to MCA params.

rhc54:
This was responsible for breaking things like ```map-by foo```

Fixes https://github.com/open-mpi/ompi/issues/1623


rhc54:
Spent a bunch of time tracking down this "error" from the Mellanox Jenkins, and I don't think the test is correct. The cmd line shown in fact works correctly:

```shell
$  mpirun -np 2 -mca mca_base_env_list 'XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E' env | grep XXX
OMPI_MCA_mca_base_env_list=XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E
XXX_E=5
XXX_D=4
XXX_C=3
OMPI_MCA_mca_base_env_list=XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E
XXX_E=5
XXX_D=4
XXX_C=3
$
```
So when you run that thru ```wc -l```, you will get an answer of 8, not 6.

rhc54:
Should note: the Jenkins test only exports valus for C, D, and E. Perhaps it was expecting that A and B would be set? If so, then I'm not sure where and how that was supposed to be done, and will have to look further at some point.

ggouaillardet:
@rhc54 I previously noticed Jenkins script explicitly export A and B before invoking mpirun

rhc54:
Sadly, there is no documentation anywhere on these tests, nor are they available to me - nor is the functionality particularly documented in the code. So I'm working my way thru them as best I can.

Progress is being made...

ggouaillardet:
I looks like there is a
set -x
in the Jenkins script, this is how I found it when reading the logs
I previously suggested all Jenkins scripts should be public (read only github repo looks fine) so anyone can understand what is tested and how. shall I add this to the agenda of the next telcon ?

rhc54:
sure - it would definitely help

jsquyres:
@jladd-mlnx @Di0gen can you please comment on this conversation?

miked-mellanox:
@ggouaillardet - the script is public and here (also mindist.c test)

https://github.com/mellanox-hpc/jenkins_scripts

ggouaillardet:
thanks @miked-mellanox 
@rhc54 here is what is broken
```bash
export XXX_A=1
echo XXX_B=2 > xxx.conf
mpirun -np 1 -am xxx.conf env | grep ^XXX
```
```XXX_B``` is not exported

git bisect indicates open-mpi/ompi@58dd41facf4c6f56066158dc739f953f4ad8d639 is the guilty commit
according to the logs open-mpi/ompi@2839484737ed9b5162f13254b91fd52bf268df32 might have been an attempt to fix it, but it did not fix this specific test case.

rhc54:
Yes, I know - I pointed that out in an issue I filed, and requested help to decipher the AMCA stuff as it isn't very well commented 😄

PR #1650, opened 2016-05-07
Move the registration of MCA params out of the init of the var system…

rhc54:
… - put them in with the rest of the OPAL MCA param registrations

### DIFF
--- a/opal/mca/base/mca_base_parse_paramfile.c
+++ b/opal/mca/base/mca_base_parse_paramfile.c
@@ -29,7 +29,7 @@
 #include "opal/mca/base/base.h"
 #include "opal/mca/base/mca_base_vari.h"
 #include "opal/util/keyval_parse.h"
-
+#include "opal/util/output.h"
 static void save_value(const char *name, const char *value);
 
 static char * file_being_read;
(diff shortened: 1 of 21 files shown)


rhc54:
@jjhursey @jladd-mlnx 

I'm going to need some assistance in tracking this problem down. I'm not sure what this test is supposed to do, but I'm guessing it has some relation to reading and processing of aggregate MCA param files? I've tried tracing that code, but there are few comments about what is supposed to happen where, and it traverses several locations - so I'm having no luck figuring it out.

Basically, I think the problem is that we are registering an MCA param too early. This PR moves the two most obvious cases to a better place, but that still didn't fix this problem.

I'd appreciate it if someone familiar with the AMCA (and I guess we now have an env file?) could take a look and suggest a fix.

rhc54:
Adding @hjelmn since he also worked on the param code.

I surrender - the code for reading MCA param files loops over and around and thru itself across multiple files. The ```-am``` option itself is rather confusing as it claims to set a path, but seems to specify the file within a path?? Very confusing.

I'm afraid this is going to remain dead until someone who remembers something about how all this is supposed to work can attack it. When they do, please (a) untangle the spaghetti, and (b) comment the code so someone else can figure out what is going on.

Meantime, please disable the AMCA-related tests.

jjhursey:
Related PR #1675

ibm-ompi:
Test passed.

rhc54:
This doesn't fully solve the envlist problem, but these are all necessary steps, so let's commit that part.

lanl-ompi:
Test FAILed.

PR #1656, opened 2016-05-08
In case, we do not build Fortran, Fortran 2008 or CXX, the regexp in …

hpcraink:
…make_manpage.pl will delete all lines up to the next ".fi" -- which for functions that do not implement the corresponding interface "as code" will have all text eliminated. Change regexp to delete this language-specific content up to the next newline+section header ".SH"

Also in case of make V=1, we'd like to see the command line, too (see Makefile.ompi-rules)

Amend OMPI_Affinity_str according to the other man-pages definitions.

### DIFF
--- a/Makefile.ompi-rules
+++ b/Makefile.ompi-rules
@@ -21,7 +21,7 @@ endif
 
 .1in.1:
 	@ echo "  BUILDING $@"
-	@ $(top_srcdir)/ompi/mpi/man/make_manpage.pl \
+	$(AM_V_CC) $(top_srcdir)/ompi/mpi/man/make_manpage.pl \
 	  --package-name='@PACKAGE_NAME@' \
 	  --package-version='@PACKAGE_VERSION@' \
 	  --ompi-date='@OMPI_RELEASE_DATE@' \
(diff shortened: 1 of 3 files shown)


hpcraink:
@ggouaillardet Could You please have a look at the changes in `make_manpage.pl`? Thanks.

ggouaillardet:
@hpcraink i am fine with this PR, thanks for filling it.

out of curiosity, did you run into the trouble you fixed (e.g. all text deleted) ?
at first glance, the root cause seems there is a missing ```.fi``` in the man page ?

hpcraink:
@ggouaillardet Thanks!
Well, some man-pages, that did not define a Fortran (Or C++ Interface) were lacking any content when shown... Took me a while to figure out, that it was the parsing through the Script, and that it's the code section missing; Now I think the \n.SH assumption is fine.

Makefile.ompi-rules
```diff
@@ -21,7 +21,7 @@ endif
 
 .1in.1:
 	@ echo "  BUILDING $@"
-	@ $(top_srcdir)/ompi/mpi/man/make_manpage.pl \
+	$(AM_V_CC) $(top_srcdir)/ompi/mpi/man/make_manpage.pl \
```
jsquyres: I think this should be `$(OMPI_V_GEN)`.  Shouldn't the `echo` statement be removed, too?

ibm-ompi:
Test passed.

jsquyres:
bot:retest

PR #1659, opened 2016-05-09
sync_builtin: check for 64-bit atomic support

hjelmn:
This commit adds an additional check for 64-bit atomic support for __sync
builtins. If 64-bit support is not available the opal_atomic_*_64 atomics
are disabled.

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -110,6 +110,9 @@ __sync_add_and_fetch(&tmp, 1);],
 
   AC_DEFINE_UNQUOTED([OPAL_ASM_SYNC_HAVE_64BIT],[$opal_asm_sync_have_64bit],
 		     [Whether 64-bit is supported by the __sync builtin atomics])
+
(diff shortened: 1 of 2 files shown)


hjelmn:
@ggouaillardet @PHHargrove I went ahead and added a check for 64-bit __sync atomics. It shouldn't be necessary (a macro exists that indicated whether a platform supports 64-bit asm) but this is safer than relying on that macro.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

PHHargrove:
Works for me now.  :+1:
-Paul

hjelmn:
@PHHargrove addressed your other concern. The second commit will fall back on inline asm on x86 if the compiler does not implement 64-bit __sync atomics.

PHHargrove:
@hjelmn 
This looks better now, using the inline asm on two IA32 platforms I tried, and  still using __sync atomics on an x86-64.

Issue #1655, opened 2016-05-07
Libraries are not being added to wrapper compiler when building with external hwloc


rhc54:
When building with an external hwloc and devel headers, I'm finding that the wrapper compilers are not getting the required flags to build the ORTE tests:

```c
$ mpicc --showme
gcc -I/home/common/openmpi/build/foobar/include/openmpi -I/home/common/local/include
-I/home/common/openmpi/build/foobar/include -pthread -L/home/common/local/lib -Wl,-rpath
-Wl,/home/common/local/lib -Wl,-rpath -Wl,/home/common/openmpi/build/foobar/lib
-Wl,--enable-new-dtags -L/home/common/openmpi/build/foobar/lib -lmpi
```

I have to manually add ```-lhwloc -lopen-rte -lopen-pal``` to get it to build. Will check to see if this impacts 2.x as well

ggouaillardet:
@rhc54 i do not understand what the issue is ...

as far as i am concerned, ```mpicc``` should only use ```-lmpi```
if you are building ORTE tests, then shouldn't you use ```ortecc``` instead ?

in my (outdated) build with all (pmix, libevent and hwloc) external libs, ```mpicc``` uses ```-lmpi```
and ```ortecc``` uses ```-lopen-rte -lopen-pal```

also, if ORTE tests do not invoke OPAL directly, then ```-lopen-pal``` should not even be needed.

of course, feel free to disagree if you consider ```mpicc``` should work to build apps that directly invoke ORTE and/or OPAL.

rhc54:
It's an interesting point, but it has worked just fine for years, which indicates that something changed recently. Note that it continues to work today _unless_ I build with external (pmix, libevent, and hwloc), which would appear to indicate a problem with how we are handling that scenario.

ggouaillardet:
how do you evidence the issue ?
make check ?
other ?

ggouaillardet:
can you also please post your configure command line, and your LD_LIBRARY_PATH ?

rhc54:
Sure - the issue can be evidenced by just changing to the orte/test/mpi directory and doing "make clean hello"

Configure line:
```shell
./configure --enable-orterun-prefix-by-default --prefix=foo --with-hwloc=bar --with-libevent=bar --with-pmix=bar --with-devel-headers
```

ggouaillardet:
ok, that is something different than what i initially expected.

[...]

PR #1663, opened 2016-05-10
hwloc/external: set WRAPPER_EXTRA_* vars in proper location

jsquyres:
WRAPPER_EXTRA flags are checked *before* the POST_CONFIG macro is
invoked.  So set them in the main CONFIG macro.

Refs #1655 -- but is not a complete fix.  Someone needs to update `opal/mca/pmix/external/configure.m4` per the discussion in https://github.com/open-mpi/ompi/issues/1655#issuecomment-218176532.


rhc54:
I'll take care of the PMIx configure change. I know @jjhursey is looking at how we support multiple PMIx release versions, but I'll clean this up for the interim.

👍

Issue #1660, opened 2016-05-09
PMIx configure logic should fail configure if external cannot be used


jjhursey:
Configure with:
```shell
./configure --with-pmix=/installs/pmix-1.1.3 --enable-mca-no-build=pmix-pmix114
```

On this particular machine PMIx 1.1.3 is installed, but `pmix_common.h` is missing a header so the configure check fails:
```shell
checking if user requested PMIx support(/installs/pmix-1.1.3)... yes
checking pmix.h usability... no
checking pmix.h presence... yes
configure: WARNING: pmix.h: present but cannot be compiled
configure: WARNING: pmix.h:     check for missing prerequisite headers?
configure: WARNING: pmix.h: see the Autoconf documentation
configure: WARNING: pmix.h:     section "Present But Cannot Be Compiled"
configure: WARNING: pmix.h: proceeding with the compiler's result
configure: WARNING:     ## ------------------------------------------------------ ##
configure: WARNING:     ## Report this to http://www.open-mpi.org/community/help/ ##
configure: WARNING:     ## ------------------------------------------------------ ##
checking for pmix.h... no
checking if user requested PMI support... no
checking which components should be disabled... pmix-pmix114 
checking which components should be direct-linked into the library... 
checking which components should be run-time loadable... all
checking which components should be static... none
checking for projects containing MCA frameworks... opal, orte, ompi, oshmem
``` 

However, when we get down to the pmix component selection it is still trying to build the pmix112
```shell
+++ Configuring MCA framework pmix
checking for no configure components in framework pmix... isolated
checking for m4 configure components in framework pmix... cray, external, pmix114, s1, s2

--- MCA component pmix:isolated (no configuration)
checking for MCA component pmix:isolated compile mode... dso
checking if MCA component pmix:isolated can compile... yes

--- MCA component pmix:cray (m4 configuration macro)
checking for MCA component pmix:cray compile mode... dso
checking for Cray PMI support... checking for CRAY_PMI... no
checking if MCA component pmix:cray can compile... no

--- MCA component pmix:external (m4 configuration macro)
checking for MCA component pmix:external compile mode... dso
checking if MCA component pmix:external can compile... no

--- MCA component pmix:pmix114 (m4 configuration macro)
checking for MCA component pmix:pmix114 compile mode... dso
configure: OPAL configuring in opal/mca/pmix/pmix114/pmix
configure: running /bin/sh './configure' 
...
configure: creating ./config.status
config.status: creating Makefile
[...]

PR #1667, opened 2016-05-11
Ensure we fail to configure if external PMIx was requested and is not found.

rhc54:
Fixes https://github.com/open-mpi/ompi/issues/1660


jjhursey:
I'll start a build of this to confirm, but from a code scan it looks ok. Thanks!

jjhursey:
👍  Per comment on Issue #1660 Thanks!

This should be moved to v2.0.0 if possible.

rhc54:
Looks like we are still hitting Jenkins issues on Mellanox server:

```shell
18:43:03 [jenkins01][[39435,1],0][btl_openib_endpoint.c:115:mca_btl_openib_endpoint_post_send] error posting send request error 12: Cannot allocate memory. size = 14
```

Issue #1654, opened 2016-05-07
Patcher issues reported on mailing list


jsquyres:
As reported by @PHHargrove http://www.open-mpi.org/community/lists/devel/2016/05/18928.php and http://www.open-mpi.org/community/lists/devel/2016/05/18930.php, and by @bosilca http://www.open-mpi.org/community/lists/devel/2016/05/18929.php, there appear to be new segv's in master -- potentially caused by patcher...?

Here's one stack trace posted by @PHHargrove on LITTLE-ENDIAN Power8:

```
Program terminated with signal SIGSEGV, Segmentation fault.

(gdb) where
#0  0x0000000000000000 in ?? ()
#1  0x00003fff897adb38 in intercept_munmap (start=0x3fff89670000, length=65536)
    at /home/phargrov/OMPI/openmpi-v2.x-dev-1410-g81e0924-linux-ppc64el-xlc/openmpi-gitclone/opal/mca/memory/patcher/memory_patcher_component.c:155
#2  0x00003fff8933bc80 in __GI__IO_setb () from /lib64/libc.so.6
#3  0x00003fff89339528 in __GI__IO_file_close_it () from /lib64/libc.so.6
#4  0x00003fff89327f74 in fclose@@GLIBC_2.17 () from /lib64/libc.so.6
#5  0x0000000010000f7c in do_test ()
    at /home/phargrov/OMPI/openmpi-v2.x-dev-1410-g81e0924-linux-ppc64el-xlc/openmpi-gitclone/ompi/debuggers/dlopen_test.c:97
#6  0x00000000100010e0 in main (argc=1, argv=0x3fffff332888)
    at /home/phargrov/OMPI/openmpi-v2.x-dev-1410-g81e0924-linux-ppc64el-xlc/openmpi-gitclone/ompi/debuggers/dlopen_test.c:135

"start" is valid:
(gdb) print *(char*)0x3fff89670000
$1 = 35 '#'

Frame 1:
155         opal_mem_hooks_release_hook (start, length, true);
```

Here's another: 

BIG-endian PPC64 w/ xlc V13.1 experiences a nearly identical failure.
However, this time gdb appears to have been able to resolve frame #0 to a PLT slot (instead of "??").

```
#0  0x00000fff8904ef88 in 00000010.plt_call.opal_mem_hooks_release_hook+0 ()
[...]

PR #1673, opened 2016-05-17
rcache: fix deadlock in multi-threaded environments

hjelmn:
This commit fixes several bugs in the registration cache code:

 - Fix a programming error in the grdma invalidation function that can
   cause an infinite loop if more than 100 registrations are
   associated with a munmapped region. This happens because the
   mca_rcache_base_vma_find_all function returns the same 100
   registrations on each call. This has been fixed by adding an
   iterate function to the vma tree interface.

 - Always obtain the vma lock when needed. This is required because
   there may be other threads in the system even if
   opal_using_threads() is false. Additionally, since it is safe to do
   so (the vma lock is recursive) the vma interface has been made
   thread safe.

 - Avoid calling free() while holding a lock. This avoids race
   conditions with locks held outside the Open MPI code.

Fixes open-mpi/ompi#1654.

### DIFF
--- a/opal/mca/btl/vader/btl_vader_module.c
+++ b/opal/mca/btl/vader/btl_vader_module.c
@@ -539,6 +539,17 @@ static void mca_btl_vader_endpoint_constructor (mca_btl_vader_endpoint_t *ep)
     ep->fifo = NULL;
 }
 
+#if OPAL_BTL_VADER_HAVE_XPMEM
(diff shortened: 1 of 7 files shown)


hjelmn:
@bosilca Please test.

jsquyres:
@gpaulsen Please also test.  Thanks.

bosilca:
@hjelmn so far so good. Not a single deadlock over few tens of runs with large number of threads. I will switch everything to an optimized mode and start an overnight series of tests. I'll update the status tomorrow.

bosilca:
👍

Issue #1676, opened 2016-05-18
openmpi-2.0.0rc2: mtl/ofi/verbs being chosen over pml/ob1/btl/openib


larrystevenwise:
I haven't figured out how I get the systems in this state, but somehow I'm seeing the cm pml is being chosen over ob1.  This results in the following error:
```
[root@stevo3 ~]# /opt/ompi-2.0.0rc2/bin/mpirun --map-by node -np 2 --host stevo3,stevo4 --allow-run-as-root --mca mpi_add_procs_cutoff 1024 --mca btl openib,sm,self /usr/local/src/osu-micro-benchmarks-5.0/mpi/pt2pt/osu_bw
[stevo4.asicdesigners.com:19152] [[32156,1],1] selected pml cm, but peer [[32156,1],0] on stevo3 selected pml ob1
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
--------------------------------------------------------------------------
MPI_INIT has failed because at least one MPI process is unreachable
from another.  This *usually* means that an underlying communication
plugin -- such as a BTL or an MTL -- has either not loaded or not
allowed itself to be used.  Your MPI job will now abort.

You may wish to try to narrow down the problem;

 * Check the output of ompi_info to see which BTL/MTL plugins are
   available.
 * Run your application with MPI_THREAD_SINGLE.
 * Set the MCA parameter btl_base_verbose to 100 (or mtl_base_verbose,
   if using MTL-based communications) to see exactly which
   communication plugins were considered and/or discarded.
--------------------------------------------------------------------------
```
Bumping up the ob1 priority works around the problem:
```
[root@stevo3 ~]# /opt/ompi-2.0.0rc2/bin/mpirun --map-by node -np 2 --host stevo3,stevo4 --allow-run-as-root --mca mpi_add_procs_cutoff 1024 --mca pml_ob1_priority 90 --mca btl openib,sm,self /usr/local/src/osu-micro-benchmarks-5.0/mpi/pt2pt/osu_bw
# OSU MPI Bandwidth Test v5.0
# Size      Bandwidth (MB/s)
1                       1.92
2                       3.96
4                       1.92
8                      12.41
16                      7.27
32                     49.66
64                     28.93
128                   192.53
256                   358.52
512                   779.84
1024                  389.50
2048                  579.23
4096                  492.21
8192                  550.24
16384                1077.49
[...]

Issue #1664, opened 2016-05-10
OpenMPI 2.0.0rc2 with openib/iWARP btl stalls running OSU/IMB benchmarks


larrystevenwise:
I have installed this OpenMPI on my lab machines. I find that the
osu_latency test runs OK. However, the osu_bw test stalls. Upon doing some more
investigation, I find the that calls to MPI_ISend  on the first node all
complete and the calls to MPI_irecv on the second node all complete. However
when the nodes go into MPI_Waitall both nodes stall. 

It is important that we identify where the problem is before openmpi releases
2.0 as final.

larrystevenwise:
1- I built openmpi 2..0.0rc2 with the following configure command

./configure --prefix=/usr/local/ --enable-mpi-thread-multiple 
--enable-mpirun-prefix-by-default --enable-debug --enable-mpi-cxx

Does not matter about mpi-thread-multiple. Tried it both ways.

larrystevenwise:
When stalled we see one rank with a pending SEND WR.  HW has consumed the WR and it appears the node has transmitted the SEND to the peer.

larrystevenwise:
We started bisecting the ompi release git repo.. For commits before march of ompi,
osu tests are working fine. Will update the culprit commit..

larrystevenwise:
Only osu_bcast and osu_bw have this issue and other OSU tests are running fine.
Tests are hanging with no traffic running.  IMB tests are also not working on
openmpi2.0.0rc2.

larrystevenwise:
@jsquyres, I think this is a blocker for 2.0.0.  Chelsio is working this as top priority.  I'm also pushing on Chelsio to regression test new RCs as they release.  This issue should not have made it to rc2 IMO...

jsquyres:
@larrystevenwise Yes, would have been great to know that this bug existed before now.  No one else in the community has Chelsio equipment with which to test.

larrystevenwise:
@jsquyres, sorry about that.  I can ask Chelsio to give some cards to someone who would test OMPI/openib/cxgb4 regularly.

jsquyres:
I think the community would prefer if *you* or *Chelsio* did regular testing.  😄

larrystevenwise:
agreed... :)

I've recommended they test rc1 of each release and then regression test each rc after that.

jsquyres:
[...]

PR #1677, opened 2016-05-18
bml/r2: always add btl progress function

hjelmn:
This commit changes the behavior of bml/r2 from conditionally
registering btl progress functions to always registering progress
functions. Any progress function beloning to a btl that is not yet in
use is registered as low-priority. As soon as a proc is added that
will make use of the btl is is re-registered normally.

This works around an issue with some btls. In order to progress a
first message from an unknown peer both ugni and openib need to have
their progress functions called. If either btl is not in use after the
first call to add_procs the callback was never happening. This commit
ensures the btl progress function is called at some point but the
number of progress callbacks is reduced from normal to ensure lower
overhead when a btl is not used. The current ratio is 1 low priority
progress callback for every 8 calls to opal_progress().

Fixes open-mpi/ompi#1676

### DIFF
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -166,22 +166,30 @@ static mca_bml_base_endpoint_t *mca_bml_r2_allocate_endpoint (ompi_proc_t *proc)
     return bml_endpoint;
 }
 
-static void mca_bml_r2_register_progress (mca_btl_base_module_t *btl)
+static void mca_bml_r2_register_progress (mca_btl_base_module_t *btl, bool hp)
 {
     if (NULL != btl->btl_component->btl_progress) {
         bool found = false;
+        size_t p;
 
-        for (size_t p = 0 ; p < mca_bml_r2.num_btl_progress ; ++p) {
+        for (p = 0 ; p < mca_bml_r2.num_btl_progress ; ++p) {
             if(mca_bml_r2.btl_progress[p] == btl->btl_component->btl_progress) {
                 found = true;
                 break;
             }
         }
 
-        if (found == false) {
-            mca_bml_r2.btl_progress[mca_bml_r2.num_btl_progress++] =
-                btl->btl_component->btl_progress;
-            opal_progress_register (btl->btl_component->btl_progress);
+        if (found == false || hp) {
+            if (found == false) {
+                mca_bml_r2.btl_progress[mca_bml_r2.num_btl_progress++] =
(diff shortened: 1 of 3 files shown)


hjelmn:
@larrystevenwise Please test.

jsquyres:
@bosilca Please also review -- this touches `opal_progress()`.

jsquyres:
@bharatpotnuri Please test (this is a possible solution for #1664).

hjelmn:
Huh, I referenced the wrong bug #. Will rebase it with the correct one later.

bharatpotnuri:
@jsquyres `commit a679cc072b9458957671886062c643f6a40b0351 bml/r2: always add btl progress function` #1677 is working fine with iWARP and fixes issue #1664.
Below is the output of `osu_bw` test without stall.
```
/usr/mpi/gcc/openmpi-2.0.0/bin/mpirun -np 2 --host peer1,peer2 --allow-run-as-root --mca btl openib,sm,self /usr/mpi/gcc/openmpi-2.0.0/tests/OSU/osu_bw

# OSU MPI Bandwidth Test v5.0
# Size      Bandwidth (MB/s)
1                       0.52
2                       1.04
4                       2.07
8                       3.95
16                      7.74
32                     16.12
64                     31.21
128                    60.22
256                   118.03
512                   222.06
1024                  403.43
2048                  678.97
4096                  961.80
8192                 1260.08
16384                1542.48
32768                2268.59
65536                2836.43
131072               3181.33
262144               3307.34
524288               3349.31
1048576              3363.01
2097152              3368.89
4194304              3370.98
```
Thanks @hjelmn .

jsquyres:
@bosilca please review. Thanks.

ibm-ompi:
Test passed.

hjelmn:
Needed for 2.0.0. Merging and opening 2.0.0 PR.

bosilca:
Have someone evaluated the impact of this change on the performance ? We should stop adding atomic operations and polluting the cache in our critical point.

hjelmn:
Ran some simple tests and it looked ok. I can do it without the atomics if that is preferred. Don't really care if the low-priority calls are made exactly 1/8 times.

hjelmn:
@bosilca Your request changes will make it so only one thread is ever in opal_progress() correct? If that is the case we should plan to remove all atomics from opal_progress().

bosilca:
The current version ensure that only one thread is active in opal_progress. However, we might want to be more liberal with the resources, allowing multiple threads to concurrently drain the networks.

PR #1678, opened 2016-05-19
Unlock the mutex prior to destructing it.

rhc54:
Thanks to Nicolas Joly for the report


rhc54:
@hppritcha Looks like Travis is bombing over the monitoring test library (no rule to make target) which is weird - obviously unrelated, but it seems strange that only some compilers are impacted. Any ideas?

jsquyres:
@rhc54 Remember that we have travis do "make -k", so it's building *everything*, even long after it hit the first compile error.  The actual error is with opal/util/ethtool.c -- should be fixed by https://github.com/open-mpi/ompi/pull/1679.  I'm waiting for some more CI completions before merging.

PR #1679, opened 2016-05-19
ethtool: better handle portability

jsquyres:
Be sure to handle the case where we don't have ethtool support at all.


jsquyres:
@Di0gen @jladd-mlnx Mellanox Jenkins failure seems to have something to do with a local UCX compilation error.

jladd-mlnx:
@jsquyres I'll take your word for it. I'll disable UCX checks in Jenkins.

jsquyres:
@jladd-mlnx Check it out here: http://bgate.mellanox.com/jenkins/job/gh-ompi-master-pr/2372/

PR #1682, opened 2016-05-20
opal/util/ethtool: use system ethtool_cmd_speed when available

ggouaillardet:
Refs: open-mpi/ompi#1679


ggouaillardet:
@jsquyres can you please double check this ?

jsquyres:
Good catch; thank you. 👍

PR #1688, opened 2016-05-23
Fix base64 implementation in pmix framework.

artpol84:
In the commit 80f07b65f16e9538aca7fc5e124d2074e7e0b69e setting of '-' marker used as the string termination sign was moved from base64 code:
from: https://github.com/open-mpi/ompi/commit/80f07b65f16e9538aca7fc5e124d2074e7e0b69e#diff-1b10896c267d2591dc2c08fd0542ab67L491
to: https://github.com/open-mpi/ompi/commit/80f07b65f16e9538aca7fc5e124d2074e7e0b69e#diff-1b10896c267d2591dc2c08fd0542ab67R189

However the decoding function wasn't fixed and still expects on extra byte at the end of the encoded string.
This leads to **data truncation during decoding** (was noticed on standalone code that was using base64 from OMPI).

### DIFF
--- a/opal/mca/pmix/base/pmix_base_fns.c
+++ b/opal/mca/pmix/base/pmix_base_fns.c
@@ -5,6 +5,8 @@
  * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -556,7 +558,7 @@ int opal_pmix_base_cache_keys_locally(const opal_process_name_t* id, const char*
     }
 
     /* search for each key in the decoded data */
-    for (offset = 0 ; offset < len && '\0' != tmp_val[offset] ; ) {
+    for (offset = 0 ; offset < len ; ) {
         /* type */
         tmp = tmp_val + offset + strlen (tmp_val + offset) + 1;
         /* size */
@@ -669,11 +671,11 @@ static inline unsigned char pmi_base64_encsym (unsigned char value) {
     assert (value < 64);
 
     if (value < 26) {
-	return 'A' + value;
+        return 'A' + value;
     } else if (value < 52) {
-	return 'a' + (value - 26);
+        return 'a' + (value - 26);
     } else if (value < 62) {
-	return '0' + (value - 52);
+        return '0' + (value - 52);
     }
 
     return (62 == value) ? '+' : '/';
@@ -681,19 +683,18 @@ static inline unsigned char pmi_base64_encsym (unsigned char value) {
 
 static inline unsigned char pmi_base64_decsym (unsigned char value) {
     if ('+' == value) {
-	return 62;
+        return 62;
     } else if ('/' == value) {
-	return 63;
+        return 63;
     } else if (' ' == value) {
-	return 64;
+        return 64;
     } else if (value <= '9') {
-	return (value - '0') + 52;
+        return (value - '0') + 52;
     } else if (value <= 'Z') {
-	return (value - 'A');
+        return (value - 'A');
     } else if (value <= 'z') {
-	return (value - 'a') + 26;
+        return (value - 'a') + 26;
     }
-
     return 64;
 }
 
@@ -715,12 +716,12 @@ static inline int pmi_base64_decode_block (const char in[4], unsigned char out[3
 
     out[0] = in_dec[0] << 2 | in_dec[1] >> 4;
     if (64 == in_dec[2]) {
-	return 1;
+        return 1;
     }
 
     out[1] = in_dec[1] << 4 | in_dec[2] >> 2;
     if (64 == in_dec[3]) {
-	return 2;
+        return 2;
     }
 
     out[2] = ((in_dec[2] << 6) & 0xc0) | in_dec[3];
@@ -736,7 +737,7 @@ static char *pmi_encode(const void *val, size_t vallen)
 
     outdata = calloc (((2 + vallen) * 4) / 3 + 2, 1);
     if (NULL == outdata) {
-	return NULL;
+        return NULL;
     }
 
     for (i = 0, tmp = outdata ; i < vallen ; i += 3, tmp += 4) {
@@ -750,7 +751,7 @@ static char *pmi_encode(const void *val, size_t vallen)
 
 static uint8_t *pmi_decode (const char *data, size_t *retlen)
 {
-    size_t input_len = (strlen (data) - 1) / 4;
+    size_t input_len = strlen (data) / 4;
     unsigned char *ret;
     int out_len;
     size_t i;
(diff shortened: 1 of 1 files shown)


artpol84:
This needs to go to v2.x as well.

artpol84:
@ggouaillardet @rhc54 @hppritcha please review.

artpol84:
:bot:retest

ggouaillardet:
that does not look quite right to me
if I understand correctly, the base64 input (aka data) is "- " terminated
the length of the binary result data is
(strlen(data)-1)*3/4
since these are integer, multiplication should be performed first

for example, if you encode 1 byte, the encoded string will be "xy-"
with this commit, 0 bytes are calloc'ed

makes sense ?

artpol84:
@ggouaillardet, disagree. You are talking about encoding logic. And this PR addresses **decoding** logic. The formula you referring wasn't changed here.
If you have `n` bytes to encode you will get `m = ceil (n / 3) * 4` encoded bytes. `n % 3 <> 0` you will have some empty bytes (equal to 64) at the end of the encoded sequence. But in any case it will be exhaustive.

In current code I was encoding 4 bytes and getting only 3 of them back:
original: 4 input bytes = 3 + 1 = > encoded 4 + 4 = 8 bytes.

With current formula you'll have  strlen() = 8 => (8  - 1) / 4 = 7 / 4 = 1 - one block to unpack instead of 2!

ggouaillardet:
so the string to be decoded (aka data) is no more "-" terminated ?
I am surprised 1 byte is not encoded into 2 ascii chars (3 if you add the "-" terminator) but 4.
anyway, I will review the encoding/decoding logic more thoroughly tomorrow

artpol84:
1. I think that strings are still '-' terminated but not by base64 code anymore, so this symbol is accounted in the plain data length already.
2. As far as I understood the code each 3 chars of the plain data are encoded with 4 characters of base64 data. And if you have 4 bytes you will end up having 2 4-byte blocks. The first will be full and contain first 3 bytes and the second one (tail) will not be full and represent only one byte. It is fine since this only happens with the last base64 block.

ggouaillardet:
ok, the encoded string length is a multiple of 4 (+1 if "-" terminated)
in pmix, pmi_decode null terminates the result, which makes no sense since the result is binary
(on top of that, this causes an overflow...)

artpol84:
Indeed we don't need to null terminate the binary.
But what do you mean by overflow. As far as I was able to see during
encoding we account null byte. And during decoding this doesn't have any
effect.

понедельник, 23 мая 2016 г. пользователь Gilles Gouaillardet написал:

> ok, the encoded string length is a multiple of 4 (+1 if "-" terminated)
> in pmix, pmi_decode null terminates the result, which makes no sense since
> the result is binary
> (on top of that, this causes an overflow...)
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/open-mpi/ompi/pull/1688#issuecomment-220970009>
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

ggouaillardet:
if we null terminate the binary, then we need to calloc +1 byte.
if the to be encoded binary is a null terminated string, then we have to encode the string and the null terminator, so a null terminated string will be decoded
in any case, we cannot null terminate in pmi_decode unless an extra byte is calloc'ed

artpol84:
I recalled why we null terminate the binary. Encoded string goes to SLURM
PMI which assumes a string. So encoded string has to be null terminated.

понедельник, 23 мая 2016 г. пользователь Gilles Gouaillardet написал:

> if we null terminate the binary, then we need to calloc +1 byte.
> if the to be encoded binary is a null terminated string, then we have to
> encode the string and the null terminator, so a null terminated string will
> be decoded
> in any case, we cannot null terminate in pmi_decode unless an extra byte
> is calloc'ed
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/open-mpi/ompi/pull/1688#issuecomment-220982156>
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
As for encoding null terminated string - I don't see any issues with this -
you just need to pass the right length to encode, base64 will handle all
null chars. In fact current OMPI meta keys have a lot of nulls.

понедельник, 23 мая 2016 г. пользователь Gilles Gouaillardet написал:

> if we null terminate the binary, then we need to calloc +1 byte.
> if the to be encoded binary is a null terminated string, then we have to
> encode the string and the null terminator, so a null terminated string will
> be decoded in any case, we cannot null terminate in pmi_decode unless an
> extra byte is calloc'ed
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/open-mpi/ompi/pull/1688#issuecomment-220982156>
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

ggouaillardet:
slurm uses ascii strings only, that is why binary is base64 encoded.
my question is why do we null terminate the decoded binary ?
binary is not sent to slurm, isn't it ?

artpol84:
That is a good point. We shouldn't need null at the end and this is a bug
in my patch. I'll fix it now.

понедельник, 23 мая 2016 г. пользователь Gilles Gouaillardet написал:

> slurm uses ascii strings only, that is why binary is base64 encoded.
> my question is why do we null terminate the decoded binary ?
> binary is not sent to slurm, isn't it ?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/open-mpi/ompi/pull/1688#issuecomment-220986718>
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
This is a good point, @ggouaillardet. I fixed that. We either need to continue allocating +1 byte or do not add the '\0'.
If my understanding is correct, there is the only one datapath that uses pmi_decode:
* https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/base/pmix_base_fns.c#L509
* https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/base/pmix_base_fns.c#L553

And in the `for` loop here:
https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/base/pmix_base_fns.c#L553

in the exit condition we check for both `len` overflow and reaching of nul-terminator. For completeness I removed nul-terminator checking there.

artpol84:
I've updated the branch but it seems that github experiences some problems so changes are not visible in the web interface yet:
https://twitter.com/githubstatus

however I can see them if I `git clone` my branch.

artpol84:
:bot:retest

lanl-ompi:
Test FAILed.

ibm-ompi:

ibm-ompi:

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

jjhursey:
Looking at the Jenkins from this it looks like it is related to the GitHub outage (the error is from doing a GitHub checkout)
```shell
 > git rev-parse b24546bf347e20600a1d0a6a893adf55978ac6aa^{commit} # timeout=10
ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
```

Give GitHub some time to resolve, and by try to trigger the bots to retest.

artpol84:
Yes,
still not everything is correctly visible. Will need to give GH some time.

artpol84:
For some reasons it shows @ggouaillardet ROMIO commit as this PR's commit even though it is already in OMPI master:
https://github.com/open-mpi/ompi/commit/b159587325aaf3a06aaba0b3f205b8c4fa032f26

Will wait for tomorrow and reopen if this won't be fixed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ibm-ompi:

ibm-ompi:

hppritcha:
I'm feeling really lucky.
bot:retest

rhc54:
👍  tested on my Slurm setup under Slurm master, both PMI and PMI2.

jladd-mlnx:
bot:retest

hjelmn:
:bot:retest:

jladd-mlnx:
bot:retest

artpol84:
what is the status? Can we merge?

artpol84:
@ggouaillardet I plan to merge it if no objections.

ggouaillardet:
@artpol84 no objection, please go ahead !

artpol84:
Thank you!

PR #1692, opened 2016-05-24
Fix the --tune problem by searching the argv for MCA params in advance of opal_init_util

rhc54:
Only search the first app_context as we historically have done - we can debate whether or not to search all app_contexts


rhc54:
Well, this gets us past the prior bottleneck - will check out the dist mapping policy later

PR #1697, opened 2016-05-24
win: add support for accumulate_ordering info key

hjelmn:
This commit adds support for the MPI-3.1 accumulate_ordering info
key. The default value is rar,war,raw,waw and is supported using an
MCA variable flag enumerator.


hjelmn:
@gpaulsen This info key was missing from master. It is targeted for 2.0.1 for MPI-3.1 completeness.

gpaulsen:
Thanks.  How did you find this?  just lucky?  or do you have some tests?

hjelmn:
Went to use it for some acc optimizations and realized it was missing. It is another key osc/rdma will be able to use in set_info.

lanl-ompi:
Test FAILed.

PR #1700, opened 2016-05-24
CMA: Fixing logic for CMA system call detection

shamisp:
The OPAL_CMA_NEED_SYSCALL_DEFS is always defined/set to 0 or 1.  Therefore
instead of checking if the macro is defined, we have to look at the value
itself.

### DIFF
--- a/opal/mca/btl/sm/btl_sm.c
+++ b/opal/mca/btl/sm/btl_sm.c
@@ -19,6 +19,7 @@
  * Copyright (c) 2013-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      ARM, Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


shamisp:
FYI @hjelmn

hjelmn:
@shamisp Good catch. btl/sm is on its way out but I will merge this once jenkins finishes.

jsquyres:
@shamisp Has ARM updated its schedule A form to include your name?

shamisp:
@jsquyres - I don't think so ? Is it a blocker ? (I can start the process)

jsquyres:
We do try to hold off on unauthorized commits until the paperwork has been emailed to IU.

Updating the schedule A is trivial; it doesn't require a lawyer's signature.

shamisp:
@jsquyres - the CLA schedule A was updated. You should have the copy. 10x

jsquyres:
Excellent; thank you!

shamisp:
@hjelmn - are we good to go ?

PR #1701, opened 2016-05-24
VADER: Replacing VADER_MAX_ADDRESS with XPMEM_MAXADDR_SIZE

shamisp:
XPMEM already defines the max address value XPMEM_MAXADDR_SIZE,
which has special handling on the kernel level. Therefore we should
avoid usage of VADER_MAX_ADDRESS, which actually causes failures
on some systems.

### DIFF
--- a/opal/mca/btl/vader/btl_vader_xpmem.c
+++ b/opal/mca/btl/vader/btl_vader_xpmem.c
@@ -5,6 +5,7 @@
  * Copyright (c) 2014      The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
+ * Copyright (c) 2016      ARM, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -21,7 +22,7 @@
 
 int mca_btl_vader_xpmem_init (void)
 {
-    mca_btl_vader_component.my_seg_id = xpmem_make (0, VADER_MAX_ADDRESS, XPMEM_PERMIT_MODE, (void *)0666);
+    mca_btl_vader_component.my_seg_id = xpmem_make (0, XPMEM_MAXADDR_SIZE, XPMEM_PERMIT_MODE, (void *)0666);
     if (-1 == mca_btl_vader_component.my_seg_id) {
         return OPAL_ERR_NOT_AVAILABLE;
     }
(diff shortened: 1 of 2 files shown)


shamisp:
FYI @hjelmn

hjelmn:
@shamisp The problem is XPMEM_MAXADDR_SIZE in the Cray version of xpmem is wrong. If xpmem attempts to attach to an address over that value hard-coded in vader an error is returned. I plan to update my version of xpmem to expose the real maximum not just -1. This was affecting the bound not the attached region.

shamisp:
On the other hand VADER_MAX_ADDRESS does not work for all systems. Do you have any other idea how to fix this ?

hjelmn:
Not sure. We may have to define VADER_MAX_ADDRESS by the architecture I fear. Otherwise we may see attach failures. I wonder if SGI ever solved this problem... I will look at their xpmem source code.

shamisp:
@hjelmn how did you come up with the number that you use now ?

hjelmn:
I got originally got the value from trial and error with the api. Confirmed it from the kernel source later.

hjelmn:
For reference the largest address that can be attached to is TASK_SIZE which is (1 << 47)  - 4096 on linux. See http://lxr.free-electrons.com/source/arch/x86/include/asm/processor.h#L747. This line enforces the max attach address https://github.com/hjelmn/xpmem/blob/master/kernel/xpmem_attach.c#L372.

hjelmn:
I wonder if it will be fine to just use the VADER_MAX_ADDRESS on x86. The problem is that linux x86 reserves the last 4096 bytes of the user virtual address space. The causes an attach in the top 2MB of the stack to fail because vader by default tries to attach to 2MB aligned/sized regions. This by definition includes the reserved page. If the linux kernel doesn't reserve the last page on other platforms we could simply use (size_t) -1 on those platforms. There may still be problems if the user attempts to further increase the attach size but that might be ok.

hjelmn:
Doing a spot check around different archs. ARM, PPC, MIPS, etc all use a power of two for TASK_SIZE.

shamisp:
@hjelmn - not a bad idea. In UCX we only register the memory that we need, so we don't have this issue.

hjelmn:
Yeah, this comes from trying to keep the rcache small so lookups remain fast.

hjelmn:
I can make a new commit to define VADER_MAX_ADDRESS as XPMEM_MAXADDR_SIZE on non-x86 systems if you want or you can submit a new patch. If another platform runs into an issue we can always special-case other platforms.

shamisp:
No problem. I will update the patch.

shamisp:
Based on the discussion I'm closing this PR, and will be submitting a new one.

PR #1704, opened 2016-05-25
Fix configure for FreePGI on OSX

hpcraink:
Our checks and the ones of libevent are somewhat flawed.
If adding multiple "-framework" to CXXFLAGS or CFLAGS, we strip
the keyword from the command-line, not good.
libevent however assumes plain gcc without testing properly
that the compiler supports -Wno-deprecated-declarations.

### DIFF
--- a/config/opal_functions.m4
+++ b/config/opal_functions.m4
@@ -391,9 +391,13 @@ AC_DEFUN([OPAL_FLAGS_UNIQ],[
                 opal_found=0
                 opal_i=`expr $opal_count + 1`
                 ;;
+        -framework)
+                opal_found=0
+                opal_i=`expr $opal_count + 1`
+                ;;
         --param)
-                ompi_found=0
-                ompi_i=`expr $ompi_count + 1`
+                opal_found=0
+                opal_i=`expr $opal_count + 1`
                 ;;
         esac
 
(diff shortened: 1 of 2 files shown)


jsquyres:
The libevent fix should probably be pushed upstream, too (along with proper indenting -- I assume you didn't re-indent here in OMPI to make the minimal change).

Otherwise, 👍

jsquyres:
I hit "submit" too soon -- I meant to mention that libevent's upstream is here on github.  You might just want to file a PR with them: https://github.com/libevent/libevent

jladd-mlnx:
@hpcraink @jsquyres The Mellanox Jenkins failure is new and legit. It's coming from Libevent. Please do not merge this without fixing.

jsquyres:
The problem did not come from this PR; the problem came from https://github.com/open-mpi/ompi/pull/1422, which I asked @bosilca not to merge.

rhc54:
Agreed - this is a harmless change.

hppritcha:
I think the MLNX jenkins failure is owing to PR #1422 which ought not to have been merged till we sorted out why that PR seemed to have broken oshmem hello world.

hjelmn:
I would go ahead and merge this one. Looks good to me.

PR #1711, opened 2016-05-25
Test commit

hjelmn:


hjelmn:
Testing a theory. This will have some warnings but makes oshmem_request_t match ompi_request_t again. If this passes mellanox jenkins then it gives an indication of where the problem is.

hjelmn:
Or I will get a random unrelated ucx failure instread.....

hjelmn:
:bot:retest:

PR #1715, opened 2016-05-25
btl/ugni: reduce overhead of progress function

hjelmn:
This commit reduces the overhead of calling the ugni progress
function. It does the following:

 - Check for new connections once every eight calls.

 - Do not call remote smsg progress unless we are connected to at
   least one remote peer.

 - Do not call rdma progress unless at least one rdma fragment is
   outstanding.

 - Check endpoint wait list size before obtaining a lock.

### DIFF
--- a/opal/mca/btl/ugni/btl_ugni.h
+++ b/opal/mca/btl/ugni/btl_ugni.h
@@ -129,6 +129,8 @@ typedef struct mca_btl_ugni_module_t {
     int nlocal_procs;
 
     volatile int active_send_count;
+    volatile int64_t connected_peer_count;
+    volatile int64_t active_rdma_count;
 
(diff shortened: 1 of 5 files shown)


hjelmn:
@hppritcha This should reduce the effect the ugni btl has on shared memory performance. I tested it on trinitite and the shared memory latency is much better now with -mca btl self,vader,ugni and there seems to be no effect on ugni latency.

hjelmn:
Might not be quite right yet. Still running some testing to make sure it is working as expected.

hjelmn:
:bot:retest:

hppritcha:
mlnx jenkins failure is unrelated to this pr.

hppritcha:
:+1:

PR #1719, opened 2016-05-26
Removal of ompi_request_lock from pml/ucx.

thananon:
request_lock in UCX is used to protect the completion of the request. There is no need for it anymore. However, there's some part that still need to make sure that I got it right in pml_ucx.c where we attach the callback function.

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -403,11 +403,9 @@ mca_pml_ucx_blocking_recv_completion(void *request, ucs_status_t status,
                     (void*)req, ucs_status_string(status), info->sender_tag,
                     info->length);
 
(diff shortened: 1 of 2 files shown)


jsquyres:
@jladd-mlnx I can see that the openib BTL is still failing -- I presume someone is working on this. Is there any way to know if the UCX fix in this PR is working?  I honestly cannot tell from the jenkins output if any UCX-specific tests are running.

jsquyres:
@jladd-mlnx Actually, I see that Mellanox wrote the UCX PML -- can you guys review this and merge if ok?

jladd-mlnx:
@jsquyres Looks fine to me. The OpenIB BTL bug has nothing to do with this; in that case OpenIB BTL is failing in a thread multiple test.

jsquyres:
@jladd-mlnx Ok, thanks.

PR #1583, opened 2016-04-26
pmix/external114: External component supporting 1.1.4

jjhursey:
**(This is not ready to merge. Posting for comment)**

 * Add an 'extern114' component that is targeted at the PMIx 1.1.4 release.
 * Add a opal_check_version macro that can help with checking version ranges.
 * Adjust the PMIx check to include getting the version information from the library.

PR #1722, opened 2016-05-27
Enable PMIx external support for both 1.1.4 and 2.0 versions

rhc54:
Stealing some pieces of Josh Hursey's PR #1583 and modifying a bit, allow the opal/pmix external component to handle both PMIx 1.1.4 and PMIx 2.0 versions. Automatically detect the version of the target external library and adjust the only two APIs that changed (PMIx_Init and PMIx_Finalize)


rhc54:
@jjhursey See what you think!

rhc54:
This is the openib malloc issue

PR #1725, opened 2016-05-27
ompi/request: fix bugs in MPI_Wait_some and MPI_Wait_any

hjelmn:
This commit fixes two bugs in MPI_Wait_any:

 - If all requests are inactive then the sync wait would hang forever
   because no requests are attached to the sync.

 - The request pointer was pointing to the request before the completed
   request which caused the wrong request to be freed or marked inactive.

MPI_Wait_some had a similar issue if all the requests were pending.

These issues were identified by MTT.

### DIFF
--- a/ompi/request/req_wait.c
+++ b/ompi/request/req_wait.c
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2010 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca Fixing MTT failures from last night. Please take a look.

thananon:
👍

bosilca:
We are still not handling correctly the error cases. I am done with the Wait* functions, I need to port it on the Test*. Will be done in about a day.

bosilca:
I would add 50cec45, d9fb59be and bfcf145 to this.

PR #1729, opened 2016-05-30
start bug fixes

hjelmn:
@bosilca I think I have the problem fixed. Took less work than I expected. In order to detect the difference between a new request (fresh from isend_init) and an incomplete buffered send both the cm and ob1 isend_init functions now mark the request as pml_complete. This allows start to accurately use pml_complete to decide if a new request is needed. I kept the remainder of the cleanup as we know that no receive request can be in this state. Let me know if you are happy with the changes. I want to attach this to the request rework as osc/pt2pt hangs without this

This replaces #1726.

### DIFF
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_data_move.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_data_move.c
@@ -1661,9 +1661,10 @@ static int ompi_osc_pt2pt_callback (ompi_request_t *request)
 
     osc_pt2pt_gc_clean (module);
 
+    ompi_osc_pt2pt_frag_start_receive (module);
+
     /* put this request on the garbage colletion list */
     osc_pt2pt_gc_add_request (module, request);
-    ompi_osc_pt2pt_frag_start_receive (module);
 
     OPAL_OUTPUT_VERBOSE((50, ompi_osc_base_framework.framework_output,
                          "finished posting receive request"));
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_data_move.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_data_move.c
@@ -1673,6 +1673,7 @@ static int ompi_osc_pt2pt_callback (ompi_request_t *request)
 
 int ompi_osc_pt2pt_frag_start_receive (ompi_osc_pt2pt_module_t *module)
 {
+    module->frag_request = MPI_REQUEST_NULL;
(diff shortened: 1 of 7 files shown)


lanl-ompi:
Test FAILed.

ibm-ompi:

ibm-ompi:

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ibm-ompi:

ibm-ompi:

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ibm-ompi:

ibm-ompi:

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

bosilca:
The code looks better than the previous version, but I did not had a chance to test it yet. I'm traveling and will not be able to check more thoughtfully before Thursday.

hjelmn:
@bosilca Let me know if you are ok with this change. I would like to get it in before 9pm EDT tonight so it can run in MTT.

bosilca:
@hjelmn the original way of testing for the request status was correct (looking at req_status being ACTIVE), while testing for MPI and PML completion can lead to issues in case the request is indeed MPI and PML complete but has never been waited on.

hjelmn:
@bosilca I still check if the request is active on req_status. If it is active OMPI_ERR_REQUEST is returned because per MPI-3.1 pp77-78 a request should be inactive. I don't like the should wording (too weak) but I think the intent is must be inactive. Setting the request as pml_complete is just a way to differentiate between a new request (not yet started) and subsequent starts. On an initial start it is not ok to just mark the request as freed because it will never be waited on. Does that make sense?

bosilca:
@hjelmn thanks for the explanation, the diff is becoming difficult to read. One last question: these requests are persistent, and you never downgrade them to more normal state. Thus, I don't see how they are released when you create a new one because of the previous one not being PML complete?

hjelmn:
Yeah, it is enough code change to make the diffs a pain to read.

In most cases the request is re-used so it doesn't need to be freed on start. The only case where this is not true is buffered sends. In that case the old request is marked as freed if it is not pml complete.

bosilca:
👍 Let's give it a try.

PR #1731, opened 2016-05-31
Provide ETIMEDOUT as the mpirun exit code if the timeout limit was hit

rhc54:


rhc54:
This is the standing openib issue - unrelated

PR #1734, opened 2016-05-31
opal/progress: make progress function registration mt safe

hjelmn:
This commit fixes a bug in opal progress registration that can cause
crashes when a progress function is registered while another thread is
in opal_progress(). Before this commit realloc is used to allocate
more space for progress functions but it is possible for a thread in
opal_progress() to try to read from the array that is freed by realloc
before the array is re-assigned when realloc returns. To prevent this
race use malloc + memcpy to fill the new array and atomically swap out
the old and new array pointers.

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -55,11 +55,11 @@ int opal_progress_spin_count = 10000;
 static opal_atomic_lock_t progress_lock;
 
 /* callbacks to progress */
-static opal_progress_callback_t *callbacks = NULL;
+static volatile opal_progress_callback_t *callbacks = NULL;
 static size_t callbacks_len = 0;
 static size_t callbacks_size = 0;
 
-static opal_progress_callback_t *callbacks_lp = NULL;
+static volatile opal_progress_callback_t *callbacks_lp = NULL;
 static size_t callbacks_lp_len = 0;
 static size_t callbacks_lp_size = 0;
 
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca Fixes an occasional crash when running multi-threaded tests. Looking a lot better after this change.

jsquyres:
@thananon Can you have a look, too?

thananon:
@jsquyres I'm not familiar with this at all. This looks great to me but I will let people with more experience say that.

opal/runtime/opal_progress.c
```diff
@@ -322,38 +325,74 @@ opal_progress_set_event_poll_rate(int polltime)
...
-    size_t index;
 
-    /* just in case there is a low-priority callback remove it */
-    (void) opal_progress_unregister (cb);
+    if (OPAL_ERR_NOT_FOUND != opal_progress_find_cb (cb, *cbs, *cbs_len)) {
+        return OPAL_SUCCESS;
+    }
 
-    opal_atomic_lock(&progress_lock);
+    fprintf (stderr, "Registering callback %p\n", cb);
```
jsquyres: This debugging statement should be removed.

hjelmn:
Looks like there is still a bug in this branch. Looking into it.

opal/runtime/opal_progress.c
hjelmn: Yup. Removed in my local branch. Debugging the crash now.

bosilca:
@hjelmn I took a quick look at this patch and I don't see how you removed the potential bug. The opal_progress continues to use the callbacks array while the other might have free it (after replacing it with the newly allocated array). I think you have to declare the 2 callback arrays as volatile to solve at least part of the issue. Also, I would just double the size from the previous allocation instead of just adding 4 (and start with a default size of 8).

hjelmn:
Makes sense. I will make those changes.

hjelmn:
@bosilca It did help in my case but it was not an optimized build. volatile will help when optimized.

hjelmn:
Fixed. Found a typo in there as well. Will merge once jenkins finishes.

PR #1735, opened 2016-05-31
atomic: add support for __atomic builtins

hjelmn:
This commit adds support for the gcc __atomic builtins. The __sync
builtins are deprecated and have been replaced by these atomics. In
addition, the new atomics support atomic exchange which was not
supported by __sync.

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -116,6 +116,84 @@ __sync_add_and_fetch(&tmp, 1);],
 ])
 
 
+AC_DEFUN([OPAL_CHECK_GCC_BUILTIN_CSWAP_INT128], [
+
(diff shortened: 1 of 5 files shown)


hjelmn:
@bosilca This is a work in progress. Might make it part of a larger change to the atomics that adds C11 support.

hjelmn:
@bosilca Can you double-check the memory-model arguments? I am pretty sure they match the intended semantics of the opal_atomic_* functions.

bosilca:
I think our code assume that opal_atomic_cmpset_32 uses the ACQUIRE memory model and not the RELAXED. IF we want to change this we should audit the code to make sure we are not assuming a stricter memory model (based on the semantic of the sync_* atomic intrinsics).

hjelmn:
I think you are correct. I updated the PR to use acquire for opal_cmpset_*.

bosilca:
Looks good. 👍

PR #1741, opened 2016-06-02
Update to 1.1.4rc3

rhc54:


rhc54:
I honestly have no idea how to interpret the Travis issue - if I'm reading it correctly, it is saying that a test called "opal_lifo" in our test subdirectory is failing, which has nothing to do with updating the PMIx code.

PR #1743, opened 2016-06-02
atomic/gcc: add check for 128-bit CAS being lock-free

hjelmn:
Compiler implementations are free to include support for atomics that
use locks. Unfortunately lock-free and lock atomics do not mix. Older
versions of llvm on OS X use locks to provide
__atomic_compare_exchange on 128-bit values but are lock-free on
64-bit values. This screws up our lifo implementation which mixes
64-bit and 128-bit atomics on the same values to improve
performance. This commit adds a configure-time check if 128-bit
atomics are lock free. If they are not then the 128-bit __atomic CAS
is disabled and we check for the __sync version as a fallback.

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -148,11 +148,21 @@ AC_DEFUN([OPAL_CHECK_GCC_BUILTIN_CSWAP_INT128], [
 
 	  CFLAGS=$CFLAGS_save
       fi
+
+      if test $atomic_compare_exchange_n_128_result = 1 ; then
+         AC_MSG_CHECKING([if __int128 atomic compare-and-swap is always lock-free])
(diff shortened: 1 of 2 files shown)


hjelmn:
Really just after a Travis check. Not sure why the new atomics are failing on Travis since I can't get them to fail on my Mac.

hjelmn:
:bot:retest:

hjelmn:
Hmm, digging deeper this looks like an OS bug. 10.9 does not implement __atomic_compare_exchange with the appropriate instruction. Instead it uses a spin lock, memcmp, and memcpy. This doesn't mix with the __atomic_compare_exchange used on the 64-bit type in push. I will probably withdraw this and add a configure check for this.

PR #1748, opened 2016-06-03
Cleanup the forced termination a bit by restoring the delay before is…

rhc54:
…suing the sigkill, and eliminating the large time loss spent checking if the proc died. The latter is responsible for a large number of test timeouts in MTT


lanl-ompi:
Test FAILed.

PR #1766, opened 2016-06-07
ompi/request: fix loop conditional

hjelmn:
This commit fixes a bug in waitany that causes the code to go past the
beginning of the request array. The loop conditional i >= 0 is invalid
since i is unsigned. Changed to loop to check (i+1) > 0.

### DIFF
--- a/ompi/request/req_wait.c
+++ b/ompi/request/req_wait.c
@@ -125,7 +125,7 @@ int ompi_request_default_wait_any(size_t count,
   after_sync_wait:
     /* recheck the complete status and clean up the sync primitives. Do it backward to
      * return the earliest complete request to the user. */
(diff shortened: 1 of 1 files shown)


hjelmn:
@jsquyres @sjeaugey @bosilca Fixes failures identified by MTT.

sjeaugey:
It fixes the issue on my side. Tried MPI_Waitany_c and MPI_Waitany_f.

 :+1:

bosilca:
i shouldn't be unsigned. I noticed you kept changing it to size_t, but I fail to see any valid reason.

hjelmn:
Because count is unsigned. Changed it because of warnings from the compiler. Simple enough to just check i+1.

PR #1768, opened 2016-06-07
btl/openib: fix cq resize calculation

hjelmn:
Before dynamic add_procs the openib_btl_size_queues was called exactly
once for non-dynamic jobs. Now the function is called on each new
connection so the calculation was wrong. Re-wrote the function to
correctly calculate the CQ size and only attempt to adjust the CQ if
the requested size has changed. This fixes a bug when using the openib
btl on psm2 hardware that is caused by the time needed to resize a
CQ. The overhead was causing udcm to timeout and fail.

### DIFF
--- a/opal/mca/btl/openib/btl_openib.c
+++ b/opal/mca/btl/openib/btl_openib.c
@@ -425,13 +425,20 @@ static int openib_btl_prepare(struct mca_btl_openib_module_t* openib_btl)
 static int openib_btl_size_queues(struct mca_btl_openib_module_t* openib_btl)
 {
     uint32_t send_cqes, recv_cqes;
-    int rc = OPAL_SUCCESS, qp;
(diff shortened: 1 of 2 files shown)


hjelmn:
@hppritcha This is the error I was telling you about. The CQ sizing was completely wrong and the overhead of resizing the CQ was breaking UDCM on omnipath hardware. This change was developed and tested on kit.

jladd-mlnx:
@hjelmn @hppritcha This looks to be a bug UCX. We are tracking a few issues related to memory hooks being deregistered after the component is unloaded. Adding `-mca pml ^ucx` fixes the hang and there is no issue when running with `-mca pml ucx`. I'll update the Jenkins test now.

jladd-mlnx:
bot:retest

PR #1769, opened 2016-06-07
RFC: ignore more local TCP interfaces by default

abouteiller:
By default, the ignored list of TCP interface includes only lo(127.0.0.1) and sppp. 

There are many interfaces that are not routable. However, the TCP btl will try to use them between nodes of a pair that both feature a similar (address range) interface. This case happen often with virtual machines, docker containers, Xeon Phi, etc. 

A similar discussion happened in the past, the solution was to add an item in the FAQ. 
https://www.open-mpi.org/community/lists/devel/2012/02/10390.php

The current proposal is to go one step further and increase the list of interfaces that are ignored by default with common cases (complete list TBD). 

One may note that some users may actually want to use the virtual machine interface. In this case, the proposed change will yield an error "some processes could not be reached etc". While this may be unsettling, this is hinting the user in the right direction as to how to fix it (i.e. alter the default if_exclude). In comparison, the current behavior is to silently select non-routable interfaces, then the program runs for a while as-if things were fine, then deadlocks (the deadlock may be delayed until multirail kicks in, thus triggering the first use of the offending non-routable interface). This gives no hint about the possible remediation and looks like to an end-user as if Open MPI has a bug rather than a system configuration issue.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -246,7 +246,7 @@ static int mca_btl_tcp_component_register(void)
     mca_btl_tcp_param_register_uint("links", NULL, 1, OPAL_INFO_LVL_4, &mca_btl_tcp_component.tcp_num_links);
     mca_btl_tcp_param_register_string("if_include", "Comma-delimited list of devices and/or CIDR notation of networks to use for MPI communication (e.g., \"eth0,192.168.0.0/16\").  Mutually exclusive with btl_tcp_if_exclude.", "", OPAL_INFO_LVL_1, &mca_btl_tcp_component.tcp_if_include);
     mca_btl_tcp_param_register_string("if_exclude", "Comma-delimited list of devices and/or CIDR notation of networks to NOT use for MPI communication -- all devices not matching these specifications will be used (e.g., \"eth0,192.168.0.0/16\").  If set to a non-default value, it is mutually exclusive with btl_tcp_if_include.",
-                                      "127.0.0.1/8,sppp",
+                                      "127.0.0.1/8,sppp,virbr,docker,mic",
                                       OPAL_INFO_LVL_1, &mca_btl_tcp_component.tcp_if_exclude);
 
     mca_btl_tcp_param_register_int ("free_list_num", NULL, 8, OPAL_INFO_LVL_5,  &mca_btl_tcp_component.tcp_free_list_num);


jsquyres:
@abouteiller and I talked about this in person a bit:

1. The original mail thread cited `vibr0`, not `vibr`.  Either the naming convention has changed, or this PR should be updated.
1. We think it might be better to exclude these interfaces because it is *far* more common to **not** want to use these interfaces (vs. the other way around).  This PR should probably also update the help message to say something along the lines of "you might want to check the btl_tcp_if_exclude list yadda yadda yadda..."

Also, the original mail in the thread @abouteiller cited is actually https://www.open-mpi.org/community/lists/devel/2012/02/10380.php.

jsquyres:
bot:retest

bosilca:
I don't think this is the right approach because frankly, we will never be able to maintain the list of undesirable interfaces. However, there is a list that we can easily maintain, aka. the list of interfaces that usually exists and works.

I would even go one step forward, and disable all interfaces by default, and instead print a message pointing to a well detailed FAQ entry on what users will have to do to make it work. Meanwhile, we can add a timeout (MCA parameter) to drop any interface where the handshake does not complete in under the timeout.

jsquyres:
@bosilca With the advent of the ethernet/PCI-device specific naming, I don't know if we can maintain a list of interfaces that work by default.  The `vibr*` (and friends) names haven't changed in a few years.  Having TCP not work for *all* users "out of the box" would be... not a good decision.

bosilca:
In the same scenario will not be more able to maintain a list of interfaces that do not work (and remember we don't have * as a regex for interfaces). We can have eth0 by default, so it will work in the many cases.

abouteiller:
The exclude patterns are prefixes (that is exclude "vibr" excludes both "virb0, virb1, virb18"), this is unlike the include patterns that are exact matches (see tcp_proc.c line 330).

jsquyres:
@abouteiller Prefixes: ah, ok!

@bosilca In RHEL 7, there is no "eth0" any more.  We could always add primitive regexp-like support into the interface include/exclude lists if we need to...?

bosilca:
What if eth11 is working and should be used, but one wants to exclude eth1 ? That code is badly broken!

jsquyres:
@bosilca Fair point.  What do you suggest?

bwbarrett:
I think we should throw this patch away.  I'm not a fan of default excluding docker networking devices, for obvious reasons.  Amazon's been working on finally bringing the reachable framework into the tcp btl (and hopefully eventually the runtime), which should help significantly here without the sharp edges for the containers / vm cases.

@jsquyres / @abouteiller , comments?  If we're not going to throw away, we should make progress.  This PR is over 15 months old and getting a bit ripe.

abouteiller:
I was looking at it 2 weeks ago after Ralph pointed out about old PRs. 

Given all the discussion it is clear that having whitelists or blacklists will not get us anywhere. 

One option that I liked is to shorten connection timeouts in the TCP BTL so that trying to use a non-routable interface does not deadlock for an inordinate amount of time.

Let me know if there is interest @bosilca @bwbarrett . If so I'll create a new PR with this idea and close this one. Otherwise we'll just close this one and wait for the reachable framework to come in.

jsquyres:
I agree -- this PR can go away.  The ultimate solution will be the `reachable` framework, I think.

bwbarrett:
@abouteiller, even with the reachable framework, shorter timeouts might be a good idea.  I'm sitting on a stack of patches for the reachable framework / components an intern worked on this year; I'll try to bang them into shape this week and slip them in before the 3.1 branch.

PR #1773, opened 2016-06-10
ompi/datatype: Fix args of HINDEXED_BLOCK

kawashima-fj:
According to MPI-3.1 P.121, `ni` for `MPI_COMBINER_HINDEXED_BLOCK` should be `2`, not `2 + count`.

This bug was introduced in 113b45b4 (when `MPI_Type_create_hindexed_block` support was added in Open MPI) and fixed partially in 7f5314ee and 8de93982. This commit fixes the remaining part.

Probably this bug has no user impact. It only consumes a bit more memory on one-sided communication (pt2pt OSC).

@bosilca Please review.


bosilca:
👍

PR #1778, opened 2016-06-13
Added MPI_THREAD_MULTIPLE support for btl/usnic.

thananon:
This PR added thread support for usnic btl.

PR #1784, opened 2016-06-15
Fix btl/usnic deadlock when the connectivity check is turned off.

thananon:
This patch fix the deadlock found in coverity test after #1778 . We will take the lock after we check that the connectivity check is needed. 

@jsquyres


jsquyres:
👍

PR #1785, opened 2016-06-15
opal/memory: disable __malloc_initialize_hook if poisoned

hjelmn:
Newer versions of gcc have "poisoned" the __malloc_initialize_hook
name and it can no longer be used. Added a configure check and
protection around its usage.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -854,7 +854,7 @@ OPAL_SEARCH_LIBS_CORE([ceil], [m])
 # -lrt might be needed for clock_gettime
 OPAL_SEARCH_LIBS_CORE([clock_gettime], [rt])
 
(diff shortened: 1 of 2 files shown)


hjelmn:
@opoplawski Can you see if this fixes the error for you? If it does we will apply this to 2.0.0.

opoplawski:
Appears to compile fine here.  Thanks.

hjelmn:
@opoplawski Great, thanks! The change in gcc+glibc will affect 1.10.x negatively as it will disable the memory hooks in that release. 2.x will use better memory hooks.

hjelmn:
Mellanox Jenkins looks like it is out to lunch.

PR #1786, opened 2016-06-16
opal/progress: use 32-bit atomics for call counter

hjelmn:
This commit fixes a compile error on 32-bit platforms. The
low-priority call counter was always using 64-bit atomics which will
not work if 64-bit atomic math is not available. Updated to use 32-bit
instead.

### DIFF
--- a/opal/mca/patcher/overwrite/patcher_overwrite_module.c
+++ b/opal/mca/patcher/overwrite/patcher_overwrite_module.c
@@ -28,9 +28,9 @@
 #include <dlfcn.h>
 #include <assert.h>
 
-#if defined(__i386__) || defined(__x86_64__) || defined(__ia64__)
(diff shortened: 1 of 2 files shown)


hjelmn:
@PHHargrove Fixed PPC32

hjelmn:
@PHHargrove I think the second commit might help with your ppc64 problem. I can't reproduce before or after this commit. Can you send me your configure line for powerpc64le?

PHHargrove:
@hjelmn 
On little-endiam ppc64 I am configuring with just `--prefix=... --enable-debug CC=xlc CXX=xlC FC=xlf --disable-mpi-fortran`

hjelmn:
Ah, ok. So xlc instead of gcc.... Hmmm.

hjelmn:
My commit might help then. Please let me know. We don't seem to have xlc on our power8 systems.

PHHargrove:
The first commit applied, but I can't see it being the problem on ppc64.
The second commit didn't apply to the 2.0.0rc3 tarball and I don't have time to look at why, nor to build a tarball from this PR.  So, no improvement to report at this time.

hjelmn:
@PHHargrove I will create a 2.0.1 PR off for you to test.

PR #1788, opened 2016-06-16
comm/split_type: allow MPI_UNDEFINED for split_type

hjelmn:
It is valid for any rank to deviate on the split_type argument if they
specify MPI_UNDEFINED. The code was incorrectly not allowing this
condition. Changed the split_type uniformity check and allow
local_size to be 0 if the local split_type is MPI_UNDEFINED.


hppritcha:
@hjelmn please add a test to OMPI tests to catch this.

hppritcha:
:+1:

hppritcha:
Do we have this problem on 1.10 too?

hjelmn:
I am not sure. Will check and add a test to ompi-tests.

PR #1800, opened 2016-06-20
Fixed common symbol error in btl/usnic.

thananon:
- This patch fixed the common symbol error when do make install. 
- Move btl_usnic_lock declaration to btl_usnic.h

@jsquyres


jsquyres:
Please update the commit message to:

Fixed common symbol in btl/usnic

* This commit fixes the accidental common symbol `btl_usnic_lock`
* It also moves the  `btl_usnic_lock` declaration to `btl_usnic.h`

PR #1803, opened 2016-06-21
Java bindings exception handling fix

### LINKED ISSUE #1698: java bindings error handling could use improving
The OMPI native method error handling code ```ompi_java_exceptionCheck``` could use some improving.  In particular, there are some cases where the MPI native code might return MPI_SUCCESS, but actually the JNI support code detected something was amiss and had thrown an exception.  Currently this case is not being handled correctly.

The two attached test cases illustrate the problem.  In the first case, the MPI native code returns an error, which ```ompi_jav_exceptionCheck``` correctly handles.  The second case, an out-of-bounds error from 
```java
java.lang.ArrayIndexOutOfBoundsException
```
is not handled correctly.

Thanks to Siegmar for reporting ( http://www.open-mpi.org/community/lists/users/2016/05/29256.php)
[Exception_1_Main.txt](https://github.com/open-mpi/ompi/files/280068/Exception_1_Main.txt)
[Exception_2_Main.txt](https://github.com/open-mpi/ompi/files/280071/Exception_2_Main.txt)

nrgraham23:
Fixed an error where if there were no MPI exceptions, a
JNI error could still exist and not get handled.

Fixes #1698 

@hppritcha @jsquyres

### DIFF
--- a/ompi/mpi/java/c/mpi_MPI.c
+++ b/ompi/mpi/java/c/mpi_MPI.c
@@ -1124,6 +1124,8 @@ void ompi_java_releasePtrArray(JNIEnv *env, jlongArray array,
 
 jboolean ompi_java_exceptionCheck(JNIEnv *env, int rc)
 {
+    jboolean jni_exception;
+
     if (rc < 0) {
         /* handle ompi error code */
         rc = ompi_errcode_get_mpi_code (rc);
@@ -1131,16 +1133,13 @@ jboolean ompi_java_exceptionCheck(JNIEnv *env, int rc)
          * all Open MPI MPI error codes should be > 0. */
         assert (rc >= 0);
     }
+    jni_exception = (*env)->ExceptionCheck(env);
 
-    if(MPI_SUCCESS == rc)
+    if(MPI_SUCCESS == rc && JNI_FALSE == jni_exception)
     {
         return JNI_FALSE;
     }
-    else if((*env)->ExceptionCheck(env))
-    {
-        return JNI_TRUE;
-    }
-    else
+    else if(MPI_SUCCESS != rc)
     {
         int     errClass = ompi_mpi_errcode_get_class(rc);
         char    *message = ompi_mpi_errnum_get_string(rc);
@@ -1152,6 +1151,9 @@ jboolean ompi_java_exceptionCheck(JNIEnv *env, int rc)
         (*env)->Throw(env, mpiex);
         (*env)->DeleteLocalRef(env, mpiex);
         (*env)->DeleteLocalRef(env, jmessage);
+        return JNI_TRUE;
+    }
+    else if (JNI_TRUE == jni_exception) {
         return JNI_TRUE;
     }
 }


hppritcha:
@nrgraham23 could you open a PR on ompi-release for this commit?  milestone 2.0.1.  Thanks.

ompi/mpi/java/c/mpi_MPI.c
```diff
@@ -1154,6 +1153,9 @@ jboolean ompi_java_exceptionCheck(JNIEnv *env, int rc)
         (*env)->DeleteLocalRef(env, jmessage);
         return JNI_TRUE;
     }
+    else if (JNI_TRUE == jni_exception) {
+        return JNI_TRUE;
```
ggouaillardet: @nrgraham23 the result of this function is not checked in collectives. shouldn't we "cancel" the native exception and throw a MPIException here ?
my knowledge of C/Java interaction is pretty limited, so the terminology I used might be incorrect.

nrgraham23: @ggouaillardet @hppritcha @jsquyres @siegmargross 

I would say no.  JNI exceptions are directly related to problems with something the JNI has tried to do.  It seems to me like this is a separate issue from MPI because it is a result of a user trying to do something they shouldn't in Java.  In the case of the index out of bounds error that has brought this up, a java array is too short for the attempted call.

That said, I have thoroughly traced the way the exceptions are being thrown, and discovered that it is actually acting differently from the way I had originally thought.  The ompi_java_exceptionCheck method is returning a true when a JNI error occurs, however it's return is almost never checked (you can do a "git grep ompi_java_exceptionCheck" from the top ompi directory to see what I am talking about).

Instead, in almost every case, the code continues to execute in the C code, and the Java exception is not actually thrown until it returns to Java land.

I propose we change it to either throw an exception in ompi_java_exceptionCheck back to Java so the code stops executing and change the method return to void, or change all of the places where ompi_java_exceptionCheck is being called to actually check the return value.

The first option is far less code, but the second option would allow us to do memory cleanup stuff.  Any thoughts?

ggouaillardet: currently, in such a case, the cleanup is performed (e.g. ```ompi_java_releaseWritePtr``` or ```ompi_java_releaseReadPtr```) but a ```java.lang.*Exception``` is thrown instead of a ```mpi.MPIException```, am i right ?

what if there is an MPI error (for example, root rank is negative ) ? in my understanding, no cleanup is performed.

at first, could/should we
- do the cleanup first
- and then ```ompi_java_exceptionCheck``` ?

an other way to put things is, is there any reason __not__ to do any cleanup if the MPI subroutine returned an error ?

if the buffer is too short, throwing a ```mpi.MPIException``` is still half baked to me, since blocking MPI subroutines would throw a ```mpi.MPI_Exception``` but non blocking MPI subroutines would throw a ```java.lang.*Exception``` and likely after the MPI subroutine returned successfully.

bottom line, i do not feel comfortable with that kind of approach.
- having Java throw a ```java.lang.*Exception``` is a valid option to me
- having MPI Java bindings check user buffer (a la ```memchecker```) when directed by the user (this is not cheap from a performance point of view) can be seen as an interesting but __new__ feature

nrgraham23: From what I have seen so far as I have started changing some of the error code is that cleanup is happening in the case of any exception (MPI included), but objects are being returned to Java that should not be.  The JNI stuff gets cleaned up, but its possible there is some MPI related stuff that is not getting cleaned up.  Ill have to look more closely at that, but a first pass probably won't include those changes.

I think the non blocking MPI subroutines get their errors through the Request object that is created, but I could be wrong about that.  Its been a while since I have looked at that code.

I personally don't think we should be checking user buffers, but I suppose if someone wanted to add it as a feature they could.  So long as there was a switch to turn it on or off.  And that is probably something that would need to be discussed with more people.

PR #1805, opened 2016-06-21
osc/rdma: fix typo

hjelmn:
Need to increment the total size after checking the local offset not
before. This typo causes large allocations with MPI_Win_allocate() to
fail.


lanl-ompi:
Test FAILed.

Issue #1806, opened 2016-06-21
Single threaded latency with latest request changes


nysal:
I was seeing a slowdown in single threaded IMB pingpong latency with the new request code in v2.x branch. ompi_request_wait_completion seems to use the multithreaded version even though I built with --disable-mpi-thread-multiple. The code seems to be guarded with OPAL_ENABLE_MULTI_THREADS, which doesn't look right. There are a few other locations too that seem to use the same check. On discussing this at the weekly meeting, Nathan suggested we fix this for 2.0.

PR #1807, opened 2016-06-21
ompi/request: fix performance regression

hjelmn:
This commit fixes a performance regression introduced by the request
rework. We were always using the multi-thread path because
OPAL_ENABLE_MULTI_THREADS is either not defined or always defined to 1
depending on the Open MPI version. To fix this I removed the
conditional and added a conditional on opal_using_threads(). This path
will be optimized out in 2.0.0 in a non-thread-multiple build as
opal_using_threads is #defined to false in that case.

Fixes open-mpi/ompi#1806

### DIFF
--- a/ompi/request/request.h
+++ b/ompi/request/request.h
@@ -375,27 +375,25 @@ static inline int ompi_request_free(ompi_request_t** request)
  * Wait a particular request for completion
  */
 
-#if OPAL_ENABLE_MULTI_THREADS
 static inline void ompi_request_wait_completion(ompi_request_t *req)
 {
-    ompi_wait_sync_t sync;
-    WAIT_SYNC_INIT(&sync, 1);
+    if (opal_using_threads ()) {
+        ompi_wait_sync_t sync;
+        WAIT_SYNC_INIT(&sync, 1);
 
-    if(OPAL_ATOMIC_CMPSET_PTR(&req->req_complete, REQUEST_PENDING, &sync)) {
-        SYNC_WAIT(&sync);
-    }
+        if(OPAL_ATOMIC_CMPSET_PTR(&req->req_complete, REQUEST_PENDING, &sync)) {
+            SYNC_WAIT(&sync);
+        }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
@nysal Confirmed the latency is back down to what is expected with this change.

thananon:
There's still macro left in opal/thread/wait_sync.h that could affect the performance. I can fix this.

`#if OPAL_ENABLE_MULTI_THREADS

#define OPAL_ATOMIC_ADD_32(a,b)         opal_atomic_add_32(a,b)
#define OPAL_ATOMIC_SWP_PTR(a,b)        opal_atomic_swap_ptr(a,b)
#define SYNC_WAIT(sync)                 sync_wait_mt(sync)
#define PTHREAD_COND_INIT(a,b)          pthread_cond_init(a,b)
#define PTHREAD_MUTEX_INIT(a,b)         pthread_mutex_init(a,b)

#define WAIT_SYNC_RELEASE(sync)                       \
    do {                                              \
       pthread_cond_destroy(&(sync)->condition);      \
       pthread_mutex_destroy(&(sync)->lock);          \
    } while(0)

#define WAIT_SYNC_SIGNAL(sync)                        \
    do {                                              \
        pthread_mutex_lock(&(sync->lock));            \
        pthread_cond_signal(&sync->condition);        \
        pthread_mutex_unlock(&(sync->lock));          \
    } while(0)

#else

#define OPAL_ATOMIC_ADD_32(a,b)         (*(a) += (b))
#define OPAL_ATOMIC_SWP_PTR(a,b)        *(a) = (b)
#define PTHREAD_COND_INIT(a,b)
#define PTHREAD_MUTEX_INIT(a,b)
#define SYNC_WAIT(sync)                 sync_wait_st(sync)
#define WAIT_SYNC_RELEASE(sync)
#define WAIT_SYNC_SIGNAL(sync)

#endif /* OPAL_ENABLE_MULTI_THREADS */`

hjelmn:
Yup. Have that fixed in my local branch. Plan to push later today after testing.

PR #1812, opened 2016-06-23
Fix Java Coverity issue

nrgraham23:
Fixing a possible error that Coverity pointed out in
ompi_java_exceptionCheck.

Signed-off-by: Nathaniel Graham <nrgraham23@gmail.com>


jsquyres:
Once CI finishes, 👍

PR #1816, opened 2016-06-24
opal/sync: fix race condition

### LINKED ISSUE #1813: Overlap test sometimes failing in Mellanox Jenkins with Yalla

hjelmn:
This commit fixes a race condition discovered by @artpol84. The race
happens when a signalling thread decrements the sync count to 0 then
goes to sleep. If the waiting thread runs and detects the count == 0
before going to sleep on the condition variable it will destroy the
condition variable while the signalling thread is potentially still
processing the completion. The fix is to add a non-atomic member to
the sync structure that indicates another process is handling
completion. Since the member will only be set to false by the
initiating thread and the completing thread the variable does not need
to be protected. When destoying a condition variable the waiting
thread needs to wait until the singalling thread is finished.

Thanks to @artpol84 for tracking this down.

Fixes #1813

### DIFF
--- a/opal/threads/wait_sync.h
+++ b/opal/threads/wait_sync.h
@@ -5,6 +5,7 @@
  *                         reserved.
  * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2016      Mellanox Technologies. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
--- a/opal/threads/wait_sync.h
+++ b/opal/threads/wait_sync.h
@@ -24,6 +24,7 @@ typedef struct ompi_wait_sync_t {
     pthread_mutex_t lock;
     struct ompi_wait_sync_t *next;
     struct ompi_wait_sync_t *prev;
+    volatile bool signaling;
 } ompi_wait_sync_t;
 
 #define REQUEST_PENDING        (void*)0L
@@ -31,10 +32,21 @@ typedef struct ompi_wait_sync_t {
 
 #define SYNC_WAIT(sync)                 (opal_using_threads() ? sync_wait_mt (sync) : sync_wait_st (sync))
 
+/* The loop in release handles a race condition between the signaling
+ * thread and the destruction of the condition variable. The signaling
+ * member will be set to false after the final signaling thread has
+ * finished opertating on the sync object. This is done to avoid
+ * extra atomics in the singalling function and keep it as fast
+ * as possible. Note that the race window is small so spinning here
+ * is more optimal than sleeping since this macro is called in
+ * the critical path. */
 #define WAIT_SYNC_RELEASE(sync)                       \
     if (opal_using_threads()) {                       \
-       pthread_cond_destroy(&(sync)->condition);      \
-       pthread_mutex_destroy(&(sync)->lock);          \
+        while ((sync)->signaling) {                  \
+            continue;                                 \
+        }                                             \
+        pthread_cond_destroy(&(sync)->condition);     \
+        pthread_mutex_destroy(&(sync)->lock);         \
     }
 
 #define WAIT_SYNC_SIGNAL(sync)                        \
@@ -42,6 +54,7 @@ typedef struct ompi_wait_sync_t {
         pthread_mutex_lock(&(sync->lock));            \
         pthread_cond_signal(&sync->condition);        \
         pthread_mutex_unlock(&(sync->lock));          \
+        sync->signaling = false;                     \
     }
 
 OPAL_DECLSPEC int sync_wait_mt(ompi_wait_sync_t *sync);
@@ -61,6 +74,7 @@ static inline int sync_wait_st (ompi_wait_sync_t *sync)
         (sync)->next = NULL;                                    \
         (sync)->prev = NULL;                                    \
         (sync)->status = 0;                                     \
+        (sync)->signaling = true;                               \
         if (opal_using_threads()) {                             \
             pthread_cond_init (&(sync)->condition, NULL);       \
             pthread_mutex_init (&(sync)->lock, NULL);           \
@@ -81,8 +95,9 @@ static inline void wait_sync_update(ompi_wait_sync_t *sync, int updates, int sta
         }
     } else {
         /* this is an error path so just use the atomic */
-        opal_atomic_swap_32 (&sync->count, 0);
         sync->status = OPAL_ERROR;
+        opal_atomic_wmb ();
+        opal_atomic_swap_32 (&sync->count, 0);
     }
     WAIT_SYNC_SIGNAL(sync);
 }


hjelmn:
@artpol84 Getting this running through Jenkins. Will try to get it retested multiple times to see if it fails. In theory it shouldn't.

jladd-mlnx:
bot:retest

hjelmn:
:bot:retest:

hjelmn:
:bot:retest:

jladd-mlnx:
Retest after adding the original Jenkins command line with ob1 (allow UCX to load and unload.) 

bot:retest

hjelmn:
I can make this even faster as the lock/unlock is not really needed for the condition signal here. Will leave that for later though.

hjelmn:
Looking good. Sending through again.

:bot:retest:

artpol84:
@hjelmn I guess that race condition is still possible here
1. `(sync)->signalling` is initialized to `false` and this is correct because you may wait late (after all `wait_sync_update` was called)
2, Assume instruction reordering in `wait_sync_update` (correct be if I'm wrong, but it's possible):

| count | PML progress  | MPI_Wait thread |
| ------------- | ------------- | ------------- |
| 1 | OPAL_THREAD_ADD32(&sync->count, -1)  |   |
|0 |  | if( sunc->count > 0 ) - false => exit the loop |
| 0 | | running till the end and free'ing `sync` |
|0| WAIT_SYNC_SIGNAL(sync); - error | |

Same error here. I also assume there will be problems with several independently running PML progress threads. 

So it seems that you need a memory fence here:

```C
static inline void wait_sync_update(ompi_wait_sync_t *sync, int updates, int status)
  {
     sync->signalling = true;

     mem_fence(); // <---------   

     if( OPAL_LIKELY(OPAL_SUCCESS == status) ) {
          if( 0 != (OPAL_THREAD_ADD32(&sync->count, -updates)) ) {
              return;
         }
     } else {
         /* this is an error path so just use the atomic */
         opal_atomic_swap_32 (&sync->count, 0);
         sync->status = OPAL_ERROR;
     }
     WAIT_SYNC_SIGNAL(sync);
 }
```

This race will be much more rare compared to original one, and harder to track.

hjelmn:
Instruction reordering will not happen. pthread calls are effectively a memory barrier. We could add an isync just to be safe if you want think it is warranted. We have isync for both ppc and arm.

hjelmn:
Oh, wait. *before* the atomic. Sorry, didn't catch that part. You are absolutely correct.

hjelmn:
An atomic wmb would be fine there. An isync as I suggestedmight be too weak.

artpol84:
Setting of signal and changing count doesn't have pthreads, they are coming
lately

суббота, 25 июня 2016 г. пользователь Nathan Hjelm написал:

> Instruction reordering will not happen. pthread calls are effectively a
> memory barrier. We could add an isync just to be safe if you want think it
> is warranted. We have isync for both ppc and arm.
>
> —
> <https://github.com/open-mpi/ompi/pull/1816#issuecomment-228507294>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHL5Pkr67OoyDmNxE209ANgTPm1ZPMfpks5qPKI0gaJpZM4I-HOt>
> .
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

hjelmn:
@artpol84 Yup. Just caught that after hitting comment :)

hjelmn:
Will add the barrier now.

hjelmn:
And correct to the more common signaling spelling :D

hjelmn:
@artpol84 Thanks for catching that. Should be fixed now.

hjelmn:
Won't commit this until I get a +1 from @artpol84.

Looks like we had a finalize hang. Looks unrelated. @artpol84 Any clue as to what happened?

hjelmn:
rdmacm failure here too. I will merge the rdmacm fix I have open and see if it fixes it.

hjelmn:
:bot:retest:

hjelmn:
@artpol84 Feel free to open a PR vs my branch and add the mellanox copyright. I prefer the commit adding a copyright line be from a member the copyrighting organization.

opal/threads/wait_sync.h
```diff
@@ -75,6 +89,9 @@ static inline int sync_wait_st (ompi_wait_sync_t *sync)
  */
 static inline void wait_sync_update(ompi_wait_sync_t *sync, int updates, int status)
 {
+    sync->signaling = true;
+    /* ensure the signalling value is commited before updating the count */
+    opal_atomic_wmb ();
```
artpol84: @hjelmn I may be wrong but it seems that wmb addresses CPU-level reordering possibility.
Do we address compiler reordering?

artpol84: I mean adding something like `asm volatile("" ::: "memory");` may be needed

artpol84: Ok, this is already done in `wmb()`

opal/threads/wait_sync.h
```diff
@@ -75,6 +89,9 @@ static inline int sync_wait_st (ompi_wait_sync_t *sync)
...
     if( OPAL_LIKELY(OPAL_SUCCESS == status) ) {
         if( 0 != (OPAL_THREAD_ADD32(&sync->count, -updates)) ) {
```
artpol84: @hjelmn @jladd-mlnx I have a question here.
Consider we have 2 concurrently running wait_sync_updates. Opal thread add is atomic in the sense that intermediate results of the op will not intermix.
**however** we can get both of them receiving 0 if they called this primitive together.
This means another possible race condition:
* Both threads will go to release path
* First thread will release signaling flag
* waiting thread will release sync
* second thread running wait_sync_updates will hit the same bug.

artpol84: By the "same bug" I mean signaling on uninitialized condition.

artpol84: I guess I'm not correct here.
Only one (the last) thread will get 0.
So no race here.

bosilca: Nice catch. The scope of this race condition is restricted to cases where
the synchronization primitive is used in a wait or waitall but not waitany
or waitsome (where the sync primitive is atomically removed from the
requests before release).

As the only thing we need is to guarantee that the signaling thread is out
of the signaling sequence, so we can start with the signaling field set to
true. This will save the memory barrier in the wait_sync_update, and will
only require the change in WAIT_SYNC_SIGNAL.

Also the error path is now broken, the following patch is necessary.
```
diff --git a/opal/threads/wait_sync.h b/opal/threads/wait_sync.h
index b29f01c..d2b5003 100644
--- a/opal/threads/wait_sync.h
+++ b/opal/threads/wait_sync.h
@@ -80,9 +80,10 @@ static inline void wait_sync_update(ompi_wait_sync_t
*sync, int updates, int sta
             return;
         }
     } else {
-        /* this is an error path so just use the atomic */
-        opal_atomic_swap_32 (&sync->count, 0);
+        /* We need the status to be visible when the count become 0 */
         sync->status = OPAL_ERROR;
+        opal_atomic_wmb();
+        opal_atomic_swap_32 (&sync->count, 0);
     }
     WAIT_SYNC_SIGNAL(sync);
 }
```

artpol84: @bosilca I'll add this as a separate commit in my PR https://github.com/open-mpi/ompi/pull/1815 on your behalf.

artpol84:
Looks like waitany and waitsome are broken. I filed the PR to @hjelmn branch:
https://github.com/artpol84/ompi/commit/49b1c0ffadde8341cbf1a0186325bd3614f4634a

Can be easily reproduced with following modifications of overlap test:
[overlap_waitany.txt](https://github.com/open-mpi/ompi/files/334573/overlap_waitany.txt)
[overlap_waitsome.txt](https://github.com/open-mpi/ompi/files/334574/overlap_waitsome.txt)

@hjelmn @bosilca please, check.

artpol84:
@jladd-mlnx @miked-mellanox I think we (Mellanox) will need to include this tests to our jenkins/MTT suite.
They exercise both corner cases: late completion and early completion of all requests.
We can also add versions without delay to exercise other possible paths.
Since those are race conditions - probably multiple executions is needed.

jsquyres:
@artpol84 It would be good if these could be standalone tests added to the ibm test suite in ompi-tests; then they'll be run in MTT every night by a bunch of organizations.

artpol84:
@jsquyres, sure, we can do that as well.

hjelmn:
@bosilca, @artpol84 Are we all ok with this change the way it is? @artpol84 You can close your PR. Due to a rebase it would no longer merge so I cherry-picked the commit.

artpol84:
@hjelmn, what do you mean? I don't see the wait some waitany fix in this Pr.

bosilca:
I think this patch fixes the original issue, and as such it is ready to go. The performance issues discovered meanwhile should be fixed by a subsequent ticket, once we have the correct fix.

artpol84:
There was a) a bug and b) performance issue.
I PRed a bug fix for wait some and waitany here.
And perf is pending on another request

вторник, 28 июня 2016 г. пользователь bosilca написал:

> I think this patch fixes the original issue, and as such it is ready to
> go. The performance issues discovered meanwhile should be fixed by a
> subsequent ticket, once we have the correct fix.
>
> —
> <https://github.com/open-mpi/ompi/pull/1816#issuecomment-228907225>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHL5PoIxz2ev2r02WH_ZDSnOaj2igRMEks5qQF3ogaJpZM4I-HOt>
> .
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
https://github.com/hjelmn/ompi/pull/9/commits/42e6251a25e4b9064f72352a150af1105098899f
This is a bug fix

artpol84:
Without this bugfix the following tests (already mentioned above) are hanging 100%
[overlap_waitany.txt](https://github.com/open-mpi/ompi/files/336056/overlap_waitany.txt)
[overlap_waitsome.txt](https://github.com/open-mpi/ompi/files/336057/overlap_waitsome.txt)

I think PR must fix without new hang introduction.

bosilca:
42e6251 introduces its own issues (I commented on it on the other issue about the possible race condition). A possible fix is to generate a sync_update (which will take care of the signaling field) in the  wait* operation if there is nothing to wait for.

artpol84:
@bosilca I think that your comment https://github.com/open-mpi/ompi/pull/1820#issuecomment-228905750 is reasonable for MPI_Waitany only. So in my understanding it is related to this PR because I want this commit https://github.com/open-mpi/ompi/pull/1820/commits/42e6251a25e4b9064f72352a150af1105098899f to go along with this PR.
I will fix that now. For MPI_Waitsome I believe we are fine.

artpol84:
@hjelmn I rebased my PR https://github.com/hjelmn/ompi/pull/9 and added the fix for the problem that @bosilca pointed, please don't rebase untill we agreed here.

@bosilca please have a look.

artpol84:
@bosilca btw I solved the problem with Waitany with sync_sets/unsets instead of sync_update. I can do the sync_update but with sync_sets/unsets we avoiding one atomic. What do you think?

artpol84:
ohh, and according to the backtrace I had when I was debugging:
https://github.com/open-mpi/ompi-release/pull/1240#issuecomment-227990044

sync_wait will then try to call condition which has internal lock (see the backtrace of the second thread).
So I guess we wont to avoid sync_wait call when possible.

artpol84:
@bosilca @hjelmn I'll be on today's OMPI call available for discussion.

bosilca:
I wont be able to join the call today.

artpol84:
I think I've got all your points now. Will update PRs later today. Thank
you!

вторник, 28 июня 2016 г. пользователь bosilca написал:

> I wont be able to join the call today.
>
> —
> <https://github.com/open-mpi/ompi/pull/1816#issuecomment-229005736>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHL5PqkZyU1bvP8o1DYuNz5cejjaA_t7ks5qQPCHgaJpZM4I-HOt>
> .
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
Again updated the PR https://github.com/hjelmn/ompi/pull/9 with slightly improved MPI_Waitsome performance according @bosilca [suggestion](https://github.com/open-mpi/ompi/pull/1820#issuecomment-228905750)

jladd-mlnx:
@hjelmn @artpol84 @bosilca Do we have any idea why it's consistently hanging in init when using RDMACM? This is new behavior.

artpol84:
I don't think it's related. The proble is in MPI Init. Unlikely request

вторник, 28 июня 2016 г. пользователь Joshua Ladd написал:

> @hjelmn <https://github.com/hjelmn> @artpol84
> <https://github.com/artpol84> @bosilca <https://github.com/bosilca> Do we
> have any idea why it's consistently hanging in init when using RDMACM? This
> is new behavior.
>
> —
> <https://github.com/open-mpi/ompi/pull/1816#issuecomment-229039372>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHL5PiE6ZZUNF2sBKtWmNBxl5sU2cfgqks5qQRi7gaJpZM4I-HOt>
> .
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

hjelmn:
Well, it shouldn't be crashing. It should be exiting. This is the problem:

```
06:44:23 --------------------------------------------------------------------------
06:44:23 No OpenFabrics connection schemes reported that they were able to be
06:44:23 used on a specific port.  As such, the openib BTL (OpenFabrics
06:44:23 support) will be disabled for this port.
06:44:23 
06:44:23   Local host:           jenkins01
06:44:23   Local device:         mlx5_0
06:44:23   Local port:           1
06:44:23   CPCs attempted:       rdmacm
06:44:23 --------------------------------------------------------------------------
```

rdmacm is working for me. Not sure why it is failing to load on Mellanox Jenkins.

hjelmn:
Unless that is the other port.

hjelmn:
BTW, if you need udcm to work with a router it wouldn't take much work. No need to use rdmacm for that :).

artpol84:
I will check tomorrow.

artpol84:
@hjelmn could you accept my PR?
@bosilca approved it.

hjelmn:
@artpol84 Will merge after jenkins finishes.

jsquyres:
@hjelmn @artpol84 We're less than 45 mins from the call -- do you want to wait for merging whichever PRs are going to get merged until we can do final coordination on the phone?

artpol84:
Sure.
I was talking about my PR to @hjelmn branch. Let's talk.

PR #1817, opened 2016-06-24
OSHMEM: Removing erroneous initialization check

shamisp:
Since the introduction of the on-demand proc allocation
the check become erroneous and irrelevant.
Moreover, it completely breaks OpenSHMEM support in OMPI.

### DIFF
--- a/oshmem/proc/proc.c
+++ b/oshmem/proc/proc.c
@@ -4,6 +4,7 @@
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


shamisp:
@yosefe @hjelmn please take a look.

jsquyres:
@shamisp Are you saying this is a blocker for v2.0.0?  If so, I'm curious: why wasn't this discovered before now?

shamisp:
@jsquyres The bug can be exposed with any SHMEM code running on NP > 2. So I suspect nobody is really testing the code with more than two processes.

jsquyres:
Is there going to be a v2.0.0 PR for this?  If so, please file it ASAP.

hppritcha:
It does appear the same problem exists on 2x.

shamisp:
I will prepare the patch.

PR #1820, opened 2016-06-27
MPI_Waitsome performance improvement

artpol84:
This PR makes sense only if
* sub PR https://github.com/hjelmn/ompi/pull/9 will be accepted
* PR https://github.com/open-mpi/ompi/pull/1816 will be merged

will need to rebase afterwards.

@bosilca @hjelmn please check.

I'm addressing the following drawback:
https://github.com/hjelmn/ompi/blob/request_perfm_regression/ompi/request/req_wait.c#L412:L417

Still checking but waitsome test seems to pass.

### DIFF
--- a/ompi/request/req_wait.c
+++ b/ompi/request/req_wait.c
@@ -15,6 +15,7 @@
  * Copyright (c) 2012      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2016      Mellanox Technologies. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -87,6 +88,7 @@ int ompi_request_default_wait_any(size_t count,
     int rc = OMPI_SUCCESS;
     ompi_request_t *request=NULL;
     ompi_wait_sync_t sync;
+    int sync_sets = 0, sync_unsets = 0;
 
     WAIT_SYNC_INIT(&sync, 1);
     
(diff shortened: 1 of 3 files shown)


artpol84:
running stress-test againts this now. Will update later.

bosilca:
I have few comments about this patch.

Introducing WAIT_SYNC_SIGNALLED have the drawback to bring back the original issue. Now the sync primitive might have few requests associated with, but will be released by the creator thread even if one of the progress thread is actually trying to activate the synchronization primitive (because now signaling doesn't protect the path where a thread is still having a hand on the synchronization primitive). I think a simpler solution will be to simply call the sync_update to remove the initial value and if necessary set the signalling field to false).

sync_sets isn't equal to count - num_requests_null_inactive - num_requests_done in the first part of wait_some ? and then to count - num_requests_null_inactive - num_requests_done on the second loop ?

The patch improving MPI_Waitsome saves one atomic operation for all requests that were originally completed in exchange for increasing the size of the request structure. Why not using the user provided array (indices) instead of altering the request structure ? I dont think there is anything in the standard that prevents us from using the array as temporary buffer, as long as upon return it contains the expected values, and we know it must have the right size (or the application is incorrect).

artpol84:
@bosilca 
1. Your concerns are only applicable to MPI_Waitany, I believe. For MPI_Waitsome we call SIGNALLED only if we are 100% sure that nobody is waiting on the sync.
2. introducing a flag helps to avoid index search (if you will put request indexes compact as they appear to be finished OR index rearrangement at the end of the function if you will use original indexes).

artpol84:
@bosilca Ok, let me think more about that and try to consider your suggestions.

artpol84:
Closing this in favor of PR https://github.com/open-mpi/ompi/pull/1821

Issue #1831, opened 2016-06-29
Significant degradation in message rates observed on Master. 


jladd-mlnx:
Opening this issue for tracking purposes. Measured with Master nightly build against 1.10.3. Possible fix on master. 

@hjelmn or @bosilca please comment. 

```
PML - Yalla
OMPI – 1.10.3
$mpirun -np 2 --map-by node --bind-to core -mca pml yalla -x MXM_RDMA_PORTS=mlx5_0:1  -mca btl_openib_if_include mlx5_0:1 /hpc/local/benchmarks/hpc-stack-gcc/install/ompi-v1.10/tests/osu-micro-benchmarks-5.2/osu_mbw_mr
# OSU MPI Multiple Bandwidth / Message Rate Test v5.2
# [ pairs: 1 ] [ window size: 64 ]
# Size                  MB/s        Messages/s
1                       4.01        4005006.11
2                       8.24        4121056.15
4                      16.39        4097311.09
8                      32.45        4055766.73
16                     64.16        4010025.24
32                    127.13        3972687.66
64                    237.04        3703703.70
128                   455.11        3555555.62
256                   860.96        3363110.99
512                  1592.23        3109815.42
1024                 2811.50        2745602.68
2048                 4972.38        2427921.16
4096                 5430.79        1325875.29
8192                 5933.54         724309.64
16384                6155.42         375697.10
32768                6328.16         193120.10
65536                6398.15          97627.95
131072               6433.23          49081.64
262144               5161.27          19688.67
524288               5731.10          10931.20
1048576              6046.06           5765.97
2097152              6215.12           2963.60
4194304              6306.30           1503.54


---------------------
PML – Yalla
OMPI – Master
$mpirun -np 2 --map-by node --bind-to core -mca pml yalla -x MXM_RDMA_PORTS=mlx5_0:1  -mca btl_openib_if_include mlx5_0:1 /hpc/local/benchmarks/hpc-stack-gcc/install/ompi-master/tests/osu-micro-benchmarks-5.2/osu_mbw_mr
# OSU MPI Multiple Bandwidth / Message Rate Test v5.2
# [ pairs: 1 ] [ window size: 64 ]
# Size                  MB/s        Messages/s
1                       1.89        1887305.40
2                       3.80        1898890.08
4                       7.56        1889678.24
8                      15.31        1914346.13
16                     30.41        1900517.95
32                     60.30        1884510.12
64                    119.99        1874796.93
128                   227.47        1777098.03
[...]

PR #1832, opened 2016-06-30
bml/r2: be more restrictive on rdma endpoints

hjelmn:
This commit makes bml/r2 more restrictive on which endpoints end up in the rdma
endpoint list. Before this commit an endpoint was added if it supported either
put or get. This was done to ensure that endpoints are available for RMA.
Thought it is possible to support put or get endpoints we only currently
support endpoints that have put, get, and amos. bml/r2 now reflects this
support.

### DIFF
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -267,8 +267,9 @@ static int mca_bml_r2_endpoint_add_btl (struct ompi_proc_t *proc, mca_bml_base_e
         }
     }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
This might have an impact on #1831. @jladd-mlnx Let me know if that is the case.

hjelmn:
The intent is to support RMA over btls when only some true RMA operations are supported by the btl. As this support is not ready this PR is fine for now. If I finish the intended RMA support we will need to figure out how to prevent ob1 from using a btl it was not supposed to use.

Issue #1822, opened 2016-06-28
Open MPI parsing of proc mounts needs help


hppritcha:
Open MPI tries to parse proc/mounts (not the hwloc part), which dies if entries exceed the size of the buffer used to read a line from the file.  The technique in hwloc using `setmntent` and `getmntent` to parse /proc/mounts is probably the way to go.

This problem shows up regularly on Cray XC's running CLE 6.

hjelmn:
I think I fixed this on master with the rcache/mpool rewrite. Will need to make a 2.0.x specific fix.

ggouaillardet:
@hjelmn 
i made #1846 for master
a similar fix should be done for v2.x

@hppritcha 
v1.10 hwloc manually parses /proc/mounts, is a fix required for this branch too ?

jsquyres:
@hppritcha Per @ggouaillardet's question, do you only care about v2.x for this fix, or do you need a fix for v1.10.x as well?  If you only need v2.x, then is https://github.com/open-mpi/ompi-release/pull/1298 sufficient?

hppritcha:
I think the problematic points in Open MPI recent releases have all be addressed so this can be closed.

AlexanderKurtz:
Since I was just bitten by this, here is an example error message when running OpenMPI under docker. This hopefully improves the odds of $search_engine finding this bug report!

```
Unexpected end of /proc/mounts line `overlay / overlay rw,relatime,lowerdir=/var/lib/docker/overlay2/l/P6LJKWCEKPXJWROJMJK3TEQEQ6:/var/lib/docker/overlay2/l/AUWTYHXASAF36QV5RKRCMUSFKR:/var/lib/docker/overlay2/l/R6UEDWX7LPYJAAG2FU3ATF2V3C:/var/lib/docker/overlay2/l/VOL35J5R5MJ3IT24W6CNYJTL4O:/var/lib/docker/overlay2/l/Z2MK7LR3EUI55CT2NUQPOTE2XI:/var/lib/docker/overlay2/l/KV24PDZFQBZCG5RDHRDAOCKWYH:/var/lib/docker/overlay2/l/A5XWI2JFDDU35WYIE4IJKKHHLD:/var/lib/docker/overlay2/l/IEQ77LFHHVS3JHGRXBTRTIC4TU:/var/lib/docker/overlay2/l/4HJEKXYLDUYTM'
```

jsquyres:
[...]

PR #1846, opened 2016-07-04
mpool/hugepage: set mntent API instead of manually parsing /proc/mounts

ggouaillardet:
Refs open-mpi/ompi#1822

### DIFF
--- a/opal/mca/mpool/hugepage/mpool_hugepage_component.c
+++ b/opal/mca/mpool/hugepage/mpool_hugepage_component.c
@@ -55,6 +55,9 @@
 #ifdef HAVE_SYS_MMAN_H
 #include <sys/mman.h>
 #endif
+#ifdef HAVE_MNTENT_H
+#include <mntent.h>
+#endif
 
 #include <fcntl.h>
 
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@hjelmn is hugetlbfs upported by Open MPI on OS X ?
in other words, is it ok to disqualify this module if mntent is not available ?

hppritcha:
I plan to open a PR for the Cray large page support this week.

The manual parsing doesn't work reliably on CLE 6

Am Montag, 4. Juli 2016 schrieb Gilles Gouaillardet :

hjelmn:
@ggouaillardet OS X has hugepage support using an mmap flag. Problem is that I have not been able to reliably allocate huge pages more than 5 min after a boot. It looks like Apple only allows huge page allocation in a pool that is very quickly exhausted.

hppritcha:
@hjelmn i think this needs to be checked on a Cray  if in fact you'd like the huge page mpool to work on that system.

hppritcha:
This fix is needed for Cray KNL systems (or more precisely CLE 6) otherwise we see fairly frequent aborts in MPI_Init (unless one is using ofi mtl).  So putting this at milestone 2.0.1.

hppritcha:
@ggouaillardet I'm going to merge this and test the cherry pick on 2.x.  Unfortunately native launch is aprun on KNL nodes is all messed up on master.

PR #1850, opened 2016-07-06
Patchs on mtl portals4

PDeveze:
These patchs concern the mtl/portals4 module.

PR #1853, opened 2016-07-07
Patchs on osc portals4

PDeveze:
I have here a few patchs for osc/portals4


PDeveze:
Sorry, in that pull request are also the patchs for mtl/portals4. But I think this will not be a problem to integrate the patchs.

jsquyres:
@regrant @tkordenbrock Please review.

jsquyres:
@PDeveze Are these related to the commits you have on #1850?

If so, and if you didn't already know: you can continue to add / modify commits on a PR at any time before it is merged (e.g., you could have added these new osc/portals4 commits on #1850, if you wanted to).

PDeveze:
Thanks for your tip !
The reason why I created a new branch is to separate the patchs concerning mtl/portals4 and osc/portals4. The next days I will also have patchs and PR on coll/portals4 and btl/portals4. And I will take attention not to mix the different patchs on these different branches.

PR #1860, opened 2016-07-12
configury: fix configure --disable-mpi-cxx when no C++ compiler is av…

ggouaillardet:
…ailable/working

note:
if GNU C compiler is used but no GNU C++ compiler is available, then
CPPFLAGS='-x c'
must be added to the configure command line

### DIFF
--- a/config/opal_config_pthreads.m4
+++ b/config/opal_config_pthreads.m4
@@ -198,7 +198,8 @@ AC_DEFUN([OPAL_INTL_POSIX_THREADS_PLAIN_CXX], [
 #
 # C++ compiler
 #
-if test "$opal_pthread_cxx_success" = "0"; then
+if test "$opal_pthread_cxx_success" = "0" && \
+   test "$WANT_MPI_CXX_SUPPORT" -eq; then
   AC_MSG_CHECKING([if C++ compiler and POSIX threads work as is])
 
   AC_LANG_PUSH(C++)
@@ -251,9 +252,10 @@ AC_PROVIDE_IFELSE([AC_PROG_CC],
                   [OPAL_INTL_POSIX_THREADS_PLAIN_C],
                   [opal_pthread_c_success=1])
 
-AC_PROVIDE_IFELSE([AC_PROG_CXX],
-                  [OPAL_INTL_POSIX_THREADS_PLAIN_CXX],
-                  [opal_pthread_cxx_success=1])
+AS_IF([test "$WANT_MPI_CXX_SUPPORT" = 1],
+      [AC_PROVIDE_IFELSE([AC_PROG_CXX],
+                         [OPAL_INTL_POSIX_THREADS_PLAIN_CXX],
+                         [opal_pthread_cxx_success=1])])
 
 AC_PROVIDE_IFELSE([AC_PROG_FC],
                   [OPAL_INTL_POSIX_THREADS_PLAIN_FC],
@@ -379,9 +381,10 @@ AC_PROVIDE_IFELSE([AC_PROG_CC],
                   [OPAL_INTL_POSIX_THREADS_SPECIAL_FLAGS_C],
                   [opal_pthread_c_success=1])
 
-AC_PROVIDE_IFELSE([AC_PROG_CXX],
-                  [OPAL_INTL_POSIX_THREADS_SPECIAL_FLAGS_CXX],
-                  [opal_pthread_cxx_success=1])
+AS_IF([test "$WANT_MPI_CXX_SUPPORT" = 1],
+      [AC_PROVIDE_IFELSE([AC_PROG_CXX],
+                         [OPAL_INTL_POSIX_THREADS_SPECIAL_FLAGS_CXX],
+                         [opal_pthread_cxx_success=1])])
 
 AC_PROVIDE_IFELSE([AC_PROG_FC],
                   [OPAL_INTL_POSIX_THREADS_SPECIAL_FLAGS_FC],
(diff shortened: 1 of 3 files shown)


ggouaillardet:
@jsquyres could you please have a look at it ?
if you build Open MPI on OSX with FreePGI, two issues remain
- libtool pass -flat_namespace to the linker, it works fine with gcc, but pgcc expects -Wl,-flat_namespace
I made a simple proof of concept (patch ltmain.sh, smilitary to what was done for NAG)
- atomic prototypes are missing ```static inline``` which ends up with many symbols defined in multiple places, and link fails. I tried to understand the logic, but I could not get it all. some macro explicitly reference GCC, and that is wrong here. @hjelmn can you please have a look at it too ?

config/opal_config_pthreads.m4
```diff
@@ -198,7 +198,8 @@ AC_DEFUN([OPAL_INTL_POSIX_THREADS_PLAIN_CXX], [
 #
 # C++ compiler
 #
-if test "$opal_pthread_cxx_success" = "0"; then
+if test "$opal_pthread_cxx_success" = "0" && \
+   test "$WANT_MPI_CXX_SUPPORT" -eq; then
```
jsquyres: Missing RHS 1.

config/opal_config_pthreads.m4
```diff
@@ -251,9 +252,10 @@ AC_PROVIDE_IFELSE([AC_PROG_CC],
                   [OPAL_INTL_POSIX_THREADS_PLAIN_C],
                   [opal_pthread_c_success=1])
 
-AC_PROVIDE_IFELSE([AC_PROG_CXX],
-                  [OPAL_INTL_POSIX_THREADS_PLAIN_CXX],
-                  [opal_pthread_cxx_success=1])
+AS_IF([test "$WANT_MPI_CXX_SUPPORT" = 1],
```
jsquyres: You use `-eq` for `$WANT_MPI_CXX_SUPPORT` above -- should probably use it here, too.  Then you don't need to use the quotes for `$WANT_MPI_CXX_SUPPORT, either (because it's guaranteed to be a number).

config/opal_config_pthreads.m4
```diff
@@ -251,9 +252,10 @@ AC_PROVIDE_IFELSE([AC_PROG_CC],
...
+      [AC_PROVIDE_IFELSE([AC_PROG_CXX],
```
jsquyres: I actually can't find any documentation on `AC_PROVIDE_IFELSE` -- I see that we use it all throughout `opal_config_pthreads.m4`, but I can't quite figure out what it is *for*.  My best guess is that it is "if `AC_PROG_CXX` has been substituted earlier, then substitute $1, otherwise substitute $2."

But if that's correct, it's been used entirely wrong in this m4 file (and likely should be removed).

Specifically: I'm trying to figure out why your new test on `$WANT_MPI_CXX_SUPPORT` is necessary.  And if it is, is the `AC_PROVIDE_IFELSE` unnecessary (or incorrect)?

config/opal_config_pthreads.m4
```diff
@@ -661,6 +665,10 @@ if test "$OMPI_TRY_FORTRAN_BINDINGS" = "$OMPI_FORTRAN_NO_BINDINGS" || \
   opal_pthread_fortran_success=1
 fi
 
+if test "$WANT_MPI_CXX_SUPPORT" -ne 1; then
```
jsquyres: If you use the numeric test operators, you shouldn't need to use quotes on the variable under test (because it had better be guaranteed to be a number).

config/opal_config_pthreads.m4
ggouaillardet: on OS X with FreePGI, there is no c++ compiler, so configure falls back to g++, and pgcc does not work out of the box with g++, so pthread tests fail.
the purpose of the test is simply not to run tests that might fail and we should not care about.

makes sense ?

jsquyres: Right -- I understand what `$WANT_MPI_CXX_SUPPORT` is for, but I want to understand what `AC_PROVIDE_IFELSE` was *supposed* to be for.

ggouaillardet: my understanding is that this is here to support autotools that do not define/provide the ```AC_PROG_CXX``` macro.
As you already pointed, that might not be what we want.

jsquyres: I see `AC_PROVIDE_IFELSE` used in a few places throughout the tree, but aside from `opal_config_pthreads.m4`, all of them are 3rd-party locations (e.g., libtool, libevent, romio, etc.).

Do you want to get rid of the use of `AC_PROVIDE_IFELSE` here in this file, and instead, use the proper "if we have FOO support" OMPI shell variables?  (e.g., `$WANT_MPI_CXX_SUPPORT`)

ggouaillardet: Should the test be
Is there a working c++ compiler ?
AND
Do we want to build c++ bindings ?

Which raises the question, how do we test if there is a working c++ compiler ?
```test -n $CXX``` ?
assuming AC_PROG_CXX was already invoked

jsquyres: I think `$CXX` will be set to `no` (or empty?) if there is no valid C++ compiler.

/me goes to test...

It looks like configure sets `$CXX` to `g++` if it can't find a valid C++ compiler (!).  But then if you try to run it, for example, it will fail (e.g., if it can't be found). 😞 

The only place we have C++ in the code base is in the MPI C++ bindings (and examples), but those aren't even built by default any more.

I wonder if this whole C++ section of configure is in need of a bit of revamping / simplification...

bwbarrett:
@ggouaillardet, Can you either update or close this PR?

ggouaillardet:
moving this PR to https://github.com/open-mpi/ompi/wiki/Archived-PR

PR #1863, opened 2016-07-12
openmpi.spec: don't export FFLAGS

jsquyres:
The Open MPI configure script has long-since only paid attention to
FCFLAGS.  Indeed, it will warn if you set FFLAGS or F77FLAGS.  So
remove them from the spec file.


jsquyres:
None of the CI's check the specfile, so I'm just going to merge without waiting.

lanl-ompi:
Test FAILed.

PR #1855, opened 2016-07-08
ompi/comm: refactor communicator cid code

hjelmn:
This commit simplifies the communicator context ID generation by
removing the blocking code. The high level calls: ompi_comm_nextcid
and ompi_comm_activate remain but now call the non-blocking variants
and wait on the resulting request. This was done to remove the
parallel paths for context ID generation in preperation for further
improvements of the CID generation code.

Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

PR #1873, opened 2016-07-13
Improve MPI_Comm_split_type scalability

hjelmn:
This is built on top of the refactor work in #1855.

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -358,13 +358,7 @@ int ompi_comm_create ( ompi_communicator_t *comm, ompi_group_t *group,
     }
 
     /* Determine context id. It is identical to f_2_c_handle */
-    rc = ompi_comm_nextcid ( newcomp,  /* new communicator */
-                             comm,     /* old comm */
-                             NULL,     /* bridge comm */
-                             NULL,     /* local leader */
-                             NULL,     /* remote_leader */
-                             mode,     /* mode */
-                             -1 );     /* send first */
+    rc = ompi_comm_nextcid (newcomp, comm, NULL, NULL, NULL, false, mode);
     if ( OMPI_SUCCESS != rc ) {
         goto exit;
     }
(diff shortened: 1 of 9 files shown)


hjelmn:
@bosilca The idea behind this change is most programs are not going to either reorder or drop ranks (with MPI_UNDEFINED). We can optimize this case and leave off the global all gather operation.

hjelmn:
comm_split_type

hjelmn:
comm_split_type

rhc54:
I'm not sure why there is concern over the performance of comm_split, but @hjelmn did raise a question with me about the time lost in looking up hostnames when that operation is tagged as "optional". Looking into it, the time appears to be spent performing several hash table lookups because we don't have general rules about naming conventions.

Specifically, we could improve this algorithm by:

- requiring that any lookup request for job-level info (i.e., info that is not directly tied to a specific rank) by done by specifying a rank of ```PMIX_RANK_WILDCARD``` or ```PMIX_RANK_INVALID```. This would let us eliminate 1/3 of the time.

- requiring that any modex info use a non-PMIx namespace key - i.e., that user-provided data cannot have a key starting with "pmix". This would eliminate an additional 1/3 of the time.

I cannot see a way to eliminate more than this, but the combination should reduce the time by nearly 70%, and I don't think these rules would be particularly onerous. Is this worth pursuing?

hjelmn:
This is comm_split_type so it is a little more important to optimize it. At a minimum it is a not a great idea to do a all gather on a large comm when any rank only needs to know the result on a small subset of ranks. BTW, in order to do a fair comparison of old vs new I removed the modex receive for the old algorithm as well. On master no ompi_proc_t means not node local.

rhc54:
Okay, I'll propose adding those requirements to the PMIx specification - if nobody objects, I can update the algorithm.

rhc54:
Working its way thru the PMIx approval process: https://github.com/pmix/master/pull/114

bosilca:
@hjelmn nice catch, this patch indeed improves the case where no processes are processes are using MPI_UNDEFINED. 👍

PR #1879, opened 2016-07-16
Fix dynamic operations by ensuring that we only fire the debugger rel…

rhc54:
…ease if the debugger is attached, and that the OPAL pmix key for directing events to non-default handlers matches the PMIx spelling


ggouaillardet:
@rhc54 It seems file permissions were accidentally changed from 644 to 755.
Thanks for the fix !

rhc54:
stupid smb file system - thanks for pointing it out. Fixed!

PR #1881, opened 2016-07-18
pmix/pmix2x: add missing increment in loop

hjelmn:
This commit fixes a bug in the pmix2x client code where a loop
variable is not correctly incremented. This was leading to hangs and
crashes when creating intercommunicators.


hjelmn:
@rhc54 This is what was causing my CID changes to hang and crash. Pretty obvious bug.

rhc54:
👍

hjelmn:
@rhc54 Found another error. Will update the PR shortly.

rhc54:
thanks for finding these!

👍

PR #1897, opened 2016-07-21
Cleanup the new naming requirements to ensure that info is correctly retrieved

rhc54:


artpol84:
Yoohoo!!!
@rhc54  green jenkins

artpol84:
Let's push it and retrigger other PRs

hppritcha:
folks who use native launch methods might want to check this PR out.  The use of the wildcard for storing wildcard id rather than ```ORTE_PROC_MY_NAME``` messes things up for non-pmix-native pmix components.

I'm endeavouring to fix on cray.

hppritcha:
@hjelmn I don't know if this what you hit on our system or not.

jladd-mlnx:
@artpol84 @karasevb Please test with srun launch.

hppritcha:
@jladd-mlnx I'll leave it to your folks to fix the s1 and s2.  Should be pretty simple.

jladd-mlnx:
@hppritcha roger that. 

@karasevb Please align s1 and s2 components accordingly with the new PMIx naming convention/wildcard handling.

artpol84:
Ok, thank you @hppritcha!

artpol84:
@hppritcha I just returned to this and while analyzing your changes I found that you changed some invocations of  `opal_pmix_base_store` while other remains:
https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/cray/pmix_cray.c#L377:L420
what was the motivation behind that?

artpol84:
Ok, I think I figured out why - some data has to be stored with wildcard while the other is dedicated to this rank

hppritcha:
Correct.  I've no idea what the motivation was for the change but just followed the pattern being used for native PMIX.

artpol84:
@hppritcha 
I see that `OPAL_PMIX_LOCAL_PEERS` is also retrieved using wildcard: 
https://github.com/open-mpi/ompi/pull/1897/files#diff-3f6a054da48e108773c5ffa42ba74237R313

while in your fix it's still a proc:
https://github.com/open-mpi/ompi/pull/1904/files#diff-a8e6d270000c565f65ae4b78e4e03b50L376

orte/mca/ess/pmi/ess_pmi_module.c
```diff
@@ -203,7 +203,7 @@ static int rte_init(void)
     /* get the number of local peers - required for wireup of
      * shared memory BTL */
     OPAL_MODEX_RECV_VALUE(ret, OPAL_PMIX_LOCAL_SIZE,
-                          ORTE_PROC_MY_NAME, &u32ptr, OPAL_UINT32);
+                          &name, &u32ptr, OPAL_UINT32);
```
artpol84: @rhc54 @hppritcha 
Can we change `name` to `widcard` so it'll be much easier to read the code?

artpol84:
@hppritcha 
I guess that `OPAL_PMIX_UNIV_SIZE` should be also wildcarded:
https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/cray/pmix_cray.c#L318

orte/mca/ess/pmi/ess_pmi_module.c
rhc54: It isn't the wildcard name - it is the name of this nspace, but with a wildcard rank.

artpol84: I meant the name of the variable. Name `name` is confusing to me. The following line would look much nicer for me:
```C
OPAL_MODEX_RECV_VALUE(ret, OPAL_PMIX_LOCAL_SIZE,
                          &wildcard, &u32ptr, OPAL_UINT32);
```

rhc54:
This lengthy thread explains the change:

https://github.com/pmix/master/pull/114

Simplified: reducing the number of hash table lookups required that we place some constraints on how the info was stored. We have some info that is rank-specific, but other info that is job-specific and node-specific (and spans ranks). So we had to create and enforce some new rules.

We are still working them out, but hopefully are over the worst of it.

orte/mca/ess/pmi/ess_pmi_module.c
rhc54: But it isn't the wildcard name - and that could just as easily confuse someone!

Maybe call it something that indicates "wildcard_rank"?

artpol84: yep, looks fine as well.
Currently I'm constantly worry that `name.rank` may change somewhere in between and I can't be sure that at certain point it still contains `rank == wildcard`.
Technically the same applies to `wildcard_rank` but here you understand that original developer ment a wildcard here and unlikely that he would change it somewhere in between.

rhc54: Sounds good - can you go ahead and make the change?

artpol84: Sure, will do that along with s1/s2 fixes today

PR #1903, opened 2016-07-26
btl/openib: set send flags only after endpoint is connected

hjelmn:
The max inline send size on a queue pair is not available until after
the endpoint is connected. Before this commit the send flags
(including the inline flag) were set before this value was
initialized. This commit moves setting the send_flags down to
mca_btl_openib_put_internal which is only called after the endpoint is
connected. This fixes a bug when using osc/rdma.


hjelmn:
@regrant This fixes the issue seen on stampede.

regrant:
@hjelmn Thanks, we'll test it right away.

PR #1904, opened 2016-07-26
pmix/cray: switch to using wildcards for some

hppritcha:
items so that at least srun native launch on
cray works again.

More issues to fix when using alps.

@hjelmn - alps fixes coming next.


hppritcha:
This will need to be PR'd to ompi-release if the equivalent of #1897 goes in to that branch.

PR #1908, opened 2016-07-27
rcache/udreg: make reference count thread safe

hjelmn:


hjelmn:
@hppritcha I will probably need to back-port this to 2.0.1. The reference counts were getting screwed up when using multiple threads with RMA.

PR #1912, opened 2016-07-28
mpisync: Fix a compilation error.

rivis:
This commit fixes the undefined `OPAL_MAXHOSTNAMELEN` error
which arises only when `--enable-timing` is specified for
`configure`.

This bug exists only in master branch because the commit 3322347
is not merged into other branches.

### DIFF
--- a/ompi/tools/mpisync/sync.c
+++ b/ompi/tools/mpisync/sync.c
@@ -8,6 +8,8 @@
  * $HEADER$
  */
 
+#include "opal_config.h"
+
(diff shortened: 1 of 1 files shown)


jsquyres:
@rivis I believe 3322347 has been merged to the v2.x branch now -- can you check, and file a PR if necessary?  Thanks!

rivis:
@jsquyres Sorry, I looked very old v2.x branch. I filed a PR open-mpi/ompi-release#1290 for v2.x.

PR #1920, opened 2016-07-29
PMIX keys passed as process keys instead of namespace keys

dsolt:
Under the new get model we are more careful about whether keys are related
to a specific rank or to the entire namespace.  In OMPI were passing several
keys down as part of the namespace because they are the same for all local
ranks, but that is technically incorrect because they can vary for ranks on
other nodes.   This change makes them part of the PROC_BLOB data as they
should be.  An example would be pmix.lranks which is not necessarily the same 
for all processes in a namespace, but was being passed down as a namespace key
because it was the same for all processes on the node.

### DIFF
--- a/orte/orted/pmix/pmix_server_register_fns.c
+++ b/orte/orted/pmix/pmix_server_register_fns.c
@@ -18,6 +18,7 @@
  *                         All rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      IBM Corporation. All rights reservered.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -171,12 +172,6 @@ int orte_pmix_server_register_nspace(orte_job_t *jdata)
         OPAL_LIST_RELEASE(info);
         return ORTE_ERR_NOT_FOUND;
     }
-    /* pass our node ID */
-    kv = OBJ_NEW(opal_value_t);
-    kv->key = strdup(OPAL_PMIX_NODEID);
-    kv->type = OPAL_UINT32;
-    kv->data.uint32 = node->index;
-    opal_list_append(info, &kv->super);
 
     /* identify our local node object within the map,
      * if we were included */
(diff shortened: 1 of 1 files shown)


orte/orted/pmix/pmix_server_register_fns.c
```diff
@@ -324,6 +288,43 @@ int orte_pmix_server_register_nspace(orte_job_t *jdata)
...
         opal_list_append(pmap, &kv->super);
 
+        /* local size - same for all on host, but not same across namespace */
+        kv = OBJ_NEW(opal_value_t);
+        kv->key = strdup(OPAL_PMIX_LOCAL_SIZE);
+        kv->type = OPAL_UINT32;
+        kv->data.uint32 = jdata->num_local_procs;
+        opal_list_append(pmap, &kv->super);
+
+        /* pass our node ID - same for all on host, but not same across namespace */
```
rhc54: @dsolt I moved the nodeID the other day (and fixed a typo this morning) to correct a problem reported by a user. You probably need to rebase to pickup that change and resolve the conflict - don't want to pass it twice

orte/orted/pmix/pmix_server_register_fns.c
```diff
@@ -324,6 +288,43 @@ int orte_pmix_server_register_nspace(orte_job_t *jdata)
...
+        kv = OBJ_NEW(opal_value_t);
+        kv->key = strdup(OPAL_PMIX_NODEID);
+        kv->type = OPAL_UINT32;
+        kv->data.uint32 = node->index;
+        opal_list_append(pmap, &kv->super);
+
+        /* construct the list of peers for transmission */
+        /* - same for all on host, but not same across namespace */
+        if (NULL != list) {
+            tmp = opal_argv_join(list, ',');
```
rhc54: I don't think this is correct - isn't this list the one we generated above that is specific to this one node? I think you'll need to regenerate this for each proc using the peers on just its node.

orte/orted/pmix/pmix_server_register_fns.c
```diff
@@ -324,6 +288,43 @@ int orte_pmix_server_register_nspace(orte_job_t *jdata)
...
+            kv->key = strdup(OPAL_PMIX_LOCAL_PEERS);
+            kv->type = OPAL_STRING;
+            kv->data.string = tmp;
+            opal_list_append(pmap, &kv->super);
+        }
+        /* construct the list of cpusets for transmission */
+        /* - same for all on host, but not same across namespace */
+        if (NULL != procs) {
+            tmp = opal_argv_join(procs, ':');
+            /* pass the list of cpusets */
```
rhc54: same thing here - this data is just for the one node. If you really want the list of cpusets for the remote procs, then you'll need to regenerate that each time thru the loop.

In that case, you are probably better off looping by node instead of by proc - that way, you can generate the list and procs arrays only once, and then loop across all procs on that node and generate their "blob".

Let me know if you want help with that optimization.

rhc54:
@dsolt Please see https://github.com/open-mpi/ompi/pull/1926 for an example of how to resolve the problems cited here. Note that the example still doesn't get the retrievals correctly defined as the job hangs in vader on my machine.

artpol84:
bot:mellanox:retest

rhc54:
This has now been committed via a different PR

Issue #1753, opened 2016-06-06
Missing Fortran named constants for predefined datatypes


kawashima-fj:
@bosilca, I found a Fortran/datatype bug and am trying to fix it but I want to hear your opinion.
The bug is that a Fortran program cannot be compiled if it uses some predefined datatypes.
The cause is that those datatypes are not defined or incorrectly defined in `include/mpif-handles.h`, which is created by `ompi/include/mpif-values.pl`.

The missing datatypes are:
- `MPI_C_BOOL`
- `MPI_LONG_LONG`

The incorrectly defined datatypes are:
- `MPI_LONG_DOUBLE_INT`
- `MPI_CXX_FLOAT_COMPLEX`
- `MPI_CXX_DOUBLE_COMPLEX`
- `MPI_CXX_LONG_DOUBLE_COMPLEX`

The incorrectly defined datatypes can be fixed easily:
```diff
--- a/ompi/include/mpif-values.pl
+++ b/ompi/include/mpif-values.pl
@@ -166,14 +166,14 @@ $handles->{MPI_DOUBLE} = 46;
 $handles->{MPI_LONG_DOUBLE} = 47;
 $handles->{MPI_FLOAT_INT} = 48;
 $handles->{MPI_DOUBLE_INT} = 49;
-$handles->{MPI_LONGDBL_INT} = 50;
+$handles->{MPI_LONG_DOUBLE_INT} = 50;
 $handles->{MPI_LONG_INT} = 51;
 $handles->{MPI_2INT} = 52;
 $handles->{MPI_SHORT_INT} = 53;
 $handles->{MPI_CXX_BOOL} = 54;
-$handles->{MPI_CXX_CPLEX} = 55;
-$handles->{MPI_CXX_DBLCPLEX} = 56;
-$handles->{MPI_CXX_LDBLCPLEX} = 57;
+$handles->{MPI_CXX_FLOAT_COMPLEX} = 55;
+$handles->{MPI_CXX_DOUBLE_COMPLEX} = 56;
+$handles->{MPI_CXX_LONG_DOUBLE_COMPLEX} = 57;
 $handles->{MPI_INT8_T} = 58;
 $handles->{MPI_UINT8_T} = 59;
 $handles->{MPI_INT16_T} = 60;
```

The problem is the two missing datatypes.

[...]

PR #1927, opened 2016-08-02
fortran: Correct named constants and a datatype name

kawashima-fj:
This PR consists of four commits and three of them are trivial. The third commit (b5c34c1) may require discussion. Please read my lengthy commit message. This issue was previously discussed in #1753 but I forgot it for a while so the chance to keep the compatibility between master and v2.x was lost. Sorry.

If this PR is pulled to master, I'll make a PR for v2.x but modify the commit to keep the binary compatibility of `MPI_C_COMPLEX` between v2.0.0 and v2.0.x.

@bosilca please review

Closes open-mpi/ompi#1753


bosilca:
@kawashima-fj Down in the datatype engine I prefer to keep the full name of the type, ompi_mpi_c_float_complex instead of ompi_mpi_c_complex (to keep the consistency with the other similar types such as ompi_mpi_c_double_complex).

kawashima-fj:
@bosilca Do you mean `MPI_C_COMPLEX` in `mpi.h` and `mpif.h` should point `ompi_mpi_c_float_complex` (C) and its index (Fortran) instead of `ompi_mpi_c_complex`? In other words,  make `ompi_mpi_c_float_complex` the canonical name and make `ompi_mpi_c_complex` the alias of it (or delete `ompi_mpi_c_complex`) in Open MPI implementation? If this understanding is correct, I'll update the PR.

bosilca:
@kawashima-fj this is indeed what I meant. thanks.

kawashima-fj:
@bosilca I updated the PR. I didn't delete `ompi_mpi_c_complex` this time because it is used some other places especially `ompi_datatype_basicDatatypes` array.

bosilca:
👍

PR #1928, opened 2016-08-02
mpool/hugepage mntent intro fallout

hppritcha:
On Cray, PR #1846 introduced a double free
situation which leads to all kinds of random memory
corruption problems.

@ggouaillardet - this will need to be included when PR #1846 is pushed back to v2.x


ggouaillardet:
:+1: 
if I read correctly, the free can also be moved from line 251 up to line 237

PR #1931, opened 2016-08-02
ess/base: set up nidmap after pmix

hjelmn:
This fixes a SEGV when the nidmap code attempts to use
opal_pmix.store_local before pmix is set up.


hjelmn:
@rhc54 This fixes the setup part but we still fail due to orted using the server_init instead of client_init.

PR #1936, opened 2016-08-03
osc/pt2pt: do not set rdma_frag after start

hjelmn:
It is possible for the start call to complete the requests. For this
reason the module rdma_frag field should be filled in before start is
called. If the request completes the completion callback will reset
the rdma_frag field to NULL. Fixes a bug discovered by @tkordenbrock.


hjelmn:
@tkordenbrock Turns out this bug was unrelated to the double-free datatype bug. Was storing a stale request and freeing it.

tkordenbrock:
@hjelmn Bummer.  OTOH, this does fix the assert() I saw in the onesided tests.  I also applied this commit to ompi-release/v2.x with the same positive results.

artpol84:
bot:mellanox:retest

PR #1939, opened 2016-08-04
btl/ugni: protect against re-entry and races in connections

hjelmn:
This commit fixes two issues that can occur during a connection:

 - Re-entry to connection progress from modex lookup. Added an
   additional endpoint state that will keep the code from re-entering
   the common endpoint create.

 - Fixed a race between a process posting a directed datagram through
   a send and a connection being progressed through opal_progress().
   The progress code was not obtaining the endpoint lock before
   attempting to update the endpoint. To limit the amount of code
   changed for 2.0.1 this commit makes the endpoint lock recursive. In
   a future update this may be changed.

### DIFF
--- a/opal/mca/btl/ugni/btl_ugni_component.c
+++ b/opal/mca/btl/ugni/btl_ugni_component.c
@@ -489,7 +489,14 @@ mca_btl_ugni_progress_datagram (mca_btl_ugni_module_t *ugni_module)
                  data, (void *) ep, remote_id));
 
     /* NTH: TODO -- error handling */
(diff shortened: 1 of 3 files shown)


hjelmn:
@hppritcha Found this bug with a threaded application on Trinity. I can probably write a better fix but this is sufficient to fix the bug and keeps the changes minimal. Target 2.0.1.

artpol84:
bot:mellanox:retest

artpol84:
bot:mellanox:retest

PR #1940, opened 2016-08-05
code cleanup: clang is now a happier panda

hppritcha:
Clang 5.1 on my mac was a sad panda compiling a couple
of files,  complaining about uninitialized stack variables.

This commit makes clang a happier panda (or at least not so sad).
@rhc54 

### DIFF
--- a/orte/runtime/orte_data_server.c
+++ b/orte/runtime/orte_data_server.c
@@ -10,7 +10,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2007      Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2012      Los Alamos National Security, LLC.
+ * Copyright (c) 2012-2016 Los Alamos National Security, LLC.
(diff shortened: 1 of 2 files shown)


orte/runtime/orte_data_server.c
```diff
@@ -178,7 +178,7 @@ void orte_data_server(int status, orte_process_name_t* sender,
     uint32_t uid = UINT32_MAX;
     opal_pmix_data_range_t range;
     orte_data_req_t *req, *rqnext;
-    orte_jobid_t jobid;
+    orte_jobid_t jobid = 0;
```
rhc54: I guess it must think there is some path that allows jobid to be used without having been unpacked? I can't readily see it - would be nice to know what it is and fix it.

Regardless, you'd be better off initializing it to ORTE_JOBID_INVALID as job=0 is the daemon job and a legitimate value.

hppritcha:
looks like mlnx has an HCOLL install issue.

hppritcha:
part of this commit needs to be patched back in to v2.x so assigning 2.0.1

PR #1941, opened 2016-08-05
configury: make memory/patcher symbol detection more robust

ggouaillardet:
by defaults, Sun compiler optimize out the test, fail detecting a symbol is missing.

### DIFF
--- a/opal/mca/memory/patcher/configure.m4
+++ b/opal/mca/memory/patcher/configure.m4
@@ -36,57 +36,17 @@ AC_DEFUN([MCA_opal_memory_patcher_COMPILE_MODE], [
 AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
     AC_CONFIG_FILES([opal/mca/memory/patcher/Makefile])
 
-    OPAL_VAR_SCOPE_PUSH([memory_patcher_have___curbrk memory_patcher_have___mmap memory_patcher_have___syscall memory_patcher_have___mmap_prototype memory_patcher_have___syscall_prototype])
-
-    memory_patcher_have___curbrk=0
-    memory_patcher_have___mmap=0
-    memory_patcher_have___mmap_prototype=0
-    memory_patcher_have___syscall=0
-    memory_patcher_have___syscall_prototype=0
-
-    AC_MSG_CHECKING([for __curbrk symbol])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([extern char *__curbrk;],[char *tmp = __curbrk;])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___curbrk=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___CURBRK], [$memory_patcher_have___curbrk],
-                       [Whether the glibc __curbrk exists])
-
-    AC_MSG_CHECKING([whether __mmap prototype exists])
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include <sys/mman.h>],[char *tmp = __mmap (NULL, 0, 0, 0, 0, 0);])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___mmap_prototype=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___MMAP_PROTO], [$memory_patcher_have___mmap_prototype],
-                       [Whether the internal __mmap call has a prototype])
-
-    AC_MSG_CHECKING([whether __mmap symbol exists])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([void *__mmap ();],[char *tmp = __mmap ();])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___mmap=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___MMAP], [$memory_patcher_have___mmap],
-                       [Whether the internal __mmap call exists])
-
-    AC_MSG_CHECKING([whether __syscall prototype exists])
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include <sys/syscall.h>],[char *tmp = __syscall (SYS_mmap, NULL);])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___syscall_prototype=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___SYSCALL_PROTO], [$memory_patcher_have___syscall_prototype],
-                       [Whether the internal __syscall call has a prototype])
-
-    AC_MSG_CHECKING([whether __syscall symbol exists])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([void *__syscall ();],[char *tmp = __syscall ();])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___syscall=1],
-                   [AC_MSG_RESULT([no])])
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@jsquyres @hjelmn could you please have a look ?

this has been discussed on the user ML

opal/mca/memory/patcher/configure.m4
```diff
@@ -45,7 +45,7 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
     memory_patcher_have___syscall_prototype=0
 
     AC_MSG_CHECKING([for __curbrk symbol])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([extern char *__curbrk;],[char *tmp = __curbrk;])],
+    AC_LINK_IFELSE([AC_LANG_PROGRAM([extern char *__curbrk;],[if (*__curbrk) return 1;])],
```
jsquyres: Could that cause a segv?  I.e., if `__curbrk` is NULL (or otherwise invalid), won't the dereference cause a problem?  Or is it guaranteed that dereferencing `__curbrk` is always ok?

jsquyres: Perhaps a safer form of this test could be:

```c
printf("%p\n", __curbrk);
```

I.e., call a function that is not in this compilation unit that takes the symbol we want to test, and does something harmless (i.e., print to stdout, which won't be seen on the configure stdout).  If `__curbrk` doesn't exist, the test will fail to compile.

BTW, is there a reason we're not using `AC_CHECK_DECLS` or `AC_CHECK_FUNCS` for `__curbrk`, `__mmap`, and `__syscall`?

ggouaillardet: i do not know why standard macros are used.

but why do you think printf is safer ?
the test is compiled and linked, it is never ran.
using printf means we should #include <stdio.h> to make picky compilers happy, and we have to hope it dones not pull a conflicting definition of __curbrk

jsquyres: You're right of course -- I neglected to notice that you were linking, not running.  So your original code is fine (and you're right, let's not get into including stdio.h).

Could you try with `AC_CHECK_DECLS` and see if that works for `__curbrk`, `__mmap`, `__syscall`?

artpol84:
bot:mellanox:retest

artpol84:
bot:mellanox:retest

ggouaillardet:
@jsquyres i made the changes and that looks good to me now.

note OSX was also broken, since ```gcc -O2``` does not generate any code with the tested symbol.
```AC_CHECK_FUNCS``` work just fine (e.g. no ```___curbrk``` on OSX```)

File: opal/mca/memory/patcher/configure.m4 [hunk 2]
```diff
@@ -38,53 +38,18 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
 
     OPAL_VAR_SCOPE_PUSH([memory_patcher_have___curbrk memory_patcher_have___mmap memory_patcher_have___syscall memory_patcher_have___mmap_prototype memory_patcher_have___syscall_prototype])
```
jsquyres: Should probably remove `memory_patcher_have___syscall` (and friends) from this `OPAL_VAR_SCOPE`, too.

opal/mca/memory/patcher/configure.m4
```diff
@@ -38,53 +38,18 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
...
+    AC_CHECK_FUNCS([__curbrk])
 
-    AC_MSG_CHECKING([for __curbrk symbol])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([extern char *__curbrk;],[char *tmp = __curbrk;])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___curbrk=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___CURBRK], [$memory_patcher_have___curbrk],
-                       [Whether the glibc __curbrk exists])
+    AC_CHECK_HEADERS([linux/mman.h sys/syscall.h])
```
jsquyres: What's this for?

opal/mca/memory/patcher/configure.m4
```diff
@@ -38,53 +38,18 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
...
 
-    AC_MSG_CHECKING([whether __mmap prototype exists])
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([#include <sys/mman.h>],[char *tmp = __mmap (NULL, 0, 0, 0, 0, 0);])],
-                   [AC_MSG_RESULT([yes])
-                    memory_patcher_have___mmap_prototype=1],
-                   [AC_MSG_RESULT([no])])
-    AC_DEFINE_UNQUOTED([OPAL_MEMORY_PATCHER_HAVE___MMAP_PROTO], [$memory_patcher_have___mmap_prototype],
-                       [Whether the internal __mmap call has a prototype])
+    AC_CHECK_DECLS([__mmap], [], [], [#include <sys/mman.h>])
```
jsquyres: I notice that you use `<sys/mman.h>` here, but check for `<linux/mman.h>`, above.  Was that intentional?

Did you want to make this test somehow dependent upon the result of the prior `AC_CHECK_HEADERS` result for `mman.h`?

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -90,7 +90,7 @@ opal_memory_patcher_component_t mca_memory_patcher_component = {
        it out) */
 };
 
-#if OPAL_MEMORY_PATCHER_HAVE___SYSCALL_PROTO && OPAL_MEMORY_PATCHER_HAVE___SYSCALL
+#if HAVE_DECL___SYSCALL && defined(HAVE___SYSCALL)
```
jsquyres: Shouldn't the order of checking be the other way around here?

```c
#if defined(foo) && foo
```

opal/mca/memory/patcher/configure.m4
ggouaillardet: will do

opal/mca/memory/patcher/configure.m4
ggouaillardet: note that was already there, i just moved it.

from ```opal/mca/memory/patcher/memory_patcher_component.c```
```c
#include <sys/mman.h>
#if defined(HAVE_SYS_SYSCALL_H)
#include <sys/syscall.h>
#endif
#if defined(HAVE_LINUX_MMAN_H)
#include <linux/mman.h>
#endif
```

bottom line, the current code requires it, and i do not know the details.

opal/mca/memory/patcher/configure.m4
ggouaillardet: my understanding of the code is that this is intentional, and i do not know the purpose of ```linux/mman.h``` vs ```sys/mman.h```

opal/mca/memory/patcher/memory_patcher_component.c
ggouaillardet: nope
if you give it a closer look, you will notice the test is
```c
#if foo && defined(bar)
```

two distinct macros with two distinct usages for two distinct purposes, courtesy of autotools

fwiw

```
AC_CHECK_FUNCS([foo])
```
will either generate
```#undef HAVE_foo``` or ```#define HAVE_foo 1```
whereas
```
AC_CHECK_DECLS([foo])
```
will either generate
```#define HAVE_DECL_foo 0``` or ```#define HAVE_DECL_foo 1```

ggouaillardet:
@jsquyres i pushed the latest changes.
i do not know much about the memory patcher internals, so i just made minimal changes in order to fix the initial issue (with optimizations, some tests are compiled out, hence some false positives)

this issue was reported on the ML, so it would be great to have the fix in v2.0.1.
shall we merge this, and PR to v2.x now ?
then we will have more time to discuss possible improvements with @hjelmn

hjelmn:
Looks good to me. Lets get this merged and PR'd to 2.0.1 today.

Issue #1877, opened 2016-07-14
OMPI interfering with rand/random series


rhc54:
Per mail to devel, from @matcabral 

Doing quick test with rand()/srand() I found that MPI_Init() seems to be calling a function in their family  that is affecting the values in the user application.  Please see below my simple test and the results. Yes, moving the second call to srand() after MPI_init() solves the problem. However, I’m confused since this was supposedly addressed in version 1.7.5. From release notes:
 
1.7.5 20 Mar 2014:
 
- OMPI now uses its own internal random number generator and will not perturb srand() and friends.
 
I tested on OMPI 1.10.2 and 1.10.3. The result is deterministic.
 
Any ideas?
 
 
Thanks,
Regards,
 
```c
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>
int main(int argc, char *argv[])
{
    int rand1;
    int rand2;
  int name_len;
    srand(100000);
    rand1 = rand();
    srand(100000);
    MPI_Init(&argc, &argv);
    rand2 = rand();
    if (rand1 != rand2) {
        printf("%d != %d\n", rand1, rand2);
        fflush(stdout);
    }
    else {
        printf("%d == %d\n", rand1, rand2);
        fflush(stdout);
    }
    MPI_Finalize();
    return 0;
}
 ```

```shell
host1:/tmp> mpirun -np 1 -host host1 -mca pml ob1 -mca btl tcp,self ./rand1
964940668 != 865007240
```

rhc54:
Doing a quick scan thru the code, I see several places that violate the "leave rand/random alone" rules:

```shell
ompi/mca/io/romio314/romio/adio/common/shfp_fname.c
25:   is rank 0's process id and xxxx is a random number. If the
42:        srand(time(NULL));
43:        i = rand();


mpi/mca/io/romio314/romio/test-internal/heap_test.c
41:void fill_random_test(test_params_t *params);
123:	fill_random_test(&test);
179:	"   -r <size>  Create a random test and verify of size <size>\n"
217:void fill_random_test(test_params_t *params) {
224:    srand(seed);
235:	params->offsets[i] = (rand() % max_key);


ompi/mca/topo/treematch/treematch/tm_malloc.c
74:  srandom(0);
77:    extra_data[i] = (char) random() % 256;

opal/mca/event/libevent2022/libevent/poll.c
183:	i = random() % nfds;

opal/mca/event/libevent2022/libevent/select.c
180:	i = random() % nfds;

[...]

PR #1944, opened 2016-08-05
romio: make romio use internal opal_random instead of rand(3).

thananon:
Cross referenced to #1877 

As I'm not familiar with romio, I don't know how to test this properly. Please review.
The code does compile.

### DIFF
--- a/ompi/mca/io/romio314/romio/adio/common/shfp_fname.c
+++ b/ompi/mca/io/romio314/romio/adio/common/shfp_fname.c
@@ -17,6 +17,13 @@
 #ifdef HAVE_TIME_H
 #include <time.h>
 #endif
+
+/*
+ * Open MPI: we have to use internal opal_random() instead of rand(3)
+ * to prevent pertubing user's randon seed
+ */
+#include <opal/util/alfg.h>
+
 /* The following function selects the name of the file to be used to
    store the shared file pointer. The shared-file-pointer file is a
    hidden file in the same directory as the real file being accessed.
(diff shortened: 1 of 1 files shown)


jsquyres:
@Edgargabriel can you recommend a test for this PR?

edgargabriel:
without knowing too much about opal_srand etc. , but I can have a look at it and run some tests. Will let you know.

jsquyres:
Sorry -- what I meant was: is there a simple MPI File-based test we could run to just verify simple functionality?  I was kinda hoping you were pseudo-familiar with the internals of ROMIO to know what top-level MPI_File_* function(s) call ADIOI_Shfp_fname.  😄

edgargabriel:
it is a shared file pointer operation, so I *think* anything using e.g. MPI_File_write_shared etc. should invoke that.

jsquyres:
@edgargabriel Cool.  Do we have any shared file pointer tests in ompi-tests?

edgargabriel:
yes, of course. E.g. you can go to
ompi-tests/mpi2basic_tests/file

adjust the Makefile.in one level above, and compile e.g. the write_shared or the read_shared test
make write_shared read_shared

You have to run them with exactly 6 processes on a file system that support file locking (e.g. local laptop/desktop should be fine most of the time), forcing using romio314. I just confirmed on my machine that those two test cases do pass with romio314

edgargabriel:
I can run those tests on this branch if you want, but I need to finish something else first urgently

jsquyres:
Excellent; thanks for running them for us!

:+1:

jsquyres:
Hah; we commented at the same time.  We can run the tests - you gave us all the relevant info. 

Arm: why not run the tests Edgar mentioned, just to be sure.

thananon:
All mpi2basic_tests/file passed 👍

edgargabriel:
just to make sure, you did add --mca io romio314 to your execution? Otherwise it would use ompio

thananon:
I did not. Will retest soon.
Thank you for catching this.

edgargabriel:
ok, just make sure to compile and run only write_shared and read_shared, one of the other testcases leads to a problem with ROMIO, but those two should work.

thananon:
test with -mca io romio read_shared and write_shared passed 👍

jsquyres:
Arm: please pr to v2.0.1

PR #1950, opened 2016-08-09
opal_random_* helpers

ggouaillardet:


ggouaillardet:
@jsquyres @thananon 
can you please review this ?

on big endian machines, ```long int``` vs ```int``` might be a big issue if sizes do differ.
ideally, this should land into v2.0.1 ...

bwbarrett:
@ggouaillardet, this is the oldest active PR, and it looks like you have next steps.  Do you still want to do something?

thananon:
I think this PR is outdated.  These changes have been merged with another commit. We can safely close this.

rhc54:
good enough - thanks!

PR #1951, opened 2016-08-10
ompi request callback update

hjelmn:
This PR changes when the callback happens on a completed request. This is completely internal to Open MPI and currently only affects osc/pt2pt. This fixes a race condition between returning a request and its completion.

### DIFF
--- a/ompi/mca/coll/hcoll/coll_hcoll_rte.c
+++ b/ompi/mca/coll/hcoll/coll_hcoll_rte.c
@@ -5,7 +5,7 @@
  * Copyright (c) 2013      The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
- * Copyright (c) 2015      Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2015-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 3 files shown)


hjelmn:
@bosilca Came across this race while testing osc/pt2pt with threads. Very nasty race. There shouldn't be a problem with calling the callback after setting req_complete because the completion callbacks are entirely internal.

lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/65d02fb1257d3f401e8b96939c97fa40

ibm-ompi:

 Gist: https://gist.github.com/5e4a1f5048eff91fedbb7aa5b245f8e4

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

bosilca:
We need to check this as we are heavily using the callback for some of our collective works.

hjelmn:
@bosilca Alright. Let me know.

bosilca:
@hjelmn we do have an issue with this change. Talking with @thananon we remember why we made the opposite change: in a multithreaded environment, if another thread is waiting on the request it might complete and free the request between the moment where the request is unlocked and the callback is called. Thus, we call the callback earlier, to make sure the request is still valid when the callback is triggered.

hjelmn:
So your code mixed ompi_request_wait/ompi_request_test with the callback code? I can see that being a problem. I was thinking about that case yesterday but since it didn't already exist in the codebase I put off fixing it.

One possibility is to save the callback and the callback data and have the callback only pass the callback data. This would work for my case. Would it work for yours?

hjelmn:
Hmm, though I would still need a way to cause the request to be freed..... That would take more callback data. Will have to think about this.

bosilca:
@hjelmn let's talk about this next week in Dallas.

PR #1968, opened 2016-08-15
Further cleanup getpwuid usage - try it first (unless completely disa…

### LINKED ISSUE #1964: Open MPI was unable to obtain the username in order to create a path
Due to some unresolved misconfiguration, `whoami` returns a number rather than a name. This situation causes OpenMPI to give the error:
```
Open MPI was unable to obtain the username in order to create a path 
for its required temporary directories. This type of error is usually 
caused by a transient failure of network-based authentication services 
(e.g., LDAP or NIS failure due to network congestion), but can also be 
an indication of system misconfiguration. 
Please consult your system administrator about these issues and try 
again.
```

This issue has been reported here:
https://www.open-mpi.org/community/lists/users/2014/10/25485.php

The workaround is to recompile OpenMPI using `--disable-getpwuid`. Recompiling OpenMPI is a rather heavy weight workaround, when all that is needed is a temporary directory.

Could the above error be downgraded to a warning instead?

Is there a workaround that doesn't require recompiling OpenMPI?

Thanks!

rhc54:
…bled), and then silently failover to try other methods.

Fixes #1964

### DIFF
--- a/orte/mca/plm/rsh/plm_rsh_module.c
+++ b/orte/mca/plm/rsh/plm_rsh_module.c
@@ -1557,15 +1557,10 @@ static int setup_shell(orte_plm_rsh_shell_t *rshell,
         struct passwd *p;
 
         p = getpwuid(getuid());
-        if( NULL == p ) {
-            /* This user is unknown to the system. Therefore, there is no reason we
-             * spawn whatsoever in his name. Give up with a HUGE error message.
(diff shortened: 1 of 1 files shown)


rhc54:
Refs https://github.com/open-mpi/ompi-release/pull/1320 as it needs to be added there

PR #1971, opened 2016-08-16
Update the session dir structure. Restore the creation of a top-level…

rhc54:
… dir based on userid so that everything is contained under the user's top-level dir. Make the next level down (the "job family" level) be either the pid (indicated by a name of "pid.N") or the job family if not launched by mpirun. This allows for proper rendezvous by direct-launched procs.


Di0gen:
bot:mellanox:retest

PR #1972, opened 2016-08-16
rsh: robustify the check for plm_rsh_agent default value

jsquyres:
Don't strcmp against the default value -- the default value may change
over time.  Instead, check to see if the MCA var source is not
DEFAULT.

### DIFF
--- a/orte/mca/plm/rsh/plm_rsh_component.c
+++ b/orte/mca/plm/rsh/plm_rsh_component.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2008-2009 Sun Microsystems, Inc.  All rights reserved.
  * Copyright (c) 2010      Oracle and/or its affiliates.  All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


jsquyres:
@rhc54 I'm not 100% sure this is the right robustification.

This check simply checks to ensure that the value was set by someone other than the default.  But what if someone set it to exactly the same value as the default value?

I guess my question is: what exactly do we need to test here?

rhc54:
I think this is exactly what we need - if someone "sets" it to the default value, then they have "confirmed" that this is what they want and we should obey it. If they specified nothing, then we want to open the door to look for alternatives.

So I'd say this is fine and should be committed - can then PR it across to 2.0.1

jsquyres:
@rhc54 +1  I'll merge.

I think this can wait until v2.0.2 -- unless you strongly feel it needs to be in v2.0.1?  I.e., I think your existing code is more-or-less fine; this is a bit stronger enforcement, but not a critical issue.  Thoughts?

rhc54:
agreed, it can wait

PR #1974, opened 2016-08-17
MCA component filter

hjelmn:
This RFC is based on discussion on how to add a top-level component (ala --net) to Open MPI to enable or disable all components of a particular network type. This is a work in progress and will likely evolve.

### DIFF
--- a/ompi/mca/coll/hcoll/coll_hcoll_component.c
+++ b/ompi/mca/coll/hcoll/coll_hcoll_component.c
@@ -29,6 +29,11 @@ static int hcoll_open(void);
 static int hcoll_close(void);
 static int hcoll_register(void);
 int mca_coll_hcoll_output = -1;
+
+mca_base_component_filter_t mca_coll_hcoll_filter = {
+    .mca_filter_include = {"infiniband", "mxm", NULL},
+};
+
 mca_coll_hcoll_component_t mca_coll_hcoll_component = {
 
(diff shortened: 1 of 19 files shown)


hjelmn:
@jjhursey This is a working example of the low-level support needed for --net. The enable functionality is not quite ready but disable works. For example --mca mca_base_component_disable shared-memory disabled btl/sm and btl/vader.

hjelmn:
On way this diverges from what we discussed is I put a mxm filter on all mxm components. I did this to give a better example of how filtering could be done.

jjhursey:
A few notes from the discussion in the Open MPI Developer's meeting:
 * If I disable `tcp` does that disable it for the ORTE daemons. Maybe this is just for the MPI application.
   - Two users of a common set of functionality (`btl`) that might make different decisions: `osc` and `ob1`
 * What is the (expansiveness of) intent of the end user when they specify an option?
 * `disable` is easy and done.
 * `enable` is more difficult. Maybe we need a `net:` prefix to the keys. Central issue is what about components that do not supply any flags and how do we know to ignore `ib` when enabling `tcp`?
 * Should we approach this as a call down model from the framework, instead of built into the low level logic? *top level filtering approach*
   - orte took a similar approach that might be simpler when you get into the weeds of details
 * Open question: is there a use case to _not_ `dlopen` the component? We would need to add a metadata structure parallel to the `.so`

PR #1977, opened 2016-08-17
osc/pt2pt: make receive count an unsigned int

hjelmn:
This receive_count MCA variable should never be negative. Change it
to an unsigned int.


hjelmn:
@regrant This will fix the 2.0.1 PR.

PR #1978, opened 2016-08-17
configure: fail if CMA requested but not supported

hjelmn:
This commit updates the CMA configure to fail if CMA is requested with
--with-cma=yes but not available. Before this commit we would likely
fail in the build if this was the case.


hjelmn:
@jsquyres Does this make the CMA check ok for you?

PR #1981, opened 2016-08-17
Support the PMIX_TIMEOUT key at the PMIx server when timeout=0 - this…

rhc54:
… indicates that the user doesn't want a lookup of any data from the host RM.


rhc54:
Fixes https://github.com/pmix/master/issues/128

PR #1982, opened 2016-08-18
pkg-config: fix static linking

jsquyres:
We need to list all major project libraries in the private libraries line to enable static linking to work properly.

@hjelmn Please test / review.


jsquyres:
@hjelmn bump

PR #1987, opened 2016-08-19
comm/cid: fix threaded CID allocation

hjelmn:
This commit should restore the pre-non-blocking behavior of the CID
allocator when threads are used. There are two primary changes: 1)
do not hold the cid allocator lock past the end of a request callback,
and 2) if a lower id communicator is detected during CID allocation
back off and let the lower id communicator finish before continuing.


hjelmn:
@edgargabriel This should work. Its passing the reproducer for me. I changed the locking and added a check to only allow the communicator with the lowest parent CID and lowest tag (create group) to proceed.

I will add the reproducer to MTT once this is in master.

edgargabriel:
great! Looks good to me.

PR #1988, opened 2016-08-19
mtl/ofi: fix a botched assignment of av_type

hppritcha:
Well now the av_type is being assigned correctly
@yburette 

hopefully finally fixed this simple thing.


yburette:
:+1:

PR #1989, opened 2016-08-20
Update usnic BTL to libfabric v1.4

jsquyres:
Do not merge until https://github.com/ofiwg/libfabric/issues/2052 is closed.

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic_component.c
+++ b/opal/mca/btl/usnic/btl_usnic_component.c
@@ -763,7 +763,6 @@ static mca_btl_base_module_t** usnic_component_init(int* num_btl_modules,
         mca_btl_usnic_component.cq_readerr_try_again_value = -FI_EAGAIN;
     }
 
-    /* libnl initialization */
     opal_proc_t *me = opal_proc_local_get();
     opal_process_name_t *name = &(me->proc_name);
     mca_btl_usnic_component.my_hashed_rte_name =
--- a/opal/mca/btl/usnic/btl_usnic_component.c
+++ b/opal/mca/btl/usnic/btl_usnic_component.c
@@ -378,7 +378,7 @@ static int check_usnic_config(opal_btl_usnic_module_t *module,
                    "not enough usnic resources",
                    true,
                    opal_process_info.nodename,
-                   info->fabric_attr->name,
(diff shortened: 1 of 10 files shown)


jsquyres:
@bturrubiates Please review.

opal/mca/btl/usnic/btl_usnic_component.c
```diff
@@ -590,6 +592,30 @@ static void free_filter(usnic_if_filter_t *filter)
...
+
+    hints.caps = FI_MSG;
+    hints.mode = FI_LOCAL_MR | FI_MSG_PREFIX;
+    hints.addr_format = FI_SOCKADDR;
+    hints.ep_attr = &ep_attr;
+    hints.fabric_attr = &fabric_attr;
+
+    int ret;
+    struct fi_info *info_list_local;
+    ret = fi_getinfo(version, NULL, 0, 0, &hints, &info_list_local);
```
bturrubiates: why not:

```c
return fi_getinfo(version, NULL, 0, 0, &hints, *info_list_local);
```

and get rid of `info_list_local` and `ret`?

opal/mca/btl/usnic/btl_usnic_component.c
```diff
@@ -841,6 +881,8 @@ static mca_btl_base_module_t** usnic_component_init(int* num_btl_modules,
         module->fabric = fabric;
         module->domain = domain;
         module->fabric_info = info;
+        module->linux_device_name = strdup(linux_device_name);
```
bturrubiates: Shouldn't you be checking the return value of `strdup` here?

jsquyres: Yes, I should.  Good catch.

opal/mca/btl/usnic/btl_usnic_component.c
jsquyres: I think you meant

```c
return fi_getinfo(version, NULL, 0, 0, &hints, info_list);
```

But yes, you're right -- that's a bit simpler.

jsquyres:
@bturrubiates Pushed new commits addressing your comments.  Still waiting for Travis to clear CI.

bturrubiates:
👍

PR #1990, opened 2016-08-20
steps towards making file I/O operations thread safe

edgargabriel:
This pr addresses two problems of file I/O operations in a multi-threaded execution

1. the framework initialization was not thread safe, since it is not performed in ompi_mpi_init (lazy initialization)
2. added a mutex lock to the ompi_file_t structure to enable per file handle protection of the component. A global mutex lock can and will lead to deadlocks in many scenarios.

### DIFF
--- a/ompi/file/file.c
+++ b/ompi/file/file.c
@@ -31,6 +31,10 @@
 #include "ompi/mca/io/base/base.h"
 #include "ompi/info/info.h"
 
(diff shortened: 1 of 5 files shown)


edgargabriel:
bot:retest

edgargabriel:
bot:mellanox:retest

PR #1993, opened 2016-08-22
topo/treematch: fix hang in mca_topo_treematch_dist_graph_create()

ggouaillardet:
since topo->weighted (boolean) might be differ between tasks,
always perform the gather.

### DIFF
--- a/ompi/mca/topo/treematch/topo_treematch_dist_graph_create.c
+++ b/ompi/mca/topo/treematch/topo_treematch_dist_graph_create.c
@@ -6,7 +6,7 @@
  * Copyright (c) 2011-2015 INRIA.  All rights reserved.
  * Copyright (c) 2012-2015 Bordeaux Poytechnic Institute
  * Copyright (c) 2015      Intel, Inc. All rights reserved
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:

the ```dist-graph/distgraph_test_4``` from the ```ibm``` test suite currently hangs in my VM (1 socket and 4 cores) with ```Reorder=1``` and ```root=0```

bosilca:
@ggouaillardet MPI requires a consistent value across processes for the weights.
> It is erroneous to supply MPI_UNWEIGHTED, or in C++ omit the weight arrays, for some but not all processes of comm_old.

I don't think your patch is necessary for a correct MPI application.

ggouaillardet:
ok, I will fix the test then, and close thus pr

thanks

ggouaillardet:
@bosilca so i fixed the test, here is the patch FYI

```patch
diff --git a/ibm/dist-graph/distgraph_test_4.c b/ibm/dist-graph/distgraph_test_4.c
index a5203ec..8bb6e4e 100644
--- a/ibm/dist-graph/distgraph_test_4.c
+++ b/ibm/dist-graph/distgraph_test_4.c
@@ -60,7 +60,7 @@ static bool correct_result(int cnt,
             if(!diff_procs(cnt, rank, Procs1[1], dest, outdegree,0))return false;
             if(!diff_procs(cnt, rank, Weight[0], destw, outdegree,0))return false;
         }else{
-            if(indegree!=1 || outdegree!=0 || weighted!=0 ||
+            if(indegree!=1 || outdegree!=0 || weighted!=1 ||
                *srce!=root || dest!=NULL || destw!=NULL)
                 return false;
 
@@ -92,7 +92,7 @@ static bool correct_result(int cnt,
             if(!diff_procs(cnt, rank, Procs1[1], srce, indegree,0))return false;
             if(!diff_procs(cnt, rank, Weight[0], srcew, indegree,0))return false;
         }else{
-            if(indegree!=0 || outdegree!=1 || weighted!=0 ||
+            if(indegree!=0 || outdegree!=1 || weighted!=1 ||
                srce!=NULL || *dest!=root || srcew!=NULL)
                 return false;
         }
@@ -693,6 +693,7 @@ static void create_graph(int cnt, int root, MPI_Comm *newcomm){
             sources =      NULL;
             degrees =      NULL,
             destinations = NULL;
+            weights = MPI_WEIGHTS_EMPTY;
         }
         break;
 
@@ -709,6 +710,7 @@ static void create_graph(int cnt, int root, MPI_Comm *newcomm){
             sources =      NULL;
             degrees =      NULL,
             destinations = NULL;
+            weights = MPI_WEIGHTS_EMPTY;
         }
         break;
 
```

PR #1994, opened 2016-08-22
Unify the PMIx2x components and minor cleanup of coll/sync

rhc54:


artpol84:
@rhc54 @jladd-mlnx @jsquyres @hppritcha 

I think that this PR was merged too quickly: submitted Aug 23 3:33 AM my time and merged 5:20 AM my time same day. There was no @-mention so I had no chance to notice this.

It might be a straightforward change but periodically I'm completely loosing the track of what's going on and things that was working previously mysteriously stop working. This is OK for the development branch but I need to be aware. We are working with PMIx as well and sometimes I have to spend half of the day rearranging things to just compile OMPI.

Another example is  b0cc9b0bc885ece2f09cfcb55c2b86a11be8bc16. Why do we have ext20 aligned with master and not pmix/release/v2.x ?? This means that currently one has no way to use/test pmix-v2.0 with any ompi: ompi-release supports only pmix-v1 and ompi/master now supports only pmix/master.

@rhc54 can you please @-mention me in all commits that affect the pmix side of OMPI and not merge important commits until you hear from me?

artpol84:
Also I think that we should separate indentation changes from the functionality changes so it will be easier to review them. For example in b0cc9b0bc885ece2f09cfcb55c2b86a11be8bc16 indentation-related changes take reasonable part of the commit complicating the review.

PR #1995, opened 2016-08-22
Change the behavior of cpus-per-rank. 

rhc54:
We previously counted each cpu against the #slots. However, IBM has pointed out that "slot" is equated to the number of processes allowed to run on each node, and not the number of cpus on the node. This has been a continuing source of confusion, so make the distinction a "hard" one.

Each process occupies a "slot". We automatically set #slots = #cpus if nothing else is told to us. If you want to run more procs than slots, you must tell us to allow oversubscription.

A process can utilize multiple pe's if that option is given. If you try to bind more than one proc to a given pe, then we will error out unless you tell us to allow overloading.

@markalle Please see if this matches your expectations.

### DIFF
--- a/orte/mca/rmaps/base/rmaps_base_support_fns.c
+++ b/orte/mca/rmaps/base/rmaps_base_support_fns.c
@@ -583,7 +583,7 @@ orte_proc_t* orte_rmaps_base_setup_proc(orte_job_t *jdata,
     proc->node = node;
     node->num_procs++;
     if (node->slots_inuse < node->slots) {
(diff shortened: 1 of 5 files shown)


jjhursey:
Our of curiosity, if the user passes `--use-hwthread-cpus` then #slots = #hwthreads without the need for oversubscription, right? I just wanted to confirm that this case is still covered with this change.

rhc54:
I didn't change anything about how the number of pe's is computed, so this shouldn't be impacted.

markalle:
I haven't checked out and built the change, but I think the description matches what I would expect. Thanks.

PR #2001, opened 2016-08-24
fix sec/native module under Solaris and other misc issues

ggouaillardet:


ggouaillardet:
this contains reference to commits that have not yet been merged into upstream pmix,
so this PR should not be merged before pmix/master#135

rhc54:
Merged upstream

PR #2007, opened 2016-08-24
usnic: show the local UDP ports

jsquyres:

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic_module.c
+++ b/opal/mca/btl/usnic/btl_usnic_module.c
@@ -69,6 +69,30 @@
 static void finalize_one_channel(opal_btl_usnic_module_t *module,
                                  struct opal_btl_usnic_channel_t *channel);
 
+static int channel_addr2str(opal_btl_usnic_module_t *module, int channel,
+                            char *str, size_t len_param)
+{
+    size_t len;
+
+    len = len_param;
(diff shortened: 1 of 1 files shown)


jsquyres:
@bturrubiates Please review.

This functionality came up the other day when helping debug some cluster-setup issues; we wanted to trace the traffic from one MPI process to another, but MPI processes didn't provide diag output indicating their source IP address *and UDP port*.  This commit adds verbose output (when running with `--mca btl_base_verbose`) that shows both the local UDP ports, and also, when "connecting" to a remote peer, show both the local and remote IP/UDP ports.  It's solely for diagnostic / verbose output purposes.

jsquyres:
@bturrubiates actually recommends this kind of thing: https://gist.github.com/jsquyres/dd959a48b6d9d34a41261c7cf7998a5e

I don't have time to look at this at the moment, but wanted to stash it here so I don't lose it.

jsquyres:
@bturrubiates Updated with your suggestion.  Have a look.

jsquyres:
@bturrubiates Got your latest feedback in this update.

bturrubiates:
Looks good 👍

jsquyres:
I'm not sure why the Mellanox Jenkins would timeout in its test; this PR only deals with code in the usnic BTL, which the Mellanox Jenkins does not test.

PR #2010, opened 2016-08-25
common/ompio: make --disable-mpi-io work in static builds

edgargabriel:
@jsquyres do you want to review this? You are way more experienced with the autogen/Makefile magic of Open MPI. As far as I can see, this commit fixes the problem with --disable-mpi-io for static builds, but I am not entirely sure whether it is the 'correct' solution.

### DIFF
--- a/ompi/mca/common/ompio/Makefile.am
+++ b/ompi/mca/common/ompio/Makefile.am
@@ -30,7 +30,11 @@ sources = \
 	common_ompio_file_view.c   \
 	common_ompio_file_read.c   \
 	common_ompio_file_write.c
+else
 
+headers = 
+sources = 
+endif
 
 # To simplify components that link to this library, we will *always*
 # have an output libtool library named libmca_<type>_<name>.la -- even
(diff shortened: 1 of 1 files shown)


ompi/mca/common/ompio/Makefile.am
```diff
@@ -30,7 +30,11 @@ sources = \
 	common_ompio_file_view.c   \
 	common_ompio_file_read.c   \
 	common_ompio_file_write.c
+else
 
+headers = 
+soruces =
```
ggouaillardet: @edgargabriel there is a minor typo, it should be ```sources=```

fwiw, i made things a bit differently in #2018

jsquyres:
@ggouaillardet's comment about the typo is correct -- it should be fixed.

I like the general direction of @ggouaillardet's PR (#2018), but it's not quite right.  If @ggouaillardet can't make that approach work, then I think this PR should be fine (once the typo is fixed).

edgargabriel:
ok, thanks for the feedback. I fixed the typo, and we can either merge this commit to have a quick fix, followed by the nicer solution from @ggouaillardet . Otherwise, I am also happy to not commit this fix, and wait for the other solution.

jsquyres:
I think #2018 is pretty close / will be worked out shortly.  We should probably close this one.

Issue #1614, opened 2016-05-02
Only short hostname is accepted in rank file in case of affinity support


fengliplatform:
OpenMPI 1.10.1

When the rank file is like following with long name, Got error.

```
li@fengli-16-120: cat ./hostrankfile0
rank 0=fengli-17.gss.platformlab.ibm.com slot=0
rank 1=fengli-17.gss.platformlab.ibm.com slot=1
rank 2=fengli-17.gss.platformlab.ibm.com slot=2
rank 3=fengli-17.gss.platformlab.ibm.com slot=3
rank 4=fengli-16.gss.platformlab.ibm.com slot=0
rank 5=fengli-16.gss.platformlab.ibm.com slot=1
rank 6=fengli-16.gss.platformlab.ibm.com slot=2
rank 7=fengli-16.gss.platformlab.ibm.com slot=3

li@fengli-16-119: mpirun -rf ./hostrankfile0 ./hello2
--------------------------------------------------------------------------
The rankfile that was used claimed that a host was either not
allocated or oversubscribed its slots.  Please review your rank-slot
assignments and your host allocation to ensure a proper match.  Also,
some systems may require using full hostnames, such as
"host1.example.com" (instead of just plain "host1").

  Host: fengli-16
--------------------------------------------------------------------------
```
When the rank file is like following with short name, it works all right.

```
 li@fengli-16-125: cat ./hostrankfile1
rank 0=fengli-17 slot=0
rank 1=fengli-17 slot=1
rank 2=fengli-17 slot=2
rank 3=fengli-17 slot=3
rank 4=fengli-16 slot=0
rank 5=fengli-16 slot=1
rank 6=fengli-16 slot=2
rank 7=fengli-16 slot=3

li@fengli-16-124: mpirun -rf ./hostrankfile1 ./hello2
Hello world from processor fengli-16, rank 5 out of 8 processors
Hello world from processor fengli-16, rank 6 out of 8 processors
Hello world from processor fengli-16, rank 7 out of 8 processors
Hello world from processor fengli-16, rank 4 out of 8 processors
Hello world from processor fengli-17, rank 2 out of 8 processors
Hello world from processor fengli-17, rank 3 out of 8 processors
Hello world from processor fengli-17, rank 0 out of 8 processors
Hello world from processor fengli-17, rank 1 out of 8 processors
```

Note: this long hostname rank file has been tested successfully with OpenMPI 1.6.5.

***EDIT: Reformatted verbatim text***

[...]

PR #2015, opened 2016-08-25
orte: Expand use of !orte_keep_fqdn_hostnames MCA parameter

jjhursey:
* Expand the use of the `orte_keep_fqdn_hostnames` MCA parameter when it is set to false.
 * If that parameter is set to false (default) then short hostnames (e.g., `node01`) will match with the long hostnames (e.g., `node01.mycluster.org`). This allows a user (or resource manager) to mix the use of short and long hostnames.
  - Note that this mechanism does _not_ perform a DNS lookup, but instead strips off the FQDN by truncating the hostname string at the first `.` character (when not an IP address).
     - By default (`false`) the following is true: `node01 == node01.mycluster.org == node01.bogus.com` we use `node01` as the hostname.

### DIFF
--- a/orte/mca/plm/base/plm_base_launch_support.c
+++ b/orte/mca/plm/base/plm_base_launch_support.c
@@ -16,6 +16,7 @@
  * Copyright (c) 2013-2016 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -1528,7 +1529,7 @@ int orte_plm_base_setup_virtual_machine(orte_job_t *jdata)
     bool one_filter = false;
     int num_nodes;
     bool default_hostfile_used;
-    char *hosts;
+    char *hosts = NULL;
     bool singleton=false;
     bool multi_sim = false;
 
--- a/orte/mca/rmaps/base/rmaps_base_support_fns.c
+++ b/orte/mca/rmaps/base/rmaps_base_support_fns.c
@@ -13,6 +13,7 @@
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2014-2016 Intel, Inc. All rights reserved.
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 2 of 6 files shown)


jjhursey:
Related to Issue #1614 but a more general solution.

jjhursey:
@rhc54 I don't know if you want to take a look at this PR before I merge it in. I'll let it sit here a for a little while, but it would be nice to get into 2.0.2 as it fixes a problem found on an internal test environment.

rhc54:
scratching my head...it looks to me like this does exactly what happens when you set ```orte_keep_fqdn_hostnames=0```, doesn't it? I mean, you did fill in a few places where we weren't checking it, but other than that - why a new param?

jjhursey:
You made a comment in Issue #1614 about some environments where the difference between short versus long hostnames is important. The `orte_keep_fqdn_hostnames=0` behavior would make the `mpirun`/HNP hostname shortened. So I wasn't sure if there was a user depending on this mixing of hostnames behavior.

That being said, if the intention of  `orte_keep_fqdn_hostnames=0` was to do exactly what `orte_use_mixed_hostnames=t` is supposed to do then we can reuse that MCA variable. That would reduce the number of options, and future confusion about the difference between these two options.

So I'm good with just reusing `orte_keep_fqdn_hostnames=0` if you are.

I think I got all of the places in the code that needed to be modified to make this work. Certainly it worked for our use cases with LSF, rankfiles, hostfiles, and `-host`.

rhc54:
Hmmm...perhaps I should clarify that comment. What I was saying was that we don't support mixing short and long hostnames because we've had problems in the past with host confusion when we tried to do so. Thus, the param dictates that you use either _all_ short names, or _all_ long names.

The user was asking us to support mixing the two, and I refused. 😄

jjhursey:
Ah. In that case then:
 * `orte_keep_fqdn_hostnames=true` : use only long names (as available - I did not add code to do a DNS lookup on the hostnames to fill out this path.)
   - `node01 != node01.mycluster.org != node01.bogus.com`
 * `orte_keep_fqdn_hostnames=false`: (default) use only short names. If a long hostname is passed then we shorten it to the first section (before the `.`) before processing it internally.
   - `node01 == node01.mycluster.org == node01.bogus.com` : since all shorten to `node01`

rhc54:
Yep, that is correct!

jjhursey:
OK. I'll update the PR tomorrow to bring those options together.

jjhursey:
@rhc54 I updated the PR with a new commit (I'll squash them together before merge). This should match what we discussed.

ibm-ompi:

 Gist: https://gist.github.com/c34c2ef0c667c7d2968dc69e75dfa48e

ibm-ompi:

 Gist: https://gist.github.com/50fb5793fcad3937828ba301bf17ccd7

jjhursey:
I always find it funny when the IBM CI fails on one of my PRs 😄  This is due to a temporary file system issue on the test machine. I've deactivated our CI until it comes back.

rhc54:
👍 
Thanks!

jjhursey:
bot:ibm:retest

edvinas31:
Hi, I know it is a way old thread here, but is there any way to specify orte_keep_fqdn_hostnames value globally in a system instead of doing it from the command line?

rhc54:
Put the MCA param in the default param file where OMPI was installed.

Issue #2009, opened 2016-08-25
--disable-mpi-io broken on master


jsquyres:
@edgargabriel It looks like some recent OMPIO changes on master broke `--disable-mpi-io`.  Can you have a look?

https://mtt.open-mpi.org/index.php?do_redir=2344

edgargabriel:
I thought I fixed that a while back, but there seems to be an issue when using --disable-dlopen. Will look into that. Can confirm and reproduce the problem.

PR #2018, opened 2016-08-26
configury: fix --disable-mpi-io for static builds

ggouaillardet:
- move the mpi-io configury option into config/ompi_configure_options.m4
 - add ompi/mca/common/configure.m4 so this module is not built when
   Open MPI is configure'd with --disable-mpi-io

Fixes open-mpi/ompi#2009

### DIFF
--- a/config/ompi_configure_options.m4
+++ b/config/ompi_configure_options.m4
@@ -17,7 +17,7 @@ dnl Copyright (c) 2009      Los Alamos National Security, LLC.  All rights
 dnl                         reserved.
 dnl Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
 dnl Copyright (c) 2013      Intel, Inc.  All rights reserved.
-dnl Copyright (c) 2015      Research Organization for Information Science
+dnl Copyright (c) 2015-2016 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl
 dnl $COPYRIGHT$
@@ -254,6 +254,14 @@ AS_IF([test $OMPI_TRY_FORTRAN_BINDINGS -lt $OMPI_FORTRAN_USEMPIF08_BINDINGS],
 AC_DEFINE_UNQUOTED([OMPI_BUILD_FORTRAN_F08_SUBARRAYS],
                    [$OMPI_BUILD_FORTRAN_F08_SUBARRAYS],
                    [Whether we built the 'use mpi_f08' prototype subarray-based implementation or not (i.e., whether to build the use-mpi-f08-desc prototype or the regular use-mpi-f08 implementation)])
+
+AC_ARG_ENABLE([mpi-io],
+    [AC_HELP_STRING([--disable-mpi-io],
+        [Disable built-in support for MPI-2 I/O, likely because
+         an externally-provided MPI I/O package will be used.
+         Default is to use the internal framework system that uses
+         the ompio component and a specially modified version of ROMIO
+         that fits inside the romio314 component])])
 
 ])dnl
 
--- a/ompi/mca/common/ompio/configure.m4
+++ b/ompi/mca/common/ompio/configure.m4
@@ -1,0 +1,21 @@
+# -*- shell-script -*-
+#
+# Copyright (c) 2016      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# MCA_ompi_common_ompio_CONFIG([action-if-can-compile],
+#                         [action-if-cant-compile])
+# ------------------------------------------------
+AC_DEFUN([MCA_ompi_common_ompio_CONFIG],[
+    AC_CONFIG_FILES([ompi/mca/common/ompio/Makefile])
+
+    AS_IF([test "$enable_mpi_io" != "no"],
+          [$1],
+          [$2])
+])dnl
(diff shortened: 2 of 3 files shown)


ggouaillardet:
@jsquyres this is a variant of #2010  
can you please advise or merge the most suitable PR ?

@edgargabriel FYI

ompi/mca/common/configure.m4
```diff
@@ -0,0 +1,20 @@
...
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# MCA_ompi_common_CONFIG(project_name, framework_name)
+# -------------------------------------------
+AC_DEFUN([MCA_ompi_common_CONFIG],
+[
+    MCA_CONFIGURE_FRAMEWORK([$1], [$2], [$define_mpi_io])
```
jsquyres: Clever!  👍

jsquyres: ...actually, after my first "oh, that's cool!" reaction, I don't think that this is *quite* right yet.

This solution will disable *all* common components when `--disable-mpi-io` is used, even if those common components have nothing to do with MPI IO.

Granted, right now, there's only `ompi/mca/common/ompio`, but there could well be other common components that are unrelated to MPI IO (we have a bunch in `opal/mca/common`, for example).

Can you adjust?

ggouaillardet: yep, will do

ggouaillardet:
@jsquyres i think i got it right this time, can you please double check ?

note you can also directly close this PR and merge #2020 (```--disable-io-ompio```) since it includes this commit

jsquyres:
I'm inclined to merge this one and *not* merge #2020 -- do we really need `--disable-io-ompi`?  Come to think of it, why do we (still) have `--disable-io-romio`?  Both of these components (and hopefully `common/ompio` and any other ompi-related components...?) can be disabled via `--enable-mca-no-build`.  Do we need top-level switches for ROMIO / OMPIO?

The use case for disabling MPI IO *altogether* (including the PMI/MPI IO API functions) is different: it allows Open MPI with an external installation of ROMIO.

jjhursey:
The reason for having a `--disable-io-ompi` is because it's not easy to enumerate all of the frameworks (/components) that are related to OMPIO (I had a hard time making sure I got them all). For ROMIO it is pretty straightforward. That's why I would advocate for having a `--disable-io-ompi` option to pair with the `--disable-io-romio`. Even if that just turns around and enumerates the `--enable-mca-no-build` (Which seems like a fine way to go about it as long as it is maintained).

config/ompi_configure_options.m4
```diff
@@ -255,5 +255,16 @@ AC_DEFINE_UNQUOTED([OMPI_BUILD_FORTRAN_F08_SUBARRAYS],
                    [$OMPI_BUILD_FORTRAN_F08_SUBARRAYS],
                    [Whether we built the 'use mpi_f08' prototype subarray-based implementation or not (i.e., whether to build the use-mpi-f08-desc prototype or the regular use-mpi-f08 implementation)])
 
+AC_ARG_ENABLE([mpi-io],
+    [AC_HELP_STRING([--disable-mpi-io],
+        [Disable built-in support for MPI-2 I/O, likely because
+         an externally-provided MPI I/O package will be used.
+         Default is to use the internal component system and
+         its specially modified version of ROMIO])])
```
jsquyres: The default is now to use OMPIO, not ROMIO, right?

I know you just copied this help text, but I think it is now stale / outdated and needed to be updated anyway... 😄

ompi/mca/common/ompio/configure.m4
```diff
@@ -0,0 +1,21 @@
...
+# $HEADER$
+#
+
+# MCA_ompi_common_ompio_CONFIG([action-if-can-compile],
+#                         [action-if-cant-compile])
+# ------------------------------------------------
+AC_DEFUN([MCA_ompi_common_ompio_CONFIG],[
+    AC_CONFIG_FILES([ompi/mca/common/ompio/Makefile])
+
+    AS_IF([test "$define_mpi_io" = "1"],
```
jsquyres: Might be good to put a comment here explaining what `$define_mpi_io` is, and why it can enable/disable the entire common/ompi framework (I know you just copied the code from before, but that name is a bit... weird, and doesn't really follow the conventions of other enable/disable vars elsewhere in the code base).

ompi/mca/io/configure.m4
```diff
@@ -14,20 +16,11 @@
...
-    AC_ARG_ENABLE([mpi-io],
-        [AC_HELP_STRING([--disable-mpi-io],
-            [Disable built-in support for MPI-2 I/O, likely because
-             an externally-provided MPI I/O package will be used.
-             Default is to use the internal component system and
-             its specially modified version of ROMIO])])
-
     OMPI_MPIF_IO_CONSTANTS_INCLUDE=
     OMPI_MPIF_IO_HANDLES_INCLUDE=
     AS_IF([test "$enable_mpi_io" != "no"],
```
jsquyres: Should this line now be checking `$define_mpi_io` instead of `$enable_mpi_io`?

config/ompi_configure_options.m4
ggouaillardet: ok, will revamp that

ompi/mca/io/configure.m4
ggouaillardet: this is equivalent, but let's change it to be more consistent

ggouaillardet: on second thought, i'd rather use ```$enable_mpi_io``` everywhere, and locally define ```$define_mpi_io``` only where needed in ```ompi/mca/io/configure.m4```

config/ompi_configure_options.m4
ggouaillardet: in this context, i think we want to say ```ompi/mca/io``` framework vs external ```MPI-IO``` library

ggouaillardet:
@jsquyres i made the changes

PR #2019, opened 2016-08-26
orte/schizo: fix binding detection in slurm component

artpol84:
in SLURM 16.05 the SLURM_CPU_BIND_TYPE is equal to "mask_cpu:"
instead of "mask_cpu". Account for that.


artpol84:
@rhc54 please reviev

artpol84:
@rhc54 I see that there is no "slurm" schizo component in v2.x. Is to supposed to be there?

jladd-mlnx:
#2072 

Need SLURM schizo component in v2.x - it is currently in master.

jsquyres:
Is it as simple as just adding the slurm schitzo component?  I thought there were other code dependencies in the core (e.g., mods to `orterun`, corresponding `ess` components, etc.)...?

artpol84:
@jsquyres let me check that and I'll update once I know.

Issue #1854, opened 2016-07-07
dlopen_test crash with xlc on PPC in v2.x


jjhursey:
Reported by @PHHargrove on the devel mailing list:
 * https://www.open-mpi.org/community/lists/devel/2016/07/19156.php
 * https://www.open-mpi.org/community/lists/devel/2016/06/19107.php

In the email exchange, @hjelmn mentioned maybe having a fix for this - maybe can comment about what he had in mind.

On a ppc64le system with the XL compiler I was able to reproduce with the following configure arguments:
```shell
shell$ xlc -qversion
IBM XL C/C++ for Linux, V13.1.2 (5725-C73, 5765-J08)
Version: 13.01.0002.0000
shell$ ./configure --prefix=/tmp/ompi/install/ompi-2.0.0rc4 CC=xlc CXX=xlC FC=xlf --disable-mpi-fortran
```

```shell
shell$ make check
...
PASS: predefined_gap_test
PASS: predefined_pad_test
../../config/test-driver: line 107: 130708 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: dlopen_test
============================================================================
Testsuite summary for Open MPI 2.0.0rc4
============================================================================
# TOTAL: 3
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See ompi/debuggers/test-suite.log
Please report to http://www.open-mpi.org/community/help/
============================================================================
```

```shell
shell$ cat ompi/debuggers/test-suite.log
======================================================
   Open MPI 2.0.0rc4: ompi/debuggers/test-suite.log
======================================================

# TOTAL: 3
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: dlopen_test
=================

[c712f5n10:130708] *** Process received signal ***
[c712f5n10:130708] Signal: Segmentation fault (11)
[c712f5n10:130708] Signal code: Address not mapped (1)
[c712f5n10:130708] Failing at address: (nil)
[c712f5n10:130708] [ 0] [0x3fffa03f0478]
[c712f5n10:130708] [ 1] [0x0]
[c712f5n10:130708] [ 2] /lib64/power8/libc.so.6(_IO_setb+0xa0)[0x3fff9fca9c30]
[...]

PR #2021, opened 2016-08-26
opal/patcher: fix xlc support

hjelmn:
The xlc compiler seems to behave in a different way that gcc when it
comes the inline asm. There were two problems with the code with xlc:

 - The TOC read in mca_patcher_base_patch_hook used the syntax
   register unsigned long toc asm("r2") to read $r2 (the TOC
   pointer). With gcc this seems to behave as expected but with xlc
   the result in toc is not the same as $r2. I updated the code to use
   asm volatile ("std 2, %0" : "=m" (toc)) to load the TOC pointer.

 - The OPAL_PATCHER_BEGIN macro is meant to be the first thing in a
   hook. On PPC64 it loads the correct TOC pointer (thanks to
   mca_patcher_base_patch_hook) and saves the old one. The
   OPAL_PATCHER_END macro restores the TOC pointer. Because we *need*
   the TOC to be correct before it is accessed in the hook the
   OPAL_PATCHER_BEGIN macro MUST come first. We did this and all was
   well with gcc. With xlc on the other hand there was a TOC access
   before the assembly inserted by OPAL_PATCHER_BEGIN. To fix this
   quickly I broke each hook into a pair of function with the
   OPAL_PATCHER_* macros on the top level functions. This works around
   the issue but is not a clean way to fix this. In the future we
   should 1) either update overwrite to not need this, or 2) figure
   out why xlc is not inserting the asm before the first TOC read.

This fixes open-mpi/ompi#1854

### DIFF
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -98,6 +98,12 @@ opal_memory_patcher_component_t mca_memory_patcher_component = {
 #define memory_patcher_syscall syscall
 #endif
 
+/* All the hooks in this file have two levels. The first level has the OPAL_PATCHER_* macros
+ * around the call to the second level. This was done because with xlc the compiler was
+ * generating an access to r2 before the OPAL_PATCHER_* assembly. This was loading invalid
+ * data. If this can be resolved the two levels can be joined.
+ */
+
 /*
  * The following block of code is #if 0'ed out because we do not need
  * to intercept mmap() any more (mmap() only deals with memory
@@ -148,9 +154,8 @@ static void *intercept_mmap(void *start, size_t length, int prot, int flags, int
 
 static int (*original_munmap) (void *, size_t);
 
-static int intercept_munmap(void *start, size_t length)
+static int _intercept_munmap(void *start, size_t length)
 {
-    OPAL_PATCHER_BEGIN;
     int result = 0;
 
     /* could be in a malloc implementation */
@@ -162,6 +167,13 @@ static int intercept_munmap(void *start, size_t length)
         result = original_munmap (start, length);
     }
 
+    return result;
+}
+
+static int intercept_munmap(void *start, size_t length)
+{
+    OPAL_PATCHER_BEGIN;
+    int result = _intercept_munmap (start, length);
     OPAL_PATCHER_END;
     return result;
 }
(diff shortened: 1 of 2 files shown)


hjelmn:
@jjhursey Please double-check. This is running cleanly on my Power8 system with xlc with both --enable-static and without.

At some point we should probably get together with the xlc team and see why the assembly wasn't working as expected.

hjelmn:
@jsquyres Ideally this should be in 2.0.1 but IBM could use this in their own internal version until we get out 2.0.2.

jjhursey:
I'll try to test on Monday.
@markalle Can you also take a look at this? I'm not as familiar with the patcher code as you might be.

hjelmn:
Merging to let MTT have a crack at it.

PHHargrove:
I will also try to retest at least three distinct ppc64 platforms this weekend if I have sufficient net access.

PHHargrove:
My testing consisted of building the most recent master tarball (openmpi-dev-4711-gd33204b), including:
  + configure
  + make
  + make check
  + make install
  + make clean
  + run "make" in the examples subdir, with new install in PATH and LD_LIBRARY_PATH
  + "mpirun -np 2 ./ring_c" with platform-appropriate "-mca btl FOO" and/or "-mca mtl BAR"

The following two both passed all of those steps, while the 2.0.1rc1 tarball got a SEGV in dlopen_test from "make check" (and thus didn't get as far as a ring_c test which the following ran fine):

Power8 little-endian
IBM XL C/C++ for Linux, V13.1.2 (5725-C73, 5765-J08)
Version: 13.01.0002.0000

Power7 big-endian
IBM XL C/C++ for Linux, V13.1
Version: 13.01.0000.0000

The following had not tested 2.0.1rc1 previously (but when tested today failed in a DIFFERENT way to be reported soon), but passed on master:

Power7 big-endian
IBM XL C/C++ for Blue Gene, V12.1
Version: 12.01.0000.0014
[NOTE: testing was on the Power7 front-end, NOT a Blue Gene compute node]

PHHargrove:
> The following had not tested 2.0.1rc1 previously (but when tested today failed in a DIFFERENT way to be reported soon), but passed on master:
> 
> Power7 big-endian
> IBM XL C/C++ for Blue Gene, V12.1
> Version: 12.01.0000.0014
> [NOTE: testing was on the Power7 front-end, NOT a Blue Gene compute node]

When 2.0.1rc is configured with `--enable-builtin-atomics` I can get past the "DIFFERENT" failure mentioned above, and reproduce my original SEGV from dlopen_test.  So, this platform also represents one "fixed" on master while broken in 2.0.1rc1.

jjhursey:
I just tested `master` with the same environment/configuration as in https://github.com/open-mpi/ompi/issues/1854 - it passed all `make check` tests. HEAD of branch:
```b0d86388 Merge pull request #2015 from jjhursey/topic/mixed-hostnames```

I also tested the current `v2.x` branch and it passed as well. So that's odd as it had been failing before. HEAD of branch:
``` 17bbb5c1 Merge pull request #1338 from hjelmn/v2.x_ugni```

Both configured the same way:
```
./configure --prefix=/tmp/ompi/install CC=xlc CXX=xlC FC=xlf --disable-mpi-fortran
```

hjelmn:
@jjhursey I couldn't reproduce the issue with make check. I was able to reproduce using a simple MPI application (ring_c).

hjelmn:
@PHHargrove Thanks for testing! Looks like the fix worked.

Issue #2023, opened 2016-08-27
oshmem yoda spml call to add_procs seems wrong


jsquyres:
When compiling master and v2.x:

```
spml_yoda.c: In function 'mca_spml_yoda_add_procs':
spml_yoda.c:645:33: warning: passing argument 1 of 'mca_pml.pml_add_procs' from incompatible pointer type [-Wincompatible-pointer-types]
     rc = MCA_PML_CALL(add_procs(procs, nprocs));
                                 ^
../../../../ompi/mca/pml/pml.h:550:41: note: in definition of macro 'MCA_PML_CALL'
 #define MCA_PML_CALL(a) mca_pml.pml_ ## a
                                         ^
spml_yoda.c:645:33: note: expected 'struct ompi_proc_t **' but argument is of type 'oshmem_proc_t ** {aka struct oshmem_proc_t **}'
     rc = MCA_PML_CALL(add_procs(procs, nprocs));
                                 ^
../../../../ompi/mca/pml/pml.h:550:41: note: in definition of macro 'MCA_PML_CALL'
 #define MCA_PML_CALL(a) mca_pml.pml_ ## a
                                         ^
```

I note that pml add_procs takes an `(ompi_proc_t **)`, but https://github.com/open-mpi/ompi/blob/master/oshmem/mca/spml/yoda/spml_yoda.c#L645 is calling with an `(oshmem_proc_t **)`.

This was added just 10 days ago by @ggouaillardet in  6b7bc6410168f162be930f93e0b6fa18753bcdb4.

However, `oshmem_proc_t` is defined as (https://github.com/open-mpi/ompi/blob/master/oshmem/proc/proc.h#L44-L58):

```c
struct oshmem_proc_t {
    opal_proc_t                    super;
    /* endpoint data */
    void *proc_endpoints[OMPI_PROC_ENDPOINT_TAG_MAX];
    /*
     * All transport channels are globally ordered.
     * pe(s) can talk to each other via subset of transports
     * these holds indexes of each transport into global array
     *  proc -> id, where id can be btl id in yoda or mxm ptl id
     *  in ikrit
     *  spml is supposed to fill this during add_procs()
     **/
    int                             num_transports;
    char                           *transport_ids;
};
```

which does *not* contain an `(ompi_proc_t)`, thereby making the (implicit) cast from `(oshmem_proc_t**)` to `(ompi_proc_t**)` incorrect.

@ggouaillardet Can you please have a look?

[...]

PR #2024, opened 2016-08-27
spml/yoda: fix compiler warning

jsquyres:
This is really only a partial fix.  Per https://github.com/open-mpi/ompi/issues/2023, the (oshmem_proc_t) and (ompi_proc_t) types should be unified properly (e.g., oshmem_proc_t should inherit an ompi_proc_t, or something like that).  The fact that this code works by casting between two unrelated types is fragile and susceptible to break in the future.


jsquyres:
I think that this PR was made moot by #2026 and open-mpi/ompi-release#1358.

PR #2035, opened 2016-08-31
Provide a mechanism for obtaining memory profiles of daemons and application profiles for use in studying our memory footprint

rhc54:
Setting OMPI_MEMPROFILE=N causes mpirun to set a timer for N seconds. When the timer fires, mpirun will query each daemon in the job to report its own memory usage plus the average memory usage of its child processes. The Proportional Set Size (PSS) is used for this purpose.


rhc54:
I should also note that once mpirun has reported out the memory profile, the job is aborted. So to get a profile, you need to have your process "spin" before calling MPI_Finalize or terminating, and ensure you provide an adequate timeout for the job to start and get to the spinning phase.

PR #2036, opened 2016-08-31
io/ompio: fix the reference count of basic datatypes used as etypes o…

edgargabriel:
update the reference counter for predefined datatypes correctly in our internal datatype_duplicate function.

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_file_view.c
+++ b/ompi/mca/common/ompio/common_ompio_file_view.c
@@ -32,6 +32,7 @@ static int datatype_duplicate  (ompi_datatype_t *oldtype, ompi_datatype_t **newt
 {
     ompi_datatype_t *type;
     if( ompi_datatype_is_predefined(oldtype) ) {
+	OBJ_RETAIN(oldtype);
         *newtype = oldtype;
         return OMPI_SUCCESS;
     }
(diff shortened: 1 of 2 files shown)


bosilca:
@edgargabriel can we do it in the opposite way, aka. don't release predefined datatypes ?

edgargabriel:
I thought about that, but that fix would be more intrusive (has to fix file_set_view as well as file_close). In addition, I thought logically it makes sense to increase the refcount on a datatype if we have a reference to it, even if it is a basic datatype. 

I can change that however if you feel that it would be better.

jsquyres:
I just merged this on the v2.x branch so that we could move forward on the v2.0.1 release.  I'm merging it here so that we stay in sync with master.  If we want to do something better than this PR, please open a new PR.  Thanks!

ggouaillardet:
+1 on @bosilca suggestion

we already do that (e.g. treat predefined and non predefined datatypes differently) at several places,
plus this is slightly more efficient.

since this is an enhancement/optimization, there is no rush and it can wait v2.0.2

edgargabriel:
ok, thanks for the feedback @bosilca  and @ggouaillardet. I will with draw the pr for master, and redo it along those lines. (That was actually my first solution before I realized that the other solution is a one-liner). We can than decide whether to move it to the 2.0.x series as well, or stick with the current solution for this version.

ggouaillardet:
@edgargabriel note the PR has already been merged into master and v2.x
you can either revert the commit and then issue a new one, or simply make a single commit that reverts the changes and make new ones

PR #2042, opened 2016-09-01
Several fixes related to session directories:

artpol84:
* enable OMPI to retrieve paths from RM through PMIx
* cleanups related to tempdirs.


artpol84:
@rhc54 This is a draft of what we were talking about on PMIx call. I'm still verifying the code but the conception is clear. Could you take a look and see if you are OK with this.

I will also add the PMIx keys to indicate that RM has created directories and that it will clean them up for OMPI as we discussed. I will implement this as a separate commit.

artpol84:
The keys are added but I need to test them with SLURM.

artpol84:
@rhc54 This works well for me now! Please have a look.

rhc54:
As per today's PMIx telecon, this looks fine. Just replace your ```TODO``` comments with a note that we decided to let MCA params override the RM directive and we should be good.

Thanks!

👍

jladd-mlnx:
#2072 Has a dependency.

PR #2050, opened 2016-09-02
Add a show_help message to TCP BTL when peer unexpectedly disconnects

jsquyres:
We commonly see messages on the users list where a peer has hung up because it has crashed.  Instead of having just a BTL_ERROR message, make this a real opal_show_help() message that tells the user that the peer unexpectedly hung up, and they should look into *why* that peer hung up.
    
There's a second commit on this PR that disentangles two help messages that accidentally look like they got entangled.


jsquyres:
@bosilca Look good?

bosilca:
👍

jsquyres:
I slightly improved the help message.

lanl-ompi:
Test FAILed.

PR #2059, opened 2016-09-06
build: Custom libmpi name option in configure

jjhursey:
* Add a configure time option to rename `libmpi(_FOO).*`
   - `--with-libmpi-name=STRING` defaults `STRING` to `mpi` (so `libmpi(_FOO).*` is not changed)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -280,6 +280,8 @@ fi
 OPAL_SET_LIB_PREFIX([])
 m4_ifdef([project_orte],
           [ORTE_SET_LIB_PREFIX([])])
+m4_ifdef([project_ompi],
+          [OMPI_SET_LIB_NAME([])])
 
 ############################################################################
 # Libtool: part one
--- a/ompi/Makefile.am
+++ b/ompi/Makefile.am
@@ -17,6 +17,7 @@
 # Copyright (c) 2015      Intel, Inc. All rights reserved.
 # Copyright (c) 2015      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -128,9 +129,9 @@ DIST_SUBDIRS = \
 
 # Build the main MPI library
 
-lib_LTLIBRARIES = libmpi.la
-libmpi_la_SOURCES =
-libmpi_la_LIBADD = \
+lib_LTLIBRARIES = lib@OMPI_LIBMPI_NAME@.la
+lib@OMPI_LIBMPI_NAME@_la_SOURCES =
+lib@OMPI_LIBMPI_NAME@_la_LIBADD = \
         datatype/libdatatype.la \
         debuggers/libdebuggers.la \
         mpi/c/libmpi_c.la \
@@ -144,13 +145,13 @@ libmpi_la_LIBADD = \
 
 
 if OMPI_RTE_ORTE
-libmpi_la_LIBADD += \
+lib@OMPI_LIBMPI_NAME@_la_LIBADD += \
         $(OMPI_TOP_BUILDDIR)/orte/lib@ORTE_LIB_PREFIX@open-rte.la
 endif
-libmpi_la_LIBADD += \
+lib@OMPI_LIBMPI_NAME@_la_LIBADD += \
         $(OMPI_TOP_BUILDDIR)/opal/lib@OPAL_LIB_PREFIX@open-pal.la
-libmpi_la_DEPENDENCIES = $(libmpi_la_LIBADD)
-libmpi_la_LDFLAGS = \
+lib@OMPI_LIBMPI_NAME@_la_DEPENDENCIES = $(lib@OMPI_LIBMPI_NAME@_la_LIBADD)
+lib@OMPI_LIBMPI_NAME@_la_LDFLAGS = \
        -version-info $(libmpi_so_version) \
         $(OMPI_LIBMPI_EXTRA_LDFLAGS)
 
@@ -159,7 +160,7 @@ headers =
 noinst_LTLIBRARIES =
 include_HEADERS =
 dist_ompidata_DATA =
-libmpi_la_SOURCES += $(headers)
+lib@OMPI_LIBMPI_NAME@_la_SOURCES += $(headers)
 nodist_man_MANS =
 
 # Conditionally install the header files
--- a/README
+++ b/README
@@ -12,7 +12,7 @@ Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
 Copyright (c) 2006-2011 Mellanox Technologies. All rights reserved.
 Copyright (c) 2006-2012 Oracle and/or its affiliates.  All rights reserved.
 Copyright (c) 2007      Myricom, Inc.  All rights reserved.
-Copyright (c) 2008      IBM Corporation.  All rights reserved.
+Copyright (c) 2008-2016 IBM Corporation.  All rights reserved.
 Copyright (c) 2010      Oak Ridge National Labs.  All rights reserved.
 Copyright (c) 2011      University of Houston. All rights reserved.
 Copyright (c) 2013-2015 Intel, Inc. All rights reserved
@@ -829,6 +829,14 @@ INSTALLATION OPTIONS
   Load configure options for the build from FILE.  Options on the
   command line that are not in FILE are also used.  Options on the
   command line and in FILE are replaced by what is in FILE.
+
+--with-libmpi-name=STRING
+  Replace libmpi.* and libmpi_FOO.* (where FOO is one of the fortran
+  supporting libraries installed in lib) with libSTRING.* and
+  libSTRING_FOO.*. This is provided as a convenience mechanism for
+  third-party packagers of Open MPI that might want to rename these
+  libraries for their own purposes. This option is *not* intended for
+  typical users of Open MPI.
 
 NETWORKING SUPPORT / OPTIONS
 
--- a/config/opal_set_lib_prefix.m4
+++ b/config/opal_set_lib_prefix.m4
@@ -1,6 +1,7 @@
 # -*- shell-script -*-
 #
 # Copyright (c) 2014 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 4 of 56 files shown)


jjhursey:
@jsquyres This touches the build system, so I need someone with build system expertise to look over this to make sure I didn't miss anything obvious.

I'm doing some more testing now, but I think I caught all of the places that needed to change. Notice we are just focusing on `libmpi.so` not any of the other `libmpi_` libraries like `libmpi_mpifh.so`.

An example:
```shell
shell$ ./autogen.pl --libmpi-name wookie_mpi
...
shell$  grep ompi_lib_prefix config/autogen_found_items.m4 
m4_define([ompi_lib_prefix], [wookie_mpi])
shell$ grep OMPI_LIB_PREFIX= configure
    OMPI_LIB_PREFIX=wookie_mpi
shell$ # build and install
...
shell$ cd install-dir
shell$ find . -name "libmpi.*"
shell$ find . -name "libwookie_mpi.*"
./lib/libwookie_mpi.so.0.0.0
./lib/libwookie_mpi.so.0
./lib/libwookie_mpi.so
./lib/libwookie_mpi.la
```

lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/6545aa86165dc3dd9d094fcd17a2b7f4

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/d4ee8cdade0e425b2a9fc25749b7d955

jjhursey:
rebased and ready for some review (I'd like another set of eyes on this before pushing to master). @jsquyres If you get a chance can you take a look at this PR. No big rush on this, just when you get a chance.

jjhursey:
@jsquyres ping

jsquyres:
@jjhursey Any reason *not* to also do `libmpi_FOO` and friends?  It does look a little weird to do `libjosh_mpi`, but then have `libmpi_mpifh` (vs. `libjosh_mpi_mpifh`).

I mention this because from a sysadmin perspective, it's pretty obvious when you can `ls -l libmpi*` to see all the MPI libraries.

autogen.pl
```diff
@@ -1096,6 +1098,7 @@ sub in_tarball {
   --no-ompi | -no-ompi          Do not build the Open MPI layer
   --no-orte | -no-orte          Do not build the ORTE layer
   --no-oshmem | -no-oshmem      Do not build the OSHMEM layer
+  --libmpi-name | -libmpi-name  Override the default lib name of: lib$libmpi_name_arg
```
jsquyres: Just curious: why is this at autogen time (vs. configure time)?

configure.ac
```diff
@@ -278,6 +279,8 @@ fi
 OPAL_SET_LIB_PREFIX([])
 m4_ifdef([project_orte],
           [ORTE_SET_LIB_PREFIX([])])
+m4_ifdef([project_ompi],
+          [OMPI_SET_LIB_PREFIX([ompi_lib_prefix])])
```
jsquyres: I'm a little confused -- I see that we have the mechanisms here to add prefixes for the OPAL and ORTE libraries, but those don't appear to be *used* anywhere (i.e., we just pass in blank strings, so the prefixes are always empty -- there's no way to *not* have them be empty).

I'm therefore trying to remember why this functionality is here...?  😕

ompi/Makefile.am
```diff
@@ -128,9 +129,9 @@ DIST_SUBDIRS = \
 
 # Build the main MPI library
 
-lib_LTLIBRARIES = libmpi.la
-libmpi_la_SOURCES =
-libmpi_la_LIBADD = \
+lib_LTLIBRARIES = lib@OMPI_LIB_PREFIX@.la
```
jsquyres: It's a little odd to call this value a "prefix", when, in fact, it's the entire library name.  I.e., it's:

```
lib@MPI_LIBRARY_NAME@.la
```

not

```
lib@OMPI_LIB_PREFIX@mpi.la
```

The latter would follow suit for what was done with the other libraries.  I.e., you could set the prefix to be `josh_`, and then it would be `libjosh_mpi` (etc.).

Is there a reason you want it to be the entire library name, rather than a prefix?

autogen.pl
jjhursey: I thought it was a little odd to set it during configure time since as part of a release we want this value fixed so users don't change it. However, I could probably be convinced otherwise. Just seemed like a little bit of protection.

configure.ac
jjhursey: Per another email thread, those were added for a different external project (ORCM, I believe). Where they wanted to prefix `libopen-pal.so` to something like `liborcm_open-pal.so`. I just piggybacked on that mechanism to achieve this patch.

If there is no reason to have the `OPAL_LIB_PREFIX` and `ORTE_LIB_PREFIX` then we can probably remove them. But if there is a chance that other projects might want to do such renaming then we might as well leave them in. They are not hurting anything.

ompi/Makefile.am
jjhursey: We were hoping to use this mechanism to rename `libmpi` to `libmpi_ibm` - so really a postfix. When preparing the patch I made it a bit more broad to allow both prefix and postfix by replacing the whole name. However, the variable `OMPI_LIB_PREFIX` is misleading so adopting `OMPI_LIBMPI_NAME` or `OMPI_LIBRARY_NAME` is probably better.

I can make that change once we settle on what the variable should be called.

autogen.pl
jsquyres: I guess my default position on this is: we only put stuff in `autogen.pl` that *can't* go in `configure.ac`.  Otherwise, you could make an argument that just about any `configure` CLI option should also (or only) be available in `autogen.pl`.

Indeed, we already have some branding-specific `configure` CLI options.  For example, how is the library name different than `--with-package-string=STRING` or `--with-ident-string=STRING`?

ompi/Makefile.am
jsquyres: Ah, you want a suffix.

Ok, in classic "Jeff wants a long name" form, I'd like `OMPI_MPI_LIBRARY_NAME`.  I'd settle for `OMPI_MPI_LIB_NAME`, too.  I think the extra `MPI` is important in there because, as a project, OMPI has *lots* of libraries.  In this case, you're specifically talking about the MPI library.

But then again, we have *several* MPI libraries...  So it would be good to see this for *all* the MPI libraries -- Fortran and C++ too.  E.g.:

* lib$(OMPI_MPI_LIB_NAME).la
* lib$(OMPI_MPI_LIB_NAME)_mpifh.la
* ...etc.

rhc54:
To fully maintain jeff's reputation, i think the variable needs to be ```OPEN_MESSAGE_PASSING_LIBRARY_FOR_FORTRAN_AS_WELL_AS_C_AND_C++_AND_PYTHON_LIBRARY_NAME```

jjhursey:
In reply to @jsquyres comment:
> Any reason not to also do `libmpi_FOO` and friends? It does look a little weird to do `libjosh_mpi`, but then have `libmpi_mpifh` (vs. `libjosh_mpi_mpifh`).

The truth is that we do not need it right now. However, I could envision a time where that might change. For the near term we only want to change `libmpi`, in the future that might change to include all `libmpi_FOO` - so we can extend the patch to include that. However, I would need the ability to adjust one and not the other for a while.

How about this logic as a proposal:
```text
# These two variable set either via autogen or configure
# Name of libmpi
#  -- Default 'mpi' (as in libmpi)
OMPI_LIBMPI_NAME=
# Name of libmpi_FOO
#  -- Default 'mpi' (as in libmpi_mpifh)
OMPI_LIBMPI_FOO_NAME=

# Setting OMPI_LIBMPI_NAME will also set OMPI_LIBMPI_FOO_NAME, if not already specified
if( OMPI_LIBMPI_NAME != "" && OMPI_LIBMPI_FOO_NAME == "") {
   OMPI_LIBMPI_FOO_NAME=OMPI_LIBMPI_NAME
}
```

ompi/Makefile.am
gpaulsen: No objection to doing the Fortran names also.  I think longer term the consistency would be desirable.  Short term (for our 2.0 based release), we're only interested in the libmpi_suffix, and if the community default of empty string is applied, this change should be an effective no-op for the Open MPI v2.x branch.

jjhursey:
I spoke with @jsquyres on the phone a little bit about this, and I think we came up with a workaround for the `OMPI_LIBMPI_NAME` vs `OMPI_LIBMPI_FOO_NAME` comment that would not push that distinction to the community code.

So, per the review, I'm going to make the following changes:
 * Move the prefix option from autogen time to configure time
 * Create a single `OMPI_LIBMPI_NAME` variable that will update `libmpi` and `libmpi_FOO`
   - Something like `lib@OMPI_LIBMPI_NAME@` and `lib@OMPI_LIBMPI_NAME@mpifh`
   - Update all of the `libmpi_FOO` places.

ibm-ompi:

 Gist: https://gist.github.com/b656e4b3ef6b705de75aafdbac3db94c

lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/f759fd66a0ddb20e640a77d6e61e0263

gpaulsen:
libwookie will be THE new official Open MPI library name right?  It follows the Star Wars [tm] naming conventions of the components.  Plus MPI _IS_ quite Hairy in places.

jsquyres:
Per discussion on the webex this morning, a `--with` option would likely be better (e.g., to be similar to the already-existing `--with-ident-string=STRING`).  Hopefully, this is a small / trivial change to make.  It would probably be good to document this switch in README, too (perhaps even putting an admonishment that users are not expected to use this option).

jjhursey:
Latest commit changes the mechanism to a configure option: `--with-libmpi-name=STRING`
Also updated the `README`

jjhursey:
Rebased - preparing for merge. Also fixed one corner case seen in diff below (case `--with-libmpi-name` with no `=`):

```diff
diff --git a/config/opal_set_lib_prefix.m4 b/config/opal_set_lib_prefix.m4
index 46ee4f73..69b65930 100644
--- a/config/opal_set_lib_prefix.m4
+++ b/config/opal_set_lib_prefix.m4
@@ -47,7 +47,7 @@ AC_DEFUN([OMPI_SET_LIB_NAME],[
     AC_ARG_WITH([libmpi-name],
         [AC_HELP_STRING([--with-libmpi-name=STRING],
                 ["Replace \"libmpi(_FOO)\" with \"libSTRING(_FOO)\" (default=mpi)"])])
-    if test "$with_libmpi_name" = "" || test "$with_libmpi_name" = "no"; then
+    if test "$with_libmpi_name" = "" || test "$with_libmpi_name" = "no" || test "$with_libmpi_name" = "yes"; t
         with_libmpi_name="mpi"
     fi
 
```

jsquyres:
@jjhursey Do you want to error if someone specifies `--with-mpi-name` (without a string) or `--without-mpi-name`?

jjhursey:
I followed the logic used by `--with-ident-string`
https://github.com/open-mpi/ompi/blob/master/config/opal_configure_options.m4#L421-L427

Which, looking at that logic, would have the same problem I noted a couple comments ago where if the user supplies `--with-ident-string` (instead of `--with-ident-string=`) then `with_ident_string` will be set to `yes`.

In any case, having a error in the following cases is probably good.
 * `--without-libmpi-name` (`$with_libmpi_name = "no"`)
 * `--with-libmpi-name=` (`$with_libmpi_name = ""`)

This case would be the 'default', so no warning if the user specified it or not.
 * `--with-libmpi-name` (`$with_libmpi_name = "yes"`)

jjhursey:
Pushed a rebased version with the suggested change to make `--without-libmpi-name` and error:
```shell
shell$ ./configure --without-libmpi-name
...
checking if want custom libmpi(_FOO) name... Error
configure: WARNING: Invalid to specify --without-libmpi-name
```

@jsquyres I think this is good to go. Any other comments?

jsquyres:
Sorry for chiming in late: 👍

jjhursey:
In preparing the PRs I noticed a bug in the resolution of the libmpi name for the fortran wrapper compiler. PR to fix that coming shortly.

Issue #2055, opened 2016-09-05
Compilation fails for powerpc with --enable-mpi-cxx


amckinstry:
Hi,

Enabling --enable-mpi-cxx leads to:

```
make[2]: Entering directory '/home/mckinstry/openmpi-2.0.1/ompi/mpi/cxx'
/bin/bash ../../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I../../../opal/include -I../../../ompi/include -I../../../oshmem/include -I../../../opal/mca/hwloc/hwloc1112/hwloc/include/private/autogen -I../../../opal/mca/hwloc/hwloc1112/hwloc/include/hwloc/autogen -I../../../ompi/mpiext/cuda/c  -DOMPI_BUILDING_CXX_BINDINGS_LIBRARY=1 -DOMPI_SKIP_MPICXX=1 -I../../.. -I../../../orte/include -Wdate-time -D_FORTIFY_SOURCE=2   -I/usr//include -I/home/mckinstry/openmpi-2.0.1/opal/mca/event/libevent2022/libevent -I/home/mckinstry/openmpi-2.0.1/opal/mca/event/libevent2022/libevent/include  -DNDEBUG -g -O2 -fdebug-prefix-map=/home/mckinstry/openmpi-2.0.1=. -fstack-protector-strong -Wformat -Werror=format-security -finline-functions -pthread -c -o mpicxx.lo mpicxx.cc
[...]

PR #2062, opened 2016-09-07
mpi/cxx: limit exposure of internal header to C++

hjelmn:
This commit removes the #include of errhandler/errhandler.h (which
includes opal_object.h). The inlude was only needed to get the extern
declaration of ompi_mpi_errors_throw_exceptions. Since we have the
ompi_predefined_errhandler_t in mpi.h we can declare the extern in
mpicxx.c without needing the C errhandler header.

This should fix open-mpi/ompi#2055

### DIFF
--- a/ompi/mpi/cxx/mpicxx.cc
+++ b/ompi/mpi/cxx/mpicxx.cc
@@ -34,14 +34,13 @@ namespace MPI {
     const char ompi_libcxx_version_string[] = OMPI_IDENT_STRING;
 }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
@jsquyres See any problem with this? There are probably other headers that cause problems but this is the problem is this instance.

hjelmn:
Hmm, this only fixes the one file. We will run into issues with comm.cc and file.cc. Going to think about this a little more.

PR #2063, opened 2016-09-07
Error handling improvements

### LINKED ISSUE #1698: java bindings error handling could use improving

nrgraham23:
This commit improves and corrects error handling.  In
cases where existing objects are altered after a call
to ompi_java_exceptionCheck, the results of the exception
check method are checked.  In the case of an exception,
memory is cleaned up and the code returns to Java without
altering existing objects.

Fixes #1698

@hppritcha please review this when you get a chance.

### DIFF
--- a/ompi/mpi/java/c/mpi_Comm.c
+++ b/ompi/mpi/java/c/mpi_Comm.c
@@ -11,6 +11,8 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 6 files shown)


lanl-ompi:
Test FAILed.

hppritcha:
looks good but please update the LANL copyright lines on the changed files.

siegmargross:
Thank you very much for the fix. With the following addition to your version of the exception program, I get what I want. I found the solution for the root cause on stackoverflow.

    catch (IndexOutOfBoundsException e) {
      System.err.println("Handle index out of bounds here");
      /* Try to find the root cause for the exception                   */
      Throwable cause = null; 
      Throwable result = e;
      while (null != (cause = result.getCause()) && (result != cause))  {
        result = cause;
      }
      System.err.printf ("  Process %2d: Error message: %s\n",  mytid, result);
    }

Issue #2030, opened 2016-08-30
COMM_SPAWN is broken in 2.1.x HEAD


jsquyres:
Recent Cisco and NVIDIA MTT results are showing COMM_SPAWN and related dynamic operations are failing.

We need to find out if it was broken in v2.0.0 and see if this is a regression.  This will determine whether it's a v2.0.1 blocker or not (because we'd really like to release the other important fixes in v2.0.1 ASAP, and start working on the short timeline for v2.1.0 and v2.0.2, etc.).

@sjeaugey is going to test on his tree and see if the recent v2.x commit(s) about bringing over PMIx 1.1.5 are the root of this particular problem.

sjeaugey:
Strangely enough, IBM MTT shows the issue on master but not on v2.x.

Also, I looked at the 1.1.5rc commit and it is far from being a significant change. If this commit is the issue, it should be easy to fix.

sjeaugey:
I'm currently unable to reproduce the problem on v2.x outside of MTT. On MTT though, it failed both with tcp and smcuda, and the error is a clear segmentation fault in orted (not the random pmix failure when there is a remaining socket in /tmp).

jsquyres:
@sjeaugey Can you provide a backtrace?

jsquyres:
@sjeaugey Also, can you force your MTT to use a tarball or git clone with that PMIx commit reverted?

sjeaugey:
@jsquyres I still need to be able to reproduce the problem. I'm trying to launch MTT building only v2.x and only IBM dynamic/spawn, but maybe it only fails when spawn is launched after in the middle of all other tests.

jsquyres:
@sjeaugey You might be able to run this indirectly via MTT.  E.g. (I didn't look at your MTT results; I'm guessing/assuming it's the `ibm` section where you're seeing failures, and fill in `...` below for your relevant `.ini` file and scratch directory):

```sh
$ client/mtt --file ... --scratch ... --verbose --mpi-get
$ client/mtt --file ... --scratch ... --verbose --mpi-install
$ client/mtt --file ... --scratch ... --verbose --test-get --section ibm
$ client/mtt --file ... --scratch ... --verbose --test-build --section ibm
$ client/mtt --file ... --scratch ... --verbose --test-run --section ibm-run-the-bad-test
```

[...]

PR #2077, opened 2016-09-13
btl/tcp: fix double list remove

hjelmn:
This commit fixes an abort during finalize because pending events were
removed from the list twice.

References #2030

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -394,7 +394,7 @@ static int mca_btl_tcp_component_open(void)
 
 static int mca_btl_tcp_component_close(void)
 {
-    opal_list_item_t *item;
+    mca_btl_tcp_event_t *event, *next;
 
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca Found by MTT.

bosilca:
Ah it is because opal_list_remove_first doesn't really mark the item as not on a list anymore. Good catch. 👍

rhc54:
Should we update opal_list_remove_[] to update the item as no longer on a list? Feels a bit like a bug to me that it doesn't

bosilca:
It does it on a DEBUG run, but not in an optimized one (it saves 2 stores). Whatever approach we take, we should standardize the behavior.

rhc54:
I see - yeah, makes sense then that we don't update the item.

hjelmn:
Yeah, it is an error to remove an item from a list twice since we try to keep the logic simple by not keeping track of the list. It would be a large change in behavior if opal_list_remove_item can be called twice on the same item without an append/prepend between.

PR #2079, opened 2016-09-14
pmix3x: configury: correctly handle --disable-dlopen

ggouaillardet:
the LT_* macros do overwrite the enable_dlopen variable,
so it must be tested and saved before invoking LT_INIT.
use the PMIX_ENABLE_DLOPEN_SUPPORT variable later

### DIFF
--- a/opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
+++ b/opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
@@ -2,6 +2,8 @@ dnl -*- shell-script -*-
 dnl
 dnl Copyright (c) 2010-2015 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2016      Intel, Inc. All right reserved
+dnl Copyright (c) 2016      Research Organization for Information Science
+dnl                         and Technology (RIST). All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
@@ -25,7 +27,7 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
     # (we still need to configure them all so that things like "make
     # dist" work", but we just want the MCA system to (artificially)
     # conclude that it can't build any of the components.
-    AS_IF([test "$enable_dlopen" = "no"],
+    AS_IF([test $PMIX_ENABLE_DLOPEN_SUPPORT -eq 0],
           [want_pdl=0], [want_pdl=1])
 
     MCA_CONFIGURE_FRAMEWORK([pdl], [$want_pdl])
@@ -33,7 +35,7 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
     # If we found no suitable static pdl component and dlopen support
     # was not specifically disabled, this is an error.
     AS_IF([test "$MCA_pmix_pdl_STATIC_COMPONENTS" = "" && \
-           test "$enable_dlopen" != "no"],
+           test $PMIX_ENABLE_DLOPEN_SUPPORT -ne 0],
           [AC_MSG_WARN([Did not find a suitable static pmix pdl component])
            AC_MSG_WARN([You might need to install libltld (and its headers) or])
            AC_MSG_WARN([specify --disable-dlopen to configure.])
--- a/opal/mca/pmix/pmix3x/pmix/config/pmix.m4
+++ b/opal/mca/pmix/pmix3x/pmix/config/pmix.m4
@@ -670,6 +670,20 @@ AC_DEFUN([PMIX_SETUP_CORE],[
 ])dnl
 
 AC_DEFUN([PMIX_DEFINE_ARGS],[
+    # do we want dlopen support ?
+    AC_MSG_CHECKING([if want dlopen support])
+    AC_ARG_ENABLE([dlopen],
+        [AC_HELP_STRING([--enable-dlopen],
+                        [Whether build should attempt to use dlopen (or
+                         similar) to dynamically load components.
+                         Disabling dlopen implies --disable-pdl-dlopen
+                         (default: enabled)])])
+    AS_IF([test "$enable_dlopen" = "no"],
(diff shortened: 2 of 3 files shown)


ggouaillardet:
@jsquyres could you please advise on that PR ?

a more elegant approach would be to check ```$enable_dlopen``` with the other ```AC_ARG_ENABLE``` in ```config/pmix.m4``` and then invoke ```LT_*``` __after__
i do not know enough about libtool to tell whether this is the right thing to do here or not ...

opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
```diff
@@ -25,15 +27,15 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
     # (we still need to configure them all so that things like "make
     # dist" work", but we just want the MCA system to (artificially)
     # conclude that it can't build any of the components.
-    AS_IF([test "$enable_dlopen" = "no"],
+    AS_IF([test "$PMIX_ENABLE_DLOPEN_SUPPORT" = "0"],
```
jsquyres: If you're going to change the variable to be numeric, you might as well change `=` to `-eq`.

opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
```diff
@@ -25,15 +27,15 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
...
           [want_pdl=0], [want_pdl=1])
 
     MCA_CONFIGURE_FRAMEWORK([pdl], [$want_pdl])
 
     # If we found no suitable static pdl component and dlopen support
     # was not specifically disabled, this is an error.
     AS_IF([test "$MCA_pmix_pdl_STATIC_COMPONENTS" = "" && \
-           test "$enable_dlopen" != "no"],
+           test "$PMIX_ENABLE_DLOPEN_SUPPORT" != "0"],
```
jsquyres: Ditto here -- change `=` to `-eq`.

opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
ggouaillardet: done

opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4
ggouaillardet: done

ggouaillardet:
@jsquyres i made the requested changes.

yes, i am 100% sure ```LT_INIT``` overwrite ```$enable_dlopen```
you can ```grep enable_dlopen config/libtool.m4``` or use the following patch to confirm it
```patch
diff --git a/opal/mca/pmix/pmix3x/pmix/configure.ac b/opal/mca/pmix/pmix3x/pmix/configure.ac
index caf16d2..0d98635 100644
--- a/opal/mca/pmix/pmix3x/pmix/configure.ac
+++ b/opal/mca/pmix/pmix3x/pmix/configure.ac
@@ -163,7 +163,10 @@ AS_IF([test "$enable_dlopen" = "no"],
       [PMIX_ENABLE_DLOPEN_SUPPORT=1
        AC_MSG_RESULT([yes])])
 
+enable_dlopen_saved=$enable_dlopen
 LT_INIT()
+AS_IF([test "$enable_dlopen" != "$enable_dlopen_saved"],
+      [AC_MSG_ERROR([enable_dlopen was altered, got \"$enable_dlopen\" but expected \"$enable_dlopen_saved\"])])
 LT_LANG([C])
 LT_LANG([C++])
 
```

jsquyres:
LT_* macros: ugh, that's pretty horrible.  😦 

That being said, in the main OMPI `configure.ac`, we do the `LT_*` macros very late; after most processing is done.  Have you tried moving the `LT_*` macros to be late in PMIx's `configure.ac`?

ggouaillardet:
@jsquyres i move ```LT_*``` invokation after all the ```AC_ARG_ENABLE``` and it seems to work ...
can you please review ?

jsquyres:
@rhc54 Good question: @ggouaillardet are we passing the enable/disable dlopen option from the main OMPI configure?  Or does that pass through in the environment automatically to the sub-`configure` script?  ...or is the sub-`configure` script ignoring the enable/disable option from the parent OMPI `configure` script?

ggouaillardet:
from ```opal/mca/pmix/pmix3x/configure.m4```
```
    OPAL_CONFIG_SUBDIR([$opal_pmix_pmix3x_basedir/pmix],
                       [$opal_pmix_pmix3x_args $opal_subdir_args 'CFLAGS=$CFLAGS' 'CPPFLAGS=$CPPFLAGS'],
                       [opal_pmix_pmix3x_happy=1], [opal_pmix_pmix3x_happy=0])
```

```--disable-dlopen``` is coming from Open MPI configure command line and is in ```$opal_subdir_args```

then from ```opal/mca/pmix/pmix3x/pmix/src/mca/pdl/configure.m4```
```
    # If --disable-dlopen was used, then have all the components fail
    # (we still need to configure them all so that things like "make
    # dist" work", but we just want the MCA system to (artificially)
    # conclude that it can't build any of the components.
    AS_IF([test "$enable_dlopen" = "no"],
          [want_pdl=0], [want_pdl=1])
```

currently, in (upstream) pmix
- there is no ```AC_ARG_ENABLE([dlopen])```
- ```$enable_dlopen``` is tested vs "no" or anything else
- ```$enable_dlopen``` value at the time of the test is ```unknown```, and that has been set by the ```LT_INIT``` macro

bottom line, ```pmix``` sub configure is correctly passed the ```--disable-dlopen``` argument by Open MPI (pmix3x glue), but it is not correctly handled by upstream ```pmix```.

i am not sure i got the question right, did i answer it ?

jsquyres:
@ggouaillardet Thanks for all the homework.  I think this looks good.

PR #2092, opened 2016-09-21
configury: improve NAG compiler support

ggouaillardet:
Fixes open-mpi/ompi#4034

### DIFF
--- a/config/opal_setup_wrappers.m4
+++ b/config/opal_setup_wrappers.m4
@@ -231,6 +231,7 @@ AC_DEFUN([OPAL_SETUP_RUNPATH],[
                     AC_MSG_RESULT([yes (-Wl,--enable-new-dtags)])],
                    [AC_MSG_RESULT([no])])
     AC_LANG_POP([C])
+m4_ifdef([project_ompi],[
     # Output goes into globally-visible $rpath_args.  Run this in a
     # sub-process so that we don't pollute the current process
     # environment.
@@ -268,7 +269,7 @@ EOF
 end program]])],
                    [runpath_fc_args="${wl_fc}--enable-new-dtags"],
                    [runpath_fc_args=""])
-    AC_LANG_POP([Fortran])
+    AC_LANG_POP([Fortran])])
     LDFLAGS=$LDFLAGS_save
 
     OPAL_VAR_SCOPE_POP
@@ -302,7 +303,6 @@ AC_DEFUN([RPATHIFY_LDFLAGS_INTERNAL],[
 AC_DEFUN([RPATHIFY_LDFLAGS],[RPATHIFY_LDFLAGS_INTERNAL([$1], [rpath_args], [runpath_args])])
 
 AC_DEFUN([RPATHIFY_FC_LDFLAGS],[RPATHIFY_LDFLAGS_INTERNAL([$1], [rpath_fc_args], [runpath_fc_args])])
-
 
 dnl
 dnl Avoid some repetitive code below
--- a/config/opal_setup_wrappers.m4
+++ b/config/opal_setup_wrappers.m4
@@ -172,7 +172,33 @@ EOF
 
     AS_IF([test -n "$rpath_args"],
           [WRAPPER_RPATH_SUPPORT=rpath
-           AC_MSG_RESULT([yes ($rpath_args)])],
+           cat > $rpath_script <<EOF
+#!/bin/sh
+
+# Slurp in the libtool config into my environment
+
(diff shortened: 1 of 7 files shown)


ggouaillardet:
Refs open-mpi/ompi#4034

ggouaillardet:
@jsquyres this PR was tested successfully by @nncarlson in open-mpi/ompi#4034
fwiw, it is already in ```v3.0.x``` and working just fine

could you please review this ?

@hppritcha this PR was basically waiting for someone with a NAG compiler to test it.
this has been done so i moved back the milestone to ```v2.1.2```

hppritcha:
This PR is soon going to have a birthday.  I'd like to close this if it's never going to be merged.

config/opal_setup_wrappers.m4
```diff
@@ -231,6 +231,7 @@ AC_DEFUN([OPAL_SETUP_RUNPATH],[
                     AC_MSG_RESULT([yes (-Wl,--enable-new-dtags)])],
                    [AC_MSG_RESULT([no])])
     AC_LANG_POP([C])
+m4_ifdef([project_ompi],[
```
jsquyres: I understand that this was done to make this commit be a minimal/trivial commit.  But effectively putting a new block in the middle of an existing block without re-indenting everything is... misleading, at best.  If this section is going to be conditionally used, it should probably be noted better somehow -- perhaps by becoming its own M4 macro.

ggouaillardet:
New activity occured 2 weeks ago. As far as i am concerned, this PR should be reviewed and merged

config/opal_setup_wrappers.m4
```diff
@@ -205,6 +231,45 @@ AC_DEFUN([OPAL_SETUP_RUNPATH],[
...
+    # Output goes into globally-visible $rpath_args.  Run this in a
+    # sub-process so that we don't pollute the current process
+    # environment.
+    rpath_script=conftest.$$.sh
+    rpath_outfile=conftest.$$.out
+    rm -f $rpath_script $rpath_outfile
+    cat > $rpath_script <<EOF
```
jsquyres: Ditto comment for this new code block as was given above.

config/opal_setup_wrappers.m4
```diff
@@ -172,7 +172,33 @@ EOF
 
     AS_IF([test -n "$rpath_args"],
           [WRAPPER_RPATH_SUPPORT=rpath
-           AC_MSG_RESULT([yes ($rpath_args)])],
+           cat > $rpath_script <<EOF
```
jsquyres: This whole new code block is a giant amount of replicated code -- it looks nearly identical to the block in earlier in `OPAL_SETUP_RPATH`.

This code block should likely be extracted into its own macro that both `OPAL_SETUP_RPATH` and `OPAL_SETUP_RUNPATH` can invoke.  Indeed, it should probably be parameterized so that we can use this one macro for all 3 cases: 1) `OPAL_SETUP_RPATH` C, 2) `OPAL_SETUP_RPATH` Fortran, and 3) `OPAL_SETUP_RUNPATH` Fortran.  That would make the code *much* more readable / maintainable.

I only mention this because it took me 15 minutes of closely reading to code to figure out why there were 3 copies of essentially the same Libtool-flag-extracting code (only *one* of them is for C, the other two are for Fortran, but that's only different by a single `--tag=FC` token in 2 of the 3 copies).

nncarlson:
This is important to some of us at LANL.  In the past we've not used the Fortran interface at all, preferring our own application-specific Fortran layer over MPI that interfaced with the C library, but recently we started moving in a direction that requires us to use the Fortran interface, and that has forced the issue of us getting open mpi to build correctly with the NAG compiler.  Modulo technical obstacles I'd be willing to help out by setting up some automated testing with the NAG compiler.

jsquyres:
@nncarlson Thank you, that would be great!  Can you sync up with @hppritcha to discuss internally how you'd like to setup some automated testing?  For compiler-specific stuff like this, we don't need a whole lot (i.e., we don't need to test all the deep-magic MPI layer stuff -- we basically need to verify that the compilation works and that basic MPI Fortran programs run correctly).

hppritcha:
@jsquyres I think we should take this for rc3.

jsquyres:
@ggouaillardet Will you promise to make the other 2 changes on master?  😄   If those changes don't make it for the release, that's no problem -- I just want to make sure that those changes at least get made on master so that we have a hope of maintaining this stuff in the future.

hppritcha:
assuming @ggouaillardet will address @jsquyres concerns on master, merging this for rc3.

PR #2098, opened 2016-09-21
v2.x: misc travis config update

ggouaillardet:


ggouaillardet:
@jsquyres first commit should have been back-ported earlier.
from today, glibtoolize does not work any more, the root cause is likely on travis-ci.org side.
i tried a workaround on master that is working fine and cannot really hurt us

PR #2099, opened 2016-09-21
v2.0.x: misc travis config update

ggouaillardet:


ggouaillardet:
similar to #2098 but for the ```v2.0.x``` branch

PR #2101, opened 2016-09-21
v2.1 mpool/rcache rewrite

hjelmn:
This PR reverts several spot fixes made to the v2.0.x release series then applies the mpool rewrite and the associated fixes. This PR should be complete but will need some testing to make sure I didn't miss anything.


hjelmn:
Looks like I missed one.

jsquyres:
Per discussion on the call today, we encourage the community to check out this PR:

* It's been on master for months (the number of commits reflect shaking out the bugs on master)
* It (greatly) simplifies mpool and rcache:
  * Mpool is now just a memory pool
  * Rcache is now just a registration cache
* We've talked about this in several face-to-face meetings

If this PR breaks things, it's likely because it's inadvertently missing some master commits.

This PR is the first in a two-step process: the next will be to bring over the MCA flag enumerator stuff, which will then enable a small number of other PRs to come over.

Issue #2106, opened 2016-09-22
v2.1.x: Open BSD 6.0 and patcher problems


jsquyres:
Filing a duplicate issue to #2049 for v2.1.0 (it's a blocker on v2.0.2, so it needs to be a blocker on v2.1.0, too).

-----

As reported by Paul H., Open BSD 6.0 disallows the patcher code's mprotect call.

See https://mail-archive.com/devel@lists.open-mpi.org/msg19527.html

hppritcha:
@hjelmn could you get this working on v2.x?

hppritcha:
Waiting on #2107 to get merged in.

jsquyres:
Per discussion on the 2017-02-21 webex, @hjelmn said he would try to get this in today.

hppritcha:
@jsquyres should we push this to 2.1.1?

jsquyres:
Agreed.  Pushed to v2.1.1.

PR #2107, opened 2016-09-22
patcher/overwrite: add runtime check for mprotect support

hjelmn:
This commit adds a runtime check to mca_patcher_overwrite_query to see if
mprotect (.., PROT_WRITE|...) works on a function address. This is needed
because on some platforms it is not possible to make a function page
writeable.

Signed-off-by: Nathan Hjelm hjelmn@me.com

### DIFF
--- a/opal/mca/patcher/overwrite/patcher_overwrite_component.c
+++ b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2016-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
@@ -12,6 +12,12 @@
 #include "patcher_overwrite.h"
 #include "opal/mca/mca.h"
 #include "opal/mca/base/base.h"
+#include "opal/util/sys_limits.h"
+
+#include <sys/mman.h>
+#if HAVE_SYS_SYSCTL_H
+#include <sys/sysctl.h>
+#endif
 
 static int mca_patcher_overwrite_priority;
 
@@ -28,6 +34,28 @@ static int mca_patcher_overwrite_register (void)
 
 static int mca_patcher_overwrite_query (mca_base_module_t **module, int *priority)
 {
+    unsigned long page_size = opal_getpagesize ();
+    int wxabort = 0;
+    int ret;
+
+#if defined(CTL_KERN) && defined(KERN_WXABORT)
+    (void) sysctl ((int []) {CTL_KERN, KERN_WXABORT}, 2, &wxabort, &(int) {sizeof (wxabort)}, NULL, 0);
+#endif
+    if (1 != wxabort) {
+        /* try to modify the protection on a function. if we can't change the protection we
+         * can't support the overwrite style of runtime patching. */
+        ret = mprotect ((void *)((intptr_t) mca_patcher_overwrite_query & ~(page_size - 1)),
+                        page_size, PROT_EXEC|PROT_READ|PROT_WRITE);
+    } else {
+        /* kern.wxabort is set. can not change memory protection */
+        ret = -1;
+    }
+
+    if (0 != ret) {
+        *priority = -1;
+        return OPAL_ERR_NOT_AVAILABLE;
+    }
+
     *module = &mca_patcher_overwrite_module.super;
     *priority = mca_patcher_overwrite_priority;
     return OPAL_SUCCESS;


hjelmn:
@PHHargrove This should fix running on OpenBSD.

jsquyres:
@hjelmn Please put "Fixes #2106" in your comment message or comment body.

PHHargrove:
@hjelmn 
I have not tested on OpenBSD yet, but suspect that this is NOT a fix.

The behavior I originally saw was that mprotect() called abort() when asked to make a page both executable and writable.  So, I suspect this change will just lead to an *earlier* abort rather than preventing it.  That is why I had originally suggested that the probe be done at configure time.

I *do* plan to confirm or refute my supposition about still aborting... stand by.

hjelmn:
@jsquyres I will add the fixes into the PRs for 2.0.x and 2.x.

hjelmn:
@PHHargrove Ok, I have a configure time version. Will update this PR to use that.

ggouaillardet:
@PHHargrove @hjelmn  i am trying to follow the logic here ...

if the behaviour is influenced by the kernel parameter ```kern.wxabort```, then should we instead
- at ```configure``` time, check the __existence__ of the ```kern.wxabort``` parameter (and ignore its value)
- at ```mpirun``` time, check the __value__ of the ```kern.wxabort``` parameter only if this parameter existence was detected at ```configure``` time. this check should be perform by both ```mpirun``` and ```orted```

makes sense ?

strictly speaking, we do not even need to check the existence at configure time, and we could only check both existence and value at run time. if we do that, OpenMPI built on OpenBSD 5.9 would work or fail with a nice error message if installed on OpenBSD 6.0.

PHHargrove:
Gilles's comments made me think this through a little better than I had.
I withdraw my previous suggestion that a configure-time check is useful/necessary.

The *ability* to mprotect() with both the W and X bits requires that the executable have a specific ELF header flag set, and the filesystem mounted with a special flag.
When those conditions are not met, the `kern.wxabort` sysctl determines **how** the mprotect() call fails: return failure or raise SIGABRT.

So, if `kern.wxabort==1` then a runtime probe of `mprotect()` could abort.
That is the "weakness" of Nathan's original fix.


I propose the following as "good enough":
```
    if (kern.wxabort is set)  DISABLE_PATCHER; // not safe to try mprotect()
    else if (0 > mprotect(...)) DISABLE_PATCHER; // mprotect() failed
    else ENABLE_PATCHER;
```

Basically, this comes down adding a sysctl() call before Nathan's runtime check to ensure the check is not attempted if the wxabort behavior is enabled.

The next step of sophistication might be to `oldhand = signal(SIGABRT, SIG_IGN)` before the runtime check of mprotect(), and `signal(SIGABRT, oldhand)` after.  Use of `sigaction()` is left as an exercise for the reader :-)

ggouaillardet:
@PHHargrove why not testing the ELF header flag and the mount options ?
To me, it looks smoother than "mprotect() and recover if it crashes"

Btw, whose ELF headers should be checked ? a.out ? Open MPI shared libs ? a.out and __all__ shared libs ?

PHHargrove:
@ggouaillardet asked
> why not testing the ELF header flag and the mount options ?

One answer would be "because I don't know how".
Another would be that *trying* the `mprotect()` will continue to work if the mount option or ELF header flag changes.

> Btw, whose ELF headers should be checked ? a.out ? Open MPI shared libs ? a.out and all shared libs ?

I don't know if the main executable or the dlopen()ed library is the one that matters - it might be both.
This falls under the previous response "because I don't know how".

ggouaillardet:
@PHHargrove got it

What if we trim your previous proposal like this

```
    if (kern.wxabort is set)  DISABLE_PATCHER; // not safe to try mprotect()
    else ENABLE_PATCHER;
```

Given that Open MPI would work out of the box on OpenBSD 6 and Open MPI is not widely used on OpenBSD, that still looks good enough to me

PHHargrove:
@ggouaillardet 

If you test "if (kern.wxabort exists)" then that would be sufficient.
Perhaps we are saying the same thing if you had interpreted "is set" as "exists" (though I had intended it as "has non-zero value", I admit I wasn't clear)

However, if take "is set" as "has non-zero value" and we abridge my proposal as you suggest then we are right back where we started, when I reported the problem on the devel list.  The mprotect() call will fail when patcher/overwrite runs.

ggouaillardet:
@PHHargrove I got it wrong indeed ... Thanks for the explanation
We should keep your initial proposal with three cases, in which "is set" means "exists and is non zero"

hjelmn:
@PHHargrove I updated the check based off your comments. Can you test?

PHHargrove:
Currently unable to test due to the following failure:
```
{openbsd-amd64 examples}$ mpirun -np 2 ./hello_c
[openbsd-amd64.my.domain:60925] [[20772,0],0] ORTE_ERROR_LOG: Not found in file ../../../../../ompi/orte/mca/ess/hnp/ess_hnp_module.c at line 629
Open MPI developer):

  opal_pmix_base_select failed
```

rhc54:
Must be some problem with updating the patch - the master is working fine for me

hjelmn:
Was updated to master this morning. Just updated it again to make sure it is in sync.

PHHargrove:
@rhc54 
You said "Must be some problem with updating the patch - the master is working fine for me"

Is master working for you on Open BSD 6?
That is where this testing is taking place.

I've just now tried last night's master tarball and the idential problem is also present on Open BSD 6.
This is an issue I first observed in late September and mentioned: [here](https://github.com/open-mpi/ompi/pull/2110#issuecomment-249101144)
At the time, Gilles indicated that `--disable-dlopen` got him past that error, and I will be trying that next.

-Paul

rhc54:
Ah - no, I have no way of testing against Open BSD. Let me know what you discover.

PHHargrove:
@rhc54 
With `--disable-dlopen` I now get a compile error in pmix on Open BSD 6.
I'll open an issue shortly.

PHHargrove:
I have opened issue #2533 for the pmix problem I mentioned in the previous comment, and I am working-around by manually editing the pmix_config.h file to define HAVE_NET_IF_H

Alas, I can run ring_c on Open BSD 6 only if I configure with `--disable-dlopen` and edit pmix_config.h as noted above.
However, since that also disables patcher:overwrite I am left unable to test this PR.

PHHargrove:
@hjelmn 
Not looking particularly promising:

```
Making all in mca/patcher/overwrite
  CC       patcher_overwrite_module.lo
patcher_overwrite_module.c: In function 'mca_patcher_overwrite_patch_address':
patcher_overwrite_module.c:255: error: implicit declaration of function 'mca_patcher_overwrite_apply_patch'
*** Error 1 in opal/mca/patcher/overwrite (Makefile:1907 'patcher_overwrite_module.lo': @echo "  CC      " patcher_overwrite_module.lo;depba...)
```

I don't find `sysctlbyname` in the manpages or anywhere in /usr/include.

-Paul

PHHargrove:
You probably want a normal sysctl() call with
```
    int mib[2] = { CTL_KERN, KERN_WXABORT };
```
and of course you need to `#include <sys/sysctl.h>`!


Additionally, we need to figure out the reason for the following:
```
$ grep OPAL_ASSEMBLY_ARCH opal/include/opal_config.h
#define OPAL_ASSEMBLY_ARCH OPAL_UNSUPPORTED
```
I think the issue there is `$host=amd64` where x86_64 is expected by opal_config_asm.m4

PHHargrove:
I don't see the OPAL_ASSEMBLY_ARCH issue with the master tarball.
I suspect I botched something merging Ralph's pmix header fixes with this PR. 

However, please consider the following as a step in the right direction:
```
--- a/opal/mca/patcher/overwrite/patcher_overwrite_component.c
+++ b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
@@ -15,6 +15,9 @@
 #include "opal/util/sys_limits.h"

 #include <sys/mman.h>
+#if HAVE_SYS_SYSCTL_H
+#include <sys/sysctl.h>
+#endif

 static int mca_patcher_overwrite_priority;

@@ -32,11 +35,14 @@ static int mca_patcher_overwrite_register (void)
 static int mca_patcher_overwrite_query (mca_base_module_t **module, int *priority)
 {
     unsigned long page_size = opal_getpagesize ();
-    size_t value_len = sizeof (int);
     int wxabort = 0;
     int ret;

-    (void) sysctlbyname ("kern.wxabort", &wxabort, &value_len, NULL, 0);
+#if defined(CTL_KERN) && defined(KERN_WXABORT)
+    int mib[2] = { CTL_KERN, KERN_WXABORT };
+    size_t value_len = sizeof (wxabort);
+    (void) sysctl(mib, 2, &wxabort, &value_len, NULL, 0);
+#endif

     if (1 != wxabort) {
         /* try to modify the protection on a function. if we can't change the protection we

```
Signed-off-by: Paul H. Hargrove <PHHargrove@lbl.gov>

hjelmn:
Ok, sysctlbyname must be a FreeBSD only call. Will fix now.

jsquyres:
In F2F Jan 2017 meeting, @hjelmn will finish this in the near future and PR to v2.x.

hjelmn:
@PHHargrove Please take a look and let me know if the latest version of this PR works for you.

PHHargrove:
@hjelmn I will try to get this tested tonight.
If not, bug me again on Thu (I am traveling on Wed).

PHHargrove:
Still something broken with respect to `OPAL_ASSEMBLY_ARCH`
```
  CC       patcher_overwrite_module.lo
patcher_overwrite_module.c: In function 'mca_patcher_overwrite_patch_address':
patcher_overwrite_module.c:255: error: implicit declaration of function 'mca_patcher_overwrite_apply_patch'

$ grep OPAL_ASSEMBLY_ARCH opal/include/opal_config.h
#define OPAL_ASSEMBLY_ARCH OPAL_UNSUPPORTED
```
See my fourth comment of Dec 6, 2016  re: "am64" vs "x86_64" in  opal_config_asm.m4

Manually fixing OPAL_ASSEMBLY_ARCH I can complete the compilation.


The next issue is that you are passing an int for the 4th arg to sysctl, while OpenBSD (and FreeBSD, NetBSD and Darwin) expects size_t:
```
  CC       patcher_overwrite_component.lo
patcher_overwrite_component.c: In function 'mca_patcher_overwrite_query':
patcher_overwrite_component.c:42: warning: passing argument 4 of 'sysctl' from incompatible pointer type
```

Manually fixing that gets me a clean build.
However, I still cannot get far enough to actually run any test that will run the code:
```
{openbsd-amd64 examples}$ mpirun -np 2 ./ring_c
[openbsd-amd64.my.domain:08476] [[40389,0],0] ORTE_ERROR_LOG: Not found in file ess_hnp_module.c at line 640
Open MPI developer):

  opal_pmix_base_select failed
```

rhc54:
@PHHargrove Can you take a look in your prefix/lib/openmpi directory and see what ```mca_pmix_*``` libraries are there? The error indicates that a required component didn't build. We'll likely need to see the config.log file from opal/mca/pmix/pmix112/pmix to determine why.

PHHargrove:
@rhc54 It looks like the "isolated" and "flux" components built:
```
{openbsd-amd64 openmpi}$ ls *pmi*
mca_ess_pmi.la        mca_pmix_flux.la      mca_pmix_isolated.la
mca_ess_pmi.so        mca_pmix_flux.so      mca_pmix_isolated.so
```
I am attaching [opal/mca/pmix/pmix2x/pmix/config.log](https://github.com/open-mpi/ompi/files/794047/config.log.txt) (w/ an added .txt suffix to make github happy).
The path you gave involving "pmix112" is not present in this PR's branch.

jjhursey:
It looks like PMIx `master` failed when checking for `dlopen`:
```
configure:26106: checking for dlopen in -ldl
configure:26131: gcc -o conftest -g  -g -I/home/phargrov/OMPI/PR2017 -I/home/phargrov/OMPI/PR2017 -I/home/phargrov/OMPI/PR2017/opal/include -I/home/phargrov/OMPI/PR2017/opal/include -I/home/phargrov/OMPI/PR2017/opal/mca/event/libevent2022/libevent -I/home/phargrov/OMPI/PR2017/opal/mca/event/libevent2022/libevent/include -I/home/phargrov/OMPI/PR2017/opal/mca/hwloc/hwloc1113/hwloc/include      conftest.c -ldl  -lm   -lz >&5
/usr/bin/ld: cannot find -ldl
collect2: ld returned 1 exit status
configure:26131: $? = 1
...
configure:26140: result: no
configure:27047: checking if MCA component pdl:pdlopen can compile
configure:27049: result: no
configure:27189: WARNING: Did not find a suitable static pmix pdl component
configure:27191: WARNING: You might need to install libltld (and its headers) or
configure:27193: WARNING: specify --disable-dlopen to configure.
configure:27195: error: Cannot continue
```

@rhc54 Maybe the `--disable-dlopen` from the top level OMPI configure didn't make it's way down to PMIx?

PHHargrove:
@jjhursey As I discovered/reported in December I *can* get past the PMIX issue if I `--disable-dlopen`.  However, at least in my testing back then, patcher:overwrite is not built when dlopen is disabled and there is nothing to test!

PHHargrove:
In the top-level config.log I see dl:dlopen is getting built:
```
configure:5946: --- MCA component dl:dlopen (m4 configuration macro, priority 80)
configure:127282: checking for MCA component dl:dlopen compile mode
configure:127285: result: static
configure:127392: checking dlfcn.h usability
configure:127392: gcc -std=gnu99 -c -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -pthread -I/home/phargrov/OMPI/PR2017/opal/mca/hwloc/hwloc1113/hwloc/include   conftest.c >&5
configure:127392: $? = 0  
configure:127392: result: yes
configure:127392: checking dlfcn.h presence
configure:127392: gcc -E -I/home/phargrov/OMPI/PR2017/opal/mca/hwloc/hwloc1113/hwloc/include   conftest.c
configure:127392: $? = 0  
configure:127392: result: yes
configure:127392: checking for dlfcn.h
configure:127392: result: yes
configure:127486: result: looking for library without search path
configure:127488: checking for library containing dlopen
configure:127519: gcc -std=gnu99 -o conftest -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -pthread -I/home/phargrov/OMPI/PR2017/opal/mca/hwloc/hwloc1113/hwloc/include      conftest.c  -lm -lutil  -lz  -lm -lutil  -lz  >&5
conftest.c:639: warning: function declaration isn't a prototype
conftest.c:642: warning: function declaration isn't a prototype
configure:127519: $? = 0
configure:127536: result: none required
configure:128664: checking if MCA component dl:dlopen can compile
configure:128666: result: yes
```

So, something may be wrong w/ the pmix probe for dlopen?

rhc54:
I found some problems and have a patch working thru CI on pmix master that I'll bring over to OMPI once completed. The dlopen support isn't in OMPI release branches, so this doesn't affect them.

jsquyres:
Where are we on this one?  We're rapidly running out of time.

rhc54:
I fixed and committed the PMIx disable-dlopen problem.

jsquyres:
@hjelmn @PHHargrove Can you guys test again?

PHHargrove:
I will attempt to cherry-pick Ralph's e86a0dbf39 to Nathan's PR branch, autogen and retest.

PHHargrove:
@rhc54 Still no luck:

OPAL finds dlopen (in libc) and pmix does not (configure output below).
This is with e86a0db cherry-picked to this PR's branch, so perhaps that is not enough?
I am doing the best I can with the (lack of) information from you and Nathan.

I am attaching the pmix [config.log](https://github.com/open-mpi/ompi/files/798130/config.log.txt)

Let me know what info you want to help pursue this.


```
+++ Configuring MCA framework dl
checking for no configure components in framework dl...
checking for m4 configure components in framework dl... dlopen, libltdl

--- MCA component dl:dlopen (m4 configuration macro, priority 80)
checking for MCA component dl:dlopen compile mode... static
checking dlfcn.h usability... yes
checking dlfcn.h presence... yes
checking for dlfcn.h... yes
looking for library without search path
checking for library containing dlopen... none required
checking if MCA component dl:dlopen can compile... yes

--- MCA component dl:libltdl (m4 configuration macro, priority 50)
checking for MCA component dl:libltdl compile mode... static
checking --with-libltdl value... simple ok (unspecified)
checking --with-libltdl-libdir value... simple ok (unspecified)
checking for libltdl dir... compiler default
checking for libltdl library dir... linker default
checking ltdl.h usability... no
checking ltdl.h presence... no
checking for ltdl.h... no
checking if MCA component dl:libltdl can compile... no
```
...
```
+++ Configuring MCA framework pdl
checking for no configure components in framework pdl...
checking for m4 configure components in framework pdl... pdlopen, plibltdl

--- MCA component pdl:pdlopen (m4 configuration macro, priority 80)
checking for MCA component pdl:pdlopen compile mode... static
checking dlfcn.h usability... yes
checking dlfcn.h presence... yes
checking for dlfcn.h... yes
looking for library without search path
checking for dlopen in -ldl... no
checking if MCA component pdl:pdlopen can compile... no

--- MCA component pdl:plibltdl (m4 configuration macro, priority 50)
checking for MCA component pdl:plibltdl compile mode... static
checking --with-plibltdl value... simple ok (unspecified) 
checking --with-plibltdl-libdir value... simple ok (unspecified)
checking for libltdl dir... compiler default
checking for libltdl library dir... linker default
checking ltdl.h usability... no
checking ltdl.h presence... no
checking for ltdl.h... no
checking if MCA component pdl:plibltdl can compile... no
configure: WARNING: Did not find a suitable static pmix pdl component
configure: WARNING: You might need to install libltld (and its headers) or
configure: WARNING: specify --disable-dlopen to configure.
configure: /bin/sh './configure' *failed* for opal/mca/pmix/pmix2x/pmix
```

ggouaillardet:
@PHHargrove can you please confirm your environment
- OpenBSD 6.0 on x86_64
- @hjelmn branch (patcher) based on ```master```
- cherry-pick of @rhc54 e86a0dbf39b9b8fde9c1618c90920adb875b816d
- the patch you posted on Dec 7
if this is correct, then can you please post your ompi ```configure``` command line ?

i will then try to investigate this on a VM

PHHargrove:
@ggouaillardet 

Set up as follows:
  + OpenBSD 6.0 on x86-64 (though they call it amd64)
  + @hjelmn branch WHERE EVER HE BASED IT: `git checkout f2bf1dcb2`
  + `git cherry-pick e86a0db`
  + Bootstrap on another system w/ required autotools + rsync to OpenBSD host
  + No *previously-posted* patches (such as from Dec)
  + The patches below to address concerns that I've brought up but Nathan has not addressed
  + `configure --prefix=... --enable-debug --disable-io-romio`

```
diff --git a/config/opal_config_asm.m4 b/config/opal_config_asm.m4
index e8bd193cd9..4b4b9fbfab 100644
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1008,7 +1008,7 @@ AC_DEFUN([OPAL_CONFIG_ASM],[
             OPAL_ASM_SUPPORT_64BIT=1
             OPAL_GCC_INLINE_ASSIGN='"xaddl %1,%0" : "=m"(ret), "+r"(negone) : "m"(ret)'
             ;;
-        i?86-*|x86_64*)
+        i?86-*|x86_64*|amd64)
             if test "$ac_cv_sizeof_long" = "4" ; then
                 opal_cv_asm_arch="IA32"
             else
diff --git a/opal/mca/patcher/overwrite/patcher_overwrite_component.c b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
index d1e4f7a2c7..b6ce12012d 100644
--- a/opal/mca/patcher/overwrite/patcher_overwrite_component.c
+++ b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
@@ -39,7 +39,7 @@ static int mca_patcher_overwrite_query (mca_base_module_t **module, int *priorit
     int ret;
 
 #if defined(CTL_KERN) && defined(KERN_WXABORT)
-    (void) sysctl ((int []) {CTL_KERN, KERN_WXABORT}, 2, &wxabort, &(int) {sizeof (wxabort)}, NULL, 0);
+    (void) sysctl ((int []) {CTL_KERN, KERN_WXABORT}, 2, &wxabort, &(size_t) {sizeof (wxabort)}, NULL, 0);
 #endif
     if (1 != wxabort) {
         /* try to modify the protection on a function. if we can't change the protection we
```
Signed-off-by: Paul H. Hargrove PHHargrove@lbl.gov

PHHargrove:
I am now trying Nathan's single commit rebased onto master (which already includes Ralph's commit for PMIx+dlopen).

rhc54:
@PHHargrove Can you please try the following patch:

```diff
diff --git a/opal/mca/pmix/pmix2x/pmix/config/pmix.m4 b/opal/mca/pmix/pmix2x/pmix/config/pmix.m4
index b06e57c..bde0572
--- a/opal/mca/pmix/pmix2x/pmix/config/pmix.m4
+++ b/opal/mca/pmix/pmix2x/pmix/config/pmix.m4
@@ -729,7 +729,6 @@ AC_DEFUN([PMIX_DEFINE_ARGS],[
         [AC_HELP_STRING([--enable-dlopen],
                         [Whether build should attempt to use dlopen (or
                          similar) to dynamically load components.
-                         Disabling dlopen implies --disable-pdl-dlopen
                          (default: enabled)])])
     AS_IF([test "$enable_dlopen" = "no"],
           [enable_mca_dso="no"
@@ -738,6 +737,8 @@ AC_DEFUN([PMIX_DEFINE_ARGS],[
            AC_MSG_RESULT([no])],
           [PMIX_ENABLE_DLOPEN_SUPPORT=1
            AC_MSG_RESULT([yes])])
+    AC_DEFINE_UNQUOTED(PMIX_ENABLE_DLOPEN_SUPPORT, $PMIX_ENABLE_DLOPEN_SUPPORT,
+                      [Whether we want to enable dlopen support])
 
     # Embedded mode, or standalone?
     AC_MSG_CHECKING([if embedded mode is enabled])
diff --git a/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/configure.m4 b/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/configure.m4
index 1de8f9b..938346c
--- a/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/configure.m4
+++ b/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/configure.m4
@@ -27,7 +27,7 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
     # (we still need to configure them all so that things like "make
     # dist" work", but we just want the MCA system to (artificially)
     # conclude that it can't build any of the components.
-    AS_IF([test $PMIX_ENABLE_DLOPEN_SUPPORT -eq 0],
+    AS_IF([test "$enable_dlopen" = "no"],
           [want_pdl=0], [want_pdl=1])
 
     MCA_CONFIGURE_FRAMEWORK([pdl], [$want_pdl])
@@ -35,7 +35,7 @@ AC_DEFUN([MCA_pmix_pdl_CONFIG],[
     # If we found no suitable static pdl component and dlopen support
     # was not specifically disabled, this is an error.
     AS_IF([test "$MCA_pmix_pdl_STATIC_COMPONENTS" = "" && \
-           test $PMIX_ENABLE_DLOPEN_SUPPORT -ne 0],
+           test "$enable_dlopen" != "no"],
           [AC_MSG_WARN([Did not find a suitable static pmix pdl component])
            AC_MSG_WARN([You might need to install libltld (and its headers) or])
            AC_MSG_WARN([specify --disable-dlopen to configure.])
diff --git a/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/pdlopen/configure.m4 b/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/pdlopen/configure.m4
index ea2328e..975e0da
--- a/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/pdlopen/configure.m4
+++ b/opal/mca/pmix/pmix2x/pmix/src/mca/pdl/pdlopen/configure.m4
@@ -44,13 +44,13 @@ AC_DEFUN([MCA_pmix_pdl_pdlopen_CONFIG],[
 
     dnl This is effectively a back-door for PMIX developers to
     dnl force the use of the libltdl pdl component.
-    AC_ARG_ENABLE([dlopen],
-        [AS_HELP_STRING([--disable-dlopen],
+    AC_ARG_ENABLE([dl-dlopen],
+        [AS_HELP_STRING([--disable-dl-dlopen],
                         [Disable the "dlopen" PDL component (and probably force the use of the "libltdl" PDL component).])
         ])
 
     pmix_pdl_pdlopen_happy=no
-    AS_IF([test "$enable_dlopen" != "no"],
+    AS_IF([test "$enable_dl_dlopen" != "no"],
           [PMIX_CHECK_PACKAGE([pmix_pdl_pdlopen],
               [dlfcn.h],
               [dl],
```

PHHargrove:
@rhc54 No change.
Still not finding dlopen, despite opal finding it:
```
--- MCA component pdl:pdlopen (m4 configuration macro, priority 80)
checking for MCA component pdl:pdlopen compile mode... static
checking dlfcn.h usability... yes
checking dlfcn.h presence... yes
checking for dlfcn.h... yes
looking for library without search path
checking for dlopen in -ldl... no
checking if MCA component pdl:pdlopen can compile... no

```

You do understand that I *want* dlopen support, right?
I am not passing a configure option to disable it (which is all I see your previous patch addressing).

ggouaillardet:
@rhc54 i will give this a try to

at first glance, what happens here is ```libdl.so``` is not available.
```ompi``` handles this nicely (e.g. automatically disable dlopen), whereas ```PMIx``` aborts because ```--disable-dlopen``` was not passed to the embedded ```PMIx```

so i guess we have two options
- update ```PMIx``` so it behaves like ```ompi``` (e.g. does __not__ abort but simply disable ```dlopen```)
- update the glue, so it passes ```--disable-dlopen``` to the embedded ```PMIx```

makes sense ?

**Comment by PHHargrove** [comment]
@ggouaillardet 

I feel like I am either talking to a brick wall or I am missing something important.

I am not passing `--disable-dl-open` nor `--disable-dlopen` or any other variant (nor do I want to).
This system *does* build "dl:dlopen" (as I have shown multiple times in quoted configure outputs).
However, "pdl:pdlopen" is not getting built.

On this system there is no `libdl.so` and dlopen() is instead found in `libc.so`.
Opal's configury for dl:dlopen handles this correctly.
**PLEASE** notice the "none required" in the next-to-last line below:
```
--- MCA component dl:dlopen (m4 configuration macro, priority 80)
checking for MCA component dl:dlopen compile mode... static
checking dlfcn.h usability... yes
checking dlfcn.h presence... yes
checking for dlfcn.h... yes
looking for library without search path
checking for library containing dlopen... none required
checking if MCA component dl:dlopen can compile... yes
```

**Comment by ggouaillardet** [comment]
@PHHargrove i wast just noticing this before i read your comment !
i will double check the difference between ```ompi``` and ```PMIx```
```PMIx``` should really be happy with ```none required```

**Comment by ggouaillardet** [comment]
@PHHargrove @rhc54 
```opal``` uses ```AC_SEARCH_LIBS``` whereas ```PMIx``` uses ```AC_CHECK_LIB```
i am now testing a fix

**Comment by PHHargrove** [comment]
Just to be clear, the configury for patcher:overwrite requires OPAL_ENABLE_DLOPEN_SUPPRORT.
So, disabling dlopen (at the OPAL layer) is not an option if I am to test this PR.

If there is a way to disable/circumvent the dlopen issues with PMIx I am open to trying them.  
However, I assume Ralph wants PMIx to work correctly on OpenBSD.

FWIW: libltdl lib and header are in /usr/local and I am trying `--with-libltdl=/usr/local` next.

**Comment by ggouaillardet** [comment]
you can try
```export ac_cv_lib_dl_dlopen=yes```
before running ```configure```
(disclaimer, i did not try)

**Comment by PHHargrove** [comment]
> FWIW: libltdl lib and header are in /usr/local and I am trying --with-libltdl=/usr/local next.

That is not working either and appears to be *yet another bug*.

Output of `configure --help` says:
```
  --with-libltdl(=DIR)    Build libltdl support, optionally adding
                          DIR/include, DIR/lib, and DIR/lib64 to the search
                          path for headers and libraries
```

However, when I configure using `--with-libltdl=/usr/local` I can clearly see that my `DIR` was *not* added to the search path for headers (no `-I/usr/local/include` in the gcc command below)
```
configure:5946: --- MCA component dl:libltdl (m4 configuration macro, priority 50)
configure:129251: checking for MCA component dl:libltdl compile mode
configure:129254: result: static
configure:129316: checking --with-libltdl value
configure:129336: result: sanity check ok (/usr/local)
configure:129348: checking --with-libltdl-libdir value
configure:129351: result: simple ok (unspecified)
configure:129396: checking for libltdl dir
configure:129398: result: /usr/local (from --with-libltdl)
configure:129400: checking for libltdl library dir
configure:129402: result: linker default
configure:129468: result: looking for header without includes
configure:129472: checking ltdl.h usability
configure:129472: gcc -std=gnu99 -c -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -pthread -I/home/phargrov/OMPI/PR2107/opal/mca/hwloc/hwloc1113/hwloc/include   conftest.c >&5
conftest.c:664:18: error: ltdl.h: No such file or directory
configure:129472: $? = 1

```

And before somebody asks, below are the answers to two obvious follow-up questions:

Q1: Is the header really there?
A1: YES, as shown below.
```
{openbsd-amd64 PR2107}$ ls -l /usr/local/include/ltdl.h
-r--r--r--  2 root  bin  5717 Aug  7  2015 /usr/local/include/ltdl.h
```

Q2: Isn't `/usr/local` in the default search path?
A2: No, the default search path contains only `/usr/include`, as shown below:
```
{openbsd-amd64 PR2107}$ gcc --verbose -c empty.c
Reading specs from /usr/lib/gcc-lib/amd64-unknown-openbsd6.0/4.2.1/specs
Target: amd64-unknown-openbsd6.0
Configured with: OpenBSD/amd64 system compiler
Thread model: posix
gcc version 4.2.1 20070719 
 /usr/lib/gcc-lib/amd64-unknown-openbsd6.0/4.2.1/cc1 -quiet -v empty.c -quiet -dumpbase empty.c -auxbase empty -version -o /tmp//cc1vXNP8.s
ignoring duplicate directory "/usr/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/include
End of search list.
GNU C version 4.2.1 20070719  (amd64-unknown-openbsd6.0)
        compiled by GNU C version 4.2.1 20070719 .
GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
Compiler executable checksum: f998539979469a547ba236dfb6cc1b01
 as -V -Qy -o empty.o /tmp//cc1vXNP8.s
GNU assembler version 2.17 (amd64-unknown-openbsd6.0) using BFD version 2.17
```

**Comment by rhc54** [comment]
@ggouaillardet Making PMIx handle this scenario like OMPI is fine - I appreciate the help as I am unable to replicate the problem on any machine to which I have access.

@PHHargrove I personally don't care about OpenBSD as I am unaware of anyone who uses it. I'm doing the best I can to shoot in the dark and help with the problem solely because disable-dlopen is important. However, it appears to be working everywhere else, so please stop being mad about it. 😄

**Comment by PHHargrove** [comment]
BTW:

Of the two patches I provided a few comments back, the first was incorrect (earlier version than the one I ended on).
The correct one:
```
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1008,7 +1008,7 @@ AC_DEFUN([OPAL_CONFIG_ASM],[
             OPAL_ASM_SUPPORT_64BIT=1
             OPAL_GCC_INLINE_ASSIGN='"xaddl %1,%0" : "=m"(ret), "+r"(negone) : "m"(ret)'
             ;;
-        i?86-*|x86_64*)
+        i?86-*|x86_64*|amd64-*)
             if test "$ac_cv_sizeof_long" = "4" ; then
                 opal_cv_asm_arch="IA32"
             else
```
Signed-off-by: Paul H. Hargrove PHHargrove@lbl.gov

**Comment by PHHargrove** [comment]
> you can try
> export ac_cv_lib_dl_dlopen=yes
> before running configure
> (disclaimer, i did not try)

Nope.

**Comment by PHHargrove** [comment]
@rhc54 

> @PHHargrove I personally don't care about OpenBSD as I am unaware of anyone who uses it. I'm doing the best I can to shoot in the dark and help with the problem solely because disable-dlopen is important. However, it appears to be working everywhere else, so please stop being mad about it. 😄

I am not mad at you, Nathan or Gilles - just frustrated by my inability to communicate my problems to you guys.  It is frustrating that I have had to repeat the same thing multiple times before anybody understood me. 

I don't personally care about OpenBSD specifically, but *do* care about helping OMPI (and its related projects such as hwloc and pmix) remain portable.  That means identifying non-portable assumptions such as "dlopen() will always be found in libdl".

FWIW: On Darwin dlopen is also in the standard library (libSystem.dylib), but since there is an (unrelated!) library named libdl.dynlib the logic in PMIx "just works" (but only by accident).

Please keep in mind that I do not write MPI code, I do not run MPI code, and I do not build or install MPI implementations for users.  I could stop testing Open MPI at any time if you guys really don't want me here and I would suffer no consequences other than feeling extremely awkward when drinking beer with you guys at SC each year.


FWIW: `CC='gcc -I/usr/local/include -L/usr/local/lib'` lets PMIx find libltdl in /usr/local even though `--with-plibltdl=` explicitly discards `/usr/local`.  So, I think I finally have a build in which I can test the *actual* code of this PR.

**Comment by PHHargrove** [comment]
> FWIW: CC='gcc -I/usr/local/include -L/usr/local/lib' lets PMIx find libltdl in /usr/local even though --with-plibltdl= explicitly discards /usr/local. So, I think I finally have a build in which I can test the actual code of this PR.

Unfortunately that leads to new problems (perhaps conflict w/ a system libevent, but I am not planning to dig).

Up next I am going to try to implement Gilles's `AC_CHECK_LIB` -> `AC_SEARCH_LIBS` change.

**Comment by ggouaillardet** [comment]
@PHHargrove here it is

it does compile, but i did not even try to run a test program yet.

i PR upstream PMIx at pmix/master#318

```patch
diff --git a/opal/mca/pmix/pmix2x/pmix/config/pmix_check_package.m4 b/opal/mca/pmix/pmix2x/pmix/config/pmix_check_package.m4
index f4833c3..babb542 100644
--- a/opal/mca/pmix/pmix2x/pmix/config/pmix_check_package.m4
+++ b/opal/mca/pmix/pmix2x/pmix/config/pmix_check_package.m4
@@ -13,6 +13,8 @@
 # Copyright (c) 2012-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2014      Intel, Inc. All rights reserved.
+# Copyright (c) 2017      Research Organization for Information Science
 # $COPYRIGHT$
 #
@@ -76,7 +78,7 @@ AC_DEFUN([_PMIX_CHECK_PACKAGE_LIB], [
           [ # libdir was specified - search only there
            $1_LDFLAGS="$$1_LDFLAGS -L$6"
            LDFLAGS="$LDFLAGS -L$6"
-           AC_CHECK_LIB([$2], [$3],
+           AC_SEARCH_LIBS([$3], [$2],
                         [pmix_check_package_lib_happy="yes"],
                         [pmix_check_package_lib_happy="no"], [$4])
            AS_IF([test "$pmix_check_package_lib_happy" = "no"],
@@ -88,7 +90,7 @@ AC_DEFUN([_PMIX_CHECK_PACKAGE_LIB], [
            AS_IF([test "$pmix_check_package_libdir" = "" || test "$pmix_check_package_libdir" = "/usr" || test "$pmix_check_package_libdir" = "/usr/local"],
                [ # try as is...
                 AC_VERBOSE([looking for library without search path])
-                AC_CHECK_LIB([$2], [$3],
+                AC_SEARCH_LIBS([$3], [$2],
                         [pmix_check_package_lib_happy="yes"],
                         [pmix_check_package_lib_happy="no"], [$4])
                 AS_IF([test "$pmix_check_package_lib_happy" = "no"],
@@ -102,7 +104,7 @@ AC_DEFUN([_PMIX_CHECK_PACKAGE_LIB], [
                     [$1_LDFLAGS="$$1_LDFLAGS -L$pmix_check_package_libdir/lib"
                      LDFLAGS="$LDFLAGS -L$pmix_check_package_libdir/lib"
                      AC_VERBOSE([looking for library in lib])
-                     AC_CHECK_LIB([$2], [$3],
+                     AC_SEARCH_LIBS([$3], [$2],
                                [pmix_check_package_lib_happy="yes"],
                                [pmix_check_package_lib_happy="no"], [$4])
                      AS_IF([test "$pmix_check_package_lib_happy" = "no"],
@@ -116,7 +118,7 @@ AC_DEFUN([_PMIX_CHECK_PACKAGE_LIB], [
                     [$1_LDFLAGS="$$1_LDFLAGS -L$pmix_check_package_libdir/lib64"
                      LDFLAGS="$LDFLAGS -L$pmix_check_package_libdir/lib64"
                      AC_VERBOSE([looking for library in lib64])
-                     AC_CHECK_LIB([$2], [$3],
+                     AC_SEARCH_LIBS([$3], [$2],
                                [pmix_check_package_lib_happy="yes"],
                                [pmix_check_package_lib_happy="no"], [$4])
                      AS_IF([test "$pmix_check_package_lib_happy" = "no"],
@@ -128,6 +130,14 @@ AC_DEFUN([_PMIX_CHECK_PACKAGE_LIB], [
     AS_IF([test "$pmix_check_package_lib_happy" = "yes"],
           [$1_LIBS="-l$2 $4"
            $7], [$8])
+    AS_IF([test "$pmix_check_package_lib_happy" = "yes"],
+          [ # The result of AC SEARCH_LIBS is cached in $ac_cv_search_[function]
+           AS_IF([test "$ac_cv_search_$3" != "no" &&
+                  test "$ac_cv_search_$3" != "none required"],
+                 [$1_LIBS="$ac_cv_search_$3 $4"],
+                 [$1_LIBS="$4"])
+           $7],
+          [$8])
 
     AS_VAR_POPDEF([pmix_Lib])dnl
 ])
diff --git a/opal/mca/pmix/pmix2x/pmix/src/mca/ptl/base/ptl_base_frame.c b/opal/mca/pmix/pmix2x/pmix/src/mca/ptl/base/ptl_base_frame.c
index 66477d9..00799c4 100644
--- a/opal/mca/pmix/pmix2x/pmix/src/mca/ptl/base/ptl_base_frame.c
+++ b/opal/mca/pmix/pmix2x/pmix/src/mca/ptl/base/ptl_base_frame.c
@@ -33,6 +33,9 @@
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
+#ifdef HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif
 
 #include "src/mca/mca.h"
 #include "src/mca/base/base.h"
```

ggouaillardet:
@PHHargrove @rhc54 @jsquyres the ```/usr/local``` thing is something which is not fully addressed.
here and there (but not everywhere), we assume the compiler will look for headers and libs in ```/usr/local/include``` and ```/usr/local/lib``` by default, and we silently refuse to ```-I/usr/local/include -L/usr/local/lib```, just like we do not ```-I/usr -L/usr/lib -L/usr/lib64```  not to pull some unwanted external libraries.
we were previously bitten by Oracle compilers that do not search in ```/usr/local``` by default, and same thing happens with GNU on ```OpenBSD 6```.

i do not know how to automatically detect whether ```/usr/local``` is searched or not, but we could ask the user to specify it via yet an other ```configure``` option.

@PHHargrove fwiw and in this case, a more correct configure cli is
(i am by no mean suggesting that will fix anything)
```
./configure CPPFLAGS=-I/usr/local/include LDFLAGS=-L/usr/local/lib
```

PHHargrove:
@ggouaillardet 
It is my understanding from the info pages that `AC_SEARCH_LIBS` has a side-effect of appending to `$LIBS`, which `AC_CHECK_LIB` will only do if no action-on-success was supplied.  So, if my reading is correct, your modifications *might* have unintended consequences.  I *do* see that `PMIX_CHECK_PACKAGE` does a save/restore of `$LIBS` (among others), so there might be nothing to worry about.

PHHargrove:
@ggouaillardet Your pmix/master#318 is at least sufficient for configure to determine that pdl:pdlopen can build.  I can now (at least) run `mpirun --oversubscribe -np 2 ./ring_c`.

@rhc54 Since I never was trying to get `--disable-dlopen` or `--disable-dl-open` to work on this platform, I honestly don't know if they do or do not work (either with or without the patches you supplied).  If you want me to follow-up on that, please let me know.

I will next proceed to finally testing Nathan's PR.

Here is the "recipe" to get where I am now:
1) Started with master@99ec16ed
2) Cherry-picked Nathan's one commit from this PR
3) Applied Gilles's pmix/master#318 (using the diff in an earlier comment)
4) Applied my 2 patches (given previously, and repeated below)

```
diff --git a/config/opal_config_asm.m4 b/config/opal_config_asm.m4
index e8bd193cd9..f505ea031a 100644
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1008,7 +1008,7 @@ AC_DEFUN([OPAL_CONFIG_ASM],[
             OPAL_ASM_SUPPORT_64BIT=1
             OPAL_GCC_INLINE_ASSIGN='"xaddl %1,%0" : "=m"(ret), "+r"(negone) : "m"(ret)'
             ;;
-        i?86-*|x86_64*)
+        i?86-*|x86_64*|amd64-*)
             if test "$ac_cv_sizeof_long" = "4" ; then
                 opal_cv_asm_arch="IA32"
             else
diff --git a/opal/mca/patcher/overwrite/patcher_overwrite_component.c b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
index d1e4f7a2c7..b6ce12012d 100644
--- a/opal/mca/patcher/overwrite/patcher_overwrite_component.c
+++ b/opal/mca/patcher/overwrite/patcher_overwrite_component.c
@@ -39,7 +39,7 @@ static int mca_patcher_overwrite_query (mca_base_module_t **module, int *priorit
     int ret;
 
 #if defined(CTL_KERN) && defined(KERN_WXABORT)
-    (void) sysctl ((int []) {CTL_KERN, KERN_WXABORT}, 2, &wxabort, &(int) {sizeof (wxabort)}, NULL, 0);
+    (void) sysctl ((int []) {CTL_KERN, KERN_WXABORT}, 2, &wxabort, &(size_t) {sizeof (wxabort)}, NULL, 0);
 #endif
     if (1 != wxabort) {
         /* try to modify the protection on a function. if we can't change the protection we
```
Signed-off-by: Paul H. Hargrove PHHargrove@lbl.gov

PHHargrove:
@hjelmn 
After a flurry of work with help from Gilles and Ralph, I can finally build with the single commit in this PR, and pass "make check" and run "ring_c".
However, it does not appear that your code is actually running as far as I can tell.
Use of `-mca patcher_base_verbose 100` does not produce any output and an abort() call I added to   mca_patcher_overwrite_query() makes it clear that function is not running.
So, suggestions what I need to do to actually *run* your code?

hjelmn:
Odd that the code isn't running. Can you try it again and post a gist with the config.log?

PHHargrove:
@hjelmn 

Requested gist: [here](https://gist.github.com/PHHargrove/5e7b3ca9c0ba1ff64e20d332740c8e94)

Could you please update the PR.  
Without the various fixes detailed in Feb I cannot build.

hppritcha:
@hjelmn could you please update or close this PR?  Its one of the oldest ones still open.

hppritcha:
bot:retest

hppritcha:
closing.  we can reopen if we have a need for it.

PR #2131, opened 2016-09-28
v2.0.x: pmix{cray,s1,s2}: make pmi_opcaddy_t class static

ggouaillardet:
theses three pmix components use the same class name,
declare it as static so Open MPI can be built with --disable-dlopen

Thanks Limin Gu for the report

(back-ported from commit 871ade9231c7f555a2ef6b38f27956b1ad7722dc)


hppritcha:
@jsquyres  good to go

PR #2136, opened 2016-09-29
v2.x: oob/usock: error if the unix socket path is too long

ggouaillardet:
This has been reported on OSX where $TMPDIR can be long,
so fail with a user-friendly error message instead of using
a silently truncated path

master does things differently (and better), so this is a one-off
commit for the v2 branch


ggouaillardet:
this is the ```v2.x``` equivalent for #2135

hppritcha:
reassigning to myself.

hppritcha:
@jsquyres this is ready to go

PR #2137, opened 2016-09-29
Fix opal_cleanup() destructor

ggouaillardet:


ggouaillardet:
@hjelmn @bosilca can you please review this ?

currently, a simple MPI hello world programs invokes ```MPI_Init()``` that sets ```opal_initialized``` to ```1```, and then ```MPI_Finalize()``` resets ```opal_initialized``` to ```0```.
consequently, when the ```OPAL``` destructor ```opal_cleanup()``` is invoked, ```opal_initialized``` is ```0``` and nothing occurs (e.g. ```opal_class_finalize()``` is not invoked).

this is fixed by the first commit.

the second commit is an optimization that is based on the assumption it is not possible to invoke ```MPI_T_init_thread()``` after ```MPI_Finalize()```
in this case, and if ```ompi_mpi_finalize()``` has reset ```opal_initialized``` to ```0```, then invoke ```opal_cleanup()``` (e.g. release all the memory instead of waiting for the ```OPAL``` destructor)

can you please comment on my assumption and both commits ?

bosilca:
I think 4c1cebf is good. The way we define the validity epoch for the classes, allows the code to work even if the epoch does not increase monotonically. Thus, it is safe to have multiple call to opal_class_finalize.

👍  for your second patch, it looks like a nice optimization for the most mainstream usage.

Btw, what is the external variable **initted** in ompi/mpi/tool/init_thread.c

hjelmn:
@ggouaillardet There was a lot of discussion around MPI_T_Init after finalize. It is a case we have to support. @bosilca is correct in pointing out that we have fixed the opal_class issue and we should be able to init/finalize the class system as many times as we like. It might be worth just putting the class finalize back into opal_finalize_util.

ggouaillardet:
@hjelmn got it, let's follow up in #2220

ggouaillardet:
This PR has been made obsolete by @2220, so i am closing it

PR #2140, opened 2016-09-30
Small style changes to libnbc

planetA:
I also have a question, why the code checks return value of ompi_datatype_get_extent? This function always returns 0. Can I just remove error checking here?


jsquyres:
@bosilca Can you respond?  Thanks!

lanl-ompi:

lanl-ompi:

lanl-ompi:

bosilca:
Why would you remove the error check ? The function returns an int and not a void, it makes sense to always check the return value (in case one day we change the return from opal_datatype_get_extent). You can add an OPAL_LIKELY/OPAL_UNLIKELY to hint the compiler if you think this is a performance hit.

planetA:
@bosilca OK. I see your point.

hppritcha:
discussed at OMPI F2F and decided to close.  @planetA can open a new PR if desired.

Issue #2151, opened 2016-10-03
'nonblocking3' BVT test fails


tjcw:
This is the second of the MPICH BVT tests which fails with today's 'git clone' of OMPI.
- https://gist.github.com/tjcw/530ba5bb593a0f480b08de89d0497dc6

It fails with the following message

``` shell
--------------------------------------------------------------------------
[1,0]<stderr>:[oc0436844531:11983] *** Process received signal ***
[1,0]<stderr>:[oc0436844531:11983] Signal: Segmentation fault (11)
[1,0]<stderr>:[oc0436844531:11983] Signal code:  (128)
[1,0]<stderr>:[oc0436844531:11983] Failing at address: (nil)
[1,1]<stderr>:[oc0436844531:11984] *** Process received signal ***
[1,1]<stderr>:[oc0436844531:11984] Signal: Segmentation fault (11)
[1,1]<stderr>:[oc0436844531:11984] Signal code:  (128)
[1,1]<stderr>:[oc0436844531:11984] Failing at address: (nil)
[1,0]<stderr>:[oc0436844531:11983] [ 0] /lib64/libpthread.so.0(+0xf100)[0x7f55dc43d100]
[1,0]<stderr>:[oc0436844531:11983] [ 1] /usr/local/lib/libmpi.so.0(PMPI_Ialltoallw+0x13e)[0x7f55dc6aadce]
[1,1]<stderr>:[oc0436844531:11984] [ 0] [1,0]<stderr>:[oc0436844531:11983] [ 2] nonblocking3[0x40383d]
[1,0]<stderr>:[oc0436844531:11983] [ 3] nonblocking3[0x403ab5]
[1,0]<stderr>:[oc0436844531:11983] [ 4] [1,1]<stderr>:/lib64/libpthread.so.0(+0xf100)[0x7f57c14f1100]
[1,1]<stderr>:[oc0436844531:11984] [ 1] [1,0]<stderr>:/lib64/libc.so.6(__libc_start_main+0xf5)[0x7f55dc08db15]
[1,0]<stderr>:[oc0436844531:11983] [ 5] nonblocking3[0x402bf9]
[1,0]<stderr>:[oc0436844531:11983] *** End of error message ***
[1,1]<stderr>:/usr/local/lib/libmpi.so.0(PMPI_Ialltoallw+0x13e)[0x7f57c175edce]
[1,1]<stderr>:[oc0436844531:11984] [ 2] nonblocking3[0x40383d]
[...]

PR #2154, opened 2016-10-04
non-blocking collectives: retain MPI_op and MPI_Datatype(s)

ggouaillardet:


ggouaillardet:

when a non blocking collective is invoked, __user__ operators and datatypes (e.g. ```MPI_Op``` and ```MPI_Datatype```) are ```OBJ_RETAIN```'ed, and then ```OBJ_RELEASE```'d when the request completes.

this solves issues when user op/datatype are free'd by the app __before__ a non-blocking collective completes. i previously made a prototype but at the ```coll/libnbc``` level. these patches do it at the ```ompi``` level.

there cannot be any issues with intrinsic operators nor predefined datatypes, so i did not bother retain/release them. my understanding is there is a plan to stop doing that for non blocking point-to-point (e.g. ```MPI_Isend```, ```MPI_Irecv```), so i think that is fine.

i implemented that by "inserting" a completion callback (```req_complete_cb```) in a non blocking request.
ideally, i guess that should be done in the ```req_free``` callback (or maybe in both ```req_complete_cb``` and ```req_free``` ?) to properly handle cancelled requests.
a pointer is required to release op/datatype(s), and currently, only ```req_complete_cb_data``` is available, so it was easier to fix ```req_complete_cb```

@jsquyres @hjelmn FYI

ibm-ompi:

jjhursey:
bot:ibm:retest

jjhursey:
@ggouaillardet Is this PR is supposed to fix Issue #2151?

ggouaillardet:
@jjhursey yes 
@bosilca this is the PR we briwfly discussed yesterday
Note i might rework it a bit after #2393 is merged

bosilca:
@ggouaillardet as we briefly discussed during SC, I think the best approach will be to stop doing any refcount inside the library and instead move them all up into the MPI layer. This will allow us to release/retain all the non-predefined types and ops only once per operation.

hppritcha:
moving to 2.0.3

ggouaillardet:
@bosilca this PR can be seen as a first step.

it handles ddt/ops for non blocking collectives at the MPI level.
next step is to move ddt handling for pt2pt to the MPI level

meanwhile, do you have any comments on this PR ?

bosilca:
This patch looks extremely big, a lot of code for a little benefit. I wonder why can't we simply add the datatype and op release code next to the calls to ```ompi_request_complete``` ?

Aren't we supposed to retain the datatype (ompi_coll_base_retain_datatypes) and op (ompi_coll_base_retain_op) before the call to the non-blocking functions themselves ?

ibm-ompi:

 Gist: https://gist.github.com/3a4218f168345072c128648ce107aeef

ibm-ompi:

 Gist: https://gist.github.com/2d5c1a28af517e6eef3a24c202f7cce9

ibm-ompi:

 Gist: https://gist.github.com/c7c9ec868e4ece21fece1f8e503c878c

ibm-ompi:

 Gist: https://gist.github.com/3b47b61d703cc97ac46275531114cc15

ibm-ompi:

 Gist: https://gist.github.com/e0d9e4795c9d80a35b550f2ed3ae84ea

ibm-ompi:

 Gist: https://gist.github.com/4f43a357ab707b7c8d2515f88a9029a6

ggouaillardet:
:bot:aws:retest

ggouaillardet:
@bosilca this is quite an old PR we discussed way before it was issued.
Currently, onesided operations retain/release the datatype in the `pml` component.
I initially wrote some code (PR ?) that retain/release the datatypes and operator in the `coll` component (e.g. `coll/libnbc`) but you argued it would be better do this for all components at the `ompi` level, so I ended up with this PR, that will work for components other than `coll/libnbc` but that is a bit heavy (e.g. we end up chaining two callbacks when completing a request). We might be able to avoid this chaining by moving some data into the `ompi_request_t` struct (and I wanted to avoid this in the first place)

jjhursey:
Someone asked me about the state of this PR today. Just checking in to see where things are standing? This seems to fix Issue #2151

ggouaillardet:
I am waiting for directions from @bosilca since he objected this PR.

ggouaillardet:
@jjhursey note I did not test this PR vs persistent non blocking collectives.

AboorvaDevarajan:

jsquyres:
@bosilca Thoughts on this PR?

bosilca:
Here are few questions and suggestions:
- Create a specialized class of collective requests that contains the retain_op_data in the request (save a memory allocation/deallocation)
- In ompi_coll_base_retain_datatypes_w do you really need to make a copy of the datatype arrays ? I could not find anything in the MPI standard that define the life-expectancy of the array argyments for the non-blocking calls. But it would make sense if they were expected to exists for the entire duration of the collective. Anyway, the answer to this question might change the implementation of ompi_coll_base_retain_datatypes_w.
- I can see why we want to have distinct functions for all possible combinations of ops and datatype. However, in these functions we only use the opal_object_t capabilities of the different objects, so we could reduce the number of callbacks to 2 (2 opal_objects and 1 array of opal_objects).

ggouaillardet:
Thanks @bosilca, I will update the PR based on your feedback.

The MPI standard is explicit about the life expectancy of the array arguments for the non blocking calls, and this makes things even easier as you suggested.
From MPI 3.1 chapter 5.12 page 197

> Once initiated, all associated send buffers and buffers associated with input arguments (such
> as arrays of counts, displacements, or datatypes in the vector versions of the collectives)
> should not be modified, and all associated receive buffers should not be accessed, until the
> collective operation completes.

bosilca:
Thanks @ggouaillardet, I've been looking for that sentence for hours.

ggouaillardet:
I pushed a new commit to address the first point

ggouaillardet:
@bosilca I addressed all your points. Can you please give this PR a final review ?
@jjhursey FYI

ompi/mca/coll/base/coll_base_util.c
```diff
@@ -103,3 +103,165 @@ int ompi_rounddown(int num, int factor)
...
+        request->data.op.op = op;
+        retain = true;
+    }
+    if (!ompi_datatype_is_predefined(type)) {
+        OBJ_RETAIN(type);
+        request->data.op.datatype = type;
+        retain = true;
+    }
+    if (OPAL_UNLIKELY(retain)) {
+        if (req->req_persistent) {
```
bosilca: This is a good catch. Please add a detailed comment on why you need to trap either the complete or the free callback.

ompi/mca/coll/base/coll_base_util.c
```diff
@@ -103,3 +103,165 @@ int ompi_rounddown(int num, int factor)
...
+            req->req_complete_cb = complete_objs_callback;
+            req->req_complete_cb_data = request;
+        }
+    }
+    return OMPI_SUCCESS;
+}
+
+static void release_vecs_callback(ompi_coll_base_nbc_request_t *request) {
+    ompi_communicator_t *comm = request->super.req_mpi_object.comm;
+    int count = OMPI_COMM_IS_INTER(comm)?ompi_comm_remote_size(comm):ompi_comm_size(comm);
```
bosilca: How does this work for persistent neighbor collectives ?

ompi/mca/coll/libnbc/nbc.c
```diff
@@ -655,15 +655,15 @@ int NBC_Start(NBC_Handle *handle) {
...
+  handle->super.super.req_state = OMPI_REQUEST_ACTIVE;
+  handle->super.super.req_status.MPI_ERROR = OMPI_SUCCESS;
   res = NBC_Start_round(handle);
   if (OPAL_UNLIKELY(OMPI_SUCCESS != res)) {
     return res;
   }
 
   OPAL_THREAD_LOCK(&mca_coll_libnbc_component.lock);
-  opal_list_append(&mca_coll_libnbc_component.active_requests, &(handle->super.super.super));
+  opal_list_append(&mca_coll_libnbc_component.active_requests, &(handle->super.super.super.super));
```
bosilca: Oh my, this is getting ugly ;) It would be nicer if we simply cast it to opal_list_item_t directly.

ompi/mca/coll/retain.diff
```diff
@@ -0,0 +1,179 @@
+diff --git a/ompi/mca/coll/base/coll_base_util.h b/ompi/mca/coll/base/coll_base_util.h
```
bosilca: I don't think this should be part of the merge.

gpaulsen:
Two questions:
1) Is this at all dependent on https://github.com/open-mpi/ompi/pull/6695 ?
2) How difficult would it be to port this to v4.0.x?

ggouaillardet:
@gpaulsen 
 - this PR and #6695 are independent
 - I do not expect any difficulty backporting this to the `v4.0.x` branch, but I'll see when I get there.

ompi/mca/coll/base/coll_base_util.c
ggouaillardet: I guess you meant non blocking neighbor alltoallw (regardless this is a persistent operation or not), right ? in this case, this is a good catch I fixed in the latest commit.

ggouaillardet:
@bosilca I made the requested changes. Unless some changes are requested, I will squash the commits and merge this PR tomorrow.

ggouaillardet:
@gpaulsen fwiw, this PR applies nicely on the `v4.0.x` branch once 65660e5999fb7ed422536b79e1092520bed96694 has been backported.

gpaulsen:
@ggouaillardet @hppritcha, This functionality (along with https://github.com/open-mpi/ompi/commit/65660e5999fb7ed422536b79e1092520bed96694) was never backported to either the v4.0.x, or v4.1.x release streams.

My opinion is that this is quite a bit of change for this late in the release streams, and that we should probably wait for v5.0.0 for this functionality.  What are your thoughts?

PR #2155, opened 2016-10-04
use xcode8

ggouaillardet:


hjelmn:
@ggouaillardet There is a problem with using brew upgrade here. If a package is already the newest version then $? == 1. I work around it in another project by calling true immediately after brew upgrade. See https://github.com/libusb/libusb/blob/master/.travis.yml#L46

hjelmn:
hmm, you changed it to brew install. i think that will just say something is installed even if it isn't the newest version.

ggouaillardet:
@hjelmn i will update the master ```.travis.yml``` and ignore ```brew upgrade``` failure.

```brew upgrade gcc``` failed because ```gcc``` was not installed on the ```xcode8``` travis image.
that being said, that error occurs just before github.com yesterday major outage (that lasted less than half an hour) so i could try this again.

i google'd ```xcode8``` vs ```clock_gettime``` and it seems the error only occurs on ```El Capitan``` with ```xcode8```
some say ```xcode8``` is a really a cross-compiler for ```Sierra```, autotools detect ```clock_gettime``` is present even on ```El Capitan``` but that will fail at runtime.
according to travis doc, ```xcode8``` image is ```El Capitan``` but i could not reproduce the issue
(e.g. ```clock_gettime``` is not found by ```configure```)

bottom line, i am not sure what the issue really is, and i cannot reproduce it
(fwiw, my laptop is still running ```Yosemite```)

lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/0a114f3ffb880d9734b43641916de1d8

ibm-ompi:

 Gist: https://gist.github.com/56e2204c024dff1f75deb9b8d9be1310

ggouaillardet:
we do not use travis for OSX testing any more

PR #2157, opened 2016-10-04
v2.1.0: usNIC bug fixes

jsquyres:
Especially with regards to the forthcoming Libfabric v2.1.0.

@bturrubiates This is the v2.1.0 version of #2152.

PR #2169, opened 2016-10-05
v2.0.2: usnic bug fixes

jsquyres:
@bturrubiates This is the same as #2157, but with several more bug fix commits that hadn't yet come over from master --> v2.0.x (but were already on v2.1.x).  You should probably review individual commits; not the union of all commits.

A few commits had to be back-ported here (i.e., to the v2.0.x branch) because the usNIC BTL on master/v2.1.x support MPI_THREAD_MULTIPLE, but the usNIC BTL on v2.0.x does not.


jsquyres:
@hppritcha Good to go.

Issue #2150, opened 2016-10-03
'nonblocking' BVT test fails


tjcw:
I'm trying to run the MPICH BVT tests ( available as part of the MPICH source code ) under OpenMPI, and I find 2 of them fail. One of them is 'nonblocking.c'
  - https://gist.github.com/tjcw/530ba5bb593a0f480b08de89d0497dc6
 which fails with today's git clone of OMPI with the following message

```shell
[1,0]<stderr>:before calling Ialltoall, args NUM_INTS 2 sbuf 0xa75c80 rbuf 0xa75ca0
[1,0]<stderr>:First Ialltoall is complete
[1,1]<stderr>:before calling Ialltoall, args NUM_INTS 2 sbuf 0x1b5aa20 rbuf 0x1b5dc90
[1,1]<stderr>:First Ialltoall is complete
[oc0436844531:10615] *** An error occurred in MPI_Ialltoall
[oc0436844531:10615] *** reported by process [674365441,0]
[oc0436844531:10615] *** on communicator MPI_COMM_WORLD
[oc0436844531:10615] *** MPI_ERR_INTERN: internal error
[oc0436844531:10615] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[oc0436844531:10615] ***    and potentially your MPI job)
[oc0436844531.ibm.com:10606] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[oc0436844531.ibm.com:10606] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
[tjcw@oc0436844531 coll]$ 
```

tjcw:
[nonblocking.zip](https://github.com/open-mpi/ompi/files/506355/nonblocking.zip)
GIST here [https://gist.github.com/tjcw/3959abbe7407febc9a9df59a51957c4a](url)

ggouaillardet:
can you please double check your source file ? or post a link to mpich git
i do not see any MPI_Ialltoall in the gist nor mpich nonblocking.c nor nonblicking3.c

tjcw:
I have corrected the gist link in my comment above; I think it is right now. The failing line is the MPI_Ialltoall at line 143 : 
fprintf(stderr, "First Ialltoall is complete\n") ; 
    MPI_Ialltoall(MPI_IN_PLACE, -1, MPI_DATATYPE_NULL, rbuf, NUM_INTS, MPI_INT, comm, &req);
[...]

PR #2172, opened 2016-10-05
support MPI_IN_PLACE in MPI_Ialltoall*

ggouaillardet:
Fixes open-mpi/ompi#2150

### DIFF
--- a/ompi/mca/coll/libnbc/nbc_ialltoall.c
+++ b/ompi/mca/coll/libnbc/nbc_ialltoall.c
@@ -8,7 +8,7 @@
  * Copyright (c) 2013-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2014      NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2014-2015 Research Organization for Information Science
+ * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  *
  * Author(s): Torsten Hoefler <htor@cs.indiana.edu>
@@ -25,6 +25,8 @@ static inline int a2a_sched_pairwise(int rank, int p, MPI_Aint sndext, MPI_Aint 
 static inline int a2a_sched_diss(int rank, int p, MPI_Aint sndext, MPI_Aint rcvext, NBC_Schedule* schedule,
                                  const void* sendbuf, int sendcount, MPI_Datatype sendtype, void* recvbuf,
                                  int recvcount, MPI_Datatype recvtype, MPI_Comm comm, NBC_Handle *handle);
+static inline int a2a_sched_inplace(int rank, int p, NBC_Schedule* schedule, void* buf, int count,
+                                   MPI_Datatype type, MPI_Aint ext, ptrdiff_t gap, MPI_Comm comm);
 
 #ifdef NBC_CACHE_SCHEDULE
 /* tree comparison function for schedule cache */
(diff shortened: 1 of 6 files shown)


ggouaillardet:
Refs #2150

bosilca:
@ggouaillardet I am not sure I understand how the scheduling in libnbc works, but for scalability reasons in the a2a_sched_inplace function I would have round-robin starting from the current rank instead of starting every process from 0.

ggouaillardet:
@bosilca 
current implementation clearly does *not* scale from a performance point of view,
but from a memory usage point of view, it only requires one buffer, regardless the communicator size.

the pseudo blocking equivalent algo is
```c
for (int r = 0; r < p; r++) {
    MPI_Isend(from user buffer, to rank r);
    MPI_Irecv(into a temporary, from rank r);
   MPI_Waitall(2);
   MPI_Sendrecv(from temporary buffer, into user buffer, with myself);
}
```
so starting from the current rank, or doing something similar to pairwise algo will either deadlock or require more temporary buffers.

an other option is to mimick the ```ompi_coll_base_alltoall_intra_basic_linear``` algo
(a comment in the code states it is slow and should be replaced)

yet an other option is to
```c
for (int r = 1; r < p; r++) {
    peer = myrank ^ r;
    MPI_Isend(from user buffer, to peer);
    MPI_Irecv(into a temporary, from peer);
   MPI_Waitall(2);
   MPI_Sendrecv(from temporary buffer, into user buffer, with myself);
}
```
that works fine when communicator size is a power of 2, additional steps are required otherwise.

note that in any cases, all these algos require all peers communicate with each other, and establishing all connections can take a while at scale.

any thoughts ?

ggouaillardet:
@bosilca 

i refactored ```nbc_ialltoall*``` and mimicked ```ompi_coll_base_alltoall_intra_basic_linear```
i left the previous version for reference, i might remove it before merging

bosilca:
No doubt that all connections would be made, but we can at least try to spread the load, and prevent all ranks from hammering the same rank. Repetitively.

If you do a version of the pairwise, where you receive from r - step and send to r + step, and then you do the inverse (receive from r + step and send to r - step), you obtain a nicely spread shift of data between the ranks.

ggouaillardet:
@bosilca i just pushed my latest changes (i realized i pushed none of them yesterday)

If i implement the pairwise version you describe, that basically means i have to allocate and copy about half the user buffer, and this is what i want to avoid, for scalability reason.

to keep it simple, let's consider ```MPI_Ialltoall``` of n ```MPI_BYTE``` on a communicator of size ```p```
the data to be transferred is ```n * (p-1)``` bytes
my initial implementation allocates a temporary buffer of ```n``` bytes
the pairwise version requires a temporary buffer of ```n * p / 2``` bytes

i finally implemented a "xor-based" version for ```MPI_IN_PLACE``` and left other and unused versions for reference

bosilca:
@ggouaillardet you would have to allocate only the largest buffer sent or received. Still pretty scalable.

Let's take a case where at step i a rank will exchange data with (r+i and r-i). This is divided in 2 parts send to r+i and receive from r-i (in a temporary), and then send to r-i and receive from r+i (in the real position because the data has been already sent to r+i right before). Then it is enough to copy the temporary into the location of the r-i and the step is done in place.

ggouaillardet:
@bosilca that is clever and i will implement it
thanks for the guidance !

ggouaillardet:
@bosilca this PR should be ready for primetime now,
would you mind having a look at it first ?

PR #2173, opened 2016-10-05
v2.x: travis: cope with brew upgrade failure

ggouaillardet:
also try to brew install gcc before brew upgrade gcc

(cherry picked from commit ebb43c59d944c6ba949df3f02de1495a804bc4f9)


ggouaillardet:
similar to #2174

jsquyres:
@hppritcha I'm letting this in immediately so that we can get PRs get through Travis.

PR #2177, opened 2016-10-05
v2.x ppc fixes

hjelmn:
Assembly workarounds for PGI.

PR #2180, opened 2016-10-05
v2.x asm updates

hjelmn:
This PR brings over most of the assembly changes from master. This includes:

 - ARMv8 support
 - __sync builtin support (this is really desirable since the __builtin atomics are deprecated in gcc 5.x)
 - Allow PGI to use the inline assembly. We work around the bugs for ppc in 16.x so no need to have it disabled.
 - Add checks that builtin atomics support 64-bit values. This fixes a bug on some systems when using --enable-builtin-atomics.

This PR *does not* enable builtin atomics by default.

@artpol84 I ran performance tests before and after these commits and see no performance regression. You should probably do the same just to be sure.

### DIFF
--- a/opal/include/opal/sys/arm64/timer.h
+++ b/opal/include/opal/sys/arm64/timer.h
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2008      The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#ifndef OPAL_SYS_ARCH_TIMER_H
+#define OPAL_SYS_ARCH_TIMER_H 1
+
+#include <sys/times.h>
+
+typedef uint64_t opal_timer_t;
+
+static inline opal_timer_t
+opal_sys_timer_get_cycles(void)
+{
+    opal_timer_t ret;
+    struct tms accurate_clock;
+
+    times(&accurate_clock);
+    ret = accurate_clock.tms_utime + accurate_clock.tms_stime;
+
+    return ret;
+}
+
+#define OPAL_HAVE_SYS_TIMER_GET_CYCLES 1
+
+#endif /* ! OPAL_SYS_ARCH_TIMER_H */
--- a/opal/include/opal/sys/arm64/timer.h
+++ b/opal/include/opal/sys/arm64/timer.h
@@ -1,8 +1,11 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2008      The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2016      Broadcom Limited. All rights reserved.
+ * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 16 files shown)


hppritcha:
Is commit 19d3884 (add timer support) really related to this PR?

hjelmn:
@hppritcha It is related to the ARMv8 stuff but I can remove it if it is a problem.

opal/include/opal/sys/arm64/timer.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <sys/times.h>
+
+typedef uint64_t opal_timer_t;
+
+static inline opal_timer_t
+opal_sys_timer_get_cycles(void)
+{
+    opal_timer_t ret;
+
```
hjelmn: Not sure. @abjoshi-brcm What do you think? I can leave this commit out of this PR to be safe if we don't have consensus.

hjelmn:
@hppritcha I am pulling the PGI inline asm out of this one and putting it into #2177. That way if this PR fails we still have PGI on ppc fixed.

opal/include/opal/sys/arm64/timer.h
shamisp: @hjelmn - we need good timer (as part of this PR or separate one).

bosilca:
The arm64 timer should be monotonic isn't it?

shamisp:
correct.

opal/include/opal/sys/arm64/timer.h
abjoshi-brcm: @shamisp  
You are right, reads can occur out of order and so it should be
isb(); asm volatile("mrs %0, cntvct_el0" : "=r" (ret));  
And yes, this timer increments at a fixed frequency.

shamisp: @bosilca - AFAIK this is the timer that is invoked in opal_progress (after your fix). Do we use it in some other parts of the code ?

jsquyres:
@hjelmn Is this complete?

hjelmn:
@jsquyres Yes. Though it sounds like there is a desire from ARM to have the timer come over as well. I can either re-add it or file a separate PR.

shamisp:
Yep, we need the timer.

bosilca:
This is not yet complete. Based on the discussions above there are 2 things missing the isb and the MONOTONIC define.

hjelmn:
@shamisp Can you make the changes @bosilca mentions and PR them to master. I will add all the timer commits if that can be done by the end of the day tomorrow.

hjelmn:
Hmm, it might be faster for me to just go ahead and make the changes on master. I will have @shamisp review.

hjelmn:
@bosilca Looks like a MONOTONIC define isn't required as we always set it to 1. Non-monotonic timers have to undef it and redefine to 0.

bosilca:
@hjelmn indeed thanks for pointing this out.

shamisp:
@hjelmn can you please forward the original patch (timer) that you have been trying to merge ?

opal/include/opal/sys/arm64/timer.h
shamisp: @abjoshi-brcm FYI, I did some additional checks, which confirmed that isb() is required.

hjelmn:
@bosilca, @shamisp The ARM64 timer is back in the PR. Please review.

shamisp:
@hjelmn 👍

jsquyres:
@shamisp Can you give an actual github review?  That's what we're using these days -- not just the 👍

hppritcha:
@jsquyres I think this is ready to go

PR #2181, opened 2016-10-05
Lazy wait v2.x

artpol84:
@jladd-mlnx @jsquyres @hppritcha this recent fix is important for SLURM.

Needs to be rebased once #2176 is merged.

### DIFF
--- a/ompi/runtime/ompi_mpi_init.c
+++ b/ompi/runtime/ompi_mpi_init.c
@@ -639,7 +639,7 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided)
         if (NULL != opal_pmix.fence_nb) {
             opal_pmix.fence_nb(NULL, opal_pmix_collect_all_data,
                                fence_release, (void*)&active);
-            OMPI_WAIT_FOR_COMPLETION(active);
+            OMPI_LAZY_WAIT_FOR_COMPLETION(active);
         } else {
             opal_pmix.fence(NULL, opal_pmix_collect_all_data);
         }
@@ -809,7 +809,7 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided)
     if (NULL != opal_pmix.fence_nb) {
         opal_pmix.fence_nb(NULL, opal_pmix_collect_all_data,
                            fence_release, (void*)&active);
-        OMPI_WAIT_FOR_COMPLETION(active);
+        OMPI_LAZY_WAIT_FOR_COMPLETION(active);
     } else {
         opal_pmix.fence(NULL, opal_pmix_collect_all_data);
     }


artpol84:
@jladd-mlnx @jsquyres @hppritcha this recent fix is important for SLURM.

I removed dependency of this PR from #2176. Please review. We definitely want this in v2.1.0, while I'm not sure we'll have the time to provide all the info requested in #2176.

rhc54:
one question: is there ever a reason not to do the lazy wait? I'm wondering if we really need another parameter here, or just change the default. Note that changing this is no way impacts the speed of the underlying fence operation as that takes place in the daemons.

artpol84:
@rhc54,
If I understood you question correctly then I'm confused. It was you who requested this parameter originally when I was PRing to master:
https://github.com/open-mpi/ompi/pull/2089
(here in particular https://github.com/open-mpi/ompi/pull/2089#issuecomment-248148281)

jladd-mlnx:
https://github.com/open-mpi/ompi/pull/2089

rhc54:
I asked for it in master since it was an experimental feature. What I want to know _before_ we bring it to a release branch is the result of those tests. Did people find that it made a difference in general? Is it a detriment to anyone? Etc.

We don't just move experimental things to release branches without at least first discussing the results of the experiments. That's how we get into trouble.

jladd-mlnx:
@rhc54 it should be the default.

jladd-mlnx:
@rhc54 Has Intel run any experiments here? If so, could you share the results?

rhc54:
@matcabral Have you tested this? IIRC, the psm MTL did some progressing during MPI_Init, but I don't know if it matters how aggressively we do it.

@jsquyres How about usnic?

@hjelmn How about ugni and vader? Anything they need to do?

jsquyres:
Can someone explain what this PR is about?  (I don't know offhand if usNIC needs to do anything for this PR)

hppritcha:
Is this specific for direct launch using srun?

jladd-mlnx:
It is not specific to using a SLURM direct launch, mpirun launches could
benefit as well. Any runtime agent that uses shared CPU resources to
progress its operations with the application could benefit. The question is
what's a reasonable heuristic for time sharing the CPU between runtime and
application; this is both transport dependent AND runtime dependent (e.g.
if a runtime implements true offloaded nonblocking collectives, then it's
reasonable to allocate a larger fraction of the CPU resources to the
application.)

hppritcha:
Consensus seems to be to do more testing in master with lazy wait enabled by default per dev con call 10/18/16.  So park this PR and see if it can be reworked to not include the new mca parameter.

jsquyres:
Possibly just need to add 7910aa23eb6e365a9d97d52ac1f97379a25cf5e1 to this PR.

matcabral:
@rhc54 , I have not played with this and the PSM2 MTL

rhc54:
@matcabral We have now turned it "on" by default on master - can you just run a quick smoke test on it? I don't believe it will impact anything, but worth a quick check before it comes to the 2.x series

matcabral:
Apologize the delayed answer. I finally got to run some osu_ tests and saw no impact in comparison to 2.0.1

artpol84:
@jsquyres , done with signed-off

rhc54:
@artpol84 I have confirmed that you can remove the parameters and just hard-code use of the lazy wait. Would you please update this PR?

jsquyres:
@rhc54 mentioned on the call today that he's checked with everyone, and it seems like no transports are adversely affected by using the lazy wait.  Hence, this PR can likely be changed to just always use the lazy wait -- i.e., no need for an MCA param.

artpol84:
@rhc54 @jsquyres This will force unsync with master that currently has this option. Is this fine?

rhc54:
Ah, I was expecting that you would update master, and then PR the appropriate changes back to the release branch.

artpol84:
So we accept this one and then do a separate PR which removes the MCA param?

rhc54:
No, I'd just close this one out, update master, and then create a new PR that does the right thing.

jladd-mlnx:
In my opinion, the MCA parameter should be removed in master as well.

Josh

rhc54:
definitely

artpol84:
Ok, I'll do that

artpol84:
Closed per discussion above

PR #2200, opened 2016-10-10
v2.x: osc-portals4: fix datatype errors in put()

tkordenbrock:
Instead of ompi_datatype_get_extent(), use ompi_datatype_get_true_extent()
to get the origin and target lower bound.  For derived types like
subarray, true_lb is the correct offset for RDMA operations.  Also,
instead of the extent use the size of the datatype.

(cherry picked from open-mpi/ompi@cc863ff9fb7130cf3cd3d4a8b903b20af66848ca)


hppritcha:
@jsquyres ready to merge

PR #2207, opened 2016-10-11
Update PMIx support to latest PMIx master

rhc54:
Enable PMIx shared memory support by default


lanl-ompi:
Test FAILed.

ibm-ompi:

 Gist: https://gist.github.com/b70bd8b42b52a56c92f59c69c2ca31fa

lanl-ompi:
Test FAILed.

rhc54:
The failure shown here has nothing to do with this PR. Someone in the code base has created a link to a library that isn't being included properly when configured with ```--disable-dlopen```. The signature of the failure is:

```shell
$ orterun -n 1 -host rhc001 hostname
orterun: route/tc.c:973: rtnl_tc_register: Assertion `0' failed.

orterun:89564 terminated with signal 6 at PC=7fe6a8acf5f7 SP=7fffdf0d04d8.  Backtrace:
/usr/lib64/libc.so.6(gsignal+0x37)[0x7fe6a8acf5f7]
/usr/lib64/libc.so.6(abort+0x148)[0x7fe6a8ad0ce8]
/usr/lib64/libc.so.6(+0x2e566)[0x7fe6a8ac8566]
/usr/lib64/libc.so.6(+0x2e612)[0x7fe6a8ac8612]
/usr/lib64/libnl-route-3.so.200(+0x21649)[0x7fe6a8439649]
/lib64/ld-linux-x86-64.so.2(+0xf3a3)[0x7fe6aa7c23a3]
/lib64/ld-linux-x86-64.so.2(+0x146a)[0x7fe6aa7b446a]
```

I've tried to trace it down, but without success. Does anyone recognize where this came from?

ggouaillardet:
@rhc54 at first glance, there are two issues here
1. ```--disable-dlopen``` is not correctly supported by embedded pmix, i fixed that in pmix/master#181
   (currently, all MCA modules are built as dynamic libs that cannot be open)
2. ```route/tc.c:973: rtnl_tc_register: Assertion `0' failed``` typically occurs when Open MPI uses both ```libnl``` and ```libnl3``` (@jsquyres and i left the related PR on the back burner, and i will revive it from now)

ggouaillardet:
fwiw, i ran into the ```libnl``` vs ```libnl3``` issue on Centos 7
(```libibverbs``` is linked vs ```libnl```, and ```mca/reachable``` choses ```libnl3``` if ```libnl3-devel``` package is installed)

i just noted that ```libibverbs``` package has been updated and now links with ```libnl3```,
that means that an Open MPI master works out of the box on an up to date Centos 7, with both ```libnl``` and ```libnl3``` libs and devel packages installed

the related PR is #1014

rhc54:
Hmmm...I have both packages installed, but it turned out that I didn't have libnl3-devel:

```shell
libnl.x86_64                           1.1.4-3.el7                     @anaconda
libnl-devel.x86_64                     1.1.4-3.el7                     @anaconda
libnl3.x86_64                          3.2.21-10.el7                   @anaconda
libnl3-cli.x86_64                      3.2.21-10.el7                   @anaconda
```

I'll add libnl3-devel and see how that impacts things.

ggouaillardet:
in this case you might run into the same issue but the other way around
- ```libibverbs``` depends on ```libnl3```
- ```mca/reachable``` cannot use ```libnl3``` since ```libnl3-devel``` is missing, so it fallbacks to ```libnl```

```ompi_info --all``` crashes if there is a ```libnl``` vs ```libnl3``` conflict

rhc54:
Okay, installing libnl3-devel and updating all the yum rpms did the trick. Very annoying waste of time...sigh. Thanks!

jsquyres:
image

libnl -- you win again!

PR #2215, opened 2016-10-12
v2.x communicator code updates

hjelmn:
This PR contains the following:
- Cleanup of the CID code to only have non-blocking. This reduces the overhead of trying to maintain multiple CID paths. (Cleanup)
- Optimization for CID generation on intercomm.
- Optimization for MPI_Comm_split_type. 

The last "feature" is probably the most important part of the PR. Before this PR MPI_Comm_split_type used MPI_Allgather on the communicator. This scales very poorly and is a performance bug in Open MPI. As such I see this PR as a performance bug fix not necessarily a new feature. If you disagree then postpone this to 2.2.0.


hjelmn:
@jsquyres FYI. Please let me know whether this meets the requirements of a bug fix. This was originally intended for 2.1.0 when the target was December. The code has been soaking on master for awhile and is probably ready to go now.

bosilca:
@hjelmn I just checked the MPI standard and it is illegal to supply different values for the split_type (page 247 lien 45) with the exception of MPI_UNDEFINED. Thus, I wonder if we really need the validity check. Second, I understand this operation as a different form of MPI_Comm_split, where the color is globally defined based on prior local knowledge. In other words, as each process has the information about the entire process placement and architecture, it can decide the local color based on this. Once the color defined it can simply call MPI_Comm_split.

hjelmn:
@bosilca Processes that supply MPI_UNDEFINED still need to know what the split type is or we will hang.

MPI_Comm_split_type is indeed just a special case of MPI_Comm_split but the restrictions allow us to do some optimization. The algorithm I implemented does the following:
 - Form the local and remote groups based on information about the process placement.
 - Use the above groups to create a new intermediary communicator.
 - Perform an all-gather on the (hopefully much-smaller) intermediate communicator if either 1) the procs may need to be reordered, or 2) if any procs supplied MPI_UNDEFINED.
 - If reordering/dropping ranks is needed MPI_Comm_split is run on the intermediate communicator to do the dirty work.

bosilca:
I see. So instead of the MPI_Comm_split all gather you assume that a 4 int reduction, followed by a communicator creation, and by smaller allgather will lead to better results. Do you have any pointer to what the improvement is ?

hjelmn:
There is a graph on #1873 that shows the improvement on an XC40 on up to 2048 ranks.

See https://cloud.githubusercontent.com/assets/1226817/16821220/5658435c-4912-11e6-8e9c-bde7e8639711.png

jsquyres:
bot:lanl:retest

bosilca:
the code looks good. 👍

jsquyres:
@bosilca We're using the Github reviews these days -- the 👍  is no longer enough.  😄

bosilca:
For some reason I don't have the review at the top on this ticket (but I did on Gilles's PR).

hppritcha:
This is way to big a code change this late in the attempt to release 2.1.0.  If the release is delayed considerably we'll think about merging this in.

jsquyres:
After discussion with @hppritcha, I moved the milestone back to v2.1.0.

I have also confirmed that this fixes COMM_SPAWN (i.e., #2234).

It would be nice if we could have a smaller version of this for v2.0.x -- e.g., could we leave out 91337bf5c818b752f8ab7e63705b4fea8f4c8dd7?

hjelmn:
@jsquyres Sure. I don't see why it can't be done. Will take a look tomorrow.

jsquyres:
@hjelmn Any progress on making a smaller PR for v2.0.2?

PR #2219, opened 2016-10-13
ess: tear down pmix (if any) before oob

ggouaillardet:


rhc54:
just curious - does this order matter?

ggouaillardet:
It does
If you ```mca_base_framework_close(&opal_pmix_base_framework)``` before ```orte_ess_base_app_finalize()```, a crash will occur when trying to unregister a PMIx handler (notification handler IIRC)

rhc54:
that i can believe - but your title here is that it needs to be torn down before calling the oob, and that is what i am questioning. i'm unaware of anything in pmix that would call the oob, so why that specific dependency?

ggouaillardet:
I did not dig too much, but my initial observation is that on ```orted``` side,  ```odls``` sees the task is not unregistered if the ```oob/tcp``` socket is closed before ```PMIx_Finalize``` is invoked, and then tries to kill it.
```oob/usock``` did not bother closing the socket, so we previously did not run into this scenario.

**Comment by ggouaillardet** [comment]
In ```odls_base_default_wait_local_proc()``` we currently end up in
```c
        /* check to see if a sync was required and if it was received */
        if (ORTE_FLAG_TEST(proc, ORTE_PROC_FLAG_REG)) {
            if (ORTE_FLAG_TEST(proc, ORTE_PROC_FLAG_HAS_DEREG) ||
                orte_allowed_exit_without_sync || 0 != proc->exit_code) {
...
            } else {
                /* we required a finalizing sync and didn't get it, so this
                 * is considered an abnormal termination and treated accordingly
                 */
                state = ORTE_PROC_STATE_TERM_WO_SYNC;
                OPAL_OUTPUT_VERBOSE((5, orte_odls_base_framework.framework_output,
                                     "%s odls:waitpid_fired child process %s terminated normally "
                                     "but did not provide a required finalize sync - it "
                                     "will be treated as an abnormal termination",
                                     ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
                                     ORTE_NAME_PRINT(&proc->name)));
            }
```

rhc54:
ok, that makes sense - thanks!

PR #2235, opened 2016-10-18
v2.x: add support for MPI_IN_PLACE in MPI_Ialltoall*

ggouaillardet:


jsquyres:
@hppritcha Good to go.

jjhursey:
@ggouaillardet It looks like git messed up the cherry pick on your first change to the mpi/c/ interface. It just updated the copyright, but not the code. I think it is because the code structure is slightly different than `master`. In the v2 series the check looks like:
```c
    } else if (MPI_IN_PLACE == sendbuf || MPI_IN_PLACE == recvbuf) {
            return OMPI_ERRHANDLER_INVOKE(MPI_COMM_WORLD, MPI_ERR_ARG,
                                          FUNC_NAME);
```

So in the v2.x series this is still disabled. I can work up a patch tomorrow.

jjhursey:
PR #2339 fixed the issue noted in the previous comment. Thanks!

Issue #2230, opened 2016-10-17
coll/libnbc does not handle internal errors


jjhursey:
If coll/libnbc encounters a failure on one of it's internal requests it enters an infinite loop displaying:
```shell
MPI Error in MPI_Testall() (18)
MPI Error in MPI_Testall() (18)
...
```

The particular test case that caused this has the PML issuing an `MPI_ERR_TRUNCATE` on one of the PML messages used inside `ibcast`.

libnbc detects the error [here](https://github.com/open-mpi/ompi/blob/9c3ae64297e034b30cb65298908014764216c616/ompi/mca/coll/libnbc/nbc.c#L326-L330), but does not destroy the handle. If the handle is destroyed, and an error is returned then the problem moves [here](https://github.com/open-mpi/ompi/blob/9c3ae64297e034b30cb65298908014764216c616/ompi/mca/coll/libnbc/coll_libnbc_component.c#L242-L251) where the user-facing request is never completed leading to an application hang.

ggouaillardet:
@jjhursey 
does the patch below fix the issue ?
```patch
diff --git a/ompi/mca/coll/libnbc/coll_libnbc.h b/ompi/mca/coll/libnbc/coll_libnbc.h
index 66f0f75..c9966cb 100644
--- a/ompi/mca/coll/libnbc/coll_libnbc.h
+++ b/ompi/mca/coll/libnbc/coll_libnbc.h
@@ -63,6 +63,7 @@ BEGIN_C_DECLS
 #define NBC_NOT_IMPLEMENTED 6
 #define NBC_INVALID_PARAM 7 /* invalid parameters */
 #define NBC_INVALID_TOPOLOGY_COMM 8 /* invalid topology attached to communicator */
+#define NBC_ERR_IN_STATUS 9 /* generic error */
 
 /* number of implemented collective functions */
 #define NBC_NUM_COLL 17
diff --git a/ompi/mca/coll/libnbc/coll_libnbc_component.c b/ompi/mca/coll/libnbc/coll_libnbc_component.c
index c49ad7b..bb2b9df 100644
--- a/ompi/mca/coll/libnbc/coll_libnbc_component.c
+++ b/ompi/mca/coll/libnbc/coll_libnbc_component.c
@@ -13,6 +13,8 @@
  * Copyright (c) 2008      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2013-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
[...]

PR #2245, opened 2016-10-18
coll/libnbc: Fix error path on internal error

jjhursey:
* If an error is detected internal to libnbc (e.g., PML truncation error)
   this patch makes sure that the request is completed and the `MPI_ERROR`
   field is set appropriately.
 * Fixes Issue #2230


bosilca:
Forgive me for my light understanding of the internal working of libnbc. If I correctly remember between 2 NBC scheduling barriers several operations might have been posted. As far as I see in this patch you are not even trying to cancel or clean them up, instead entierly focusing on propagating the error code to the user. As your request will be eventually removed, if any of the pending requests complete it will automatically lead to segfaults, basically prohibiting the usage of any error handled other than the default ABORT one. Right ?

jjhursey:
Yes, with the current version of the PR I'm not trying to cleanup the other possible outstanding internal requests. This version just makes sure that we propagate an error on the user request, so that the default ABORT error handler can be triggered at the MPI API level. That seems to be working now.

What we should do is cancel/complete the outstanding internal requests before marking the external request as complete.

The part I'm pondering at the moment is that we would need to call cancel on all of the other outstanding internal requests, then wait on them to complete. Since we are in a progress function, we can't wait at this location so we would need to add a 'draining'/'canceling' status to the handle. So that on successive calls we keep checking for completion until all of the requests are complete. Since not all PMLs support cancel (I believe that they return `OMPI_SUCCESS` regardless), is it possible that we could hang here?

bosilca:
For the PML that do not support cancel, the request is not internally marked as cancelled (the field _cancelled in the request status), so this should provide a hint on what you should expect from the request. However, if we assume that the send cancel will be removed from the next version of the MPI standard, this makes the problem almost insolvable through the MPI concepts, as all the sends must now have a matching receive (and if they are not correctly matched they will match the next collective).

jjhursey:
@bosilca I added a new commit that attempts to cleanup the outstanding requests. I also improves the error message printed. What do you think of that?

It did improve the error message displayed as part of Issue #2256 
```shell
MPI Error in MPI_Testall() (req 0 = 15)
```
Error code `15` is `MPI_ERR_TRUNCATE` which is coming from the PML.

bosilca:
The error message suggest that all non-blocking communications are unsafe when in fact only the non-blocking on this communicator might be affected. Otherwise it looks good.

PR #2246, opened 2016-10-19
Fix no-disconnect dynamics test

### LINKED ISSUE #2030: COMM_SPAWN is broken in 2.1.x HEAD

### LINKED ISSUE #2234: COMM_SPAWN broken in v2.0.x
This is a mirror for #2030, but for v2.0.x (instead of v2.1.x).

Keep all general discussion about the issue over on #2030; put any specific v2.0.x information here.

rhc54:
Bring across open-mpi/ompi@a681837ba8dc0cb3c6530b3ee2ae943a71565400 by hand to avoid conflicts.

Fixes #2234

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -16,7 +16,7 @@
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2015 NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


jsquyres:
@hppritcha Good to go.

PR #2259, opened 2016-10-20
Fix concurrency issue with MPI_Comm_accept

pietern:
I couldn't find `dpm_orte.c` in the 2.x or master trees. This commit only fixes the 1.10 tree.

Maybe this bug exists in those trees as well, but in a different location.

----

```
When hammered with many connecting clients, a server running the
function to assemble one big communicator would sometimes crash with
`dpm_orte.c: Unpack beyond buffer size` (or something along those
lines). The accept function executes 2 receive operations from the
remote peer. This function does receives on the wildcard address, which
means that these two receives could be fulfilled by different clients!
This change simply records the sender of the first receive and uses that
to filter the second receive.

The conversion from `WAIT_FOR_COMPLETION` to `LAZY_WAIT_FOR_COMPLETION`
is done to prevent unconnected clients (waiting in line to connect to
the server) from looping on `poll(2)` without timeout (at 100% CPU).
```

### DIFF
--- a/ompi/mca/dpm/orte/dpm_orte.c
+++ b/ompi/mca/dpm/orte/dpm_orte.c
@@ -206,7 +206,7 @@ static int connect_accept(ompi_communicator_t *comm, int root,
                                     ORTE_RML_NON_PERSISTENT,
                                     orte_rml_recv_callback, &xfer);
             /* wait for response */
-            OMPI_WAIT_FOR_COMPLETION(xfer.active);
+            OMPI_LAZY_WAIT_FOR_COMPLETION(xfer.active);
             i=1;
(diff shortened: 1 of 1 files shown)


lanl-ompi:

lanl-ompi:

lanl-ompi:

ibm-ompi:

 Gist: https://gist.github.com/ibm-ompi/7f110ab6c94da788f0fed01d83c12215

jjhursey:
bot:ibm:retest

rhc54:
@hppritcha Why is LANL's bot generating multiple comments to verify this patch? Is this a new feature?

pietern:
Thanks!

PR #2263, opened 2016-10-20
Compiler fixes for `win.c`

jjhursey:
PGI compiler stumbled on a couple of compiler fixes that did not get PR'ed over to the 2.x branch. One cases the PGI compiler to error out (initializer in `accumulate_ops_values`), the other two are warnings.


jjhursey:
A gist with the compiler error is [here](https://gist.github.com/ibm-ompi/c33b44ddf0768d16cc21b045015889d0):
```shell
PGC-S-0155-Empty initializer not supported  (win/win.c: 64)
PGC-W-0095-Type cast required for this conversion (win/win.c: 100)
PGC/power Linux 16.7-beta: compilation completed with severe errors
```

jsquyres:
@jjhursey Do these need to go in v2.0.x?

jjhursey:
@jsquyres No, v2.0.x is not impacted. The root cause changes were not ported to the v2.0.x series (I think because it relies on the addition of the enum_flags features in opal/mca)

PR #2271, opened 2016-10-21
Properly mark a node as down and decrease the number of daemons so any

rhc54:
subsequent grpcomm collectives can correctly operate. Note that only the
direct grpcomm component knows how to deal with down nodes.


rhc54:
this was caused by github being down for awhile this morning

bot:retest

rhc54:
Looks like all CI is dead in the DDOS attack. This is a minor change that affects only the DVM, so let's bring it along

PR #2279, opened 2016-10-24
v2.x: fortran and misc fixes

ggouaillardet:


ibm-ompi:

 Gist: https://gist.github.com/072d5c144b32174ba0fce122277fa38b

jjhursey:
IMB-CI (PGI Compiler) failure related to PR #2263 and not this ticket.

ompi/include/mpif-values.pl
```diff
@@ -431,26 +436,32 @@ sub write_file {
...
 
 sub write_fortran_file {
-    my ($header, $vals, $file) = @_;
+    my ($header, $vals, $lvals, $file) = @_;
 
     foreach my $key (sort(keys(%{$vals}))) {
         $header .= "        integer $key\n";
     }
+    foreach my $key (sort(keys(%{$lvals}))) {
+        $header .= "        integer(KIND=MPI_OFFSET_KIND) $key\n";
```
jsquyres: @ggouaillardet How did you find this issue -- do you have a test case?

I *think* that we're ok in terms of backwards compatibility because this constant (`MPI_DISPLACEMENT_CURRENT`) would not have compiled when passing in to any of the displacement parameters, right?  I.e., so there must be no applications using it.

If that's correct, I think we can make this technically-backwards-compatibility-breaking change because no one could have been using it before this.

Is my analysis correct?

jsquyres:
bot:ibm:retest

jsquyres:
Actually, I think my comment is true for *all* the changed symbols in this PR (i.e., I think we're ok in terms of back-compat because no one could have been using them before now).

ggouaillardet:
@jsquyres i found this when trying to build the mpich test suite ( ```test/mpi``` from https://github.com/pmodels/mpich )
build failed because the type was incorrect.
build could have succeeded with a less picky (F77 ?) compiler, but bad things would likely happen at runtime : alignment might be wrong, value is definitely wrong on big endian, and generally wrong on little endian (unless you are lucky and MSB is initialized to zero)
so i agree with your analysis : we did not break anything

ggouaillardet:
added Signed-off-by

jsquyres:
@ggouaillardet There should probably be a note about this in both NEWS and README.  I.e., we broke backwards compat, but it shouldn't be an issue because no one could have been using it, anyway, ...etc.

hppritcha:
@jsquyres do you want NEWS and README updates as part of this PR?

jsquyres:
I made #2392 for the NEWS changes.

PR #2281, opened 2016-10-24
v1.10: fortran and misc fixes

ggouaillardet:


ggouaillardet:
@jsquyres i also added the ```use-mpi-tkr``` fixes

jsquyres:
bot:lanl:retest

rhc54:
@jsquyres review??

rhc54:
@jsquyres Need for 1.10.6?

rhc54:
@jsquyres I think at this point it is safe to say nobody cares about this for the 1.10 series? If they do, then can we get this reviewed?

jsquyres:
Oh shoot, 2b504a2a8b261e9ae8ccbefffe78ab79db630fac from this PR will change the ABI (because we change the type of `MPI_DISPLACEMENT_CURRENT`.

@rhc54 I'm not quite sure what to do here.  On the one hand, we had the type of the variable wrong.  And Fortran is a strongly typed language, so correct MPI programs should not have been compiling previously.  ...but they may have adapted their programs to use our (incorrect) type.  But from the "this was a bug" perspective, I think we're ok to release.

...but then the question becomes: what should the `.so` version number be?  Do we actually bump the version such that the new `.so` version number marks that the prior version is incompatible?  I'm inclined to say "no" -- should we indicate that the library has changed, but it is still compatible, even though, technically speaking, that's a lie.  I think that this one variable is fairly obscure and so few people use the Fortran modules that we're ok.

We should probably add something in NEWS and/or README about this, though.

Thoughts?

jsquyres:
Also, technically, 1ba7519c8262bafe06503d075f1729009af99425 renames two APIs (which was clearly wrong, but still).

rhc54:
Given that these are fairly obscure, and that the changes were actually bug fix/corrections to match the standard, I'm comfortable with just indicating that the library changed and adding something in the NEWS about it

PR #2290, opened 2016-10-24
Open the conduits for application procs

rhc54:
We probably can remove all the RML-related frameworks from MPI applications now, but let's wait a bit to ensure we have cleaned up all the points where messaging might occur.


rhc54:
@anandhis Should fix the problem you are seeing with ofi_stress test program

PR #2294, opened 2016-10-25
fortran/use-mpi-tkr misc fixes

ggouaillardet:


ggouaillardet:
Also, it seems MPI_Sizeof is missing the implementation (e.g. compilation is ok, but link fails because of missing (mangled) symbols. Symbols should be there, right ?

ggouaillardet:
So gcc 4.8 cannot build MPI_Sizeof
I revamped a few things so MPI_Sizeof is not even part of mpi.mod
(E.g. compile will fail, not only link)

ompi/mpi/fortran/use-mpi-tkr/mpi.F90
```diff
@@ -58,4 +58,8 @@ module mpi
   include "pmpi-f90-file-interfaces.h"
 #endif
 
+#if OMPI_FORTRAN_BUILD_SIZEOF
+  include "mpi-tkr-sizeof.h"
```
jsquyres: I don't see this file in this commit -- did you forget to add it?

jsquyres:
bot:mellanox:retest

jsquyres:
@ggouaillardet Have no idea why that comment just showed up now -- I sent it last night while looking at this on a small phone.  I think you already answered my question and I missed it (the file was already generated; we just weren't using it).

ggouaillardet:
@jsquyres you had a good point about the ```.gitignore``` and i pushed open-mpi/ompi@52a1f967269712185aa0f082004380de0789ce58 to address it

PR #2302, opened 2016-10-27
v2.x: mpi: support MPI_Dims_create(..., ndims=0, ...)

ggouaillardet:
this is a bozo case, but it should not fail with MPI_ERR_DIMS

Signed-off-by: Gilles Gouaillardet gilles@rist.or.jp

(cherry picked from commit open-mpi/ompi@ad7f3f93b0b45588b172ed7f70f79cac2716512a)

### DIFF
--- a/ompi/mpi/c/dims_create.c
+++ b/ompi/mpi/c/dims_create.c
@@ -13,7 +13,7 @@
  *                         reserved.
  * Copyright (c) 2014      Intel, Inc. All rights reserved
  * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


bosilca:
The definition of the dims arguments is "number of Cartesian dimensions", which from a mathematical perspective must be at least 1.

ggouaillardet:
@bosilca i am fine with this interpretation. however the MPI standard does not explicitly states this is an error, and ```mpirun --mca mpi_param_check false ...``` will noop with ```ndims=0```

has this ever been discussed by the MPI Forum ? shall i ask the forum an official interpretation ?

fwiw, i hit this when running the mpich test suite (they expect ```ndims==0``` is just fine), but that would
not be the first time one test of the mpich test suite is incorrect ...

bosilca:
I merely pointed out that I can hardly find a sense in having 0 dimensions in a cartesian space, but otherwise I do not really care how we support such a corner case... Let's go for the cheapest solution, this code is already implemented and it provide the expected behavior for existing test codes.

nysal:
Just a minor suggestion "if ((NULL == dims) && (0 != ndims))" might be a bit more efficient for the common case.

ggouaillardet:
this topic is now being discussed on the mpi-comments mailing list.
```ndims == 0``` is currently unspecified by the standard, and MPI-4 will likely update it

jsquyres:
What is the status of this PR -- are we doing it?

ggouaillardet:
@jsquyres let's close that PR and revive it when/if MPI-4 is explicit about this bozo case

PR #2304, opened 2016-10-27
v1.10: mpi: support MPI_Dims_create(..., ndims=0, ...)

ggouaillardet:
this is a bozo case, but it should not fail with MPI_ERR_DIMS

Signed-off-by: Gilles Gouaillardet gilles@rist.or.jp

(back-ported from commit open-mpi/ompi@ad7f3f93b0b45588b172ed7f70f79cac2716512a)


jsquyres:
What about @bosilca's comment from #2302 (comment)?

PR #2307, opened 2016-10-27
btl/self: rewrite to decrease memory usage

hjelmn:
This commit rewrites much of the btl/self component to fix a long
standing memory usage bug. Before this commit the prepare_src path
would always allocate a max send fragment (256kB). This caused the
rank to allocate 32 * 256k useless buffers from one send. This commit
makes the following changes:

 - Add the MCA_BTL_FLAGS_GET flag by default. No reason not to set it.

 - Reduce the eager limit, max send size, buffers per allocation, and
   maximum buffer count per fragment size. These changes should have
   no noticible affect on performance but should greatly reduce the
   memory usage of the component.

 - Implement the sendi function. This should reduce self send latency
   somewhat.

 - Rewrite prepare_src to never allocate a eager or max send fragment
   for contiguous data.

 - add_procs needs to return something in the peer array for the proc
   self not just set the reachability bit. Now stores (void *) 1.

 - Various cleanups. Removed and unused file.


hjelmn:
@bosilca See any issues with this?

ibm-ompi:

jsquyres:
@hjelmn Do you really want to target v2.0.2?  Or did you mean v2.1.0?

hjelmn:
@jsquyres It is a performance bug. Its up to you if it should go to 2.0.2.

jsquyres:
Could you PR it to v2.1.0 for now?  Let me talk this over with Howard to see if we want it in v2.0.2.  It might be ok to push people in the direction of v2.1.0.

PR #2308, opened 2016-10-27
btl/vader: reduce memory footprint when using xpmem

hjelmn:
The vader btl kept a per-peer registration cache to keep track of
attachments. This is not really a problem with small numbers of local
ranks but can be a problem with large SMP machines. To reduce the
footprint there is now one registration cache for all xpmem
attachments. This will probably increase the lookup time for large
transfers but is a worthwhile trade-off.

Signed-off-by: Nathan Hjelm hjelmn@lanl.gov

### DIFF
--- a/opal/mca/btl/vader/btl_vader.h
+++ b/opal/mca/btl/vader/btl_vader.h
@@ -48,6 +48,7 @@
 #include "opal/sys/atomic.h"
 #include "opal/mca/btl/btl.h"
 #include "opal/mca/rcache/rcache.h"
+#include "opal/mca/rcache/base/rcache_base_vma.h"
 #include "opal/mca/btl/base/base.h"
 #include "opal/mca/rcache/rcache.h"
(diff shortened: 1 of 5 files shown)


hjelmn:
@hppritcha Just need another pair of eyes on this. I tested these changes on tr2 and it helps a lot with the memory footprint. Will run it through more tests on trinitite today.

hppritcha:
@jsquyres this seems like a 2.1 thing to me.

jsquyres:
@hppritcha I agree.

hppritcha:
This is okay with me, but its kind of sad that ```mca_rcache_base_vma_module_alloc ``` consumes enough memory that this is noticeable.

hppritcha:
@jsquyres ready to go

jsquyres:
@hppritcha This is master, so I assume @hjelmn could merge whenever he wants.  😄   But I'm a few timezone hours ahead of you guys, so I'll go ahead and merge.

PR #2322, opened 2016-10-31
Update the routed components as we no longer need to init_routes.

rhc54:
Fixes case of direct launch via srun


rhc54:
looks like github barfed

bot:mellanox:retest

Issue #2328, opened 2016-11-01
Fix case when RM/hostfile-provided names do not match hostname(1)


martinkontsek:
We have found out that running OpenMPI job via Moab/TORQUE will lead to this error:

```
[pacini002:01000] [[37987,0],1] bind() failed on error Address already in use (98)
[pacini002:01000] [[37987,0],1] ORTE_ERROR_LOG: Error in file oob_usock_component.c at line 247
[pacini002:01003] [[37987,1],0] usock_peer_recv_connect_ack: received unexpected process identifier [[37987,0],0] from [[37987,0],1]
[pacini002:00989] [[37987,0],0]-[[37987,1],0] mca_oob_usock_peer_recv_handler: invalid socket state(1)
[pacini002:00989] [[37987,0],0]-[[37987,1],0] mca_oob_usock_peer_recv_handler: invalid socket state(1)
[pacini002:00989] [[37987,0],0]-[[37987,1],0] mca_oob_usock_peer_recv_handler: invalid socket state(1)
[pacini002:01004] [[37987,1],1] usock_peer_recv_connect_ack: received unexpected process identifier [[37987,0],0] from [[37987,0],1]
```

Running the program directly via mpirun without Moab/TORQUE works perfectly. The same issue is present in `stable v2.0.1` and `nightly v2.0.1-134-g52bea1d from 2.0.x branch` and `nightly v2.0.1-270-g87a79fa from 2.x branch`.

However issue is NOT present in `stable v1.10.4` and `nightly v2.x-dev-3109-g6074c2a from master` as it is working properly directly via mpirun and also with Moab/TORQUE. We use Moab v9.0.2 and TORQUE v6.0.2 supplied as RPM by Adaptive Computing, LibFabric 1.3.0 and we are running on RHEL 7.2. We have multiple OpenMPI versions installed and we use Environment Modules to select version needed.

We build RPM from tarball using this command with specfile included in the tarball.

```
[...]

PR #2329, opened 2016-11-01
ras/*: Fix !orte_keep_fqdn_hostnames for RAS components

jjhursey:
* By default, make sure that we are using the short hostnames and not the fully qualified hostnames when running under LSF and other resource managers with RAS components.
 * The previous commit missed updating the RAS components to strip of the hostnames that they discover from the environment.
 * Related to commit open-mpi/ompi@d26dd2c20e19be864bd5a2e20770d1773fd75406

### DIFF
--- a/orte/mca/ras/gridengine/ras_gridengine_module.c
+++ b/orte/mca/ras/gridengine/ras_gridengine_module.c
@@ -10,6 +10,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006-2010 Oracle and/or its affiliates.  All rights reserved
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -28,6 +29,7 @@
 #include <string.h>
 
 #include "opal/util/output.h"
+#include "opal/util/net.h"
 #include "orte/util/show_help.h"
 #include "orte/mca/errmgr/errmgr.h"
 #include "orte/runtime/orte_globals.h"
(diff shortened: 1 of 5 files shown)


lanl-ompi:

lanl-ompi:

lanl-ompi:

rhc54:
can someone please explain these "Can one of the admins verify this patch?" messages, and why we get three of them? Is there a way to turn these things off??

jjhursey:
@rhc54 This fixes LSF, but other RAS components may also need updating. I think this might be related to Issue #2328, but I'm not sure.

I will work on a similar patch for the rest of the RAS components, but will not be able to test them. So I'll add them as separate commits to this PR for individual review. We can broaden the scope of the PR as needed.

jjhursey:
I just pushed a similar fix for the follow `ras` components
 * `tm`
 * `gridengine`
 * `loadleveler`
 * `slurm`

I did not update `alps` since it uses an integer value for the hostname, so should not be impacted.
I did not update `simulator` since I'm not sure exactly where to do so.

rhc54:
Don't worry about simulator as it doesn't actually launch anything, and thus the hostname length is irrelevant.

hppritcha:
the message was coming from jenkins because I tried to fix something per jeff's request but it wasn't correct.  it should be fixed now.

jjhursey:
@rhc54 This fixes an issue for LSF that I need to get into the v2.0.2 release. My question is really about the other RAS components since I cannot test them. Can you take a quick look over the change and see if you notice anything that I might have missed based on your experience with these environments?

My inclination is to merge this into the `master` branch then let MTT have a pass at it tonight. It should be obvious if something is wrong. Then I'll PR it over into the `v2.x` and `v2.0.x` branches.

rhc54:
I can't test them all either, but I don't see anything obviously wrong, so I agree with your plan.

jjhursey:
Sounds good. I'm going to keep these as 5 separate commits instead of squashing down to 1. This should help us isolate the part of this PR that might have problems identified by MTT testing.

hppritcha:
Would it be okay just to push the commit for LSF component to 2.0.x?  I'm nervous about making changes to components we don't think we'll get good testing for in a bug fix release.  Do we know if all of the ras components in question are exercised by MTT?  I'm not concerned about alps ras because it uses "nids" as was pointed out.
@jsquyres

jjhursey:
@hppritcha We can if you would prefer. Then bring all of the patches into `v2.x`?

I expect the other environments will need these patches if they are using a mixed short/long hostname setup. If they are all using short hostnames, they will be fine. If they use long hostnames, then they will have a problem.

PR #2334, opened 2016-11-01
mpi/neighbor_allgatherv: fix copy&paste error and add helpers

### LINKED ISSUE #2324: Neighborhood Collectives
@jsquyres I've found serious issues in the implementation of neighborhood collectives. These issues showed up after a memory allocation optimization I've implemented in mpi4py in commit [c2eaa292](https://bitbucket.org/mpi4py/mpi4py/commits/c2eaa292b2aa1d4bf809c91a1c19d988629f009a), which failed badly with Open MPI 2.0.1 in this [Bitbucket Pipelines](https://bitbucket.org/mpi4py/mpi4py/addon/pipelines/home#!/results/%7B8c67bd67-9cd6-4251-94b9-c96bf375c7a3%7D) build.

After checking the source code in ompi/v2.x, I found a major issue for the case of intracommunicators with a topology. The code simply do not check the topology to determine the number or sources and destinations, and use ``ompi_comm_size()`` instead. This leads to wrong checks and reads after end of arrays.

In mpi4py, I'm using the following [code](https://bitbucket.org/mpi4py/mpi4py/src/master/src/mpi4py/MPI/commimpl.pxi?fileviewer=file-view-default#commimpl.pxi-49) to determine the number of sources and destinations. This piece of code only handles intracommunicators, but could be extended to intercommunicators.

BTW, the wording of the MPI standard seems to require a topology for neighborhood collectives, although supporting intracomms and intercomms seems possible and the result should be equivalent to regular allgather and alltoall. Or maybe the right thing to do is to just error in case of intercomms or intracomms with no topology.

Am I missing something?

PS: MPICH seems to implement neighbor collectives just for Cartesian/graph topologies, the helper routine `MPIR_Topo_canon_nhb_count` is quite similar to what I'm using in mpi4py (though the MPICH one fails in case of no topology). @roblatham00 Any comments to you competitors :-) about this?

```
$[dalcinl@localhost mpich.git]$ git grep MPIR_Topo_canon_nhb_count
src/include/mpir_topo.h:int MPIR_Topo_canon_nhb_count(MPIR_Comm *comm_ptr, int *indegree, int *outdegree, int *weighted);
src/mpi/topo/inhb_allgather.c:    mpi_errno = MPIR_Topo_canon_nhb_count(comm_ptr, &indegree, &outdegree, &weighted);
src/mpi/topo/inhb_allgatherv.c:    mpi_errno = MPIR_Topo_canon_nhb_count(comm_ptr, &indegree, &outdegree, &weighted);
src/mpi/topo/inhb_alltoall.c:    mpi_errno = MPIR_Topo_canon_nhb_count(comm_ptr, &indegree, &outdegree, &weighted);
src/mpi/topo/inhb_alltoallv.c:    mpi_errno = MPIR_Topo_canon_nhb_count(comm_ptr, &indegree, &outdegree, &weighted);
src/mpi/topo/inhb_alltoallw.c:    mpi_errno = MPIR_Topo_canon_nhb_count(comm_ptr, &indegree, &outdegree, &weighted);
src/mpi/topo/topoutil.c:#define FUNCNAME MPIR_Topo_canon_nhb_count
src/mpi/topo/topoutil.c:int MPIR_Topo_canon_nhb_count(MPIR_Comm *comm_ptr, int *indegree, int *outdegree, int *weighted)
```

hjelmn:
This commit adds a helper function to get the inbound and outbound
neighbor count and updates the neighbor_allgatherv bindings to use the
correct count when checking the input parameters.

Fixes #2324

### DIFF
--- a/ompi/mca/topo/base/base.h
+++ b/ompi/mca/topo/base/base.h
@@ -195,6 +195,9 @@ OMPI_DECLSPEC int
 mca_topo_base_dist_graph_neighbors_count(ompi_communicator_t *comm,
                                          int *inneighbors, int *outneighbors, int *weighted);
 
+
+int mca_topo_base_neighbor_count (ompi_communicator_t *comm, int *indegree, int *outdegree);
+
 END_C_DECLS
 
 #endif /* MCA_BASE_TOPO_H */
--- a/ompi/mca/topo/base/topo_base_frame.c
+++ b/ompi/mca/topo/base/topo_base_frame.c
@@ -71,6 +71,33 @@ static int mca_topo_base_open(mca_base_open_flag_t flags)
     return mca_base_framework_components_open(&ompi_topo_base_framework, flags);
 }
 
+int mca_topo_base_neighbor_count (ompi_communicator_t *comm, int *indegree, int *outdegree) {
+  if (!OMPI_COMM_IS_TOPO(comm)) {
+    return OMPI_ERR_BAD_PARAM;
+  }
+
+  if (OMPI_COMM_IS_CART(comm)) {
+    /* cartesian */
+    /* outdegree is always 2*ndims because we need to iterate over
+       empty buffers for MPI_PROC_NULL */
+    *outdegree = *indegree = 2 * comm->c_topo->mtc.cart->ndims;
+  } else if (OMPI_COMM_IS_GRAPH(comm)) {
+    /* graph */
+    int rank, nneighbors;
+
+    rank = ompi_comm_rank (comm);
+    mca_topo_base_graph_neighbors_count (comm, rank, &nneighbors);
+
+    *outdegree = *indegree = nneighbors;
+  } else if (OMPI_COMM_IS_DIST_GRAPH(comm)) {
+    /* graph */
+    *indegree = comm->c_topo->mtc.dist_graph->indegree;
+    *outdegree = comm->c_topo->mtc.dist_graph->outdegree;
+  }
+
+  return OMPI_SUCCESS;
+}
+
 MCA_BASE_FRAMEWORK_DECLARE(ompi, topo, "OMPI Topo", NULL,
                            mca_topo_base_open, mca_topo_base_close,
                            mca_topo_base_static_components, 0);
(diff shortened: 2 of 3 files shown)


ompi/mca/topo/base/base.h
```diff
@@ -194,6 +194,9 @@ OMPI_DECLSPEC int
 mca_topo_base_dist_graph_neighbors_count(ompi_communicator_t *comm,
                                          int *inneighbors, int *outneighbors, int *weighted);
 
+
```
ggouaillardet: we already have a ```ompi_comm_neighbors_count``` subroutine that does the same think (at least at first glance)
that being said, your new subroutine only uses internal subroutine (vs ```MPI_*```) so i like it better.

should we only keep your subroutine ?

dalcinl: ``ompi_comm_neighbors_count`` using ``MPI_*`` interfaces interfere with MPI profiling. I'm in favor of the new routine. Moreover, the new routine should be used everywhere and the former should be removed.

dalcinl:
@hjelmn Please also fix the check ``if (NULL == displs) ...``  the following way

```c
if (in_size > 0 && NULL == displs) {
    return OMPI_ERRHANDLER_INVOKE(comm, MPI_ERR_BUFFER, FUNC_NAME);
}
```

jsquyres:
@hjelmn Looks like this PR still needs some work.

ibm-ompi:

 Gist: https://gist.github.com/752b78b8734de54ad555409612822542

hppritcha:
@hjelmn please take a look at this.

hjelmn:
Ok, should be better now. @ggouaillardet Please take a look.

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

hjelmn:
:bot:retest:

ibm-ompi:

jjhursey:
bot:ibm:pgi:retest

PR #2356, opened 2016-11-03
v2.02: enable mca no build help update

jsquyres:
Also update the configure.ac help message for --enable-mca-no-build to
avoid using a framework name that does not exist any more.

(cherry picked from commit c54dc87f719da65b2446dee16e0578f259541f43)

@hppritcha Please review.


hppritcha:
merging to make sure we get a new MTT run tomorrow

PR #2357, opened 2016-11-03
SPML/UCX: Fixing incorrect handling of UCX return codes

shamisp:
For UCX it is legal to return UCS_INPROGRESS (1) code, which means that the
operation was successfully started but not immediately completed. This is a
"good" return code that should not be handled as an error.

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx.h
+++ b/oshmem/mca/spml/ucx/spml_ucx.h
@@ -3,6 +3,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      ARM, Inc. All rights reserved.
  * $COPYRIGHT$
  * 
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


shamisp:
@yosefe @miked-mellanox @jladd-mlnx please take a look. This is a fatal bug that I would suggest to stage for a release.

oshmem/mca/spml/ucx/spml_ucx.h
```diff
@@ -130,7 +131,7 @@ mca_spml_ucx_get_mkey(int pe, void *va, void **rva)
 
 static inline int ucx_status_to_oshmem(ucs_status_t status)
 {
-    return OPAL_LIKELY(UCS_OK == status) ? OSHMEM_SUCCESS : OSHMEM_ERROR;
+    return OPAL_LIKELY(status >= 0) ? OSHMEM_SUCCESS : OSHMEM_ERROR;
```
yosefe: @shamisp IMHO we should change this only for non-blocking calls, by itroducing another function e.g ucx_status_to_oshmem_nb()

shamisp: @yosefe - not a bad idea.

shamisp:
@yosefe - done.

shamisp:
@yosefe - can you please take a look ?

miked-mellanox:
@shamisp - could you please PR it into v2.x and v1.10 branches?
Thanks

shamisp:
@miked-mellanox - did you merge the *_nbi() interface the v2.x and v1.10. If so, I will prepare a patch.

miked-mellanox:
right, only v2.x

shamisp:
@miked-mellanox it is not in v2.x

miked-mellanox:
it is in v2.x
https://github.com/open-mpi/ompi/blob/v2.x/oshmem/mca/spml/ucx/spml_ucx.c#L460

shamisp:
The patch does not apply. I will take a look.

Issue #2358, opened 2016-11-03
"mpirun --debug ..." broken in v2.0.1


gpaulsen:
It's also broken on top of v2.0.x branch.

line 1184(v2.x) of orte/tools/orterun/orterun.c is:

```
   /* Do we want a user-level debugger? */

    if (orterun_globals.debugger) {
#if 0
        /* see https://github.com/open-mpi/ompi/issues/1225
         * Once things are fixed, the orte_show_help/exit can be removed
         * and this code re-enabled.
         */
        run_debugger(orte_basename, cmd_line, argc, argv, orterun_globals.num_procs);
#else
        orte_show_help("help-orterun.txt", "no_debugger_launch_support", false);
        exit(-1);
#endif
    }
```

This causes --debug to print this message on v2.0.1.

related to https://github.com/open-mpi/ompi/issues/1225

***EDIT: Added proper block verbatim notation***

jsquyres:
@gpaulsen Github pro tip: when doing a multi-line verbatim block on github, use *3* single-tick marks to denote the beginning / end (va. the *1* single tick mark which is for inline verbatim).

jsquyres:
@gpaulsen This looks like a blocker for v2.0.2, right?  If so, please apply labels and the v2.0.2 milestone.

rhc54:
Hmmm..I'm pretty sure this is actually fixed - will look into it tonight.

rhc54:
Just looked at the 2.x branch, and there is no reason for the code not to be enabled. I'll test it and update things here, but I believe all that is required is to simply remove the ```#if``` clause.

gpaulsen:
great.  We're testing that tonight also

rhc54:
I tested both branches using the ORTE_TEST_DEBUGGER_ATTACH envar, and the debugger release was correctly sent and received.

gpaulsen:
I also tested, that it correctly launched totalview.

jsquyres:
Fixed have been merged.

PR #2359, opened 2016-11-03
Remove the stale protection around the debugger option for orterun - …

rhc54:
…debugger attach is fully functional


rhc54:
Fixes #2358

PR #2341, opened 2016-11-02
Poc/v2.x: mpirun scalability under slurm

ggouaillardet:

Issue #2270, opened 2016-10-21
mpirun doesn't scale to large node counts


rhc54:
There is a fundamental issue with mpirun as it stands today when launched in any environment other than ssh. In the absence of static ports, every daemon opens a socket back to mpirun to "phone home" upon launch. This is done because the only address the daemon knows is that of mpirun. The daemon reports its uri back to mpirun, who then includes that info in its xcast of the application's launch message.

The xcast algorithm always starts with mpirun sending to some subset of daemons. Upon receipt of the launch message, each daemon unpacks the global daemon URI wireup blob and processes it to extract all the connection info. It then opens sockets to whatever daemons are next in the xcast pattern.

Thus, the socket back to mpirun is _only_ required for the initial "phone home" communication - yet we hold that socket open throughout the length of the job. This winds up consuming N sockets on the node where mpirun is executing, where N = number of nodes in the job. For large scale, that is a big number.

What would be better is to have mpirun close each socket upon receiving the "phone home" message from a daemon. However, this means that the daemon cannot use that socket as a lifeline and terminate if it closes. Instead, the lifeline must come from the xcast wireup, which means there is a window of vulnerability where the daemon is "alone" and cannot know if the job died somewhere "upstream" of it.

There isn't that much time difference between an ssh and "native RM" launch for our daemons when using mpirun. So perhaps one answer is to just say "use ssh at scale" - but some environments may not allow that option.

@bosilca @hjelmn @abouteiller Any thoughts?

rhc54:
That pretty much describes what we do today. ssh only phones back to the prior wave, xcast and allgather follow the topology, etc. Not sure what you think isn't already there, but I'd welcome any pointers to things we missed.

As I said, the problem is that you cannot use ssh everywhere. Many systems will not allow it on the backend nodes, and so the wave-based model won't work. We have the static port method, but that assumes one can find/assign the same port on every node, and that isn't always possible. Eventually, we can try to get the RMs to assign a port, but that may or may not be something they are willing to do - SLURM had it for awhile, but dropped it last year as they are trying to push people away from mpirun.

So I'm looking for an alternative solution for those (increasing number of) systems where ssh isn't allowed, and we cannot reliably set a static port.

bosilca:
All the above steps were combined together in a single larger step coordinated between multiple components, that at all moments was maintaining a connected topology (to completely remove any window of vulnerability).

[...]

PR #2362, opened 2016-11-04
POC : slurm resv ports

ggouaillardet:


ggouaillardet:
Refs #2270 

let's follow-up #2341 (v2.x based) here from now (master based)

since all orted are started at the same time and might communicate directly with each other, there is a race condition. by default ```oob_tcp_retry_delay``` MCA param is ```0``` and it should be bumped to something higher to enable retry when the race condition occurs.

a typical ```mpirun``` command line should look like

```
mpirun --mca plm_slurm_resv_ports true --mca oob_tcp_retry_delay 1 ...
```

the second commit is a temporary workaround for the ```ORTE_PROC_FLAG_ALIVE``` thing, and is not intended to land into master

ibm-ompi:

 Gist: https://gist.github.com/fdb2fa4a4815e6107ec906eece96a88b

bwbarrett:
removing the 3.1.0 milestone for the merge, since it's against master.

hppritcha:
closing due to lack of activity. reopen if this will become active again.

PR #2364, opened 2016-11-04
Since static ports are only used by ORTE if the runtime option is given,

rhc54:
there is no need for a configure option as well - so remove the
--enable-orte-static-ports configure option. When decoding the daemon
nidmap, mark new daemons as ALIVE by default - we will discover dead
ones as we go.

Refs #2362 

### DIFF
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -493,7 +493,6 @@ int orte_ess_base_orted_setup(char **hosts)
         goto error;
     }
 
-#if ORTE_ENABLE_STATIC_PORTS
     /* if we are using static ports, then we need to setup
      * the daemon info so the RML can function properly
      * without requiring a wireup stage. This must be done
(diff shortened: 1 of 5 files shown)


orte/mca/ess/base/ess_base_std_orted.c
```diff
@@ -493,33 +493,29 @@ int orte_ess_base_orted_setup(char **hosts)
...
      * own port, which we need in order to construct the nidmap
      */
     if (orte_static_ports) {
-        /* define the routing tree so we know the pattern
-         * if we are trying to setup common or static ports
-         */
-        orte_routed.update_routing_plan(NULL);
         /* extract the node info from the environment and
-         * build a nidmap from it
+         * build a nidmap from it - this will update the
```
ggouaillardet: @rhc54 that commit broke static ports.
simply
```n0 $ mpirun --host=n1:1 --mca oob_tcp_static_ipv4_ports 1234-np 1 hostname```

```orte_util_build_daemon_nidmap(hosts)``` is invoked with ```hosts == NULL```, so this is really a no-op (the comment suggests otherwise)

then the comment in the ```else``` clause refers to static ports, though static ports are handled in the ```if``` clause

the inline patch belows works for me, and it fixes #2362 

```patch
diff --git a/orte/mca/ess/base/ess_base_std_orted.c b/orte/mca/ess/base/ess_base_std_orted.c
index 99fb81b..374d1db 100644
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -493,6 +493,12 @@ int orte_ess_base_orted_setup(char **hosts)
         goto error;
     }
 
+    /* be sure to update the routing tree so the initial "phone home"
+     * to mpirun goes through the tree if static ports were enabled - still
+     * need to do it anyway just to initialize things
+     */
+    orte_routed.update_routing_plan(NULL);
+
     /* if we are using static ports, then we need to setup
      * the daemon info so the RML can function properly
      * without requiring a wireup stage. This must be done
@@ -509,12 +515,6 @@ int orte_ess_base_orted_setup(char **hosts)
             error = "construct daemon map from static ports";
             goto error;
         }
-    } else {
-        /* be sure to update the routing tree so the initial "phone home"
-         * to mpirun goes through the tree if static ports were enabled - still
-         * need to do it anyway just to initialize things
-         */
-        orte_routed.update_routing_plan(NULL);
     }
 
     /* Now provide a chance for the PLM
```

rhc54: Odd - it works just fine for me on a multi-node cluster. However, it also works fine the other way, so it really doesn't seem to matter. Go ahead and make the change if it makes a difference for you.

ggouaillardet: ok, will do.

btw, i missed some MCA params ...

can you try
```n0 $ mpirun --host=n1:1 --mca oob_tcp_static_ipv4_ports 1234-np 1 --mca routed,radix --mca oob tcp hostname```

rhc54:
@ggouaillardet Are you still planning to fix this? Or is there something I need to do?

ggouaillardet:
@rhc54 i made #2536 for that
feel free to merge it once it completes CI

rhc54:
mucho thanks!

PR #2365, opened 2016-11-04
Update schizo framework

rhc54:
Update the schizo framework to support checking the environment during startup. This allows better detection of singleton scenarios, and provides a mechanism by which we can recognize and support SLURM-imposed bindings.

Discussed with @jsquyres for pre-approval - addresses problem reported on the user list:

https://mail-archive.com/users@lists.open-mpi.org/msg30212.html

### DIFF
--- a/config/opal_check_singularity.m4
+++ b/config/opal_check_singularity.m4
@@ -1,0 +1,60 @@
+# -*- shell-script ; indent-tabs-mode:nil -*-
+#
+# Copyright (c) 2016      Intel, Inc. All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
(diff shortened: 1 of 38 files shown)


hppritcha:
I'd like to do some testing in SLURM and ALPS environments before merging.

jsquyres:
@rhc54 Sorry -- we let this get stale, and it now has conflicts.  Can you resolve?

hppritcha:
actually I can't test this right now because I don't know where to pull the branch from.  I'd rather do that then pull from the patch file, esp. since there are conflicts with v2.x

jsquyres:
Yo @rhc54 -- this is a new one to me: GitHub can't find the repo from which you created this PR (!).  Do you need to re-create this PR?  (there's conflicts that need to be fixed, too)

rhc54:
This has sat so long it might have been done off the old release repo - I'll just kill it and forget it.

PR #2379, opened 2016-11-07
coll/libnbc: Work around for non-uniform data types in ibcast

jjhursey:
* If (legal) non-uniform data type signatures are used in ibcast
   then the chosen algorithm may fail on the request, and worst case
   it could produce wrong answers.
 * Add an MCA parameter that, by default, protects the user from this
   scenario. If the user really wants to use it then they have to
   'opt-in' by setting the following parameter to false:
   - `-mca coll_libnbc_ibcast_skip_dt_decision f`
 * Once the following Issues are resolved then this parameter can
   be removed.
   - https://github.com/open-mpi/ompi/issues/2256
   - https://github.com/open-mpi/ompi/issues/1763

(cherry picked from commit 350ef67fe0505ab0b7c151c024432fa128f5711f)

### DIFF
--- a/ompi/mca/coll/libnbc/coll_libnbc.h
+++ b/ompi/mca/coll/libnbc/coll_libnbc.h
@@ -15,6 +15,7 @@
  *                         reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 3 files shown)


jjhursey:
Fixes #2256 for the `v2.x` branch.

jsquyres:
This was merged into v2.0.x, so I'm also merging it here for v2.x.

PR #2381, opened 2016-11-07
v2.0.x comm CID refactor

### LINKED ISSUE #2380: MPI_Comm_idup hang in v2.0.x with MPI_THREAD_MULTIPLE
I created a test case in ompi-tests under`ibm/communicator/comm_idup_many.c`

This test is doing a simple:
```c
    for (i = 0; i < num_comms_to_create; i++) {
        MPI_Comm_idup(MPI_COMM_WORLD, &comms[i], &req[i]);
    }

    MPI_Waitall(num_comms_to_create, req, MPI_STATUSES_IGNORE);
```

If you compile the `v2.0.x` branch with `--enable-mpi-thread-multiple` this test will hang when you try to create more than one communicator this way.
```shell
./configure --enable-debug --enable-mpi-thread-multiple
```

The problem is that the first request will progress past stage one, but the second request will block the progress thread in this section of code:
 * https://github.com/open-mpi/ompi/blob/v2.0.x/ompi/communicator/comm_cid.c#L517-L530

The `master` and `v2.x` branches pass this test. So I believe that the CID allocation fix that went into those two branches 'fixed' this problem.

hjelmn:
Fixes #2380


jjhursey:
Might also fix #2234 - @jsquyres to check.

jjhursey:
👍  I can confirm that this fixes Issue #2380. Thanks @hjelmn!

jsquyres:
I confirm that this fixes COMM_SPAWN problems (#2234).

PR #2384, opened 2016-11-08
pthread cleanup

ggouaillardet:


bosilca:
Make sense. Please also remove the "unsuported progress thread" section from help-mpi-btl-tcp.txt

ggouaillardet:
good catch, i just did

PR #2385, opened 2016-11-08
v2.x: powerpc: Add support for powerpcle in timer/pstat.

AboorvaDevarajan:
Signed-off-by: Aboorva Devarajan <abodevar@in.ibm.com>


jsquyres:
@hppritcha I think this is good to go for v2.x.

PR #2398, opened 2016-11-11
Protect the tcp_endpoints list from concurrent accesses.

bosilca:
This patch will fix some [hopefully all] of the segfaults when TCP progress thread is on. Thanks Gilles for your help.


hppritcha:
Is there a corresponding commit on master?

bosilca:
This is the PR for master, that will be pulled in once we agree on the solution.

jsquyres:
@ggouaillardet @bosilca Should we merge to master / PR to v2.0.x and v2.x?

bosilca:
I will let @ggouaillardet comment and merge it if he is happy with.

ggouaillardet:
i read the patch and it looks good to me, so i will go ahead and merge it.

i will fully test it next week, and then PR to the release branches

Issue #2320, opened 2016-10-31
TCP BTL in docker container freezes (again, sorry)


d-a-v:
Hello,

When trying to run openmpi in a docker container using a bridge as the only existing network interface available, the application gets stuck.

A setup is available for anyone who wishes to [test](https://github.com/d-a-v/openmpi-bridge-docker).

This is a still real old problem. This https://github.com/open-mpi/ompi/issues/160 issue from 2014 expected the problem to be solved by ompi-v1.9+. 
I have not come to beg. I just want to highlight this bug again, because it would rock to run setups with slurm+openmpi in bridged docker containers, so to allow far easier HPC administration.

ggouaillardet:
which version of Open MPI do you run ?

master branch blacklists interfaces named ```vir*```, and this is hard coded in ```oob/tcp```
so i am aware you cannot run ```mpirun``` on your host and put containers in your hostfile

the inline patch can help you
```patch
diff --git a/orte/mca/oob/tcp/oob_tcp_component.c b/orte/mca/oob/tcp/oob_tcp_component.c
index b610812..19ba1a9 100644
--- a/orte/mca/oob/tcp/oob_tcp_component.c
+++ b/orte/mca/oob/tcp/oob_tcp_component.c
@@ -514,10 +514,12 @@ static int component_available(void)
         /* get the name for diagnostic purposes */
         opal_ifindextoname(i, name, sizeof(name));
 
+#if 0
         /* ignore any virtual interfaces */
         if (0 == strncmp(name, "vir", 3)) {
             continue;
         }
+#endif
 
         /* handle include/exclude directives */
         if (NULL != interfaces) {
```

d-a-v:
Hi, thanks.
I tried the master branch and the v2.0.x branch, with same issue.
The name of the interface inside the docker container is eth0, so the patch above has no effect (but still I tried too, and the name of the interface is indeed eth0 inside the 'name' variable).

rhc54:
You might want to check on the slurm mailing list as I know people are doing this today - not sure what configuration was required.

ggouaillardet:
I will try to reproduce this.
At first glance, all happens inside containers, and eth0 is the only avsilable interface, so from an Open MPI point of view, thst should be business as usual ...
The hang suggests ip forwarding on the hosts for container-to-container traffic does not work as it should 
If you have any chance, can you double check with tcpdump traffic gies through ?

ggouaillardet:
[...]

PR #2400, opened 2016-11-11
Report dropped connections.

bosilca:
This is a collective fix for the issue #2320. Instead of silently dropping connections that do not match the IP addresses known for the peer, we nicely inform the user about the mismatch.

### DIFF
--- a/opal/util/net.c
+++ b/opal/util/net.c
@@ -385,7 +385,6 @@ opal_net_get_hostname(const struct sockaddr *addr)
         if(NULL == inet_ntop(AF_INET6, &((struct sockaddr_in6*) addr)->sin6_addr,
                              name, NI_MAXHOST)) {
             opal_output(0, "opal_sockaddr2str failed with error code %d", errno);
-            free(name);
             return NULL;
         }
         return name;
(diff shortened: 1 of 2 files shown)


opal/util/net.c
```diff
@@ -405,7 +403,6 @@ opal_net_get_hostname(const struct sockaddr *addr)
        int err = errno;
        opal_output (0, "opal_sockaddr2str failed:%s (return code %i)\n",
                     gai_strerror(err), error);
-       free (name);
```
jsquyres: @bosilca Why did you remove these 3 `free()` statements?

bosilca: Because I was curious why we release this buffer in all error cases but never on the successful path, so I went read the code of the get_hostname_buffer function. The returned buffer is associated with the calling thread (TLS), and we better don't free it at all.

jsquyres: Why is it associated with TLS data?

bosilca: To be similar to inet_ntoa I guess (the returned string is in a static area that is overwritten by successive calls).

ibm-ompi:

 Gist: https://gist.github.com/3892e955c74781a6f4eb9e7b8e08fcf5

bosilca:
This has been integrated into the master in ec4a235e6a5 and then by #3077.

PR #2401, opened 2016-11-12
ompi/init: always lazy-wait in ompi_mpi_init

artpol84:
According to discussion in #2181 we don't need MCA
parameter any more.


artpol84:
@rhc54 @jladd-mlnx @hppritcha please review!
I will need to PR the same change to v2.1.

PR #2412, opened 2016-11-14
topo/base: Fix module reference in collective call

jjhursey:


jsquyres:
@hppritcha I think this is fine once CI finishes (Travis previously had a false failure; I just restarted the job -- it should finish in a while).

PR #2416, opened 2016-11-14
v2.0.x: coll: fix coll_base_comm_get_reqs usage

ggouaillardet:
- do  not coll_base_comm_get_reqs() zero requests
- use stack allocated request for irecv/send instead of irecv/isend with two dynamically allocated requests
- misc code cleanup


jsquyres:
@hppritcha I think this is good to go.

PR #2423, opened 2016-11-15
io/ompio: change the default value of mca parameter

edgargabriel:
change the default value of the mca_io_ompio_cycle_buffer_size parameter in order to avoid accidental truncation of a file for very large individual operations.
Thanks to @cniethammer for reporting it.


edgargabriel:
@cniethammer would you mind reviewing this change, since you reported the problem and successfully ran the application after applying this change?

jsquyres:
Per https://github.com/open-mpi/ompi/pull/2424#issuecomment-264532855, we'll merge this into v2.1.x for now.  More work to come on this issue.

PR #2424, opened 2016-11-15
io/ompio: change the default value of mca parameter

edgargabriel:
change the default value of the mca_io_ompio_cycle_buffer_size parameter in order to avoid accidental truncation of a file for very large individual operations.
Thanks to @cniethammer for reporting it.

### DIFF
--- a/ompi/mca/io/ompio/io_ompio.h
+++ b/ompi/mca/io/ompio/io_ompio.h
@@ -71,6 +71,7 @@ OMPI_DECLSPEC extern int mca_io_ompio_coll_timing_info;
  * General values
  */
 #define OMPIO_PREALLOC_MAX_BUF_SIZE   33554432
+#define OMPIO_DEFAULT_CYCLE_BUF_SIZE  536870912
 #define OMPIO_PERM_NULL               -1
 #define OMPIO_IOVEC_INITIAL_SIZE      100
 #define OMPIO_ROOT                    0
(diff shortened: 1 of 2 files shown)


edgargabriel:
@cniethammer would you mind reviewing this change, since you reported the problem and successfully ran the application after applying this change?

edgargabriel:
That's a good point, but I would ask maybe @jsquyres to chime in on that: to the best of knowledge, the Open MPI philosophy with MCA parameters was always to follow the request of the user, even if this leads to an error. In this particular situation, you are absolutely right that if the user would disable the cycle_buffer_size feature by setting it to -1, his code might fail. However, after this merge, this would only happen if he explicitly requested that behavior.

Your suggestion would basically overwrite the request of the user.  

A second more practical difficulty is, that there is to the best of my knowledge no predefined constant in header files that we could just use. There is RLIMIT_FSIZE which indicates the maximum file size overall, the posix write manpages talks about SSIZE_MAX, both of which are however way too large. So this would require developing the entire configure logic etc. to make it portable. I can do that if the community decides that this is the way to go, but in that case this fix might come too late for the 2.0.2 release.

jsquyres:
We talked about this in email.  The decision was:

* This PR is a good fix for the initial problem.  It seems to give good enough performance and fix the bug.
* That being said, the 512MB buffer size has not been fully evaluated.  We need to keep working on a proper solution for the long term -- that will likely involve either fully-vetting a fixed size for the default, and/or evaluating using "-1" as the default and having appropriate loops to ensure completion of blocking and non-blocking writes.

In short: more work still needs to be done -- this is a short-term fix while this more detailed work is ongoing.

PR #2426, opened 2016-11-16
Performance Degradation in Derived Datatype creation - Open MPI

AboorvaDevarajan:
There is a performance degradation in ompi - derived datatype creation under a specific corner case, when compared to MPICH.

It occurs in OpenMPI MPI_Type_vector - create,  whenever the allocated DDTs are not freed simultaneously.

This suspected case of performance degradation can be reproduced by,
```c
#include "mpi.h"
#include <stdio.h>

#define free_loop 200000
#define main_tv_loop 600000

{
    int rank, size, i=0, j=0;
    double t_start = 0, t_end = 0, t_tv_start=0,t_tv_end=0,t_tv_total=0,t_commit_start=0,t_commit_end=0,t_commit_total=0,t_free_start=0,t_free_end=0,t_free_total=0;

    MPI_Datatype type[main_tv_loop], type2;
    MPI_Status status;

    MPI_Init(&argc, &argv);
   
    while(i<main_tv_loop) {
        for (j=0;j<free_loop;j++){
           
            t_tv_start = MPI_Wtime();
            MPI_Type_vector( 24, 24,  3000 , MPI_LONG, &type[j] );
            t_tv_total += (MPI_Wtime() - t_tv_start);
           
            MPI_Type_commit(&type[j]);
            i++;
            if(i == main_tv_loop)
                break;
        }
        for (j=0;j<free_loop;j++){
            MPI_Type_free(&type[j]);

            if(i == main_tv_loop)
                break;
         }
    } 
    printf("time taken type_vector : %0.3f s \n",  (t_tv_total));

    MPI_Finalize();
    return 0;
}
```
**Results:**

**MPICH:**
```
[c712f5n12][/u/abdevara/ompi-tv]> mpicc tv.c -o tv
[c712f5n12][/u/abdevara/ompi-tv]> mpirun -np 2 tv
time taken type_vector : 0.457 s
time taken type_vector : 0.460 s
```

**OMPI:**
```
[c712f5n12][/u/abdevara/ompi-tv]> mpicc tv.c -o tv
[c712f5n12][/u/abdevara/ompi-tv]> mpirun -np 2 tv
time taken type_vector : 116.760 s
time taken type_vector : 116.774 s
```
This is specifically due to the insertion of temporary datatype in the list `ompi_datatype_f_to_c_table`, which could be skipped [ompi/datatype/ompi_datatype_create_vector.c L:57 ](   https://github.com/open-mpi/ompi/blob/master/ompi/datatype/ompi_datatype_create_vector.c#L57)

Here is a possible fix for the issue,
https://github.com/AboorvaDevarajan/ompi/commit/f1ae4a814668c3a484fcbd0037bac09ed19646c1

By skipping the addition of temporarily created datatype to `ompi_datatype_f_to_c_table`, we get this result with the same test,

**OMPI:**

```
[c712f5n12][/u/abdevara/ompi-tv]> mpicc tv.c -o tv
[c712f5n12][/u/abdevara/ompi-tv]> mpirun -np 2 tv
time taken type_vector : 1.439 s 
time taken type_vector : 1.445 s 
```

Extended `opal_object` to accept a custom construtor during object creation, using the macro 
```
OBJ_NEW_CONSTRUCTOR(type, constructor) 
```
where, 
```
type - Type of the object.
constructor - Function pointer to the custom constructor.
```

This could also be done using the same macro OBJ_NEW and extend its functionality to accept variable arguments (variadic macro)

### DIFF
--- a/ompi/datatype/ompi_datatype.h
+++ b/ompi/datatype/ompi_datatype.h
@@ -120,6 +120,7 @@ OMPI_DECLSPEC int32_t ompi_datatype_default_convertors_fini( void );
 
 OMPI_DECLSPEC void ompi_datatype_dump (const ompi_datatype_t* pData);
 OMPI_DECLSPEC ompi_datatype_t* ompi_datatype_create( int32_t expectedSize );
+OMPI_DECLSPEC ompi_datatype_t* ompi_datatype_create_temporary( int32_t expectedSize );
 
 static inline int32_t
 ompi_datatype_is_committed( const ompi_datatype_t* type )
--- a/ompi/datatype/ompi_datatype_create.c
+++ b/ompi/datatype/ompi_datatype_create.c
@@ -26,16 +26,27 @@
 #include "ompi/datatype/ompi_datatype.h"
 #include "ompi/attribute/attribute.h"
 
-static void __ompi_datatype_allocate( ompi_datatype_t* datatype )
+__opal_attribute_always_inline__  static void __ompi_datatype_base_allocate( ompi_datatype_t* datatype )
 {
     datatype->args               = NULL;
-    datatype->d_f_to_c_index     = opal_pointer_array_add(&ompi_datatype_f_to_c_table, datatype);
     /* Later generated datatypes will have their id according to the Fortran ID, as ALL types are registered */
     datatype->id                 = datatype->d_f_to_c_index;
     datatype->d_keyhash          = NULL;
     datatype->name[0]            = '\0';
     datatype->packed_description = NULL;
     datatype->pml_data           = 0;
+}
+
+static void __ompi_datatype_allocate( ompi_datatype_t* datatype )
+{
+    datatype->d_f_to_c_index     = opal_pointer_array_add(&ompi_datatype_f_to_c_table, datatype);
+    __ompi_datatype_base_allocate(datatype);
+}
+
+static void __ompi_datatype_temporary_allocate( ompi_datatype_t* datatype )
+{
+    datatype->d_f_to_c_index     = -1;
+    __ompi_datatype_base_allocate(datatype);
 }
 
 static void __ompi_datatype_release(ompi_datatype_t * datatype)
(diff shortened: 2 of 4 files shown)


nysal:
@bosilca Can you please review this enhancement?

bosilca:
With SC on, I didn't had the time to thoughtfully analyze the patch, but I don't think we can skip the c2f index creation for Fortran types. However, instead of always registering the datatype with the c2f array upon datatype creation, we can only do it in the MPI_Type_c2f (with the exception of predefined datatypes).

ggouaillardet:
@bosilca my understanding is that this patch only optimizes temporary ddt (e.g. they are created, use to build the ddt the user requested, and then deleted), so there is no way this type can be used by the end user (and hence there cannot be a need for the Fortran index)

@AboorvaDevarajan ```OBJ_NEW_CONSTRUCTOR``` is one option, an other option is to declare a new class
(e.g. ```ompi_temp_datatype_t``` "inherits" from ```ompi_datatype_t``` and has the ```__ompi_datatype_temporary_allocate()``` constructor, or the other way around).
as far as i am concerned, that would be more consistent with what we currently do in other places.

nysal:
@ggouaillardet Thanks for your comments. Yes, the issue we are trying to fix is only for temporary data types. In this case, just the ones used in vector/hvector. I believe there might be a few other places that might benefit too, such as collective algorithms that create a temporary data type (Bruck comes to mind), where we really don't expose the data type to the application. If ompi_temp_datatype_t inherits from ompi_datatype_t, OBJ_NEW will still call all the base class constructors right? If so, then that might not help. We might have to inherit from opal_datatype_t and duplicate ompi_datatype_t struct declaration. We could alternatively make this a little more elegant by making OBJ_NEW a variadic macro and then try to handle it (Not sure we can do this portably). This would also mirror the C++ feature of being able to call different constructors in a class, which is what ompi tries to mimic. Please let us know your comments/suggestions.

nysal:
Another option is if we can really defer the insertion of the data type to MPI_Type_c2f, like George mentioned. Then that would be a bigger win, as we improve performance for all cases of data type creation.

nysal:
@ggouaillardet yes, that is a bit more compact. 

@bosilca  We'll take a shot at the MPI_Type_c2f() approach and see how it goes. 
Additionally, I see some potential for MPI_THREAD_MULTIPLE improvements for intrinsic data types in fortran too. Maybe worth exploring that as well. PMPI_Type_f2c calls opal_pointer_array_get_item() in the fast path, which involves taking a lock. If we cache the f2c conversion for predefined data types in a static array, we can avoid the lock for this case.

bosilca:
@nysal sounds good.

ibm-ompi:

 Gist: https://gist.github.com/ibm-ompi/56872b8a6e4819a04736383bc540bc5a

jjhursey:
bot:ibm:retest
(Network issue on relay machine)

ibm-ompi:

 Gist: https://gist.github.com/e4dee3aaab66238d7ccc5629b081fb6c

bosilca:
A different solution has been integrated into master ( #3369 ).

PR #2439, opened 2016-11-18
Allow the SLURM PMI components to operate in a Flux environment

rhc54:


rhc54:
no longer required - flux is adding its own component

PR #2441, opened 2016-11-21
v2.x: coll/libnbc: fix race condition with multi threaded apps

ggouaillardet:
protect the mca_coll_libnbc_component.active_requests list with
the new mca_coll_libnbc_component.lock mutex.

Thanks Jie Hu for the report

(cherry picked from commit open-mpi/ompi@2c94a3a6f398e187dff67386e9e1a7fe4bc26f15)


hppritcha:
@bosilca can you review when you get a chance?  Also #2442

jjhursey:
@ggouaillardet I'm curious why you didn't just reuse `mca_coll_libnbc_component.progress_lock`? Since the progress function is already protected by that lock then you would not have to grab another lock in there, but instead just protect the `opal_list_append`.

bosilca:
There is an ongoing discussion about this on #2443.

PR #2445, opened 2016-11-21
oshmem: memheap: removes find_offset

alex-mikheev:
@jladd-mlnx @yosefe @miked-mellanox 
Reasons for removal are:
- the function is only used by the shmem_lock code
- only a subset of the function is used by the shmem_lock
- for the general case the function is not correct


yosefe:
:+1:

yosefe:
@alex-mikheev pls  port to v2.x as well

PR #2448, opened 2016-11-22
Completely removed ompi_request_lock and ompi_request_cond

thananon:
ompi_request_lock and ompi_request_cond was meant to be completely removed from the request refactoring. This PR removed all of them.  We don't need to lock before calling ompi_request_complete() anymore. 

The removal will prevent the confusion and wrong assumption in the future. 
However there's still some part in iboffload and cm_request_free that I'm not familiar with the code base. So I changed the code the way I think it should be. Please review.

### DIFF
--- a/ompi/mca/bcol/iboffload/bcol_iboffload_component.c
+++ b/ompi/mca/bcol/iboffload/bcol_iboffload_component.c
@@ -931,9 +931,8 @@ static int progress_one_device(mca_bcol_iboffload_device_t *device)
 
             if (coll_request->n_frag_mpi_complete ==
                             coll_request->n_fragments) {
-                coll_request->super.req_complete = true;
-                opal_condition_broadcast(&ompi_request_cond);
-                IBOFFLOAD_VERBOSE(10, ("After opal_condition_broadcast.\n"));
+                OPAL_ATOMIC_SWAP_PTR(&coll_request->super.reg_complete, REQUEST_COMPLETED);
+                IBOFFLOAD_VERBOSE(10, ("After request completion.\n"));
             }
 
             rc = handle_collfrag_done(coll_frag, coll_request, device);
--- a/ompi/debuggers/ompi_mpihandles_dll.c
+++ b/ompi/debuggers/ompi_mpihandles_dll.c
@@ -84,10 +84,6 @@ struct mpidbg_name_map_t *mpidbg_status_name_map = NULL;
  * cases like the above with compilers that require the symbol (like
  * Sun Studio) we add in these definitions here.
  */
-size_t ompi_request_completed;
-opal_condition_t ompi_request_cond;
-size_t ompi_request_waiting;
-opal_mutex_t ompi_request_lock;
 opal_mutex_t opal_event_lock;
 int opal_progress_spin_count;
 bool opal_mutex_check_locks;
--- a/ompi/debuggers/ompi_msgq_dll.c
+++ b/ompi/debuggers/ompi_msgq_dll.c
@@ -1060,10 +1060,6 @@ static void dump_request( mqs_taddr_t current_item, mqs_pending_operation *res )
 }
 
 /**
- * TODO: ompi_request_completed can be used to detect any changes in the request handles.
- */
-
-/**
  * Handle the send queue as well as the receive queue. The unexpected queue
  * is a whole different story ...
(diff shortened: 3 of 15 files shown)


ompi/mca/bcol/iboffload/bcol_iboffload_component.c
```diff
@@ -931,9 +931,8 @@ static int progress_one_device(mca_bcol_iboffload_device_t *device)
 
             if (coll_request->n_frag_mpi_complete ==
                             coll_request->n_fragments) {
-                coll_request->super.req_complete = true;
-                opal_condition_broadcast(&ompi_request_cond);
-                IBOFFLOAD_VERBOSE(10, ("After opal_condition_broadcast.\n"));
+                OPAL_ATOMIC_SWAP_PTR(&coll_request->super.reg_complete, REQUEST_COMPLETED);
```
abouteiller: This has lost the cond bcast semantic. Probably it needs an update_sync instead.

bosilca: I think the original code already had an issue. The request is marked as completed and the waiting thread is informed. As the request is an MPI request, it will eventually be freed by the calling thread. However, few lines below, the coll_request is passed to handle_collfrag_done, which also put the request in an internal freelist, without checking the status of the user level request.

This component has no owner and is not maintained. I would propose we scrap it out completely.

rhc54: I agree - let's suggest it at next telecon and see if there are any objections

rhc54:
@thananon @bosilca Can we commit this and then (after soak) bring it to the v2.x? I'd like to see if it solves the performance issues over there.

thananon:
@rhc54 I am onto testing this patch on my side right now. It takes some time because we dont have MXM locally here.

jsquyres:
@thananon How's it going on this PR?

thananon:
There are some parts of the code that needs discussion because it is obsolete, unused and clearly wrong. We might as well remove that part but its not our teritory.

@bosilca told me to wait on the weekly telecon  discussion.

I'm not sure about the status.

jsquyres:
@bosilca What do we need to discuss on this issue on the weekly telecon?

bosilca:
We need to decide what we do with iboffload.

rhc54:
Unless someone raises their hand to take responsibility for it, I'd say we remove it - there is precedence for such action.

jsquyres:
IIRC, `bcol` (which contains `iboffload`) and `sbgp` are *only* used by `coll/ml`, which is ompi_ignored.  ORNL has been unwilling to un-ignore it / maintain it properly.

Perhaps this is the nail in the coffin for which we should remove:

* bcol
* sbgp
* coll/ml

?

shamisp:
Probably you want to check with ORNL folks - @manjugv

jsquyres:
@manjugv hasn't been active in a long while with Open MPI, and I couldn't find a github ID for Geoffroy Vallee -- so I emailed him off-list asking him to comment here.  😄

manjugv:
@jsquyres @rhc54 Thanks for asking. iboffload can be removed.  We have worked on other parts (coll/ml, sbgp) internally at ORNL last year, on improving the initialization performance and scalability with encouraging results. Unfortunately, I personally don’t have the cycles to push it or maintain the code. So, I can’t justify keeping this code as a part of main tree, particularly, if it is causing issues for others and if we don’t have volunteers to maintain it. 

As a courtesy, I would also ask @hjelmn 's opinion.

thananon:
If we decide to go with iboffload removal, this PR can be merged right away. We can have another PR to remove iboffload.

hppritcha:
I am okay with removal of ibofflaod.

rhc54:
I'm going to go ahead and commit this, then, and add the PR to remove the other areas so that @matcabral can see if this resolves the problem.

jsquyres:
@thananon @bosilca @manjugv @gvallee iboffload (and friends) have been removed.  This PR is now clear to rebase / merge.

jsquyres:
Hah -- never mind, it's already been merged.  😄 

@rhc54 @matcabral Now that this has been merged, it would be interesting to see if this affects the performance measured in #2644.

PR #2451, opened 2016-11-23
master: Add arguments to rpmbuild script and update README.

martinkontsek:
We have updated buildrpm script with external arguments. Script was tested on RedHat 7.3

### DIFF
--- a/contrib/dist/linux/README
+++ b/contrib/dist/linux/README
@@ -8,7 +8,7 @@ Copyright (c) 2004-2006 High Performance Computing Center Stuttgart,
                         University of Stuttgart.  All rights reserved.
 Copyright (c) 2004-2006 The Regents of the University of California.
                         All rights reserved.
-Copyright (c) 2006-2009 Cisco Systems, Inc.  All rights reserved.
+Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
 $COPYRIGHT$
 
 Additional copyrights may follow
@@ -24,9 +24,9 @@ version after the dash in the SRPM filename.  For example,
 v1.6.3.  Subsequent releases of SRPMs typically contain bug fixes for
 the RPM packaging, but not Open MPI itself.
 
-The buildrpm.sh script takes a single argument -- a filename pointing
-to an Open MPI tarball (may be either .gz or .bz2).  It will create
-one or more RPMs from this tarball:
+The buildrpm.sh script takes a single mandatory argument -- a filename
+pointing to an Open MPI tarball (may be either .gz or .bz2).  It will
+create one or more RPMs from this tarball:
 
 1. Source RPM
 2. "All in one" RPM, where all of Open MPI is put into a single RPM.
@@ -36,8 +36,62 @@ one or more RPMs from this tarball:
    - openmpi-devel
    - openmpi-docs
 
-The prefix, target architecture, and choice of RPM(s) to build are all
-currently hard-coded in the beginning of the buildrpm.sh script.
+The folowing arguments could be used to affect script behaviour.
+Please, do NOT set the same settings with parameters and config vars.
+
+-b
+   If you specify this option, only the all-in-one binary RPM will
+   be built. By default, only the source RPM (SRPM) is built. Other
+   parameters that affect the all-in-one binary RPM will be ignored
+   unless this option is specified.
+
+-n name
+   This option will change the name of the produced RPM to the "name".
+   It is useful to use with "-o" and "-m" options if you want to have
+   multiple Open MPI versions installed simultaneously in the same
+   enviroment. Requires use of option "-b".
+
+-o
+   With this option the install path of the binary RPM will be changed
+   to /opt/_NAME_/_VERSION_. Requires use of option "-b".
+
+-m
+   This option causes the RPM to also install modulefiles
+   to the location specified in the specfile. Requires use of option "-b".
+
+-i
+   Also build a debuginfo RPM. By default, the debuginfo RPM is not built.
+   Requires use of option "-b".
+
+-f lf_location
+   Include support for Libfabric. "lf_location" is Libfabric install
+   path. Requires use of option "-b".
+
+-t tm_location
+   Include support for Torque/PBS Pro. "tm_location" is path of the
+   Torque/PBS Pro header files. Requires use of option "-b".
+
+-d
+   Build with debugging support. By default,
+   the RPM is built without debugging support.
+
+-c parameter
+   Add custom configure parameter.
+
+-r parameter
+   Add custom RPM build parameter.
+
+-s
+   If specified, the script will try to unpack the openmpi.spec
+   file from the tarball specified on the command line. By default,
+   the script will look for the specfile in the current directory.
+
+-h
+   Prints script usage information.
+
+
+Target architecture is currently hard-coded in the beginning
+of the buildrpm.sh script.
 
 Alternatively, you can build directly from the openmpi.spec spec file
 or SRPM directly.  Many options can be passed to the building process
--- a/contrib/dist/linux/buildrpm.sh
+++ b/contrib/dist/linux/buildrpm.sh
@@ -3,7 +3,63 @@
 # Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
 #                         University Research and Technology
 #                         Corporation.  All rights reserved.
-# Copyright (c) 2006      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
+#
+
+#
+# External script parameters
+# The folowing parameters could be used to affect script behaviour.
+# Please, do NOT set the same settings with parameters and config vars.
+#
+# -b
+#    If you specify this option, only the all-in-one binary RPM will
+#    be built. By default, only the source RPM (SRPM) is built. Other
+#    parameters that affect the all-in-one binary RPM will be ignored
+#    unless this option is specified.
+#
+# -n name
+#    This option will change the name of the produced RPM to the "name".
+#    It is useful to use with "-o" and "-m" options if you want to have
+#    multiple Open MPI versions installed simultaneously in the same
+#    enviroment. Requires use of option "-b".
+#
+# -o
+#    With this option the install path of the binary RPM will be changed
+#    to /opt/_NAME_/_VERSION_. Requires use of option "-b".
+#
+# -m
+#    This option causes the RPM to also install modulefiles
+#    to the location specified in the specfile. Requires use of option "-b".
+#
+# -i
+#    Also build a debuginfo RPM. By default, the debuginfo RPM is not built.
+#    Requires use of option "-b".
+#
+# -f lf_location
+#    Include support for Libfabric. "lf_location" is Libfabric install
+#    path. Requires use of option "-b".
+#
+# -t tm_location
+#    Include support for Torque/PBS Pro. "tm_location" is path of the
+#    Torque/PBS Pro header files. Requires use of option "-b".
+#
+# -d
+#    Build with debugging support. By default,
+#    the RPM is built without debugging support.
+#
+# -c parameter
+#    Add custom configure parameter.
+#
+# -r parameter
+#    Add custom RPM build parameter.
+#
+# -s
+#    If specified, the script will try to unpack the openmpi.spec
+#    file from the tarball specified on the command line. By default,
+#    the script will look for the specfile in the current directory.
+#
+# -h
+#    Prints script usage information.
 #
 
 #
@@ -11,8 +67,6 @@
 # The following vars can be set from outside and will affect script behave:
 # prefix,rpmbuild_options,configure_options,build_srpm,build_single,build_multiple,rpmtopdir
 #
-
-
 specfile="openmpi.spec"
 prefix=${prefix:-"/opt/openmpi"}
 rpmbuild_options=${rpmbuild_options:-"--define 'mflags -j4' --define '_source_filedigest_algorithm md5'  --define '_binary_filedigest_algorithm md5'"}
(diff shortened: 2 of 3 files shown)


contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
 
-The prefix, target architecture, and choice of RPM(s) to build are all
-currently hard-coded in the beginning of the buildrpm.sh script.
+The folowing arguments could be used to affect script behaviour.
+Please, do NOT set the same settings with parameters and config vars.
+
+-b
+   If you specify this option, all-in-one binary RPM would be created
+   instead of source RPM (SRPM). This option is necessary for other
+   parameters to work, as they won't be influence SRPM package.
```
jsquyres: I'm not sure what this 2nd sentence means ("This option is necessary...")?

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+
+-n <name>
+   This option will change the name of the produced RPM to the <name>.
+   It is useful to use with "-o" and "-m" options if you want to have
+   multiple OpenMPI versions installed simultaneously in the same
```
jsquyres: "Open MPI" (it's 2 words 😄 ).

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+   enviroment.
+
+-o
+   With this option the install path of the binary RPM will be changed
+   to /opt/<name>/<version>
+
+-m
+   RPM would also install modulefiles to location specified in RPM
+   spec file.
```
jsquyres: Suggestion: "This option causes the RPM to also install modulefiles to the location specified..."

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+
+-d
+   Script will also build debuginfo RPM.
```
jsquyres: Suggestion: "Also build a debuginfo RPM (the default is to *not* build a debuginfo RPM)."

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+
+-u
+   Build RPM with usNIC support. It will include all the recommended
+   compile options for usNIC as well.
+
+-f <lf_location>
+   Include support for LibFabric. <lf_location> is path of the LibFabric
```
jsquyres: It's just "Libfabric" (no capitol F).

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+   header files.
+
+-t <tm_location>
+   Include support for Torque. <tm_location> is path of the Torque
```
jsquyres: Torque and PBS Pro.

contrib/dist/linux/README
```diff
@@ -36,8 +36,53 @@ one or more RPMs from this tarball:
...
+   header files.
+
+-s
+   If specified, script will try to unpack openmpi.spec file
+   from tarball
```
jsquyres: I'm not sure what this means -- do you mean that it will unpack the tarball specified on the command line and use the openmpi.spec file that is in the tarball?  And if this option is not specified, where does it expect to find the openmpi.spec file?

contrib/dist/linux/buildrpm.sh
```diff
@@ -45,12 +88,69 @@ build_multiple=${build_multiple:-"no"}
...
+# save original parameters
+#
+orig_param="$@"
+
+#
+# usage information
+#
+usage="Usage: $0 [-b][-o][-m][-d][-u][-h] [-n <name>][-f <lf_location>][-t <tm_location>] <tarball>
+
+	-b                do NOT build SRPM, only all-in-one binary RPM (required for other flags to work)
```
jsquyres: What other flags require `-b`?

contrib/dist/linux/buildrpm.sh
```diff
@@ -80,6 +180,14 @@ unset first
 echo "--> Found Open MPI version: $version"
 
 #
+# Try to unpack spec file from tarball
+#
+
+if test "$unpack_spec" = "1"; then
```
jsquyres: Super minor nit, but something I try to propagate to shell coders... 😄 

When using numerical values in shell variables:
* You can assign the value without quotes: `foo=1`
* You can test the value without quotes (i.e., without promoting the integer to a string) and use `test` operators such as `-eq`, `-ne`, etc.  E.g.: `test $foo -eq 1`

contrib/dist/linux/buildrpm.sh
```diff
@@ -45,12 +88,69 @@ build_multiple=${build_multiple:-"no"}
...
+#
+while getopts bn:omduf:t:sh flag; do
+    case "$flag" in
+      b) build_srpm="no"
+         build_single="yes"
+         ;;
+      n) rpmbuild_options="$rpmbuild_options --define '_name $OPTARG'"
+         ;;
+      o) rpmbuild_options="$rpmbuild_options --define 'install_in_opt 1'"
+	 configure_options="$configure_options --enable-mpirun-prefix-by-default --enable-orterun-prefix-by-default"
```
jsquyres: These two options are synonyms of each other -- you probably only need to list one of them.

contrib/dist/linux/buildrpm.sh
```diff
@@ -7,12 +7,55 @@
...
+#
+# -m
+#    RPM would also install modulefiles to location specified in RPM spec file.
+#
+# -d
+#    Script will also build debuginfo RPM.
+#
+# -u
+#    Build RPM with usNIC support. It will include all the recommended compile
+#    options for usNIC as well.
```
jsquyres: As much as I appreciate the Cisco-specific option here 😄 , it seems like we should either do this for other vendors, or we shouldn't do it for any of them.  I'd personally be ok removing this Cisco-specific feature and relegating that kind of functionality to build scripts here inside Cisco.

Anyone else have an opinion on this?

contrib/dist/linux/buildrpm.sh
```diff
@@ -45,12 +88,69 @@ build_multiple=${build_multiple:-"no"}
...
+      o) rpmbuild_options="$rpmbuild_options --define 'install_in_opt 1'"
+	 configure_options="$configure_options --enable-mpirun-prefix-by-default --enable-orterun-prefix-by-default"
+         ;;
+      m) rpmbuild_options="$rpmbuild_options --define 'install_modulefile 1'"
+         ;;
+      d) rpmbuild_options="$rpmbuild_options --define 'build_debuginfo_rpm 1'"
+         ;;
+      u) configure_options="$configure_options --with-usnic --without-memory-manager --enable-mca-no-build=btl-openib,common-verbs,oob-ud --enable-mpi-ext=all"
+         ;;
+      f) configure_options="$configure_options --with-libfabric=$OPTARG \"LDFLAGS=-Wl,--build-id -Wl,-rpath -Wl,$OPTARG/lib64 -Wl,--enable-new-dtags\""
```
jsquyres: I'm not sure that `-Wl,$OPTARG/lib64` is a good idea here.  What if `$OPTARG/lib64` does not exist?

ibm-ompi:

 Gist: https://gist.github.com/ibm-ompi/2746fecf92fda4d80ff1b80b23edbfc7

jjhursey:
bot:ibm:retest
(Network issue on relay machine)

contrib/dist/linux/README
martinkontsek: I have fixed the description. It should be a bit better now.

Basically, all the parameters influence only build of the  binary RPM. If user build only Source RPM, those parameters won't have any effect as they are not modifying specfile.

contrib/dist/linux/README
martinkontsek: I am sorry. Fixed.

contrib/dist/linux/README
martinkontsek: Thanks for suggestion.

contrib/dist/linux/README
martinkontsek: Thanks for suggestion.

contrib/dist/linux/README
martinkontsek: I am sorry. Fixed. It just looks better in my opinion 😄 (same with previous suggestion for Open MPI name).

contrib/dist/linux/README
martinkontsek: Thank you.

contrib/dist/linux/README
martinkontsek: Reworded. Thank you for pointing out.

contrib/dist/linux/buildrpm.sh
martinkontsek: I have removed the option, just in case.

contrib/dist/linux/buildrpm.sh
martinkontsek: I have added comments to all parameters.

contrib/dist/linux/buildrpm.sh
martinkontsek: Which option do you suggest to leave there? Why there are 2 synonymous options? Is it for legacy purposes?

contrib/dist/linux/buildrpm.sh
martinkontsek: Thank you. I have fixed the code.

contrib/dist/linux/buildrpm.sh
martinkontsek: Yes, you are right. It will cause problems if the `$OPTARG/lib64` doesn't exist.

I have 2 ideas how to fix this:
1. Create another parameter, which will be used for specifying lib directory. It will be used together with `-f` parameter. For example:
buildrpm.sh -b -f /opt/libfabric/1.4.0 -l /opt/libfabric/1.4.0/lib64

2. Have only `-f` (-f /opt/libfabric/1.4.0) and try to find out in code, if there is `lib*` subdirectory. If not, Error would be shown and script will stop.

I like 2nd idea more. Which idea do you prefer? Or do you have different idea how to solve it? Thank you

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
 Please, do NOT set the same settings with parameters and config vars.
 
 -b
-   If you specify this option, all-in-one binary RPM would be created
-   instead of source RPM (SRPM). This option is necessary for other
-   parameters to work, as they won't be influence SRPM package.
+   If you specify this option, only all-in-one binary RPM would be build.
+   By default, only source RPM (SRPM) is build. This option is necessary
+   for all other parameters to work, as they work only with binary RPM.
```
jsquyres: I apologize, but the English language is totally terrible -- it's full of special case grammar rules, especially in this paragraph.  Here's probably how it should read:

> If you specify this option, only the all-in-one binary RPM will be built.  By default, only the source RPM (SRPM) is built.  Other parameters that affect the all-in-one binary RPM will be ignored unless this option is specified.

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
 -o
    With this option the install path of the binary RPM will be changed
-   to /opt/<name>/<version>
+   to /opt/_NAME_/_VERSION_. Requires use of option "-b".
 
 -m
-   RPM would also install modulefiles to location specified in RPM
-   spec file.
+   This option causes the RPM to also install modulefiles
+   to the location specified in specfile. Requires use of option "-b".
```
jsquyres: ...to the location in **the** specfile.

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
 
--d
-   Script will also build debuginfo RPM.
+-i
+   Also build a debuginfo RPM. By default, debuginfo RPM is not build.
+   Requires use of option "-b".
```
jsquyres: By default, **the** debufinfo RPM is not **built**.

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
+
+-f lf_location
+   Include support for Libfabric. "lf_location" is path of the Libfabric
+   header files. Requires use of option "-b".
```
jsquyres: I think lf_location needs to be the prefix where Libfabric is installed, right?  I.e., it needs to be `/path/to/libfabric/installation`, not `/path/to/libfabric/installation/include`, right?

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
 
--u
-   Build RPM with usNIC support. It will include all the recommended
-   compile options for usNIC as well.
+-t tm_location
+   Include support for Torque/PBS Pro. "tm_location" is path of the
+   Torque/PBS Pro header files. Requires use of option "-b".
```
jsquyres: Same comment as Libfabric comment, above.

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
+
+-d
+   Build with debugging support. By default,
+   RPM is build without debugging support.
```
jsquyres: By default, the RPM is built without debugging support.

contrib/dist/linux/README
```diff
@@ -40,42 +40,50 @@ The folowing arguments could be used to affect script behaviour.
...
-   header files.
+-r parameter
+   Add custom RPM build parameter.
 
 -s
-   If specified, script will try to unpack openmpi.spec file
-   from tarball
+   If specified, script will try to unpack openmpi.spec file from
+   tarball specified on the command line. By default, script is
+   looking for specfile in current directory. Requires use of option "-b".
```
jsquyres: If specified, the script will try to unpack the openmpi.spec file from the tarball specified on the command line.  By default, the script will look for the specfile in the current directory.

contrib/dist/linux/buildrpm.sh
```diff
@@ -12,45 +12,55 @@
 # Please, do NOT set the same settings with parameters and config vars.
 #
 # -b
-#    If you specify this option, all-in-one binary RPM would be created
-#    instead of source RPM (SRPM). This option is necessary for other
-#    parameters to work, as they won't be influence SRPM package.
+#    If you specify this option, only all-in-one binary RPM would be build.
+#    By default, only source RPM (SRPM) is build. This option is necessary
+#    for all other parameters to work, as they work only with binary RPM.
```
jsquyres: Same comments here in buildrpm.h as in README.

contrib/dist/linux/buildrpm.sh
```diff
@@ -95,48 +105,79 @@ orig_param="$@"
...
 #
-usage="Usage: $0 [-b][-o][-m][-d][-u][-h] [-n <name>][-f <lf_location>][-t <tm_location>] <tarball>
+usage="Usage: $0 [-b][-o][-m][-d][-u][-s][-h] [-n name][-f lf_location][-t tm_location] tarball
+
+  -b
+             build all-in-one binary RPM only (required for all other flags to work)
+             {default: build only SRPM}
+
+  -n name  
+             name of the resulting RPM package set to "name". Requires "-b" flag.
```
jsquyres: Is the quoting problematic here?  I.e., you're closing the usage string and then re-opening it.

contrib/dist/linux/buildrpm.sh
```diff
@@ -95,48 +105,79 @@ orig_param="$@"
...
+
+  -m         install modulefiles during RPM installation. Requires "-b" flag.
+             {default: modulefiles will NOT be installed}
+
+  -i         build debuginfo RPM. Requires "-b" flag.
+             {default: do NOT build debuginfo RPM}
+
+  -f lf_location  
+             include Libfabric support from <lf_location>. Requires "-b" flag.
+             {default: build without Libfabric support}
```
jsquyres: I think the default is to build Libfabric support only if it is found in default paths, right?  (i.e., it will *try* to build Libfabric support, but it won't error if Libfabric support is not found)

contrib/dist/linux/buildrpm.sh
```diff
@@ -95,48 +105,79 @@ orig_param="$@"
...
-	-m                create modulefiles
-	-d                build debuginfo RPM
-	-u                include usNIC support
-	-f <lf_location>  include LibFabric support from <lf_location>
-	-t <tm_location>  include Torque support from <tm_location>
-	-h                print this message and exit
-	-s                try to unpack openmpi.spec file from <tarball>
+  -t tm_location  
+             include Torque/PBS Pro support from "tm_location". Requires "-b" flag.
+             {default: build without Torque/PBS Pro}
```
jsquyres: Same comment as Libfabric, above.

contrib/dist/linux/buildrpm.sh
jsquyres: Yes, it's mainly for legacy / backwards compatibility purposes.

contrib/dist/linux/buildrpm.sh
jsquyres: In the Open MPI configury, we handle this complicated/sucky situation, but it takes a bunch of code.  😦 

* Test for the presence of `$dir/lib64`.  If that's there, use it.
* Test for the presence of `$dir/lib`.  If that's there, use it.
* If you didn't find either of them, error.

This is not bullet proof, either, because the above scheme clearly assumes a 64 bit build.  But what if the user also passed in relevant compiler/linker options to build in 32 bit?  It's a rat hole.  ☹️   So Open MPI *also* has a secondary, optional param for each package.  See https://github.com/open-mpi/ompi/blob/master/README#L874-L887.

martinkontsek: Thank you. I have adjusted script accordingly.

contrib/dist/linux/README
martinkontsek: I am sorry. As I'm not a native speaker, I often forget about English grammar :-) Thanks.

contrib/dist/linux/README
martinkontsek: Thanks.

contrib/dist/linux/README
martinkontsek: Thanks.

contrib/dist/linux/README
martinkontsek: Yes, you are right, fixed.

contrib/dist/linux/README
martinkontsek: Well, actually, at least in our Torque installation, path to devel include files is completely different than Torque install directory.

contrib/dist/linux/README
martinkontsek: Thanks.

contrib/dist/linux/README
martinkontsek: Thanks.

contrib/dist/linux/buildrpm.sh
martinkontsek: Thanks.

contrib/dist/linux/buildrpm.sh
martinkontsek: Oh, I am sorry. This was really a problem.

contrib/dist/linux/buildrpm.sh
martinkontsek: Yes, you are right.

contrib/dist/linux/buildrpm.sh
martinkontsek: Yes, you are right.

martinkontsek:
I have squashed all the pull request suggestions into first commit.

jsquyres:
Thanks @martinkontsek

PR #2453, opened 2016-11-23
ompi/mpi_init: lazy-wait during PMIx fence execution

artpol84:
Relax CPU usage pressure from the application processes when doing
modex and barrier in ompi_mpi_init.

We see significant latencies in SLURM/pmix plugin barrier progress
because app processes are aggressively call opal_progress pushing
away daemon process doing collective progress.

(ported from 08618845a4de7ce7010f3ad0cb65aa7b4bf39fe1 and 06a73da5ea62992527995303c3c159de12db332d)


artpol84:
@rhc54 @hppritcha @jladd-mlnx please review

PR #2456, opened 2016-11-24
fcoll/dynamic_gen2: fix bug exposed by uneven distribution of data

edgargabriel:
This fixes a bug reported in-house occuring with this component. It is triggered if the data assigned to different aggregators is highly differing, leading to different number of internal iterations required to handle it.


edgargabriel:
bot:lanl:retest

PR #2457, opened 2016-11-24
Fixed -Werror=unused-result warnings in comm_cid.c by adding error checking

OMGtechy:
This is my first commit (and therefore pull request), so please let me know if I've not followed any of your conventions.

This pull request does the following:
- Fixes -Werror=unused-result warnings in comm_cid.c
- Adds error checking for when asprintf fails

### DIFF
--- a/ompi/communicator/comm_cid.c
+++ b/ompi/communicator/comm_cid.c
@@ -840,6 +840,9 @@ static int ompi_comm_allreduce_pmix_reduce_complete (ompi_comm_request_t *reques
     opal_pmix_pdata_t pdat;
     opal_buffer_t sbuf;
     int rc;
+    int bytes_written;
+    const int output_id = 0;
+    const int verbosity_level = 1;
 
     OBJ_CONSTRUCT(&sbuf, opal_buffer_t);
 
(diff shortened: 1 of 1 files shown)


hppritcha:
@hjelmn please review

ompi/communicator/comm_cid.c
```diff
@@ -840,6 +840,7 @@ static int ompi_comm_allreduce_pmix_reduce_complete (ompi_comm_request_t *reques
     opal_pmix_pdata_t pdat;
     opal_buffer_t sbuf;
     int rc;
+    int bytesWritten;
```
jsquyres: Minor nit: we don't typically use camelCase in Open MPI; we typically use multiple_words.  This is pretty trivial -- we would not reject a patch just for this reason, though -- I mention it just for "style fit" reasons.

ompi/communicator/comm_cid.c
```diff
@@ -858,16 +859,36 @@ static int ompi_comm_allreduce_pmix_reduce_complete (ompi_comm_request_t *reques
...
+    }
+
+    if (bytesWritten == -1) {
+        // write with separate calls,
+        // just in case the args are the cause of failure
+        fprintf (stderr, "send first: %d\n", cid_context->send_first);
+        fprintf (stderr, "port string: %s\n", cid_context->port_string);
+        fprintf (stderr, "pmix tag: %s\n", cid_context->pmix_tag);
+        fprintf (stderr, "iter: %d\n", cid_context->iter);
+        return OMPI_ERR_FATAL;
```
jsquyres: You can probably return `OMPI_ERR_OUT_OF_RESOURCE` here; it should be treated as a fatal error by the caller.

ompi/communicator/comm_cid.c
```diff
@@ -858,16 +859,36 @@ static int ompi_comm_allreduce_pmix_reduce_complete (ompi_comm_request_t *reques
...
-                       cid_context->iter);
+    bytesWritten = asprintf(&info.key,
+                            cid_context->send_first ? "%s:%s:send:%d"
+                                                    : "%s:%s:recv:%d",
+                            cid_context->port_string,
+                            cid_context->pmix_tag,
+                            cid_context->iter);
+
+    if (bytesWritten == -1) {
+        fprintf (stderr, "writing info.key failed\n");
```
jsquyres: We don't typically `fprintf()` from deep within the library.  Instead, we typically use `opal_output()` if something needs to unconditionally be written to stderr (because it will prefix the output with a label indicating the process it came from).  For errors like this, however, we typically will use `opal_output_verbose()`, which lets the error only be displayed conditionally (i.e., if the user has selected a high enough verbose value at run time), or just let the caller decide whether to display the error or not.

OMGtechy: Thank you! I used fprintf because other bits of code around here used it (`fprintf (stderr, "pack failed. rc  %d\n", rc);` for example). Would it be worth me fixing that in another pull request, or is there a specific reason in that case?

jjhursey:
bot:ibm:retest
(I'm trying to bring the IBM CI back online)

ompi/communicator/comm_cid.c
jsquyres: We used `fprintf` elsewhere in here?  Oh, how embarrassing...  😳 

Yes, if you could have a 2nd commit on this PR that fixes all the other `fprintf`s, that would be great.  Thanks!

ompi/communicator/comm_cid.c
OMGtechy: Fixed in 7fda6fe5a7c55ec3b1e057a8fc63afd0491f0bde

ompi/communicator/comm_cid.c
OMGtechy: Fixed in df356adfe6140fc82a9764a68196a777593ad2b4

ompi/communicator/comm_cid.c
OMGtechy: Fixed in 8c7da1a39db3e25df99320862102f079d6d15e62

hjelmn:
Please squish.

jsquyres:
Sweet; thanks!

One last request: can you squash down to just two commits?

* replace fprintf calls with opal_output_verbose
* everything else

That way, we don't have a commit followed by multiple followup commits to fix the elements identified in the review.

OMGtechy:
@jsquyres I have squashed it all into one commit as it was easier, and semantically speaking these are all one change. Let me know if that's a problem :)

jsquyres:
Works for me; thanks.  Will merge when CI finishes.

PR #2462, opened 2016-11-27
oshmem: memheap: removes find_offset

alex-mikheev:
@miked-mellanox @yosefe @jladd-mlnx 
Reasons for removal are:
- the function is only used by the shmem_lock code
- only a subset of the function is used by the shmem_lock
- for the general case the function is not correct

(cherry picked from commit d1723355d31e7a625daabf6a64cd2084a5d76c6a)


yosefe:
:+1:

PR #2466, opened 2016-11-28
pml/ob1: Reduce per-rank memory footprint slightly

kawashima-fj:
`sturct mca_pml_ob1_comm_proc_t`, which is allocated per
connected rank in a communicator, had two paddings after
`expected_sequence` and `send_sequence` by alignments.
By changing the order of the members, the size of
`mca_pml_ob1_comm_proc_t` is reduced by 8 bytes on 64-bit
architectures.


jsquyres:
Do you want to PR these to v2.0.x and v2.x?

kawashima-fj:
No. This is a very small memory footprint improvement thanks to the dynamic add_procs feature. If someone wants this in v2.x, I'll make a PR.

PR #2469, opened 2016-11-29
coll/hcoll: hcoll_context_free

vspetrov:

    @bureddy @jladd-mlnx Plz review.


bureddy:
:+1:

PR #2471, opened 2016-11-29
v2.x: coll/hcoll: hcoll_context_free

vspetrov:


jsquyres:
This looks like the v2.x version of #2469.

jsquyres:
@hppritcha Good to go.

PR #2474, opened 2016-11-29
v2.0.x: coll/hcoll: hcoll_context_free

vspetrov:

    v2.0.x version of #2469 


jsquyres:
@vspetrov Github pro tip: you're indenting your message in the description, which causes Github you render it in the fixed-width font (i.e., fixed indenting is Github Markdown for "put this in verbatim mode").  If you don't indent, Github will render the text normally.

Also, we figured out on the call this morning that you weren't a member of the open-mpi Github org; I added you sometime in the last 30 mins or so.  I think you have to click through an email, and then you'll have access to set labels and milestones and whatnot on issues/PRs.

jsquyres:
@hppritcha Good to go when CI finishes.

PR #2477, opened 2016-11-30
oob/v2.x/msg drop

artpol84:
The problem was observed for direct modex used with recursive doubling
algorithm (used for collective ID calculation prior to d52a2d081e9598a9ac9a50fb4b013a6d2a72375b)
that has pairwise nature and counter-connections are highly likely.

The following scenario was uncovering the issue:
* ranks `x` and `y` want to communicate with each other, `x` < `y`;
* rank `x` initiates the connection and sends the ack;
* rank `y` starts to `connect()` and gets the ack from `x`;
* `y` identifies that it already started connecting and `y` > `x` so it rejects incoming connection.
* `x` sees that his connection was rejected in `mca_oob_tcp_peer_recv_connect_ack()` when trying to
read the message header using `tcp_peer_recv_blocking()` which calls `mca_oob_tcp_peer_close()`
that effectively flushes all the messages in the peer->send_queue.
* `y` send the ack to `x` and the connection is established, however all the messages for the peer
at `x` are vanished (except the front one in peer->send_msg).

This commit introduces a "nack" function that will be used at `y` side to tell `x` that `y` has the
priority and `x`'s connection should be closed. This allows to avoid "guessing" on the unexpectedly
closed connection.

(cherry-picked from ada93e0c026df8f28b9078d47af25c74eb398656)

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -862,6 +862,7 @@ int mca_oob_tcp_peer_recv_connect_ack(mca_oob_tcp_peer_t* pr,
              */
             mca_oob_tcp_peer_close(peer);
         }
+        free(msg);
         return ORTE_ERR_UNREACH;
     }
 
@@ -876,6 +877,7 @@ int mca_oob_tcp_peer_recv_connect_ack(mca_oob_tcp_peer_t* pr,
           MCA_OOB_TCP_CONNECTING == peer->state ||
          MCA_OOB_TCP_CONNECT_ACK == peer->state ) ) {
         if (retry(peer, sd, false)) {
+            free(msg);
             return ORTE_ERR_UNREACH;
         }
     }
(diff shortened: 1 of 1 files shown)


rhc54:
Please include open-mpi/ompi@30ff8be9c98da040c264b15d11b0e2827ecc0651

artpol84:
@rhc54 - I'm not sure who's "sign-off" should be there. I'll change it if not correct.

rhc54:
It doesn't matter, so long as someone's is there

ibm-ompi:

 Gist: https://gist.github.com/777cab00ef8baee35cadb0eb7e52a618

artpol84:
@jjhursey looks like an fs failure.

jjhursey:
bot:ibm:retest

jsquyres:
@rhc54 Can you review?  Thanks.

jsquyres:
@hppritcha Good to go.

artpol84:
@hppritcha any updates on this? We would like to get it in so we won't see MTT failures.

PR #2484, opened 2016-11-30
orte/oob/tcp: Plug the memory leak.

artpol84:
Plug coverity defect CID 1396541.


artpol84:
@rhc54 please, have a look. It's a simple thing.

PR #2488, opened 2016-12-01
Topic/misc fixes

ggouaillardet:


ibm-ompi:

 Gist: https://gist.github.com/ibm-ompi/2746fecf92fda4d80ff1b80b23edbfc7

ibm-ompi:

 Gist: https://gist.github.com/ibm-ompi/56872b8a6e4819a04736383bc540bc5a

ibm-ompi:

opal/mca/btl/openib/btl_openib_mca.c
```diff
@@ -641,6 +641,9 @@ int btl_openib_register_mca_params(void)
         /* Don't try to recover from this */
         return OPAL_ERR_OUT_OF_RESOURCE;
     }
+    if (NULL != mca_btl_openib_component.default_recv_qps) {
+        free(mca_btl_openib_component.default_recv_qps);
+    }
     mca_btl_openib_component.default_recv_qps = default_qps;
```
hjelmn: Just an FYI. free (NULL) is valid in C. No need to check for NULL, just free it.

ggouaillardet: thanks @hjelmn
iirc, we do this (useless) test virtually everywhere else, so i did it again to keep some consistency in the code.

maybe we should all agree on a policy about that, for example
- keep testing vs ```NULL``` before ```free()```
- stop testing vs ```NULL``` before ```free()``` from now
- remove all tests vs ```NULL``` before ```free()``` (that could be quite a massive change)

@jsquyres @rhc54 @bosilca @hppritcha any thoughts ?

bosilca: I would rather save the function call if there is no need for it and if the call is in the critical path. Otherwise, exert your freedom of choice.

jsquyres: Yeah, I think it was mostly a side effect of a conservative style, and to help with debugging (i.e., if we *never* call `free()` (and friends) with a NULL pointer, then if OPAL intercepts ever saw `free()` with NULL, it was a bug).

PR #2494, opened 2016-12-02
v2.0.x: misc coll fixes

ggouaillardet:


jsquyres:
@hppritcha I think that this is good to go for v2.0.2.

PR #2496, opened 2016-12-02
Always provide a SIGTERM before SIGKILL to allow procs a chance to cleanup

rhc54:
Thanks to Noel Rycroft for the report


rhc54:
Replaced by #2529

PR #2498, opened 2016-12-02
Ensure that we always send a SIGTERM prior to SIGKILL to give child processes a chance to cleanup

rhc54:
Thanks to Noel Rycroft for the report

### DIFF
--- a/orte/mca/odls/base/odls_base_default_fns.c
+++ b/orte/mca/odls/base/odls_base_default_fns.c
@@ -1367,13 +1367,30 @@ typedef struct {
 
 static void kill_cbfunc(int fd, short args, void *cbdata)
 {
-    odls_kill_caddy_t *cd = (odls_kill_caddy_t*)cbdata;
+    orte_timer_t *tm = (orte_timer_t*)cbdata;
+    odls_kill_caddy_t *cd = (odls_kill_caddy_t*)tm->payload;
 
-    if (!ORTE_FLAG_TEST(cd->child, ORTE_PROC_FLAG_ALIVE) || 0 == cd->child->pid) {
-        free(cd);
-        return;
-    }
+    OPAL_OUTPUT_VERBOSE((5, orte_odls_base_framework.framework_output,
+                         "%s SENDING FORCE SIGKILL TO %s",
+                         ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                         ORTE_NAME_PRINT(&cd->child->name)));
+
     cd->kill_local(cd->child->pid, SIGKILL);
+    /* indicate the waitpid fired as this is effectively what
+     * has happened
+     */
+    ORTE_FLAG_SET(cd->child, ORTE_PROC_FLAG_WAITPID);
+    cd->child->pid = 0;
+
+    /* ensure the child's session directory is cleaned up */
+    orte_session_dir_finalize(&cd->child->name);
+    /* check for everything complete - this will remove
+     * the child object from our local list
+     */
+    if (ORTE_FLAG_TEST(cd->child, ORTE_PROC_FLAG_IOF_COMPLETE) &&
+        ORTE_FLAG_TEST(cd->child, ORTE_PROC_FLAG_WAITPID)) {
+        ORTE_ACTIVATE_PROC_STATE(&cd->child->name, cd->child->state);
+    }
     free(cd);
 }
 
@@ -1387,6 +1404,7 @@ int orte_odls_base_default_kill_local_procs(opal_pointer_array_t *procs,
     int i, j;
     opal_pointer_array_t procarray, *procptr;
     bool do_cleanup;
+    odls_kill_caddy_t *cd;
 
     OBJ_CONSTRUCT(&procs_killed, opal_list_t);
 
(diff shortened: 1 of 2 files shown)


orte/mca/odls/default/odls_default_module.c
```diff
@@ -761,9 +761,9 @@ static int send_signal(pid_t pid, int signal)
                          signal, (long)pid));
 
     if (orte_forward_job_control) {
```
gpaulsen: Why did you change the last if (orte_forward_job_control) on line 395 to #if HAVE_SETPGID, but left this if(orte_forward_job_control)?  was this intentional?

orte/mca/odls/base/odls_base_default_fns.c
```diff
@@ -1503,53 +1521,30 @@ int orte_odls_base_default_kill_local_procs(opal_pointer_array_t *procs,
...
+               running, then SIGTERM will not get delivered.  Ignore return
+               value. */
             OPAL_OUTPUT_VERBOSE((5, orte_odls_base_framework.framework_output,
-                                 "%s SENDING FORCE SIGKILL TO %s pid %lu",
+                                 "%s SENDING SIGCONT TO %s",
                                  ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
-                                 ORTE_NAME_PRINT(&child->name), (unsigned long)child->pid));
-            kill_local(child->pid, SIGKILL);
+                                 ORTE_NAME_PRINT(&child->name)));
+            kill_local(child->pid, SIGCONT);
```
gpaulsen: Do we need a small delay between sending the SIGCONT and the SIGTERM?  It is possible to set a SIGCONT handler, so perhaps a very small delay would be useful to allow that to finish if set.

jsquyres:
I have a reply from Noel (the user who reported the issue) -- I provided him with a bootstrapped tarball of this PR.

-----

That doesn't seem to fix the issue unfortunately..

What I did:
* Unpack, build, and install Open MPI
    * downloaded and unpacked tarball
    * ./configure --prefix <my install path>
    * make
    * make install
* Build reproducer against new version
    * I updated the MPI_ROOT variable in the Makefile of the reproducer
    * make clean to remove old objects and remove the sleep.openmpi executable
    * make to rebuild sleep.openmpi executable
* Run 'sleep' reproducer against new version
    * set the OMPI_HOME variable to the new install
    * run `sleeptest.sh 30`
    * send SIGTERM to the top level script `kill -TERM 14866`

I've annotated the output below..

```sh
ncr@dougie.lebanon.example.com:~/data/RUN/377/openmpi> ./sleeptest.sh 30
[14866] : Installing signal handler for signal : INT
[14866] : Installing signal handler for signal : TERM
[14866] : Installing signal handler for signal : 0
[14866] : Parallel Sleep Test MPI
# this (i.e., the next line) is the mpirun command line that is launched by the top level sleeptest.sh script
[14866] : /users/ncr/data/RUN/377/openmpi/openmpiinstall/gitclone/bin/mpirun --prefix /users/ncr/data/RUN/377/openmpi/openmpiinstall/gitclone -np 2 sleep.sh 30 &                                                
# this is the output from the first MPI tasks (sleep.sh)
[14872] : Installing signal handler for signal : INT
[14872] : Installing signal handler for signal : TERM
[14872] : Installing signal handler for signal : 0
# the actual MPI application binary is launched
[14872] : ./sleep.openmpi 30                                 
# this is the output from the second MPI tasks (sleep.sh)
[14873] : Installing signal handler for signal : INT        
[14873] : Installing signal handler for signal : TERM
[14873] : Installing signal handler for signal : 0
# the actual MPI application binary is launched
[14873] : ./sleep.openmpi 30                                 
# the two MPI ranks start up and the application will go into a do/while loop sleeping
# for 1 second until total is reached (in this case 30)
Hello world! I'm 0 out of 2 on dougie.lebanon.example.com  
My rank = 0 PID = 14874                                      
0 sleeping for 1 seconds
Hello world! I'm 1 out of 2 on dougie.lebanon.example.com
My rank = 1 PID = 14875
1 sleeping for 1 seconds
2 sleeping for 1 seconds
3 sleeping for 1 seconds
4 sleeping for 1 seconds
5 sleeping for 1 seconds
6 sleeping for 1 seconds
7 sleeping for 1 seconds
8 sleeping for 1 seconds
9 sleeping for 1 seconds
10 sleeping for 1 seconds
# This is where the top level script receives SIGTERM and propagates it to its child,
# which is mpirun
[14866] : sleeptest.sh Terminated                            
[14866] : Killing child : 14867                               
[14866] : kill -TERM 14867
11 sleeping for 1 seconds
12 sleeping for 1 seconds
13 sleeping for 1 seconds
# The MPI binary application continues which implies that SIGTERM was never propagated
# to it. If I look at the process tree I can see that the child scripts (sleep.sh) have
# been killed and the binaries are orphaned. I also don't see any output showing that the 
# child sleep.sh scripts propagated SIGTERM to the MPI application binaries.
14 sleeping for 1 seconds                                    
15 sleeping for 1 seconds                                    
16 sleeping for 1 seconds                                    
17 sleeping for 1 seconds                                    
18 sleeping for 1 seconds
19 sleeping for 1 seconds
20 sleeping for 1 seconds
21 sleeping for 1 seconds
22 sleeping for 1 seconds
23 sleeping for 1 seconds
24 sleeping for 1 seconds
25 sleeping for 1 seconds
26 sleeping for 1 seconds
27 sleeping for 1 seconds
28 sleeping for 1 seconds
29 sleeping for 1 seconds
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.

  Local host:  dougie.lebanon.example.com
  System call: unlink(2) /tmp/openmpi-sessions-1014@dougie_0/16719/1/1/vader_segment.dougie.1
  Error:       No such file or directory (errno 2)
[dougie.lebanon.example.com:14867] 1 more process has sent help message help-opal-shmem-mmap.txt / sys call fail
[dougie.lebanon.example.com:14867] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
[14866] : Normal Exit : 1
```

rhc54:
I think we've had a complete communication failure. The problem I resolved was the one where mpirun spawns a child, and that child then spawns its own children. This PR solves that problem by ensuring that (a) signals sent by mpirun are given to the child plus its children, and (b) we provide a time delay between SIGTERM and a subsequent SIGKILL.

What the user is testing is something very different, where a script that captures SIGTERM is launching mpirun. They then are hitting mpirun their script with SIGTERM, which is supposed to then hit mpirun with SIGTERM - and waiting to see mpirun propagate it to the spawned procs.

I have no idea how that is going to work as there are multiple signal traps occurring. Regardless, that isn't what I was trying to fix, so I suggest we treat that as a separate use-case and issue.

orte/mca/odls/base/odls_base_default_fns.c
rhc54: feel free :-)  we've never bothered to do so before.

gpaulsen: yeah, i doubt anyone ever does that.

noelrycroft:
I re-tested by sending SIGTERM directly to mpirun.

In Open MPI 1.8.4 mpirun must be propagating SIGTERM to the tasks because I can see verbosity from my task scripts which indicate that they've received SIGTERM and then go-ahead and propagate to their children.

**1.8.4**

```
24 sleeping for 1 seconds
# sleep.sh is a child task of mpirun
[12612] : sleep.sh Terminated
# sleep.sh propagates SIGTERM to it's child
[12612] : Killing child : 12614
[12612] : kill -TERM 12614
[12612] : Normal Exit : 143
# this is the 2nd sleep.sh MPI task which also receives SIGTERM etc...
[12613] : sleep.sh Terminated
[12613] : Killing child : 12615
[12613] : kill -TERM 12615
[12613] : Normal Exit : 143
```

In 2.0.1 I see different behaviour and the child tasks never receive SIGTERM.

```
14 sleeping for 1 seconds
mpirun: Forwarding signal 18 to job
# this indicates the sleep binary exec continues to execute and di not receive SIGTERM
15 sleeping for 1 seconds
16 sleeping for 1 seconds
17 sleeping for 1 seconds
18 sleeping for 1 seconds
19 sleeping for 1 seconds
20 sleeping for 1 seconds
21 sleeping for 1 seconds
22 sleeping for 1 seconds
23 sleeping for 1 seconds
24 sleeping for 1 seconds
25 sleeping for 1 seconds
26 sleeping for 1 seconds
27 sleeping for 1 seconds
```

I see identical behaviour between version 2.0.1 and the tarball that Jeff sent to me..

rhc54:
I've reviewed the 1.8 code and I see the difference. I think there may be a quick fix to resolve the problem. Let me take a quick peek.

rhc54:
Replaced by https://github.com/open-mpi/ompi/pull/2528

PR #2509, opened 2016-12-03
Fixed memory leak and some -Werror=unused-result warnings

OMGtechy:
Signed-off-by: Joshua Gerrard <enquiries@joshuagerrard.com>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -165,6 +165,7 @@ Josh Hursey, Indiana University, Oak Ridge National Laboratory, Los Alamos Natio
   jjhursey@open-mpi.org
 Joshua Gerrard
   enquiries@joshuagerrard.com
+  joshuagerrard+ompi-commit@protonmail.com
 Joshua Ladd, Mellanox
   jladd.mlnx@gmail.com
   joshual@mellanox.com
--- a/ompi/errhandler/errhandler_predefined.c
+++ b/ompi/errhandler/errhandler_predefined.c
@@ -177,43 +177,84 @@ static void backend_fatal_aggregate(char *type,
                                     char *name, int *error_code,
                                     va_list arglist)
 {
-    char *arg, *prefix, *err_msg = "Unknown error";
-    bool err_msg_need_free = false;
+    char *arg = NULL, *prefix = NULL, *err_msg = NULL;
+    const char* const unknown_error_code = "Error code: %d (no associated error message)";
+    const char* const unknown_error = "Unknown error";
+    const char* const unknown_prefix = "[?:?]";
+
+    // these do not own what they point to; they're
+    // here to avoid repeating expressions such as
+    // (NULL == foo) ? unknown_foo : foo
+    const char* usable_prefix = unknown_prefix;
+    const char* usable_err_msg = unknown_error;
 
     arg = va_arg(arglist, char*);
     va_end(arglist);
 
-    asprintf(&prefix, "[%s:%d]", ompi_process_info.nodename,
-             (int) ompi_process_info.pid);
+    if (asprintf(&prefix, "[%s:%d]",
+                 ompi_process_info.nodename,
+                 (int) ompi_process_info.pid) == -1) {
+        prefix = NULL;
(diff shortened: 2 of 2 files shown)


AUTHORS
```diff
@@ -165,6 +165,8 @@ Josh Hursey, Indiana University, Oak Ridge National Laboratory, Los Alamos Natio
   jjhursey@open-mpi.org
 Joshua Gerrard
   enquiries@joshuagerrard.com
+  technopiano@gmail.com
+  joshuagerrard+ompi-commit@protonmail.com
```
jsquyres: @OMGtechy Are these the emails that you committed with?  We typically only show the emails that show up in `git log`.

ompi/errhandler/errhandler_predefined.c
```diff
@@ -177,21 +177,30 @@ static void backend_fatal_aggregate(char *type,
...
     arg = va_arg(arglist, char*);
     va_end(arglist);
 
-    asprintf(&prefix, "[%s:%d]", ompi_process_info.nodename,
-             (int) ompi_process_info.pid);
+    if (asprintf(&prefix, "[%s:%d]",
+                 ompi_process_info.nodename,
+                 (int) ompi_process_info.pid) == -1) {
+        // non-fatal, we could still go on to give useful information here...
+        opal_output(0, "%s", "Could not write node and PID to prefix");
```
jsquyres: Just curious -- why do the double-string form?  You shouldn't need to use the copy through %s -- you could just put the 3rd argument directly as the 2nd arg.

ompi/errhandler/errhandler_predefined.c
```diff
@@ -177,21 +177,30 @@ static void backend_fatal_aggregate(char *type,
...
 
     if (NULL != error_code) {
         err_msg = ompi_mpi_errnum_get_string(*error_code);
         if (NULL == err_msg) {
-            err_msg_need_free = true;
-            asprintf(&err_msg, "Error code: %d (no associated error message)",
-                     *error_code);
+            if (asprintf(&err_msg, unknown_error_code,
+                         *error_code) == -1) {
+                opal_output(0, "%s", "Could not write to err_msg");
```
jsquyres: Same comment as above -- %s doesn't seem to be needed...?

ompi/errhandler/errhandler_predefined.c
```diff
@@ -177,21 +177,30 @@ static void backend_fatal_aggregate(char *type,
...
+    }
 
     if (NULL != error_code) {
         err_msg = ompi_mpi_errnum_get_string(*error_code);
         if (NULL == err_msg) {
-            err_msg_need_free = true;
-            asprintf(&err_msg, "Error code: %d (no associated error message)",
-                     *error_code);
+            if (asprintf(&err_msg, unknown_error_code,
+                         *error_code) == -1) {
```
jsquyres: Will a failed `asprintf()` leave NULL in `err_msg`?  Or could it be some random value (that `free(err_msg)`, below, will choke on)?

ompi/errhandler/errhandler_predefined.c
OMGtechy: It's a habit to stop situations like this:

`opal_output(0, user_input, ...)`

Quite happy to change it if you want though :)

AUTHORS
OMGtechy: One of them is - I've updated accordingly.

ompi/errhandler/errhandler_predefined.c
OMGtechy: Good spot! I also made the same fix for `prefix`

OMGtechy:
> Sorry for the delay in reviewing -- December gets busy!

No worries! Your review comments should be addressed now.

ompi/errhandler/errhandler_predefined.c
jsquyres: FWIW, I think compilers only warn on that for libc functions (e.g., printf).  They don't for opal_output.  But it doesn't matter that much...

PR #2542, opened 2016-12-07
Correctly cleanup the local children and node map info on remote orte…

rhc54:
…ds upon job completion. Ensure that register_nspace only includes procs from that job in the proc map

Thanks to Ashley Pittman for the report


rhc54:
@ashleypittman

Issue #2550, opened 2016-12-09
early termination is not correctly handled in v2


ggouaillardet:
@rhc54 this is a followup of the issue initially reported at https://mail-archive.com/users@lists.open-mpi.org/msg30424.html

here is a simple reproducer

```c
#include <stdio.h>
#include <mpi.h>

int main(int argc, char *argv[]) {
    MPI_Init(&argc, &argv);
    int rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
    if (0 != rank) {
        MPI_Recv(NULL, 0, MPI_INT, 0, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
    }
    return 0;
}
```

the correct behaviour (master is fine) is to report an error since task 0 exited without calling ```MPI_Finalize()``` nor ```MPI_Abort()```

with both ```v2.x``` and ```v2.0.x```, the program
- exits normally if ran with only one MPI task
- hangs otherwise

can you please take care of that ?

hppritcha:
this problem was reported by Christof Koehler.  this is a note to help me with NEWS updates.

jsquyres:
Fix merged in v2.0.x and v2.x.  Done!

PR #2552, opened 2016-12-09
Set the registered/deregistered flags so we correctly detect exit without calling finalize

rhc54:
Fixes https://github.com/open-mpi/ompi/issues/2550


hppritcha:
Mellanox jenkins unresponsive lately.  Going ahead and merging.

PR #2553, opened 2016-12-09
Set the registered/deregistered flags so we correctly detect exit without calling finalize

rhc54:
Refs #2550 


jsquyres:
@hppritcha This was merged for v2.0.x, so I'm merging here.

PR #2558, opened 2016-12-12
ompi/datatype: fix bug in darray that causes MPI/IO failures

hjelmn:
This commit fixes errors in the lb and extent of darray datatypes. For
these datatypes the lb should be the start offset of the rank's data
in the array and the extent should be the size of the entire
datatype. In master the lb was always 0 and the extent was always to
small. This commit updates the call to opal_datatype_resize to set the
correct lb and fixes the extent calculation.

### DIFF
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -1,4 +1,4 @@
-/* -*- Mode: C; c-basic-offset:4 ; -*- */
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
@@ -13,6 +13,8 @@
  * Copyright (c) 2009      Sun Microsystems, Inc. All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -169,6 +171,7 @@ int32_t ompi_datatype_create_darray(int size,
     ptrdiff_t orig_extent, *st_offsets = NULL;
     int i, start_loop, end_loop, step;
     int *coords = NULL, rc = OMPI_SUCCESS;
+    ptrdiff_t displs[2], tmp_size = 1;
 
     /* speedy corner case */
     if (ndims < 1) {
@@ -187,10 +190,13 @@ int32_t ompi_datatype_create_darray(int size,
         int tmp_rank = rank, procs = size;
 
         coords = (int *) malloc(ndims * sizeof(int));
+        displs[1] = orig_extent;
         for (i = 0 ; i < ndims ; i++) {
             procs = procs / psize_array[i];
             coords[i] = tmp_rank / procs;
             tmp_rank = tmp_rank % procs;
+            /* compute the upper bound of the datatype, including all dimensions */
+            displs[1] *= gsize_array[i];
         }
     }
 
@@ -246,7 +252,6 @@ int32_t ompi_datatype_create_darray(int size,
         lastType = *newtype;
     }
 
-
     /**
      * We need to shift the content (useful data) of the datatype, so
      * we need to force the displacement to be moved. Therefore, we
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca The bug in v1.10.x is a bit less complex. Here is the patch destined for v1.10:

```
rom a8ac1c254fbb43441443e0e05bb152299c072a66 Mon Sep 17 00:00:00 2001
From: Howard Pritchard <hppritcha@gmail.com>
Date: Fri, 9 Dec 2016 15:12:32 -0700
Subject: [PATCH] datatypes: fix incorrect extent setting in create_darray

fixes a bug introduced with open-mpi/ompi-release#859

diff --git a/ompi/datatype/ompi_datatype_create_darray.c b/ompi/datatype/ompi_datatype_create_darray.c
index 6d394ca..0dc96fd 100644
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -278,7 +278,7 @@ int32_t ompi_datatype_create_darray(int size,
             rc = ompi_datatype_create_hindexed( 1, blength_hindexed, displs_hindexed, lastType, newtype);
         }
         ompi_datatype_destroy(&lastType);
-        opal_datatype_resize( &(*newtype)->super, 0, displs[1] );
+       opal_datatype_resize( &(*newtype)->super, displs[1], displs[2] );
         /* need to destroy the old type even in error condition, so
            don't check return code from above until after cleanup. */
         if (MPI_SUCCESS != rc) goto cleanup;
--
1.7.1
```

hjelmn:
I have a regression test that shows the error. Will add it to ompi-tests after this is in master.

hjelmn:
@hppritcha FYI. This is the master fix. Once this is in I will open the PRs for the release branches.

bosilca:
I would propose the following patch, based on your but integrated with the existing loops:

```diff
diff --git a/ompi/datatype/ompi_datatype_create_darray.c b/ompi/datatype/ompi_datatype_create_darray.c
index 56caa2a8b0..976e33faaf 100644
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -169,6 +169,7 @@ int32_t ompi_datatype_create_darray(int size,
     ptrdiff_t orig_extent, *st_offsets = NULL;
     int i, start_loop, end_loop, step;
     int *coords = NULL, rc = OMPI_SUCCESS;
+    ptrdiff_t displs[2], tmp_size = 1;

     /* speedy corner case */
     if (ndims < 1) {
@@ -187,10 +188,13 @@ int32_t ompi_datatype_create_darray(int size,
         int tmp_rank = rank, procs = size;

         coords = (int *) malloc(ndims * sizeof(int));
+        displs[1] = orig_extent;
         for (i = 0 ; i < ndims ; i++) {
             procs = procs / psize_array[i];
             coords[i] = tmp_rank / procs;
             tmp_rank = tmp_rank % procs;
+            /* compute the upper bound of the datatype, including all dimensions */
+            displs[1] *= gsize_array[i];
         }
     }

@@ -210,9 +214,14 @@ int32_t ompi_datatype_create_darray(int size,
     }

     /* Build up array */
+    displs[0] = st_offsets[start_loop];
     for (i = start_loop; i != end_loop; i += step) {
         int nprocs, tmp_rank;

+        /* Update the lower bound of the local type */
+        tmp_size *= gsize_array[i - step];
+        displs[0] += tmp_size * st_offsets[i];
+
         switch(distrib_array[i]) {
         case MPI_DISTRIBUTE_BLOCK:
             rc = block(gsize_array, i, ndims, psize_array[i], coords[i],
@@ -245,7 +254,7 @@ int32_t ompi_datatype_create_darray(int size,
         if (MPI_SUCCESS != rc) goto cleanup;
         lastType = *newtype;
     }
-
+    displs[0] *= orig_extent;  /* with the original datatype extent */

     /**
      * We need to shift the content (useful data) of the datatype, so
@@ -255,29 +264,19 @@ int32_t ompi_datatype_create_darray(int size,
      * new data, and insert the last_Type with the correct
      * displacement.
      */
-    {
-        ptrdiff_t displs[2], tmp_size = 1;
-
-        displs[0] = st_offsets[start_loop];
-        displs[1] = orig_extent;
-        for (i = start_loop + step; i != end_loop; i += step) {
-            tmp_size *= gsize_array[i - step];
-            displs[0] += tmp_size * st_offsets[i];
-            displs[1] *= gsize_array[i];
-        }
-        displs[0] *= orig_extent;
-
-        *newtype = ompi_datatype_create(lastType->super.desc.used);
-        rc = ompi_datatype_add(*newtype, lastType, 1, displs[0], displs[1]);
-        ompi_datatype_destroy(&lastType);
-        opal_datatype_resize( &(*newtype)->super, 0, displs[1] );
-        /* need to destroy the old type even in error condition, so
-           don't check return code from above until after cleanup. */
-        if (MPI_SUCCESS != rc) newtype = NULL;
+    *newtype = ompi_datatype_create(lastType->super.desc.used);
+    rc = ompi_datatype_add(*newtype, lastType, 1, displs[0], displs[1]);
+    ompi_datatype_destroy(&lastType);
+    /* need to destroy the old type even in error condition, so
+     * don't check return code from above until after cleanup. */
+    if (MPI_SUCCESS != rc) {
+        ompi_datatype_destroy (newtype);
+    } else {
+        (void)opal_datatype_resize( &(*newtype)->super, 0, displs[1] );
     }

  cleanup:
-    if (NULL != st_offsets) free(st_offsets);
-    if (NULL != coords) free(coords);
-    return OMPI_SUCCESS;
+    free(st_offsets);
+    free(coords);
+    return rc;
 }
```

hjelmn:
@bosilca I am fine with that. Will update now.

hjelmn:
Opps. @bosilca that doesn't work as written. st_offsets is not initialized where you are updating displs[0].

hjelmn:
At the very least we can calculate the extent in the loop @ L187. The lb calculation probably should probably be left where it was.

hjelmn:
@bosilca Added the change to the extent calculation. Passing my regression test.

hjelmn:
Tracking release issues. #2559, #2560, #2561.

hjelmn:
As far as I can tell true_lb *should* be the same as lb in this case. I added a test case to the ibm suite to check the lb and extent (see ibm/datatype/darray-pack.c). This is how it is in Open MPI releases prior to 2.0.0.

ggouaillardet:
@hjelmn my understanding of MPI 3.1 chapter 4.1 p99 (see the definition of ```cyclic``` is that the lower bound should be ```0``` _most of the time_

_most of the time_ means that this could not be true if ```oldtype``` has a negative lower bound.
(and upper bound might also differ if ```oldtype``` has a _very large_ upper bound)

hjelmn:
@ggouaillardet Makes sense. I don't care either way as long as we get it working. Updating the patch to set lb to 0 and true lb to displs[1].

hjelmn:
Looks like the true_lb is correct on master if we pass 0 for the lb. The issue that affected 1.10.x was apparently fixed. Updated the patch and now we get the same result as mpich.

bosilca:
In the MPI 3.1 page 99 line 38 it is explicitly stated that the special markers LB and UB must be set (these markers do not exist anymore, but their conceptual meaning is kept with the resize operation). Thus, lb must be 0, and the new extent must be gsize * extent(original_data) on every rank.

hjelmn:
@bosilca. Ok. Then the current PR is correct. Will merge once CI finishes.

Issue #2561, opened 2016-12-12
darray datatypes broken


hjelmn:
Distributed array datatypes give the wrong lower bound and extent.

Fix is already in testing. This bug is to track when the fix has been applied.

hppritcha:
This has been fixed by #2558, #2566, #2567.

PR #2567, opened 2016-12-13
ompi/datatype: fix bug in darray that causes MPI/IO failures

### LINKED ISSUE #2560: darray datatypes broken
Distributed array datatypes give the wrong lower bound and extent.

Fix is already in testing. This bug is to track when the fix has been applied.

hjelmn:
This commit fixes errors in the lb and extent of darray datatypes. For
these datatypes the lb should be the start offset of the rank's data
in the array and the extent should be the size of the entire
datatype. In master the lb was always 0 and the extent was always to
small. This commit updates the call to opal_datatype_resize to set the
correct lb and fixes the extent calculation.

Fixes #2560

(cherry picked from commit eb439228b12bd94315d4f00f4cb7875a71f69b09)


jsquyres:
@hppritcha Good to go once CI finishes.

hppritcha:
Fixes #2561

Issue #2572, opened 2016-12-14
PMIx: Cannot run simultaneous jobs by different users on the same node


jsquyres:
On master HEAD, it looks like PMIx is using a hard-coded file name as its contact point: `/tmp/pmix-contact.txt`.

This is a problem if multiple different users try to run a job on the same node at the same time.  One of them will be able to create/write `/tmp/pmix-contact.txt`, but the next one will not be able to.

For example, if my MTT is running on a bunch of nodes, but then I go to do some short tests on the same nodes:

```
$ mpirun --hostfile file_containing_mtt_servers hostname
[mpi021:10716] Impossible to open the file /tmp/pmix-contact.txt in write mode
[mpi021:10716] PMIX ERROR: PMIX_ERR_FILE_OPEN_FAILURE in file ptl_tcp_component.c at line 547
--------------------------------------------------------------------------
The PMIx server's listener thread failed to start. We cannot
continue.
--------------------------------------------------------------------------
[hang]
```

Notice too, that mpirun hangs -- it doesn't quit.

rhc54:
Huh - that shouldn't happen. We are supposed to pass the session dir head down to pmix, which should use that instead of /tmp. Probably something out-of-sequence in an update from all the various branches. I'll get it fixed this morning between meetings.

rhc54:
I guess I forgot to mark it "fixed" in the PR

PR #2575, opened 2016-12-14
Use the server tmpdir instead of the system tmpdir for tool contact files

rhc54:


rhc54:
Fixes #2572

jsquyres:
👍

PR #2577, opened 2016-12-14
Ensure jobs that fail always return a non-zero exit code.

rhc54:
Thanks to Ashley Pittman for the report.


rhc54:
@ashleypittman

Issue #2505, opened 2016-12-02
osc pt2pt wrong answer


markalle:
Using the testcase small_1sided.c from this test harness pull request:
    https://github.com/open-mpi/ompi-tests/pull/25
I get wrong answers from the osc pt2pt component:

```
% mpicc -o x small_1sided.c
% mpirun -host hostA,hostB -mca osc pt2pt -mca pml ob1 -mca btl openib,self,vader ./x
% mpirun -host hostA,hostB -mca osc pt2pt -mca pml ob1 -mca btl tcp,self,vader ./x
```

In cases where it's able to use osc rdma it seems to work.

This was from a vanilla build of openmpi-master-201612022109-0366f3a.

jsquyres:
@hjelmn Can you look at this one, too?  (perhaps related to #2530?)

jjhursey:
A couple more data points running `1sided.c` like this on a single node:
```shell
mpirun -np 2 -mca pml ob1 -mca osc pt2pt -mca btl VALUE ./small_1sided
```

`VALUE` | Result
------------ | -------------
 `self,sm` | **PASS**
 `self,vader` | **Wrong answer** (near the start of the test)
 `self,tcp` | **HANG** (near the start of the test)
 `self,openib` | **HANG**

For one of the hangs this is the backtrace (both processes are in the same location):
```gdb
(gdb) bt
#0  0x00003fff77da2e44 in clock_gettime () from /lib64/power8/libc.so.6
#1  0x00003fff7773d37c in gettime (base=0x10034a1f2e0, tp=0x10034a1f450) at event.c:370
#2  0x00003fff77741f40 in event_base_loop (base=0x10034a1f2e0, flags=<optimized out>) at event.c:1624
#3  0x00003fff778a38b8 in opal_progress () at ompi/opal/runtime/opal_progress.c:207
#4  0x00003fff74f9a990 in opal_condition_wait (c=0x10034bbac38, m=0x10034bbabd0)
    at ompi/opal/threads/condition.h:72
#5  0x00003fff74f9d0dc in ompi_osc_pt2pt_wait (win=0x10034bba9a0)
    at ompi/ompi/mca/osc/pt2pt/osc_pt2pt_active_target.c:519
#6  0x00003fff780c8318 in PMPI_Win_wait (win=0x10034bba9a0) at pwin_wait.c:52
#7  0x000000001000317c in main_test_fn ()
#8  0x0000000010004348 in runtest ()
[...]

PR #2590, opened 2016-12-16
Topic/osc pt2pt 1 thread fixes

jjhursey:
Fixes Issue #2505 for the single threaded case.

@hjelmn I would appreciate a review of these three commits before they go into `master`.

The multithreaded test case against the `osc/pt2pt` component is still failing with wrong answers, hangs, and segvs. We'll open another ticket for that.

Note that this commit should go into the `v2.x` and `v2.0.x` branches. If included in the `v2.0.2` release we should also put in a protection against using `osc/pt2pt` with `MPI_THREAD_MULTIPLE` as that currently can lead to wrong answers. We'll need a separate, `v2.0.x` temporary commit for that. I can work on generating that change.

### DIFF
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_active_target.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_active_target.c
@@ -10,7 +10,7 @@
  *                         All rights reserved.
  * Copyright (c) 2007-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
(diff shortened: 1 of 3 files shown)


jjhursey:
@hjelmn Thanks!

We have one more patch regarding lock ordering that might need some discussion. I'm preparing that now. We will bring that in as a separate PR since it might require some discussion (might turn into a MPI standard interpretation issue).

Issue #2592, opened 2016-12-16
Performance regression: placeholder for v2.0.2 release


jsquyres:
See #2591 -- that issue might also affect the v2.0.x series, so I wanted to open an issue on the v2.0.2 milestone to track it here, too.

It is potentially a blocker, but we'll know more when @hjelmn reports findings on #2591.

matcabral:
Hi @jsquyres,  I see we may have different issues and/or sources degradation  and impacts. As discussed in #1831, I have seen a latency and bw_mr degradation (especially in small messages) when moving from 1.10.4 to 2.0.x. in PSM2 and OFI MTLs. I've been looking at it but have found nothing concrete.  Saw an increase in cache misses and minor page faults, but suspect this is not the source of the degradation since vader btl shows the same percentage of misses but increased performance.

hppritcha:
closed via #2596, #2611, and #2612

matcabral:
Hi @hppritcha and @jsquyres, 
Would the scope of this issue be limited to the "clocks" degradation only? I'm asking because I'm still seeing the regression described in #1831 when comparing #2611 with #2599 (%5 to %13 latency degradation in osu_latency for message sizes<=4k running over PSM). 
thanks,

rhc54:
Yeah, I think you should probably open another issue to track that further problem. There are clearly several things going on, so let's focus this one on the clock problem.

matcabral:
opened #2644

Issue #2591, opened 2016-12-16
Performance regression in 1.10.3


jsquyres:
@hjelmn is working on identifying a performance regression that was introduced in 1.10.3.  He doesn't have any further info yet beyond the fact that one of their apps runs a bunch slower with 1.10.3 vs 1.10.2.

He thinks this will also affect v2.0.x and v2.x.

This is potentially a blocker for v1.10.5 and v2.0.2.

@hjelmn Says he'll have more information shortly.

hjelmn:
Bisect shows for 1.10 the regression starts with 9f2a6da4b266cbd5907b7bdf3ab14b62c61047ec. This commit forces the use of clock_gettime () over the rtdtsc instruction. IMHO this is a really bad idea. This forces us to call clock_gettime () on every single call to opal_progress ().

@bosilca We do not need a monotonic timer for opal_progress (), correct? If not we should force opal_progress to always use the asm timer and not clock_gettime ().

hjelmn:
@rhc54, @jladd-mlnx. I would think this would affect message rates as well. The pre-1.10.3 behavior can be forced using --mca timer_require_monotonic 0. Might be worth rerunning some of the performance regressions with this option set.

bosilca:
@hjelmn I think we can survive in opal_progress without a monotonic timer, in which case we might run the libevent progress at an erratic rate. However, this will force us to expose 2 timers, one that is monotonic and one that might not be, and we do not have right now the infrastructure.

hjelmn:
@bosilca. Yeah. Not sure how to handle that.

For now I can do something for newer Intel processors. There is a cpuid bit to indicate whether the cpu has a core invariant tsc. I am adding a function to check that bit. If it is set it will use the rtdtsc instruction as a monotonic timer.

bosilca:
👍

ggouaillardet:
Iirc, we moved to clock_gettime() because MPI_Wtime() could go back in time if a (thread of a) task migrates within a node.
@hjelmn do you plan to test the cpuid bit at configure time  or at run time ?
Iirc once again, we take several shortcuts in the timer, and it does not behave as a standard component.
Having two timers (one internal, one external and always monotonic) could be the most efficient and "portable" way

hjelmn:
[...]

PR #2596, opened 2016-12-16
opal/timer: add code to check if rtdtsc is core invariant

hjelmn:
Newer x86 processors have a core invariant tsc. On these systems it is
safe to use the rtdtsc instruction as a monotonic timer. This commit
adds a new function to the opal timer code to check if the timer
backend is monotonic. On x86 it checks the appropriate bit and on
other architectures it parrots back the OPAL_TIMER_MONOTONIC value.

### DIFF
--- a/opal/include/opal/sys/amd64/timer.h
+++ b/opal/include/opal/sys/amd64/timer.h
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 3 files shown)


hjelmn:
This is a enhancement that helps #2591 and #2592 when running on newer Intel processors where the tsc is core invariant. We still will need to find something for older Intel processors.

hppritcha:
@hjelmn have you tested this on KNL?

jsquyres:
bot:lanl:retest

hjelmn:
@hppritcha Haven't tested this on KNL yet. Being a newer processor I would hope the invariant tsc is available.

BTW, I now I have reports back from 3 code teams. They all report getting the same or better performance as 1.6.5 on TLCC2 systems when using the tsc timer. CTS-1 systems show a similar speedup.

PR #2599, opened 2016-12-17
opal/timer: add code to check if rtdtsc is core invariant

hjelmn:
Newer x86 processors have a core invariant tsc. On these systems it is
safe to use the rtdtsc instruction as a monotonic timer. This commit
adds a new function to the opal timer code to check if the timer
backend is monotonic. On x86 it checks the appropriate bit and on
other architectures it parrots back the OPAL_TIMER_MONOTONIC value.

(cherry picked from commit a718743a5c7017c13fb56950d16edd0b59f07108)

### DIFF
--- a/opal/include/opal/sys/amd64/timer.h
+++ b/opal/include/opal/sys/amd64/timer.h
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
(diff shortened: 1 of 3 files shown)


hjelmn:
@rhc54 This fixes as performance regression found in 1.10.3 that is cause LANL codes problems. The fix doesn't help with all Intel processors but helps with the latest iterations. I am working on a more complete fix but we need a 1.10.5 on Monday or Tuesday if possible.

rhc54:
can you get someone to review this by then?

hjelmn:
@rhc54 Sure. If George can't get to it I will see who I can find.

rhc54:
Fixes #2591

Issue #2616, opened 2016-12-20
Build failure with --with-hwloc=external


opoplawski:
If I try to build the Fedora openmpi package with --with-hwloc=external I end up with:
```
In file included from ../../opal/mca/hwloc/hwloc.h:131:0,
                 from proc.h:21,
                 from proc.c:19:
../../opal/mca/hwloc/external/external.h:20:35: fatal error: /include/hwloc.h: No such file or directory
 #include MCA_hwloc_external_header
                                   ^
compilation terminated.
```
Full logs - https://kojipkgs.fedoraproject.org//work/tasks/5738/17005738/build.log
It does work with --with-hwloc=/usr, but I noticed a post on the mailing list suggesting that that was bad practice.

ggouaillardet:
@opoplawski thanks for the report, i will be pushing a fix shortly

opoplawski:
Looks good here.  Thanks.

opoplawski:
So, it worked okay with the #2618 patch, but now with 2.0.2rc2 I get lots of errors:
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../opal/include -I../../../ompi/include -I../../../oshmem/include -I../../../opal/mca/hwloc/hwloc1112/hwloc/include/private/autogen -I../../../opal/mca/hwloc/hwloc1112/hwloc/include/hwloc/autogen -I../../../ompi/mpiext/cuda/c -I../../../opal/include -I../../.. -I../../../orte/include -I/builddir/build/BUILD/openmpi-2.0.2rc2/opal/mca/event/libevent2022/libevent -I/builddir/build/BUILD/openmpi-2.0.2rc2/opal/mca/event/libevent2022/libevent/include -DNDEBUG -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -finline-functions -fno-strict-aliasing -mcx16 -pthread -MT base/hwloc_base_dt.lo -MD -MP -MF base/.deps/hwloc_base_dt.Tpo -c base/hwloc_base_dt.c  -fPIC -DPIC -o base/.libs/hwloc_base_dt.o
In file included from ../../../opal/mca/hwloc/base/base.h:18:0,
[...]

PR #2617, opened 2016-12-21
hwloc: correctly handle --with-hwloc=external

ggouaillardet:
- simply #include "hwloc.h" to use the external hwloc header
- do use the external hwloc header instead of opal/mca/hwloc/hwloc.h

Thanks Orion Poplawski for the report

Fixes open-mpi/ompi#2616


hppritcha:
uff, hit milestone button by accident

Issue #1092, opened 2015-11-03
configure help wrong for --with-libltdl, --with-hwloc


loveshack:
configure --help says that --with-libltdl=external and --with-hwloc=external can be specified to use external libraries, but "external" is actually interpreted as a directory name in both cases, and configuration fails.

rhc54:
I'm not able to replicate that problem on the current 1.10.1 release candidate:

```
--- MCA component hwloc:external (m4 configuration macro, priority 90)
checking for MCA component hwloc:external compile mode... static
checking --with-hwloc-libdir value... simple ok (unspecified)
checking looking for external hwloc in... (default search paths)
checking hwloc.h usability... no
checking hwloc.h presence... no
checking for hwloc.h... no
checking if MCA component hwloc:external can compile... no
configure: WARNING: MCA component "external" failed to configure properly
configure: WARNING: This component was selected as the default
configure: error: Cannot continue
```

Looking at the configure log, it appears it did exactly what it was supposed to do:

```
configure:66802: checking for m4 configure components in framework hwloc
configure:66804: result: external, hwloc191
configure:66891: checking for MCA component hwloc:external compile mode
configure:66894: result: static
configure:66978: checking --with-hwloc-libdir value
configure:66981: result: simple ok (unspecified)
configure:67010: checking looking for external hwloc in
configure:67017: result: (default search paths)
configure:67079: checking hwloc.h usability
configure:67079: gcc -std=gnu99 -c -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -pthread   conftest.c >&5
conftest.c:568:19: fatal error: hwloc.h: No such file or directory
 #include <hwloc.h>
                   ^
compilation terminated.
configure:67079: $? = 1
```

Can you confirm what version of OMPI you were using?

loveshack:
rhc54 <notifications@github.com> writes:

> I'm not able to replicate that problem on the current 1.10.1 release candidate:

Strangely, I now can't either with hwloc -- apologies -- but I can with
ltdl:

```
$ ./configure --with-libltdl=external |& tail -n5
--- MCA component dl:libltdl (m4 configuration macro, priority 50)
checking for MCA component dl:libltdl compile mode... static
checking --with-libltdl value... not found
configure: WARNING: Directory external not found
configure: error: Cannot continue
```

[...]

PR #2622, opened 2016-12-21
v2.x: opal_configure_options: remove stale option help

jsquyres:
--with-libltdl is now added (via AC_ARG_WITH) in
opal/mca/dl/libltdl/configure.m4 -- it no longer belongs up here in
this top-level m4 file.  Plus, the help string in this stale entry is
also stale/incorrect.

(cherry picked from commit 678e314c0ef0325946b88878db487e73553f9982)

Refs #1092.


jsquyres:
bot:mellanox:retest

Issue #2614, opened 2016-12-20
osc/pt2pt fails with multiple threads


jjhursey:
Testing with the `osc/pt2pt` component revealed multiple hangs and wrong answers when running with two threads. Each thread is working with it's own communicator copy of `MPI_COMM_WORLD` and their own private windows.

The test is here:
 * https://github.com/open-mpi/ompi-tests/pull/25
 * https://gist.github.com/markalle/56f739c3c907e53960d22d5de2fe4bc9

```shell
% mpicc -o x mt_1sided.c mt_1sided_td1.c mt_1sided_td2.c
% mpirun -host hostA,hostB -mca osc pt2pt -mca pml ob1 -mca btl openib,self,vader ./x
```

PR #2630 will need to be reverted when a resolution to this issue is committed on v2.0.x branch.

hjelmn:
I can't get this to hang with osc/rdma. @markalle Was the 1sided.c test hanging with osc/rdma or just osc/pt2pt?

hjelmn:
Ok, I see what is happening. The restore of %ebx isn't working as expected. Looks like it might be selecting a register that is clobbered by cpuid. If the code is causing a problem here then it also is in the patcher. Will fix both.

hjelmn:
I see the problem. The code I copied for patcher only saves the lower 32 bits of b. Needs to save all 64 bits. Not sure why we are just discovering this now. Patch incoming.

hjelmn:
Wait, the other bit is a 32-bit only path. Thats why that worked. *face-palm*.

hjelmn:
Opps. Wrong bug :). Deleting those.

hjelmn:
Ok, I can reproduce the issue with osc/pt2pt. Looks like something is still not right with PSCW. Taking a look now.

hjelmn:
Found the bug. Its an artifact of the original design. I haven't had the time to move the counters into the sync object. Think I have a workaround. Testing it now.

hjelmn:
Ok, definitely fixed. Running it through the tests a couple more times to shake out any remaining bugs. Will have a PR open for master, v2.x, and v2.0.3.

hjelmn:
Still have one stubborn bug holding up the fix. It now hangs about 10% of the time. Will hopefully have this resolved today.

hppritcha:
@hjelmn can this issue be closed?

markalle:
[...]

PR #2626, opened 2016-12-21
Topic/osc controls

gpaulsen:
Adding mca parameter "enable" to enable / disable osc_rdma and osc_pt2pt. 
Adding mca parameter "allow_thread_multiple" to enable (default) / disable osc_rdma and osc_pt2pt when user passed in MPI_THREAD_MULTIPLE to MPI_Init_threads().

Due to the Issue #2614, we recommend that v2.0.2 gets released with the following additions to the etc/openmpi-mca-params.conf file:
```
osc_rdma_enable = 1
osc_rdma_allow_thread_multiple = 0
osc_pt2pt_enable = 1
osc_pt2pt_allow_thread_multiple = 0
```

This PR does NOT make any changes to the mca params.conf file, the release engineer should make that decision / change.

### DIFF
--- a/ompi/include/ompi/constants.h
+++ b/ompi/include/ompi/constants.h
@@ -10,6 +10,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2016      Intel, Inc. All rights reserved.
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -75,6 +76,7 @@ enum {
     OMPI_ERR_WIN                    = OMPI_ERR_BASE - 7,
     OMPI_ERR_RMA_FLAVOR             = OMPI_ERR_BASE - 8,
 };
+/* it'd be nice to have a new OMPI_ERR_DISABLED error code. */
 
 #define OMPI_ERR_MAX                    (OMPI_ERR_BASE - 100)
 
--- a/ompi/mca/osc/pt2pt/osc_pt2pt.h
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt.h
@@ -15,6 +15,7 @@
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      FUJITSU LIMITED.  All rights reserved.
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -88,6 +89,21 @@ struct ompi_osc_pt2pt_component_t {
 
     /** Is the progress function enabled? */
     bool progress_enable;
+
+    /** opal_output_verbose stream id */
+    int output;
+
+    /** MCA parameter: priority of this component */
+    int priority;
+
+    /** MCA parameter: Verbose level of this component */
+    int verbose;
+
+    /** MCA parameters: Enable osc_pt2pt component */
+    bool enable;
+
+    /** Allow osc_pt2pt to run when user passes in MPI_THREAD_MULTIPLE */
+    bool allow_thread_multiple;
 };
 typedef struct ompi_osc_pt2pt_component_t ompi_osc_pt2pt_component_t;
 
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_component.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_component.c
@@ -16,6 +16,7 @@
  * Copyright (c) 2012-2013 Sandia National Laboratories.  All rights reserved.
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -33,6 +34,7 @@
 #include "osc_pt2pt_data_move.h"
 
 #include "ompi/mca/osc/base/osc_base_obj_convert.h"
+
 
 static int component_register(void);
 static int component_init(bool enable_progress_threads, bool enable_mpi_threads);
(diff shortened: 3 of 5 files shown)


bosilca:
If we add these 4 MCA parameters we are bound to provide them in future releases (at least as synonyms), even after we fix the threading issues in the OSC components. Instead, why not simply asking for the osc MCA parameter to be explicitly defined in he multi-threaded case ?

I can't test right now to see when the OSC module component_init function called ? If it is called directly in MPI_Init, then we will have a lot of spurious failures when threading is enabled, even when one-sided is not used.

gpaulsen:
Sorry, I can't quite parse: "why not simply asking for the osc MCA parameter to be explicitly defined in he multi-threaded case ?"  Can you elaborate?

Yes, component_init is called directly or indirectly at MPI_Init and MPI_Init_thread time.  That's when this PR checks to see if the user requested MPI_Init_thread, but also has osc_pt2pt_allow_thread_multiple = 0, and then disallows that component.   What sort of errors are you envisioning in this model?

hppritcha:
We will only retain these MCA params for 2x series.
:+1:

gpaulsen:
So should I not push to master?  Only PR them against 2.0.x and 2.x ?

bosilca:
@gpaulsen in the same way as we manually specify the BTL or PML we want to use (--mca btl ...), we can specify the OSC components to be used. Thus, if the user does not manually add "--mca osc rdma" to his MCA params, we can disqualify the RDMA module in the multi-threaded case.

If component_init is called during the initialization phase, what will happen if all OSC components disqualify themselves when threading support is requested ? Will the initialization succeed or it will fail because no OSC components are available ? If the latter, then we have a problem.

@hppritcha this is not how we did things in the past. We kept the MCA parameters, either as synonyms or as deprecated parameters for a while.

hppritcha:
@gpaulsen I'm not thinking it would be better to abandon this approach and simply patch 2.0.x to exclude P2P OSC in the select operation until we have a fix.  I don't see a need for any MCA parameter or even environment variable as developers would fix the problem on master then push the fix back to 2.0.x branch.  I'll open an issue and link it with 2614 to remind the developer(s) to revert the commit that will be associated with this patch.

gpaulsen:
So, the 'enable' parameter is similar to some other components that have an enable parameter.  Obveously there are a few ways to disable a component:
1) specify --mca btl ^<component>
2) remove the component shared library from the install.
3) some components have an 'enable' MCA parameter.

The reason that we prefer #3 is that the distro or the system admin can disable a component via the OPAL_PREFIX/etc/openmpi-mca-params.conf file easily.

As far as the 'enable_thread_multiple' MCA parameter, that is a bit weirder as it's conditional upon the value of MPI_THREAD_INIT, and we hope that it gets fixed in a future release.

I tested with no osc components, and it passes through MPI_Init (or MPI_Init_thread), but then fails at MPI_Win_create with the following message:
[host:pid] *** An error occurred in MPI_Win_create
[host:pid] *** reported by process [pid,3]
[host:pid] *** on communicator MPI_COMM_WORLD
[host:pid] *** MPI_ERR_WIN: invalid window
[host:pid] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[host:pid] ***    and potentially your MPI job)

It would be nice to have a better error message in this case, but we should be able to detect this and produce a nicer error message (I just don't know how yet).

rhc54:
I'm a little confused here - can someone please clarify something for me? We have never included values in the template default MCA param file before, and we shouldn't be doing so. It is impossible to predict what different installations need.

If something needs to be disabled, let the system admin do so. I would urge that we not set a new precedent in this regard.

hppritcha:
I opened a PR #2630 which ought to address @bosilca 's concerns about mca parameters for workarounds like this one.

hppritcha:
If we don't reach a consensus by COB today on how to do a hack workaround for #2614, I'm going to declare 2.0.2rc2 dead wrt to a release in 2016.  We can do an rc2 in first week of January, and hopefully get a release out either end of first or second week of January.

rhc54:
FWIW: I think #2630 looks fine to me.

gpaulsen:
I think #2630 would be sufficient for osc_pt2pt for v2.0.2, but we've also seen hangs in osc_rdma component as well.  We haven't tested osc_rdma enough to know that there aren't wrong answers, but we're concerned enough that we recommend also disabling osc_rdma for the MPI_THREAD_MULTIPLE case.

We should discuss this situation at the face to face and see if we can come up with some general procedures (disable things via conf files, or hard coded disables, or documentation, etc) that works for everyone.  MCA parameter seems very useful for this situation, so I'd like to hear folk's opinions.

rhc54:
It depends on the type of problem. If something truly doesn't work for a particular case (e.g., ```MPI_THREAD_MULTIPLE```), then we disable it internally by checking for that case. We wouldn't require someone to set an MCA parameter for that purpose as we _know_ it doesn't work in all environments.

If we have a scenario where something does work, but not in a particular environment, then we try to detect that environment and again disable it internally.

If something works, but there is an application-dependent tradeoff (e.g., a performance hit) associated with using it, then we use MCA params to help dictate selection. This is what @bosilca was alluding to.

This seems to be to be a situation where we honestly aren't sure these components work in threaded situations. This is detectable, and we shouldn't be using MCA params to deal with it - we should simply detect that this is a multi-threaded scenario and disable the component.

We should also continue our policy of never putting a param in the default MCA param file we distribute. It can take a lot of effort for someone to track down that unexpected behavior was caused by some unknown param being set.

hppritcha:
I'm reluctant to disable OSC rdma, but if the consensus is that we should do that, I'll modify #2630 to handle disable rdma as well. @hjelmn

hjelmn:
osc/rdma is thread safe. do you have a reproducer that shows otherwise?

hjelmn:
btw, we already have a way to enable/disable a component. --mca osc.

gpaulsen:
That all sounds reasonable to me.  I'm still learning, so I appreciate everyone's patience.

I only question the "never put a param in the default MCA param file".  If the value is NOT set in that file, there is STILL a (more hidden in my way of thinking) default that's passed in via the mca_base_component_var_register() function.  In other projects I've seen a policy of always putting the same default that's in the registration call in the conf file also, so that all options are more visible to customers.  Of course those values can be sometimes be queried using ompi_info and mpirun if run in the same environment as the job (not always easy to do).

Regarding the hard coding the disabling of components, we have seen cases in the past where customers are upset because we've had to disable a component for a particular release due to wrong answers, but they are certain that their app is not tickling the wrong answers and want to compare apples to apples (sometimes even if they DO get a wrong answer for performance comparisons).  This is why, in the past, we've provided some runtime tunable to enable even the wrong-answers case, if possible.  I agree, generally not as useful, but this is our rational.

gpaulsen:
@hjelmn - osc_rdma hang with MPI_THREAD_MULTIPLE can be seen with 1sided_mt.c here:
https://github.com/open-mpi/ompi-tests/pull/25

rhc54:
What happens when you remove the parameter from the default param file? If the default value is still applied, then doesn't that confuse people? After all, they saw the param in the file, and then removed it - so why isn't it "gone"?

I hear what you are saying about the hard coding issue. Alternative is to shut it off _unless_ someone specifically requests it via MCA param. We've used that before too.

hjelmn:
Ok, will take a look later today.

gpaulsen:
Yeah.  users...  Can't make em all happy.  

I put a topic in the face to face agenda to discuss... I'll try to frame the conversation.  It's just an extension of the general task of making the UI better, which of course is subjective and never ending, but good to talk about, and make progress on.

gpaulsen:
I'm closing this PR, as the community chose to go a different direction in:
PR #2630.

PR #2636, opened 2016-12-25
configury: fix OMPI_UNIQUE -> OMPI_FLAGS_UNIQUE

### LINKED ISSUE #2635: Configure script mangles custom CFLAGS
Attempting to configure ompi-2.0.1 fails when attempting to use certain sets of custom cflags due to the fact that the configure script incorrectly strips additional --param statements after the first.

MWE for this is (parameters for haswell processors):
`CFLAGS="--param l1-cache-size=32 --param l1-cache-line-size=64 --param l2-cache-size=6144 -march=haswell"`

This is a major issue when attempting to optimize binaries for systems where "-march=native" cannot be used for whatever reason.

jsquyres:
Looks like we missed one place where we needed to swap OMPI_UNIQUE for OMPI_FLAGS_UNIQUE.  Thanks to Phil Tooley (@Telemin) for reporting the issue.

Fixes #2635.


jsquyres:
@telmin can you please check to see if this fixes your issue?

Telemin:
Hi unfortunately not fully.  It fixes the ./configure issue, but there are still problems later in the build.  The bad CFLAGS are still making their way into for example `opal/mca/pmix/pmix2x/pmix/src/util/keyval/Makefile` and the build fails as soon as one of these makefiles is encountered.

jsquyres:
@rhc54 Can you check PMIx?  I suspect that there's a case similar to this in PMIx.

rhc54:
@Telemin If you can, please pull the referenced change and add it to this one - should fix the remaining issue

jsquyres:
@Telemin @rhc54 I just rebased this PR to include Ralph's commit.  So if you pull this PR, you should be good to go.

jsquyres:
@Telemin I'm quite sure that this PR fixes at least part of the problem, so I'm going to merge it.  But let us know if there's still more that needs to be done.

PR #2671, opened 2017-01-05
Per recommendation from Gilles, use writev instead of sending the hea…

### LINKED ISSUE #2657: comm_spawn seems very slow
This mostly becomes an issue on the loop_spawn test, which is taking several 10's of minutes. I haven't had a chance to check this out - I believe it is the communicator formation codes that are taking the time, but haven't verified that.

@ggouaillardet Would you have a chance to dig into this? Comm_spawn should be a pretty fast operation, really

rhc54:
…der and message body in two separate write cmds. Operates much faster!

Fixes #2657

### DIFF
--- a/opal/mca/pmix/pmix2x/pmix/src/event/pmix_event_notification.c
+++ b/opal/mca/pmix/pmix2x/pmix/src/event/pmix_event_notification.c
@@ -525,7 +525,7 @@ static void _notify_client_event(int sd, short args, void *cbdata)
                                     "pmix_server: notifying client %s:%d of code %d",
                                     pr->peer->info->nptr->nspace, pr->peer->info->rank, cd->status);
                 PMIX_RETAIN(cd->buf);
-                PMIX_SERVER_QUEUE_REPLY(pr->peer, 0, cd->buf);
+                PMIX_PTL_SEND(pr->peer, 0, cd->buf);
             }
         }
     }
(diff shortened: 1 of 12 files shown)


bosilca:
@rhc54 the 2 codes are not equivalent. The new code only works when the entire message is sent into a single syscall (basically the message is less than the kernel buffer for your socket). I would suggest yo take a look at the TCP BTL is has support for all these operations in a thread-safe and efficient way.

bosilca:
Sorry I should ahve mentionned the 2 functions of interest: mca_btl_tcp_frag_send and mca_btl_tcp_frag_recv (both in btl_tcp_frag.c).

rhc54:
@bosilca quick question: I take it that someone fragments the message into fixed-sized buffers prior to calling the frag_send function? So that way you know the size of the iovec data buffers, and how many iovec's are in each complete message?

rhc54:
@bosilca never mind - i'm beginning to parse thru what you guys did. i think this code is actually pretty close, so it shouldn't take too much to add the missing loop

rhc54:
@bosilca @ggouaillardet I've made a first pass at resolving the concerns raised by @bosilca, but I've run into an issue on the recv end - I'm now getting EAGAIN when trying to read, and then we hang. The send seems to complete just fine. However, all of our messages are tiny, and so the resend code path is never actually being exercised.

I'm afraid I have no time to continue this right now. If someone else can pick it up, I'd appreciate it. Meantime, I'm going to just set the usock component priority higher than TCP so we recover performance.

ggouaillardet:
@rhc54 i will take care of that from Tuesday

rhc54:
Closed in favor of #2686

PR #2673, opened 2017-01-06
Raise the priority of the usock component so it gets preferentially picked

rhc54:


rhc54:
Fixes #2657

PR #2682, opened 2017-01-07
v1.10: amd64: save/restore all 64 bits of rbx around cpuid

### LINKED ISSUE #2678: PR #2596 introduced regression when configuring with -O3 optimization
#2596, which has been merged back in to v2.x and v.2.0.x introduced a major regression.  Open MPI can no longer be compiled with gcc (at least) at -O2 or -O3 for x86_64 without immediately segfaulting in MPI_Init.   The problem has been observed both for gcc 4.9 and 6.3.

orterun also fails with a similar segfault pattern.  If one uses the cpuid instruction directly (avoiding all the exb exchange stuff - which isn't necessary for x86_64), the problem vanishes.  

See comments to commit a718743a5c7017c13fb56950d16edd0b59f07108.
Although the problem was originally reported on a system with Opteron processors, 

@marksantcroos thanks for catching this.

hjelmn:
This commit fixes a bug in the timer check. When -fPIC is used we need
to save/restore ebx. The code copied from patcher was meant for 32-bit
systems and did not work correctly on 64-bit systems. This commit
updates the save/restore to use rbx instead of ebx.

Fixes #2678

(cherry picked from commit 5b70ae3ec0f36e422c6401c5f7d4ec363a1059e1)

### DIFF
--- a/opal/include/opal/sys/amd64/timer.h
+++ b/opal/include/opal/sys/amd64/timer.h
@@ -56,16 +56,18 @@ opal_sys_timer_get_cycles(void)
 
 static inline bool opal_sys_timer_is_monotonic (void)
 {
-    int32_t cpuid1, cpuid2, tmp;
+    int64_t tmp;
+    int32_t cpuid1, cpuid2;
     const int32_t level = 0x80000007;
+
     /* cpuid clobbers ebx but it must be restored for -fPIC so save
      * then restore ebx */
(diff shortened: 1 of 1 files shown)


hjelmn:
@rhc54 Not sure how I could have missed this. Causes a SEGV when -fPIC is used.

hppritcha:
bot:lanl:retest

PR #2686, opened 2017-01-09
pmix2x: ptl/base: send header and message data together via writev()

ggouaillardet:
on Linux, sending the header and then the message data does severely
impact performances of ptl/tcp :
on the receiver, reading the data can often result in an PMIX_ERR_RESOURCE_BUSY
or PMIX_ERR_WOULD_BLOCK, which ends up degrading performances)
this commit send both header and message data at the same time via writev()
and makes ptl/tcp virtually as efficient as ptl/usock.

fwiw, no such degradation was observed on OSX.

Refs open-mpi/ompi#2657


ggouaillardet:
@rhc54 @bosilca i kept it simple and humbly suggest this is way less invasive PR than #2671

bosilca:
Looks good. I do have one comment about the performance of the patch. In general the writev would send less data than expected when the kernel buffer is full. In this case there is no reason to do one additional syscall before returning from the send_msg function. Thus, in the case when writev returns less data then required I would update the pointers and then force a return with the code PMIX_ERR_RESOURCE_BUSY instead of going back into another writev.

ggouaillardet:
@bosilca thanks for the insight ! i made the suggested changes

PR #2687, opened 2017-01-09
README: add a note about OpenBSD issues

hppritcha:
PPhargrove's testing shows one cannot build the 2.0.x
release currently on OpenBSD unless the --enable-mca-no-build=patcher
option.

Add a note about this in the README.


jsquyres:
This is a workaround for #1654 until we actually fix the issue.

PR #2689, opened 2017-01-09
mca/base: account for NULL string_value in verbose set

hjelmn:
The MCA variable code calls the string from value function with a NULL
string to verify values. The verbosity enumerator was not correctly
checking for a non-NULL value before trying to set the string.

(cherry picked from commit d6bd69dc93b99137f759c3908dcd3cf934799d24)


hjelmn:
If too late for 2.0.2 please just mark as 2.0.3. Its a very simple fix for a SEGV in MPI_T_cvar_write.

jsquyres:
bot:mellanox:retest

PR #2692, opened 2017-01-09
master: usnic queue fixes

jsquyres:
A series of minor improvements to the usnic BTL, with three important fixes that happen at large scale:

1. Fix a CQ overrun (d70aa3e)
1. Fix a TX queue overrun (6df502c)
1. Fix a memory leak for ACKs (7ccbac9)

PR #2695, opened 2017-01-10
v2.0.2: usnic queue fixes

jsquyres:
This is the v2.0.x version of master PR #2692.

It does not include all off the commits from #2692 -- some of those are just extra error checking, and one is a change in the output format of statistics.

This is a blocker because it fixes some usnic bugs when running at scale.


jsquyres:
bot:mellanox:retest

gpaulsen:
We'll wait a few hours to see if this gets reviewed and pulled.  If not we'll pull from PR.

jsquyres:
@bturrubiates Updated to include fixes from review of #2692.

hppritcha:
lanl-disable-dlopen failure is spurious.  Its catching the cpuid error.

jsquyres:
@hppritcha This is good to go.  I'm going to merge so that it's in the nightly tarball.

PR #2699, opened 2017-01-10
v1.10: usnic queue fixes

jsquyres:
This is the v1.10 version of master PR #2692.

It includes all but the last commit from #2692 (i.e., the one that change the output format of statistics).

This is a blocker because it fixes some usnic bugs when running at scale.


jsquyres:
@bturrubiates Updated to include fixes from review of #2692.

PR #2703, opened 2017-01-11
rcache/base: do not release vma stuctures in vma_tree_delete

hjelmn:
This commit fixes a deadlock that can occur when the libc version
holds a lock when calling munmap. In this case we could end up calling
free() from vma_tree_delete which would in turn try to obtain the lock
in libc. To avoid the issue put any deleted vma's in a new list on the
vma module and release them on the next call to vma_tree_insert. This
should be safe as this function is not called from the memory hooks.

### DIFF
--- a/opal/mca/rcache/base/rcache_base_vma.h
+++ b/opal/mca/rcache/base/rcache_base_vma.h
@@ -13,7 +13,7 @@
  *
  * Copyright (c) 2006      Voltaire. All rights reserved.
  * Copyright (c) 2009      IBM Corporation.  All rights reserved.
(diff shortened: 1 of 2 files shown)


hjelmn:
@bosilca Please test and let me know if this fixes #1654.

hppritcha:
tentatively giving a 2.0.2 milestone.

hjelmn:
@bosilca I am going to go ahead and merge this. Please test the v2.0.x PR.

bosilca:
I got a short allocation on the machine (a dozen runs) and I was not able to replicate. The original issue was only seldom triggered, so there is a chance I just got lucky this time.

PR #2711, opened 2017-01-11
Protect default mapping/binding options for cases where no NUMA or SOCKET objects exist - like VMs

rhc54:


rhc54:
bot:mellanox:retest

PR #2713, opened 2017-01-11
amd64/timer: bring timer code up to date with master

hjelmn:
There is a bug with the timer code used in v1.10 that causes
MPI_Wtime() to reset more frequently than expected (~ every 8000
seconds instead of 500 years). The timer code was fixed on master but
was never fixed on v1.10.

Backport of 6772b077925f5143988457b66085685efdab73de


hjelmn:
@bosilca The rdtscp configure check on master is broken. The one included here is tested and is working. I will update the master, v2.0.x, and v2.x checks as well.

hjelmn:
whoops. this is not quite right. forgot to make some changes based on 1.10 differences.

PR #2714, opened 2017-01-12
timer/linux: prevent 64-bit overflow

hjelmn:
The linux timer code was multiplying the result of the x86 time stamp
counter by 1000000 before dividing by the cpu frequency. This can
cause us to overflow 64 bits if the time stamp counter grows larger
than ~ 1.8e13 (about 8400 seconds after boot). To fix the issue the
1000000 factor has been moved to avoid overflow.

### DIFF
--- a/opal/mca/timer/linux/timer_linux_component.c
+++ b/opal/mca/timer/linux/timer_linux_component.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2015-2016 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2015-2017 Los Alamos National Security, LLC. All rights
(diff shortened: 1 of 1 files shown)


hjelmn:
We may potentially loose some accuracy with the timer for cpus with non-integer clock speeds in MHz. Since this is unlikely we probably won't see any issues.

hjelmn:
@hppritcha This is for 2.0.2 as well. Will open a PR tonight. If @bosilca can't get to it either you or @jsquyres can probably take a look. Pretty straightforward.

bosilca:
Good catch. Btw, why not changing the scale of opal_timer_linux_freq to completely avoid the overhead of the second integer division in the opal_timer_base_get_usec_sys_timer function ?

hjelmn:
@bosilca Sure. Can do it that way. Was keeping the change simple but getting rid of the division is probably the better way to go. Will update the PR.

hjelmn:
@bosilca Updated.

PR #2717, opened 2017-01-12
oshmem: sshmem: adds UCX allocator

alex-mikheev:
Also some code cleanup
@yosefe @jladd-mlnx @igor-ivanov

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx.c
+++ b/oshmem/mca/spml/ucx/spml_ucx.c
@@ -425,7 +425,8 @@ sshmem_mkey_t *mca_spml_ucx_register(void* addr,
     size_t len;
     int my_pe = oshmem_my_proc_id();
     ucp_mem_map_params_t mem_map_params;
-    int seg;
+    int segno;
+    map_segment_t *mem_seg;
     unsigned flags;
 
     *count = 0;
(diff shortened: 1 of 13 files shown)


oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -405,26 +406,32 @@ sshmem_mkey_t *mca_spml_ucx_register(void* addr,
         return NULL;
     }
 
-    seg = memheap_find_segnum(addr);
+    segno   = memheap_find_segnum(addr);
```
miked-mellanox: what if segno is invalid?

jladd-mlnx:
@alex-mikheev please add signed-off-by

jladd-mlnx:
@alex-mikheev should we deprecate the verbs allocator? Please PR into 1.10, 2.0.2, and 2.1.

yosefe:
@alex-mikheev i don't see where memory is actually allocated with UCX?
also - better put the cleanups in existing code in separate commit

alex-mikheev:
@jladd-mlnx not yet. ucx allocator can only be used by the ucx spml
@yosefe I forgot to 'add' new files

oshmem/mca/spml/ucx/spml_ucx.c
alex-mikheev: It is always valid because the register function is only called for memory segments that were already allocated and belong to the memheap.

oshmem/mca/sshmem/sshmem_types.h
```diff
@@ -112,9 +113,10 @@ typedef struct map_segment {
     sshmem_mkey_t       *mkeys;          /* includes local segment bases in va_base */
     segment_flag_t       flags;          /* enable/disable flag */
     int                  seg_id;
-    char                 seg_name[OPAL_PATH_MAX];
```
igor-ivanov: @alex-mikheev do you believe that `seg_name` is needless? I see that original code has `unlink(ds_buf->seg_name)` in `shmem_ds_reset()`. I do not remember if it is something useful but pay attention to avoid related issues.

alex-mikheev: I see unlink() and memset two zero. However I do not see where seg_name is set to something and I do not see where it is used. (except in debug prints)

yosefe:
bot:retest

hppritcha:
@alex-mikheev  Please update v3.0.0 whitelist  #3107 when you merge this.  Check off ucx allocator button.  Thanks.

jladd-mlnx:
@yosefe Can we remove the DNM label?

jsquyres:
I think it needs to pass Mellanox Jenkins, too.  Otherwise, all future PRs will fail Mellanox Jenkins.

jladd-mlnx:
Looks like it's pulling in an older version of UCX that doesn't have the new flags defined.

yosefe:
right, we need to make sure it passes with Mellanox jenkins.

yosefe:
@alex-mikheev pls pick to v2.1

jladd-mlnx:
This is too late for 2.1

PR #2723, opened 2017-01-12
rcache/base: do not release vma stuctures in vma_tree_delete

hjelmn:
This commit fixes a deadlock that can occur when the libc version
holds a lock when calling munmap. In this case we could end up calling
free() from vma_tree_delete which would in turn try to obtain the lock
in libc. To avoid the issue put any deleted vma's in a new list on the
vma module and release them on the next call to vma_tree_insert. This
should be safe as this function is not called from the memory hooks.

(cherry picked from commit 79cabc92fd0bc5307649597f5ef9c6006d193f5d)


hjelmn:
:bot:retest:mellanox

jsquyres:
@hppritcha Good to go.

PR #2712, opened 2017-01-11
ess/hnp: add support for forwarding additional signals

hjelmn:
This commit adds support to the hnp ess module to forward additional
signals beyond the default SIGUSR1, SIGUSR2, SIGSTP, and SIGCONT.

Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

PR #2724, opened 2017-01-12
ess/hnp: add support for forwarding additional signals (#2712)

hjelmn:
* ess/hnp: add support for forwarding additional signals

This commit adds support to the hnp ess module to forward additional
signals beyond the default SIGUSR1, SIGUSR2, SIGSTP, and SIGCONT.

* Generalize this a bit to allow a broader range of signals to be forwarded. Turns out that SIGURG is now a "standard" signal, though the value differs across systems. So setup to forward it (and some friends) if they are defined. Allow users to provide the signal name (instead of the integer value) as the value of even the more common signals does vary across systems. Don't limit the number that can be supported.

* ess/hnp: fix some bugs in the signal forwarding code

This commit fixes two bugs:

 - signals_set needs to be set even if no signals are being
   forwarded. If it is not set we will SEGV in libevent if
   ess_hnp_forward_signals == none.

 - SIGTERM and SIGHUP are handled with a different type of handler. Do
   not allow the user to specify these to be forwarded.

* We are sure to get "dinged" if error messages aren't nicely output via show_help, so do so here

(cherry picked from commit 110840fc877a0b0d759d3a787341fc22ebb29eac)

### DIFF
--- a/orte/mca/ess/hnp/Makefile.am
+++ b/orte/mca/ess/hnp/Makefile.am
@@ -10,12 +10,17 @@
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
 # Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2017      Los Alamos National Security, LLC. All rights
+#                         reseved.
+# Copyright (c) 2017      Intel, Inc.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
 #
 # $HEADER$
 #
+
+dist_ortedata_DATA = help-ess-hnp.txt
 
 sources = \
         ess_hnp.h \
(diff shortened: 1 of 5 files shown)


hjelmn:
This is an actual failure. Looks like I missed something on the merge. Fixing.

rhc54:
@hjelmn I have seen that error before, and it is a Jenkins failure - not a real issue. All I had to do was tell it to retest.

jsquyres:
@rhc54 @hjelmn Is this needed for v2.0.2?

hppritcha:
bot:mellanox:retest

hjelmn:
@jsquyres This restores functionality we lost somewhere in the 1.8.x/1.10.x series. We were not aware of it until earlier this month due to the fact that our codes are only now on 1.10.x.

hppritcha:
bot:mellanox:retest

PR #2730, opened 2017-01-12
Do not mark a -host specification as "slots given" unless it actually specifies the #slots

rhc54:
This allows the system to dynamically detect the number of available processors and set the #slots accordingly.

### DIFF
--- a/orte/tools/orterun/orterun.1in
+++ b/orte/tools/orterun/orterun.1in
@@ -1,6 +1,7 @@
 .\" -*- nroff -*-
 .\" Copyright (c) 2009-2014 Cisco Systems, Inc.  All rights reserved.
 .\" Copyright (c) 2008-2009 Sun Microsystems, Inc.  All rights reserved.
+.\" Copyright (c) 2017      Intel, Inc.  All rights reserved.
 .\" $COPYRIGHT$
 .\"
 .\" Man page for ORTE's orterun command
@@ -732,10 +733,14 @@ Here, we list both the host names (aa, bb, and cc) but also how many "slots"
 there are for each.  Slots indicate how many processes can potentially execute
 on a node.  For best performance, the number of slots may be chosen to be the
 number of cores on the node or the number of processor sockets.  If the hostfile
-does not provide slots information, a default of 1 is assumed.
-When running under resource managers (e.g., SLURM, Torque, etc.),
-Open MPI will obtain both the hostnames and the number of slots directly
-from the resource manger.
+does not provide slots information, Open MPI will attempt to discover the number
+of cores (or hwthreads, if the use-hwthreads-as-cpus option is set) and set the
+number of slots to that value. This default behavior also occurs when specifying
+the \fI-host\fP option with a single hostname. Thus, the command
+.
+.TP 4
+mpirun -H aa ./a.out
+launches a number of processes equal to the number of cores on node aa.
 .
 .PP
 .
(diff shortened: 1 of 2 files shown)


hppritcha:
Do we need to change any documentation for this change in behavior?

rhc54:
Good point - I have updated the man page

hppritcha:
@jsquyres what do you think about this kind of change in behavior within a release stream? 
I recall a lot of discussion about behavior of ```--host``` vs ```--hostfile``` etc.
at one of our recent F2F's:
https://github.com/open-mpi/ompi/wiki/Meeting-2016-02
just wanting to make sure we are keeping with the original decisions.

hppritcha:
Here's the spread sheet that we worked on at the Dallas F2F 2/16:

https://docs.google.com/spreadsheets/d/1poOwNKtYxnDnpF7-D15lmcFVtLRNrmf4_hu3Obtu95M/edit

rhc54:
@hppritcha Hate to tell you, but that spreadsheet isn't accurate - at least, that isn't the current behavior.

ggouaillardet:
@hppritcha @rhc54 imho, this specific case is kind of undocumented/unspecified.

one one hand, we specify a slot-list, and on the other hand, we specify a host.

for example, what if the slot-list contains 12 slots but we specify a different number of slots with the ```--host``` option ? should we try to handle this in a "nice" way ? or should we simply abort because of an inconsistency in the options requested by the user ?

in this very specific case, e.g. ```mpirun --slot-list 0:0-5,1:0-5 --host xxx```, yet an other option could be to set the ```ORTE_NODE_FLAG_SLOTS_GIVEN``` flag, and set the number of slots to ```12``` (e.g. the number of slots in the list)

rhc54:
I've said this before, but I'll reiterate here - I am no longer supporting the -host/-hostfile code. We keep coming up with every imaginable corner-case, and the complexity of trying to handle all of them - while preserving existing behavior - is insane. Clear evidence: the current master behavior no longer mirrors @jsquyres spreadsheet.

This patch makes the branch follow the master and fixes the user's problem, but the result isn't the behavior in the spreadsheet. Someone else can figure out what they want to do.

jladd-mlnx:
bot:mellanox:retest

jladd-mlnx:
bot:mellanox:retest

jladd-mlnx:
bot:mellanox:retest

hppritcha:
Here's what we have at the moment across several releases for the ```--host/---hostfile``` behavior where host_foo has a single entry foo.  The table below shows how many processes are launched as a function of mpirun command line options and Open MPI release:

Release          | mpirun --host foo | mpirun --host foo:2 | mpirun --hostfile host_foo | mpirun --hostfile host_foo -np 1»
------------- | ------------------ | -------------------- | -------------------------- | -------------------------------------
1.8.8 | one process | doesn't work | nslot processes | one process
1.10.0  | one process  | doesn't work | nslot processes | one process
1.10.5  | one process  | doesn't work | nslot processes | one process
1.10.x |  one process  | doesn't work  | nslot processes | one process
2.0.0 | one process | two processes | nslot processes | one process
2.0.1 | one process | two processes | nslot processes | one process
master | nslot processes | two processes | nslot processes | one process

where nslots is the number of cores on the host foo

hppritcha:
Discussed at the devel F2F and decided to keep the behavior of 1.10.x series (modulo the new --host foo:X).  So closing this PR.

jsquyres:
For reference: https://github.com/open-mpi/ompi/wiki/Meeting-2017-01

jjhursey:
Food for thought: It would be nice if we had a system regularly running a handful of `-host` and `-hostfile` runs and compare the output to confirm that we stay in compliance with what we intend to do. Maybe a CI test on a specific system (keeping to a specific system makes it easy to setup without handling the generic system problem).

PR #2731, opened 2017-01-12
Do not mark a -host specification as "slots given" unless it actually specifies the #slots

rhc54:
This allows the system to dynamically detect the number of available processors and set the #slots accordingly.


hppritcha:
See #2730 for more on this.  This represents a significant break in behavior from the earlier 2.0.0 and 2.0.1 releases and probably should not be accepted for 2.x release streams.

jladd-mlnx:
bot:mellanox:retest

hppritcha:
Per discussion at the Open MPI developer F2F today we're not going to merge this in to 2.x release stream.  See #2730 for more info.

PR #2732, opened 2017-01-12
Missed one spot - plug fd leaks in orteds

### LINKED ISSUE #2691: regression: orte_submit_job() high error rates during startup
With the latest master I get frequent errors during the launch of jobs where the launch callback returns `-117` for the first failing jobs and then `69` for the remaining failing jobs.
Still trying characterise it fully. Seems to work fine for ~190 consecutive jobs and then starts to fail.

Also tried to find out what these errors maps to, but couldn't find that either...

The workload consists of very short jobs.

This is a regression in comparison to early december, didn't track it down to a specific commit yet.

rhc54:
Fixes #2691


rhc54:
looks like everything is wedged...again

PR #2738, opened 2017-01-16
Cancel the waitpid callback once the waitpid on a process has fired to avoid multiple notifications

rhc54:


ggouaillardet:
Thanks Ralph !

PR #2740, opened 2017-01-17
Add an MCA param "hnp_on_smgmt_node"

rhc54:
Add an MCA param "hnp_on_smgmt_node" that mpirun can use to tell the orteds to ignore its topology signature as mpirun is executing on a system mgmt node, and hence a different topology than the compute nodes


hjelmn:
@hppritcha Turns out mpirun was unusable slow because every orted was reporting the topology xml. This change is targeted for 2.1 and will have to go in the lanl XC platform file. I can confirm from tests on tr2 that the launch times are greatly improved with this option.

PR #2747, opened 2017-01-18
Try a different approach for scalably dealing with hetero clusters

rhc54:
Revert e9bc293 and try a different approach for scalably dealing with hetero clusters. Have each orted send back its topo "signature". If mpirun detects that this signature has not been seen before, then ask for that daemon to send back its full topology description. This allows the system to only get the topology once for each unique topo in the cluster.


rhc54:
@hjelmn This isn't quite right yet - will fix shortly

rhc54:
@hjelmn should be good to go - let me know if it works for you as I'd like verification before committing

rhc54:
@jsquyres you will need to adjust your MTT cmd line if/when this is committed as the hetero-nodes option is being removed

hjelmn:
Will be testing this overnight.

PR #2756, opened 2017-01-18
Cleanup the os_dirpath logic

### LINKED ISSUE #2754: Issues with opal session dir caused by change to opal/util/os_dirpath.c
Hi!

Sorry to hassle you guys again.  We are just in the process of trying to migrate to v2.x, and are hitting a condition that appears to have been introduced in 58d1b3f4d0e295184c25bca862eb0aadd0799c87.

Some of our nodes have a per-job TMPDIR on Lustre, and we are getting tripped up because when calling mkdir on an existing path, where the user would not have permissions to create the directory, Lustre returns EACCES rather than EEXIST.  This breaks the new form of the check in opal/util/os_dirpath.c

```
[user@host ompi ((58d1b3f4d0...))]$ mkdir /p5
mkdir: cannot create directory `/p5': File exists
(i.e. root filesystem behaves as you expected)

[user@host ompi ((58d1b3f4d0...))]$ mkdir /p5/000scratch
mkdir: cannot create directory `/p5/000scratch': Permission denied
(this causes a problem)

[user@host ompi ((58d1b3f4d0...))]$ mkdir /p5/000scratch/my_temp_dir
(but if we ignore it and just carry on trying to make the deeper components, it will work fine)
```

Without wanting to start an argument over whether Lustre is doing the right thing here, this seems like it might be a not-terribly-uncommon configuration?

Cheers,
Tim

rhc54:
Cleanup the os_dirpath logic so it doesn't error out if the directory actually gets created (regardless of what mkdir returns), and pretty-prints the error if it does error out.

Fixes #2754 

(cherry picked from commit b257c32d2c264f78897e24c6952ea5297709e44c)

### DIFF
--- a/opal/util/os_dirpath.c
+++ b/opal/util/os_dirpath.c
@@ -113,10 +113,11 @@ int opal_os_dirpath_create(const char *path, const mode_t mode)
         }
 
         /* Now that we have the name, try to create it */
-        ret = mkdir(tmp, mode);
+        mkdir(tmp, mode);
+        ret = errno;  // save the errno for an error msg, if needed
         if (0 != stat(tmp, &buf)) {
             opal_show_help("help-opal-util.txt", "mkdir-failed", true,
-                        tmp, strerror(errno));
+                        tmp, strerror(ret));
             opal_argv_free(parts);
             free(tmp);
             return OPAL_ERROR;
--- a/opal/util/os_dirpath.c
+++ b/opal/util/os_dirpath.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016-2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
@@ -121,6 +121,12 @@ int opal_os_dirpath_create(const char *path, const mode_t mode)
             opal_argv_free(parts);
             free(tmp);
             return OPAL_ERROR;
+        } else if (i == (len-1) && (mode != (mode & buf.st_mode)) && (0 > chmod(tmp, (buf.st_mode | mode)))) {
+            opal_show_help("help-opal-util.txt", "dir-mode", true,
+                           tmp, mode, strerror(errno));
+            opal_argv_free(parts);
+            free(tmp);
+            return(OPAL_ERR_PERM); /* can't set correct mode */
         }
     }
 
--- a/opal/util/os_dirpath.c
+++ b/opal/util/os_dirpath.c
@@ -11,6 +11,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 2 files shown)


tjb900:
Awesome, thanks!  You guys sure do respond quickly.  One thing though - with the current patch, I think the stat() might destroy the errno from mkdir(), in which case the message from strerror will end up being wrong.

rhc54:
crumby - you are quite correct. will fix. thanks!

opal/util/os_dirpath.c
```diff
@@ -112,14 +112,12 @@ int opal_os_dirpath_create(const char *path, const mode_t mode)
...
-        ret = mkdir(tmp, mode);
-        if ((0 > ret && EEXIST != errno) || 0 != stat(tmp, &buf)) {
-            opal_output(0,
-                        "opal_os_dirpath_create: "
-                        "Error: Unable to create the sub-directory (%s) of (%s), mkdir failed [%d]\n",
-                        tmp, path, ret);
+        /* Now that we have the name, try to create it */
+        mkdir(tmp, mode);
+        ret = errno;  // save the errno for an error msg, if needed
+        if (0 != stat(tmp, &buf)) {
```
ggouaillardet: the pointed commit was meant to solve a race condition (e.g. ```mkdir()``` after ```stat()```)
 
there still is a tricky race condition here ... someone else might have ```mkdir()``` but with the wrong permissions, so it will likely crash later.
so i guess we have to check the permissions at the end.

this is for master
```patch
diff --git a/opal/util/os_dirpath.c b/opal/util/os_dirpath.c
index 9ef9ebb..2328179 100644
--- a/opal/util/os_dirpath.c
+++ b/opal/util/os_dirpath.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
- * Copyright (c) 2015-2016 Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  * Copyright (c) 2016-2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
@@ -121,6 +121,12 @@ int opal_os_dirpath_create(const char *path, const mode_t mode)
             opal_argv_free(parts);
             free(tmp);
             return OPAL_ERROR;
+        } else if (i == (len-1) && (mode != (mode & buf.st_mode)) && (0 > chmod(tmp, (buf.st_mode | mode)))) {
+            opal_show_help("help-opal-util.txt", "dir-mode", true,
+                           tmp, mode, strerror(errno));
+            opal_argv_free(parts);
+            free(tmp);
+            return(OPAL_ERR_PERM); /* can't set correct mode */
         }
     }

```

rhc54: sounds reasonable - if you want to commit to master, I'll cherry-pick it into this PR

ggouaillardet: ok, i made #2759 
will merge when CI completes

rhc54:
@ggouaillardet Thanks! Cherry-picked your update so it is included here

jsquyres:
@hppritcha This is good to go when CI finishes.  It fixes an issue on Lustre.

tjb900:
Hi all,

Just a heads up, I hit this again on v2.1.0 - it appears to not have been cherrypicked into v2.x.

Cheers,
Tim

ggouaillardet:
@tjb900 that is correct, i just filled #3345 for this

tjb900:
Thanks!

PR #2760, opened 2017-01-19
Expose opal_set_using_threads in python bindings

marksantcroos:
Expose opal_set_using_threads() in python bindings so that the client can request to enable mutexes.
Improve error message on missing ompi_info while we are at it.
Increase version for pypi.

Signed-off-by: Mark Santcroos mark.santcroos@rutgers.edu

### DIFF
--- a/orte/bindings/python/setup.py
+++ b/orte/bindings/python/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup
 
 setup(
     name = "orte-cffi",
-    version = "0.4.0",
+    version = "0.4.2",
     author = "Mark Santcroos",
     author_email = "mark.santcroos@rutgers.edu",
     description = "CFFI-based Python wrapper for Open RTE",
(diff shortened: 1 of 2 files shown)


jsquyres:
This may be ignorance on my part, but why are we exposing an OPAL API through ORTE Python bindings?  Isn't that an abstraction violation?

marksantcroos:
> This may be ignorance on my part, but why are we exposing an OPAL API through ORTE Python bindings? Isn't that an abstraction violation?

Fair point. I guess you are right in principle.

I think we never considered doing the binding on the "root" project level, as mpi4py does that pretty well already.
Practically I considered the bindings on the ORTE level, as everything required to use ORTE.
I wouldn't be against doing something similar on the OPAL layer, although I'm not sure how that would work out in practice (i.e. im currently not linking against libopal, that gets pulled in by orte).

Another alternative would be to hide this call in the binding layer and not expose it.

Last alternative would be to provide an ORTE equivalent/wrapper for this OPAL call.

I'm sure there are more alternatives, what do you folks think?

rhc54:
While it is a fair point, I think it is probably not necessary. This isn't meant to be a pure abstraction overlay - it is just meant to provide access to the ORTE/OPAL collection. So I'd just leave it as-is.

PR #2764, opened 2017-01-19
If a tool sees the HNP it is attached to die (thereby losing connecti…

rhc54:
…on), then stop the event loop instead of going through the abort code path. This will allow the tool to cleanup before exiting


rhc54:
bot:mellanox:retest

PR #2765, opened 2017-01-20
Allow parallel processing of launch msg while relaying

rhc54:
Have mpirun and orteds activate the oob/tcp progress thread by default, leaving a way to turn it off via MCA param. Provide a method by which the add_procs command can be processed in parallel with relaying the cmd message to the next daemons down the tree.


rhc54:
bot:mellanox:retest

PR #2769, opened 2017-01-20
v2.x: snapshot: fix hash comparison

jsquyres:
- Don't use "-i" CLI option to perl; it's unnecessary here and causes
  a warning
- Branch names may not be entirely letters (e.g., "v1.11"), so take
  any character in the regexp to match the branch name

(cherry picked from commit 328b654626223f17da0ab413c1bf2f4290a63ca0)


jsquyres:
Not waiting for CI; CI doesn't test these scripts at all.

PR #2772, opened 2017-01-20
master: opal/stacktrace improvements

jjhursey:
The first commit fixes a bug that I saw with this test program:
 * https://gist.github.com/jjhursey/1af4dac993f11f4d6d3e9b4247b96475

If I run the program with Open MPI, it catches the first `SIGFPE` and prints a stack through the opal/stacktrace mechanism, but let's the program continue executing 😞 . When it hits the second `SIGFPE` then the system's default signal handler is invoked and kills the program.

I would expect Open MPI to print the stack and kill the program on the first instance of this signal. And if I run Open MPI with `-mca opal_signal 6,7,11` (skipping `8 = SIGFPE`) then it does exactly that (exits the program at the first encounter).

So the first commit makes sure that we re-raise the signal after processing the stacktrace.

The second commit adds some more flexibility to the stacktrace mechanism in particular where the stacktrace output is directed. Default is still `stderr`, but now we have options for `none`, `stdout`, and a set of per-process files.

See the commit messages for some more detail.

### DIFF
--- a/opal/mca/backtrace/backtrace.h
+++ b/opal/mca/backtrace/backtrace.h
@@ -12,6 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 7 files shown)


jjhursey:
Adding @hjelmn since they were seeing sometime similar to the problem I attempted a fix for in this PR.

I found the discussion here helpful:
 * https://www.gnu.org/software/libc/manual/html_node/Termination-in-Handler.html#Termination-in-Handler

jsquyres:
@jjhursey @hjelmn We're going to make an RC for 2.0.2 tomorrow.  If you want this in 2.0.2, please merge to master and PR to v2.0.2 ASAP.  Thanks.

jsquyres:
Refs #2801

PR #2773, opened 2017-01-20
Add a 'hook' framework

jjhursey:
Add a hook static framework to the `ompi` layer.

Among other uses, this framework is useful as one mechanism for licensing. For licensing it is important that components have the option to be built statically, and not allow the user to force it to be unloaded. The two commits here do just that.

This includes a demo component that provides an example of how the framework might be used.

The framework allows for static components, dynamic components, and dynamic registration of 'hooks' by components in other frameworks. The latter enhancement has proven useful in components that need to know if the application is between `MPI_Init` and `MPI_Finalize`.

### DIFF
--- a/ompi/mca/hook/base/base.h
+++ b/ompi/mca/hook/base/base.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#ifndef OMPI_HOOK_BASE_H
+#define OMPI_HOOK_BASE_H
+
+#include "ompi_config.h"
+
+#include "ompi/mca/mca.h"
+#include "opal/mca/base/mca_base_framework.h"
+
+#include "ompi/mca/hook/hook.h"
+
+BEGIN_C_DECLS
+
+/**
+ * Framework struct declaration for this framework
+ */
+OMPI_DECLSPEC extern mca_base_framework_t ompi_hook_base_framework;
+
+
+/**
+ * Dynamically register function pointers to be called from outside of the hook
+ * framework. For example, a collective component could register a callback
+ * at the bottom of init to perform some action.
+ */
+OMPI_DECLSPEC int ompi_hook_base_register_callbacks(ompi_hook_base_component_t *comp);
+OMPI_DECLSPEC int ompi_hook_base_deregister_callbacks(ompi_hook_base_component_t *comp);
+
+/**
+ * Wrapper functions matching the interface functions
+ */
+OMPI_DECLSPEC void ompi_hook_base_mpi_initialized_top(int *flag);
+OMPI_DECLSPEC void ompi_hook_base_mpi_initialized_bottom(int *flag);
+
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_thread_top(int *argc, char ***argv, int required, int *provided);
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_thread_bottom(int *argc, char ***argv, int required, int *provided);
+
+OMPI_DECLSPEC void ompi_hook_base_mpi_finalized_top(int *flag);
+OMPI_DECLSPEC void ompi_hook_base_mpi_finalized_bottom(int *flag);
+
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_top(int argc, char **argv, int requested, int *provided);
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_top_post_opal(int argc, char **argv, int requested, int *provided);
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_bottom(int argc, char **argv, int requested, int *provided);
+OMPI_DECLSPEC void ompi_hook_base_mpi_init_error(int argc, char **argv, int requested, int *provided);
+
+OMPI_DECLSPEC void ompi_hook_base_mpi_finalize_top(void);
+OMPI_DECLSPEC void ompi_hook_base_mpi_finalize_bottom(void);
+
+END_C_DECLS
+
+#endif /* OMPI_BASE_HOOK_H */
--- a/ompi/mca/hook/base/hook_base.c
+++ b/ompi/mca/hook/base/hook_base.c
@@ -1,0 +1,297 @@
+/*
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "ompi/mca/mca.h"
+#include "opal/mca/base/base.h"
+
+#include "opal/runtime/opal.h"
+#include "opal/util/output.h"
+#include "opal/util/show_help.h"
+#include "opal/class/opal_list.h"
+#include "opal/include/opal/prefetch.h"
+
+#include "ompi/constants.h"
+#include "ompi/mca/hook/hook.h"
+#include "ompi/mca/hook/base/base.h"
+
+/*
+ * The following file was created by configure.  It contains extern
+ * statements and the definition of an array of pointers to each
+ * component's public mca_base_component_t struct.
+ */
+#include "ompi/mca/hook/base/static-components.h"
+
+
+// Is the framework open - or has it been closed and we need to reopen it.
+static bool ompi_hook_is_framework_open = false;
+
+static opal_list_t *additional_callback_components = NULL;
+
+
+static int ompi_hook_base_register( mca_base_register_flag_t flags )
+{
+    return OMPI_SUCCESS;
+}
+
+static int ompi_hook_base_open( mca_base_open_flag_t flags )
+{
+    int ret;
+    const mca_base_component_t **static_components = ompi_hook_base_framework.framework_static_components;
+    mca_base_component_list_item_t *cli = NULL;
+    mca_base_component_t *component = NULL;
+    bool found = false;
+
+    additional_callback_components = OBJ_NEW(opal_list_t);
+
+    /* Open up all available components */
+    ret = mca_base_framework_components_open( &ompi_hook_base_framework, flags );
+    if (ret != OMPI_SUCCESS) {
+        return ret;
+    }
+
+    /*
+     * Make sure that the `MCA_BASE_COMPONENT_FLAG_REQUIRED` components defined
+     * as static are loaded. If we find one that was avoided then error out.
+     */
+    if( NULL != static_components ) {
+        for (int i = 0 ; NULL != static_components[i]; ++i) {
+            if( static_components[i]->mca_component_flags & MCA_BASE_COMPONENT_FLAG_REQUIRED ) {
+                // Make sure that this component is in the list of components that
+                // were included in the earlier framework_components_open() call.
+                found = false;
+                OPAL_LIST_FOREACH(cli, &ompi_hook_base_framework.framework_components, mca_base_component_list_item_t) {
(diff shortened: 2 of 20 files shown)


ompi/mca/hook/hook.h
```diff
@@ -0,0 +1,219 @@
...
+#include "mpi.h"
+#include "ompi/mca/mca.h"
+#include "opal/mca/base/base.h"
+
+BEGIN_C_DECLS
+
+/*
+ * Some functions are specially marked. See decoder below.
+ *
+ * *** Static Only (Allways) ***
```
jjhursey: SP: `always`

jjhursey:
Comments from Open MPI [face-to-face](https://github.com/open-mpi/ompi/wiki/Meeting-Minutes-2017-01) meeting:
 * Concern that we are putting in a feature upstream for a component that is not upstream (i.e., licensing example)
 * By adding this option we are misleading the user when they say `-mca fwk foo` but component `bar` is marked as always-consider then they are actually getting `foo,bar`
   - Need to allow the user to know that this is happening. Maybe add a help message to be displayed letting them know that we forced `bar` to be included.
   - Concern about allowing this to be applied to all frameworks.
      - Move the `flag` to be a `hook` framework selection only flag?
 * Use cases mentioned in the room for the always-consider flag: licensing, `btl/self`, `coll/self`
   - Push back on all but the licensing case

**Resolution**:
 * `Required` / `Mandatory` flag : Put flag in base, but enforcement in `ompi/mca/hook/base`
 * `Suggested` / `Recommended` flag : `btl/self`, `coll/self` - Separate PR to add this option.
 * `Disabled` flag : Separate PR to add this option

jjhursey:
If the user tries to avoid a required component then a mess like the following is printed just before `MPI_Init` fails out.
```
Error: A request was made to exclude a hook component from consideration that
is required to be included. This component (noted below) can -not- be excluded
from consideration. The program will fail at this time.
 
Framework: hook
Component: demo
```

jjhursey:
@jsquyres I'm going to ask you to review, if you have time. We'd like to get this into the `v2.x` series if at all possible.

ompi/mca/hook/configure.m4
```diff
@@ -0,0 +1,17 @@
...
+#
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# we only want one :)
+#m4_define(MCA_ompi_hook_CONFIGURE_MODE, STOP_AT_FIRST)
```
jsquyres: I think you need the delete the above 2 lines.

ompi/mca/hook/demo/hook_demo_fns.c
```diff
@@ -0,0 +1,76 @@
...
+#include "hook_demo.h"
+
+#define DEBUG_OUTPUT( fn_name )                     \
+    do {                                            \
+        opal_output(0, "hook:demo: " # fn_name );   \
+    } while(0)
+
+void ompi_hook_demo_mpi_initialized_top(int *flag)
+{
+    DEBUG_OUTPUT( mpi_initialized_top );
```
jsquyres: Did you not want to use `__func__` for some reason?

opal/mca/mca.h
```diff
@@ -315,9 +329,13 @@ struct mca_base_component_2_1_0_t {
...
   /**< Method for registering the component's MCA parameters */
 
+  int32_t mca_component_flags;
+  /**< flags for this component */
+
   /** Extra space to allow for expansion in the future without
-      breaking older components. */
-  char reserved[32];
+      breaking older components.
+      NTH/JJH: reduced by 4 bytes for mca_component_flags */
```
jsquyres: This comment should likely either be explained further or moved to the commit message.  I.e., if you look at this comment 12 months from now, will it still be relevant?  Or stale?  Do we need to know in the code that the padding was reduced by 4?  Or only when spelunking through old commits?  (I'm guessing the latter)

ompi/mca/hook/base/hook_base.c
```diff
@@ -0,0 +1,299 @@
...
+
+/* ***********************************************************************
+ * ***********************************************************************
+ * *********************************************************************** */
+
+/*
+ * If the framework has not been opened, then we can only use the static components.
+ *
+ * Otherwise we would need to initialize opal outside of ompi_mpi_init and possibly
+ * after ompi_mpi_finalize which gets messy (especially when tring to cleanup).
```
jsquyres: tring -> trying

ompi/mca/hook/base/hook_base.c
```diff
@@ -0,0 +1,299 @@
...
+    } while(0)
+
+/*
+ * Once the framework is open then call all available components with
+ * the approprate function pointer. Call order:
+ * 1) static components
+ * 2) dynamic components
+ * 3) 'registered' components (those registered by ompi_hook_base_register_callbacks)
+ */
+#define HOOK_CALL_COMMON_IN_MPI(fn_name, ...)                           \
```
jsquyres: This name (and HOOK_CALL_COMMON_NOT_IN_MPI) is misleading.  You call this macro (or the other one) depending on whether the hook framework has been initialized -- not whether we are in MPI or not.

This may be nit-picky, but it also could apply to OSHMEM someday.

ompi/mca/hook/base/hook_base.c
```diff
@@ -0,0 +1,299 @@
...
+ * after ompi_mpi_finalize which gets messy (especially when tring to cleanup).
+ */
+#define HOOK_CALL_COMMON_NOT_IN_MPI(fn_name, ...)                       \
+    do {                                                                \
+        ompi_hook_base_component_t *component;                          \
+        int idx;                                                        \
+                                                                        \
+        for(idx = 0; NULL != mca_hook_base_static_components[idx]; ++idx ) { \
+            component = (ompi_hook_base_component_t*)mca_hook_base_static_components[idx]; \
+            if( NULL != component->hookm_ ## fn_name && ompi_hook_base_ ## fn_name != component->hookm_ ## fn_name ) { \
```
jsquyres: Nit pick: could you put the 2nd test on a second line?  It will make it easier to read (same for the 2 other places this occurs in the other macro).  Thanks.

ompi/mca/hook/base/hook_base.c
```diff
@@ -0,0 +1,299 @@
...
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "ompi_config.h"
```
jsquyres: `ompi_config.h` should always be first.

ompi/mca/hook/base/hook_base.c
```diff
@@ -0,0 +1,299 @@
...
+
+int ompi_hook_base_register_callbacks(ompi_hook_base_component_t *comp)
+{
+    mca_base_component_list_item_t *cli;
+    bool found = false;
+
+    // Check if it is already there
+    OPAL_LIST_FOREACH(cli, additional_callback_components, mca_base_component_list_item_t) {
+        if( cli->cli_component == (mca_base_component_t*)comp ) {
+            found = true;
```
jsquyres: To-MAY-to, to-MAH-to: you could just `return OMPI_SUCCESS` here and skip the use of `found` altogether.  Doesn't really matter either way -- just makes the code 0.00003% simpler.  😄

ompi/mca/hook/demo/configure.m4
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2016      IBM Corporation.  All rights reserved.
```
jsquyres: 2017.

Should probably update all copyrights to 2017, because that is when it was introduced into the Open MPI community code base.

ompi/mca/hook/base/base.h
```diff
@@ -0,0 +1,56 @@
...
+BEGIN_C_DECLS
+
+/**
+ * Framework struct declaration for this framework
+ */
+OMPI_DECLSPEC extern mca_base_framework_t ompi_hook_base_framework;
+
+
+/**
+ * Register another component to be called
```
jsquyres: Can you put a little more explanation of this functionality here?

ompi/mca/hook/demo/hook_demo_component.c
```diff
@@ -0,0 +1,97 @@
...
+    .hookm_mpi_finalize_top = ompi_hook_demo_mpi_finalize_top,
+    .hookm_mpi_finalize_bottom = ompi_hook_demo_mpi_finalize_bottom,
+};
+
+ompi_hook_base_component_t hook_demo_extra_component = {
+    /* Component functions */
+    .hookm_mpi_init_bottom = ompi_hook_demo_extra_mpi_init_bottom,
+};
+
+static int ompi_hook_demo_component_open(void)
```
jsquyres: Since this is a demo component, can you add some verbose comments here explaining the use of this extra component and the extra registration / deregistration?

ompi/mca/hook/demo/hook_demo_component.c
```diff
@@ -0,0 +1,97 @@
...
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+
+#include "hook_demo.h"
```
jsquyres: Should probably also include `opal_output.h` in here, since you call `opal_output()`.

ompi/mca/hook/demo/hook_demo_fns.c
```diff
@@ -0,0 +1,76 @@
...
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+
+#include "hook_demo.h"
```
jsquyres: Should probably also include `opal_output.h` in here, since you call `opal_output()`.

ompi/mca/hook/demo/hook_demo_fns.c
```diff
@@ -0,0 +1,76 @@
...
+
+void ompi_hook_demo_mpi_finalize_bottom(void)
+{
+    DEBUG_OUTPUT( mpi_finalize_bottom );
+}
+
+
+void ompi_hook_demo_extra_mpi_init_bottom(int argc, char **argv, int requested, int *provided)
+{
+    DEBUG_OUTPUT( mpi_init_bottom (extra) );
```
jsquyres: I could not find where `mpi_init_bottom` was defined, nor `extra`..., but yet this file still compiled for me.  Whaaa...?  I have to run now and can't investigate further...

jsquyres:
Also, given that this framework will almost certainly be used with out of tree components, there should definitely be thought given to versioning of the components and modules, and also how to expand the component and or module without screwing out-of-tree components.

jsquyres:
One last thought: can you put a README[.md] in the `demo` component?  It would be good to explain a few things to someone who is coming in cold to the `hook` framework.

ompi/mca/hook/demo/hook_demo_fns.c
jjhursey: No reason. I suppose I just wanted to make work for myself. :) I'll clean that up on the next iteration.

ompi/mca/hook/demo/hook_demo_fns.c
jjhursey: I think the macro was masking it by interpreting it as a string. It was a leftover from an older version - I'll clean that out.

ompi/mca/hook/base/hook_base.c
jjhursey: 👍

jjhursey:
@jsquyres I just pushed an update that I think addresses all of your comments except the `demo/Readme.md`. I can try to add that today if you think it's still necessary.

This was delayed a bit because I was experimenting with how best to support forwards compatible components. I have enough of a solution (outside of this PR on a side branch) to feel comfortable moving forward.

jjhursey:
@jsquyres Thanks for the review!

PR #2779, opened 2017-01-22
oob/base: fix num_threads registration type

hjelmn:
This commit fixes a bug in the registration of the num_threads MCA
variable. The variable is of type int and was being registered as
a boolean.


hjelmn:
@rhc54 Simple fix.

PR #2786, opened 2017-01-23
Pr/sharedfp append fix v2.x

edgargabriel:
Fixes a bug reported on the mailing list. ompio did only reposition the individual
file pointer when the file was opened in append mode. Set the shared file
pointer also to point to the end of the file, similarly to the individual
file pointer.

This is the equivalent to commit d3a8d38
on master, cannot be cherry-picked because of differences in the organization
of the ompio files ( usage of mca/common/ompio on master).

### DIFF
--- a/ompi/mca/io/ompio/io_ompio_component.c
+++ b/ompi/mca/io/ompio/io_ompio_component.c
@@ -49,7 +49,7 @@ int mca_io_ompio_bytes_per_agg = OMPIO_PREALLOC_MAX_BUF_SIZE;
 int mca_io_ompio_num_aggregators = -1;
 int mca_io_ompio_record_offset_info = 0;
 int mca_io_ompio_coll_timing_info = 0;
-int mca_io_ompio_sharedfp_lazy_open = 1;
+int mca_io_ompio_sharedfp_lazy_open = 0;
 
 int mca_io_ompio_grouping_option=5;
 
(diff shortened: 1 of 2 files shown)


jsquyres:
Just like #2787:

@edgargabriel Got a compile fail:

```
io_ompio_file_open.c: In function 'ompio_io_ompio_file_open':
io_ompio_file_open.c:257:17: error: 'shared_fp_base_module' undeclared (first use in this function)
                 shared_fp_base_module = ompio_fh->f_sharedfp;
                 ^
io_ompio_file_open.c:257:17: note: each undeclared identifier is reported only once for each function it appears in
```

Can you have a look?

edgargabriel:
argh, copy-and-paste error, a line is missing, Will fix.

jsquyres:
@hppritcha Good to go.

jsquyres:
Shoot -- @edgargabriel just mentioned on the v2.0.x version (https://github.com/open-mpi/ompi/pull/2787#issuecomment-274885235) that we should have waited merging.  My fault for not seeing that fast enough... ☹️

edgargabriel:
I will provide a separate pr for 2.x for the outstanding item. its probably more important to get it in correctly for 2.0.x right now

jsquyres:
Thanks @edgargabriel

PR #2788, opened 2017-01-23
libevent2022: Fix broken configure AC_LANG_PROGRAM

jjhursey:
* The AC_LANG_PROGRAM macro adds the `main()` so it is erroneous
   to add it to the test program.
 * This was detected with the XL compilers which will fail to
   build the program in this situation. The GNU compiler does not
   error out or warn, but successfully compiles the program.

The resulting test program looked like
```c
int
main()
{
  {
  // ...
  }
}
```

And XL complained with
```shell
error: function definition is not allowed here
```


jjhursey:
Note that I do not see this section of configure logic in the `libevent 2.0.22` release. So we must have added this logic and carried it foward?

PR #2792, opened 2017-01-24
libevent2022: Fix broken configure AC_LANG_PROGRAM

jjhursey:
* Similar to commit https://github.com/open-mpi/ompi/commit/029964a748230475bc4c950334bd85d593e2352e This removes an extra `int main` during configure.
 * This commit should have been included in the previous PR.

PR #2793, opened 2017-01-24
libevent2021: Fix broken configure AC_LANG_PROGRAM

jjhursey:
* The AC_LANG_PROGRAM macro adds the `main()` so it is erroneous
   to add it to the test program.
 * This was detected with the XL compilers which will fail to
   build the program in this situation. The GNU compiler does not
   error out or warn, but successfully compiles the program.


jjhursey:
References PR #2788 and PR #2792 from `master` - custom patch because of the difference in `libevent` versions.

PR #2808, opened 2017-01-24
coll: Move reduce_local into the coll framework

jjhursey:
Per Face-to-Face in Aug. 2016:
 * https://github.com/open-mpi/ompi/wiki/Meeting-2016-08

### DIFF
--- a/ompi/communicator/communicator.h
+++ b/ompi/communicator/communicator.h
@@ -20,6 +20,7 @@
  * Copyright (c) 2014-2015 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -232,6 +233,28 @@ typedef struct ompi_communicator_t ompi_communicator_t;
  * size so when the bitness changes the size of the handle changes.
  * This is done so we don't end up needing a structure that is
  * incredibly larger than necessary because of the bitness.
+ *
+ * This padding mechanism works as a (likely) compile time check for when the
+ * size of the ompi_communicator_t exceeds the predetermined size of the
+ * ompi_predefined_communicator_t. It also allows us to change the size of
+ * the ompi_communicator_t without impacting the size of the
+ * ompi_predefined_communicator_t structure for some number of additions.
+ *
+ * As an example:
+ * If the size of ompi_communicator_t is less than the size of the _PAD then
+ * the _PAD ensures that the size of the ompi_predefined_communicator_t is
+ * whatever size is defined below in the _PAD macro.
+ * However, if the size of the ompi_communicator_t grows larger than the _PAD
+ * (say by adding a few more function pointers to the structure) then the
+ * 'padding' variable will be initialized to a large number often triggering
+ * a 'array is too large' compile time error. This signals two things:
+ * 1) That the _PAD should be increased.
+ * 2) That users need to be made aware of the size change for the
+ *    ompi_predefined_communicator_t structure.
+ *
+ * Q: So you just made a change to communicator structure, do you need to adjust
+ * the PREDEFINED_COMMUNICATOR_PAD macro?
+ * A: Most likely not, but it would be good to check.
  */
 #define PREDEFINED_COMMUNICATOR_PAD (sizeof(void*) * 192)
 
--- a/ompi/mpi/c/reduce_local.c
+++ b/ompi/mpi/c/reduce_local.c
@@ -15,6 +15,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016-2017 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -45,13 +46,15 @@ static const char FUNC_NAME[] = "MPI_Reduce_local";
 int MPI_Reduce_local(const void *inbuf, void *inoutbuf, int count,
                      MPI_Datatype datatype, MPI_Op op)
 {
+    int err;
+
     MEMCHECKER(
         memchecker_datatype(datatype);
     );
 
     if (MPI_PARAM_CHECK) {
         char *msg;
-        int err = MPI_SUCCESS;
+        err = MPI_SUCCESS;
         OMPI_ERR_INIT_FINALIZE(FUNC_NAME);
 
         if (MPI_OP_NULL == op || NULL == op) {
@@ -75,9 +78,14 @@ int MPI_Reduce_local(const void *inbuf, void *inoutbuf, int count,
 
     /* Invoke the op component to perform the back-end operation */
     OBJ_RETAIN(op);
-    /* XXX -- CONST -- do not cast away const -- update mca/coll */
-    ompi_op_reduce(op,(void *) inbuf, inoutbuf, count, datatype);
+    OBJ_RETAIN(datatype);
+    // Since there is no 'comm' parameter to this interface use 'self' since
+    // this is a local operation to this process.
+    ompi_communicator_t *comm = &ompi_mpi_comm_self.comm;
+    err = comm->c_coll.coll_reduce_local(inbuf, inoutbuf, count, datatype, op,
+                                         comm->c_coll.coll_reduce_local_module);
+    OBJ_RELEASE(datatype);
     OBJ_RELEASE(op);
 
-    return MPI_SUCCESS;
+    OMPI_ERRHANDLER_RETURN(err, MPI_COMM_WORLD, err, FUNC_NAME);
 }
--- a/ompi/mca/coll/base/coll_base_comm_select.c
+++ b/ompi/mca/coll/base/coll_base_comm_select.c
@@ -19,7 +19,7 @@
  *                         reserved.
  * Copyright (c) 2014      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2017 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -52,7 +52,7 @@ struct avail_coll_t {
     opal_list_item_t super;
 
     int ac_priority;
-    mca_coll_base_module_2_1_0_t *ac_module;
+    mca_coll_base_module_2_2_0_t *ac_module;
     const char * ac_component_name;
 };
 typedef struct avail_coll_t avail_coll_t;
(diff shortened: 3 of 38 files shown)


jjhursey:
@hjelmn @bosilca Since I'm changing the size of `mca_coll_base_comm_coll_t` and `mca_coll_base_module_2_1_0_t` do I need to:
 * Change the `ompi_communicator_t` padding?
 * Update the version of the framework from `2.1.0` to `2.2.0`?

bosilca:
We were just talking about this during the f2f meeting. Temporary solution is to increase the size of the communicator as long as you don't grow it more than the predefined object.

jjhursey:
@bosilca Would you mind checking my change to the padding to make sure I interpreted how to change that variable correctly. There is a separate commit for the versioning/padding than the main commit. The versioning just touched a bunch of files.

ibm-ompi:

 Gist: https://gist.github.com/fe0e91f97591ef2cc5260f717698aada

ibm-ompi:

 Gist: https://gist.github.com/5b686d647f5337d119c35a3b34cb336a

ibm-ompi:

 Gist: https://gist.github.com/61aa2bb2dceabfcf8568b2f40fc8f93a

jjhursey:
Disregard the IBM CI failures - our relay failed. I've disabled Jenkins while that gets resolved.

artpol84:
bot:mellanox:retest

bosilca:
@jjhursey I added a sketch of the approach we discussed during the devel meeting. The collective are now in their own structure, and the ompi_communicator_t is down to 352 bytes (thus we can make the predefined 512 bytes long).

We can try to save a little extra memory (and certainly time during communicators creation), by delaying the initialization of the collective modules until needed, aka. the first collective communication is called. Thus, communicators that will never participate in collective communications we will not have to instantiate this structure.

gpaulsen:
:bot:retest

ibm-ompi:

 Gist: https://gist.github.com/602e24f79ffa7b8f01f6887a6bf38824

ibm-ompi:

 Gist: https://gist.github.com/29039a92f5e38f664623bdd095759af1

ibm-ompi:

 Gist: https://gist.github.com/718fc59cb72eb6adb23f502123200d2f

jjhursey:
@gpaulsen There are 3 commits so far in the PR. 2 from IBM and 1 from UTK. @bosilca is working on restructuring how we hang the collective module structure on the communicator (per the face-to-face meeting). Since adding a new pointer to the module structure (for `reduce_local`) would force an ABI break it is a good time to try to do the communicator update too.

The CI breaks seem to indicate that the `master` has changed and this PR needs to be re-sync'ed. I'll put that on my todo for today

jjhursey:
bot:mellanox:retest

jjhursey:
@bosilca What do you think about splitting this into two PRs. One for the reduce_local and a follow-on for the coll restructuring?

I think your patch is complete now (with one addition from me), so it should be easy to spin up another PR. I like the idea of having mca/coll be a lazy initialization framework. That should help communicator creation time.

bosilca:
👍  for the split. Can you do it ?

jjhursey:
@bosilca Yeah I'll do it tonight/tomorrow and tag you in the PR.

jjhursey:
I just rolled this commit back to just be the `reduce_local` change. I'll file a separate PR for the coll restructuring.

jjhursey:
@gpaulsen @bosilca I think this is ready to go if you want to have another look at it.

@gpaulsen per your review comments:
 -  Re `memchecker.h`: A number of the collectives are not listed there (e.g., nonblocking collectives). That file should be re-audited in a different PR.
 - For the `libnbc` I did not add the `reduce_local` there as it has been added to `basic`, `inter`, and `self` which cover all of the bases so there is no need for `libnbc` to have a version of it as well.
 - I've updated the copyrights on all of the files.

jjhursey:
@gpaulsen Can you re-review?

ompi/mpi/c/reduce_local.c
```diff
@@ -76,8 +79,12 @@ int MPI_Reduce_local(const void *inbuf, void *inoutbuf, int count,
...
-    ompi_op_reduce(op,(void *) inbuf, inoutbuf, count, datatype);
+    // Since there is no 'comm' parameter to this interface use 'self' since
+    // this is a local operation to this process.
+    ompi_communicator_t *comm = &ompi_mpi_comm_self.comm;
+    err = comm->c_coll.coll_reduce_local(inbuf, inoutbuf, count, datatype, op,
+                                         comm->c_coll.coll_reduce_local_module);
     OBJ_RELEASE(op);
 
-    return MPI_SUCCESS;
+    OMPI_ERRHANDLER_RETURN(err, MPI_COMM_WORLD, err, FUNC_NAME);
```
gpaulsen: The ErrHandler specifies MPI_COMM_WORLD, but the reduce local was done on MPI_COMM_SELF.  I believe we should invoke the error handler on MPI_COMM_SELF instead.

You should also OBJ_RETAIN and OBJ_RELEASE the datatype.

ompi/communicator/communicator.h
```diff
@@ -233,7 +234,7 @@ typedef struct ompi_communicator_t ompi_communicator_t;
  * This is done so we don't end up needing a structure that is
  * incredibly larger than necessary because of the bitness.
  */
-#define PREDEFINED_COMMUNICATOR_PAD (sizeof(void*) * 192)
+#define PREDEFINED_COMMUNICATOR_PAD (sizeof(void*) * 194)
```
gpaulsen: Please add a comment of where 192 / 194 came from.  I presume you incremented by two since you added two pointers to struct mca_coll_base_comm_coll_t ?

jjhursey: The comment above this section of code describes how to set the padding. So I don't think we need any further additions here.

ompi/mpi/c/reduce_local.c
jjhursey: The use of `MPI_COMM_WORLD` matches the other error handling in the param check earlier in the file.

I used comm_self as the reference point for the `reduce_local`.

Other collectives don't do an `OBJ_RETAIN` / `OBJ_RELEASE` on the datatype so I did not do that here.

ompi/communicator/communicator.h
gpaulsen: Well, when *I* read all of that, it seems like you shouldn't EVER change the PREDEFINED_COMMUNICATOR_PAD (since that was set long ago), and that the padding will auto-shrink as you add fields to ompi_communicator_t), though perhaps we could reset it at a MAJOR.0.0 release (since we require users to recompile/relink at that time).

So, I still don't understand how you can change the PREDEFINED_COMMUNICATOR_PAD without breaking backwards compatibility.

ompi/mpi/c/reduce_local.c
gpaulsen: I think the problem is that users who set error handlers rarely set them on MPI_COMM_SELF, so I suppose what you have is "fine" (rather than changing the param checking lines to MPI_COMM_Self which would be more correct).

The OBJ_RETAIN / OBJ_RELEASE is a bug waiting to happen, but it does take time, and I doubt anyone but a test case would actually do that.  And when it does we can handle that all in a different PR.

bosilca: MPI 3.1 section 8.3 page 340 line 21, specify that MPI calls not attached to an MPI object (communicator, file or window) are attached to MPI_COMM_WORLD. Basically, all error handling for such call are required to be raised in the context of MPI_COMM_WORLD.

@gpaulsen is right about the retain/release. However, the right approach will be to do it in the API and not in the actual implementation (such that when we use the function internally we don't waste time in unnecessary retain/release).

jjhursey: @bosilca That sounds fine. I can add the retain/release at the API level. Where to do the retain/release is something that is being discussed on Issue #2151, right?

ompi/communicator/communicator.h
jjhursey: That's a fair point. I can add the following language to the comment:
```
To keep the ompi_predefined_communicator_t structure stable in size padding is
added to it beyond the size of the ompi_communicator_t. For example, If a single
pointer is added to ompi_communicator_t then the PREDEFINED_COMMUNICATOR_PAD
must by increased by 1 to keep the size of the ompi_predefined_communicator_t
structure the same.
```

What do you think?

jjhursey: In discussing this more with @gpaulsen I'm now seeking clarification myself. Maybe @bosilca and @jsquyres can clarify how (if at all) the `PREDEFINED_COMMUNICATOR_PAD` should be changed when the communicator changes.

For example:
```
Assume:
  sizeof(void*) = 8 bytes
  ompi_communicator_t = 80 bytes
  PREDEFINED_COMMUNICATOR_PAD = 800 bytes (sizeof(void*) * 100)
Then:
  ompi_predefined_communicator_t = 80 + (800 - 80) = 800 bytes
```

So if I were to add 1 pointer to the `ompi_communicator_t` and **not** change `PREDEFINED_COMMUNICATOR_PAD` that works out to:
```
Assume:
  sizeof(void*) = 8 bytes
  ompi_communicator_t = 88 bytes
  PREDEFINED_COMMUNICATOR_PAD = 800 bytes (sizeof(void*) * 100)
Then:
  ompi_predefined_communicator_t = 88 + (800 - 88) = 800 bytes
```

So from my understanding I don't think the padding should ever have to change until the `sizeof(ompi_communicator_t)` becomes larger than `PREDEFINED_COMMUNICATOR_PAD`.

Is that logic correct or are we missing something here?

jjhursey:
@gpaulsen I've updated `reduce_local` to `OBJ_RETAIN/OBJ_RELESE` the datatype. I'm going to hold on merging this until we get a comment back about the padding question. Whatever the resolution I'll add a comment to that section of the code to help clarify how that variable should/shouldn't be modified in a scenario like this.

ompi/communicator/communicator.h
bosilca: This is the correct logic, the padding only changes when the sizeof the communicator structure outgrows it.

jjhursey: Ok. Let me double check the math to see if we needed to change it. I don't think we do, but I'd like to make sure.

So the padding works as a compile time check for backwards compatibility on that structure size. You can add pointers to the communicator structure (or something hanging off of it) and if at any point you exceed the padding then you should get a compiler error flagging this as an issue.

gpaulsen:
Thanks.  I like the page long high level comment above it describing the overall design.  But something RIGHT THERE that says "DONT TOUCH THIS NUMBER UNLESS <foo>" is what I'd like.  :)
Yes, all caps! :)

jjhursey:
I updated the comment above `PREDEFINED_COMMUNICATOR_PAD`. Can folks take a look and let me know that it's 👌 ? I kept it as a separate commit for now to make it easy to review. I'll squash that down once we are ready to go.

I also reverted the size of the pad back to `192` since we did not need to adjust it in the first place.
The current size of:
```
sizeof(void*) = 8 bytes
PREDEFINED_COMMUNICATOR_PAD = 1536 bytes (192 pointers)
sizeof(ompi_communicator_t) = 1064 bytes
```

jjhursey:
Travis CI failure due to script timeout while downloading gcc. All other passed.

PR #2809, opened 2017-01-24
opal/output: Make sure verbose gets updated when id 0 gets updated.

jjhursey:
- This allows the following MCA option to have an impact on the
   framework verbose output as well.
   * `-mca mca_base_verbose stdout`

### DIFF
--- a/opal/util/output.c
+++ b/opal/util/output.c
@@ -15,6 +15,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


jjhursey:
@jsquyres Just wanted another set of eyes on this before pushing to `master`.

jjhursey:
Adding some context here. So in the process of trying to help a user we found that `mca_base_verbose` was not working. We wanted to be able to redirect the verbose output on stream `0` to something other than the default `stderr` (which is then flagged as `stddiag` by the orted if you run with `--tag-output`).

This MCA parameter has the following description:

> MCA mca base: parameter "mca_base_verbose" (current value: "stderr", data source: default, level: 9 dev/all, type: string, synonyms: mca_verbose)
>
>  Specifies where the default error output stream goes (this is separate from distinct help messages).  Accepts a comma-delimited list of: stderr, stdout, syslog, syslogpri:<notice|info|debug>, syslogid:<str> (where str is the prefix string for all syslog notices), file[:filename] (if filename is not specified, a default filename is used), fileappend (if not specified, the file is opened for truncation), level[:N] (if specified, integer verbose level; otherwise, 0 is implied)

However, we found that it was not being honored when running an example like:
```shell
shell$ mpirun -np 2 --tag-output -mca mca_base_verbose stdout -mca coll_base_verbose 9 ./hello_c
[1,0]<stddiag>:[mpi01:34218] coll:tuned:component_open: done!
[1,1]<stddiag>:[mpi01:34219] coll:tuned:component_open: done!
[1,0]<stddiag>:[mpi01:34218] coll:base:comm_select: new communicator: MPI_COMM_WORLD (cid 0)
...
```

With this patch the output looks like:
```shell
shell$ mpirun -np 2 --tag-output -mca mca_base_verbose stdout -mca coll_base_verbose 9 ./hello_c
[1,0]<stdout>:[mpi01:20905] coll:tuned:component_open: done!
[1,1]<stdout>:[mpi01:20906] coll:tuned:component_open: done!
[1,0]<stdout>:[mpi01:20905] coll:base:comm_select: new communicator: MPI_COMM_WORLD (cid 0)
...
```

This is helpful if the user wants the Open MPI verbose output to be routed over `stdout` along with their application out for debugging purposes.

This bug fix pairs nicely with the `orte_map_stddiag_to_stderr` and (new) `orte_map_stddiag_to_stdout` options as well.

Issue #2811, opened 2017-01-24
Failure in Finalize when using Java Bindings


nrgraham23:
A pointer that was not allocated is being freed in the Finalize call.  This causes an endless loop of errors to occur, and you have to use the kill command to stop the Java processes (ctrl + c does not work).

This problem exists on master, but not in v2.0.x.  I will check the other release branches.

ggouaillardet:
@nrgraham23 i will take care of that.

i recently plugged a bunch of memory leaks, and we end up in a non supported case here :
```MPI_Init()``` is not invoked from the main thread, so some data ends up being free'd twice

nrgraham23:
I suspected it was related to the memory leak fixes.  Your pull request fixes the problem.

nrgraham23:
@jsquyres @hppritcha Would it be possible to add a very simple sanity test to MTT for the Java bindings?  It was quite a while from when the Java bindings must have been broken to when I noticed they were broken.  Maybe something as simple as the Hello.java in the examples directory.

PR #2815, opened 2017-01-25
opal/threads: protect opal_tsd_keys_destruct() to fix Java bindings.

ggouaillardet:
When Java bindings are used, MPI_Init() is not invoked
by the main thread, and this causes some keys being destructed twice.
Reset the per thread values to NULL in order to correctly handle this

Fixes open-mpi/ompi#2811


nrgraham23:
+1 This fixes the problem I was seeing.

hppritcha:
@ggouaillardet does this need to go in to v2.x/v2.0.x?

ggouaillardet:
@hppritcha i did not PR any of my memory leak plugs into v2.x/v2.0.x yet, so for now, this PR should not be backported.
once all leaks are plugged in master (some leaks involve PMIx, see pmix/master#265), i will likely PR all of them at once into v2.x

PR #2816, opened 2017-01-25
Improve launch performance

rhc54:
Improve launch performance by:

* eliminating nested loop - use the "node mapped" flag to determine if a node needs to be added to the job map, and then clear all the used nodes' flags after completing the assignments.

* eliminating the return of local node topology unless mpirun detects a different signature that it hasn't seen before


jsquyres:
@hjelmn Please review.

hppritcha:
bot:lanl:retest

hppritcha:
LANL-CI down today as NERSC is doing quarterly maintenance.

PR #2829, opened 2017-01-25
Allow all tuned MCA parameters to be modified programatically.

bosilca:
This brings back a lost feature that would allow the users to fine-grain tune what collectives will be used for all new communicators.
In addition: Fix a comment in the MCA header.


hppritcha:
@bosilca please fix the signed off.

artpol84:
bot:mellanox:retest

PR #2831, opened 2017-01-25
pml/base: Expose some bsend varaibles so PMLs may reference them

jjhursey:
* This was useful when trying to optimize our PML - having access to the user pointers.

### DIFF
--- a/ompi/mca/pml/base/pml_base_bsend.c
+++ b/ompi/mca/pml/base/pml_base_bsend.c
@@ -15,6 +15,7 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC.  All rights
  *                         reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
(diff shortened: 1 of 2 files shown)


bosilca:
@jjhursey I would be curious to see the rationale behind this change ?

artpol84:
bot:mellanox:retest

jjhursey:
@bosilca We needed access to the `mca_pml_bsend_base` to check if it is a CUDA buffer that they passed to `MPI_Buffer_attach` then called `MPI_Bsend` later. I couldn't find a better way to access that user address without re-working where `mca_pml_base_bsend_attach` is located.

jjhursey:
@ggouaillardet has a commit [here](https://github.com/open-mpi/ompi/commit/ad44ecb2ba371c10642344f90b01a97e957d241e#commitcomment-20694838) that initializes these variables that needs to be included in the PRs to the release branches (when I make them).

jjhursey:
Ref PR #2895 and PR #2896

PR #2833, opened 2017-01-25
oob/usock: Protect against possible NULL peer

jjhursey:
* It is possible that this function is called with a `NULL` peer in an error path where the peer is unclear. This protects us from a segv.
 * Not on `master` so nothing to cherry-pick


artpol84:
bot:mellanox:retest

PR #2838, opened 2017-01-25
Remove the enable-heterogeneous config option

### LINKED ISSUE #2802: remove --enable-heterogenous configury option
Mixed big endian/little endian no longer works in Open MPI and there are no plans currently to fix it.  To avoid user confusion, remove the configure --enable-heterogenous option.  Related to issue #639.

Rather than just removing entirely from the configury, it would be better to have configure print out an error message if this option is attempted and abort.

nrgraham23:
This commit removes the enable-heterogeneous option.
If the user attempts to use this option, configure
will print a warning message and error out.

fixes #2802 

@hppritcha @jsquyres

### DIFF
--- a/config/opal_configure_options.m4
+++ b/config/opal_configure_options.m4
@@ -297,11 +297,12 @@ AC_ARG_ENABLE([heterogeneous],
                      platform support (default: disabled)])])
 if test "$enable_heterogeneous" = "yes" ; then
      AC_MSG_RESULT([yes])
-     opal_want_heterogeneous=1
-else
-     AC_MSG_RESULT([no])
-     opal_want_heterogeneous=0
+     AC_MSG_WARN([Heterogeneous support is not currently supported in Open MPI])
+     AC_MSG_ERROR([cannot continue])
 fi
+
+opal_want_heterogeneous=0
+
 AC_DEFINE_UNQUOTED([OPAL_ENABLE_HETEROGENEOUS_SUPPORT],
                    [$opal_want_heterogeneous],
                    [Enable features required for heterogeneous support])


nrgraham23:
Here is the fix without removing the dead code this PR creates.  I can modify it to go ahead and remove that code if it is decided that is what is wanted.

nrgraham23:
@jsquyres @hppritcha Is this ready to be merged?

rhc54:
I believe that @ggouaillardet wanted to retain it and was offering to maintain it - if that is true, then we really shouldn't remove it.

nrgraham23:
He did.  I assumed that Jeff and Howard had seen his comments on issue #2802 and had decided to go ahead with the removal anyways, but if this is not the case, let me know and I can close this.

hppritcha:
I think we should go ahead with this PR and when @ggouaillardet fixes existing issues with hetero he can reenable.

rhc54:
@ggouaillardet We talked about this on the telecon this morning, and bottom line is that we will apply our usual policy here: if someone is willing to maintain and test this feature, then no problem leaving it in the code. This probably applies sooner to v2.1 than v2.0.3, but if you can update the support and at least post on the PR your MTT results, then we are content to leave this alive.

jsquyres:
@ggouaillardet Do you have plans to fix the heterogeneous support?

ggouaillardet:
@jsquyres i made #2940 with some fixes i did not yet push into master
i will likely ask some extra pair of eyes to review them before commiting

jsquyres:
Ok.  So the intent is definitely to fix heterogeneous -- let's close this PR, then.

PR #2840, opened 2017-01-26
verbs: remove extra event user increment/decrement operation

### LINKED ISSUE #2839: TCP and OpenIB BTLs cause performance degradation in unrelated frameworks
We have observed this both in OSHMEM UCX testing and now in HCOLL shared memory testing on POWER8 Garrison platform. In the 2.x series, when the BTLs are silently loaded to cover OSC we see the following when running HCOLL shared memory collectives. 

Running with 160 threads
 `mpirun -np 160 -mca coll_hcoll_enable 1 ./IBM-MPI1 Allreduce`

8-byte allreduce latency:
`21.9`

 `mpirun -np 160 -mca btl ^tcp,openib -mca coll_hcoll_enable 1 ./IBM-MPI1 Allreduce`

8-byte allreduce latency:
`11.58`

 @hjelmn and I discussed this at the meeting, and the root cause appears to be the asynchronous progress thread in these two BTLs.

hjelmn:
Since the oob and connections systems do not work the same way they
did in older versions of Open MPI these operations are no longer
necessary. At best they do nothing and at worst they hurt performance
by making us enter the event library more often in opal_progress().

Fixes #2839

### DIFF
--- a/opal/mca/btl/openib/btl_openib_endpoint.c
+++ b/opal/mca/btl/openib/btl_openib_endpoint.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006-2013 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2006-2016 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2006-2017 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2006-2007 Voltaire All rights reserved.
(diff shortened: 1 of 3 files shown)


hjelmn:
This may be responsible for performance regressions on v2.0.x from v1.10.x on any platform that support unconnected datagrams using libibverbs. This is a bug and I will move it into master after CI. I will open the 2.0.x and 2.1.x PRs now.

jladd-mlnx:
We confirmed that this Fixes #2839. Thanks, @hjelmn !!

hjelmn:
bot:mellanox:retest

hjelmn:
Since we have confirmation from Intel and mellanox that the issue is fixed i will go ahead and merge this. Mellanox Jenkins is wedged.

PR #2824, opened 2017-01-25
Fix for Ireduce + MPI_IN_PLACE

gpaulsen:
Fixes a wrong answer from MPI_Ireduce when the red_sched_chain()
path was taken (which only happens for np<=4 and mesgsize>=64k).

The way libnbc treats MPI_ANY_SOURCE is to set sbuf == rbuf, and
whether an algorithm will work cleanly or not after that depends on the
details.

In this case the last steps of the algorithm amounted to
    (right neighbor is sending us reduction results from ranks 1..n-1)
    recv into rbuf from right neighbor
    add the contribution from our sbuf into rbuf
this would be fine in general, but if sbuf==rbuf, that recv overwrites
the sbuf. I changed it to recv into a tmpbuf if ANY_SRC was used.

Signed-off-by: Geoffrey Paulsen <gpaulsen@us.ibm.com>

PR #2854, opened 2017-01-26
Fix for Ireduce + MPI_IN_PLACE.

hppritcha:
Fixes a wrong answer from MPI_Ireduce when the red_sched_chain()
path was taken (which only happens for np<=4 and mesgsize>=64k).

The way libnbc treats MPI_IN_PLACE is to set sbuf == rbuf, and
whether an algorithm will work cleanly or not after that depends on the
details.

In this case the last steps of the algorithm amounted to
    (right neighbor is sending us reduction results from ranks 1..n-1)
    recv into rbuf from right neighbor
    add the contribution from our sbuf into rbuf
this would be fine in general, but if sbuf==rbuf, that recv overwrites
the sbuf. I changed it to recv into a tmpbuf if MPI_IN_PLACE was used.

related to #2824
@gpaulsen 
(cherry picked from commit 045d0c5f4caffdf31fb2dd1cb4d7b8940c07d365)


jjhursey:
Refs PR #2824

PR #2857, opened 2017-01-27
pmix/ext11 fixes, plugs and rename

ggouaillardet:
- fix compilation issue
- rename pmix/ext11 component into pmix/ext1x
- plugs misc memory leaks


ggouaillardet:
@rhc54 @jjhursey could you please review these ?

the first commit is hopefully a no brainer.

the second fixes a race condition i observe at startup with Open MPI ```master``` and external PMIx ```v1.2```.
i reviewed the code of other branches, and it seems this should be backported to ```v2.0.x``` and ```v2.x``` ```pmix/pmix112``` glue too.

btw, in master, the external component is called ```pmix/ext11``` but it is used for PMIx ```v1.2```.
is this supposed to work ? or should there be yet an other ```pmix/ext12``` component ? if not, could we
rename this ```pmix/ext1x``` in order to avoid this kind of confusion ?

rhc54:
@ggouaillardet we should rename the component in master ext1x as you suggest - there is no integration difference between PMIx v1.1 and v1.2

ggouaillardet:
@rhc54 ok, i will take care of that from Monday
(e.g. merge this PR, PR to the v2 branches and rename the ext11 component)

PR #2859, opened 2017-01-27
v2.x: pmix/pmix112: Correct `--enable-pmix-dstore` option

kawashima-fj:
@jjhursey This is the intended option, right?

The Open MPI `configure` option to enable PMIx shared meory dstore
is now `--enable-pmix-dstore`, not `--enable-pmix3-dstore`.
Without this fix, The `--disable-dstore` is always passed to PMIx
`configure` regardless of the Open MPI `configure` option.

And, `==` is bash extention.

This bug exists only in v2.x branch. No master and v2.0.x.

### DIFF
--- a/opal/mca/pmix/pmix112/configure.m4
+++ b/opal/mca/pmix/pmix112/configure.m4
@@ -43,7 +43,7 @@ AC_DEFUN([MCA_opal_pmix_pmix112_CONFIG],[
                   [AC_HELP_STRING([--enable-pmix-dstore],
                                   [Enable PMIx shared memory data store (default: disabled)])])
     AC_MSG_CHECKING([if PMIx shared memory data store is enabled])
-    if test "$enable_pmix3_dstore" == "yes"; then
+    if test "$enable_pmix_dstore" = "yes"; then
         AC_MSG_RESULT([yes])
         opal_pmix_pmix_sm_flag=--enable-dstore
     else


hppritcha:
bot:lanl:retest

opal/mca/pmix/pmix112/configure.m4
```diff
@@ -43,7 +43,7 @@ AC_DEFUN([MCA_opal_pmix_pmix112_CONFIG],[
                   [AC_HELP_STRING([--enable-pmix-dstore],
                                   [Enable PMIx shared memory data store (default: disabled)])])
     AC_MSG_CHECKING([if PMIx shared memory data store is enabled])
-    if test "$enable_pmix3_dstore" == "yes"; then
+    if test "$enable_pmix_dstore" = "yes"; then
```
jjhursey: Let's change this selection statement to match `master` to make it easier to compare the branches:
```m4
    if test "$enable_pmix_dstore" != "no"; then
```

kawashima-fj:
@jjhursey Updated. Please review again.

I confirmed options passed to PMIx with this change.
- Open MPI with `--enable-pmix-dstore` → PMIx `--enable-dstore`
- Open MPI with `--disable-pmix-dstore` → PMIx `--disable-dstore`
- Open MPI without `--{enable,disable}-pmix-dstore` → PMIx `--enable-dstore`

kawashima-fj:
Failure of LANL-disable-dlopen is caused by enabling dstore because `--disable-dstore` was default for the embedded PMIx before this PR and this PR changes it. LANL-disable-dlopen runs an MPI program as a singleton. So singleton + dstore seems to have a problem.

I can reproduce the problem on my local environment. An MPI program spawned by mpiexec runs successfully but singleton fails. The singleton ESS in MPI program forks orted with the `--hnp` option and receives an orted URI via `read`. The URL does not have an expected format (probably an empty string). We need to check `orted --hnp` with dstore.

In source code, the `fork_hnp` function in `orte/mca/ess/singleton/ess_singleton_module.c` returns `ORTE_ERR_BAD_PARAM`.

https://github.com/open-mpi/ompi/blob/v2.x/orte/mca/ess/singleton/ess_singleton_module.c#L615

rhc54:
Ok, will take a look - thx!

jjhursey:
We can address the LANL-disable-dlopen runtime issue in a separate PR since this PR is only uncovering an existing issue.
@jsquyres @hppritcha I think this PR is ready to go.

kawashima-fj:
@jjhursey @rhc54 @jsquyres @hppritcha Pulling this PR without fixing the singleton + dstore problem will cause LANL-disable-dlopen failure on all PRs for v2.x branch. Let's split the changes if we cannot resolve the problem soon.

1. Fix `--enable-pmix3-dstore` option
2. Fix singleton + dstore
3. Change the default

I reverted the default in the commit a few minutes ago to address only 1. Without 1, we cannot enable dstore of embedded PIMx through the Open MPI configure option.

I confirmed that the problem exists only in v2.x  branch.

jjhursey:
@kawashima-fj This looks good, and preserves the default disabled behavior for this feature. I agree with your plan noted in [this comment](https://github.com/open-mpi/ompi/pull/2859#issuecomment-276831906).

Now that PR #2909 is in, I think all that is left is a PR making dstore default enabled. Is that correct?

kawashima-fj:
@jjhursey Yes. Correct. I made a PR #2932 to enable it by default to complete our work. Thanks.

PR #2837, opened 2017-01-25
orterun: Add parameter to control when we give up on stack traces

jjhursey:
* MCA option to control how long we wait for stack traces:
   - `orte_timeout_for_stack_trace INTEGER`
     Default: 30
     Setting to <= 0 will cause it to wait forever
 * Useful when gathering stack traces from large jobs which might take
   a long time.

This is a custom patch for `v2.x` since this functionality takes a different form here than in `master`. I need to backport this to `master`.

This improves https://github.com/open-mpi/ompi-release/pull/1317

PR #2861, opened 2017-01-27
orterun: Add parameter to control when we give up on stack traces

jjhursey:
* MCA option to control how long we wait for stack traces:
   - `orte_timeout_for_stack_trace INTEGER`
     Default: 30
     Setting to <= 0 will cause it to wait forever
 * Useful when gathering stack traces from large jobs which might take
   a long time.
 * Back port of PR #2837 which required a custom patch


jjhursey:
@rhc54 PR #2837 and PR #2861 are the same feature. I just had to port them differently due to the code reorg on `master`

Issue #2843, opened 2017-01-26
master: "mpirun --host a,b" via ssh seems to be broken


jsquyres:
On master, when I try to `mpirun --host a,b hostname` (via ssh -- *outside* of a SLURM job), I get the following error (the error seems to be occurring in plm_spawn, so I ran with `plm_base_verbose` set to 100):

```
$ mpirun --host mpi001,mpi002 --mca plm_base_verbose 100 hostname
[savbu-usnic-a:16604] mca: base: components_register: registering framework plm components
[savbu-usnic-a:16604] mca: base: components_register: found loaded component slurm
[savbu-usnic-a:16604] mca: base: components_register: component slurm register function successful
[savbu-usnic-a:16604] mca: base: components_register: found loaded component rsh
[savbu-usnic-a:16604] mca: base: components_register: component rsh register function successful
[savbu-usnic-a:16604] mca: base: components_register: found loaded component isolated
[savbu-usnic-a:16604] mca: base: components_register: component isolated has no register or open function
[savbu-usnic-a:16604] mca: base: components_open: opening plm components
[savbu-usnic-a:16604] mca: base: components_open: found loaded component slurm
[savbu-usnic-a:16604] mca: base: components_open: component slurm open function successful
[savbu-usnic-a:16604] mca: base: components_open: found loaded component rsh
[savbu-usnic-a:16604] mca: base: components_open: component rsh open function successful
[savbu-usnic-a:16604] mca: base: components_open: found loaded component isolated
[savbu-usnic-a:16604] mca: base: components_open: component isolated open function successful
[savbu-usnic-a:16604] mca:base:select: Auto-selecting plm components
[savbu-usnic-a:16604] mca:base:select:(  plm) Querying component [slurm]
[savbu-usnic-a:16604] mca:base:select:(  plm) Querying component [rsh]
[savbu-usnic-a:16604] [[INVALID],INVALID] plm:rsh_lookup on agent ssh : rsh path NULL
[savbu-usnic-a:16604] mca:base:select:(  plm) Query of component [rsh] set priority to 10
[savbu-usnic-a:16604] mca:base:select:(  plm) Querying component [isolated]
[savbu-usnic-a:16604] mca:base:select:(  plm) Query of component [isolated] set priority to 0
[savbu-usnic-a:16604] mca:base:select:(  plm) Selected component [rsh]
[savbu-usnic-a:16604] mca: base: close: component slurm closed
[...]

PR #2864, opened 2017-01-27
Repair rsh/ssh tree spawn

rhc54:
Repair rsh/ssh tree spawn by unpacking and updating the nidmap in remote_spawn.

Add more specific error messages so the cause of a messaging problem is a little clearer. Remove some stale code. Ensure we stop trying to send a message after a few times.

### DIFF
--- a/orte/mca/errmgr/base/help-errmgr-base.txt
+++ b/orte/mca/errmgr/base/help-errmgr-base.txt
@@ -10,7 +10,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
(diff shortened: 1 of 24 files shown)


rhc54:
If the HNP is going to request topology info, it cannot do so via a routed OOB message as the intervening daemons may not be ready. So disable routing until the VM is ready, and have daemons start routing as they receive the xcast launch msg (which includes the data they need to talk to their peers).

Do a little optimization and minimize recomputation of the routing plan.

This completes repair of the master branch

rhc54:
Fixes #2843

rhc54:
@artpol84 this is the PR that needs to come into OMPI v4, if not already there.

jsquyres:
The 2 commits from this PR are already included in the v4.0.x branch.

artpol84:
@janjust, FYI

PR #2865, opened 2017-01-27
Fix configure RDTSCP check

Zzzoom:
RDTSCP support checking is currently being skipped due to a bug in the arch test.
The RDTSCP check itself, which doesn't compile because it clobbers the wrong registers, then fails due to an autoconf macro typo.

This PR plugs all 3. I haven't tested for issues in the RDTSCP codepath yet.

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1193,17 +1193,17 @@ AC_MSG_ERROR([Can not continue.])
 
     # Check for RDTSCP support
     result=0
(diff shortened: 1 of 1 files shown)


jsquyres:
Please sign off the commit (git commit --amend -s). 

@hjelmn @bosilca please have a look

hjelmn:
Hmm, thought I had already done this. Must not have made it upstream. This should probably go into 2.0.3.

Zzzoom:
Not needed anymore due to #2926, needs new PR to remove the check altogether.

PR #2874, opened 2017-01-29
yalla: fix memory leak with blocking non-contig send.

yosefe:
@miked-mellanox @alex-mikheev @jladd-mlnx


miked-mellanox:
+1
Please PR it into v1.x, v2,0 and v2.x

PR #2905, opened 2017-02-02
libevent/external: Add opal_event_include to this component

jjhursey:
* Adds a parameter to adjust the method used by libevent.
   - Matches that of the libevent2022 component.

Refs PR #2807


jjhursey:
@gpaulsen I wanted to make this match the internal component logic as closely as possible. Adding a note about the default value is a good idea. If someone does that they should update the internal and external modules with the same logic. But that will need to be a different PR.

Issue #2897, opened 2017-02-02
v2.x: MPI singleton + PMIx dstore fails


kawashima-fj:
@rhc54 As discussed in #2859, when I enable the PMIx dstore, An MPI process of singleton execution (launch directly; no mpiexec) fails with the following message on v2.x branch.

```
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):
  
  orte_ess_init failed
  --> Returned value Bad parameter (-5) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):
  
  ompi_mpi_init: ompi_rte_init failed
  --> Returned "Bad parameter" (-5) instead of "Success" (0)
--------------------------------------------------------------------------
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```

The problem will be in `fork_hnp` function of the singleton ESS. It checks the number of PMIx parameters. But the number varies if dstore is enabled. Probably `PMIX_DSTORE_ESH_BASE_PATH` is added.

https://github.com/open-mpi/ompi/blob/v2.x/orte/mca/ess/singleton/ess_singleton_module.c#L615

```
615             if (4 != opal_argv_count(argv)) {
(gdb) n
616                 opal_argv_free(argv);
(gdb) p cptr
$8 = 0x6533cc "PMIX_NAMESPACE=399310849,PMIX_RANK=0,PMIX_SERVER_URI=pmix-server:22133:/tmp/openmpi-sessions-1000@imtofu2_0/6093/pmix-22133,PMIX_SECURITY_MODE=native,PMIX_DSTORE_ESH_BASE_PATH=/tmp/openmpi-sessions-10"...
```

The master seems to have the solution. Probably a1e8e58a8a9a06d2e4297c4dc479c8b359ca8d05. Cherry-picking this commit is sufficient?

[...]

PR #2909, opened 2017-02-02
Fix singleton operations

rhc54:
Cherry-pick several commits from master to fix singleton operations. Includes some partial custom repair of singleton comm_spawn - still some problems that need to be resolved.

Refs #2897 

### DIFF
--- a/orte/mca/ess/singleton/ess_singleton_module.c
+++ b/orte/mca/ess/singleton/ess_singleton_module.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2011      Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2013-2015 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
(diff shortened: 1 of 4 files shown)


rhc54:
yeah, the rest of the problem with singleton comm_spawn is in the orte layer - i'll address it tomorrow

rhc54:
@hppritcha @jsquyres I would suggest going ahead with this PR and #2897 for now. With the changes in this PR, singletons will work just fine with dstore enabled or disabled. Singleton comm_spawn hangs, but that is a secondary problem and can be dealt with later. I don't have time right now, so I'd suggest someone file an issue for v2.1 and we'll deal with it when time is available.

jsquyres:
@hppritcha Signed-off-by checker failed because @rhc54 cherry-picked some master commits from before we required signed-off-by.  They're commits from @ggouaillardet, who a) is a proven community member, and b) has a signed contribution form on file for commits from before we required signed-off-by.

My $0.02: good to go.

hppritcha:
Note I have a pr in a similar signature sign off issue.   Do we give founding members who commit directly to master without sign offs a pass?   Note that isn't the case with this PR.

rhc54:
This isn't a question of getting a "pass". All we are doing is porting a commit already in our code base over to a release branch. There is no code being introduced into the repository, and so technically no sign-off is required.

New code requires a sign-off, regardless of who is committing it.

jsquyres:
👍  on @rhc54's answer.

PR #2911, opened 2017-02-02
For performance, try to send the oob/tcp message a few times before dropping back into the event library

rhc54:


rhc54:
@hjelmn Please take a quick look

PR #2914, opened 2017-02-02
swat some compiler warnings

hppritcha:
I got tired of seeing these compiler warnings while using clang on my Mac.
@hjelmn 

### DIFF
--- a/ompi/mca/coll/libnbc/nbc_ialltoallv.c
+++ b/ompi/mca/coll/libnbc/nbc_ialltoallv.c
@@ -7,7 +7,7 @@
  *                         rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2015-2017 Los Alamos National Security, LLC. All rights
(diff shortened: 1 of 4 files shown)


hjelmn:
I think we have an opal_attribute_unused. I think we are supposed to use that. @jsquyres Would know better.

ompi/mca/coll/libnbc/nbc_ialltoallv.c
```diff
@@ -214,6 +214,7 @@ int ompi_coll_libnbc_ialltoallv_inter (const void* sendbuf, const int *sendcount
   return OMPI_SUCCESS;
 }
 
+__attribute__((unused))
```
jsquyres: @hjelmn is right -- we have `__opal_attribute_unused__` as a portable form of `__attribute__((unused))`.  I think this was a bigger deal a while ago, before most (all?) compilers supported the `__attributed__((unused))` form.

...that being said, I guess I actually have no idea which compilers do/do not support `__attributed__((unused))`.  So it's probably safer to use `__opal_attribute_unused__` everywhere (which will be defined to be "" if the compiler doesn't support it).

hppritcha:
@jsquyres switched to using the ```_opal_attribute_unused__```

Issue #2920, opened 2017-02-03
master: pmix MPI spawn / list assertion error


jsquyres:
From @siegmargross post on users (https://www.mail-archive.com/users@lists.open-mpi.org/msg30564.html):

-----
 
I have installed openmpi-master-201702010209-6cb484a on my "SUSE Linux Enterprise Server 12.2 (x86_64)" with Sun C 5.14 and gcc-6.3.0. Unfortunately, I get errors when I run my spawn programs.

```
loki spawn 107 mpiexec -np 1 --host loki,loki,nfs1 spawn_intra_comm
Parent process 0: I create 2 slave processes
[nfs1:27716] PMIX ERROR: ERROR in file ../../../../../../../openmpi-master-201702010209-6cb484a/opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c at line 1029
[nfs1:27716] PMIX ERROR: ERROR in file ../../../../../../../openmpi-master-201702010209-6cb484a/opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_get.c at line 501
--------------------------------------------------------------------------
At least one pair of MPI processes are unable to reach each other for
MPI communications.  This means that no Open MPI device has indicated
that it can be used to communicate between these processes.  This is
an error; Open MPI requires that all MPI processes be able to reach
each other.  This error can sometimes be the result of forgetting to
specify the "self" BTL.

 Process 1 ([[42193,2],1]) is on host: nfs1
 Process 2 ([[42193,1],0]) is on host: unknown!
 BTLs attempted: self tcp

Your MPI job is now going to abort; sorry.
--------------------------------------------------------------------------
[nfs1:27727] [[42193,2],1] ORTE_ERROR_LOG: Unreachable in file ../../openmpi-master-201702010209-6cb484a/ompi/dpm/dpm.c at line 426
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):

 ompi_dpm_dyn_init() failed
 --> Returned "Unreachable" (-12) instead of "Success" (0)
--------------------------------------------------------------------------
[nfs1:27727] *** An error occurred in MPI_Init
[nfs1:27727] *** reported by process [2765160450,1]
[...]

PR #2925, opened 2017-02-04
orte: Fix MPI_Spawn

artpol84:
Register namespace even if there is no node-local processes that
belongs to it. We need this for the MPI_Spawn case.

Addressing https://github.com/open-mpi/ompi/issues/2920.
Was introduced in be3ef777392347aa4560fb4eaa13075d2e77ed6e.

### DIFF
--- a/orte/orted/pmix/pmix_server_register_fns.c
+++ b/orte/orted/pmix/pmix_server_register_fns.c
@@ -74,11 +74,6 @@ int orte_pmix_server_register_nspace(orte_job_t *jdata)
                         ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
                         ORTE_JOBID_PRINT(jdata->jobid));
 
-    /* if this job has no local procs, then no need to register them */
-    if (0 == jdata->num_local_procs) {
-        return ORTE_SUCCESS;
-    }
-
     /* setup the info list */
     info = OBJ_NEW(opal_list_t);
     uid = geteuid();


artpol84:
v2.0.x and v2.x are intact.

artpol84:
@siegmargross @jsquyres

artpol84:
```
$ ./mpirun_my.sh 1 ./spawn  
Parent process 0: I create 3 slave processes



Child process 0 running on cn1
    MPI_COMM_WORLD ntasks:              3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        1

Child process 2 running on cn1
    MPI_COMM_WORLD ntasks:              3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        3
Parent process 0 running on cn1
    MPI_COMM_WORLD ntasks:              1
    COMM_CHILD_PROCESSES ntasks_local:  1
    COMM_CHILD_PROCESSES ntasks_remote: 3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        0
Child process 1 running on cn2
    MPI_COMM_WORLD ntasks:              3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        2
```

artpol84:
Note, I sometimes see the following error at once app is finished:
```
$ ./mpirun_my.sh 1 ./spawn  
Parent process 0: I create 3 slave processes

Child process 0 running on cn1
    MPI_COMM_WORLD ntasks:              3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        1


Child process 1 running on cn2
    MPI_COMM_WORLD ntasks:              3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        2

Child process 2 running on cn1
Parent process 0 running on cn1
    MPI_COMM_WORLD ntasks:              1
    MPI_COMM_WORLD ntasks:              3
    COMM_CHILD_PROCESSES ntasks_local:  1
    COMM_CHILD_PROCESSES ntasks_remote: 3
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        0
    COMM_ALL_PROCESSES ntasks:          4
    mytid in COMM_ALL_PROCESSES:        3
mpirun noticed that process rank 0 with PID 2086 on node cn1 exited on signal 13 (Broken pipe).
```

I'm running a cluster of linux containers on my laptop and in my environment delays are more visible than on the real cluster. So this may be some sort of a race condition appearing.

artpol84:
@jsquyres I'm not sure that Ralph will have the time to review. So it's up to you who should review instead.

jjhursey:
I think this looks good. I'm ok with merging it in and letting MTT have a try this evening.

artpol84:
Ok, thanks!

rhc54:
I'm not convinced this change is what we really want - the "connect" code is supposed to register any missing nspaces, so it sounds instead like we aren't correctly waiting for connect to complete. If we look at the error being reported on #2920, it is when dstore is attempting to store modex data. There are no comments in the code, but it _appears_ as though it is trying to do this prior to ORTE having registered the nspace.

Arbitrarily registering every nspace even when there are no local procs causes problem for the DVM users as it consumes space on nodes that don't have any involvement in a job. We should instead fix the real source of the problem.

artpol84:
Dstore was failing trying to store jobinfo for the namespace that wasn't registered. Registering namespace solved the problem. So I believe that dstore is working correctly.
As far as I recall I got to [this place](https://github.com/open-mpi/ompi/pull/2925/files#diff-1491da77fa56c7a6f9c42f66ae199eacL77) from connect code path. There is no way to distinguish between the originator of namespace register call.
So I guess that the right solution is to not to call register namespace  if it is not desirable instead of blindly decide in `orte_pmix_server_register_nspace`

rhc54:
Sorry if I wasn't clear - I agree that dstore is working correctly. My point was that we aren't calling the functions in the correct order. I'll take a look at where you left off.

Thanks!

artpol84:
thanks,
to clarify as well: I'm suggesting to decide whether to register namespace or not on the higher level that has the information about what is going on rather than on the pmix level that should just do what it's being said.
Or as another option to have an additional argument for the `orte_pmix_server_register_nspace` that identifies if this is mandatory to register this namespace or not.
As I noted this function was called from the connect code so there was no race condition that caused namespace to be not yet registered. Registration was just immediately "aborted" because of removed statement.

rhc54:
got it - thanks! Will fix

PR #2926, opened 2017-02-04
Improve x86-64 timer performance

Zzzoom:
WIP

On x86-64 a significant amount of processor time is currently being spent on ```opal_timer_base_get_usec_sys_timer```, mainly used by ```opal_progress```. The function calls ```opal_sys_timer_get_cycles``` to get a cycle count and divides it by a frequency.
In asm this looks like either ```cpuid; rdtsc; salq; orq; divq``` or (with RDTSCP support) ```rdtscp; cpuid; salq; orq; divq```. The problem is that both ```cpuid``` and ```divq``` are expensive.

This PR tackles both:
- ```cpuid``` is replaced by the cheaper ```lfence``` for serialization, like ```rdtsc_ordered``` does in the linux kernel. This is safe on Intel processors (http://www.intel.com/content/www/us/en/architecture-and-technology/64-ia-32-architectures-software-developer-vol-3a-part-1-manual.html page 8-17), but apparently not on AMD, which requires ```mfence``` instead. I'm not sure of where in openmpi it would be best to perform runtime cpu detection and setting up the dispatch.
- ```divq``` is avoided, using cycles directly instead of usecs in ```opal_progress``` if possible. This is done by flipping the timer priority from ```USEC_NATIVE > CYCLES_NATIVE > fallback cycles``` to ```CYCLES_NATIVE > USEC_NATIVE > fallback cycles```.

I haven't looked for a good microbenchmark that isolates the effect, but I've been getting decent time improvements on NPB (mainly FT) and Quantum Espresso.

### DIFF
--- a/opal/include/opal/sys/amd64/timer.h
+++ b/opal/include/opal/sys/amd64/timer.h
@@ -31,31 +31,14 @@ typedef uint64_t opal_timer_t;
 
 #if OPAL_GCC_INLINE_ASSEMBLY
 
-/**
- * http://www.intel.com/content/www/us/en/intelligent-systems/embedded-systems-training/ia-32-ia-64-benchmark-code-execution-paper.html
- */
+/* TODO: add AMD mfence version and dispatch at init */
 static inline opal_timer_t
 opal_sys_timer_get_cycles(void)
 {
-     unsigned l, h;
-#if !OPAL_ASSEMBLY_SUPPORTS_RDTSCP
-     __asm__ __volatile__ ("cpuid\n\t"
+     uint32_t l, h;
+     __asm__ __volatile__ ("lfence\n\t"
                            "rdtsc\n\t"
-                           : "=a" (l), "=d" (h)
-                           :: "rbx", "rcx");
-#else
-     /* If we need higher accuracy we should implement the algorithm proposed
-      * on the Intel document referenced above. However, in the context of MPI
-      * this function will be used as the backend for MPI_Wtime and as such
-      * can afford a small inaccuracy.
-      */
-     __asm__ __volatile__ ("rdtscp\n\t"
-                           "mov %%edx, %0\n\t"
-                           "mov %%eax, %1\n\t"
-                           "cpuid\n\t"
-                           : "=r" (h), "=r" (l)
-                           :: "rax", "rbx", "rcx", "rdx");
-#endif
+                           : "=a" (l), "=d" (h));
      return ((opal_timer_t)l) | (((opal_timer_t)h) << 32);
 }
 
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -37,6 +37,7 @@
 #include "opal/runtime/opal_params.h"
 
 #define OPAL_PROGRESS_USE_TIMERS (OPAL_TIMER_CYCLE_SUPPORTED || OPAL_TIMER_USEC_SUPPORTED)
+#define OPAL_PROGRESS_ONLY_USEC_NATIVE (OPAL_TIMER_USEC_NATIVE && !OPAL_TIMER_CYCLE_NATIVE)
 
 #if OPAL_ENABLE_DEBUG
 bool opal_progress_debug = false;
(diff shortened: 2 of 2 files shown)


ibm-ompi:

 Gist: https://gist.github.com/3926f5191428aa85815430a5681ff170

ibm-ompi:

 Gist: https://gist.github.com/75df96ec0449e2d2111580030c743e4c

ibm-ompi:

 Gist: https://gist.github.com/28520303aa33907b5d3d4ca299d379c5

Zzzoom:
Who maintains @ibm-ompi? Those logs aren't very useful.

jjhursey:
@Zzzoom I maintain @ibm-ompi - It looks like our CI got stuck. You can re-run it by adding a comment:
bot:ibm:retest

ibm-ompi:

 Gist: https://gist.github.com/7d688cb9be5106ff9adeaec6ca6d35f0

ibm-ompi:

 Gist: https://gist.github.com/5200ba1d829c8bf888e232516546b2b2

ibm-ompi:

 Gist: https://gist.github.com/b658eba46c3fa76386c7c7eb1f94bca7

jjhursey:
The IBM CI is hitting a local file system issue that I need to resolve with the sysadmins. I'm turning it off for now. Disregard the CI failures for now. I'll retest when it comes back online.

hjelmn:
:+1: Thanks for doing this.

opal/include/opal/sys/amd64/timer.h
```diff
@@ -31,31 +31,14 @@ typedef uint64_t opal_timer_t;
...
-      * can afford a small inaccuracy.
-      */
-     __asm__ __volatile__ ("rdtscp\n\t"
-                           "mov %%edx, %0\n\t"
-                           "mov %%eax, %1\n\t"
-                           "cpuid\n\t"
-                           : "=r" (h), "=r" (l)
-                           :: "rax", "rbx", "rcx", "rdx");
-#endif
+                           : "=a" (l), "=d" (h));
```
hjelmn: One question. Why remove the rdtscp path? Is it slower than the rdtsc instruction?

hjelmn:
:bot:mellanox:retest

hjelmn:
bot:ibm:retest

hjelmn:
@Zzzoom I don't even know if we care about instruction ordering with the rdtsc instruction. All we need for MPI_Wtime () is monotonic. For opal_progress all we care about is that the event library fires at some point.

opal/include/opal/sys/amd64/timer.h
Zzzoom: According to http://agner.org/optimize/instruction_tables.pdf rdtscp has higher latency than lfence+rdtsc and it has similar or worse ordering guarantees.

Zzzoom:
@hjelmn people use MPI_Wtime() to time code so I didn't want to introduce a regression by making it less accurate in some weird case where the processor is particularly crafty when reordering instructions.
```opal_progress``` could use a lower precision timer that only does ```rdtsc```, but the 4 cycle latency penalty in doing ```lfence``` probably isn't worth the refactoring.

opal/include/opal/sys/amd64/timer.h
hjelmn: Ok. Thanks! I plan to merge this soon and PR it to 2.0.x and 2.1.x.

hjelmn:
bot:ibm:retest

opal/include/opal/sys/amd64/timer.h
Zzzoom: On AMD processors ```mfence``` should be used instead of ```lfence```, but I don't know where in opal I should detect the processor and change the function pointer.

bosilca: Such an approach will force us to lose the benefit of the inlined function. I like how the Linux kernel solves this issue, but it might be a little too intrusive to bring all the mechanism into OMPI.

What happens on AMD processors if we use lfence instead of mfence ? Would the timer stop being monotonic ?

Zzzoom: @bosilca no, it will have more variance.

Zzzoom: @bosilca I just took another look at the code and ```opal_sys_timer_get_cycles``` only gets inlined into ```opal_timer_base_get_usec_sys_timer``` and ```opal_timer_base_get_cycles_sys_timer```. Then these two functions are called through the ```opal_timer_base_get_usec``` and ```opal_timer_base_get_cycles``` pointers, so there might be some gains to be had if we got rid of all the indirection and allowed inlining.

Issue #1666, opened 2016-05-11
MPI_Accumulate Error when using  datatype "MPI_SHORT_INT"


Zhiming-Wang:
Hi, all

There is a MPI_Accumulate error at Open MPI when using  datatype "MPI_SHORT_INT".

I thought that it was about the compress/uncompress of predefined datatype "MPI_SHORT_INT" at osc module.

The datatype "MPI_SHORT_INT" used at this case, its size of storage space is 8 byte (4-byte aligned),  but its payload size is 6 byte (2 byte for short, and 4 byte for int.). When updating data from remote ranks, it  was compressed (6 byte for each), but uncompress was not done when reading data, 8 byte was get for each element, so generated an error.
Please confirm and help to fix it. Thanks very much.

Source code of my test case:

```[C]
#include <stdio.h>
#include "mpi.h"

#define LEN 100
#define  SHIFT -1

int main( int argc, char **argv ) {

    int my_rank, num_tasks;
    int i, j;
    MPI_Win win;
    MPI_Group all_group, origin_group, target_group;
    int root = 0;

    typedef struct {
        short val;
        short pad;
        int rank;
    } short_int_t;

    short_int_t Win[LEN], Array[LEN];

    MPI_Init( &argc, &argv );
    MPI_Comm_rank( MPI_COMM_WORLD, &my_rank );
    MPI_Comm_size( MPI_COMM_WORLD, &num_tasks );
    if (2 != num_tasks) {
        printf("There should be 2 ranks at this case. \n");
        MPI_Abort(MPI_COMM_WORLD, 1);
    }



    MPI_Win_create( Win, sizeof(Win), sizeof(short_int_t),
                    MPI_INFO_NULL, MPI_COMM_WORLD, &win );

    MPI_Win_fence( 0, win );
    MPI_Comm_group( MPI_COMM_WORLD, &all_group );
    MPI_Group_incl( all_group, 1, &root, &target_group );
    MPI_Group_difference( all_group, target_group, &origin_group );

    /* initialize all arrays for MPI_MINLOC*/
    for( i=0; i<LEN; i++ ) {
        if (my_rank == i%2) {
            Array[i].val = Win[i].val = SHIFT;
        }
        else {
            Array[i].val = Win[i].val = -1*SHIFT;
        }
        Array[i].rank = my_rank;
        Win[i].rank = my_rank;
        Win[i].pad = -i;
    }

    if( my_rank == root ) {

[...]

PR #2832, opened 2017-01-25
osc/base: Detect unsupported data types and abort

jjhursey:
Using MPI_MINLOC or MPI_MAXLOC with the following data types
leads to data corruption:
 * MPI_DOUBLE_INT
 * MPI_LONG_INT
 * MPI_SHORT_INT
 * MPI_LONG_DOUBLE_INT

Detect this print a error message and abort.
This workaround should be removed once the following issue is resolved:
 * https://github.com/open-mpi/ompi/issues/1666

PR #2927, opened 2017-02-06
osc/base: Detect unsupported data types and abort

jjhursey:
Using MPI_MINLOC or MPI_MAXLOC with the following data types
leads to data corruption:
 * MPI_DOUBLE_INT
 * MPI_LONG_INT
 * MPI_SHORT_INT
 * MPI_LONG_DOUBLE_INT

Detect this print a error message and abort.
This workaround should be removed once the following issue is resolved:
 * https://github.com/open-mpi/ompi/issues/1666

(cherry picked from commit 94f92f6b49212545355123d711abc91af9647d8e)

### DIFF
--- a/ompi/mca/osc/base/Makefile.am
+++ b/ompi/mca/osc/base/Makefile.am
@@ -7,12 +7,15 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
+# Copyright (c) 2017 IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
 #
 # $HEADER$
 #
+
+dist_ompidata_DATA = base/help-mca-osc-base.txt
 
 headers += \
         base/base.h \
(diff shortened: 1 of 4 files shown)


jjhursey:
Refs PR #2832

hjelmn:
These datatypes and ops need to die :-/.

nysal:
@jjhursey I think this patch will only work for osc/pt2pt. osc/rdma will probably need a similar check in ompi_osc_rdma_rget_accumulate_internal()

hjelmn:
osc/pt2pt *should* be trivial to update to support for minloc/maxloc. It is a slight modification to ```osc_pt2pt_accumulate_buffer```. The hetero #ifdef needs to be removed  and the if needs to check if the datatype is non-contiguous. I have no interest in making either operation work with osc/rdma (outside just treating MPI_SHORT_INT as contiguous) so I welcome a check that makes it error.

jjhursey:
@nysal I updated the PR to include a commit for `ompi_osc_rdma_rget_accumulate_internal()` - can you and/or @hjelmn confirm that this is the correct/best placement for that logic in the `osc/rdma` component?

jsquyres:
@hppritcha Good to go

PR #2932, opened 2017-02-07
pmix/pmix112: Enable PMIx shared memory dstore by default

kawashima-fj:
The problem of PMIx shared memory dstore + MPI singleton is fixed now. See #2859 and #2909. We need PMIx shared memory dstore in v2.1.x.

The master branch had this change long ago and the commit includes many changes so this commit is a direct commit (no cherry-pick).

@jjhursey Please review.

### DIFF
--- a/opal/mca/pmix/pmix112/configure.m4
+++ b/opal/mca/pmix/pmix112/configure.m4
@@ -41,9 +41,9 @@ AC_DEFUN([MCA_opal_pmix_pmix112_CONFIG],[
 
     AC_ARG_ENABLE([pmix-dstore],
                   [AC_HELP_STRING([--enable-pmix-dstore],
-                                  [Enable PMIx shared memory data store (default: disabled)])])
+                                  [Enable PMIx shared memory data store (default: enabled)])])
     AC_MSG_CHECKING([if PMIx shared memory data store is enabled])
-    if test "$enable_pmix_dstore" = "yes"; then
+    if test "$enable_pmix_dstore" != "no"; then
         AC_MSG_RESULT([yes])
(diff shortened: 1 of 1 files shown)


jjhursey:
Thanks for this PR. This is good to go.

However, we should have a discussion in the OMPI Teleconf about if we should have this enabled by default in the 2.1.x series. I cannot remember if we decided on this yet or not. @gpaulsen Can you add this to the agenda for tomorrow.

kawashima-fj:
@jjhursey @gpaulsen OK. I added the WIP-DNM label. Please remove it if you confirmed it.

hppritcha:
LANL CI down for today due to NERSC quarterly maintenance.

jsquyres:
Per 2017-02-07 webex discussion, there appears to be a bug in dstore on master.  When that gets fixed and is brought over to v2.x, it seems like we want dstore to be the default on v2.x.

hppritcha:
@hjelmn has reason to believe there is a memory corruption problem if shared memory is enabled by default.

jjhursey:
This is the PMIx issue that @hjelmn @artpol84 are working on.
 * https://github.com/pmix/master/issues/280

Once that is fixed in PMIx, then I'll update the `v2.x` branch with updated PMIx `v1.2.1rc`. We'll need those fixes in before merging this PR - so we don't break default builds.

kawashima-fj:
bot:mellanox:retest

hppritcha:
merge this after #2958

jsquyres:
#2958 has been merged.  Now merging this one to enable PMIx shared memory datastore by default.

PR #2934, opened 2017-02-07
Move coll structure outside of the communicator

jjhursey:
As we changed the ABI (forcing a major release), we can limit
the size of the predefined communicators by moving the collective
structure outside the communicator. This might have a minimal,
but unnoticeable, impact on performance. This approach has been
discussed during the January 2017 devel meeting.

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2013 The University of Tennessee and The University
+ * Copyright (c) 2004-2017 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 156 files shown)


bosilca:
It won't be fait to review pieces of my own code. I'll ask @ggouaillardet if he has some time to review.

ggouaillardet:
@bosilca the code looks fine to me

that being said, per the commit message, the purpose of this PR is to reduce the size of the predefined communicators.
from a pragmatic point of view, we now save a bit less than 1k per MPI task since we do not need to allocate ```c_coll``` for ```MPI_COMM_NULL```, at the expense of adding one indirection each time a collective is invoked.
as you pointed out, the cost of this extra indirection is very likely a minimal but unnoticeable impact of performance, and i do not plan to argue with that.

bottom line, and unless i miss the full picture, that looks like a massive and acceptable change for a small benefit.
did i miss the point ?

or is there a master plan behind that ?
for example, ```c_coll``` could be made an object so it can be shared and hence improve communicator duplication performance.

bosilca:
@ggouaillardet thanks for the review. I would not call this a master plan but here are few of the options and benefits that we have evoked so far:
 * remove the ABI incompatibility from the ompi_communicator_t. Every time we amend the ompi_communicator_t structure (because of collective or other reasons) we breaks the ABI with previous versions. Moving the coll structure outside gives us some slack on the size of the predefined communicator structure.
 * we can now lazily allocate the coll structure on the first collective, decreasing the time to build a communicator.
 * we can now share the collective structures between duplicated communicators (as you mentionned).
 * in the long[er] term we could split this structure in 2 (function pointers and then module info), so that we can save the space for the function pointers among communicators with identical collective support.

All in all, a some small benefits, with potentially small impact on the performance or memory consumption of the OMPI library in mainstream scenarios.

jjhursey:
Now that PR #2808 has been merged, I updated this PR to be based off of `master`. It should be ready for review now.

jjhursey:
@ggouaillardet Can you review for us? I'd like to get this into `v3.0` branch which _might_ be happening in the next week or so.

PR #2954, opened 2017-02-10
hwloc: rename the hwloc framework into something else

ggouaillardet:
Refs. open-mpi/ompi#2616


ibm-ompi:

 Gist: https://gist.github.com/e3dc606df6bb24606487f4e95e99060e

jsquyres:
I suggest that we might not want to rename anything with the string "xxx" in it.  😳

ggouaillardet:
@jsquyres i cheched　```hwlocxxx``` is not used anywhere, so it will be painless to do a ```sed``` and a few ```git mv``` once we decide the final name for this framework

rhc54:
Please - let's not do this. You are swatting a mosquito with a sledgehammer. There are much simpler solutions to this problem. Will address on the issue.

jsquyres:
We went with #2955 instead of this one.

PR #2957, opened 2017-02-10
plm/rsh: Fix signal handling for rsh launcher

jjhursey:
* Similar to the other launchers (i.e., slurm, alps) we need to put the
   children in a separate process group to prevent SIGINT (from a CTRL-C)
   from being delivered to the whole process group and prematurely
   killing the rsh/ssh connections to the remote daemons.

### DIFF
--- a/orte/mca/plm/rsh/plm_rsh_module.c
+++ b/orte/mca/plm/rsh/plm_rsh_module.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2007-2012 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2008-2009 Sun Microsystems, Inc.  All rights reserved.
- * Copyright (c) 2011      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2011-2017 IBM Corporation.  All rights reserved.
  * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


rhc54:
@jjhursey Looks like this failed on Mellanox tester. It also doesn't look correct to me - only the child should be calling setpid - per orte/mca/odls/default:

```c
#if HAVE_SETPGID
    /* Set a new process group for this child, so that any
     * signals we send to it will reach any children it spawns */
    setpgid(0, 0);
#endif
```

rhc54:
Hmmm...I see the comment about what Stevens recommends in the alps and slurm launchers, and those aren't failing, so perhaps my comment is incorrect. Regardless, the Mellanox details are gone as time has expired on them. Let's see if a retest shows anything.

bot:mellanox:retest

jjhursey:
@rhc54 Yeah I thought the `setpgid` in the parent was redundant since the child is doing that already. But then I saw the comment in the [plm/slurm](https://github.com/open-mpi/ompi/blob/master/orte/mca/plm/slurm/plm_slurm_module.c#L702-L705) and [plm/alps](https://github.com/open-mpi/ompi/blob/master/orte/mca/plm/alps/plm_alps_module.c#L684-L687) so I followed the same pattern.

I might update this PR with an error check on the return code of `setpgid`. So I'll add the `HAVE_SETPGID` when I push that update. I'll mark this to hold until then.

rhc54:
@jjhursey looks like it was a momentary thing as it now passes - feel free to merge after adding the tests. Thx!

PR #2964, opened 2017-02-14
Protect the embedded libraries when updating copyrights

rhc54:
we shouldn't be overwriting their copyrights with our own

### DIFF
--- a/contrib/update-my-copyright.pl
+++ b/contrib/update-my-copyright.pl
@@ -68,7 +68,12 @@ my $my_search_name = "Cisco";
 my $my_formal_name = "Cisco Systems, Inc.  All rights reserved.";
 
 # Protected directories
-my @protected = qw(pmi2x/pmix/ hwloc1113/hwloc/ libevent2022/libevent/);
+my @protected = qw(
+    opal\\/mca\\/pmi\\/pmix.+?\\/pmix\\/
+    opal\\/mca\\/hwloc\\/hwloc.+?\\/hwloc\\/
+    opal\\/mca\\/libevent\\/libevent.+?\\/libevent\\/
+    contrib\\/update-my-copyright.pl
+);
 
 # Override the defaults if some values are set in the environment
 $my_search_name = $ENV{OMPI_COPYRIGHT_SEARCH_NAME}
@@ -155,7 +160,7 @@ foreach my $f (@files) {
     # the copyright
     my $ignore = 0;
     foreach my $p (@protected) {
-        if (index($f, $p) != -1) {
+        if (eval("\$f =~ /$p/")) {
             quiet_print "Ignoring protected file $f\n";
             $ignore = 1;
             last;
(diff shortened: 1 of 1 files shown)


jsquyres:
@rhc54 Are we *overwriting* their copyrights, or adding our own?

I guess there's two cases when we modify any of these embedded projects:

1. If the upstream project has a copyright by org X, and then X modifies an embedded copy of a file in OMPI.  Do we modify the original/upstream copyright?  Or should we add another copyright statement for org X that is separate from the original/upstream copyright?
    * I'm not sure what the Right answer is here.
1. the upstream project has no copyright on the embedded via (via the upstream release).  We should probably *add* a copyright for org X, and not disturb any of the original/upstream copyrights.
    * Are we currently disturbing the original/upstream copyrights in this case?

rhc54:
I'm afraid it is both cases - if the org is the same, then we "update" the copyrights. If the org is different, then we "add" a copyright.

Either way is wrong, IMO, as we are not modifying their files. We are simply updating our embedded copy to track their changes.

contrib/update-my-copyright.pl
```diff
@@ -67,6 +67,9 @@
 my $my_search_name = "Cisco";
 my $my_formal_name = "Cisco Systems, Inc.  All rights reserved.";
 
+# Protected directories
+my @protected = qw(pmi2x/pmix/ hwloc1113/hwloc/ libevent2022/libevent/);
```
jsquyres: Should we make these regexps to make them a bit more future-proof?  Otherwise, we'll need to remember to edit this script whenever we update these components.

Eg:

```diff
diff --git a/contrib/update-my-copyright.pl b/contrib/update-my-copyright.pl
index bed0f0a..5b068e0 100755
--- a/contrib/update-my-copyright.pl
+++ b/contrib/update-my-copyright.pl
@@ -68,7 +68,12 @@ my $my_search_name = "Cisco";
 my $my_formal_name = "Cisco Systems, Inc.  All rights reserved.";
 
 # Protected directories
-my @protected = qw(pmi2x/pmix/ hwloc1113/hwloc/ libevent2022/libevent/);
+my @protected = qw(
+    opal\\/mca\\/pmi\\/pmix.+?\\/pmix\\/
+    opal\\/mca\\/hwloc\\/hwloc.+?\\/hwloc\\/
+    opal\\/mca\\/libevent\\/libevent.+?\\/libevent\\/
+    contrib\\/update-my-copyright.pl
+);
 
 # Override the defaults if some values are set in the environment
 $my_search_name = $ENV{OMPI_COPYRIGHT_SEARCH_NAME}
@@ -155,7 +160,7 @@ foreach my $f (@files) {
     # the copyright
     my $ignore = 0;
     foreach my $p (@protected) {
-        if (index($f, $p) != -1) {
+        if (eval("\$f =~ /$p/")) {
             quiet_print "Ignoring protected file $f\n";
             $ignore = 1;
             last;
```

rhc54: sure - I'm just not enough of a perl expert to know how to do it. I'll happily take your diff.

jsquyres: Yeah, it turned out a little messier than I had hoped, but it'll do the job...

PR #2968, opened 2017-02-14
pmix/cray: performance improvements and cleanup

hjelmn:

 - Set the orte_bound_at_launch MCA variable. This resolves a launch
   performance bug when using aprun to launch Open MPI processes. If
   this variable is not set it can take minutes longer to launch with
   high ppn.

 - Do not use opal_output_verbose inside O(n) loops. This was causing
   us to make O(n) calls to snprintf which was greatly slowing launch
   at scale.

### DIFF
--- a/opal/mca/pmix/cray/pmix_cray.c
+++ b/opal/mca/pmix/cray/pmix_cray.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2007      The Trustees of Indiana University.
  *                         All rights reserved.
  * Copyright (c) 2011-2016 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2011-2016 Los Alamos National Security, LLC. All
+ * Copyright (c) 2011-2017 Los Alamos National Security, LLC. All
  *                         rights reserved.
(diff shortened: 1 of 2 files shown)


opal/mca/pmix/cray/pmix_cray.c
```diff
@@ -253,6 +253,7 @@ static void cray_get_more_info(void)
     opal_setenv("OMPI_NUM_APP_CTX", pbuf, true, &environ);
     opal_setenv("OMPI_FIRST_RANKS", firstrankstring, true, &environ);
     opal_setenv("OMPI_APP_CTX_NUM_PROCS", npstring, true, &environ);
+    opal_setenv(OPAL_MCA_PREFIX "orte_bound_at_launch", "true", true, &environ);
```
rhc54: On master, would you please move this to the orte/mca/schizo/alps component? I know it has to go here for v2.1, but it really doesn't belong here. Thanks!

hjelmn: Ok. Will do that now.

hppritcha:
going ahead and merging this in so we can test in MTT.

PR #2969, opened 2017-02-14
mca/base: fix deprecated variable help message

hjelmn:
Actually print out the original variable name.

(cherry picked from commit 33676c996082dc3c46fc9aae6f629df7063e6fec)

### DIFF
--- a/opal/mca/base/mca_base_var.c
+++ b/opal/mca/base/mca_base_var.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2012-2015 Los Alamos National Security, LLC. All rights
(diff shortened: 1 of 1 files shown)


rhc54:
@hjelmn Looks like the variable ```original``` doesn't exist in that function in 1.10.

hjelmn:
Ah yes. Let me look into it.

hjelmn:
Hmm, would have to bring over more to get this commit to work as is. Will have to backport the patch.

rhc54:
@hjelmn How critical is this for v1.10.6? I'd like to go ahead and release it since it has unintentionally been sitting here for two weeks. Shall we just push this to 1.10.7?

hjelmn:
Very low priority. I need to find the time to make a custom patch for 1.10

rhc54:
okay, let's just set this aside for now. I'll release 1.10.6, and if/when we decide to do a 1.10.7, then we can take the time to look at a custom patch - no need to worry about it for now

PR #2885, opened 2017-01-31
v2.x: pmix: fix PMIx_server_register_nspace() usage

ggouaillardet:

PR #2981, opened 2017-02-15
pmix/external: Fix memory corruption

karasevb:
This bug introduced in PR #2885. The status variable of waiting was not
activated it leads to early release of the callback data and memory
corruption.


ggouaillardet:
@karasevb thanks for catching this !
indeed, this was missing when i backported open-mpi/ompi@dccb1899e6622af38e4edc6de021ad670e02f7ff into open-mpi/ompi@1bf323751a085486bd0bcf078a8cce3b0dc579d8 from #2885 
i will take care of that

PR #2991, opened 2017-02-15
orte/errmgr: Improve help message on connection lost

jjhursey:
Clarify in this error message that it's the HNP located on node X that lost a connection to a remote daemon on node Y.

### DIFF
--- a/orte/mca/errmgr/base/help-errmgr-base.txt
+++ b/orte/mca/errmgr/base/help-errmgr-base.txt
@@ -11,6 +11,7 @@
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
 # Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
+# Copyright (c) 2017      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -61,9 +62,10 @@ route found between them. Please check network connectivity
 (including firewalls and network routing requirements).
 #
 [node-died]
-ORTE has lost communication with its daemon located on node:
+ORTE has lost communication with a remote daemon.
 
-  hostname:  %s
+  HNP daemon   : %s on node %s
+  Remote daemon: %s on node %s
 
 This is usually due to either a failure of the TCP network
 connection to the node, or possibly an internal failure of
--- a/orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
+++ b/orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
@@ -10,6 +10,7 @@
  * Copyright (c) 2011-2013 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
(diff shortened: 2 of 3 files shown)


jjhursey:
I've added the `WIP-DNM` flag. I'm debating on making some of the `OPAL_OUTPUT_VERBOSE` in this code path `opal_output_verbose` so that folks with a non-debug build can easily see these messages without having to rebuild.

@rhc54 What do you think? It's an error path just before the HNP calls abort, so I don't think it's a performance critical issue.

orte/mca/errmgr/base/help-errmgr-base.txt
```diff
@@ -61,9 +62,10 @@ route found between them. Please check network connectivity
 (including firewalls and network routing requirements).
 #
 [node-died]
-ORTE has lost communication with its daemon located on node:
+ORTE has lost communication with a remote daemon.
```
jsquyres: Is it useful (or even possible?) to show the error message?  I.e., can we see any kind of error code indicating *why* the remote end disconnected?

jjhursey: That information is not accessible at this level. To see more information you can pass: `-mca oob_base_verbose 7` which will show messages from the `oob/tcp` component (for example) during the connection lost path.
 1. [mca_oob_tcp_peer_close](https://github.com/open-mpi/ompi/blob/master/orte/mca/oob/tcp/oob_tcp_connection.c#L1045) - multiple way to get here, but once we call it then we lose the fine grained context. The verbose output should help pinpoint where this started from.
 1. [mca_oob_tcp_component_lost_connection](https://github.com/open-mpi/ompi/blob/master/orte/mca/oob/tcp/oob_tcp_component.c#L1119)
 1. [proc_errors](https://github.com/open-mpi/ompi/blob/master/orte/mca/errmgr/default_hnp/errmgr_default_hnp.c#L288) then reports the issue and aborts the job.

jjhursey:
Example output (here I'm manually `kill -9`'ing the `orted` on `mpi12`):
```
shell on mpi10$ mpirun -np 2 -host mpi10,mpi12 -mca oob_base_verbose 7 sleep 120
...
[mpi10:58379] [[4209,0],0]:tcp:recv:handler called for peer [[4209,0],1]
[mpi10:58379] [[4209,0],0]:tcp:recv:handler CONNECTED
[mpi10:58379] [[4209,0],0]:tcp:recv:handler allocate new recv msg
[mpi10:58379] [[4209,0],0]:tcp:recv:handler read hdr
[mpi10:58379] [[4209,0],0]-[[4209,0],1] mca_oob_tcp_msg_recv: peer closed connection
[mpi10:58379] [[4209,0],0] tcp_peer_close for [[4209,0],1] sd 48 state CONNECTED
[mpi10:58379] [[4209,0],0] tcp:lost connection called for peer [[4209,0],1]

  HNP daemon   : [[4209,0],0] on node mpi10
  Remote daemon: [[4209,0],1] on node mpi12

```

That is the output once `mpirun` notices the remote `orted` has failed. It's a bit of a better error message. There might be some more details we can extract from the `oob/tcp` point of failure - I'd have to look into that some more.

rhc54:
I have no issue with changing the verbosity calls, or with improving the error message

PR #2992, opened 2017-02-15
Fix binding policy bug and support pe=1 modifier

rhc54:
Allow someone to specify the "pe=N" modifier to a mapping policy when N=1. This equates to just "bind-to core", but helps people who use a script to set the PE policy. Fix a bug where setting the binding policy left a lingering "if-supported" flag that shouldn't be there.


rhc54:
@mnd999 Hopefully tagged the right person 😄

PR #3000, opened 2017-02-16
Fix some pmix configuration code

rhc54:
Remove stale file reference that caused a check to always fail. Update psm2 function check to new libs


rhc54:
@matcabral Please give this a try

matcabral:
👍 
passed basic hello_world test

PR #3004, opened 2017-02-17
oob/tcp: Adjust TCP keepalive default values

jjhursey:
We have encountered an environment where these values were set too low. We can fix that environment by adjusting the MCA parameters.

However, after talking to @jsquyres about how these numbers were selected (in https://github.com/open-mpi/ompi/commit/1a4c9960e1df53e32cab4f6c98e882111dced72b) we thought we should broaden the discussion.

So this PR increases the default values - we can discuss in the comments about whether these are *good* values or not.

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_component.c
+++ b/orte/mca/oob/tcp/oob_tcp_component.c
@@ -18,6 +18,7 @@
  * Copyright (c) 2014      NVIDIA Corporation.  All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -417,8 +418,8 @@ static int tcp_component_register(void)
                                           &mca_oob_tcp_component.disable_ipv6_family);
 #endif // OPAL_ENABLE_IPV6
 
-    // Default to keepalives every 60 seconds
-    mca_oob_tcp_component.keepalive_time = 60;
+    // Wait for this amount of time before sending the first keepalive probe
+    mca_oob_tcp_component.keepalive_time = 300;
     (void)mca_base_component_var_register(component, "keepalive_time",
                                           "Idle time in seconds before starting to send keepalives (keepalive_time <= 0 disables keepalive functionality)",
                                           MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
@@ -426,8 +427,8 @@ static int tcp_component_register(void)
                                           MCA_BASE_VAR_SCOPE_READONLY,
                                           &mca_oob_tcp_component.keepalive_time);
 
-    // Default to keepalive retry interval time of 5 seconds
-    mca_oob_tcp_component.keepalive_intvl = 5;
+    // Resend keepalive probe every INT seconds
+    mca_oob_tcp_component.keepalive_intvl = 20;
     (void)mca_base_component_var_register(component, "keepalive_intvl",
                                           "Time between successive keepalive pings when peer has not responded, in seconds (ignored if keepalive_time <= 0)",
                                           MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
@@ -435,9 +436,8 @@ static int tcp_component_register(void)
                                           MCA_BASE_VAR_SCOPE_READONLY,
                                           &mca_oob_tcp_component.keepalive_intvl);
 
-    // Default to retrying a keepalive 3 times before declaring the
-    // peer kaput
-    mca_oob_tcp_component.keepalive_probes = 3;
+    // After sending PR probes every INT seconds consider the connection dead
+    mca_oob_tcp_component.keepalive_probes = 9;
     (void)mca_base_component_var_register(component, "keepalive_probes",
                                           "Number of keepalives that can be missed before declaring error (ignored if keepalive_time <= 0)",
                                           MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,


jjhursey:
This article was helpful as a point of reference:
 * http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html

On RHEL 7, for example, these values default to:
```
Time:     7200
Interval: 75
Probe:    9

Time + Interval * Probe = 7875 sec
 2 hrs 11 min 15 sec to socket timeout
```

The current default values:
```
Time:     60
Interval: 5
Probe:    3

Time + Interval * Probe = 75 sec
1 min 15 sec to socket timeout
```

Suggested values:
```
Time:     300
Interval: 20
Probe:    9

Time + Interval * Probe = 480 sec
8 min to socket timeout
```

I think we want to keep the total timeout value fairly low so we don't sit in an allocation for 2 hours waiting for a socket to timeout. Waiting less than 10 minutes seems to be more reasonable.

rhc54:
I gather that the problem is that the daemon doesn't use that socket once launch is complete unless the app is generating some stdout/err, and so the OS eventually allows it to timeout. I thought the keepalive was supposed to resolve that, but it sounds like it fails in some scenarios. Any ideas as to why keepalives are being dropped? Would it make sense for us to send a heartbeat down that wire every so many minutes to resolve this problem?

I personally don't have an issue with an 8 minute timeout - if MPI comm is on-going, they'll let us know sooner anyway.

jsquyres:
@rhc54 This happened in a Cisco environment while running a compute-heavy benchmark -- several hours into the run, we suddenly get a message from `mpirun` indicating that it had lost contact with an `orted`.  It was fairly repeatable behavior.  We could find no reason for this -- as far as we could tell, there were no network or node failures -- just `mpirun` deciding that it had lost contact with an `orted`.

Our best guess is that the low timeout (15 seconds) was too short for the remote OS to respond to the KEEPALIVE probe, especially when all the cores were busy doing MPI process computation.  I don't have proof of that, but it's the only thing we could think of.  

This caused us to look at the OS defaults for KEEPALIVE, and the values that OMPI is setting.  Now armed with a better understanding of what these 3 values mean (compared to when 1a4c9960e1df53e32cab4f6c98e882111dced72b was committed), the OMPI defaults seem far too low (i.e., the time is too short, and the number of probes / retries is too low).

FWIW: we confirmed that increasing the timeout and increasing the number of probes made the problem go away in the Cisco environment.  We didn't use these exact values, but these values seem good to us.

rhc54:
I understand the scenario - I'm just thinking about the best solution. To be clear, mpirun isn't deciding that it lost contact. The OS is _telling_ mpirun that it lost contact.

One way to solve the problem is to do as this suggests and just tell the OS to check less often, and to allow more failures before declaring the connection to be lost.

Another way would be to "exercise" the socket periodically via a heartbeat IFF no stdout/err had used it during that time.

Both methods introduce jitter - I'm not sure which is better, hence the question.

jsquyres:
👍  Fair enough.  I think the OS-level keepalive is letter jitter.

bosilca:
It is extremely difficult for the OS to let a socket timeout. However, if you have a NAT or masquerading, then the middleman can decide to drop the connection after a certain time. In this case keep alive should help, for as long as its value is below the NAT timeout.

jsquyres:
There was no NAT or firewalling between `mpirun` and the `orted`s.

bosilca:
The [HOWTO](http://www.tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/) mentionned above indicates that the only possibility for disconnections, due to network inactivity, is when one of the peers is behind a NAT proxy or a firewall. As you have no NAT/firewall on your machine I think we should look for the culprit somewhere else.

Going back to the info about the timeout, intvl and probes. You said that the current values were too small, leading to a total timeout of 75s. Yes, that's pretty small, but to fail a keepalive this would mean that for 75s the kernel on the other side had no opportunity to answer the keepalive request. I cannot foresee any reason why an interruptible kernel doesn't act on a keep alive for 75s (except if you have a kernel panic, or if you have thousands of open connections between the 2 peers and each one of them if generating keepalive messages).

Let think a little about how this could happen. You mentionned that the application was running, so at some point there were sane connections between the daemons. Assuming that is extremely unlikely in a normal scenario and **in a cluster setting** to timeout a connected sockets what other reason can there be ? What if there were other pending connection establishment requests, and they timeout because they were not accepted in time ? Suppose a connection storm at startup that forces one of the daemon to retry to connect. Are we sure that we correctly close all sockets that we discard ?

If this is reproducible, can you check what is the status of the socket that fails ? After how many minutes it happens ?

jsquyres:
Clarification: 15 seconds, not 75.  There is first a 60 second delay before KEEPALIVE kicks in, then it sends 3 probes, 5 seconds apart.  Hence, it's only looking for ACKs to the probes in that 15 seconds -- not 75.

Still, I agree that 15 seconds is effectively a "long" time.  I'd say it's probably the upper bound for a kernel to not respond to a KEEPALIVE probe.  Not responding for 75 seconds is pretty unbelievable.

Another possibility is that the remote kernel *did* respond to 1, 2, or even all 3 of the probes, but there was enough congestion in the network that either the probes or their ACKs got dropped.  (KEEPALIVE probes and their ACKs are not reliable).  Such congestion is definitely a possibility -- the benchmark we were running was HPCC.  I don't remember which test was running when it got disconnected, but it could have been RandomAccess (i.e., something that generates a LOT of traffic -- and usNIC traffic can be on the same network as the orted sockets).  Hence, it may not be 15 seconds is too short of a timeout, but rather that 3 probes is too few.  That might actually be a bit more believable...?

FWIW: The timeout happens after about 3 hours of running HPCC at 32 servers of 24 cores each (i.e., np=768).  HPCC is definitely progressing (via usNIC, which, while on a different VLAN than the orted traffic, is still on the same physical network, so congestion issues in one can affect the other).  We repeated the test a few times and got the same results, but we've unfortunately run out of time and had to turn the cluster over to some others for at least a little while.  Meaning: I probably can't re-run this HPCC experiment in the immediate future.

ibm-ompi:

 Gist: https://gist.github.com/91148280cf6e82678a3406055526f7f4

ibm-ompi:

 Gist: https://gist.github.com/85413791ec326a3e4426d75bddb4de8c

ibm-ompi:

 Gist: https://gist.github.com/75c07f0d7e9a6cb7367e95ff48a5e30a

bosilca:
@jsquyres if the packets are dropped regularly (aka you can reproduce this issue) in such a small context (only 768 processes doing a random access pattern), then the hardware might be to blame.

Anyway, I don't think that pushing up the values for the keep alive will hurt any normal usage scenario, while it will eventually hide the issue in your particular case. 👍

rhc54:
@bosilca Does this potentially impact any fault tolerance usage? I guess at worst it simply delays detection of a failure - is that a concern for you folks?

rhc54:
bot:ibm:retest

bosilca:
@rhc54 it does indeed have the potential to delay the detection in the case where the peer node crashes. But relaying on the TCP keep alive mechanism to detect a node failure is the worst possible solution, one that we have diverged for quite some time.

Issue #2998, opened 2017-02-16
host mapping mismatch on cray


marksantcroos:
This is on Blue Waters a Cray with Torque.

nodenames are in the shape of `nid%05d`, but entries in `$PBS_NODEFILE` are of the shape of `%d`.

- ras: tm
- plm: alps
- ess: alps

Until recently (at least Jan 18, bisect in progress) I could use the short format, but now it requires the full format. Is that an intentional change?

```
--------------------------------------------------------------------------
There are no allocated resources for the application:
  hostname
that match the requested mapping:
  -host: 19696

Verify that you have mapped the allocated resources properly for the
indicated specification.
```

rhc54:
Not intentional so far as I am aware - I'd have to look to see what tripped it

jjhursey:
Is this with `master` or one of the release branches?

marksantcroos:
@jjhursey master.

naughtont3:
Hi @marksantcroos, I happen to be testing with `master` today on a local Cray using Torque. It seems to be working, maybe we can sync up via email to see if the environments are different.

naughtont3:
Actually, I bet I didn't run the same command as you.  I do get an error if i pass `--host` argument on command-line.  My bad.

```
$ cat $PBS_NODEFILE | uniq
28
29
$ mpirun -np 2 --map-by node ./mpiexample 
I am rank: 1 of 2  (host=nid00029)
I am rank: 0 of 2  (host=nid00028)
$ mpirun -np 2 --map-by node hostname
nid00028
nid00029
$ mpirun -np 1 --map-by node hostname
nid00028
$ mpirun --host 28 -np 1 --map-by node hostname
--------------------------------------------------------------------------
There are no allocated resources for the application:
  hostname
that match the requested mapping:
  -host: 28
.
Verify that you have mapped the allocated resources properly for the
indicated specification.
--------------------------------------------------------------------------
$
$ mpirun --host nid00028 -np 1 --map-by node hostname
--------------------------------------------------------------------------
No nodes are available for this job, either due to a failure to
allocate nodes to the job, or allocated nodes being marked
as unavailable (e.g., down, rebooting, or a process attempting
to be relocated to another node when none are available).
--------------------------------------------------------------------------
$ mpirun -np 1 --map-by node hostname
nid00028
$
```

marksantcroos:
> I do get an error if i pass --host argument on command-line.
[...]

PR #3006, opened 2017-02-18
Support -host launch_id

rhc54:
Allow -host to look for the number of a host when running in a managed environment that supports launch id's. For example, this will allow someone who has been allocated a node of "nid0015" to refer to it with "-host 15".


rhc54:
@marksantcroos I don't have a way to test this directly, but believe this will provide the requested support.

Refs #2998

marksantcroos:
Thanks @rhc54 !

PR #3007, opened 2017-02-19
Fix launch_id matching of -hosts

rhc54:
Need to check the entire value instead of just the last N digits. Otherwise, "-host 15" will match nid0015, nid0115, and any other launch id ending in 15


rhc54:
@marksantcroos Can you please check that this is correct?

marksantcroos:
@rhc54 I see indeed that the previous version matched to loosely, but this additional commit breaks it altogether.

rhc54:
@marksantcroos Yeah, strtol fooled me - it can apparently return either a NULL _or_ a zero-length string, and I wasn't checking for that option. Should be fixed now

marksantcroos:
Thanks, it works now! Will wait for the tests to complete and then do the merge.

PR #3017, opened 2017-02-22
Update to PMIx master to include dlopen fixes and addition of libltdl support

rhc54:


rhc54:
Refs #2107

Issue #3026, opened 2017-02-24
dvm: add support for passing DVM URI via environ


naughtont3:
When launching with the DVM you need to pass the URI location on the command line via the `--hnp` argument.  Some users may want to simplify their command-line args (e.g., in interactive sessions, automation scripts, etc.). 

The proposed feature would be used like this:
 
```
    orte-dvm --report-uri dvmuri --daemonize                                                                      
    export ORTE_HNP_DVM_URI=$(cat dvmuri)                                                                         
    mpirun -np 1 hostname 
```
                                          
You can also override the EnvVar by passing '--hnp' arg on cmdline.                                               

```                                                                                                                  
    mpirun --hnp file:dvmuri2 -np 1 hostname 
```

I added a minor change in schizo:ompi that will allow a user to do the above.

PR #3027, opened 2017-02-24
dvm: Add envvar 'ORTE_HNP_DVM_URI' to schizo:ompi

naughtont3:
Add ability to pass DVM URI purely via environment
to simplify invocation from command-line (e.g., start dvm,
export URI, mpirun w/o needing to add `--hnp` arg).
If user passes both envvar *and* cmdline, the cmdline wins.

Refs #3026

### DIFF
--- a/orte/mca/schizo/ompi/schizo_ompi.c
+++ b/orte/mca/schizo/ompi/schizo_ompi.c
@@ -12,7 +12,8 @@
  * Copyright (c) 2006-2013 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2009-2016 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
+ * Copyright (c) 2011-2017 Oak Ridge National Labs.  All rights reserved.
+ * Copyright (c) 2017      UT-Battelle, LLC. All rights reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
@@ -478,6 +479,15 @@ static int define_cli(opal_cmd_line_t *cli)
         if (!takeus) {
             return ORTE_ERR_TAKE_NEXT_OPTION;
         }
+    }
+
+    /*
+     * Check if a HNP DVM URI is being passed via environment.
+     * Note: Place before opal_cmd_line_parse() so that
+     * if user passes both envvar & cmdln, the cmdln wins.
+     */
+    if (NULL != getenv("ORTE_HNP_DVM_URI")) {
+        orte_cmd_options.hnp = strdup(getenv("ORTE_HNP_DVM_URI"));
     }
 
     /* just add ours to the end */


rhc54:
I think this is fine. FWIW: the PMIx reference server does this a little more automatically:

```c
    char hostname[OPAL_MAXHOSTNAMELEN];

    /* ****************************************************************/
    /* we want to be able to detect that the PSRVR is up and running
     * prior to attempting to connect to it. The code logic in ORTE
     * actually supports such things, but unfortunately will emit
     * error messages about the contact file missing before we can
     * arrive at the point where retries can be done. So...let's
     * setup the path to the contact file here, check to see if it
     * exists, and then cycle a while if it doesn't. */

    if (NULL != (tdir = getenv("PMIX_SERVER_TMPDIR"))) {
        filename = opal_os_path(false, tdir, "contact.txt", NULL);
    if (NULL == filename) {
        fprintf(stderr, "OUT OF MEMORY\n");
        exit(1);
    }
    } else {
        /* get the nodename */
        orte_process_info.nodename = strdup(hostname);

        /* setup the top session directory name */
        if (ORTE_SUCCESS != orte_setup_top_session_dir()) {
            fprintf(stderr, "OUT OF MEMORY\n");
            exit(1);
        }
        filename = opal_os_path(false, orte_process_info.top_session_dir, "dvm", "contact.txt", NULL);
        if (NULL == filename) {
            fprintf(stderr, "OUT OF MEMORY\n");
            exit(1);
        }
        free(orte_process_info.nodename);
        orte_process_info.nodename = NULL;
        free(orte_process_info.top_session_dir);
        orte_process_info.top_session_dir = NULL;
    }
    /* check to see if the file exists - loop a few times if it
     * doesn't, delaying between successive attempts */
    n = 0;
    while (n < PRUN_LOOP_LIMIT &&
           0 != access(filename, R_OK)) {
        sleep(1);
        ++n;
    }
    /* if we still don't see the file, then give up */
    if (0 != access(filename, R_OK)) {
        fprintf(stderr, "PSRVR contact file %s not found - cannot continue\n", filename);
        exit(1);
    }
    free(filename);

```

For the orte-dvm, you'd just look at $TMPDIR instead of PMIX\_SERVER\_TMPDIR - no need to have the user set an envar.

naughtont3:
Cool! OK, I'll take a look and give that a try.  Thanks.

naughtont3:
I think EnvVar will be better b/c it avoids problems with the frontend (orterun) trying to guess at the proper session directory for "contact.txt".  If we could reliably get the child PID (orte-dvm) we could automate it, but that still assumes user wants to use DVM and not just run orterun w/o DVM. In EnvVar case the user is just using method to avoid additional typing at the shell or in scripts (also keeps their command-lines the same if they choose to run things w/ and w/o the dvm in tests).

Example:

```
  orte-dvm --hostfile h1 --report-uri uri1 --daemonize
  orte-dvm --hostfile h2 --report-uri uri2 --daemonize
   
 # Now the user would set the EnvVar to decide which URI to use for subsequent 'orterun'
```

In contrast it is clear if they do the following:

```
   orte-run --hostfile h1 --report-uri uri1 --daemonize
   orte-run --hostfile h2 --report-uri uri2 --daemonize

  export ORTE_HNP_DVM_URI=$(cat uri1)
  orterun -np 1 hostname
 
  orterun --hnp file:uri2 -np 1 hostname
```

That's a silly example, but it is clear and under user's control for passing URI, just with a bit less typing if they keep using EnvVar.

naughtont3:
@rhc54 Do you think the EnvVar name is ok  `ORTE_HNP_DVM_URI`?  I preferred having `ORTE_` as prefix as this is specific to that runtime but maybe it should be `OMPI_`?

rhc54:
sure - no problem with the approach or the name

naughtont3:
OK, thanks Ralph.

naughtont3:
Fixes #3026

PR #3030, opened 2017-02-25
v2.1.0: more updates for release

jsquyres:
The NEWS bullets are a first cut.  More work needs to be done on them before release.


ompiteam:
Not waiting for CI.

jjhursey:
@jsquyres Can you add `[skip ci]` and `bot:notest` to your PR message so the CI doesn't even kick off for these types of PRs? Travis will still fire, but the Jenkins setup's should just ignore it.

jsquyres:
@jjhursey Will do.  I always forget what the magic token is to make it skip; I'll try better to remember in the future. 🎗

PR #3038, opened 2017-02-27
Topic/hwloc

ggouaillardet:


rhc54:
What is this PR about?

rhc54:
if we need some other changes, could you please just create a PR against the person's branch instead of replacing the entire thing? It generates a ton of confusion and wasted time.

Issue #3032, opened 2017-02-25
Something wrong with configure for atomics on master


rhc54:
I am consistently seeing this output in builds, both on my machine and from Jenkins servers:

```shell
11:10:41 checking dependency style of gcc -std=gnu99... gcc3
11:10:41 checking for BSD- or MS-compatible name lister (nm)... /bin/nm -B
11:10:41 checking the name lister (/bin/nm -B) interface... BSD nm
11:10:41 checking for fgrep... /bin/grep -F
11:10:41 checking for __atomic builtin atomics... yes
11:10:41 checking for processor support of __atomic builtin atomic compare-and-swap on 128-bit values... no
11:10:41 checking for __atomic builtin atomic compare-and-swap on 128-bit values with -mcx16 flag... yes
11:10:41 checking if __int128 atomic compare-and-swap is always lock-free... yes
11:10:41 checking if .proc/endp is needed... no
11:10:41 checking directive for setting text section... .text
11:10:41 checking directive for exporting symbols... .globl
11:10:41 checking for objdump... objdump
11:10:41 checking if .note.GNU-stack is needed... yes
11:10:41 checking suffix for labels... :
11:10:41 checking prefix for global symbol labels... 
11:10:41 checking prefix for lsym labels... .L
11:10:41 checking prefix for function in .type... @
11:10:41 checking if .size is needed... yes
11:10:41 checking if .align directive takes logarithmic value... no
11:10:41 checking if processor supports x86_64 16-byte compare-and-exchange... yes
11:10:41 checking if compiler correctly handles volatile 128bits... yes
11:10:41 checking for assembly architecture... AMD64
11:10:41 checking for builtin atomics... BUILTIN_GCC
11:10:41 checking for pre-built assembly file... no (atomic-amd64-linux
11:10:41 amd64-linux-nongas.s not found)
11:10:41 checking whether possible to generate assembly file... failed
11:10:41 configure: WARNING: Could not build atomic operations assembly file.
11:10:41 configure: WARNING: There will be no atomic operations for this build.
11:10:41 checking for atomic assembly filename... none
```

Note that this is on Intel architectures. Digging a little deeper, it isn't clear to me where the ```host``` variable is getting set, and that appears to lead to the architecture being mistakenly identified.

ggouaillardet:
in ```OPAL_ASM_FIND_FILE``` should we also make an exception for ```BUILTIN_GCC``` just like we do for ```BUILTIN_SYNC``` and others ?

[...]

Issue #3036, opened 2017-02-26
configure: assembler stdout incorrect (cosmetic issue only)


rhc54:
I'm seeing an incorrect identification of the hardware architecture (similar to issue reported on master #3032):

```
*** Assembler
checking dependency style of gcc... gcc3
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
checking the name lister (/usr/bin/nm -B) interface... BSD nm
checking for fgrep... /usr/bin/grep -F
checking if .proc/endp is needed... no
checking directive for setting text section... .text
checking directive for exporting symbols... .globl
checking for objdump... objdump
checking if .note.GNU-stack is needed... yes
checking suffix for labels... :
checking prefix for global symbol labels... 
checking prefix for lsym labels... .L
checking prefix for function in .type... @
checking if .size is needed... yes
checking if .align directive takes logarithmic value... no
checking if processor supports x86_64 16-byte compare-and-exchange... yes
checking if compiler correctly handles volatile 128bits... yes
checking if gcc supports GCC inline assembly... yes
checking if gcc supports DEC inline assembly... no
checking if gcc supports XLC inline assembly... no
checking for assembly format... default-.text-.globl-:--.L-@-1-0-1-1-1
checking for assembly architecture... AMD64
checking for builtin atomics... BUILTIN_NO
checking for pre-built assembly file... no (atomic-amd64-linux.s not found)
checking whether possible to generate assembly file... yes
checking for atomic assembly filename... atomic-local.s

```

This was on an Intel x86_64 box.

jsquyres:
Per the discussion on #3039, this feels like a cosmetic issue, not something that is actually doing the Wrong Thing on the back-end.  Defer to v2.1.1.

rhc54:
resolved by @ggouaillardet

PR #3039, opened 2017-02-27
v2.x: configury: fix asm atomic detection

ggouaillardet:
there is no need to look for an assembly file when BUILTIN_GCC is used

Fixes open-mpi/ompi#3032
Refs open-mpi/ompi#3036

(cherry picked from commit open-mpi/ompi@2f4013ce33723e902160925847ce62a19f84bc71)


ggouaillardet:
@hppritcha @rhc54 this fixes open-mpi/ompi#3032
my analysis is that ```configure``` displays a false positive, and GCC builtin atomics are used.
so imho, this is not a blocker, though this would be nice to have when ```2.1.0``` is released

ggouaillardet:
@hjelmn @rhc54 is ```opal/asm/asm-data.txt``` ```x86_64``` is referred as ```AMD64```.
my guess is that comes back from the Opteron debut days.

since this is not specific to AMD processors, should we rename this as ```X86_64``` to avoid any confusion ? (this is a genuine question, and as far as i am concerned, i am fine with ```AMD64```)
if we can reach a consensus, i will be happy to update the PR
fwiw, 32 bits is referred as ```IA32```, so this could also be renamed to ```X86_32``` or something else but vendor neutral.

rhc54:
rename: absolutely! The two archs have diverged over time, and it is definitely misleading and confusing.

ggouaillardet:
done in #3041 
will cherry-pick after it passes CI

hppritcha:
Is this fixing a correctness problem that impacts applications?

rhc54:
probably could wait for 2.1.1

jsquyres:
Breakdown of the two commits:

1. The AMD64->x86_64 rename feels cosmetic (misleading, yes, but no one has complained for quite a long time).  Feels like it can wait for v2.1.1.
1. The BUILTIN_GCC fix feels like an optimization -- i.e., prevent some additional configure checks that aren't necessary if we're using builtins.  This also feels like it can be deferred to v2.1.1.

Please let me know if these are wrong (i.e., this PR is *needed* for v2.1.0).  Thanks!

ggouaillardet:
@jsquyres the issue was initially reported by @rhc54 on #3032 
here is a snippet of the ```configure``` output

```
11:10:41 checking for assembly architecture... AMD64
11:10:41 checking for builtin atomics... BUILTIN_GCC
11:10:41 checking for pre-built assembly file... no (atomic-amd64-linux
11:10:41 amd64-linux-nongas.s not found)
11:10:41 checking whether possible to generate assembly file... failed
11:10:41 configure: WARNING: Could not build atomic operations assembly file.
11:10:41 configure: WARNING: There will be no atomic operations for this build.
11:10:41 checking for atomic assembly filename... none
```

at first glance, an end user might legitimately think atomic operations could not be built because his/her genuine Intel processor was misidentified as an AMD processor.

the truth is everything is working just fine
- the processor is detected as a vendor neutral x86_64 (this is fixed by the ```AMD64-X86_64``` rename commit)
- since ```BUILTIN_GCC``` atomics are used, there is no need to use the atomic operations assembly file,
  so there is no need to build it nor display a warning (once again, atomics __are__ built though the warning message says otherwise), and this is fixed by the ```BUILTIN_GCC``` commit.

bottom line
- a bogus and misleading warning message is displayed
- ```BUILTIN_GCC``` atomics __are__ built on x86_64 platforms
- ```AMD64``` should be read as ```vendor neutral x86_64```

strictly speaking, and imho, this PR is not mandatory for ```v2.1.0```, though it would be nice
to add a few words about it in the README so confused users (if any) can be pointed to.

jsquyres:
@ggouaillardet Thanks for the explanation.

@hppritcha Yes, perhaps we should add something to README.

PR #3050, opened 2017-02-27
orte-clean: fix bad username usage, add orte-dvm

naughtont3:

### DIFF
--- a/config/opal_check_ps.m4
+++ b/config/opal_check_ps.m4
@@ -13,6 +13,7 @@ dnl                         All rights reserved.
 dnl Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2008      Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2014      Intel, Inc. All rights reserved.
+dnl Copyright (c) 2017      UT-Battelle, LLC. All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
@@ -30,11 +31,11 @@ PS_FLAVOR="unknown"
 ps -A -o fname > /dev/null 2>&1
 
 if test "$?" = "0"; then
-     PS_FLAVOR="ps -A -o fname,pid,user"
+     PS_FLAVOR="ps -A -o fname,pid,uid"
 else
      ps -A -o command > /dev/null 2>&1
      if test "$?" = "0"; then
-         PS_FLAVOR="ps -A -o command,pid,user"
+         PS_FLAVOR="ps -A -o command,pid,uid"
      fi
 fi
 AC_MSG_RESULT([$PS_FLAVOR])
--- a/orte/tools/orte-clean/orte-clean.c
+++ b/orte/tools/orte-clean/orte-clean.c
@@ -17,6 +17,7 @@
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2015      Intel, Inc. All rights reserved.
+ * Copyright (c) 2017      UT-Battelle, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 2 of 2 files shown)


naughtont3:
Fixes ##3049

rhc54:
Two quick questions:

1. is getlogin_r actually portable to all the various flavors (e.g., NetBSD)?

1. do we even need it? I can't find anywhere that we actually use that value, but maybe I'm missing it...

naughtont3:
I'll have to check on portability of `getlogin_r()` (or non-reentrant `getlogin()`).  

We need it to get the username of the calling user.  Alternatively I could use `getenv("USER")`, but not sure that's fully portable either.  Is there another, more typical, way you usually get the username for current user?  All we need is that username string instead of uid.

naughtont3:
Looks like `getlogin_r` is in NetBSD, but is *not* on OSX. :-(

rhc54:
Ah, okay - I now see where it gets used. I think we have used getpwuid in the past, but that is definitely _not_ portable, which is why we went away from getting usernames. Problem is that even though some OS might provide, some vendors strip the OS down and remove things like getpwuid.

naughtont3:
I could change to using UID instead of username (change would be in the configury I indicated above) for [opal_check_ps](https://github.com/open-mpi/ompi/blob/master/config/opal_check_ps.m4).  

But that seemed less good as the UID might differ across machines, while username would stay the same.

naughtont3:
Correction: It does appear the `getlogin_r()` or some variant exists on OSX.

rhc54:
I don't have getlogin_r on my Mac (sierra).

I'm not sure I really understand the problem. The _ps_ command will take an option to specify the uid of the user whose procs you want to find, yes? So why do we need the username?

Remember, orte-clean runs on every machine - so it locally discovers the uid of the user who ran it anyway. We need it for the session directory name.

naughtont3:
The `ps` command returns output of the form: `CommandName PID UserName`.  
Example:
```
$ ps -A -o fname,pid,user |grep orte
orte-dvm  4400 3t4
orte-dvm  4452 3t4
```

However, the comparison was between the `UserName` (e.g.,`3t4`) and the `UID` of the user (e.g., `1000`).  

So I used `getlogin_r()` to get the UserName for user and then we compare that against the contents from `ps`, i.e., to prune out the processes not belonging to our user.

Maybe you're suggesting we not use this form of the `ps` command, and just limit the output we'll read to just our user (i.e., forego username lookup and pruning ps output by username).  

Your point about UID appearing in session dir also means that as long as we are looking for UIDs that match UID running `orte-clean`, then we can avoid username lookups too.   

So sounds like it would be better to change the PS output to return UID and leave C code as-is (not use getlogin_r()).

rhc54:
I think so - I'm open to alternatives, but getting the username always seems to create portability problems in some fashion

naughtont3:
@rhc54 OK, so I changed things to use UIDs instead of USERNAMEs to forego use of getlogin_r().   I update the PR.  Let me know if this is more appropriate. 

```
sal9000:$ orte-dvm --report-uri dvmuri --daemonize
sal9000:$ ps auwx | grep orte | grep -v grep
3t4      15848  0.2  0.2 706544  5868 ?        Ssl  18:17   0:00 orte-dvm --report-uri dvmuri --daemonize
sal9000:$ orte-clean --verbose
orte-clean: cleaning session dir tree /tmp/ompi.sal9000.1000
orte-clean: killing any lingering procs
orte-clean: found potential rogue orted process (pid=15848,uid=1000), sending SIGKILL...
sal9000:$ ps auwx | grep orte | grep -v grep
sal9000:$
```
```
root:# ps auwx  | grep orte-dvm |grep -v grep
3t4      15993  0.0  0.2 706544  5868 ?        Ssl  18:17   0:00 orte-dvm --report-uri dvmuri --daemonize
root:# id
uid=0(root) gid=0(root) groups=0(root)
root:# orte-clean --verbose
orte-clean: cleaning session dir tree /tmp/ompi.sal9000.0
orte-clean: killing any lingering procs
orte-clean: found potential rogue orted process (pid=15993,uid=1000), sending SIGKILL...
root:# ps auwx  | grep orte-dvm |grep -v grep
root:#
```

Note: For issue trails, my MacOSX was 10.10.5 and it had getlogin_r() but did not have the manual page (i.e., tiny test code compiled/ran).

naughtont3:
OK, squashed down and added sign-off.

PR #3067, opened 2017-02-28
v2.1.x: --with-hwloc=external fix

### LINKED ISSUE #2616: Build failure with --with-hwloc=external

jsquyres:
Fixes #2616 for v2.1.x. Thanks to @opoplawski for reporting.

Set for v2.1.1 release, since the v2.1.0 workaround is "good enough" for the initial release, and we're trying to get v2.1.0 out the door.


jsquyres:
I updated this PR (and rebased it, just for good measure) to include a cherry-pick of b8dfd49e976b6c5864274008f7f72c7e55754a89, which fixes a regression introduced by one of the commits in this PR (i.e., one of the commits in this PR breaks the ability to run `autogen.pl` in a distribution tarball).

I could have just updated the offending commit in this PR, but then the commit wouldn't be a strict cherry-pick from master.  Instead, I left the original commit, and just added an additional cherry-pick from master that fixes the issue.

PR #3071, opened 2017-03-01
yalla/mtl_mxm/hcoll: open memory component to activate memory hooks.

yosefe:

PR #3072, opened 2017-03-01
pml_yalla/mtl_mxm/hcoll: open memory component to activate memory hooks.

yosefe:
Memory hooks are now set-up on demand. pml/yalla, mtl/mxm and
coll/hcoll need the memory hooks, so make sure those are installed.

picked from master #3071

### DIFF
--- a/ompi/mca/coll/hcoll/coll_hcoll_component.c
+++ b/ompi/mca/coll/hcoll/coll_hcoll_component.c
@@ -251,6 +251,8 @@ static int hcoll_open(void)
 
     cm->libhcoll_initialized = false;
 
+    (void)mca_base_framework_open(&opal_memory_base_framework, 0);
+
     /* Register memory hooks */
     if ((OPAL_MEMORY_FREE_SUPPORT | OPAL_MEMORY_MUNMAP_SUPPORT) ==
         ((OPAL_MEMORY_FREE_SUPPORT | OPAL_MEMORY_MUNMAP_SUPPORT) &
@@ -293,5 +295,8 @@ static int hcoll_close(void)
         HCOL_VERBOSE(1,"Hcol library finalize failed");
         return OMPI_ERROR;
     }
+
+    mca_base_framework_close(&opal_memory_base_framework);
+
     return OMPI_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


yosefe:
@jladd-mlnx @miked-mellanox @alex-mikheev

ompi/mca/coll/hcoll/coll_hcoll_component.c
```diff
@@ -251,6 +251,8 @@ static int hcoll_open(void)
 
     cm->libhcoll_initialized = false;
 
+    (void)mca_base_framework_open(&opal_memory_base_framework, 0);
```
jsquyres: Why do you want to ignore the return status here -- don't you want to error if the memory hooks fail to get setup properly?

ompi/mca/mtl/mxm/mtl_mxm_component.c
```diff
@@ -200,6 +201,7 @@ static int ompi_mtl_mxm_component_open(void)
     }
 
 #if MXM_API >= MXM_VERSION(2,0)
+    (void)mca_base_framework_open(&opal_memory_base_framework, 0);
```
jsquyres: Same question here (why ignore the return status?).

ompi/mca/pml/yalla/pml_yalla.c
```diff
@@ -112,6 +113,8 @@ int mca_pml_yalla_open(void)
 
     PML_YALLA_VERBOSE(1, "mca_pml_yalla_open");
 
+    (void)mca_base_framework_open(&opal_memory_base_framework, 0);
```
jsquyres: Same question here (why ignore the return status?).

jsquyres:
Are these fixes also necessary in v2.0.x?

ompi/mca/coll/hcoll/coll_hcoll_component.c
yosefe: @jsquyres just like in opal/rcache - the next line of code calls  opal_mem_hooks_support_level() to detect if memory hooks are supported, so no need to check the return value from framework open call.

jsquyres: Got it.  👍

jsquyres:
@hppritcha Good to go.

yosefe:
@jsquyres It's not needed on v2.0.x because it opens the memory base framework from opal_init

Issue #3042, opened 2017-02-27
MPI_Barrier hangs in 10-20h long runs


Welcome-Dear-Evgueni:
Hi,

After 10-20 hours, MPI_Barrier hangs with the following stack.

It looks like some error happens during mca_btl_tcp_proc_create, and then mca_btl_tcp_proc_destruct cannot proceed because mca_btl_tcp_component.tcp_lock is not recursive.

The stack below has been captured using OMPI 2.0.2a1, and the OMPI 2.0.2 release does not seem to fix this issue.

I would be greateful for any ideas about the root-cause of this error -- wrong host config, etc.

Unfortunately (as usual), the application is too large and I do not have a small reproducer.

Here is some information that seems relevant.

MPI is used to distribute jobs to CUDA cards inside one host.
There are no inter-host communications.
Each MPI process is multithreaded.
MPI is called from different threads but the calls are serialized at the application level.
MPI is built with --enable-thread-multiple, the app calls MPI_Init_thread(MPI_THREAD_MULTIPLE).
The application uses 50-100 communicators which are created at the application launch in the master thread and they all stay alive while the application is running.
Each communicator contains all MPI processes.
All processes get stuck with identical stack.
I used gdb to attach to the processes after they got stuck and verified that the processes passed the correct communicator to MPI_Barrier.

```
#0  __lll_lock_wait ()
    at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f0e250c7649 in _L_lock_909 ()
   from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007f0e250c7470 in __GI___pthread_mutex_lock (
    mutex=0x7f0e1ad547f8 <mca_btl_tcp_component+440>)
    at ../nptl/pthread_mutex_lock.c:79
#3  0x00007f0e1ab4f76d in mca_btl_tcp_proc_destruct ()
   from /home/espetrov/lib/openmpi/mca_btl_tcp.so
#4  0x00007f0e1ab4fce1 in mca_btl_tcp_proc_create ()
   from /home/espetrov/lib/openmpi/mca_btl_tcp.so
#5  0x00007f0e1ab48e2c in mca_btl_tcp_add_procs ()
   from /home/espetrov/lib/openmpi/mca_btl_tcp.so
#6  0x00007f0e1ab50270 in mca_btl_tcp_proc_lookup ()
[...]

Issue #3035, opened 2017-02-26
master: TCP BTL addressing fail


jsquyres:
Periodically, the Cisco MTT sees odd TCP BTL addressing failures.

For example, the one-sided TCP BTL failures from last night/s MTT on master (there's ORTE errors there, too -- ignore those): https://mtt.open-mpi.org/index.php?do_redir=2399

It looks like the TCP BTL fails the incoming connection, which then causes a frag mismatch size in OB1.  Here's a stack trace:

```
Program terminated with signal SIGABRT, Aborted.
#0  0x0000003370632925 in raise () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x2aaab959a700 (LWP 1394))]
(gdb) bt
#0  0x0000003370632925 in raise () from /lib64/libc.so.6
#1  0x0000003370634105 in abort () from /lib64/libc.so.6
#2  0x000000337062ba4e in __assert_fail_base () from /lib64/libc.so.6
#3  0x000000337062bb10 in __assert_fail () from /lib64/libc.so.6
#4  0x00002aaaaacdbaec in mca_pml_ob1_put_completion (frag=0x781380, 
    rdma_size=26666) at pml_ob1_recvreq.c:197
#5  0x00002aaaaacd80b6 in mca_pml_ob1_recv_frag_callback_fin (btl=0x6dd500, 
    tag=73 'I', des=0x2aaab9299080, cbdata=0x0) at pml_ob1_recvfrag.c:434
#6  0x00002aaaab4d0108 in mca_btl_tcp_endpoint_recv_handler (sd=24, flags=2, 
    user=0x739220) at btl_tcp_endpoint.c:893
#7  0x00002aaaab512393 in event_persist_closure (base=0x65c4e0, ev=0x739510)
    at event.c:1321
#8  0x00002aaaab5124a2 in event_process_active_single_queue (base=0x65c4e0, 
    activeq=0x65c9d0) at event.c:1365
#9  0x00002aaaab51276f in event_process_active (base=0x65c4e0) at event.c:1440
#10 0x00002aaaab512dc2 in opal_libevent2022_event_base_loop (base=0x65c4e0, 
    flags=1) at event.c:1644
#11 0x00002aaaab4cbe89 in mca_btl_tcp_progress_thread_engine (
    obj=0x2aaaab8fc1c0 <mca_btl_tcp_progress_thread>)
    at btl_tcp_component.c:781
[...]

PR #3078, opened 2017-03-02
v2.1.x: TCP BTL fixes

jsquyres:
Two fixes from @bosilca on master:

1. Do not `free()` a buffer on the stack.
1. Fix a code THREAD_MULTIPLE bug in the TCP BTL (per #3042)

@bosilca I had to fix a conflict when porting over the #3042 fix.  It seems to work, but I'd appreciate your review.

@hppritcha I marked this as for v2.1.0 because it fixes a real bug that a user encountered (thanks to Evgueni Petrov @Welcome-Dear-Evgueni for reporting).


jsquyres:
@hppritcha I think this PR is worthy of v2.1.0; we can discuss and see if you agree.

It is not vital to get into tonight's MTT, but is still fixing some real bugs, and paving the way to get more information to solve https://github.com/open-mpi/ompi/issues/3035.

PR #3099, opened 2017-03-04
ompi/pml/ucx: Fix uninitialized UCX request field.

artpol84:
Signed-off-by: Artem Polyakov <artpol84@gmail.com>

PR #3100, opened 2017-03-04
ompi/pml/ucx: Fix uninitialized UCX request field.

artpol84:


artpol84:
Fixes openucx/ucx#1353

artpol84:
`MCA_PML_BASE_SEND_STANDARD` looks strange for the receive request. Maybe we can change this check as an alternative:
https://github.com/open-mpi/ompi/blob/master/ompi/mca/pml/ucx/pml_ucx_request.c#L169:L171
so that only for the send requests we will check this field.

artpol84:
@yosefe @alex-mikheev I think that now fix looks more logical.

miked-mellanox:
@artpol84 - please integrate it into 2.x

artpol84:
@miked-mellanox already done #3099

PR #3102, opened 2017-03-04
Add missing definition of MPI_T_PVAR_SESSION_NULL (resolve #2652)

### LINKED ISSUE #2652: Add missing MPI_T_PVAR_SESSION_NULL
Open MPI is missing the `MPI_T_PVAR_SESSION_NULL` constant, of type `MPI_T_pvar_session`, defined in MPI-3.0 on page 674 and in MPI-3.1 on page 681.

This is probably as simple as adding
```C
#define MPI_T_PVAR_SESSION_NULL ((MPI_T_pvar_session) 0)
```
under the definitions for `MPI_T_PVAR_HANDLE_NULL` and `MPI_T_CVAR_HANDLE_NULL`.

omor1:
This pull resolves #2652 by adding the missing definition of `MPI_T_PVAR_SESSION_NULL`. It also corrects an issue with `MPI_T_pvar_session_free()` so as to match the man page description.

The only user-visible difference of the pull is that `MPI_T_pvar_session_free()` will now correctly fail with a return value of `MPI_T_ERR_INVALID_SESSION` when the `session` parameter is not valid (that is, when `MPI_T_PVAR_SESSION_NULL  == session`.

### DIFF
--- a/ompi/include/mpi.h.in
+++ b/ompi/include/mpi.h.in
@@ -776,6 +776,7 @@ enum {
  */
 #define MPI_T_PVAR_ALL_HANDLES ((MPI_T_pvar_handle) -1)
 #define MPI_T_PVAR_HANDLE_NULL ((MPI_T_pvar_handle) 0)
+#define MPI_T_PVAR_SESSION_NULL ((MPI_T_pvar_session) 0)
 #define MPI_T_CVAR_HANDLE_NULL ((MPI_T_cvar_handle) 0)
 
 /* MPI-2 specifies that the name "MPI_TYPE_NULL_DELETE_FN" (and all
--- a/ompi/mpi/tool/pvar_session_free.c
+++ b/ompi/mpi/tool/pvar_session_free.c
@@ -23,14 +23,19 @@
 
 int MPI_T_pvar_session_free(MPI_T_pvar_session *session)
 {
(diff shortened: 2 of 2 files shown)


omor1:
Signed-off-by: Omri Mor omri50@gmail.com

omor1:
I'm new to contributing to Open MPI—how do I correctly do the sign-off? Does it need to be in the first comment?

rhc54:
Each commit must have the sign-off in the commit message. What you should probably do here is just "squash" these two commits, and then add the sign-off to the new combined commit message.

In the future, you can do it most easily by just adding "-s" to the "git commit" command and it will add the sign-off for you.

omor1:
Alright, thanks! I had tried to cheap out by using the GitHub editing interface instead of actually cloning the repository, but that was clearly a mistake. It turns out that using the command-line interface is still superior.

hjelmn:
Looks good. This will need to be PRd to v2.0.x, v1.10.x, and v2.x.

omor1:
@hjelmn PRd to v2.0.x, v1.10.x, and v2.x in #3140, #3138, and #3139, respectively.

jsquyres:
Thank you Omri Mor (@omor1) -- especially for PR'ing it to all of our release branches!

PR #3011, opened 2017-02-22
ompi: Avoid unnecessary PMIx lookups when adding procs.

artpol84:
On the medium KNL cluster a linear scaling was observed with this part of the code.
This PR reduces number of lookups from O(n) to O(ppn).
`ompi_proc_complete_init()` was taking 9 ms  on 64 procs and 180 ms on 640 procs. After the fix the delay was fixed around 10ms.

@rhc54 I know you wanted to fix this in some other way but I believe we want this in v2.x for scalability. I'm going to open the PR there.

@hjelmn @jjhursey please consider this while doing your performance tests.

@jladd-mlnx, FYI.

PR #3104, opened 2017-03-05
Fix some minor compatibility issues 

rhc54:
Ensure job-level data gets stored against wildcard rank in the cray, s1, and s2 components, and that the ext1 component translates all wildcard rank requests into the peer's rank since v1.x of PMIx doesn't understand wildcard ranks

Closes #3101

Needs to come over to 2.1 since the change that exposed the issue was already taken there.

### DIFF
--- a/opal/mca/pmix/cray/pmix_cray.c
+++ b/opal/mca/pmix/cray/pmix_cray.c
@@ -5,7 +5,7 @@
  * Copyright (c) 2011-2016 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011-2017 Los Alamos National Security, LLC. All
  *                         rights reserved.
- * Copyright (c) 2013-2015 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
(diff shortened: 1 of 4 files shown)


artpol84:
@rhc54 
```
Needs to come over to 2.1 since the change that exposed the issue was already taken there.
```
This is not correct. In 2.1 we use this proc:
https://github.com/open-mpi/ompi/blob/v2.x/ompi/proc/proc.c#L348:L349

artpol84:
@karasevb, please verify `s1` and `s2` components at runtime with this PR.

rhc54:
@artpol84 Kewl - I didn't know you had used the different name. However, just to be correct, that name should be changed to preserve the abstraction - it should be ```OMPI_PROC_MY_NAME```

artpol84:
Agreed

artpol84:
@rhc54, we did runtime evaluation.

so it looks like `s1` and `s2` components was correct from the start with respect to local peers key. It was only `cray` component that was storing it in the old way.

Also it is funny, but pmix v1.2 doesn't care about wildcard to be passed on his input, the internal logic is written in the way that it will ignore that fact and work correctly. The change ig good tough because otherwise we were breaking API and it was just a luck that pmix v1.2 is working.

hppritcha:
@jsquyres I don't want to merge this till I understand what happened on 2/21/17.  After that point many tests failed with direct launch using aprun.  Even ring_c failed.

artpol84:
@hppritcha https://github.com/open-mpi/ompi/pull/3011 was merged on 02/28/2017 so it should be unrelated.
Why do you want to delay merging this? How can it help to investigate?

hppritcha:
given that cray/pmix specific code was changed in this PR,  and I haven't tried it yet. that's why.

hppritcha:
sorry I thought this was for v2.x since that's what the target label was. don't care if it goes in to master since its already broken for aprun direct launch.

Issue #3098, opened 2017-03-03
sharedfp:lockedfile: add support for multiple jobs using the same file


edgargabriel:
based on a bug report by Nicolas Joly on the mailing: the sharedfp/lockedfile component does not support right now a scenario where multiple jobs read from the same input file, due to a collision of the filenames utilized for the sharedfp handle.

PR #3105, opened 2017-03-06
sharedfp/lockedfile and sm: fix  name collision

edgargabriel:
this fixes the issue reported by Nicolas Joly on the mailing: the sharedfp/lockedfile component does not support right now a scenario where multiple jobs read from the same input file, due to a collision of the filenames utilized for the sharedfp handle. Although not part of the oroginal report, the same occurs for the sharedfp/sm component. Add therefore the jobid to be part of the lockedfilename/sm file name.

use the OMPI_CAST_RTE_NAME macro to determine jobid

Fixes: #3098

### DIFF
--- a/ompi/mca/sharedfp/lockedfile/sharedfp_lockedfile_file_open.c
+++ b/ompi/mca/sharedfp/lockedfile/sharedfp_lockedfile_file_open.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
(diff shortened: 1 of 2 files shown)


edgargabriel:
It is probably more an edge case. The way I would look at it is: if you need to spin another rc candidate anyway, we could integrate this pr as well. Otherwise lets wait for 2.1.1

hppritcha:
@jsquyres I think we should go ahead and merge in since we're waiting on other fixes anyway.

jsquyres:
Per discussion on the Webex today, we're going to merge this because the mallopocalypse is going to force another rc.

PR #3122, opened 2017-03-07
memory/patcher: do not hook madvise

hjelmn:
It is not possible to hook madvise at this time due to a deadlock when
using glibc.


jsquyres:
To cross reference:

* v2.0.x version of this PR: #3124 
* v2.x version of this PR: #3123

Issue #3117, opened 2017-03-06
NetBSD: Runtime error using Open MPI 2.1.0rc2


hppritcha:
@PHHargrove reports hitting a runtime error (using ring_c).c) on NetBSD when using the Open MPI 2.1.0rc2 tarball.  Error is in PMIX.

See https://www.mail-archive.com/devel@lists.open-mpi.org//msg19987.html

rhc54:
Here is the error report:

```
2.1.0rc2 tarball on NetBSD7/amd64.
Configured with only --prefix=... and --disable-mpi-fortran

To get past the lack of a struct timeval definition required a small source change in a previous email.
Once past that, I can build Open MPI and compile the examples.
However, I cannot run them.

Output below.

-Paul

$ mpirun -mca btl sm,self -np 2 examples/ring_c
[netbsd-amd64.kvm:20873] PMIX ERROR: ERROR in file /home/phargrov/OMPI/openmpi-2.1.0rc2-netbsd7-amd64/openmpi-2.1.0rc2/opal/mca/pmix/pmix112/pmix/src/dstore/pmix_esh.c at line 1651
[netbsd-amd64.kvm:20873] PMIX ERROR: OUT-OF-RESOURCE in file /home/phargrov/OMPI/openmpi-2.1.0rc2-netbsd7-amd64/openmpi-2.1.0rc2/opal/mca/pmix/pmix112/pmix/src/dstore/pmix_esh.c at line 820
[netbsd-amd64.kvm:20873] PMIX ERROR: ERROR in file /home/phargrov/OMPI/openmpi-2.1.0rc2-netbsd7-amd64/openmpi-2.1.0rc2/opal/mca/pmix/pmix112/pmix/src/dstore/pmix_esh.c at line 1468
[netbsd-amd64.kvm:20873] PMIX ERROR: ERROR in file /home/phargrov/OMPI/openmpi-2.1.0rc2-netbsd7-amd64/openmpi-2.1.0rc2/opal/mca/pmix/pmix112/pmix/src/server/pmix_server.c at line 592
[netbsd-amd64.kvm:16632] PMIX ERROR: ERROR in file /home/phargrov/OMPI/openmpi-2.1.0rc2-netbsd7-amd64/openmpi-2.1.0rc2/opal/mca/pmix/pmix112/pmix/src/dstore/pmix_esh.c at line 349
[...]

PR #3127, opened 2017-03-08
Fix a problem that surfaced on NetBSD/AMD64. 

rhc54:
Just because HAVE_POSIX_FALLOCATE is defined doesn't mean the function actually works, so fall back to the ftruncate code path when it fails.

Thanks to Paul Hargrove for the patch

### DIFF
--- a/opal/mca/pmix/pmix112/pmix/src/sm/pmix_mmap.c
+++ b/opal/mca/pmix/pmix112/pmix/src/sm/pmix_mmap.c
@@ -3,6 +3,7 @@
  *                         All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


rhc54:
Fixes #3117

artpol84:
@rhc54 as I stated in we are going to open the PR.
Give me 1 hr.

rhc54:
Already done

artpol84:
I see, but I want it little bit differently

artpol84:
@karasevb was going to add additional fix and verification.

artpol84:
But looks like holidays prevented this to happen.

artpol84:
Btw, isn't it going to be a PR to PMIx first?

rhc54:
It's over there too

rhc54:
OMPI v2.1 is trying to get out the door and this either holds it up, or will be delayed to 2.1.1

artpol84:
I need 1 hr, it won't prevent anything.

artpol84:
https://github.com/pmix/pmix/pull/328#event-991712589

rhc54:
Then please _do_ it 😄

artpol84:
As I've said @karasevb was doing that + runtime verification.

PR #3149, opened 2017-03-13
Improve multi-threaded RMA performance of the ugni btl

hjelmn:
This PR brings over work done to improve the MPI_THREAD_MULTIPLE performance of the btl. The improvements include:

 - Use multiple ugni device contexts. The number of contexts depends on the number of local processes and a number of other factors. RMA operations are spread across every available device context but SMSG operations are not. This limitation may be lifted in the future.

 - Added a wrapper for the ugni functionality we need. This wrapper currently uses a spin-lock to decrease overall operation latency.

 - Added support to modify the communication domain settings. This allows users to disable features like FMA sharing if they are not needed.


hjelmn:
@hppritcha If you get a chance please take a look. These changes have been fully tested on trinitite and tr2 both with haswell and knl. The performance improvement for multi-threaded RMA is significant.

hjelmn:
@regrant This is the version of the ugni btl used to collect the RMA-MT data.

hjelmn:
Need to clean up some merge errors.

hppritcha:
@hjelmn I'd suggest squashing this down.

hjelmn:
@hppritcha Ok. Was debating whether to do that or not. There are a couple I want to keep isolated but I will squash the rest.

hjelmn:
@hppritcha Should be good now. The current form will be much easier to back out of the v3.0 branch if any issues are discovered.

hjelmn:
A couple of these commits are also mean for fixing bugs that are present in v2.x.

hjelmn:
Killed travis OS X. No point in letting that run.

PR #3169, opened 2017-03-14
More btl/ugni updates

hjelmn:


hjelmn:
Travis is taking forever and is pointless for this PR.

PR #3170, opened 2017-03-14
Ensure the backend daemons know if we are in a managed allocation and if the HNP was included in the allocation

### LINKED ISSUE #3147: ORTE:hangs observed at finalize using PBS/Torque
A user is reporting a problem with both master and v2.1.0rc3 on their cluster system.  The failure signature is a hang in ```MPI_Finalize``` when running any MPI program, including hello_c.c from examples. The user was so kind as to do a git bisect and narrowed the failure down to this commit:

https://github.com/open-mpi/ompi/commit/48fc33971870dd73aa8db1ed51466df813a641eb

The user noticed that on systems running SLURM that he is unable to reproduce the problem.

rhc54:
Fixes #3147 


hppritcha:
does this need to go in to 3.x?

rhc54:
Yes, if you branched prior to it being committed

hppritcha:
looks like it just made it in!

PR #3181, opened 2017-03-15
ompi: Avoid unnecessary PMIx lookups when adding procs (step 2).

artpol84:
Follow-up for 717f3fef62b193845e9add5aaaae3543c2f2ebfb.

### DIFF
--- a/ompi/proc/proc.c
+++ b/ompi/proc/proc.c
@@ -116,6 +116,8 @@ static int ompi_proc_allocate (ompi_jobid_t jobid, ompi_vpid_t vpid, ompi_proc_t
     opal_hash_table_set_value_ptr (&ompi_proc_hash, &proc->super.proc_name, sizeof (proc->super.proc_name),
                                    proc);
 
+    /* by default we consider process to be remote */
+    proc->super.proc_flags = OPAL_PROC_NON_LOCAL;
     *procp = proc;
 
     return OMPI_SUCCESS;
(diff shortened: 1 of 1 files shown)


artpol84:
As it was suggested in #3011 this PR removes some unnecessary overhead during ompi add proc.

I consider this as a perf bug and we would like to have it in v2.1.1.

ompi/proc/proc.c
```diff
@@ -329,46 +305,71 @@ static int ompi_proc_compare_vid (opal_list_item_t **a, opal_list_item_t **b)
...
+        for (ompi_vpid_t i = 0 ; i < ompi_process_info.num_procs ; ++i ) {
+            opal_process_name_t proc_name;
+            proc_name.jobid = OMPI_PROC_MY_NAME->jobid;
+            proc_name.vpid = i;
+            (void) ompi_proc_for_name (proc_name);
         }
     }
 
+    /* TODO: do we need to lock here ? */
+    opal_mutex_lock (&ompi_proc_lock);
```
rhc54: I think you do need the lock in order to sort the proc list. However, I would suggest removing the unlock just above this loop and remaining locked until the end. The cost of unlock/relock likely outweighs any benefit as we only execute that loop if we are below the cutoff, and the cutoff defaults to zero (if that isn't true yet, then please ask that we put it on the telecon agenda).

ompi/proc/proc.c
```diff
@@ -329,46 +305,71 @@ static int ompi_proc_compare_vid (opal_list_item_t **a, opal_list_item_t **b)
...
+    /* if cutoff is larger than # of procs - add all processes
+     * NOTE that local procs will be automatically skipped as they
+     * are already in the hash table
+     */
+    if (ompi_process_info.num_procs < ompi_add_procs_cutoff) {
+        for (ompi_vpid_t i = 0 ; i < ompi_process_info.num_procs ; ++i ) {
+            opal_process_name_t proc_name;
+            proc_name.jobid = OMPI_PROC_MY_NAME->jobid;
+            proc_name.vpid = i;
+            (void) ompi_proc_for_name (proc_name);
```
artpol84: @rhc54 we need to unlock because `ompi_proc_for_name` is locking internally.

ompi/proc/proc.c
artpol84: I believe that here the cost of the lock/unlock is negligible - no contention.

rhc54: I agree that the probability of contention is low, and shouldn't exist during initial launch (might occur later during a connect/accept operation), so I think the issue is minor. I was only pointing it out as something minor to consider. One thing you could do, though, is move the unlock to right after the ```if``` statement, and then relock before leaving that clause. This way, you'd only unlock/relock if you actually need to do so.

Again, minor point, so I'm okay if you'd rather not do it.

artpol84: Good point!
will update now.

Issue #3003, opened 2017-02-17
CPU frequency-dependent timer issues starting with 2.0.2


AndreasKempfNEC:
Dear Open MPI team,

A few days ago, my colleague Daniel Tameling noticed severe performance issues when running the HPCC benchmark with Open MPI. After spending quite some time tracking down the reason, we suspect that a regression was introduced between Open MPI 2.0.1 and 2.0.2. More specifically: the Open MPI 2.0.1 release tarball seems to be okay and the 2.0.2 release shows issues that persist through the openmpi-v2.0.x-201702170256-5fa504b nightly build.

The issue is that the new releases seem to be severely affected by the CPU frequency set using the acpi-cpufreq driver. When the "ondemand" governor is active and set to allow frequencies between 1.20 GHz and 2.40 GHz, the performance difference between Open MPI 2.0.1 and versions > 2.0.2 is almost a factor of two. Only when the governor "userspace" is used to pin the frequency to the maximum+turbo, the two versions show similar performance.

It does not seem to depend on the PML, BTL, MTL or even fabric. We tested FDR, EDR, openib, mxm, ob1, yalla, cm (on IB with Slurm), and cm, psm2 (on OPA with PBS).

The following latencies were measured on 2 nodes, 2 sockets Intel Xeon E5-2680v4 connected using InfiniBand EDR:

ompi-2.0.3-5fa504b, "ondemand" at 1.20 GHz-2.40 GHz:

```
$ mpirun -n 2 -mca pml yalla -mca rmaps_dist_device mlx5_0:1 -mca coll_hcoll_enable 0 -x MXM_IB_PORTS=mlx5_0:1 -x MXM_TLS=rc,self,shm -mca rmaps_base_mapping_policy dist:span -map-by node --report-bindings bash -c 'ulimit -s 10240; ~/opt/osu-5.3-ompi2/libexec/osu-micro-benchmarks/mpi/pt2pt/osu_latency'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   WARNING:

You should always run with libnvidia-ml.so that is installed with your NVIDIA Display Driver. By default it's installed in /usr/lib and /usr/lib64. libnvidia-ml.so in TDK package is a stub library that is attached only for build purposes (e.g. machine that you build your application doesn't have to have Display Driver installed).
[...]

PR #3184, opened 2017-03-16
 mpi/c: Force wtick/wtime to use gettimeofday

jjhursey:
Workaround for Issue #3003
 * Some relatively minor changes to the timer/linux component
 * Force `wtick`/`wtime` to use `gettimeofday` instead of the high precision timers until we figure out a portable way to calibrate them.

### DIFF
--- a/ompi/mpi/c/wtime.c
+++ b/ompi/mpi/c/wtime.c
@@ -12,6 +12,7 @@
  * Copyright (c) 2006-2014 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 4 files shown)


jsquyres:
@hppritcha @bosilca @hjelmn Please comment / review.

There are 2 trivial cleanup commits for the Linux timer component; the 3rd commit does the temporary fix of nailing WTICK/WTIME down to always use `gettimeofday(3)`, per #3003.

jsquyres:
I think the main question here is: (@bosilca @bwbarrett @hjelmn) How much do we want the `clock_gettime()` preference to `gettimeofday()`?  If so, this PR needs to be amended -- likely including an additional configure check (probably involving `AC_RUN_IFELSE`) for `clock_gettime` on OS X.

ompi/mpi/c/wtime.c
```diff
@@ -40,10 +41,17 @@ double MPI_Wtime(void)
...
+     * more portable solution.
+     */
+#if 0
 #if OPAL_TIMER_CYCLE_NATIVE
     wtime = ((double) opal_timer_base_get_cycles()) / opal_timer_base_get_freq();
 #elif OPAL_TIMER_USEC_NATIVE
     wtime = ((double) opal_timer_base_get_usec()) / 1000000.0;
+#endif
 #else
     /* Fall back to gettimeofday() if we have nothing else */
```
jsquyres: @hjelmn + me suggest putting something like:

```c
#if _LINUX_ && OPAL_HAVE_CLOCK_GETTIME
// ... use clock_gettime(..) ...
#else
// ... use gettimeofday(..) ...
#endif
```

That way, we don't need a new configure test -- we *only* use `clock_gettime` if a) we're on Linux, and b) our existing configure test says that we have `clock_gettime` (i.e., we sidestep the whole OS X problem).

PR #3193, opened 2017-03-16
ompi: Avoid unnecessary PMIx lookups when adding procs (step 2).

artpol84:
Follow-up for 717f3fe.

(cherry-ported from 1f7a3a2d540ac85b2549ea2d94df016c1cbc4ac2)


jsquyres:
After RM discussion, this will go into v2.1.1 and v3.0.0.

PR #3213, opened 2017-03-21
orte/ras: remove loadleveler support

### LINKED ISSUE #3167: Remove loadleveler support
loadleveler is no longer being maintained by IBM and no appears to be willing to support it.  So time for it to be removed as a orte/ras component.  

@gpaulsen

hppritcha:
Remove loadleveler as it is obsolescent and is no longer supported.

Fixes #3167

We'll wait for final check of whether or not loadleveler even
compiles/functions before merging this.


hppritcha:
approved effectively in the ompi-devel call today.

PR #3214, opened 2017-03-21
travis: remove os-x from OS test matrix

hppritcha:


jsquyres:
We have decided to disable OS X Travis testing because it takes sooooooo looooooong.  We have the (LANL) OMPI community OSX testing now, and that should be enough.

PR #3217, opened 2017-03-21
Resolve a race condition for setting our working directory when fork/exec'ing application procs.

rhc54:
We have to ensure we do it after the fork occurs since we want to use multiple threads in the odls. Otherwise, the different threads are bouncing the entire process around.


jsquyres:
@rhc54 Looks like something went wrong in the alps module (per the "LANL-CI" results).

rhc54:
I'm trying to replicate that right now on gadget, but can't - it looks like everything compiles fine over there. Still trying to figure it out

rhc54:
Something is really screwed up in LANL's Jenkins - the orte/odls/alps component is the pre-change one, even though everything else is correctly updated by the PR. That's why it is breaking. I'll try it again in case there is something weird here.

bot:lanl:retest

Issue #3226, opened 2017-03-23
master: large memory leak in osc/pt2pt


jsquyres:
When running just about any of the communication-oriented IBM onesided tests on master (and probably v3.x), there appears to be a gigantic memory leak in the osc/pt2pt component.

For example, if I:

```
$ mpirun --host a,b --mca osc pt2pt c_transpose1
```

and in a separate window, observe `watch free -g`, I'll see the available memory on the node plunge to 0.

This does *not* appear to happen on v2.x (whew!).

jsquyres:
@hjelmn tells me that it is expected that `sm` will fail -- it's only for win_allocated.

jsquyres:
`git bisect` says that https://github.com/open-mpi/ompi/commit/0195d1540190716b5643b181a7e37e4b3a9e1459 is the bad commit:

```
commit 0195d1540190716b5643b181a7e37e4b3a9e1459
Author: Nathan Hjelm <hjelmn@lanl.gov>
Date:   Mon Mar 6 13:58:46 2017 -0700

    osc/pt2pt: flush pending fragments on lock ack

    This commit addresses an issue that can occur in cases where a lot of
    fragments are outstanding.

    Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>
```

clementFoyer:
I just run on an error that seems to be related to the same commit : When calling ompi_osc_pt2pt_frag_alloc it recursively call itself until the stack overflows.

BTW, sorry for the email I sent about this, I should have check here the opened issues before.

jsquyres:
@clementFoyer LOL!  I just replied to your mail on the list pointing to this bug before I refreshed this page and saw that you replied here.  So I'm guilty, too.  😄 

Regardless, it sounds like @hjelmn knows what the fix is and will fix it shortly.

jsquyres:
@clementFoyer It looks like the memory leak was fixed in master, but there's still some other problem.  @hjelmn plans to look at it next week (after SC papers are due!).

jsquyres:
Looks like this has been fixed.

PR #3229, opened 2017-03-23
osc/pt2pt: fix typo

hjelmn:


hppritcha:
is this relevant to an issue or something on the user mail list?

jsquyres:
I confirm that this fixes the memory problem.

However, all the ibm/onesided communication tests now hang (whereas on the v2.x branch, they all complete within a few seconds).

jsquyres:
@hppritcha This fixes the issue with osc/pt2pt described in #3226, but it surfaces another problem (perhaps hidden by the issue described by #3226) -- one sided test programs using osc/pt2pt are now hanging.

jsquyres:
This is needed to fix the memory problem (which is killing Cisco's MTT).  But per my above comment, there's still another problem to fix (the hang).

hppritcha:
Thanks.  It would have been nice for this info to be in the commit message.

PR #3230, opened 2017-03-23
v3.x: If we lose connection to the server after initiating a send/recv in P…

### LINKED ISSUE #3225: PMIx_Abort not always aborting
@rhc54 and I have talked about this on the phone and IM.  There are cases where PMIx_Abort will actually deadlock instead of completing aborting.  Ralph knows what the problem is; he hasn't figured out how to fix it yet.  Filing this issue to track it.

This is causing some MPI and OSHMEM tests to hang on the Cisco MTT cluster, and eat up lots of CPU.

I'm seeing this on master and 2.x (I haven't started testing 3.x yet, but I'm sure the problem is over there, too).

rhc54:
…MIx (e.g., in PMIx_Abort), then we need to "resolve" all pending recvs to avoid hanging.

Fixes #3225

(cherry picked from commit 55e4fba5f5633d08b283c247301c17926c4f5b6b)

### DIFF
--- a/opal/mca/pmix/pmix2x/pmix/src/client/pmix_client.c
+++ b/opal/mca/pmix/pmix2x/pmix/src/client/pmix_client.c
@@ -492,7 +492,7 @@ PMIX_EXPORT pmix_status_t PMIx_Abort(int flag, const char msg[],
     pmix_buffer_t *bfr;
     pmix_cmd_t cmd = PMIX_ABORT_CMD;
     pmix_status_t rc;
-    pmix_ptl_sr_t cb;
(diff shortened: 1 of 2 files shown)


jsquyres:
This seems to have fixed at least some of the Cisco MTT hangs.  There might be more to it, though (per https://github.com/open-mpi/ompi/issues/3225#issuecomment-289894391).

jsquyres:
Mellanox Jenkins is failing all tests due to a git error right now.

@Di0gen @jladd-mlnx @artpol84 Can you please check out what's going on?  Thanks.

PR #3233, opened 2017-03-24
QL - Disable nvml for embedded hwloc

rfehren:
This is to avoid annoying warning messages like
"NVIDIA: no NVIDIA devices found"
when compiled with CUDA support and running on machines
without Nvidia Hardware.


rfehren:
After further long analysis, I found this doesn't fix the problem. The problem is in a change of libcudart when going from version 7.5 (no such messages) to 8.0 (messages appear) and needs to be fixed upstream from Nvidia.

PR #3235, opened 2017-03-24
Stop segfault in BTL/TCP

rhc54:
Correctly check the sa_family and cast the data correctly before passing it to inet_nop, and don't be quite as fancy with the pointer arithmetic as the combination was causing us to segfault every time this debug message was called.


jsquyres:
We shouldn't ever get a BTL TCP connection request from a peer that we know from an IP address that we don't know, right?  I think that's why George had this message coming out unconditionally.

rhc54:
Well, your MTT is currently getting one every time it runs with the TCP BTL - which is why it segfaults every time. I don't care if we leave it "on" all the time - I can just stop the segfault if that's the preference.

rhc54:
done - "on" all the time, but not segfaulting

PR #3245, opened 2017-03-28
mca/base: accept y and n for bool and auto bool enumerator

hjelmn:


hjelmn:
@jsquyres Should address your concerns about the auto_bool PR.

PR #3249, opened 2017-03-29
Use the correct callback data - the callback function was expecting a…

### LINKED ISSUE #3225: PMIx_Abort not always aborting

rhc54:
… bool*, not a pmix_ptl_sr_t*.

Fixes #3225 


jsquyres:
Going to merge this before CI completes so that it makes it into the nightly tarball.  We'll check MTT tomorrow.

PR #3258, opened 2017-03-29
symbol name pollution

markalle:
Adding a test into "make check" that uses "nm" to look at the symbol names in the main libraries (libmpi, libmpi_fh, libmpi_usempi, libopen-rte, libopen-pal) that user apps are directly/indirectly linked against. It expects everything to have some prefix like ompi_, opal_, orte_ etc.

The second commit in this pull request adds "ompi_" onto a handful of symbols the test identified as bad. There are still a couple categories I didn't change that I found iffy:  mca_*, netpatterns_*, and a few that seem related to libevent. I left those pieces untouched and for now have them being accepted by the testcase.

The third commit is just a minor enhancement to update-my-copyright.pl to add a --manual-list=file option.

### DIFF
--- a/ompi/mpi/fortran/mpif-h/register_datarep_f.c
+++ b/ompi/mpi/fortran/mpif-h/register_datarep_f.c
@@ -12,6 +12,7 @@
  * Copyright (c) 2007-2012 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -95,12 +96,12 @@ typedef struct intercept_extra_state {
     ompi_mpi2_fortran_datarep_conversion_fn_t *write_fn_f77;
     ompi_mpi2_fortran_datarep_extent_fn_t *extent_fn_f77;
     MPI_Aint *extra_state_f77;
-} intercept_extra_state_t;
+} ompi_intercept_extra_state_t;
 
-OBJ_CLASS_DECLARATION(intercept_extra_state_t);
+OBJ_CLASS_DECLARATION(ompi_intercept_extra_state_t);
 
 #if !OMPI_BUILD_MPI_PROFILING || OPAL_HAVE_WEAK_SYMBOLS
-static void intercept_extra_state_constructor(intercept_extra_state_t *obj)
+static void intercept_extra_state_constructor(ompi_intercept_extra_state_t *obj)
 {
     obj->read_fn_f77 = NULL;
     obj->write_fn_f77 = NULL;
@@ -108,7 +109,7 @@ static void intercept_extra_state_constructor(intercept_extra_state_t *obj)
     obj->extra_state_f77 = NULL;
 }
 
-OBJ_CLASS_INSTANCE(intercept_extra_state_t,
+OBJ_CLASS_INSTANCE(ompi_intercept_extra_state_t,
                    opal_list_item_t,
                    intercept_extra_state_constructor, NULL);
 #endif  /* !OMPI_BUILD_MPI_PROFILING */
@@ -137,10 +138,10 @@ void ompi_register_datarep_f(char *datarep,
     char *c_datarep;
     int c_ierr, ret;
     MPI_Datarep_conversion_function *read_fn_c, *write_fn_c;
-    intercept_extra_state_t *intercept;
+    ompi_intercept_extra_state_t *intercept;
 
     /* Malloc space for the intercept callback data */
-    intercept = OBJ_NEW(intercept_extra_state_t);
+    intercept = OBJ_NEW(ompi_intercept_extra_state_t);
     if (NULL == intercept) {
         c_ierr = OMPI_ERRHANDLER_INVOKE(MPI_FILE_NULL,
                                         OMPI_ERR_OUT_OF_RESOURCE, FUNC_NAME);
@@ -210,8 +211,8 @@ static int read_intercept_fn(void *userbuf, MPI_Datatype type_c, int count_c,
 {
     MPI_Fint ierr, count_f77 = OMPI_FINT_2_INT(count_c);
     MPI_Fint type_f77 = PMPI_Type_c2f(type_c);
-    intercept_extra_state_t *intercept_data =
-        (intercept_extra_state_t*) extra_state;
+    ompi_intercept_extra_state_t *intercept_data =
+        (ompi_intercept_extra_state_t*) extra_state;
 
     intercept_data->read_fn_f77((char *) userbuf, &type_f77, &count_f77, (char *) filebuf,
                                 &position, intercept_data->extra_state_f77,
@@ -228,8 +229,8 @@ static int write_intercept_fn(void *userbuf, MPI_Datatype type_c, int count_c,
 {
     MPI_Fint ierr, count_f77 = OMPI_FINT_2_INT(count_c);
     MPI_Fint type_f77 = PMPI_Type_c2f(type_c);
-    intercept_extra_state_t *intercept_data =
-        (intercept_extra_state_t*) extra_state;
+    ompi_intercept_extra_state_t *intercept_data =
+        (ompi_intercept_extra_state_t*) extra_state;
 
     intercept_data->write_fn_f77((char *) userbuf, &type_f77, &count_f77, (char *) filebuf,
                                  &position, intercept_data->extra_state_f77,
@@ -244,8 +245,8 @@ static int extent_intercept_fn(MPI_Datatype type_c, MPI_Aint *file_extent_f77,
                                void *extra_state)
 {
     MPI_Fint ierr, type_f77 = PMPI_Type_c2f(type_c);
-    intercept_extra_state_t *intercept_data =
-        (intercept_extra_state_t*) extra_state;
+    ompi_intercept_extra_state_t *intercept_data =
+        (ompi_intercept_extra_state_t*) extra_state;
 
     intercept_data->extent_fn_f77(&type_f77, file_extent_f77,
                                  intercept_data->extra_state_f77, &ierr);
--- a/ompi/patterns/net/netpatterns.h
+++ b/ompi/patterns/net/netpatterns.h
@@ -3,6 +3,7 @@
  * Copyright (c) 2009-2012 Oak Ridge National Laboratory.  All rights reserved.
  * Copyright (c) 2012      Los Alamos National Security, LLC.
  *                         All rights reserved.
+  * Copyright (c) 2017      IBM Corporation. All rights reserved.
   * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -20,21 +21,21 @@
 
 BEGIN_C_DECLS
 
-int netpatterns_base_err(const char* fmt, ...);
-int netpatterns_register_mca_params(void);
+int ompi_netpatterns_base_err(const char* fmt, ...);
+int ompi_netpatterns_register_mca_params(void);
 
 #if OPAL_ENABLE_DEBUG
-extern int netpatterns_base_verbose; /* disabled by default */
-OMPI_DECLSPEC extern int netpatterns_base_err(const char*, ...) __opal_attribute_format__(__printf__, 1, 2);
+extern int ompi_netpatterns_base_verbose; /* disabled by default */
+OMPI_DECLSPEC extern int ompi_netpatterns_base_err(const char*, ...) __opal_attribute_format__(__printf__, 1, 2);
 #define NETPATTERNS_VERBOSE(args)                                \
     do {                                                         \
-        if(netpatterns_base_verbose > 0) {           \
-            netpatterns_base_err("[%s]%s[%s:%d:%s] ",\
+        if(ompi_netpatterns_base_verbose > 0) {           \
+            ompi_netpatterns_base_err("[%s]%s[%s:%d:%s] ",\
                     ompi_process_info.nodename,                  \
                     OMPI_NAME_PRINT(OMPI_PROC_MY_NAME),          \
                     __FILE__, __LINE__, __func__);               \
-            netpatterns_base_err args;               \
-            netpatterns_base_err("\n");              \
+            ompi_netpatterns_base_err args;               \
+            ompi_netpatterns_base_err("\n");              \
         }                                                        \
     } while(0);
 #else
@@ -121,24 +122,24 @@ netpatterns_narray_knomial_tree_node_t;
 
 
 /* Init code for common_netpatterns */
-OMPI_DECLSPEC int netpatterns_init(void);
+OMPI_DECLSPEC int ompi_netpatterns_init(void);
 
 /* setup an n-array tree */
-OMPI_DECLSPEC int netpatterns_setup_narray_tree(int tree_order, int my_rank, int num_nodes,
+OMPI_DECLSPEC int ompi_netpatterns_setup_narray_tree(int tree_order, int my_rank, int num_nodes,
         netpatterns_tree_node_t *my_node);
 /* setup an n-array tree with k-nomial levels */
-OMPI_DECLSPEC int netpatterns_setup_narray_knomial_tree( int tree_order, int my_rank, int num_nodes,
+OMPI_DECLSPEC int ompi_netpatterns_setup_narray_knomial_tree( int tree_order, int my_rank, int num_nodes,
         netpatterns_narray_knomial_tree_node_t *my_node);
 /* cleanup an n-array tree setup by the above function */
-OMPI_DECLSPEC void netpatterns_cleanup_narray_knomial_tree (netpatterns_narray_knomial_tree_node_t *my_node);
+OMPI_DECLSPEC void ompi_netpatterns_cleanup_narray_knomial_tree (netpatterns_narray_knomial_tree_node_t *my_node);
 
 /* setup an multi-nomial tree - for each node in the tree
  *  this returns it's parent, and it's children
  */
-OMPI_DECLSPEC int netpatterns_setup_multinomial_tree(int tree_order, int num_nodes,
+OMPI_DECLSPEC int ompi_netpatterns_setup_multinomial_tree(int tree_order, int num_nodes,
         netpatterns_tree_node_t *tree_nodes);
 
-OMPI_DECLSPEC int netpatterns_setup_narray_tree_contigous_ranks(int tree_order,
+OMPI_DECLSPEC int ompi_netpatterns_setup_narray_tree_contigous_ranks(int tree_order,
         int num_nodes, netpatterns_tree_node_t **tree_nodes);
 
 /* calculate the nearest power of radix that is equal to or greater
--- a/opal/dss/dss_open_close.c
+++ b/opal/dss/dss_open_close.c
@@ -14,6 +14,7 @@
  * Copyright (c) 2014-2016 Intel, Inc. All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -43,7 +44,7 @@ static opal_dss_buffer_type_t default_buf_type = OPAL_DSS_BUFFER_NON_DESC;
 /* variable group id */
 static int opal_dss_group_id = -1;
 
-mca_base_var_enum_value_t buffer_type_values[] = {
+static mca_base_var_enum_value_t buffer_type_values[] = {
     {OPAL_DSS_BUFFER_NON_DESC, "non-described"},
     {OPAL_DSS_BUFFER_FULLY_DESC, "described"},
     {0, NULL}
--- a/opal/mca/event/libevent2022/libevent/evthread-internal.h
+++ b/opal/mca/event/libevent2022/libevent/evthread-internal.h
@@ -47,55 +47,55 @@ struct event_base;
 #if ! defined(_EVENT_DISABLE_THREAD_SUPPORT) && defined(EVTHREAD_EXPOSE_STRUCTS)
 /* Global function pointers to lock-related functions. NULL if locking isn't
    enabled. */
-extern struct evthread_lock_callbacks _evthread_lock_fns;
-extern struct evthread_condition_callbacks _evthread_cond_fns;
-extern unsigned long (*_evthread_id_fn)(void);
-extern int _evthread_lock_debugging_enabled;
+extern struct evthread_lock_callbacks ompi__evthread_lock_fns;
+extern struct evthread_condition_callbacks ompi__evthread_cond_fns;
+extern unsigned long (*ompi__evthread_id_fn)(void);
+extern int ompi__evthread_lock_debugging_enabled;
 
 /** Return the ID of the current thread, or 1 if threading isn't enabled. */
 #define EVTHREAD_GET_ID() \
-	(_evthread_id_fn ? _evthread_id_fn() : 1)
+	(ompi__evthread_id_fn ? ompi__evthread_id_fn() : 1)
 
 /** Return true iff we're in the thread that is currently (or most recently)
  * running a given event_base's loop. Requires lock. */
 #define EVBASE_IN_THREAD(base)				 \
-	(_evthread_id_fn == NULL ||			 \
-	(base)->th_owner_id == _evthread_id_fn())
+	(ompi__evthread_id_fn == NULL ||			 \
+	(base)->th_owner_id == ompi__evthread_id_fn())
 
 /** Return true iff we need to notify the base's main thread about changes to
  * its state, because it's currently running the main loop in another
  * thread. Requires lock. */
 #define EVBASE_NEED_NOTIFY(base)			 \
-	(_evthread_id_fn != NULL &&			 \
+	(ompi__evthread_id_fn != NULL &&			 \
 	    (base)->running_loop &&			 \
-	    (base)->th_owner_id != _evthread_id_fn())
+	    (base)->th_owner_id != ompi__evthread_id_fn())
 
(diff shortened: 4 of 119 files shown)


rhc54:
Ummm...some of these changes aren't correct. The symbol should have just been made ```static``` instead of being left global and prefixed.

markalle:
Ah yes, in general I erred on the side of leaving things global and adding prefixes. There were only a couple symbols where I switched it to static. I figured this way is less error-prone, but it is possible to add more statics instead.

Would you still go static for symbols that exist in two files, foo.h and foo.c? To my eye that would be a symbol that could potentially later be used from another file even if it's only used from one file now.

ompi/mpi/fortran/mpif-h/register_datarep_f.c
```diff
@@ -89,28 +90,28 @@ static int extent_intercept_fn(MPI_Datatype type_c, MPI_Aint *file_extent,
 /* Data structure passed to the intercepts (see below).  It is an OPAL
    list_item_t so that we can clean this memory up during
    MPI_FINALIZE.  */
-typedef struct intercept_extra_state {
+typedef struct ompi_intercept_extra_state {
```
jsquyres: I'm a little surprised that this type name would be exposed externally.  Are we using it outside of this file?

opal/dss/dss_open_close.c
```diff
@@ -43,7 +44,7 @@ static opal_dss_buffer_type_t default_buf_type = OPAL_DSS_BUFFER_NON_DESC;
 /* variable group id */
 static int opal_dss_group_id = -1;
 
-mca_base_var_enum_value_t buffer_type_values[] = {
+mca_base_var_enum_value_t opal_buffer_type_values[] = {
```
jsquyres: This one can probably also be static.

opal/threads/wait_sync.h
```diff
@@ -75,7 +76,7 @@ typedef struct ompi_wait_sync_t {
         (sync)->signaling = false;                    \
 }
 
-OPAL_DECLSPEC int sync_wait_mt(ompi_wait_sync_t *sync);
+OPAL_DECLSPEC int opal_sync_wait_mt(ompi_wait_sync_t *sync);
 static inline int sync_wait_st (ompi_wait_sync_t *sync)
```
jsquyres: Just for symmetry -- even if the `_st` symbol isn't actually a problem -- should probably make both the `_mt` and `_st` versions be prefixed.

orte/mca/ess/base/ess_base_std_orted.c
```diff
@@ -89,7 +90,7 @@ static opal_event_t int_handler;
 static opal_event_t epipe_handler;
 static opal_event_t sigusr1_handler;
 static opal_event_t sigusr2_handler;
-char *log_path = NULL;
+char *orte_log_path = NULL;
```
jsquyres: Might be ok for this one to be static, too.

orte/util/show_help.c
```diff
@@ -79,13 +80,13 @@ typedef struct {
     int tli_count_since_last_display;
     /* Do we want to display these? */
     bool tli_display;
-} tuple_list_item_t;
+} opal_tuple_list_item_t;
```
jsquyres: Probably ok to be static.

test/symbol_name/nmcheck_prefix
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Copyright (c) 2017      IBM Corporation. All rights reserved.
+
+
+# if there's no perl, skip the test
+perl -v > /dev/null 2>&1
+if [ $? -ne 0 ] ; then exit 0 ; fi
```
jsquyres: Probably better to exit 77 instead of 0.  77 = test was skipped.

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/perl
```
jsquyres: Probably better to

```sh
#!/usr/bin/env perl
```

this will let it find perl in the path.  Slightly more portable this way.

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
...
+#     ompi/.libs/lib<mpi>.so
+#     ompi/mpi/fortran/mpif-h/.libs/lib<mpi>_mpifh.so
+#     ompi/mpi/fortran/use-mpi-tkr/.libs/lib<mpi>_usempi.so
+#     orte/.libs/libopen-rte.so
+#     opal/.libs/libopen-pal.so
+# but I hate to assume those are the locations, so I'll use 'find' and
+# test whatever ".so"s are found.
+
+@libs = ();
+$mpi = $ENV{OMPI_LIBMPI_NAME};
```
jsquyres: Might want to have a default value for `$mpi` here (or perhaps a warning/abort), just in case someone runs this test outside of `make check`.

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
...
+    @symbols = grep(!/^pmix_/i, @symbols);
+    @symbols = grep(!/^mpit_/, @symbols);
+    @symbols = grep(!/^ompit_/, @symbols);
+    @symbols = grep(!/^MPIR_/, @symbols);
+    @symbols = grep(!/^MPIX_/, @symbols);
+    @symbols = grep(!/^mpidbg_dll_locations$/, @symbols);
+    @symbols = grep(!/^mpimsgq_dll_locations$/, @symbols);
+
+    @symbols = grep(!/^mca_/, @symbols); # I'm tempted to call these bad
+    @symbols = grep(!/^netpatterns_/, @symbols); # I'm tempted to call these bad
```
jsquyres: `netpatterns_` should definitely be bad.

I agree that `mca_` is borderline (it unfortunately has its origins at the beginning of the project, and is likely deeply embedded throughout the code base).

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
...
+
+    # grep to get rid of symbol prefixes that are considered acceptable,
+    # leaving behind anything bad:
+    @symbols = grep(!/^ompi_/i, @symbols);
+    @symbols = grep(!/^opal_/i, @symbols);
+    @symbols = grep(!/^orte_/i, @symbols);
+    @symbols = grep(!/^orted_/i, @symbols);
+    @symbols = grep(!/^mpi_/i, @symbols);
+    @symbols = grep(!/^pmpi_/i, @symbols);
+    @symbols = grep(!/^pmix_/i, @symbols);
```
jsquyres: Is there a reason it's ok to expose `pmix_` public symbols?

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
...
+    @symbols = grep(!/^mpit_/, @symbols);
+    @symbols = grep(!/^ompit_/, @symbols);
```
jsquyres: What symbols are `mpit_` or `ompit_`?

test/symbol_name/nmcheck_prefix.pl
```diff
@@ -0,0 +1,112 @@
...
+
+    # This is another section I'm not confident about, I'm not
+    # familiar with libevent so for now I'm giving everything related
+    # to it a pass. I'd like to change all these to opal_event_* but
+    # not without confirmation from someone else first.
+    @symbols = grep(!/^_event_debug_mode_on$/, @symbols);
+    @symbols = grep(!/^event_enable_debug_output$/, @symbols);
+    @symbols = grep(!/^event_global_current_base_$/, @symbols);
+    @symbols = grep(!/^event_module_include$/, @symbols);
+    @symbols = grep(!/^eventops$/, @symbols);
```
jsquyres: @rhc54 Can you confirm: I think that these should probably all be prefixed with `opal_`.  I.e., we shouldn't be exposing any public symbols from libevent that aren't prefixed.

opal/mca/event/libevent2022/libevent/evthread-internal.h
```diff
@@ -47,55 +47,55 @@ struct event_base;
...
 /* Global function pointers to lock-related functions. NULL if locking isn't
    enabled. */
-extern struct evthread_lock_callbacks _evthread_lock_fns;
-extern struct evthread_condition_callbacks _evthread_cond_fns;
-extern unsigned long (*_evthread_id_fn)(void);
-extern int _evthread_lock_debugging_enabled;
+extern struct evthread_lock_callbacks opal_evthread_lock_fns;
+extern struct evthread_condition_callbacks opal_evthread_cond_fns;
+extern unsigned long (*opal_evthread_id_fn)(void);
+extern int opal_evthread_lock_debugging_enabled;
```
jsquyres: You should probably put all libevent-based renamings in `opal/mca/event/libevent2022/libevent/opal_rename.h`.

ompi/mpi/fortran/mpif-h/register_datarep_f.c
markalle: I'm pretty sure that one actually came from
    OBJ_CLASS_DECLARATION(intercept_extra_state_t);
and maybe I could have done a more targeted change but I figured all the OBJ_ macros probably work together and would all likely need the same name change. Another option is to fix it at the OBJ_CLASS_DECLARATION macro:

#define OBJ_CLASS_DECLARATION(NAME)             \
    extern opal_class_t NAME ## _class

maybe that should be extern opal_class_t opal_ ## NAME ## _class

markalle: At a glance, I think OBJ_CLASS OBJ_CLASS_INSTANCE and OBJ_CLASS_DECLARATION could all get an opal_ prefix in opal_object.h and then I think they'd all match

rhc54: that would seem really weird - so instead of just putting "static" in front of the OBJ_CLASS_foo macro, we would put "opal_" in front of every class, regardless of what layer they are declared in?

test/symbol_name/nmcheck_prefix.pl
markalle: Looks like libmpi.so had
    mpit_big_lock
    mpit_init_count
    mpit_lock
    mpit_unlock
    mpit_big_lock
    mpit_init_count
    mpit_lock
    mpit_unlock
    ompit_opal_to_mpit_error
    ompit_var_type_to_datatype
so, looks like half of them are in some way thread related, so maybe I'll rename them to ompi_td_big_lock etc.

jsquyres: I wonder if those are just poorly named.  Perhaps they should all be `ompi_`.  I.e., they have to do with the MPI_T implementation inside the OMPI layer.

markalle: Oh, and I guess some of these are MPI_T_ related as well. So we could claim mpit_* for that, or name them mpi_t_*. I don't really like how mpi_t_* looks like a fortran entrypoint, I'd kind of prefer all mpi_* be actual MPI calls. But all of MPI_T_ starts with MPI_ already...

jsquyres: Yeah, we lost that battle in the Forum (i.e., the prefix is `MPI_T_`, even though it's ugly).  😦 

But perhaps those `mpit_` symbols should really be `MPI_T_` or `ompi_mpi_t_` (they look like symbols that are internal to OMPI's MPI_T implementation).

markalle:
I think I've addressed all the above.

I didn't aggressively search for things that could have been static, but I changed the ones you mentioned above.

For OBJ_CLASS_*() in opal/class/opal_object.h where it creates a global var I made the var have opal_ prefixed onto its name, so opal_<NAME>_class.  That let me revert some of the changes that had otherwise spiraled into a bunch of changes in some of the OBJ_CLASS_*() users to keep the names consistent.

I added more changes for mpit_, ompit_, and netpatterns_. Those changes touched a lot of files. 

For libevent you were right, the opal_rename.h made that part way easier.

--------
Question: the Travis CI build failed at distcheck saying
make[5]: *** No rule to make target `nmcheck_prefix', needed by `nmcheck_prefix.log'.  Stop.
Is there something wrong with my Makefile.am for test/symbol_name/?

jjhursey:
@markalle Yeah, I think the `Makefile.am` is not quite right. `make distcheck` is failing while trying to find the compilation target, and since it's a script there is nothing to compile. I'll have to refresh my knowledge of how to address that. I can help you take a look at it today - just ping me.

jjhursey:
@markalle Give this patch a try:
```patch
diff --git a/test/symbol_name/Makefile.am b/test/symbol_name/Makefile.am
index ba3f8670..11668a65 100644
--- a/test/symbol_name/Makefile.am
+++ b/test/symbol_name/Makefile.am
@@ -7,6 +7,7 @@
 # $HEADER$
 #
 
-TESTS = nmcheck_prefix
+check_SCRIPTS = nmcheck_prefix nmcheck_prefix.pl
+TESTS = $(check_SCRIPTS)
 
 AM_TESTS_ENVIRONMENT = MYBASE='$(top_builddir)'; OMPI_LIBMPI_NAME=@OMPI_LIBMPI_NAME@; export MYBASE OMPI_LIBMPI_NAME;
```

ompi/patterns/net/netpatterns.h
```diff
@@ -20,21 +21,21 @@
...
 #define NETPATTERNS_VERBOSE(args)                                \
     do {                                                         \
-        if(netpatterns_base_verbose > 0) {           \
-            netpatterns_base_err("[%s]%s[%s:%d:%s] ",\
+        if(ompi_netpatterns_base_verbose > 0) {           \
+            ompi_netpatterns_base_err("[%s]%s[%s:%d:%s] ",\
                     ompi_process_info.nodename,                  \
                     OMPI_NAME_PRINT(OMPI_PROC_MY_NAME),          \
                     __FILE__, __LINE__, __func__);               \
             netpatterns_base_err args;               \
```
jjhursey: Missed one: `ompi_netpatterns_base_err`

ompi/mpi/fortran/mpif-h/register_datarep_f.c
markalle: I wrote a script to check and unless I did something wrong there are 300 occurrences of OBJ_CLASS_INSTANCE(name,,,) and every one of them has an OBJ_NEW(name) that occurs in a different file than the OBJ_CLASS_INSTANCE(). So I think that means they really are all globals.

I don't fully understand the OBJ_*() macros, but it looks like the OBJ_CLASS_INSTANCE() is where the global is created, and OBJ_NEW() is a user of that same global.

markalle: I added a check for OBJ_CLASS(name) as well as OBJ_NEW(name) outside the file where OBJ_CLASS_INSTANCE(name,,,) occurred. Now I'm getting:
150 self contained cases that could be static
17 self contained but used OBJ_CLASS_DELARATION() "extern" for some reason
133 actually span multiple files

As far as coding for this I still think adding an ```opal_``` prefix makes sense. It's what the 133 cases need, the others could in principle be switched to
    static OBJ_CLASS_INSTANCE(name,,,)
and wouldn't strictly need the ```opal_``` prefix, but it wouldn't hurt anything.

rhc54:
My concern is that this PR doesn't feel like the right approach. If we have symbols that weren't correctly prefixed, the solution should be to fix those specific symbols - not to automatically throw additional prefixes on top of _every_ symbol. If you do, then we wind up having to search for symbols like "opal_orte_my_object" in the debugger - which is not only annoying, but non-intuitive.

So why don't we just fix the problems, as we have always done before?

ompi/mpi/fortran/mpif-h/register_datarep_f.c
markalle: I do agree the specific string ```opal_``` is weird though... not sure what to do about that. I don't really want this to turn into a manual name change spanning all those files. Is there a var that indicates ```ompi_ / opal_ / orte_``` etc```?

rhc54:
I would change this PR to take a different route. Simply highlight to the community the symbols that aren't properly prefixed, and ask that the respective code "owners" fix it. This has worked multiple times in the past, and I see no evidence as to why it wouldn't work again. I'm happy to fix the ones in ORTE and relevant parts of OPAL.

If we all just do a little, the problem will be solved. The contribution here is the scripts that report which symbols we missed so we know what needs fixing.

This, IMO, would be a much preferable solution than automatically slapping prefixes on variables, even those that don't need it (which are likely to be the majority).

ompi/patterns/net/netpatterns.h
markalle: I can see how I missed that in the macro, but what's weird is it didn't show up as "U netpatterns_base_err" in the nm output. Anyway I've changed it.

markalle:
I pushed again with the test/symbol_name/Makefile.am change and the netpatterns_ fix Josh pointed out.

I agree with Ralph's concern that putting ```opal_``` in front of everything that uses OBJ_CLASS_*() is weird looking when the users might be ```ompi_``` or ```orte_```. But there are lots of uses of OBJ_CLASS_*() that can't just become static either. It would be a lot of code changes to modify all the names in all those files (and to add static to all the ones that didn't really need to be global).

rhc54:
I disagree with your last statement - it is a trivial thing to do. We have a script that does it in seconds:

```shell
$ search_replace.pl <current_symbol_name> orte_<current_symbol_name>
```
Done. You could even have your symbol checking script execute it. What is so hard???

jjhursey:
I'm having trouble reproducing the CI failures. I did a `make distcheck` and a `make check` on a local checkout of Mark's branch and they passed fine. Here is what I'm running:
```shell
./autogen.pl
./configure --prefix=/tmp/jjhursey/bogus
make -j 20
make distcheck
make check
```

Anyone have an idea about how I might be able to reproduce that so I can take a look at what's wrong with the build.

markalle:
That's fine, a large scale search and replace just requires more review to avoid errors, and is a little more work for people rebasing around this checkin. Without enough review, such a script might accidentally change MCA parameter names if their name happened to include the same string as a variable being changed. I don't think that will happen because of how the mca-var-register function is written, but that's the type of thing I'd worry about for a large search and replace.

For the ```mca_``` family would you rather:
1. leave them ```mca_```
2. change just the globals to ```ompi_/opal_/orte_mca_```
3. change them all, global or not

As an example, ```mca_pml``` is a global whose name would change, but ```mca_pml_cm_component_init``` is a static that doesn't necessarily need to change but could if we want it all to match (and it would be an extra step to keep a global search and replace from hitting the extras like that).

Another question: if we address ```mca_```, should they each get their own prefix based on where they're at ```ompi_/opal_/orte_``` so for example my script might look up the header declaration of each symbol being changed to decide which prefix it should receive, or can all MCA vars get a single prefix.  Right now there's an ```asprintf(&name, "mca_%s_%s",) opal_dl_lookup(,name)``` that wouldn't currently know which prefix to look for in front if it wasn't uniform. I think things would work overall with multiple-prefixes, but it's an extra step to automate the decision of what is the right prefix for every replacement. And although it's unlikely they'd differ, ```mca_foo``` and ```mca_foo_bar``` aren't guaranteed to have the same decision for what prefix they need.

rhc54:
Perhaps you could just post on gist (or a file here, or email) the list of variables you've identified so we can look them over? Not everything has to be automated - we just need to fix this once, and then have an automated method for reporting if/when it gets violated again.

markalle:
Okay, I made two lists:
mcasymbols.global.linked.txt = just the globals, and just from libmpi.so etc (not MCAs)
mcasymbols.all.txt = all libs, MCAs included, all ```mca_``` whether global or static
https://gist.github.com/markalle/164fb4662f0f6c9d97aaf36335636daa

A simple search-and-replace that processes only the strings from the globals list might as a side effect hit a portion of the other symbols that are static or just in MCAs. Do we want the result to be a mix of ```ompi_mca_*``` and ```mca_*``` or do we want to make all the un-needed changes too so things match?

I don't yet have the output ready that includes a guess for each ```mca_``` symbol whether it should have ```ompi_``` or ```opal_``` or ```orte_``` etc.

rhc54:
Well, you cannot rename the ```mca_foo_component``` variables as those are what the MCA system is looking for when opening a dll. So maybe the best answer is to simply declare "mca" to be a reserved prefix and be done with it.

bwbarrett:
I thought mca was already in our list of reserved namespacing?  We've always treated it that way (or at least, did before I left 3 years ago...)

markalle:
I think I'm okay with declaring ```mca_``` to be reserved.

markalle:
I finally got the testcase to run on the LANL-distcheck and travis-ci systems. I don't really feel like I'm using automake correctly, but at least it's working.

Previous attempts at using DIST, EXTRA_DIST, and check_SCRIPTS didn't result in the script nmcheck_prefix.pl being located in the test directory. I used what might be considered a hack, and am launching ${srcdir}/nmcheck_prefix.pl and now it's able to find the test and run it.

hppritcha:
botany bay is having a bad day.  try again.
bot:lanl:retest

ibm-ompi:

 Gist: https://gist.github.com/b5373acbf8ec2301067f9ae48996c582

markalle:
rebased again, some commit had used the old name for one of the changed symbols

I think I've addressed all the concerns, the OBJ_CLASS_*() aren't auto-adding any prefixes, the test is accepting "mca_" as an acceptable prefix. I didn't aggressively look for opportunities to make vars static instead of using the prefixes added in the second commit, but some of it has been made static.

jjhursey:
@markalle What's the state of this PR? Did we decide to go a different route or is this ready to go (once rebased)?

markalle:
I'd like to get this checkin in, but it touches so many files it goes stale fast. I can rebase it again but it'll become stale within a few days after that if we don't merge it right away.

I could use a different approach if people would  be more comfortable: currently this is a single commit that's a combination of changes done by hand (changing vars to static) and scripted name changes. The result I think is too big for people to really review in detail.

I could probably split it out into two commits, one for the by-hand changing which would cover the small-ish number of vars where it was easy to tell it could just become static. Then the scripted changes where I could include the script used.

bwbarrett:
Why is an iterative approach not the right one?  You could do one framework at a time or one library or something that you can get a handle on.  You're going to continue to be unsuccessful with an all-in-one approach.  You also need something more than good intentions if this is important (ie, a test that can be run on every PR/commit).

rhc54:
From what I see, it looks okay to me - but I'd echo Brian's comments. At least split it up by library now that we've agreed on approach, and then get it thru the CI and commit.

markalle:
I see value in splitting manual from scripted changes, but is there really value in splitting a list of symbols for scripted changes into multiple sets?

I'm re-doing it now, so if you want these symbols changes separated out from each other, now's the time. And you want a separate pull request for each as opposed to just separate commits?

(libmpi.so) :

    coll_base_comm_get_reqs
    comm_allgather_pml
    comm_allreduce_pml
    comm_bcast_pml
    fcoll_base_coll_allgather_array
    fcoll_base_coll_allgatherv_array
    fcoll_base_coll_bcast_array
    fcoll_base_coll_gather_array
    fcoll_base_coll_gatherv_array
    fcoll_base_coll_scatterv_array
    fcoll_base_sort_iovec
    mpit_big_lock
    mpit_init_count
    mpit_lock
    mpit_unlock
    netpatterns_base_err
    netpatterns_base_verbose
    netpatterns_cleanup_narray_knomial_tree
    netpatterns_cleanup_recursive_doubling_tree_node
    netpatterns_cleanup_recursive_knomial_allgather_tree_node
    netpatterns_cleanup_recursive_knomial_tree_node
    netpatterns_init
    netpatterns_register_mca_params
    netpatterns_setup_multinomial_tree
    netpatterns_setup_narray_knomial_tree
    netpatterns_setup_narray_tree
    netpatterns_setup_narray_tree_contigous_ranks
    netpatterns_setup_recursive_doubling_n_tree_node
    netpatterns_setup_recursive_doubling_tree_node
    netpatterns_setup_recursive_knomial_allgather_tree_node
    netpatterns_setup_recursive_knomial_tree_node
    ompit_opal_to_mpit_error
    ompit_var_type_to_datatype
    pml_v_output_close
    pml_v_output_open

(libmpi_mpifh.so) :

    intercept_extra_state_t_class

(libopen-rte.so) :

    odls_base_default_wait_local_proc

(libopen-pal.so) :

    _event_debug_mode_on
    _evthread_cond_fns
    _evthread_id_fn
    _evthread_lock_debugging_enabled
    _evthread_lock_fns
    cmd_line_option_t_class
    cmd_line_param_t_class
    crs_base_self_checkpoint_fn
    crs_base_self_continue_fn
    crs_base_self_restart_fn
    event_enable_debug_output
    event_global_current_base_
    event_module_include
    eventops
    sync_wait_mt
    trigger_user_inc_callback
    var_type_names
    var_type_sizes

rhc54:
Truly? I honestly don't care - I was just suggesting a way for you to get things committed without them constantly becoming stale. One thing is certain: waiting for several weeks before circling back to it will never work.

markalle:
Thanks, I'll take another stab or two at the large blob approach since I still think hitting this once is less of a nuisance than hitting it multiple times. But I agree piecemeal would work fine, and if this doesn't go through soon I'll switch to that approach (dropping the 4th commit above, and submitting a handful of smaller pull requests).

markalle:
I only just realized (because of PRBC) that there's an option --disable-dlopen that piles way more symbols into libmpi.so that would have normally been safely partitioned off in various mca_*.so. And possibly --enable-mca-static=name that does so more individually

So I've updated the nmcheck_prefix.pl test to be more generous with symbol name pollution when it appears to be coming from an MCA that was built into one of the main libs that would have normally been its own separate mca_*.so. Otherwise there would be a lot more symbols to change, and I think pollution at the MCA .so level is pretty safe.

gpaulsen:
Lets discuss at Face to face and try to get this merged in before it becomes stale again.

markalle:
Thanks, fwiw the big commit is now completely scripted so it's not too big a deal if it does go stale again, just takes a few minutes to rerun the script. But having said that, yeah, I'd love for it to go in.

PR #3259, opened 2017-03-30
Update how we pass the node regex so we pass _all_ nodes, even those without daemons. 

rhc54:
This allows the backend daemons to form a complete picture of the allocation. Include info on which nodes have daemons on them, and populate that info on the backend as well.


rhc54:
@hjelmn Could you please give this a try? At least on Gadget, it appears to be working correctly. There is still something funny about binding, but let's ensure the placement is correct.

hjelmn:
Ok, will check once the current user hero run is complete.

rhc54:
@hjelmn Ready for another shot

Issue #3255, opened 2017-03-29
Alltoallv small to medium size messages desync, if used in gather-alike manner


DrHaraldBraun:
Hi,

repeated calls to Alltoallv exchanging small to medium size messages causes significant desynchronization of ranks, if used in a gather-alike manner. This leads to a major performance degradation in ompi 2.X.X as compared to ompi 1.X.X. All ompi 2.X Releases and release candidates exhibit this behavior.

The following was observed in a reproducer-run (see zip-archive) on 12 (2x18coreBDW_IB:FDR) nodes using 8 ranks per node -> commsize was 96 ranks.
Ratio>1 means Alltoallv+Barrier is faster than Alltoallv alone.
```
00: MPI_Alltoallv           sparse(gath) avMsgSize=8       B        : total 6.063e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=32      B        : total 5.471e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=128     B        : total 5.310e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=512     B        : total 5.129e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=2048    B        : total 2.597e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=8192    B        : total 2.349e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=32768   B        : total 3.779e+01 sec; 1000 calls
00: MPI_Alltoallv           sparse(gath) avMsgSize=131072  B        : total 3.291e+00 sec; 1000 calls
00: MPI_Alltoallv+Barrier   sparse(gath) avMsgSize=8       B        : total 6.385e-02 sec; 1000 calls
00: MPI_Alltoallv+Barrier   sparse(gath) avMsgSize=32      B        : total 6.465e-02 sec; 1000 calls
00: MPI_Alltoallv+Barrier   sparse(gath) avMsgSize=128     B        : total 6.655e-02 sec; 1000 calls
00: MPI_Alltoallv+Barrier   sparse(gath) avMsgSize=512     B        : total 7.415e-02 sec; 1000 calls
00: MPI_Alltoallv+Barrier   sparse(gath) avMsgSize=2048    B        : total 1.074e-01 sec; 1000 calls
[...]

PR #3262, opened 2017-03-31
coll/sync: add support for alltoallv and alltoallw

DrHaraldBraun:
Refers to / fixes: #3255

Adds `alltoallv` to `coll/sync` analogous to other collectives.
Two additional control variables: barrier_before_alltoallv and
barrier_after_alltoallv analogous to barrier_before and barrier_after,
respectively. This way `sync` may be activated for `alltoallv` only.
However calls to `alltoallv` also increment the other counters.

edit:
and the same for` alltoallw`

### DIFF
--- a/ompi/mca/coll/sync/coll_sync_module.c
+++ b/ompi/mca/coll/sync/coll_sync_module.c
@@ -47,14 +47,17 @@ static void mca_coll_sync_module_construct(mca_coll_sync_module_t *module)
     module->after_num_operations = 0;
     module->before_num_operations_alltoallv = 0;
     module->after_num_operations_alltoallv = 0;
+    module->before_num_operations_alltoallw = 0;
+    module->after_num_operations_alltoallw = 0;
     module->in_operation = false;
 }
 
 static void mca_coll_sync_module_destruct(mca_coll_sync_module_t *module)
 {
-    OBJ_RELEASE(module->c_coll.coll_alltoallv_module);
     if (0 != mca_coll_sync_component.barrier_before_nops ||
         0 != mca_coll_sync_component.barrier_after_nops) {
+        OBJ_RELEASE(module->c_coll.coll_alltoallv_module);
+        OBJ_RELEASE(module->c_coll.coll_alltoallw_module);
         OBJ_RELEASE(module->c_coll.coll_bcast_module);
         OBJ_RELEASE(module->c_coll.coll_gather_module);
         OBJ_RELEASE(module->c_coll.coll_gatherv_module);
@@ -62,6 +65,13 @@ static void mca_coll_sync_module_destruct(mca_coll_sync_module_t *module)
         OBJ_RELEASE(module->c_coll.coll_reduce_scatter_module);
         OBJ_RELEASE(module->c_coll.coll_scatter_module);
         OBJ_RELEASE(module->c_coll.coll_scatterv_module);
+    } else {
+        if (0 != mca_coll_sync_component.barrier_before_nops_alltoallv ||
+            0 != mca_coll_sync_component.barrier_after_nops_alltoallv)
+            OBJ_RELEASE(module->c_coll.coll_alltoallv_module);
+        if (0 != mca_coll_sync_component.barrier_before_nops_alltoallw ||
+            0 != mca_coll_sync_component.barrier_after_nops_alltoallw) 
+            OBJ_RELEASE(module->c_coll.coll_alltoallw_module);
     }
     /* If the exscan module is not NULL, then this was an
        intracommunicator, and therefore scan will have a module as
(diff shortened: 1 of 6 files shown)


hppritcha:
@DrHaraldBraun do you have an app that is helped by having this sync option for alltoallv/w?

hppritcha:
okay to test

hppritcha:
@gpaulsen

edgargabriel:
@hppritcha My understanding of the status of this pr was that we could not identify a technical reason why this patch would be necessary, since Alltoall(v/w) act synchronizing anywyay. That's why we asked for an example from the author where this was an issue.

ompi/mca/coll/sync/coll_sync_module.c
```diff
@@ -45,18 +45,34 @@ static void mca_coll_sync_module_construct(mca_coll_sync_module_t *module)
...
+        OBJ_RELEASE(module->c_coll.coll_alltoallv_module);
+        OBJ_RELEASE(module->c_coll.coll_alltoallw_module);
+        OBJ_RELEASE(module->c_coll.coll_bcast_module);
+        OBJ_RELEASE(module->c_coll.coll_gather_module);
+        OBJ_RELEASE(module->c_coll.coll_gatherv_module);
+        OBJ_RELEASE(module->c_coll.coll_reduce_module);
+        OBJ_RELEASE(module->c_coll.coll_reduce_scatter_module);
+        OBJ_RELEASE(module->c_coll.coll_scatter_module);
+        OBJ_RELEASE(module->c_coll.coll_scatterv_module);
+    } else {
```
gpaulsen: I don't think this should be an 'else' just a close bracket.  Couldn't someone run with both?  Of course we don't want to obj_release twice, so might need to rework the logic a bit.

ompi/mca/coll/sync/coll_sync_module.c
```diff
@@ -114,23 +134,50 @@ mca_coll_sync_comm_query(struct ompi_communicator_t *comm,
...
-    sync_module->super.coll_scatter    = mca_coll_sync_scatter;
-    sync_module->super.coll_scatterv   = mca_coll_sync_scatterv;
-
+    if(0 == mca_coll_sync_component.barrier_before_nops &&
+       0 == mca_coll_sync_component.barrier_after_nops) {
+        sync_module->super.coll_allgather  = NULL;
+        sync_module->super.coll_allgatherv = NULL;
+        sync_module->super.coll_allreduce  = NULL;
+        sync_module->super.coll_alltoall   = NULL;
+        if (0 != mca_coll_sync_component.barrier_before_nops_alltoallv ||
```
gpaulsen: Again, can users only use barrier_[before|after]_nops_alltoall[v|w] if they don't use barrier_before_nops mechanism?

ompi/mca/coll/sync/coll_sync_module.c
```diff
@@ -156,13 +203,26 @@ int mca_coll_sync_module_enable(mca_coll_base_module_t *module,
...
+        CHECK_AND_RETAIN(alltoallv);
+        CHECK_AND_RETAIN(alltoallw);
+        CHECK_AND_RETAIN(bcast);
+        CHECK_AND_RETAIN(gather);
+        CHECK_AND_RETAIN(gatherv);
+        CHECK_AND_RETAIN(reduce);
+        CHECK_AND_RETAIN(reduce_scatter);
+        CHECK_AND_RETAIN(scatter);
+        CHECK_AND_RETAIN(scatterv);
+    } else {
```
gpaulsen: again, perhaps just a close bracket instead of the 'else', and some protection to only CHECK_AND_RETAIN alltoall[v|w] once if it goes through both sections of code.

gpaulsen:
@DrHaraldBraun, are you still interested in this?  Would you mind if I cleaned this up?

awlauria:
@gpaulsen / @DrHaraldBraun do we still want this?

awlauria:
@gpaulsen let's discuss the fate of this for 5.0.

awlauria:
Re-ping. 5.0 branching is targeted for April 30th. If you want this in for 5.0, please target to get it in master by then. Thanks!

gpaulsen:
I'll re-review this and try to get in for v5.0.0

jsquyres:
It looks like this PR was auto-closed when we renamed the `master` branch to `main`.  I'm not sure why GitHub auto-closed this PR due to the renaming -- all other open PRs were automatically re-targeted to `main`. ☹️ 

Can you re-open and/or re-submit this to the `main` branch?

bosilca (commented):
This PR makes the support for sync alltoall[v,w] widely different from the others collectives. I can't see why and there are no comments justifying this need. We need either to fix the code or justify the need to have a different implementation.

However, there is a bigger issue with this component. The in_operation field protects only in a single thread scenario, but can lead to non-deterministic applications if used in a multithreaded application. The sync should disqualify itself if the library is working in THREAD_MULTIPLE mode.

PR #3263, opened 2017-03-31
hwloc: update hwloc to 1.11.6

ggouaillardet:


rhc54:
@ggouaillardet I don't offhand see anything wrong here - just curious, though, as to what we gain by doing this upgrade? Are there bugs this helps resolve?

jjhursey:
I know there is a performance improvement in [v1.11.6](https://raw.githubusercontent.com/open-mpi/hwloc/v1.11/NEWS):
> * Make the Linux discovery about twice faster, especially on the CPU side,
  by trying to avoid sysfs file accesses as much as possible.

I've tried that and it shows a noticeable improvement with even just `lstopo`.

@ggouaillardet might have other reasons.

ggouaillardet:
The driving factor was to remove an annoying warning on KNL running XPPSL 1.5.1 based on hwloc 1.11.5
KNL file has version 2 but hwloc 1.11.3 only supports version 1

rhc54:
kewl - thx!

ggouaillardet:
@rhc54 @hppritcha i would like PR for the release branches
can you please advise which ones ? ```v3.x``` ? ```v2.x``` ? ```v2.0.x``` ?

rhc54:
Just 3.x - we wouldn't backport that big a change.

PR #3266, opened 2017-03-31
Fix the top_ompi_srcdir in Fortran mpiext building system

abouteiller:
The top_ompi_srcdir configure variable has been suppressed. This results in mpiext defining fortran 90 bindings to fail with an error 
```
make[2]: Entering directory '/global/u1/b/bouteill/cori/ulfm2/debug.build/ompi/mpi/fortran/mpiext'
  PPFC     mpi-ext-module.lo
mpi-ext-module.F90(22): error #5102: Cannot open include file '/ompi/mpiext/ftmpi/mpif-h/mpiext_ftmpi_mpifh.h'
      include '/ompi/mpiext/ftmpi/mpif-h/mpiext_ftmpi_mpifh.h'
--------------^
```
This patch replaces top_ompi_srcdir with srcdir in the configure logic directing mpiext fortran f90 mods.

### DIFF
--- a/config/ompi_ext.m4
+++ b/config/ompi_ext.m4
@@ -7,6 +7,9 @@ dnl Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2011-2012 Oak Ridge National Labs.  All rights reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
+dnl Copyright (c) 2017      The University of Tennessee and The University
+dnl                         of Tennessee Research Foundation.  All rights
+dnl                         reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@abouteiller thanks for the fix !
you need to signoff your commits so the PR can be merged
```
git commit --amend -s
```
will do the trick.

at first glance, i would use ```$top_scrdir``` instead of ```$srcdir```, and i may be wrong about that
@jsquyres can you please double check that ?

jsquyres:
It's probably ok to use either `srcdir` or `ompi_top_srcdir` -- the latter just guarantees to be an absolute directory name (vs. a possibly-relative directory name).

Regardless of which way you go, please squash this PR down into 1 commit (there isn't much sense in committing X, and then committing Y to fix X on a single PR).  Thanks.

PR #3271, opened 2017-04-03
v3.x: hwloc: update hwloc to 1.11.6

ggouaillardet:

(cherry picked from commit open-mpi/ompi@81062b7cd29ec8d53eea36084f95195ee02ce8fc)


bwbarrett:
Is there a driver for this being part of 3.0?  If not, we should consider pushing to the fall release.

ggouaillardet:
@bwbarrett as far as i am concerned
```
The driving factor was to remove an annoying warning on KNL running XPPSL 1.5.1 based on hwloc 1.11.5
KNL file has version 2 but hwloc 1.11.3 only supports version 1
```
some other factors were mentionned in #3263

hppritcha:
At the devel core we do not plan to pull this into 3.0.0, but defer to the next major release which branches in June.  We'll update NEWS for a potential workaround to use external hwloc.

Issue #3270, opened 2017-04-02
disable CMA in vader


hunsa:
Hello all,

We experienced problems with Open MPI when communication larger
messages, for example with `MPI_Gather`, `MPI_Allgather`, etc.  Large
means messages larger than 1k (for very small messages the problem did
not occur).

We received error messages like this
  Read -1, expected 8000, errno = 38

I suspected that there is something wrong with the CMA support.

And indeed, until Open MPI 2.0.2 the configure script would
have the following result
  `#define OPAL_BTL_SM_HAVE_CMA 0`

Now, the new CMA detection method in 2.1.0 leads to 
   `#define OPAL_BTL_SM_HAVE_CMA 1`

That wouldn't be a problem if we could disable CMA in vader during
configure, but it does not seem to be possible.  So, `--with-cma=no` or
`--without-cma` will have no effect and we will end up with
   `#define OPAL_BTL_SM_HAVE_CMA 1`

Currently, we can set 
  `OMPI_MCA_btl_vader_single_copy_mechanism=none`
and the error messages will not show up.

It would be great if we could manually disable CMA support during
configure.

Thank you

ggouaillardet:
Meanwhile, can you try
```
opal_check_cma_happy=0 configure ...
make install
```
and see whether it helps ?

fwiw, we currently only use --with-cma to abort if cma support cannot be built.
I will fix that so we also support --without-cma

hunsa:
`opal_check_cma_happy=0 configure` works and it is much better than having to set `OMPI_MCA_btl_vader_single_copy_mechanism` each time.

hjelmn:
Vader doesn't use BTL_SM_ variables. Look for OPAL_BTL_VADER_HAVE_XPMEM. 

Looks to me like CMA support either doesn't exist on your platform or we have the wrong syscall numbers for the platform. That errno translates to function not implemented. What is the kernel version and platform. It would also be helpful to see your config.log for the build.

hunsa:
@hjelmn Your are right. But the CMA detection macro (`OPAL_CHECK_CMA`) is the same for SM and VADER. Thus, if CMA is enabled, it will be enabled in both cases. It is as you say: our platform has no CMA support. So, I would like to disable it.
Here is the `config.log` (IP/host information filtered out) for a default configure run.

[config.log.zip](https://github.com/open-mpi/ompi/files/890660/config.log.zip)

hjelmn:
[...]

PR #3272, opened 2017-04-03
cma: restore --with-cma=no configure option

hjelmn:
This support broke when we enabled CMA by default. Addreses the issue
raised by #3270.


hunsa:
As a final comment, it now produces the expected result.
When configuring with `--with-cma=no`, we get
`#define OPAL_BTL_VADER_HAVE_CMA 0`
Thanks

hppritcha:
I vote for merging this PR as better approach for specifying no CMA usage.

PR #3280, opened 2017-04-05
Fix the DVM by ensuring that all nodes, even those that didn't partic…

rhc54:
…ipate (i.e., didn't have any local children) in a job, clean up all resources associated with that job upon its completion. With the advent of backend distributed mapping, nodes that weren't part of the job would still allocate resources on other nodes - and then start from that point when mapping the next job. This change ensures that all daemons start from the same point each time.


rhc54:
@hjelmn This should fix the DVM behavior you observed during launch tests

PR #3281, opened 2017-04-05
Adjust the timeout for direct modex requests to reflect the size of t…

rhc54:
…he job. It can take several seconds to start all the procs, and we don't want to timeout due to differences in start times of the various procs


rhc54:
@hjelmn I _think_ this might fix the direct modex problem - please let me know if it does

PR #3283, opened 2017-04-05
build nvml support only with CUDA support

ggouaillardet:


ggouaillardet:
@bgoglin can you please review these changes ?
 - configure ```hwloc``` with ```enable_nvml=no``` unless Open MPI was configure'd with ```--with-cuda```
 - add CUDA flags to ```CPPFLAGS``` if CUDA support was built
   the rationale is that hwloc misses support for ```nvml``` if configure'd with ```--with-cuda``` and CUDA is in the "usual" ```/usr/local/cuda``` place. the only way to have ```nvml``` support is to have CUDA in ```$CPPFLAGS``` which does not feel very natural
 (e.g. ```configure --with-cuda=/usr/local/cuda-8.0 CPPFLAGS=-I/usr/local/cuda-8.0/include ...```)

bgoglin:
As far I as I understand the OMPI-side of things, it looks OK.

hppritcha:
Was this PR addressing issues raised on one of the mai lists?

ggouaillardet:
it might have (i missed details on the issue that was reported)
i will also push an other commit to allow ```--disable-nvml``` and PR them all at once
(that was a previous request in the ML)

sjeaugey:
It may not. We figured out the root cause was the CUDA 8.0 lib. It could still be that NVML init is displaying the warning message, and there is no other call to CUDA, but I wouldn't be so sure.

PR #3285, opened 2017-04-05
coll/base: always send/recv zero-byte messages

ggouaillardet:
Fixes open-mpi/ompi#3255

Thanks Harald Braun for the report


ggouaillardet:
@bosilca could you please review this ?

PR #3292, opened 2017-04-05
IB fork

markalle:
This is to improve fork() support when infiniband is used. The big issue with fork in the presence of pinned pages is copy-on-write if the child touches a pinned page. There's no real fix for this, but you can reduce the likelihood of a child touching one of those pages by page aligning the beginning and end of pinned buffers when possible.

Without that it's too easy for another malloc() to happen to be located on the same page as the pinned memory.

In particular this was happening in btl_openib_connect_udcm.c where a buffer was being pinned with size 65664 (at btl_openib_component_init() time, so it was active whether openib was really being used or not). The start of the buffer was page aligned, but because of the size the end wasn't. That made it too easy for a forked child to accidentally touch pinned memory on the same page as the end of that buffer.

So this change increases the size of the allocated buffer to use the rest of the page.

I inspected the rest of the ibv_reg_mr() calls and changed one other place to page align its buffer too, although I think the above is the one that really matters.

### DIFF
--- a/opal/mca/btl/openib/connect/btl_openib_connect_base.c
+++ b/opal/mca/btl/openib/connect/btl_openib_connect_base.c
@@ -32,6 +32,9 @@
 #include "opal/util/proc.h"
 #include "opal/util/show_help.h"
 
+#include "opal/util/sys_limits.h"
+#include "opal/align.h"
+
 /*
  * Array of all possible connection functions
  */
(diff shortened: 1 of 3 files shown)


hjelmn:
Ah, fork().... Yeah, I see how this can help with that case.

jjhursey:
bot:ibm:gnu:retest

jjhursey:
bot:ibm:gnu:retest

jjhursey:
@markalle This is an old PR that was never merged. Is it still good to go?
bot:ibm:retest

hjelmn:
I think we should go ahead and merge this.

hppritcha:
@markalle and @gpaulsen please PR over to 3.1.x and 3.0.x

PR #3301, opened 2017-04-07
 Correctly identify the source of the event when notifying of abnormal termination by a proces

rhc54:
Also, Fix a potential segfault by avoiding NULL topologies prior to launching the VM.


rhc54:
@abouteiller My apologies - turned out I had this branch sitting on my machine for the last month. This fixes the problem we discussed regarding providing the correct source in event notification of abnormal termination.

PR #3304, opened 2017-04-07
de-ORTEfy the ompi tree

hppritcha:
The ompi tree should be runtime independent, but over time a few
ORTE depedent definitions and functions have escaped into the ompi
tree. I'm working on my own runtime so I've used this as an opportunity
to get rid of ORTE dependencies in the ompi/ tree. I still need to go
back and change orte to conform to the new world and these changes are
untested, but I can now compile (but not link) without orte so I'm
commiting this changeset.

Signed-off-by: Noah Evans <noah.evans@gmail.com>

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -1768,8 +1768,8 @@ int ompi_comm_determine_first ( ompi_communicator_t *intercomm, int high )
         theirproc = ompi_group_peer_lookup(intercomm->c_remote_group,0);
 
         mask = OMPI_RTE_CMP_JOBID | OMPI_RTE_CMP_VPID;
-        rc = ompi_rte_compare_name_fields(mask, (const orte_process_name_t*)&(ourproc->super.proc_name),
-                                                (const orte_process_name_t*)&(theirproc->super.proc_name));
+        rc = ompi_rte_compare_name_fields(mask, (const ompi_process_name_t*)&(ourproc->super.proc_name),
+                                                (const ompi_process_name_t*)&(theirproc->super.proc_name));
         if ( 0 > rc ) {
             flag = true;
         }
--- a/ompi/dpm/dpm.c
+++ b/ompi/dpm/dpm.c
@@ -158,8 +158,8 @@ int ompi_dpm_connect_accept(ompi_communicator_t *comm, int root,
                                               sizeof(ompi_proc_t *));
             for (i=0 ; i<group->grp_proc_count ; i++) {
                 if (NULL == (proc_list[i] = ompi_group_peer_lookup(group,i))) {
-                    ORTE_ERROR_LOG(ORTE_ERR_NOT_FOUND);
-                    rc = ORTE_ERR_NOT_FOUND;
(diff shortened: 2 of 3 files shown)


rhc54:
I'm always curious when someone embarks on completely replacing the runtime - is there some abstraction we are missing that motivates this? In other words, do you really feel you need to replace the entire thing, or is there some part that you would like to change but we haven't abstracted properly?

hppritcha:
I think one of the major issues was orte as currently written is well nigh impossible to port to non-posix systems.  I leave it to the Sandia folks to responded.

npe9:
@rhc54 Like @hppritcha said the main reason we need to be able to take out ORTE is that we don't have full posix system call support. It was actually easier to compile a static binary with a rump runtime than implement all of the system calls that ORTE needed. By only relying on PMI information to bring up MPI on the Kitten kernel we were able to get HPCG running single node relatively easily.

The main remaining dependency was libevent but I was able to hack around it by mapping socketpair() to our local pipe() implementation.

Does that help?

rhc54:
Sure, that makes sense. Be aware that there is a "pmix" component in the works for the ompi/rte framework. The long-term goal is to remove ORTE for any system that provides PMIx (which is expected to be most in the not-too-distant future), leaving ORTE around for those doing research and/or running on non-RM systems. So the new "pmix" component would basically do what you suggest.

If it would help, I'd be happy to collaborate a bit - we might even be able to just use what you are doing. 😄

npe9:
@rhc54 That's funny because that's pretty much what our rump runtime does. I'd be happy to help. Is there a mca object in a development branch somewhere?

rhc54:
I haven't pushed it up anywhere yet, but can do so. Basically, if you look at the MPI layer as it has evolved over the last couple of years, we have replaced nearly all the RTE references with calls to the opal_pmix framework. This was done in preparation for making this architectural change. The only things left are lingering references to the process identifier, all of which could be converted over to the OPAL one, and one call dealing with err handlers (which the pmix component will simply map to the opal_pmix error APIs).

Other than that, the RTE no longer really does anything directly for the MPI layer. It's all being handled by the opal_pmix layer, and since that abstracts above PMI-1 and PMI-2 as well, it is portable. You just lose features (e.g., comm_spawn) when using those older interfaces as most systems don't support the full range of APIs.

npe9:
@rhc54 That's almost exactly what I did (barring lots of hard coding of our own frameworks). If you want to see a snapshot of our nowhere near production rte we did for Kitten it's here: https://github.com/npe9/ompi/commit/f66bb09c79fadb4451c6e18922f9701d999c42ac

I'd love to see your version and see if I can get it to boot on Kitten out of the box.

rhc54:
Will do - I'll try to clean it up and push it over the next week (kinda buried with a couple of things right now). We have a PMIx reference server I use for testing - supports PMIx, PMI-1, and PMI-2 (the older variants being supported by the pmix library) - as it mimics an appropriately-enabled RM. You can find it here: https://github.com/pmix/pmix-reference-server

hppritcha:
I'll go ahead and merge this cleanup PR.  Looks like pmix rte component will address the Sandia orte porting concerns.

PR #3310, opened 2017-04-10
Bring ALPS ODLS up to par regarding wdir.

### LINKED ISSUE #3296: --wdir ignored in recent master
`--wdir` is no longer an effective option, e.g.,:

```
mkdir tmp
mpirun -np 1 -wdir $PWD/tmp pwd
```
returns `$PWD` and not `$PWD/tmp`.

I noticed some related commits (https://github.com/open-mpi/ompi/pull/3217) recently, could these be the cause?

marksantcroos:
This fixes #3296 which was caused by #3217.

Signed-off-by: Mark Santcroos <mark.santcroos@rutgers.edu>

### DIFF
--- a/orte/mca/odls/alps/odls_alps_module.c
+++ b/orte/mca/odls/alps/odls_alps_module.c
@@ -16,6 +16,8 @@
  * Copyright (c) 2011-2014 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2017      Rutgers, The State University of New Jersey.
+ *                         All rights reserved.
  *
  * $COPYRIGHT$
  *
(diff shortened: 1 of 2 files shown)


orte/mca/odls/alps/odls_alps_module.c
```diff
@@ -423,6 +423,11 @@ static int do_child(orte_odls_spawn_caddy_t *cd, int write_fd)
     sigprocmask(0, 0, &sigs);
     sigprocmask(SIG_UNBLOCK, &sigs, 0);
 
+    /* take us to the correct wdir */
+    if (NULL != cd->wdir) {
+        chdir(cd->wdir);
```
hppritcha: I'd like to see this handle the case that chdir returns with error.

marksantcroos: I don't disagree strongly, although I think your request is already covered by the call to orte_util_check_context_cwd() (https://github.com/open-mpi/ompi/blob/master/orte/util/context_fns.c#L56).

Note that my addition is mirroring the existing code in the default ODSL module  (https://github.com/open-mpi/ompi/blob/master/orte/mca/odls/default/odls_default_module.c#L417).

marksantcroos:
@rhc54 Would you like me to add the check in both places? Happy to do so.

rhc54:
Oh - yes please! Much appreciated 😄

hppritcha:
I'm okay with improving possible error return from ```chdir``` with some other PR.

rhc54:
Very nice - thanks!!!

marksantcroos:
> Very nice - thanks!!!

Hold your guns! ;)

It turned out to be less straight forward than I anticipated ...

Just noticed that basename is not set in the Cray case.
Should I possibly hardcode that to "orted" anyway?

rhc54:
I think that should be fine

marksantcroos:
Ok, thanks for the feedback. Tested on Titan and MacOS, its good to go for me.

PR #3311, opened 2017-04-10
common/libfabric: move libfabric to ofi

hppritcha:
This PR renames the common library for OFI libfabric from
libfabric to ofi.  There are a number of reasons this
is good to do:

1) its shorter and replaces 9 characters with three for
   function names for what may eventually be a fairly extensive interface
2) OFI is the term used for MTL and RML components that use
   the OFI libfabric interface
3) A planned OSC component will also use the OFI term.
4) Other HPC libraries that can use OFI libfabric tend to use
   the term "ofi" internally and also in their configure options
   relevant to OFI libfabric (i.e. MPICH/CH4, Intel MPI,  Sandia SHMEM)

There seem to be comments in places in the Open MPI source
code that indicate that this common library will be going away.
Far from it as we will want to be able to share things like
AV objects between OMPI and possibly OSHMEM components that
use the OFI libfabric interface.

This PR also adds a synonym to the --with-libfabric(-libdir)
configury options: --with-ofi and with-ofi-libdir.


ibm-ompi:

 Gist: https://gist.github.com/8974ae07b062e5e32a5886a671327997

config/opal_check_libfabric.m4
```diff
@@ -26,12 +26,26 @@ AC_DEFUN([OPAL_CHECK_LIBFABRIC],[
 	# Add --with options
 	AC_ARG_WITH([libfabric],
 		    [AC_HELP_STRING([--with-libfabric=DIR],
-				    [Specify location of libfabric installation, adding DIR/include to the default search location for libfabric headers, and DIR/lib or DIR/lib64 to the default search location for libfabric libraries.  Error if libfabric support cannot be found.])])
+				    [Specify location of OFI libfabric installation, adding DIR/include to the default search location for libfabric headers, and DIR/lib or DIR/lib64 to the default search location for libfabric libraries.  Error if libfabric support cannot be found.])])
 	AC_ARG_WITH([libfabric-libdir],
```
rhc54: Would it be better to flag these options as "deprecated in favor of --with-ofi", and put the detailed help message over there?

hppritcha: fine with me.  I'll change it.

matcabral:
I agree with the change, I think it will indeed make it simpler and more intuitive to users. Just note, that the two of the mentioned libraries (Sandia SHMEM and MPICH/CH4) also followed OMPI's initial implementation of calling components/files *ofi* and using configure option --with-libfabric.

hppritcha:
@matcabral hmph.  well mpich is messed up.  --with-ofi works for the CH3 device, but not for the CH4 device (which only takes libfabric).  I guess I'll open a PR on MPICH to OFIfy that configury too.

hppritcha:
@jsquyres pushed changes up

PR #3317, opened 2017-04-10
build: Fix platform detection on FreeBSD

bwbarrett:
Look for amd64 in addition to x86_64 as the platform
type for x86_64 assembly.  The FreeBSD-packaged
Autoconf package has a patch to return
amd64-unknown-freebsd11.0 instead of the
x86_64-unknown-freebsd11.0 that a stock Autoconf
package would return.  Since we want to run Jenkins
builds on FreeBSD, working around the FreeBSD patch
is probably the easiest thing.


jsquyres:
Looks like the same race condition false failure we've seen in the Mellanox Jenkins before.  Ignoring.

I don't see why the LANL distcheck failed; that's worth trying again...

bot:lanl:retest
bot:mellanox:retest

Issue #3247, opened 2017-03-28
orterun version 2.1.0 seg fault on tool launch via LaunchMON


lee218llnl:
I am getting seg faults when trying to launch a debug session. This is with the OpenMPI version 2.1.0 release and using LaunchMON as a test:

```
bash-4.2$ ./test.launch_1
[LMON_FE] launching the job/daemons via /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/bin/orterun

[LMON FE] 6 RM types are supported
[rzgenie6:55304] *** Process received signal ***
[rzgenie6:55304] Signal: Segmentation fault (11)
[rzgenie6:55304] Signal code: Address not mapped (1)
[rzgenie6:55304] Failing at address: (nil)
[rzgenie6:55304] [ 0] /lib64/libpthread.so.0(+0xf370)[0x2aaaabc45370]
[rzgenie6:55304] [ 1] /lib64/libc.so.6(+0x133586)[0x2aaaac89c586]
[rzgenie6:55304] [ 2] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/lib/libopen-rte.so.20(orte_schizo_base_setup_fork+0x40)[0x2aaaaadc452f]
[rzgenie6:55304] [ 3] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/lib/libopen-rte.so.20(orte_odls_base_default_launch_local+0x464)[0x2aaaaad5fbea]
[rzgenie6:55304] [ 4] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/lib/libopen-pal.so.20(+0xcce34)[0x2aaaab0e2e34]
[rzgenie6:55304] [ 5] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/lib/libopen-pal.so.20(+0xcd0a6)[0x2aaaab0e30a6]
[rzgenie6:55304] [ 6] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/lib/libopen-pal.so.20(opal_libevent2022_event_base_loop+0x298)[0x2aaaab0e36f9]
[rzgenie6:55304] [ 7] /collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-2.1.0/bin/orterun[0x405090]
[...]

Issue #3243, opened 2017-03-27
Tool attach failure


lee218llnl:
When using the latest OpenMPI commit (d782542c5ce403db835462005523afe2f2613da3) and trying to attach to an MPI job via LaunchMON (https://github.com/LLNL/LaunchMON) I get the following error:

```
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 60 slots
that were requested by the application:
  /nfs/tmp2/lee218/prefix/stat-travis/bin/STATD

Either request fewer slots for your application, or make more slots available
for use.
--------------------------------------------------------------------------
```

I am on a system that allocated nodes via SLURM and this is on 2 nodes with 64 tasks. The application was launched with mpirun/orterun with 4 MPI processes, which might explain why it's trying to fill the node with 64-4=60 slots. However, it should only try launching 1 STATD daemon process per node. Let me know if there are more diagnostics that I can gather to help diagnose this. Note I am able to attach TotalView to the a similarly launched MPI job, so LaunchMON is doing something different than TotalView for process acquisition/daemon launch.

Perhaps similarly related, I notice that if I instead launch the MPI application using all 64 tasks that were allocated, my attempt to attach a tool results in:
```
--------------------------------------------------------------------------
All nodes which are allocated for this job are already filled.
--------------------------------------------------------------------------
```

Let me know if this should be submitted as a separate issue. I can provide additional diagnostics for this too if need be.

lee218llnl:
@rhc54 can you help me look at this one or find someone else who can help?

rhc54:
I can help, but need another day or two to get to it. Is this with OMPI's "master" branch?

lee218llnl:
Thanks, and yes, this was on OMPI's master branch.

rhc54:
@ggouaillardet I'm stuck resolving another issue - would you have a chance to look at this and/or perhaps https://github.com/open-mpi/ompi/issues/3247?

ggouaillardet:
@lee218llnl can you please provide detailled instructions on how to reproduce the issue
- slurm script and sbatch command line
- launchmon command line

lee218llnl:
The same steps outlined in #3247 should demonstrate this issue, albeit pointing to different versions of OpenMPI. Here's what I did:

git clone https://github.com/llnl/launchmon.git
cd launchmon/
export PATH=/collab/usr/global/tools/openmpi/toss_3_x86_64_ib/openmpi-test/bin:$PATH
[...]

PR #3323, opened 2017-04-11
Fix a segfault when starting under a debugger by setting the personality field in the orte_job_t. 

rhc54:
Also, harden the schizo stubs by checking for NULL in that field and returning an error as this should never happen.

Refs #3247 

### DIFF
--- a/orte/mca/schizo/base/schizo_base_stubs.c
+++ b/orte/mca/schizo/base/schizo_base_stubs.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2015      Intel, Inc.  All rights reserved.
+ * Copyright (c) 2015-2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -13,6 +13,7 @@
 #include "orte/constants.h"
 
 #include "opal/class/opal_list.h"
+#include "orte/mca/errmgr/errmgr.h"
 #include "orte/mca/schizo/base/base.h"
 
 int orte_schizo_base_parse_cli(char *personality,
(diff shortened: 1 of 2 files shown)


rhc54:
@lee218llnl This should at least fix the segfault - please check and let us know

lee218llnl:
@rhc54 I can confirm that this does appear to fix the seg fault in #3247. The issue in #3243 still persists. Thanks.

hppritcha:
blocker to get in to 2.1.1

orte/mca/schizo/base/schizo_base_stubs.c
```diff
@@ -46,6 +46,11 @@ int orte_schizo_base_parse_env(char *personality,
     int rc;
     orte_schizo_base_active_module_t *mod;
 
+    if (NULL == personality) {
+        opal_output(0, "NULL PERSONALITY");
```
jsquyres: Is this a debugging output that was accidentally left in here?  (I see 2 more similar calls to opal_output below, too)

rhc54: I can make it an error log - I just wanted something to tell us that a condition which should never occur happened so we can fix it.

jsquyres: Ok, thanks.

General question: It looks like these are assert-like things, right?  I.e., these should never happen / are only there as a failsafe, right?

rhc54: yes - would you prefer they be an assert? Might help debugging, I suppose...

jsquyres: Nah -- you can do it however you want.  😄  I only used "assert" to convey the spirit of what I was thinking this was.  @bosilca would probably hate us putting an `assert()` in the middle of the code.  😉 

But then again, if this case happens, will returning an error code up the stack make it exit cleanly?  Or will it just cause a segv later, anyway?

rhc54: Updated to just error_log it and return the error. Since we return an error if no module can support that operation, this is no worse than the normal error path.

jsquyres: 👍

jsquyres:
@hppritcha Good to go.

PR #3328, opened 2017-04-11
Create a Github issue template

jsquyres:
It seems like we keep getting bug reports where the first questions we have to ask are always the same (E.g., "what version of Open MPI are you using?").

Per https://help.github.com/articles/creating-an-issue-template-for-your-repository/, perhaps we should have an github issue template for Open MPI.

If you're a community developer and want to make a free-form issue, it's easy enough to select-all/delete and write whatever you want.  But having this template may be helpful for newbies -- it may cause a little introspection during the bug reporting process, which is usually a Good Thing.

I tentatively put together a few questions and format.  I'm not tied to this at all -- I'm very open to suggestions and comments.

[skip ci]
bot:notest

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,29 @@
+Thank you for taking the time to submit an issue!
+
+## Background information
+
+### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
+
+
+
+### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
+
+
+
+### Please describe the system on which you are running
+
+* Operating system/version: 
+* Computer hardware: 
+* Network type: 
+
+-----------------------------
+
+## Details of the problem
+
+Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.
+
+**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
+```shell
+shell$ mpirun -np 2 ./hello_world
+```
+


jsquyres:
@rhc54 Done -- added some version examples.

@hppritcha @jjhursey I just force-pushed a new commit that included `[skip ci]`, but the LANL CIs and at least one IBM CI fired again.

jjhursey:
@jsquyres You need to put that in your PR description, not the commit message or a comment. I think that's the problem.

jjhursey:
👍  This looks good to me. Thanks!

jsquyres:
@jjhursey Huh.  I could have sworn it was the opposite (it had to be in the *commit message*, not the PR description).  Ok.

hppritcha:
Looks good.

jjhursey:
I don't want the template to get too long. But what do you think about adding a note about formatting code blocks in markdown with a link and/or example?

Something like:

**Note**: If you have shell output (or a code block) please use the [Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

bwbarrett:
@jsquyres , looking at the OMPI Jenkins, it appears to behave as Josh described as well...

jsquyres:
@jjhursey Good point.  Done.

Ah -- I see the issue now.  If you don't include [skip ci] in the commit message, then Travis fires.  Hence, we really need [skip ci] in *both* the commit message (for Travis) *and* the description (for Jenkins).

bwbarrett:
@jsquyres Or I get off my butt and finish replicated the travis tests on Jenkins :)

rhc54:
I was told there is also a ```bot:notest``` that you have to include to avoid the CI tests. Did that change? IIRC, ```skip ci``` handled Travis, and the other handled Jenkins.

hppritcha:
for the AWS jenkins the projects look for either [skip ci] [ci skip] or bot:notest.  The travis only looks at the first two.

hppritcha:
Since I'm shamelessly copying contributing.md from rails I looked at their issue template.  Its short.
I tried ours and now think its too long.  

https://github.com/rails/rails/issues/new

rhc54:
I think we could make it look shorter by condensing some of the requests - e.g., just ask for expected/observed behavior and drop the detailed description of the problem and 1-2 sentence description as those are redundant.

jsquyres:
@rhc54 Done.  Look ok?

rhc54:
Looks good - thx!

PR #3333, opened 2017-04-12
Add PMIx commit that disables build of pmi-1 and pmi-2 backward compa…

rhc54:
…tiblity support, and update OPAL pmix112 configure.m4 to set that option


hppritcha:
What problem does disabling backwards compatibility fix?

rhc54:
Static builds won't work on Slurm machines without it as the PMI-1/2 symbols in PMIx conflict with the ones in Slurm's PMI libraries. Suspect that would also be true for Alps.

PR #3338, opened 2017-04-12
`ompi_info --show-failed` feature

jjhursey:
The `mca_component_show_load_errors` MCA parameter will display an error to `stddiag` when a `dlopen` fails. But it would be nice to format that information and report it inline in `ompi_info`.

This makes it easy to see the failed component in the list next to other components of the same framework. Handy if you are dropping a module into a build on a system and one of the shared libraries it depends on is not present. Instead of seeing no errors (say if the sysadmin turned off `mca_component_show_load_errors` in the default params file), they will see it listed. For example, dropping the LSF modules into a build on a system that doesn't have the necessary supporting libraries just yet.

```shell
shell$  ompi_info --show-failed | grep fail
                 MCA ess: lsf (failed to load) libbat.so: cannot open shared object file: No such file or directory
                 MCA plm: lsf (failed to load) libbat.so: cannot open shared object file: No such file or directory
                 MCA ras: lsf (failed to load) libbat.so: cannot open shared object file: No such file or directory
```

### DIFF
--- a/opal/mca/base/base.h
+++ b/opal/mca/base/base.h
@@ -15,6 +15,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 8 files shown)


jjhursey:
@jsquyres I think this is ready to commit. Since it navigates through the MCA data structures I wanted to get your 👀  on it before committing.

jsquyres:
bot:ompi:retest

PR #3346, opened 2017-04-13
PML/UCX/YALLA: Fix the message release call.

alinask:
Set message to MPI_MESSAGE_NULL.


alinask:
@yosefe please review.

PR #3351, opened 2017-04-14

artpol84:

(cherry picked from commit 4af7a0827fa0bfc2d7d22016edd2ed3347fdd5ab)

### DIFF
--- a/orte/orted/pmix/pmix_server_gen.c
+++ b/orte/orted/pmix/pmix_server_gen.c
@@ -14,7 +14,7 @@
  * Copyright (c) 2009      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
- * Copyright (c) 2014      Mellanox Technologies, Inc.
+ * Copyright (c) 2014-2017 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2014      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


artpol84:
I'm setting bug label because with current SLURM versions original behavior will cause leftovers. SchedMD is going to fix it in next releases.

jsquyres:
@artpol84 You made the branch for this PR (and #3352 and #3353) in the ompi repository; please do not do that.  Please make branches for PRs in your own fork.

artpol84:
Sorry, I'll fix that. Was unintentional.

jsquyres:
Probably not worth closing / re-opening these 3 PR's (you can't change the branch on a PR); just note it for the future, and be sure to delete these branches when the PR's are merged. Thanks.

artpol84:
Sure!
This was the reason why some time ago I suggested to switch ompi master to PR only mode. Where everybody can merge PR but it is not possible to push directly.

jsquyres:
@artpol84 Not sure I understand... we pretty much *are* in a PR-only mode.

But we still request that you put branches in your own repo.  Keep in mind that if you have permission to merge on the ompi repo, then you also have permission to make a branch on the ompi repo (i.e., it's the same GitHub permission to merge a PR on a branch and to create new branches).  So as long as we allow all of our developers to merge PRs, then they can (accidentally) make branches on the ompi repo, too.

artpol84:
It is possible to force all activity through PR on github, we have such a repo.
There I can merge my PRs but if I will try to create a branch there I'll get permission denied. It is much safer from the data consistency perspective.

jsquyres:
If you can tell me the settings to enable that, that would be great; I could not find a way to do that.

artpol84:
I'll talk to my admins.

rhc54:
Hold on here - we have had this discussion, and NO, we are not going to lock all activity thru PRs.

You made a mistake. Let's not compound it.

artpol84:
@jsquyres I just checked, that repo doesn't work the way I described.
I can't push to the master there - can only go thru PR, but I  can create another branch.

@rhc54 I'm not insisting, also it turns out that I don't know the proper way to do that.

jsquyres:
@rhc54 No fear; I was not thinking of going to "everything must go through PRs" for master; you're right that we agreed not to do that (we do most through PRs, but I agree that having the freedom to occasionally push directly to master is sometimes useful).  That being said, having the ability to prevent people from creating branches on the ompi repo would be useful.  Looks like GitHub doesn't give us that ability, however.  So... we now return you to your regularly scheduled programming.  😄

PR #3354, opened 2017-04-14
Fix yalla PML: MPI_Recv does not return MPI_ERR_TRUNCATE upon overflow

derbeyn:
Applies commit from PR #3260 to v2.0.x
Signed-off-by: Nadia Derbey <Nadia.Derbey@atos.net>


derbeyn:
@yosefe : can you review please?

PR #3356, opened 2017-04-14
Use the node index to compare to daemon vpid when identifying procs to bind

rhc54:


rhc54:
@hppritcha Looks like something broke on the CI?
```
/usr/bin/grep: /usr/lib64/libpciaccess.la: No such file or directory
/usr/bin/sed: can't read /usr/lib64/libpciaccess.la: No such file or directory
libtool: link: `/usr/lib64/libpciaccess.la' is not a valid libtool archive
Makefile:1804: recipe for target 'libasm.la' failed
```

PR #3358, opened 2017-04-14

artpol84:

(cherry picked from commit 4af7a0827fa0bfc2d7d22016edd2ed3347fdd5ab)

### DIFF
--- a/orte/orted/pmix/pmix_server_gen.c
+++ b/orte/orted/pmix/pmix_server_gen.c
@@ -14,7 +14,7 @@
  * Copyright (c) 2009      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
- * Copyright (c) 2014      Mellanox Technologies, Inc.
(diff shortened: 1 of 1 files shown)


artpol84:
@hppritcha 
```
/usr/bin/grep: /usr/lib64/libpciaccess.la: No such file or directory
/usr/bin/sed: can't read /usr/lib64/libpciaccess.la: No such file or directory
libtool: link: `/usr/lib64/libpciaccess.la' is not a valid libtool archive
Makefile:1815: recipe for target 'libasm.la' failed
make[2]: *** [libasm.la] Error 1
make[2]: Leaving directory '/global/u2/h/hpp/jenkins_cori/workspace/ompi_master_PR_CLE/opal/asm'
Makefile:2386: recipe for target 'all-recursive' failed
make[1]: Leaving directory '/global/u2/h/hpp/jenkins_cori/workspace/ompi_master_PR_CLE/opal'
Makefile:1903: recipe for target 'all-recursive' failed
```
Looks like the problem is not related to this PR.

hppritcha:
I think there was an upgrade at NERSC and my attempts to use external hwloc are causing problems.
bot:lanl:retest

hppritcha:
Hmm... actually it looks like the pciaccess devel rpm's weren't installed correctly on the cori system.  I'll disable that node for now.

PR #3360, opened 2017-04-14

artpol84:

(ported from 4af7a0827fa0bfc2d7d22016edd2ed3347fdd5ab)


hppritcha:
bot:lanl:retest

artpol84:
We need this in 2.x!

artpol84:
I think it was already merged: https://github.com/open-mpi/ompi/pull/3359

rhc54:
Okay, good - I couldn't remember.

PR #3361, opened 2017-04-15
dl/dlopen: add libs to wrapper LIBS

jsquyres:
With this, libs (e.g., "-ldl") are not added to the wrapper LIBS flags.  This may work on some platforms, but on at least RHEL 7.3, it does not (i.e., compiling MPI applications fails because it can't find dlopen).

@rhc54 Does this fix your problem?

### DIFF
--- a/opal/mca/dl/dlopen/configure.m4
+++ b/opal/mca/dl/dlopen/configure.m4
@@ -1,6 +1,6 @@
 # -*- shell-script -*-
 #
-# Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
Seems to, yes - thx!

jsquyres:
@hppritcha Having the same false-failure problems others have reported with the LANL-CI.

hppritcha:
bot:lanl:retest

there was a botched install of pciaccess devel rpms on the cori system at NERSC, so removed external hwloc config option.

hppritcha:
bot:lanl:retest

PR #3324, opened 2017-04-11
osc/pt2pt: fix infinite frag allocation loop

hjelmn:
Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>
(cherry picked from commit 12b52b2b2cad64f0d98861964286ceb25bdfbd44)
Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

PR #3366, opened 2017-04-17
Pr/v3.x/osc fixes

jsquyres:
This PR cherry picks 2 fixes that went in on master but were not put on the v3.x branch.

@hjelmn Please verify.


hjelmn:
So, this is to replace #3324?

jsquyres:
Per comment above and per discussion on the webex today, I missed that #3324 exists.  😄 

- #3324 contains 12b52b2
- @hjelmn says that v3.x does not need c72fb30

So I'm closing this PR without merging.

PR #3367, opened 2017-04-17
ompi/comm: Improve MPI_Comm_create algorithm

artpol84:
  will not interfere
  same CID.

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -606,6 +606,14 @@ int ompi_comm_split( ompi_communicator_t* comm, int color, int key,
         }
     }
 
+    /* set the rank to MPI_UNDEFINED. This prevents in comm_activate
+     * the collective module selection for a communicator that will
+     * be freed anyway.
+     */
+    if ( MPI_UNDEFINED == color || (inter && my_rsize==0)) {
+        newcomp->c_local_group->grp_my_rank = MPI_UNDEFINED;
+    }
+
     /* Determine context id. It is identical to f_2_c_handle */
     rc = ompi_comm_nextcid ( newcomp,  /* new communicator */ 
                              comm,     /* old comm */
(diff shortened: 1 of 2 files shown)


artpol84:
I selected v1.10 as it's comm_create is more straightforward and it easier to experiment.
Want to see how this change goes through CI.

artpol84:
@edgargabriel @hjelmn I had concerns about this approach, but I checked MPICH and it seems like they do the same for their algo:
https://github.com/pmodels/mpich/blob/master/src/mpi/comm/contextid.c#L416:L423

artpol84:
I'm going to test this branch against MPICH test suite that has comm create tests.
I haven't noticed comm create tests in Mellanox CI, so we haven't actually tested the feature with CI.
I will post my results here tomorrow.

hjelmn:
Looks fine to me. Have you run this through MTT already to make sure it doesn't introduce any regressions?

artpol84:
@karasevb is doing that now. we have mpich test suite there - it as I recall there are a lot of comm tests there.

hjelmn:
@artpol84 Yeah, we need to bring some of those tests over to MTT one of these days...

artpol84:
Actually we are now submitting the results of our SLURM/pmix + ompi v2.x MTT to the community and MPICH tests are there: https://mtt.open-mpi.org/index.php?do_redir=2413.
It's tricky with this PR because it is v1.10, but @karasevb will take care of that.

artpol84:
Here is some comm functionality testing:
https://mtt.open-mpi.org/index.php?do_redir=2414

hjelmn:
@bosilca We are doing it because the collectives require all peers in the original communicator to participate. We could always use the group allreduce if we want to change that.

bosilca:
@hjelmn I specifically asked about the local part, not about the collective.

artpol84:
@hjelmn I checked against MTT and fixed some problems.
Now MTT is clean.

artpol84:
As we agreed I will port this to v2.x, v3.x and the master.

artpol84:
Correction: this PR was rejected in v2.x and v2.0.x. And it is pending in v3.x.
So I guess you should revert it.
Note that WIP-DNM was on this PR, I wasn't expecting you to merge it.

artpol84:
Oops, wrong place

artpol84:
Rejected PRs for the reference:
* https://github.com/open-mpi/ompi/pull/3376
* https://github.com/open-mpi/ompi/pull/3388

PR #3369, opened 2017-04-17
Improve the opal_pointer_array & more

bosilca:
This PR is related to #2426, as it originally tried to fix the performance degradation in datatype creation which was entirely due to the implementation of the opal_pointer_array. Thus, i replace the original implementation by one that in addition to the data array keeps an array of bits to find the next available items in a more compact way.

However, after the reimplementation of opal_pointer_array I also decided to remove the automatic registration of the OMPI datatype into the f2c (Fortran-to-C) translation table, and instead do the registration only when necessary (basically when we know that we will return the datatype object through the Fortran interface).

Here are some performance numbers (I could not replicate the original numbers on the #2426, but I strongly suspect they were obtained in debug mode). 

Original OMPI: 29.458 s
above + rewrite of opal_pointer_array: 0.787 s
above + not registration of the datatype: 0.581 s

The last trick (not automatically registering MPI objects) can also be used for a wide range of MPI objects: windows, files, communicators, requests, info, attributes and more.

### DIFF
--- a/opal/class/opal_pointer_array.c
+++ b/opal/class/opal_pointer_array.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2007 The University of Tennessee and The University
+ * Copyright (c) 2004-2017 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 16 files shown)


opal/class/opal_pointer_array.c
```diff
@@ -67,6 +70,103 @@ static void opal_pointer_array_destruct(opal_pointer_array_t *array)
...
+ */
+#define UNSET_BIT(IDX)                                                  \
+    do {                                                                \
+        uint32_t __b_idx, __b_pos;                                      \
+        GET_BIT_POS((IDX), __b_idx, __b_pos);                           \
+        assert( (table->free_bits[__b_idx] & (1UL << __b_pos)));        \
+        table->free_bits[__b_idx] ^= (1ULL << __b_pos);                 \
+    } while(0)
+
+#if 0
```
rhc54: @bosilca Is this intentionally defined out?

rhc54: Never mind - I see what you are doing.

bosilca: That function is too expensive to use in general (even in debug mode). But, in case something goes wrong, it is useful to debug.

PR #3371, opened 2017-04-18
coll/hcoll: performance bugfix

vspetrov:
Don't register mca_coll_hcoll_progress to opal_progress until
    the first call to hcoll non-blocking collective.


vspetrov:
Closing. Another approach is chosen.

PR #3374, opened 2017-04-19
mpi/c: allow MPI_PROC_NULL in MPI_Win_shared_query()

ggouaillardet:
This fixes a regression introduced in open-mpi/ompi@b3a20100d3d31e4937a5b23d012c8ae9b22e0cd3

(cherry picked from commit open-mpi/ompi@23dad50d515df0fa9910f7349b61e2423297eda0)

### DIFF
--- a/ompi/mpi/c/win_shared_query.c
+++ b/ompi/mpi/c/win_shared_query.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2012-2013 Sandia National Laboratories. All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@hjelmn per @bwbarrett review, ```MPI_Win_lock()``` should be updated.

currently
```c
    if (ompi_win_peer_invalid(win, rank)) 
        return OMPI_ERRHANDLER_INVOKE(win, MPI_ERR_RANK, FUNC_NAME);
    // ...
    /* NTH: do not bother keeping track of locking MPI_PROC_NULL. */
    if (MPI_PROC_NULL == rank) return MPI_SUCCESS;
````

we previously updated ```ompi_win_per_invalid``` to return true if rank is ```MPI_PROC_NULL```, so this obviously impacts the behavior of ```MPI_Win_lock()```

that being said, and per Bill Gropp in http://lists.mpi-forum.org/pipermail/mpi-forum/2016-January/006417.html
```
The standard says that the null handles are invalid for input unless explicitly permitted.
```
context was ```MPI_DATATYPE_NULL```, but imho the statement should apply to ```MPI_PROC_NULL``` too.
i checked the standard and could not find any exception saying that ```MPI_PROC_NULL``` is a valid parameter for ```MPI_Win_lock()```, so at this stage, i think that ```MPI_Win_lock()``` is now working as expect, and we should simply remove two lines of dead code.

does that make sense ?

@bosilca feel free to comment on that too

bosilca:
I stayed away from this discussion simply because I can't really wrap my head around what is a well-defined behavior for MPI_Win_lock on MPI_PROC_NULL. Should there be a corresponding MPI_Win_unlock?

bwbarrett:
Reading MPI-3.1 again, I think Gilles is right; there's no need to handle MPI_PROC_NULL in MPI_WIN_LOCK/MPI_WIN_UNLOCK.  11.3 only says that MPI_PROC_NULL is a valid target rank in all MPI RMA *communication* calls, and lock/unlock is not a communication call.  And I agree with George that I can't wrap my head around what that would mean anyway.

PR #3375, opened 2017-04-19
ompi/comm: Improve MPI_Comm_create algorithm

artpol84:
      will not interfere
      same CID.

### DIFF
--- a/ompi/communicator/comm_cid.c
+++ b/ompi/communicator/comm_cid.c
@@ -21,6 +21,7 @@
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2017      Mellanox Technologies. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -303,6 +304,7 @@ static int ompi_comm_allreduce_getnextcid (ompi_comm_request_t *request)
     ompi_request_t *subreq;
     bool flag;
     int ret;
+    int participate = (context->newcomm->c_local_group->grp_my_rank != MPI_UNDEFINED);
 
     if (OPAL_THREAD_TRYLOCK(&ompi_cid_lock)) {
         return ompi_comm_request_schedule_append (request, ompi_comm_allreduce_getnextcid, NULL, 0);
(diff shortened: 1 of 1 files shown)


artpol84:
## Algorithm efficiency verification
I've applied the following patch to verify that algorithm provides performance improvement:
```patch
diff --git a/ompi/communicator/comm_cid.c b/ompi/communicator/comm_cid.c                                                                                                
index 6f67d99..35d978f 100644                                                                                                                                           
--- a/ompi/communicator/comm_cid.c                                                                                                                                      
+++ b/ompi/communicator/comm_cid.c                                                                                                                                      
@@ -411,6 +411,15 @@ static int ompi_comm_nextcid_check_flag (ompi_comm_request_t *request)                                                                             
     }                                                                                                                                                                  
.                                                                                                                                                                       
     if (1 == context->rflag) {                                                                                                                                         
+       {                                                                                                                                                               
+           static int comm_count = 0;                                                                                                                                  
+           int rank;                                                                                                                                                   
+           MPI_Comm_rank(MPI_COMM_WORLD, &rank);                                                                                                                       
+           if( 0 == rank ){                                                                                                                                            
+               printf("Comm #%d: CID=%d, iters=%d\n", ++comm_count, context->nextcid, context->iter);                                                                  
+           }                                                                                                                                                           
+      }                                                                                                                                                               
+                                                                                                                                                                    
         if( !participate ) {                                                                                                                                           
             /* we need to provide something sane here                                                                                                                  
              * but we cannot use `nextcid` as we may have it                                                                                                           
```
This debug output demonstrates significant improvement for my comm_create benchmark (https://github.com/artpol84/poc/tree/master/benchmarks/comm_create) that I derived from Amber app :
* [old_cid_algo.txt](https://github.com/open-mpi/ompi/files/945198/old_cid_algo.txt)
Here number of iterations is constantly growing up to ~1000 and max CID is `1082`
* [new_cid_algo.txt](https://github.com/open-mpi/ompi/files/945201/new_cid_algo.txt)
In this case max number of iterations is 13 (avg is 3) and max CID is `26`. This confirms that we have more compact CID distribution with less fragmentation, because CID's are reused by non-overlapping groups of processes.

Overall new algo is about 80 times faster than the old one.

artpol84:
MTT is clean, I started one MTT with enabled reporter to `mtt.open-mpi.org`. Will post the link once it's finished.

artpol84:
@hjelmn I also fixed some error handling paths. Could you please double-check me?

artpol84:
I performed MTTs for each branches internally with mpirun launches. 
But I also tested `comm_create/v2.x` from PR #3376 branch with our SLURM/pmix MTT that is reported outside to `mtt.open-mpi.org`. For all of them all tests was successful.
Here are the publicly available results with SLURM/pmix `srun` launch: https://mtt.open-mpi.org/index.php?do_redir=2417.

jladd-mlnx:
👍

jsquyres:
@bosilca @ggouaillardet please review

jsquyres:
Please fix typo in commit message:

I assume "ne" should be "new".

artpol84:
I saw that it was already merged into 3.x and 1.10. 
So merge into master.

rhc54:
Sigh - I thought after this much time it had already been put in master. In future, please be advised that we don't work this way - it must be put into master _before_ opening PRs on release branches. Otherwise, it leads to this kind of confusion.

I won't revert it out of 1.10 - let's see what happens on master and then we'll decide.

artpol84:
I don't mind - All PRs was there and ready to go. I usually don't merge my own PRs.

rhc54:
Ummm...why would you be expecting someone else to merge your PRs to master? I can't think of anyone else who does that, or any reason for it.

PR #3376, opened 2017-04-19
ompi/comm: Improve MPI_Comm_create algorithm (v2.x)

artpol84:
      will not interfere
      same CID.

(cherry picked from commit 50a6d106f655d5c36aae62aad59121547f60648a)

### DIFF
--- a/ompi/communicator/comm_cid.c
+++ b/ompi/communicator/comm_cid.c
@@ -21,6 +21,7 @@
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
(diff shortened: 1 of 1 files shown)


artpol84:
## Efficiency verification
https://github.com/open-mpi/ompi/pull/3375#issuecomment-295864721

artpol84:
https://github.com/open-mpi/ompi/pull/3375#issuecomment-296007724

jsquyres:
Do not merge until #3375 is merged.

jsquyres:
Per discussion on the 2017-04-26 webex and per further discussion by the RMs on 2017-04-26, this appears to be an enhancement that isn't classified as a bug fix (e.g., no customer problems were cited).  As such, we're dropping it from v2.1.x.

PR #3390, opened 2017-04-21
v2.1.x: usnic: ensure to set the iov_limit to 1

jsquyres:
The usNIC BTL does not use more than 1 iov, so be sure to set it to 1 so that we don't allocate cq/rq/sq entries based on a default (i.e., >1) number of iovs per entry.

(cherry picked from commit 7bd2de9960419422a4591f4b5d286f1f911a0a47)


jsquyres:
@hppritcha Good to go when CI finishes.

PR #3400, opened 2017-04-22
 Set the default modex parameters back to full blocking modex while w…

rhc54:
…e continue to test and debug the slow modex - it seems to be having issues on the Cray


rhc54:
Looks like "pull request build checker" is having problems with GitHub.

PR #3410, opened 2017-04-25
group: Fix `ompi_group_have_remote_peers`

kawashima-fj:
`ompi_group_t::grp_proc_pointers[i]` may have sentinel values even for processes which reside in the local node because the array for `MPI_COMM_WORLD` is set up before `ompi_proc_complete_init`, which allocates `ompi_proc_t` objects for processes reside in the local node, is called in `MPI_INIT`. So using `ompi_proc_is_sentinel` against `ompi_group_t::grp_proc_pointers[i]` in order to determine whether the process resides in a remote node is not appropriate.

This bug sometimes causes an `MPI_ERR_RMA_SHARED` error when `MPI_WIN_ALLOCATE_SHARED` is called, where sm OSC uses `ompi_group_have_remote_peers`.

Bug reproducer program:

```c
// run as "mpiexec -n 2 --host localhost,localhost a.out" (all processes in one node)

#include <mpi.h>
{
    void *base;
    MPI_Win win;
    MPI_Init(&argc, &argv);
    MPI_Win_allocate_shared(8, 1, MPI_INFO_NULL, MPI_COMM_WORLD, &base, &win);
    MPI_Win_free(&win);
    MPI_Finalize();
    return 0;
}
```

This is a regression introduced in v2.0 (dynamic add_procs enhancement).

@hjelmn Could you review? This function was written by you in 0bf06de. Is my change a *correct* fix?

### DIFF
--- a/ompi/group/group.c
+++ b/ompi/group/group.c
@@ -563,10 +563,13 @@ bool ompi_group_have_remote_peers (ompi_group_t *group)
 #if OMPI_GROUP_SPARSE
         proc = ompi_group_peer_lookup (group, i);
 #else
-        if (ompi_proc_is_sentinel (group->grp_proc_pointers[i])) {
+        proc = ompi_group_get_proc_ptr_raw (group, i);
+        if (NULL == proc) {
+            /* the proc must be stored in the group or cached in the proc
+             * hash table if the process resides in the local node
+             * (see ompi_proc_complete_init) */
             return true;
         }
-        proc = group->grp_proc_pointers[i];
 #endif
         if (!OPAL_PROC_ON_LOCAL_NODE(proc->super.proc_flags)) {
             return true;


hjelmn:
Looks good to me. PR to v2.0.x, v2.x, and v3.x.

kawashima-fj:
@rhc54 Thanks and sorry. I reverted my commit in master (9841ad3).

Each rank has one of three states in the `ompi_group_have_remote_peers` function.

1. `group->grp_proc_pointers[rank]` is *not* a sentinel value and `ompi_proc_hash` points a `ompi_proc_t` pointer for the rank
2. `group->grp_proc_pointers[rank]` is a sentinel value but `ompi_proc_hash` points a `ompi_proc_t` pointer for the rank
3. `group->grp_proc_pointers[rank]` is a sentinel value and `ompi_proc_hash` points a `NULL` for the rank

My commit changed the `ompi_group_have_remote_peers` function to use the `ompi_group_get_proc_ptr_raw` function in order to solve the case of 2. But the `ompi_group_get_proc_ptr_raw` function returns a sentinel value intead of `NULL` in the case of 3. so accessing `proc->super.proc_flags` causes a segmentation fault.

Using `ompi_group_get_proc_ptr` instead of `ompi_group_get_proc_ptr_raw` will resolve the problem. I'll confirm.

PR #3412, opened 2017-04-25
hwloc1116: add CUDA include dir to CPPFLAGS if needed

ggouaillardet:
Thanks Siegmar Gross for the report


ggouaillardet:
@jsquyres this should fix and issue reported by @siegmargross related to compilation with CUDA

ideally, i'd like to add cuda include dir __only__ when building in ```opal/mca/hwloc/hwloc1116/hwloc``` but ```hwloc``` configury is embedded in ompi and i do not know how to do that

any thoughts ?

ggouaillardet:
closing as a duplicate of #3412

PR #3420, opened 2017-04-26
v2.1.1: update the VERSION file

hppritcha:
the upping of age and rev for libmpi.so/libmpif owes
to additions to the mpi.h and mpif.h include files.

[ci skip]

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -4,6 +4,8 @@
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
 # Copyright (c) 2016      IBM Corporation.  All rights reserved.
+# Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
+                          reserved.
 
 # This is the VERSION file for Open MPI, describing the precise
 # version of Open MPI in this distribution.  The various components of
@@ -24,7 +26,7 @@ release=1
 # requirement is that it must be entirely printable ASCII characters
 # and have no white space.
 
-greek=a1
+greek=rc1
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"
@@ -82,17 +84,17 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=30:0:10
+libmpi_so_version=30:1:10
 libmpi_cxx_so_version=30:0:10
-libmpi_mpifh_so_version=30:0:10
+libmpi_mpifh_so_version=31:0:11
 libmpi_usempi_tkr_so_version=30:0:10
 libmpi_usempi_ignore_tkr_so_version=30:0:10
 libmpi_usempif08_so_version=30:0:10
 
-libopen_rte_so_version=30:0:10
-libopen_pal_so_version=30:0:10
+libopen_rte_so_version=30:1:10
+libopen_pal_so_version=30:1:10
 libmpi_java_so_version=30:0:10
-liboshmem_so_version=30:0:10
+liboshmem_so_version=30:1:10
 libompitrace_so_version=30:0:10
 
 # "Common" components install standalone libraries that are run-time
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -82,17 +84,17 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=30:0:10
+libmpi_so_version=31:0:11
```
jsquyres: I agree with the libmpifh version change, but I do not see an `mpi.h` change that would cause this `libmpi_so_version` change.  I'm looking on the `v2.x` branch starting with the 62cd194e32d44242d29114779f9cda13c2a3e70b commit.  Are we looking at the same commit range?

VERSION
```diff
@@ -82,17 +84,17 @@ date="Unreleased developer copy"
...
 libmpi_usempi_ignore_tkr_so_version=30:0:10
 libmpi_usempif08_so_version=30:0:10
 
-libopen_rte_so_version=30:0:10
-libopen_pal_so_version=30:0:10
+libopen_rte_so_version=30:1:10
+libopen_pal_so_version=30:1:10
 libmpi_java_so_version=30:0:10
-liboshmem_so_version=30:0:10
+liboshmem_so_version=30:1:10
```
jsquyres: I didn't see any OSHMEM changes on `v2.x` after 62cd194e32d44242d29114779f9cda13c2a3e70b.  Did I miss them?

VERSION
```diff
@@ -108,6 +110,6 @@ libmca_orte_common_alps_so_version=30:0:10
 # OPAL layer
 libmca_opal_common_cuda_so_version=30:0:10
 libmca_opal_common_libfabric_so_version=30:0:10
-libmca_opal_common_sm_so_version=30:0:10
+libmca_opal_common_sm_so_version=30:1:10
```
jsquyres: I also didn't see any `opal/mca/common/sm` changes on `v2.x` after 62cd194e32d44242d29114779f9cda13c2a3e70b.

hppritcha:
I got the list of changed files by checking out v2.x in my repo and then doing

git diff --name-only 1cd1edff78..HEAD 

the sha is that for the v2.1.0 tag

hppritcha:
Sorry typed the sha wrong.  Typing on my iphone

hppritcha:
What got command are you using?  The pvar sessions thing got merged in as 3139.  hmmm that was the thing I missed in the readme.

jsquyres:
Looks like I started 1 commit after you, and that didn't really affect any noticeable change.

I just checked out the `v2.x` head and ran `git log --stat --no-merges > log.txt`.  I then opened the file in emacs, went down to 62cd194e32d44242d29114779f9cda13c2a3e70b, and trimmed everything after that.

hppritcha:
i dont think your command is the right one.

hppritcha:
Problem is that things like the mpi.h.in change was only merged in a few weeks ago.

hppritcha:
#3139 pulled that change in.  I merged Omir's commit on 3/29.

hppritcha:
same kind of mistake with the sm common.  The change was actually merged in by #3189 which got merged in by me on 3/28.

hppritcha:
Oh I noticed I had updated README with #3139 info.  Just forgot to delete the reminder line.  So the README should be ready to go.

VERSION
jsquyres: Ok, I guess I shouldn't be using `--no-merge` in my `git log`.

So I now see fda91572ce that modified `mpi.h`.  But it just added a `#define` -- it didn't add any interfaces.  So I don't think we should be bumping to `31:0:11`, but rather to `30:1:10`.

VERSION
jsquyres: Ok, I see the oshmem change now.

VERSION
jsquyres: And I see the common/sm change now.

jsquyres:
Only thing left now is the version for libmpi.

VERSION
hppritcha: libmpi.so got bumped by fda91572 which was merged in by 4b6d34a9 (PR #3138)

VERSION
hppritcha: commit 2379e6e8 which was merged in by 97c95c4d as part of #3189

hppritcha:
Your git command is defective for finding changes between releases, at least within a given branch.  I'll update the release checklist.

PR #3238, opened 2017-03-24
Fix segfault in TCP/BTL

rhc54:
Correctly check the sa_family and cast the data correctly before passing it to inet_nop, and don't be quite as fancy with the pointer arithmetic as the combination was causing us to segfault every time this debug message was called.

Signed-off-by: Ralph Castain <rhc@open-mpi.org>
(cherry picked from commit 470452cba03b7b5c684d9d99acc0a73c6c47746a)

PR #3430, opened 2017-04-28
Use the OPAL function to get the hostname.

bosilca:
Provides a better fix (as we use OPAL level functions instead of rewriting them) than #3238.

@rhc54  and @jsquyres sorry for the delay.


bwbarrett:
bot:ompi:retest

rhc54:
Thanks @bosilca

Issue #3392, opened 2017-04-21
2.0.2 and 2.1.0: orted assertion failure with output-filename and rankfile


tjb900:
## Background information
First encountered this problem on startup of MPI jobs with 2.0.2, under slurm.  It presents as the orted segfaulting, and appears to occur when the following two things are true:

1) --output-filename is specified to mpirun, to write each rank to a separate file
2) a rankfile is specified, and rank 0 is not specified to be on the launch node 

Yielding the error (in the --enable-debug version of 2.1.0):

```
orted: class/opal_list.c:69: opal_list_item_destruct: Assertion `0 == item->opal_list_item_refcount' failed
```

or, for non-debug versions, a segmentation fault.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
Have reproduced this issue with 2.0.2, 2.1.0.  Have attempted to reproduce on master, but it appears that a separate rankfile-related problem gets in the way first.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
2.0.2 and 2.1.0 were built from a tarball, with --enable-mpi-thread-multiple, using 2017 intel compiler

The log below was made with a tarball from 2.1.0, compiled with --enable-mpi-thread-multiple and --enable-debug

### Please describe the system on which you are running

* Operating system/version: Centos 6.6
* Computer hardware: Intel, most reproductions on KNL but also seen on regular Xeon
* Network type: TCP/Ethernet

-----------------------------

## Details of the problem

Can be reproduced with the following script:

```#!/bin/bash
LOGDIR=/d/home/timb/openmpi_issue_20170421/2.1.0-mt-debug/${SLURM_JOB_ID}
rankfile=${LOGDIR}/rankfile.${SLURM_JOB_ID}
mkdir -p ${LOGDIR}
nodelist=$(scontrol show hostname ${SLURM_JOB_NODELIST} | tac)
> $rankfile
rank=0
for node in $nodelist; do
   echo "rank ${rank}=${node} slot=0:*" >> $rankfile
   rank=$(( $rank + 1 ))
done

echo "Script running on $(hostname)"
echo "=== contents of rankfile"
cat ${rankfile}
echo "=== end contents of rankfile"

export OPENMPI=/d/sw/openmpi/2.1.0-mt-debug
[...]

PR #3432, opened 2017-04-28
v2.x: fix duplication of stdin into --output-filename files

tjb900:
Looks like the stdin stream was being written to all sinks, even the ones meant for stdout/err/diag.  Fixes open-mpi/ompi#3392

3.x and master appear from the code like they might not suffer from the same issue.

Signed-off-by: Tim Burgess <ozburgess@gmail.com>


jsquyres:
bot:retest

bwbarrett:
bot:ompi:retest

hppritcha:
@rhc54 
A comment in #3392 indicates the original problem was caused by an erroneous close of ```fd==0```.   Should this PR be closed?

rhc54:
These are separate issues - this PR should be committed.

PR #3434, opened 2017-04-28
ompi/opal: add support for HDR link speeds

dsharma283:
This patch enables to use adapters with HDR and NDR
speeds.
issue id 3431

Signed-off-by: Devesh Sharma <devesh.sharma@broadcom.com>

### DIFF
--- a/opal/mca/common/verbs/common_verbs_port.c
+++ b/opal/mca/common/verbs/common_verbs_port.c
@@ -68,6 +68,10 @@ int opal_common_verbs_port_bw(struct ibv_port_attr *port_attr,
         /* EDR: 25.78125 Gbps * 64/66, in megabits */
         *bandwidth = 25000;
         break;
+    case 64:
+        /* HDR: 50Gbps * 64/66, in megabits */
+        *bandwidth = 50000;
+        break;
     default:
         /* Who knows? */
         return OPAL_ERR_NOT_FOUND;


jsquyres:
bot:retest

bwbarrett:
bot:ompi:retest

opal/mca/common/verbs/common_verbs_port.c
```diff
@@ -68,6 +68,14 @@ int opal_common_verbs_port_bw(struct ibv_port_attr *port_attr,
         /* EDR: 25.78125 Gbps * 64/66, in megabits */
         *bandwidth = 25000;
         break;
+    case 64:
+        /* HDR: 50Gbps * 64/66, in megabits */
+        *bandwidth = 50000;
+        break;
+    case 128:
```
jsquyres: Two questions:

1. According to https://github.com/torvalds/linux/blob/master/include/rdma/ib_verbs.h#L428-L435, I do not see values for 64 and 128 defined in the kernel ABI for the `active_speed` field when querying IB port attributes.  Where are you getting these 64/128 values from?

2. As a secondary (but related) issue, is there a better way than using hard-coded integer values here in `common_verbs_port.c`?  On the kernel side, we have the nice `IB_SPEED_*` enums; do public equivalents exist in userspace that we can use here in Open MPI?  (ditto for `IB_WIDTH_*`)

dsharma283: Hi Jeff,

Thanks for your comments and feedback.
 Regarding the invalid pointer free(), I did some initial debugging and found that openmpi is raising this stack trace whenever it fails to select any of the usable btl for a given process. for example, with my transport if I select btl_openib_flags in such a way where only PUT support is available (flags value 306). I will open a seprate PR for this issue.

Yes, it is true that currently values 64 and 128 are not defined in the kernel/user IB stack. I will push the needed change to the ib-stack very soon.

To the best of my knowledge, those macros are not there in user-space. I can try pushing those to rdma-core once.

jsquyres: 1. Yes, either a separate PR or another commit on this PR to fix that segv would be great; thanks.
1. Once you get a patch accepted upstream for the 64/128 values, we can probably take this commit.  I just wouldn't want to take a commit here that is dependent upon a vendor-specific OFED stack (which then might vary between vendors).  Hope that makes sense.
1. I think getting the macros would be an extra bonus, and can likely be a separate upstream patch for libibverbs.  I.e., depending on the timing, either a separate PR or a separate commit on this PR that converts these naked values to enums would be awesome.

Many thanks.

dsharma283:
I have posted a change in upstream linux-rdma community to pull the port-speed and link-width enums. I will supply a rev of this patch once the change is accepted.
http://marc.info/?l=linux-rdma&m=149425108020822&w=2

Just for the reference I am leaving this pull-request open till I comeback with proper fix.

opal/mca/common/verbs/common_verbs_port.c
chrissamuel: Just to note that we have been pointed to this PR by Mellanox about OMPI not working with CX5 cards and ROCE.

However, we do not see the backtrace that @dsharma283 reports, just the error:

`--------------------------------------------------------------------------
WARNING: There was an error initializing an OpenFabrics device.

  Local host:   spartan-gpgpu008
  Local device: mlx5_0
WARNING: There was an error initializing an OpenFabrics device.

  Local host:   spartan-gpgpu007
  Local device: mlx5_0
srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
srun: error: spartan-gpgpu008: task 1: Exited with exit code 17
srun: error: spartan-gpgpu007: task 0: Killed`

chrissamuel:
I can confirm that the supplied change to `opal/mca/common/verbs/common_verbs_port.c` makes the verbs interface work with RoCE on Mellanox CX5 cards connected at 50Gb/s .

Before (via TCP btl):
```
       8 bytes took        53 usec (   0.301 MB/sec)
      16 bytes took        56 usec (   0.569 MB/sec)
[...]
```

After (via openib btl):
```
       8 bytes took        15 usec (   1.080 MB/sec)
      16 bytes took         6 usec (   5.096 MB/sec)
[...]
```

Patched with just:
```
[samuel@spartan-build openmpi-3.0.0]$ diff -iubw opal/mca/common/verbs/common_verbs_port.c.orig opal/mca/common/verbs/common_verbs_port.c
--- opal/mca/common/verbs/common_verbs_port.c.orig      2017-09-13 07:31:33.000000000 +1000
+++ opal/mca/common/verbs/common_verbs_port.c   2017-11-15 10:45:53.643484102 +1100
@@ -68,6 +68,14 @@
         /* EDR: 25.78125 Gbps * 64/66, in megabits */
         *bandwidth = 25000;
         break;
+    case 64:
+        /* HDR: 50Gbps * 64/66, in megabits */
+        *bandwidth = 50000;
+        break;
+    case 128:
+        /* NDR: 100Gbps * 64/66, in megabits */
+        *bandwidth = 100000;
+        break;
     default:
         /* Who knows? */
         return OPAL_ERR_NOT_FOUND;
```

jsquyres:
@chrissamuel @dsharma283 Did that patch get accepted upstream?

chrissamuel:
Not from what I see, it was discussed here:

https://patchwork.kernel.org/patch/9716209/

There was talk of rearchitecturing it in a way that might result in some distros being unsupported.

dsharma283:
This mail also made me think that this patch has been pulled in. This
action item is still open its just I am not able to prioritize on it.

-Regards
Devesh

chrissamuel:
@dsharma283 I see no evidence of any mention of anything using IBV_ in the linux-rdma repos here. https://github.com/linux-rdma

chrissamuel:
@jsquyres looking at your link it looks like the file has changed since then and IB_SPEED_HDR is now defined for ib_port_speed:

https://github.com/torvalds/linux/blob/master/include/rdma/ib_verbs.h#L464-L472

```
enum ib_port_speed {
	IB_SPEED_SDR	= 1,
	IB_SPEED_DDR	= 2,
	IB_SPEED_QDR	= 4,
	IB_SPEED_FDR10	= 8,
	IB_SPEED_FDR	= 16,
	IB_SPEED_EDR	= 32,
	IB_SPEED_HDR	= 64
};
```

hppritcha:
@jsquyres could you double check whether this is okay or not now

jsquyres:
I updated the patch to remove the "128" case (because that's not upstream).

dsharma283:
Thanks for pulling-in.

jsquyres:
No worries.  Note that this cherry pick is slated for the next releases in the v2.1.x, v3.0.x, and v3.1.x series (https://github.com/open-mpi/ompi/pull/4945, #4946, and #4947).

PR #3436, opened 2017-04-28
btl tcp: Don't set socket buffer size by default

bwbarrett:
Set the default send and receive socket buffer size to 0,
which means Open MPI will not try to set a buffer size during
startup.

The default behavior since near day one of the TCP BTL has been
to set the send and receive socket buffer sizes to 128 KiB.  A
number that works great on 1 GbE, but not so great on 10 GbE
fabrics of any real size.  Modern TCP stacks, particularly on
Linux, have gotten much smarter about buffer sizes and are much
less efficient if a buffer size is set (even if set to something
large).

(cherry picked from commit 3b991498bedc1559c14093ba0a74dd16f2dd0bce)


jsquyres:
bot:mellanox:retest

jsquyres:
@hppritcha Good to go when CI finishes (CI finished just fine on the v2.x version of this PR -- likely a false failure on the Mellanox Jenkins).

PR #3440, opened 2017-05-01
ess/singleton: do not pass "--mca state_novm_select 1" to orted

ggouaillardet:
and fix a hang when MPI_Comm_spawn is invoked from a singleton

### DIFF
--- a/orte/mca/ess/singleton/ess_singleton_module.c
+++ b/orte/mca/ess/singleton/ess_singleton_module.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
currently, the ```dynamic/spawn``` test from the ```ibm``` test suite hangs when invoked as a singleton.
i ran this under the debugger to compare singleton vs mpirun, and i finally found that removing ```--mca state_novm_select 1``` did the trick.

rhc54:
Did you test that with multiple nodes - i.e., where comm_spawn forces the launch of new daemons? I believe that was the use-case which required the novm state machine.

ggouaillardet:
i did not test that case, will do !

rhc54:
Thanks - I think the question here is: is the problem that the novm state machine is hanging? Or that singletons should instead kickoff the full VM?

rhc54:
I've tested singletons now that the --novm option has been fixed and they work just fine, so I'm closing this PR.

PR #3441, opened 2017-05-02
Topic/datatype

bosilca:
This started as a fix for #3439 but then evolved into a complete redesign of the handling of some datatype and convertor internal structures (mainly the array of counts for predefined types).

### DIFF
--- a/opal/datatype/opal_datatype_get_count.c
+++ b/opal/datatype/opal_datatype_get_count.c
@@ -39,9 +39,9 @@ ssize_t opal_datatype_get_element_count( const opal_datatype_t* datatype, size_t
     /* Normally the size should be less or equal to the size of the datatype.
      * This function does not support a iSize bigger than the size of the datatype.
      */
-    assert( (uint32_t)iSize <= datatype->size );
-    DUMP( "dt_count_elements( %p, %d )\n", (void*)datatype, iSize );
-    pStack = (dt_stack_t*)alloca( sizeof(dt_stack_t) * (datatype->btypes[OPAL_DATATYPE_LOOP] + 2) );
+    assert( iSize <= datatype->size );
+    DUMP( "dt_count_elements( %p, %ul )\n", (void*)datatype, (unsigned long)iSize );
+    pStack = (dt_stack_t*)alloca( sizeof(dt_stack_t) * (datatype->loops + 2) );
     pStack->count    = 1;
     pStack->index    = -1;
     pStack->disp     = 0;
@@ -93,9 +93,7 @@ int32_t opal_datatype_set_element_count( const opal_datatype_t* datatype, size_t
     /**
      * Handle all complete multiple of the datatype.
      */
-    for( pos_desc = 4; pos_desc < OPAL_DATATYPE_MAX_PREDEFINED; pos_desc++ ) {
-        local_length += datatype->btypes[pos_desc];
-    }
+    local_length = datatype->nbElems;
     pos_desc = count / local_length;
     count = count % local_length;
     *length = datatype->size * pos_desc;
@@ -104,7 +102,7 @@ int32_t opal_datatype_set_element_count( const opal_datatype_t* datatype, size_t
     }
 
     DUMP( "dt_set_element_count( %p, %d )\n", (void*)datatype, count );
-    pStack = (dt_stack_t*)alloca( sizeof(dt_stack_t) * (datatype->btypes[OPAL_DATATYPE_LOOP] + 2) );
+    pStack = (dt_stack_t*)alloca( sizeof(dt_stack_t) * (datatype->loops + 2) );
     pStack->count    = 1;
     pStack->index    = -1;
     pStack->disp     = 0;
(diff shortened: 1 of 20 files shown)


bosilca:
@ggouaillardet can you try this in a heterogeneous environment ?

ibm-ompi:

 Gist: https://gist.github.com/805cc76376493b5d175dca9ac64833a2

ibm-ompi:

 Gist: https://gist.github.com/d227a5018dfdd6cb9d3336835b9338bd

ibm-ompi:

 Gist: https://gist.github.com/30256b2e57e9a3a40fc8ef555877a110

ggouaillardet:
@bosilca this PR build but all tests crash
i will be off until Monday, and i will unlikely have any chance to review this before.
once this is fixed, do you plan to backport it to the release branches ? or make a simpler fix just for the release branches ?

bosilca:
I haven't checked yet, but I think this patch would allow us to drastically reduce the size of the predefined datatypes. If we implement the reduction, we will change the ABI, so this might only be good for 3.0. For the others I will take a look at having a slimmed down version.

ibm-ompi:

 Gist: https://gist.github.com/2fce5a06ccb4313d246c882baaeb5239

ibm-ompi:

 Gist: https://gist.github.com/90bfae8e90e611662bc9637eb7aca130

ibm-ompi:

 Gist: https://gist.github.com/cc4639d8b917a0e8d3bac6fd2ed0dbd2

ibm-ompi:

 Gist: https://gist.github.com/6fd8c492a06afe924bcdeedd43ed6d5c

ibm-ompi:

 Gist: https://gist.github.com/f4a14ddc5fca9a636699d7d67d76ce26

ibm-ompi:

 Gist: https://gist.github.com/4d387051d59b342b93d537dea980d4c3

bosilca:
It turns out that in addition of being a bug fix, this patch also provide a performance boost. We have shred few nsecs from our shared memory latency from 95.450 nsecs to 91.865 nsecs on a E5-2650 v3 @ 2.30GHz.

ompi/datatype/ompi_datatype_create_darray.c
```diff
@@ -216,9 +216,14 @@ int32_t ompi_datatype_create_darray(int size,
     }
 
     /* Build up array */
+    displs[0] = st_offsets[start_loop];
```
ggouaillardet: did you really mean to commit changes to this file ?
before entering the loop, the ```st_offsets``` array is uninitialized, and this commit causes a failure in the ```datatype/darray-pack``` test from the ```ibm``` test suite

opal/datatype/opal_datatype_get_count.c
```diff
@@ -143,3 +141,54 @@ int32_t opal_datatype_set_element_count( const opal_datatype_t* datatype, size_t
...
 }
 
+/**
+ * Compute the array of counts of the predefined datatypes contained in
+ * the datatype. We have no simple way to create this array, as we only
+ * sporadically need it (when we deal with heterogeneous environments or
+ * when we use get_element_count). Thus, we will pay the cost once per
+ * datatype, but we will only update this array if/when needed.
+ */
+int opal_datatype_compute_ptypes( opal_datatype_t* datatype )
```
ggouaillardet: the ```datatype/getel``` test from the ```ibm``` test suite loops forever.

i got things working with the patch below.
```opal_datatype_get_element_count``` might have to be updated as well, though i could not write a program that causes the same endless loop

```patch
diff --git a/opal/datatype/opal_datatype_get_count.c b/opal/datatype/opal_datatype_get_count.c
index a860d5f..c7c87ea 100644
--- a/opal/datatype/opal_datatype_get_count.c
+++ b/opal/datatype/opal_datatype_get_count.c
@@ -169,10 +169,13 @@ int opal_datatype_compute_ptypes( opal_datatype_t* datatype )
     while( 1 ) {  /* loop forever the exit condition is on the last OPAL_DATATYPE_END_LOOP */
         if( OPAL_DATATYPE_END_LOOP == pElems[pos_desc].elem.common.type ) { /* end of the current loop */
             if( --(pStack->count) == 0 ) { /* end of loop */
-                stack_pos--; pStack--;
-                if( stack_pos == -1 ) return 0;  /* completed */
+                if( stack_pos == 0 ) return 0;  /* completed */
+                stack_pos--;
+                pStack--;
+                pos_desc++;
+            } else {
+                pos_desc = pStack->index + 1;
             }
-            pos_desc = pStack->index + 1;
             continue;
         }
         if( OPAL_DATATYPE_LOOP == pElems[pos_desc].elem.common.type ) {
```

ompi/datatype/ompi_datatype_create_darray.c
bosilca: this was not supposed to get pushed. It is from another patch, but it is wrong and incomplete. I will remove.

opal/datatype/opal_datatype_get_count.c
bosilca: Thanks for noticing. The same pattern appears in several location, I fixed them all. Please review the updated version.

ggouaillardet:
@bosilca this looks good to me
i also tried in an hetero environment and could not find any issue

jsquyres:
bot:ompi:retest

jsquyres:
Note that per discussions, this PR caused master to become ABI-incompatible with the v3.x (which will shortly be renamed to be v3.0.x) branch.  That's not a *problem* -- it just means that the next release series with be v4.0.x (vs. v3.1.x.).

bosilca:
We have not released yet the 3.0. Why not including this change before the official release? This PR is not only a performance fix, but it addresses a real issue (identified by Gilles in #3439).

jsquyres:
@bwbarrett @hppritcha What say you for v3.0.0?

bwbarrett:
I'm not happy about a big DDT patch 2 weeks before we do our first RC.  Would have been better to fix the bug without all the other work, but I suppose I'm not going to have a choice, am I?

bosilca:
I like your pragmatism ;) In the defense of the patch, it looks big because it changes all the accesses to a datatype field (from ".btypes" into "->ptypes"), but the logical changes are rather small.

Technically it is possible to make a patch for the 3.x that does not have the  dynamic ptypes array. This will address the problem Gilles found, but: 1) the code will diverge between 3.x and master (and additional issues might be difficult to fix); 2) it will also require 47 * sizeof(size_t) extra bytes per datatype (including the predefined); and 3) the homogeneous case will not be as streamlined as with this patch. Your call.

hppritcha:
I tried to apply this patch to v3.x and it doesn't patch cleanly.

bosilca:
Indeed, we didn't made the change from OPAL_PTRDIFF_TYPE to ptrdiff_t in master. I have prepared a PR for you. https://github.com/open-mpi/ompi/pull/3504

PR #3444, opened 2017-05-04
pmix/configure.m4: always use embedded mode

jsquyres:
Looks like embedded mode was mistakenly disabled when `--with-devel-headers` was specified.

I don't know offhand which branches this affects -- assumedly at least v3.x.  I did not check the v2.x or v2.0.x branches.


jsquyres:
Looks like the same false failure we see from Mellanox periodically.

bot:mellanox:retest

Issue #3443, opened 2017-05-03
openmpi 2.1.0 fails to build on s390x


opoplawski:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

2.1.0

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Building Fedora openmpi package

### Please describe the system on which you are running

* Operating system/version: Fedora rawhide

-----------------------------

## Details of the problem

```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../../opal/include -I../../../../ompi/include -I../../../../oshmem/include -I../../../../opal/mca/hwloc/hwloc1112/hwloc/include/private/autogen -I../../../../opal/mca/hwloc/hwloc1112/hwloc/include/hwloc/autogen -I../../../../ompi/mpiext/cuda/c -I../../../.. -I../../../../orte/include -I/builddir/build/BUILD/openmpi-2.1.0/opal/mca/event/libevent2022/libevent -I/builddir/build/BUILD/openmpi-2.1.0/opal/mca/event/libevent2022/libevent/include -DNDEBUG -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -march=zEC12 -mtune=z13 -finline-functions -fno-strict-aliasing -pthread -MT mca_btl_sm_la-btl_sm_component.lo -MD -MP -MF .deps/mca_btl_sm_la-btl_sm_component.Tpo -c btl_sm_component.c  -fPIC -DPIC -o .libs/mca_btl_sm_la-btl_sm_component.o
btl_sm_component.c: In function 'create_rndv_file':
btl_sm_component.c:631:5: warning: ignoring return value of 'asprintf', declared with attribute warn_unused_result [-Wunused-result]
     asprintf(&tmpfname, "%s.tmp", fname);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from btl_sm.c:45:0:
../../../../opal/include/opal/sys/cma.h:89:2: error: #error "Unsupported architecture for process_vm_readv and process_vm_writev syscalls"
[...]

PR #3449, opened 2017-05-05
v2.x: opal: add support for s390 and s390x architectures

nmorey:

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1085,7 +1085,12 @@ AC_DEFUN([OPAL_CONFIG_ASM],[
             fi
             OPAL_GCC_INLINE_ASSIGN='"1: li %0,0" : "=&r"(ret)'
             ;;
-
+        s390-*)
+            opal_cv_asm_arch="S390"
+            ;;
+        s390x-*)
+            opal_cv_asm_arch="S390X"
+            ;;
         sparc*-*)
             # SPARC v9 (and above) are the only ones with 64bit support
             # if compiling 32 bit, see if we are v9 (aka v8plus) or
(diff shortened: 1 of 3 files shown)


jsquyres:
@nmorey The intent for this PR is to enable Vader's CMA support to compile on S390[X] platforms, is that correct?  Note that we usually commit such fixes to `master` first, and then cherry-pick them to the relevant release branches.  Can you open a corresponding PR against `master`?

@bosilca @hjelmn Can you have a quick look?  Looks fairly straightforward.

jsquyres:
Cross-reference: #3443

nmorey:
@jsquyres Yes. I had to fix 2.1.0 build for SUSE so I sent the patch as is but I can probably port it to master.

jsquyres:
@nmorey Excellent; thanks!

bwbarrett:
@jsquyres, we shouldn't merge this patch to 2.x until it's 1) in master and 2) there's a PR to 3.x as well...

nmorey:
@bwbarrett I'll do one for 3.x once master gets accepted then :)

bwbarrett:
@nmorey: yes, the usual path for PRs in Open MPI is master then to the branches.  @jsquyres, @hppritcha, and I are currently trying to make sure that we don't accidentally ship something in 2.1, but then not in 3.0, so the comment was as much for us as anyone :).

jsquyres:
#3454 was just merged to master; it is a slightly updated version of this PR (i.e., added a comment).

jsquyres:
@hppritcha Do you have an opinion on whether this should go to v2.1.1 or v2.1.2?

jsquyres:
Corresponding master PR: #3454

jsquyres:
Per discussion with @bwbarrett, this will go into v3.x, so let's go ahead and take this in v2.x.

PR #3453, opened 2017-05-05
v2.x: ompi/errhandler: cherry-pick missing commit

jsquyres:
Cherry pick part of fe68f230991 to avoid freeing a bogus string in an
errhandler case.

Note that this PR is not necessary for v3.x; the fix is already on master (and was there when v3.x was branched).


jsquyres:
@hppritcha I'm inclined to let this one in to v2.1.1 because it's a segv in a default error handler.  Thoughts?

jsquyres:
@hppritcha Forgot to ask you about this one one the webex today -- what do you think about this for v2.1.1?  I think it might be worthwhile because it's a segv in a default error handler.

hppritcha:
yeah let's take this.

PR #3454, opened 2017-05-05
master: opal: add support for s390 and s390x architectures

nmorey:


jsquyres:
@hjelmn @bosilca This is the master version of #3449.

bosilca:
From what I see in the code there is no practical difference between s390 and s390x from OMPI perspective. I wonder why are we creating 2 defines?

nmorey:
It is not completely necessary now but as they are 31 and 64 bit arch, I anticipated some differences that may appear later

jsquyres:
Thanks @nmorey!

PR #3456, opened 2017-05-05
Wtime fix for v1.10

sjeaugey:
This is fixing #3003 on v1.10.


rhc54:
I'll accept this change once reviewed and cleared

PR #3468, opened 2017-05-08
Add volatile to the pointers in the list)teim structures.

### LINKED ISSUE #3450: Stall in opal_fifo test

## Background information
I'm packaging openmpi 2.1 for SUSE and end up hitting a bug (probably GCC's fault) in the test suite

OpenMPI 2.1.0

Built using tarball from the website

* Operating system/version: 
openSUSE_Tumbleweed
gcc (SUSE Linux) 6.3.1 20170202 [gcc-6-branch revision 245119]

* Computer hardware: 
i586
* Network type: 

-----------------------------

## Details of the problem
Running the opal_fifo test (through make check) stalls.
After some debugging, it appears we end up into a broken loop here (opal_fifo.h:262)
```
       if (!opal_atomic_cmpset_ptr (&fifo->opal_fifo_tail.data.item, item, &fifo->opal_fifo_ghost)) {
            while (&fifo->opal_fifo_ghost == item->opal_list_next) {
                opal_atomic_rmb ();
            }
```
Looking into the generated assembly, it looks something like this
```
=> 0x08049675 <+357>:	cmp    %edi,%eax
   0x08049677 <+359>:	je     0x8049675 <thread_test+357>

```
which unless I'm mistaken means that GCC cached the value and doesn't load them from memory anymore.

The rmb used comes from gcc builtins.
Simply adding this:
```
diff --git a/opal/include/opal/sys/gcc_builtin/atomic.h b/opal/include/opal/sys/gcc_builtin/atomic.h
index 82b75f47d8..eea743503c 100644
--- a/opal/include/opal/sys/gcc_builtin/atomic.h
+++ b/opal/include/opal/sys/gcc_builtin/atomic.h
@@ -51,6 +51,9 @@ static inline void opal_atomic_mb(void)
 
 static inline void opal_atomic_rmb(void)
 {
+#if OPAL_ASSEMBLY_ARCH == OPAL_IA32
+    __asm__ __volatile__("": : :"memory");
+#endif
     __atomic_thread_fence (__ATOMIC_ACQUIRE);
 }
```
fixes the issue.

This really seems like a GCC bug, but I figured it might be worth notifying you.

bosilca:
In addition to making the code correct, it also makes it faster.
This commit fixes #3450.

### DIFF
--- a/opal/class/opal_lifo.h
+++ b/opal/class/opal_lifo.h
@@ -48,7 +48,7 @@ union opal_counted_pointer_t {
         /** update counter used when cmpset_128 is available */
         uint64_t counter;
         /** list item pointer */
-        opal_list_item_t *item;
+        volatile opal_list_item_t * volatile item;
     } data;
 #if OPAL_HAVE_ATOMIC_CMPSET_128 && HAVE_OPAL_INT128_T
     /** used for atomics when there is a cmpset that can operate on
@@ -138,14 +138,14 @@ static inline opal_list_item_t *opal_lifo_push_atomic (opal_lifo_t *lifo,
  */
 static inline opal_list_item_t *opal_lifo_pop_atomic (opal_lifo_t* lifo)
 {
+    opal_counted_pointer_t old_head;
     opal_list_item_t *item;
 
     do {
-        opal_counted_pointer_t old_head;
 
         old_head.data.counter = lifo->opal_lifo_head.data.counter;
         opal_atomic_rmb ();
-        item = old_head.data.item = lifo->opal_lifo_head.data.item;
+        old_head.data.item = item = (opal_list_item_t*)lifo->opal_lifo_head.data.item;
 
         if (item == &lifo->opal_lifo_ghost) {
             return NULL;
--- a/opal/class/opal_list.h
+++ b/opal/class/opal_list.h
@@ -103,9 +103,9 @@ struct opal_list_item_t
 {
     opal_object_t super;
     /**< Generic parent class for all Open MPI objects */
-    volatile struct opal_list_item_t *opal_list_next;
+    volatile struct opal_list_item_t * volatile opal_list_next;
     /**< Pointer to next list item */
-    volatile struct opal_list_item_t *opal_list_prev;
+    volatile struct opal_list_item_t * volatile opal_list_prev;
     /**< Pointer to previous list item */
     int32_t item_free;
 


jsquyres:
@bosilca This test is failing at Travis because it is generating MB worth of compile logs containing these warnings:

```
In file included from ../../../../../opal/class/opal_free_list.h:28:0,
                 from ../../../../../ompi/request/request.h:35,
                 from ../../../../../ompi/mca/coll/coll.h:73,
                 from ../../../../../ompi/mca/coll/self/coll_self.h:29,
                 from ../../../../../ompi/mca/coll/self/coll_self_alltoallw.c:26:
../../../../../opal/class/opal_lifo.h: In function ‘opal_lifo_pop_atomic’:
../../../../../opal/class/opal_lifo.h:148:14: warning: assignment discards ‘volatile’ qualifier from pointer target type [enabled by default]
         item = old_head.data.item = lifo->opal_lifo_head.data.item;
              ^
```

(Travis ultimately fails because the compile exceeds the max log size!)

Are we sure that the commit on this PR is correct?

jsquyres:
Looks like I exaggerated, sorry -- there's not MB of logs with these kinds of warning messages.  But there are at least a few dozen of these "argument discards `volatile` qualifier..." warning messages.

bosilca:
@jsquyres my compiler was more tolerant when casting a "volatile T*" into a "T*". I have fixed the warning (and travis seems clean again).

rhc54:
forgive my ignorance: could you please educate me as to why the need for double volatile declarations? I'm not objecting - just truly asking to be educated.

bosilca:
My understanding is that there are 3 combinations possible:
 - a pointer to a volatile data (2 syntax supported)
```
    volatile T* blah;
    T volatile * blah;
```
- A volatile pointer to a non-volatile data (not sure when this one might be necessary but it does exists)
```
  T * volatile blah;
```
 - volatile pointer to volatile data (with 2 syntax as above)
```
  volatile T* volatile blah;
  T volatile * volatile blah;
```

What we had before was just telling the compiler "this is a pointer to a volatile struct". We made no statements about the pointer itself, and as a result a very conservative compiler assumed that the pointer itself doesn't change. What we really need is the last form, because the pointer itself can also be changed (by another thread so the value should never be locally cached). Thus, the second ```volatile``` is an indicator that the pointer in addition to the struct should also be considered volatile, and prevents a compiler from any caching related to the pointer or the pointed struct.

rhc54:
Thanks for the explanation. I guess I'll have to read up more on that 3rd option as I'd never heard of a "volatile struct" before and I'm not sure what that means. I'll also go back to some places where I've used the volatile flag and ensure there aren't more of these to correct.

ibm-ompi:

ibm-ompi:

ibm-ompi:

PR #3455, opened 2017-05-05
Lustre configure fixes

jsquyres:
@edgargabriel This PR is a few commits to update/fix Lustre `configure.m4`:

1. Ensure that `--with-lustre` doesn't cause a compile fail (per #3447)
1. Ensure that if the user does `--with-lustre` and we can't find Lustre support, abort
1. Remove some dead code
1. Trivial style updates

Fixes #3447.

Can you review?  If it's good, I'll PR to the various 2.x/3.x branches.

PR #3482, opened 2017-05-08
v2.1.x: lustre fixes

jsquyres:
Corresponding master PR: #3455

Several `--with-lustre` configury fixes.  See #3455 for explanation.

@edgargabriel Can you review?  Thanks.

@edgargabriel @timattox Do we feel strongly that this fix should go into v2.1.1?  Or is it ok to slip to v2.1.2?


timattox:
There isn't a rush on my part to get this fix into 2.1.1, since my workaround of just specifying --with-luster=/usr worked for my immediate situation.  My concern is just that it should get fixed so others don't get avoidable build errors in this regard.

PR #3497, opened 2017-05-09
Allow all tuned MCA parameters to be modified programatically. (#2829)

bosilca:
Fix a comment in the MCA header.

### DIFF
--- a/ompi/mca/coll/tuned/coll_tuned_allgather_decision.c
+++ b/ompi/mca/coll/tuned/coll_tuned_allgather_decision.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2015 The University of Tennessee and The University
+ * Copyright (c) 2004-2017 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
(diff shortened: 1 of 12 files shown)


jsquyres:
RM's discussed on 19 June: looks like some of this is a set of new features, but it looks like there are some bug fixes, too (i.e., don't pass the address of a stack variable into registration).  @hppritcha will make a new PR with just the bug fixes from this PR.

jsquyres:
@hppritcha will also need to check the v3.x branch.

bosilca:
These are not new features, but they have been lost with the transition to the new MCA system. I merely brought them back.

jsquyres:
Fair enough -- but they're "new" to the 2.x branch, right?  If so, we probably have to forego them here (since they're a "new" feature to v2.1.x.).

bosilca:
There was a user complain on the mailing list that made me look into this. I have no idea when we lost the capability to dynamically change the collective, but you might be right, it might be a "new" feature in the 2.x.

I have no strong feeling for the v2.x, but I would rather have this capability in the 3 series.

bwbarrett:
I'm ok with the patch in 3.x when there's a CR for that.

hppritcha:
A subset of this PR will be pulled in to 2.x

Issue #3498, opened 2017-05-09
"mpirun hostname" launch failure in SLURM jobs on master and v3.x


jsquyres:
On 2017-05-09, master head (d7ebcca93f) is consistently failing to launch jobs on master and v3.x:

```
$ srun hostname
mpi009
mpi010
$ mpirun -V
mpirun (Open MPI) 3.0.0a1

Report bugs to http://www.open-mpi.org/community/help/ 
$ mpirun hostname
--------------------------------------------------------------------------
ORTE has lost communication with a remote daemon.

  HNP daemon   : [[23671,0],0] on node savbu-usnic-a
  Remote daemon: [[23671,0],1] on node mpi009

This is usually due to either a failure of the TCP network
connection to the node, or possibly an internal failure of
the daemon itself. We cannot recover from this failure, and
therefore will terminate the job.
--------------------------------------------------------------------------
```

v2.x installs with the same configure line do not have the same problem:

```
$ srun hostname
mpi009
mpi010
$ mpirun -V
mpirun (Open MPI) 2.1.1rc1

Report bugs to http://www.open-mpi.org/community/help/
$ mpirun hostname
mpi009
...
mpi010
...
```

Here's how OMPI was configured:

```
./configure --prefix=/home/jsquyres/bogus --disable-oshmem --with-usnic \
    --with-libfabric=/home/jsquyres/libfabric-current/install --enable-mpirun-prefix-by-default \
    --enable-debug --enable-mem-debug --enable-mem-profile --disable-mpi-fortran \
    --enable-debug --enable-mem-debug --enable-picky
```

jsquyres:
Interactive debugging with Ralph seems to indicate that the problem has something to do with:

* Spanning two nodes in my SLURM cluster: one with 16 cores total, one with 24 cores total
* SLURM is reporting 16 slots for each of the 2 nodes (i.e., it's under-reporting the number of slots available for the 24 core machine)
* Rmaps dies somewhere where it is not emitting an verbose or error message to tell us why it is dying

@rhc54 is going to add some more verbose statements in this area so that we can try again.

rhc54:
@jsquyres Don't worry about testing this - I can now replicate on my hetero system here.

jsquyres:
@rhc54 Sorry, it doesn't look like #3502 fixed the issue:

```
$ srun hostname
mpi010
mpi009
$ mpirun -V
mpirun (Open MPI) 4.0.0a1

[...]

PR #3502, opened 2017-05-09
Fix nidmap computation to deal with hetero nodes

rhc54:


rhc54:
@jsquyres Don't run this branch on MTT or it will generate a lot of output. However, it would be good to rerun those cmds and nodes we were using with this so we can track down the source of the error.

rhc54:
Refs #3498

rhc54:
Fixes #3498

rhc54:
<snicker> here is the reported reason for this PR to fail the PR checker:
```
Aborted build: 5 minutes without output.
```

bwbarrett:
Interesting place for that to hang.  I can't see how this patch would cause an infinite loop on autotools running :/.  @hppritcha, can you make sure the Cray isn't totally wonky?

rhc54:
I'm going to assume that something is borked with the cray tester, and go ahead on the basis that everything else checks out okay.

PR #3506, opened 2017-05-10

### LINKED ISSUE #1089: need ability to not use --enable-new-dtags option in linker scripts.

-Adam

@adammoody

hppritcha:
causes issues.

To disable use of --enable-new-dtags in the Open MPI compiler
wrappers, add the following to the Open MPI configury line

--disable-wrapper-runpath

Fixes #1089

Filer of issue verified the fix works for him.

### DIFF
--- a/config/opal_setup_wrappers.m4
+++ b/config/opal_setup_wrappers.m4
@@ -130,6 +130,16 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
                   [enable rpath/runpath support in the wrapper compilers (default=yes)])])
     AS_IF([test "$enable_wrapper_rpath" != "no"], [enable_wrapper_rpath=yes])
     AC_MSG_RESULT([$enable_wrapper_rpath])
+
+    AC_MSG_CHECKING([if want wrapper compiler runpath support])
+    AC_ARG_ENABLE([wrapper-runpath],
+                  [AS_HELP_STRING([--enable--wrapper-runpath],
+                  [enable runpath in the wrapper compilers if linker supports it (default: enabled,  unless wrapper-rpath is disabled).])])
+    AS_IF([test "$enable_wrapper_runpath" != "no"], [enable_wrapper_runpath=yes])
+    AC_MSG_RESULT([$enable_wrapper_runpath])
+
+    AS_IF([test "$enable_wrapper_rpath" = "no" && test "$enable_wrapper_runpath" = "yes"],
+          [AC_MSG_ERROR([--enable-wrapper-runpath cannot be selected with --disable-wrapper-rpath])])
 ])
 
 # Check to see whether the linker supports DT_RPATH.  We'll need to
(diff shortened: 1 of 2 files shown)


hppritcha:
@jsquyres please check out when you have a chance.  LLNL would like this fix.

config/opal_setup_wrappers.m4
```diff
@@ -205,9 +211,9 @@ EOF
 
     OPAL_VAR_SCOPE_POP
 
-    # If we found RPATH support, check for RUNPATH support, too
-    AS_IF([test "$WRAPPER_RPATH_SUPPORT" = "rpath"],
-          [OPAL_SETUP_RUNPATH])
+       # If we found RPATH support, check for RUNPATH support, too
+       AS_IF([test "$WRAPPER_RPATH_SUPPORT" = "rpath"],
+             [OPAL_SETUP_RUNPATH])
```
jsquyres: What is this indent change for?

config/opal_setup_wrappers.m4
```diff
@@ -130,6 +130,12 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
                   [enable rpath/runpath support in the wrapper compilers (default=yes)])])
     AS_IF([test "$enable_wrapper_rpath" != "no"], [enable_wrapper_rpath=yes])
     AC_MSG_RESULT([$enable_wrapper_rpath])
+
+    AC_ARG_ENABLE([new-dtags],
+                  [AS_HELP_STRING([--enable-new-dtags],
+                  [enable new-dtags in the wrapper compilers if linker supports it (default=yes)])])
+    AS_IF([test "$enable_new_dtags" != "no"], [enable_new_dtags=yes])
+    AC_MSG_RESULT([$enable_new_dtags])
```
jsquyres: There's an `AC_MSG_RESULT` without a corresponding `AC_MSG_CHECKING`.

config/opal_setup_wrappers.m4
```diff
@@ -130,6 +130,12 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
                   [enable rpath/runpath support in the wrapper compilers (default=yes)])])
     AS_IF([test "$enable_wrapper_rpath" != "no"], [enable_wrapper_rpath=yes])
     AC_MSG_RESULT([$enable_wrapper_rpath])
+
+    AC_ARG_ENABLE([new-dtags],
```
jsquyres: Probably need `wrapper` in the option name somewhere, to mirror the form of `--enable-wrapper-rpath` (because this option *only* affects the wrapper compiler flags, right?).

Idle question: should we call this option `--enable-new-dtags` or `--enable-runpath`?  "New dtags" is that the GNU linker calls it, but the generic name is "runpath".  I.e., I'm wondering if `--enable-runpath` would be a bit more user friendly...?  (this is an open question -- I'm open to opinions here)

config/opal_setup_wrappers.m4
```diff
@@ -130,6 +130,12 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
...
+                  [AS_HELP_STRING([--enable-new-dtags],
+                  [enable new-dtags in the wrapper compilers if linker supports it (default=yes)])])
```
jsquyres: This option is moot if `--disable-wrapper-rpath`, right?  If so, that should be included in the help message.

hppritcha:
@jsquyres take a look now

config/opal_setup_wrappers.m4
```diff
@@ -220,18 +228,19 @@ EOF
...
     LDFLAGS_save=$LDFLAGS
     LDFLAGS="$LDFLAGS -Wl,--enable-new-dtags"
-    AC_LANG_PUSH([C])
-    AC_LINK_IFELSE([AC_LANG_PROGRAM([], [return 7;])],
-                   [WRAPPER_RPATH_SUPPORT=runpath
-                    runpath_args="-Wl,--enable-new-dtags"
-                    AC_MSG_RESULT([yes (-Wl,--enable-new-dtags)])],
-                   [AC_MSG_RESULT([no])])
-    AC_LANG_POP([C])
+    AS_IF([test x"$enable_wrapper_runpath" = x"yes"],
```
jsquyres: Doesn't this change the default to only enable runpath if it is explicitly asked for? If you're changing the default, then you'll also need to update README. But I'm not sure that we want to change the default.

I thought you just wanted the ability to disable runpath for some specific platforms at LANL...?

hppritcha: no because by default ```$enable_wrapper_runpath``` will be set to yes.  hmm... maybe need to make it no though if enable_wrapper_rpath is also no.  I'll change that.  Forgot about the readme.  I'll fix that too.

jsquyres: Ah, because of `AS_IF([test "$enable_wrapper_rpath" != "no"], [enable_wrapper_rpath=yes])`, up above.  Got it.

hppritcha:
@jsquyres added a blurb in the readme and check for condition 1.

hppritcha:
@jsquyres when you have a chance could you check this again?

hppritcha:
@jsquyres i'd like to merge this before end of month.

PR #3514, opened 2017-05-10
PSM and PSM2 MTLs check on the max message size allowed by API.

matcabral:
OMPI send and receive mesages use size_t for the lenght while PSM and PSM2
psm(2)mq_send/receive use uint32_t. Type size_t is 64 bits in 64 bits arch.
Therefore, this patch adds a sanity check on the lenght of the message
and fails gracefully.

(cherry picked from commit 644641d06ffedda2c798ea1fbfcc1e782ca880f6)

### DIFF
--- a/ompi/mca/mtl/psm/help-mtl-psm.txt
+++ b/ompi/mca/mtl/psm/help-mtl-psm.txt
@@ -37,7 +37,10 @@ Unable to post application receive buffer (psm_mq_irecv).
 
   Error:      %s
   Buffer:     %p
-  Length:     %d
+  Length:     %llu
 #
 [path query mechanism unknown]
 Unknown path record query mechanism %s. Supported mechanisms are %s.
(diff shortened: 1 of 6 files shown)


matcabral:
@hppritcha I lack permissions to merge this. Would you please merge? thanks,

hppritcha:
Bryan and I will merge on Monday.
Just out of curiosity, what happens if the OFI MTl is used?  Will the PSM2 libfabric provider handle this correctly?

matcabral:
In the provider, this is "handled" at fi_getinfo time: 
https://github.com/ofiwg/libfabric/blob/master/prov/psm2/src/psmx2.h
...
#define PSMX2_MAX_MSG_SIZE	((0x1ULL << 32) - 1)
...
If the provider is already advertising the max size supported, it should not add a check in every message.

PR #3515, opened 2017-05-10
Add some more debug output

rhc54:


rhc54:
Fixes #3498

rhc54:
I confess I'm finding these testers rather confusing. When I look at a test log and see everything ran okay, but then at the very end the build got marked as "failed" with nothing to show why, I lose confidence in the tester.

Issue #3522, opened 2017-05-12
master: "make check" unpack_hetero datatype test segv


jsquyres:
When configured statically on a Linux x86_64 machine, `make check` fails the `unpack_hetero` test with a segv.  Here's the backtrace:

```
(gdb) bt
#0  0x000000000040fd7c in opal_datatype_compute_ptypes (datatype=0x5810c0 <opal_datatype_int4>) at opal_datatype_get_count.c:163
#1  0x000000000040b714 in opal_datatype_compute_remote_size (pData=0x5810c0 <opal_datatype_int4>, sizes=0x7f48c8) at opal_convertor.c:456
#2  0x000000000040b82f in opal_convertor_compute_remote_size (pConvertor=0x7f4760) at opal_convertor.c:483
#3  0x000000000040ba3e in opal_convertor_prepare_for_recv (convertor=0x7f4760, datatype=0x5810c0 <opal_datatype_int4>, count=1, pUserBuf=0x7fffffffca40) at opal_convertor.c:569
#4  0x0000000000405d3d in main (argc=1, argv=0x7fffffffcb68) at unpack_hetero.c:61
```

The code in question is:

```
163 =>  datatype->ptypes = (size_t*)calloc(OPAL_DATATYPE_MAX_SUPPORTED, sizeof(size_t));
```

gdb shows:

```
(gdb) p datatype
$2 = (opal_datatype_t *) 0x5810c0 <opal_datatype_int4>
(gdb) p datatype->ptypes
$3 = (size_t *) 0x0
```

I'm not quite sure why this is a segv.  Is this read-only member, perchance?

jsquyres:
Forgot to mention -- here's the MTT showing where it is happening (and I can reproduce manually): https://mtt.open-mpi.org/index.php?do_redir=2442

ggouaillardet:
@jsquyres yes, this is likely the root cause
```c
OPAL_DECLSPEC const opal_datatype_t opal_datatype_int4 = OPAL_DATATYPE_INITIALIZER_INT4(0);
```

jsquyres:
@ggouaillardet What should it be?

ggouaillardet:
@jsquyres 

with shared libs
```
61	    if( OPAL_SUCCESS != opal_convertor_prepare_for_recv( pConv, &opal_datatype_int4, 1, rbuf ) ) {
(gdb) p &opal_datatype_int4
[...]

PR #3526, opened 2017-05-15
opal/datatype: do not compute ptypes for OPAL predefined datatypes

ggouaillardet:
Fixes open-mpi/ompi#3522

### DIFF
--- a/opal/datatype/opal_convertor.c
+++ b/opal/datatype/opal_convertor.c
@@ -451,6 +451,10 @@ opal_datatype_compute_remote_size( const opal_datatype_t* pData,
     uint32_t typeMask = pData->bdt_used;
     size_t length = 0;
 
+    if (opal_datatype_is_predefined(pData)) {
+        return sizes[pData->desc.desc->elem.common.type];
+    }
+
     if( OPAL_UNLIKELY(NULL == pData->ptypes) ) {
         /* Allocate and fill the array of types used in the datatype description */
         opal_datatype_compute_ptypes( (opal_datatype_t*)pData );


ggouaillardet:
@bosilca can you please review and merge if ok ?

OPAL predefined datatypes are declared as ```const```, and hence the ```ptypes``` should not be set since this might be mapped in a read-only region (for example when configure'd with ```--enable-static```)
also, computing the ```ptypes``` is an overkill here, so this patch is also an optimization.
(if you think an ```OPAL_LIKELY()``` is a good fit here, i will update the PR)

@jsquyres FYI

jsquyres:
@ggouaillardet Ah, now I understand your earlier comments on #3522 better.  Thanks.

ggouaillardet:
@jsquyres next time i will try to use more words ...

bosilca:
This is strange. We alter the content of the OPAL predefined datatypes in opal_datatype_init (via the opal_datatype_basicDatatypes array), and that part of the code didn't trigger any segfault.

ggouaillardet:
@bosilca this is ok for me
```c
    for( i = OPAL_DATATYPE_FIRST_TYPE; i < OPAL_DATATYPE_MAX_PREDEFINED; i++ ) {
        datatype = opal_datatype_basicDatatypes[i];

        /* All of the predefined OPAL types don't have any GAPS! */
        datatype->desc.desc[0].elem.common.flags = OPAL_DATATYPE_FLAG_PREDEFINED |
                                                   OPAL_DATATYPE_FLAG_DATA |
                                                   OPAL_DATATYPE_FLAG_CONTIGUOUS |
                                                   OPAL_DATATYPE_FLAG_NO_GAPS;
        ...
```
```datatype``` points to a read-only region __but__ ```datatype->desc.desc``` is a statically initialized once for all pointer to somewhere in ```opal_datatype_predefined_elem_desc```, and that is not read-only memory
```c
OPAL_DECLSPEC extern union dt_elem_desc opal_datatype_predefined_elem_desc[2 * OPAL_DATATYPE_MAX_PREDEFINED];
```

since we do not modify the pointer (read only) but the pointed object (read/write), this is ok.

makes sense ?

bosilca:
thanks @ggouaillardet for pointing this out. Indeed most of the datatype is const with the exception of the ```desc``` and which points to a non-const array.

I also found an issue in my code in the OPAL_DATATYPE_INIT_PTYPES_ARRAY_UNAVAILABLE macro. It should be ```(size_t[OPAL_DATATYPE_MAX_PREDEFINED]){ [OPAL_DATATYPE_MAX_PREDEFINED-1] = 0 }``` instead of NULL. So, we have 2 choices:
 - we accept NULL ptypes for all OPAL predefined datatypes (this would not work for composed datatypes). This saves some memory for all the predefined types, but will require @ggouaillardet test.
 - we have an array of ptypes for all OPAL predefined datatypes, and we use this array when computing the ptypes for other types.

Looking at this I wonder if we don't have a bug in opal_datatype_compute_ptypes for predefined composed datatype.

ggouaillardet:
@bosilca what do you mean by "predefined composed datatypes" ?

```MPI_SHORT_INT``` and friends are predefined from an ompi point of view, but they are not from an opal point of view. unless i am missing something, NULL ptypes looks good to me for __opal__ predefined datatypes

PR #3528, opened 2017-05-15
mtl-portals4: in rendezvous, reissue PtlGet() if it fails

### LINKED ISSUE #173: Portals 4 get race
In the rendez-vous protocol, we post an ME before sending the Put, so
that if the message is truncated the receiver can perform a Get on the
ME to get the rest of the message.

Currently, the MTL does not wait for the EVENT_LINK event of the ME
Append. Using our simulator and implementation of Portals 4, we fell in
the case where the Get arrives before the ME is linked.

two solutions to solve this :
- Wait for the EVENT_LINK. Of course, this will work, but add latency
  in a semi-critical path.
- Re-issue the Get whenever the Get is Nacked. Since this won't happen
  often (maybe it will never happen on real hardware), it may be a better
  solution, performance-wise.

tkordenbrock:
This commit fixes a race condition in the rendezvous protocol.  The
race occurs because the sender does not wait for the link event on the
send buffer.  Even though this has not been seen in the wild, it is
possible for the receiver to issue the PtlGet() before the ME is
linked which causes a NAK at the receiver.  This commit resolves this
race by reissuing the PtlGet() when a NAK occurs.

Closes #173.

@regrant - Please review.

### DIFF
--- a/ompi/mca/mtl/portals4/mtl_portals4.h
+++ b/ompi/mca/mtl/portals4/mtl_portals4.h
@@ -71,6 +71,9 @@ struct mca_mtl_portals4_module_t {
     /* free list of message for matched probe */
     opal_free_list_t fl_message;
 
+    /* free list of rendezvous get fragments */
+    opal_free_list_t fl_rndv_get_frag;
+
     /** Network interface handle for matched interface */
     ptl_handle_ni_t ni_h;
     /** Limit given by portals after NIInit */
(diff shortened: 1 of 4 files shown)


tkordenbrock:
bot:retest

regrant:
👍 Reviewed and it looks good. I'll merge it into master.

PR #3529, opened 2017-05-16
v1.10: NEWS and VERSION updates

jsquyres:
@rhc54 Please review

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -82,13 +82,13 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=12:6:1
+libmpi_so_version=12:7:0
 libmpi_cxx_so_version=2:3:1
-libmpi_mpifh_so_version=12:1:1
-libmpi_usempi_tkr_so_version=6:0:1
+libmpi_mpifh_so_version=12:1:0
+libmpi_usempi_tkr_so_version=7:0:2
 libmpi_usempi_ignore_tkr_so_version=8:0:2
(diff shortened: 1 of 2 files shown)


VERSION
```diff
@@ -82,13 +82,13 @@ date="Unreleased developer copy"
...
+libmpi_so_version=12:7:0
 libmpi_cxx_so_version=2:3:1
-libmpi_mpifh_so_version=12:1:1
-libmpi_usempi_tkr_so_version=6:0:1
+libmpi_mpifh_so_version=12:1:0
+libmpi_usempi_tkr_so_version=7:0:2
 libmpi_usempi_ignore_tkr_so_version=8:0:2
 libmpi_usempif08_so_version=12:1:1
-libopen_rte_so_version=12:5:0
+libopen_rte_so_version=12:4:0
```
rhc54: @jsquyres There was a change made to the ORTE library - why did you revert the change to the revision number?
```
```

jsquyres: I only see the change to the rsh module (72459a77505b019fc0d96e11ed9c86fee1d832a1).  Was there another that I missed?

rhc54: That was it - but it still represents a change in the library and a change in behavior of that component

rhc54: or are you viewing the "library" in this case to be only the orte base code?

jsquyres: Yes, the `.so` version numbers only reflect what happens in `libopen-rte.so` proper, for non-`disable-dlopen` builds.  Hence, changes to plugins are not reflected in these version numbers.

PR #3534, opened 2017-05-16
v2.x: fortran: Fix `MPI_ARGV(S)_NULL` compilation error

kawashima-fj:
@ggouaillardet Could you review?

Corresponding master PR: #3401

(cherry picked from commit 0fcd964)


jsquyres:
Per #3535, we're ok with the ABI break (because it couldn't have worked before).

@hppritcha Good to go.

PR #3533, opened 2017-05-16
v3.x: fortran: Fix `MPI_ARGV(S)_NULL` compilation error

kawashima-fj:
@ggouaillardet Could you review?

Corresponding master PR: #3401

(cherry picked from commit 0fcd964)

PR #3535, opened 2017-05-16
v2.0.x: fortran: Fix `MPI_ARGV(S)_NULL` compilation error

kawashima-fj:
@ggouaillardet Could you review?

Corresponding master PR: #3401

(cherry picked from commit 0fcd964)


hppritcha:
I think given the nature of the bug that this PR fixes, it's okay even if it's technically an ABI change.

jsquyres:
I agree: it couldn't have worked before, so the fact that we're breaking ABI isn't too important here.

jsquyres:
@hppritcha Good to go.

@bwbarrett @hppritcha Merging this fix into v2.0.x/v2.x means you also need to merge #3533 into v3.0.x.  It's a definite bug, and the risk/impact is low.

PR #3539, opened 2017-05-16
usNIC: fix fi_ep_bind flag. FI_RECV should not be associated with address vector.

thananon:
FI_RECV is not a valid flag for address vector. This is a bug where usNIC libfabric provider does not validate the flag. There will be fix coming up from libfabric side for OMPI reverse compatibility.

Signed-off-by: Thananon Patinyasakdikul <tpatinya@utk.edu>


jsquyres:
Seems to be a false failure in the PRBC CI.

PR #3545, opened 2017-05-17
v3.x: mtl-portals4: in rendezvous, reissue PtlGet() if it fails

tkordenbrock:
This PR fixes a race condition in the rendezvous protocol.  The
race occurs because the sender does not wait for the link event on the
send buffer.  Even though this has not been seen in the wild, it is
possible for the receiver to issue the PtlGet() before the ME is
linked which causes a NAK at the receiver.  This PR resolves this
race by reissuing the PtlGet() when a NAK occurs.

(cherry picked from commit 27ee862964f7a1f4fa1b40c476200d2dacbb5658)

Signed-off-by: Todd Kordenbrock <thkgcode@gmail.com>


hppritcha:
bot:ompi:retest

PR #3551, opened 2017-05-19
1sided with some hosts single rank -- Fixes #3548

### LINKED ISSUE #3548: 1sided with some hosts single rank (osc=rdma)
If I launch a 1sided test where some of the hosts have a single rank while other hosts have multiple ranks, the test fails with a message like
```
[[57223,1],2][btl_openib_component.c:3556:handle_wc] from ibmgpu06 to: ibmgpu05
error polling LP CQ with status REMOTE ACCESS ERROR status number 10 for
wr_id 22ae5e8 opcode 128  vendor error 136 qp_idx 3
```

Examples that would fail:
```
    hostA:2,hostB:1
    hostA:1,hostB:2
```
examples that would pass:
```
    hostA:1,hostB:1
    hostA:2,hostB:2
```

I hit the same error with openib, eg
```--mca pml ob1 --mca btl self,vader,openib --mca osc rdma```
and MXM, eg
```--mca pml yalla --mca osc rdma```
although for the 1sided calls I think the above both end up in btl openib.

I've hit this issue with top of master, although I don't think it's anything new. I was running on x86-64, a mix of redhat 6 and 7, with mellanox IB cards.

The testcase I used is 1sided.c from the ompi-tests test harness at ibm/onesided/1sided.c. For convenience here's a gist of it
    https://gist.github.com/markalle/f245fef3a3294e3dcd5ae3db47a882c5
the failure happens very early, when the test is essentially doing "fence; get; fence" with each rank getting from its left-neighbor.

```
% mpicc -o x 1sided.c
% mpirun -host hostA:2,hostB:1 -np 3 --mca pml ob1 --mca btl self,vader,openib -mca osc rdma ./x
```
In the run mode where the ranks are laid out like this:
```
    hostA : [ R0, R1 ]
    hostB : [ R2 ]
```
R2 is the rank who prints the message about the error when polling.

Tracing backward a bit, I think R0's call to ompi_osc_rdma_peer_setup() for R2 is coming out wrong. Toward the bottom it has
```
    if (base_region->len) {
        if (registration_handle_size) {
            ex_peer->super.base_handle = malloc (registration_handle_size);
```
but that base_region->len is 0 so the base_handle stays NULL, where in the good runs the base_region->len is the window size and base_handle gets set. Beyond that point I'm not sure where that data comes from.

markalle:
See bug report
    https://github.com/open-mpi/ompi/issues/3548

If a 1sided test is launched -host hostA:2,hostB:1 some of the ranks
call allocate_state_single() and others call allocate_state_shared().
These functions were producing different values for module->state_size
but that's used when they lookup peer info from each other in
ompi_osc_rdma_peer_setup() so they need to all have matching
module->state_offset values.

This change adds a few unused bytes in the memory allocate_state_single()
creates so it matches.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -394,7 +394,8 @@ static int allocate_state_single (ompi_osc_rdma_module_t *module, void **base, s
 
     /* allocate anything that will be accessed remotely in the same region. this cuts down on the number of
      * registration handles needed to access this data. */
-    total_size = module->state_size + local_rank_array_size + leader_peer_data_size;
+    total_size = local_rank_array_size + module->region_size +
+        module->state_size + leader_peer_data_size;
 
     if (MPI_WIN_FLAVOR_ALLOCATE == module->flavor) {
         total_size += size;
@@ -409,7 +410,11 @@ static int allocate_state_single (ompi_osc_rdma_module_t *module, void **base, s
         return OMPI_ERR_OUT_OF_RESOURCE;
     }
 
-    module->state_offset = local_rank_array_size;
+// Note, the extra module->region_size space added after local_rank_array_size
+// is unused but is there to match what happens in allocte_state_shared()
+// This allows module->state_offset to be uniform across the ranks which
+// is part of how they pull peer info from each other.
+    module->state_offset = local_rank_array_size + module->region_size;
 
     module->state = (ompi_osc_rdma_state_t *) ((intptr_t) module->rank_array + module->state_offset);
     module->node_comm_info = (unsigned char *) ((intptr_t) module->state + module->state_size);


jsquyres:
@markalle GitHub pro tip: if you put "Fixes #3548" in the description of this PR, then when this PR is merged, it'll auto-close #3548.  You can go back and edit the description of this PR by clicking on the pencil icon in the top right of the description text box.

hjelmn:
Looks good to me. Will need to be PR'd to 2.0.x, 2.x, and 3.x.

hjelmn:
@markalle Thanks for tracking that down. This is a case that was not well tested.

bwbarrett:
bot:ompi:retest

test harness was broke for a bit yesterday...

markalle:
rebased (just had a merge of the copyright message)

PR #3554, opened 2017-05-19
v3.0.x: Remove stale "cr" MPI extension

jsquyres:
Checkpoint restart support was removed starting with commit 2192530.

A user identified in https://www.mail-archive.com/users@lists.open-mpi.org/msg31134.html that we accidentally left the cr MPI extension in the 3.0.x series, even though the rest of the checkpoint restart stuff was removed.


jsquyres:
...actually, @bwbarrett @hppritcha: I see that checkpoint restart support is still in the v3.x branch? 😲  ⚠️ 

Is that intended?

rhc54:
I suggest we remove it from master to avoid this confusion in the future. It doesn't work, won't even compile, and has no maintainer. If/when someone wants to restore it, they will probably need to do so from scratch anyway - or can pull it from history.

bwbarrett:
I don't think it was intentional that it was left in.  I'm ok removing it if that's what the community expects...

bwbarrett:
bot:ompi:retest

test harness was broke for a bit yesterday...

jsquyres:
I agree: remove it from master and from v3.0.x.  👍 

It can always be resurrected from git history if it is desired.

bwbarrett:
We're taking a different path for 3.0 and C/R, so closing this PR.

PR #3563, opened 2017-05-22
opal/stacktrace: Fix stderr target for opal_stacktrace_output

jjhursey:
* Fix a typo that prevents `-mca opal_stacktrace_output stderr` (default) option from going to `stderr`.


jsquyres:
After the fact 👍

Issue #3566, opened 2017-05-24
yalla with irregular contig datatypes


markalle:
Yalla has a macro PML_YALLA_INIT_MXM_REQ_DATA that checks if a datatype is contiguous via opal_datatype_is_contiguous_memory_layout(dt,count) and if so it selects a size and lb that presumably is what will rdma, as
                ompi_datatype_type_size(_dtype, &size); \
                ompi_datatype_type_lb(_dtype, &lb); \
    
This failed when I gave it a datatype constructed as ```[ ...]``` with extent 4. What I mean by that datatype is
```
        lens[0] = 3;
        disps[0] = 1;
        types[0] = MPI_CHAR;
        MPI_Type_struct(1, lens, disps, types, &tmpdt);
        MPI_Type_create_resized(tmpdt, 0, 4, &mydt);
```
So there are 3 chars at offset 1, and the LB is 0 and the UB is 4.
    
So their macro decides that size=4 and lb=0 and later I suppose size is getting updated to 3 for the final rdma, and so a send of a buffer ```[ 0 1 2 3 ]``` gets recved as ```[ 0 1 2 x ]```. 

If size and lb in that macro are to represent the offset and size of the contiguous thing to be sent, then I think the true lb and true extent would be the proper items to use.

In the normal case of "regular" contiguous datatypes it would all be the same. But in the case of irregular offsets/LB/UB I think the true lb and true extent are always the right thing to use for any datatype,count that could have conceivably been identified as contiguous by opal_datatype_is_contiguous_memory_layout().

markalle:
I meant to post the gist of the test case over here in the issue reoprt. This program hits the problem when run with the yalla PML at -np 2:
https://gist.github.com/markalle/24f101dfb20b6dfc5ebbe3c8aaa9505a

jladd-mlnx:
@alinask Can you take a look at this, please.

jjhursey:
@alinask @jladd-mlnx Mark posted PR #3567 with a fix if you want to review that.

jjhursey:
@markalle Do you want to keep this issue open until the release branches have the fix merged in?

markalle:
Fixed by https://github.com/open-mpi/ompi/pull/3567

yosefe:
@markalle this patch causes a failure in this test: [MPI_Type_vector_basic_f](https://github.com/open-mpi/ompi-tests/blob/master/intel_tests/src/MPI_Type_vector_basic_f.F)
[...]

PR #3567, opened 2017-05-24
yalla with irregular contig datatype -- Fixes 3566

markalle:
Yalla has a macro PML_YALLA_INIT_MXM_REQ_DATA that checks if a datatype
is contiguous via opal_datatype_is_contiguous_memory_layout(dt,count)
and if so it selects a size and lb that presumably is what will rdma, as
```
            ompi_datatype_type_size(_dtype, &size); \
            ompi_datatype_type_lb(_dtype, &lb); \
```

This failed when I gave it a datatype constructed as ```[ ...]``` with extent 4.
What I mean by that datatype is
```
    lens[0] = 3;
    disps[0] = 1;
    types[0] = MPI_CHAR;
    MPI_Type_struct(1, lens, disps, types, &tmpdt);
    MPI_Type_create_resized(tmpdt, 0, 4, &mydt);
```
So there are 3 chars at offset 1, and the LB is 0 and the UB is 4.

So that macro decides that size=4 and lb=0 and later I suppose size is getting
updated to 3 for the final rdma, and so a send of a buffer
```[ 0 1 2 3 ]``` gets recved as ```[ 0 1 2 _ ]```. I think it should use the true lb
and the true extent.

For "regular" contig datatypes it would be the same, and for the irregular
ones that are still deemed contiguous by that utility function it should
still be the right thing to use.

### DIFF
--- a/ompi/mca/pml/yalla/pml_yalla_datatype.h
+++ b/ompi/mca/pml/yalla/pml_yalla_datatype.h
@@ -3,6 +3,7 @@
  * Copyright (C) Mellanox Technologies Ltd. 2001-2011.  ALL RIGHTS RESERVED.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


markalle:
Here's a gist with a testcase example that hits the problem when run with the yalla pml:
https://gist.github.com/markalle/24f101dfb20b6dfc5ebbe3c8aaa9505a

gpaulsen:
@jladd-mlnx @artpol84 - Please take a look.

bwbarrett:
bot:ompi:retest

jjhursey:
Fixes Issue #3566

jladd-mlnx:
@yosefe Please review and bless.

jladd-mlnx:
@gpaulsen @markalle Please back-port to 2.0.x, 2.1.x, and PR to 3.0.

markalle:
I'm tempted to propose that a utility function that identifies is_contiguous(count, dt) should have a couple extra arguments so it could return "yes it's contiguous, and here's the offset and length of the contiguous region represented by that count and dt."

I think the answer is always a fairly simple true_lb and count*true_extent, so I guess the utility function doesn't really need that feature, but it could be a convenience.

bosilca:
```opal_datatype_span``` is your friend.

markalle:
Thanks, I agree, that's exactly the right way to compute the byte range for (count,dt) pairs. I just submitted another bug and pull request that I switched to use opal_datatype_span() if you want to review the use of opal_datatype_span()  (in osc rdma):
https://github.com/open-mpi/ompi/issues/3576
https://github.com/open-mpi/ompi/pull/3577

Issue #3569, opened 2017-05-24
osc rdma origin/target extent mixup


markalle:
The MPI_Accumulate call has a local count/type and a target count/type and a range check was being performed with remote_count * target_extent.

The result is
```
[ibmgpu06:6313] *** An error occurred in MPI_Accumulate
[ibmgpu06:6313] *** reported by process [140737475969025,140381006069760]
[ibmgpu06:6313] *** on win rdma window 3
[ibmgpu06:6313] *** MPI_ERR_RMA_RANGE: invalid RMA address range
[ibmgpu06:6313] *** MPI_ERRORS_ARE_FATAL (processes in this win will now abort,
[ibmgpu06:6313] ***    and potentially your MPI job)
```
when running a testcase like this
https://gist.github.com/markalle/43e955343ea8ed7dd8bf757a73dae23d

I was using MXM and -mca osc rdma to hit the failure.

markalle:
I had made a pull to fix this at
https://github.com/open-mpi/ompi/pull/3570
but independently
https://github.com/open-mpi/ompi/pull/3572
was also made for the same reason.

So I closed my 3570 pull and this bug should be fixed by 3572.

PR #3570, opened 2017-05-24
fix osc rdma origin/target extent mixup -- Fixes 3569

markalle:
The MPI_Accumulate call has a local count/type and a target count/type
and a range check was being performed with remote_count * target_extent.
I changed them both to target.


bwbarrett:
bot:ompi:retest

Looks like the ARM builder crashed last night.

jjhursey:
References Issue #3569 

@markalle add the "Fixes 3569" to the body of the PR description not the title.

bwbarrett:
This patch and https://github.com/open-mpi/ompi/pull/3572 are the same thing.  Someone should sort that out...

markalle:
I'm okay with using 3572. It looks like we're both fixing the same bug. Overall this code seems to be oriented toward frequent calls to ompi_datatype_get_extent() with extent being reused and referring to whatever was its most recent call.

I probably would have gone the other way with a target_extent and origin_extent calculated once, but I don't mind the existing coding style.

PR #3574, opened 2017-05-24
ompi/mpi: Fixes for `mpi_type_create_f90_(real|complex)`

jjhursey:
* Preserve parameter order of `(p,r)` so that a following `MPI_TYPE_GET_ENVELOPE` and `MPI_TYPE_GET_CONTENTS` see these parameters in the same order as they are supplied to `MPI_TYPE_CREATE_F90_(real|complex)`.
 * Adjust the handling of `MPI_UNDEFINED` so that a subsequent call to `MPI_TYPE_GET_ENVELOPE` and `MPI_TYPE_GET_CONTENTS` on the datatype will get `MPI_UNDEFINED` instead of `0` for the respective parameter.

Note I'll post a PR to `ompi-tests` with a test case for this to help illustrate the problem.

### DIFF
--- a/ompi/mpi/c/type_create_f90_complex.c
+++ b/ompi/mpi/c/type_create_f90_complex.c
@@ -16,6 +16,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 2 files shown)


jjhursey:
`ompi-tests` PR is here https://github.com/open-mpi/ompi-tests/pull/31

PR #3577, opened 2017-05-24
fix for buffer length check (rdma osc w/ odd datatypes)

markalle:
The osc_rdma_get_remote_segment() has the 3rd and 4th args as
* target_disp
* length
which it uses to determine if the rdma falls within the bounds of
the window or not (actually it only checks the upper bound, but I'm
okay with that).

Anyway the caller previously was passing in the length argument as
    target_datatype->super.size * target_count
which which doesn't really represent the number of bytes after target_disp
for which data exists. In particular I could create a datatype as
    { disp -4, len 4 } and use target_disp 4
and that would be bytes 0-3 of the window where the original code
would think it was bytes 4-7 and could abort at the range check.

Ive changed it to use the opal_datatype_span() function.

Fixes 3576

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_comm.c
+++ b/ompi/mca/osc/rdma/osc_rdma_comm.c
@@ -5,6 +5,7 @@
  * Copyright (c) 2016      Intel, Inc.  All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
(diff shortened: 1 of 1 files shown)


hjelmn:
Does the same thing apply to get and accumulate? It might be worth reworking the code for osc_rdma_get_remote_segment() to handle this case.

jjhursey:
@markalle This needs to go to the release branches, right?

markalle:
Since only weird tests will hit this I'm okay with it merely eventually making its way into the various OMPI releases from master if everybody else is.

PR #3587, opened 2017-05-26
pmix/pmix2x: fix errors in event abstration

hjelmn:
Parts of the pmix2x component called the event_* functions directly
instead of the opal_event_* wrappers. This is fine as long as we are
using libevent but becomes a problem with other event libraries.


hjelmn:
@rhc54 Pulled this out of my libev PR. Will try to get all the abstraction changes in place so I can try to add libev again.

hjelmn:
Yeah, I think we did get the ones in pmix. Will double-check today. If not I will open a PR on pmix.

hjelmn:
Yup. https://github.com/pmix/pmix/commit/4b49b34ae6e6aab9b0a31fa4cc6b8c38a6edc35f

PR #3606, opened 2017-05-30
NEWS: Remove +x rights

nmorey:


rhc54:
No CI required on this

rhc54:
Ah, just saw that this is for v1.10 - that series is closed.

nmorey:
Isn't there a 1.x maintenance branch ?
This is not a big issue, but it conflicts with SUSE packaging rules.
I patched in in the RPM but I figured it might be worth upstreaming.

2.x and 3.x do not have this issue

rhc54:
I'm afraid not - the v1.10 is passing out of maintenance as 3.0 is released.

Issue #3608, opened 2017-05-30
single_predefined_type with MPI_LB/UB


markalle:
The ompi_datatype_get_single_predefined_type_from_args() recurses down into a constructed type to identify what base datatype it's built from if it's built from a single type. But if the type has MPI_LB/MPI_UB, for example
```
        lens[0] = 1;
        lens[1] = 1;
        disps[0] = 0;
        disps[1] = 0;
        types[0] = MPI_LB;
        types[1] = MPI_INT;
        MPI_Type_create_struct(2, lens, disps, types, &mydt);
```
then this function will see the base type MPI_LB as differing from MPI_INT and will identify mydt as not being constructed from a single base type, so the type will be rejected for calls like MPI_Accumulate.
    
I think those "meta data" types shouldn't result in rejection like that, and the above mydt should still be identified as having a single base type of MPI_INT.

Currently an MPI_Accumulate() with a type like the above results in
[mpi03:26056] *** An error occurred in MPI_Accumulate
[mpi03:26056] *** reported by process [140644894179329,0]
[mpi03:26056] *** on win rdma window 3
[mpi03:26056] *** MPI_ERR_ARG: invalid argument of some other kind
[mpi03:26056] *** MPI_ERRORS_ARE_FATAL (processes in this win will now abort,
[mpi03:26056] ***    and potentially your MPI job)

PR #3609, opened 2017-05-30
single_predefined_type with MPI_LB/UB

markalle:
The ompi_datatype_get_single_predefined_type_from_args() recurses down
into a constructed type to identify what base datatype it's built from
if it's built from a single type.  But if the type has MPI_LB/MPI_UB,
for example
```
    lens[0] = 1;
    lens[1] = 1;
    disps[0] = 0;
    disps[1] = 0;
    types[0] = MPI_LB;
    types[1] = MPI_INT;
    MPI_Type_create_struct(2, lens, disps, types, &mydt);
```
then this function will see the base type MPI_LB as differing from MPI_INT
and will identify mydt as not being constructed from a single base type, so
the type will be rejected for calls like MPI_Accumulate.

I think those "meta data" types shouldn't result in rejection like that, and
the above mydt should still be identified as having a single base type
of MPI_INT.

Fixes https://github.com/open-mpi/ompi/issues/3608

### DIFF
--- a/ompi/datatype/ompi_datatype_args.c
+++ b/ompi/datatype/ompi_datatype_args.c
@@ -15,6 +15,7 @@
  *                         reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -837,16 +838,19 @@ ompi_datatype_t* ompi_datatype_get_single_predefined_type_from_args( ompi_dataty
                 return NULL;
             }
         }
-        if( NULL == predef ) {  /* This is the first iteration */
-            predef = current_predef;
-        } else {
-            /**
-             *  What exactly should we consider as identical types? If they are
-             * the same MPI level type, or if they map to the same OPAL datatype?
-             * In other words, MPI_FLOAT and MPI_REAL4 are they identical?
-             */
-            if( predef != current_predef ) {
-                return NULL;
+        if (current_predef != MPI_LB && current_predef != MPI_UB) {
+            if( NULL == predef ) {  /* This is the first iteration */
+                predef = current_predef;
+            } else {
+                /**
+                 *  What exactly should we consider as identical types?
+                 *  If they are the same MPI level type, or if they map
+                 *  to the same OPAL datatype? In other words, MPI_FLOAT
+                 *  and MPI_REAL4 are they identical?
+                 */
+                if( predef != current_predef ) {
+                    return NULL;
+                }
             }
         }
     }
--- a/ompi/mca/osc/base/osc_base_obj_convert.c
+++ b/ompi/mca/osc/base/osc_base_obj_convert.c
@@ -109,12 +109,16 @@ int ompi_osc_base_process_op (void *outbuf, void *inbuf, size_t inbuflen,
         bool done;
 
         primitive_datatype = ompi_datatype_get_single_predefined_type_from_args(datatype);
+        ompi_datatype_type_size (primitive_datatype, &primitive_size);
+
         if (ompi_datatype_is_contiguous_memory_layout (datatype, count) &&
             1 == datatype->super.desc.used) {
             /* NTH: the datatype is made up of a contiguous block of the primitive
              * datatype. fast path. do not set up a convertor to deal with the
              * datatype. */
-            count *= datatype->super.desc.desc[0].elem.count;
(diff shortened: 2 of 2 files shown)


bosilca:
OK. However, MPI_LB and MPI_UB are deprecated, and should not be used anymore.

bwbarrett:
bot:retest

hjelmn:
bot:retest

hjelmn:
bot:retest

bwbarrett:
bot:retest

There's no good way to remove the travis failure, but let's see what happens with the others.

@markalle, do you still want this in master?  In 3.0.x?

gpaulsen:
bot:retest

bosilca:
I take back what I said on this PR. After more thinking I do not agree with the change, simply because MPI_INT+MPI_UB is not equal with MPI_INT, as soon as a count is involved. If we make these 2 types identical, bad things will happen in all codes using MPI_LB and MPI_UB (which is a shrinking number of apps, but still enough not to ruin them completely).

gpaulsen:
@markalle, Can you comment?

markalle:
The function isn't saying the types are the same, it's saying they have the same signature (and that signature contains only a single type as its base type).

I think the only place this is used is onesided calls like MPI_Accumulate where the standard says
```Each datatype argument must be a predefined datatype or a derived datatype, where
all basic components are of the same predefined datatype. Both datatype arguments must
be constructed from the same predefined datatype. The operation op applies to elements of
that predefined type.```

**Comment by markalle** [comment]
For the signature I should be talking about the signature of the (count,type) combo though, not just the type. So if one type contained 2 ints and the other had 3 ints that would be fine too, as long as my count args were 3 and 2 for example so that they (count,type) combo had the same signature

**Comment by bosilca** [comment]
After looking more in details at the code it does indeed appear that this function is only used for the one-sided communications, to validate the arguments and then to execute the operation itself (in ompi_osc_base_process_op). In all these cases, my original complain does not apply, as the UB and LB are taken in account later in the code.

However, I think there might be a bug in the ompi_osc_base_process_op function at line 117,
```count *= datatype->super.desc.desc[0].elem.count;```
Despite the fact that the datatype is contiguous, you are not manipulating the optimized description (which would also be incorrect as we convert everything to bytes there), but instead use the user provided datatype description, which can be composed of multiple contiguous elements. Thus, using ```elem.count``` of the first available predefined type might only partially apply the operation. I would compute the number of predefined types by using the total size divided by the size of the predefined type (```primitive_datatype```).

**Comment by bosilca** [comment]
I can't write in your repo, so here is the additional patch I propose:
```diff
diff --git a/ompi/mca/osc/base/osc_base_obj_convert.c b/ompi/mca/osc/base/osc_base_obj_convert.c
index bb4641c659..e396258ce2 100644
--- a/ompi/mca/osc/base/osc_base_obj_convert.c
+++ b/ompi/mca/osc/base/osc_base_obj_convert.c
@@ -109,12 +109,16 @@ int ompi_osc_base_process_op (void *outbuf, void *inbuf, size_t inbuflen,
         bool done;

         primitive_datatype = ompi_datatype_get_single_predefined_type_from_args(datatype);
+        ompi_datatype_type_size (primitive_datatype, &primitive_size);
+
         if (ompi_datatype_is_contiguous_memory_layout (datatype, count) &&
             1 == datatype->super.desc.used) {
             /* NTH: the datatype is made up of a contiguous block of the primitive
              * datatype. fast path. do not set up a convertor to deal with the
              * datatype. */
-            count *= datatype->super.desc.desc[0].elem.count;
+            (void)ompi_datatype_type_size(datatype, &size);
+            count *= (size / primitive_size);
+            assert( 0 == (size % primitive_size) );

             /* in case it is possible for the datatype to have a non-zero lb in this case.
              * remove me if this is not possible */
@@ -125,8 +129,6 @@ int ompi_osc_base_process_op (void *outbuf, void *inbuf, size_t inbuflen,
             return OMPI_SUCCESS;
         }

-        ompi_datatype_type_size (primitive_datatype, &primitive_size);
-
         /* create convertor */
         OBJ_CONSTRUCT(&convertor, opal_convertor_t);
         opal_convertor_copy_and_prepare_for_recv(ompi_mpi_local_convertor, &datatype->super,
```

markalle:
I think I agree. I tried a few little testcases and it keeps coming out with an apparently optimized layout where item [0] has the right count, but I could certainly see that not being a guaranteed thing. Your approach seems safer.

markalle:
repushed

markalle:
Okay, geoffrey said to put it into v3.1.x as well, so I made a pull for that part too:
    https://github.com/open-mpi/ompi/pull/4302

PR #3611, opened 2017-05-30
contrib: add a release checklist template

hppritcha:
Some of us have been finding the release checklist issue
very helpful for making sure all the steps are taken to
generate a perfect tarball, and hence a wonderful experience
getting releases out.

Check a copy of this release checklist template in to the
ompi repo to simplify creating new issues with each release.

See #3557 for an example of such a release checklist

[skip ci]

### DIFF
--- a/contrib/dist/release_checklist_issue_template
+++ b/contrib/dist/release_checklist_issue_template
@@ -1,0 +1,102 @@
+## Make a perfect tarball
+
+- [ ] Update the major,minor,release,greek version numbers in `VERSION`
+- [ ] Update the `c:r:a` shared library version number(s) in `VERSION` [per the GNU Libtool shared library version number rules](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html)
+   - **NOTE:** It may be helpful to use a command like `git checkout BRANCH; git pull --rebase; git log --stat --topo-order --decorate TAG_FROM_PREVIOUS_RELEASE..HEAD` to examine the Git logs and see what has changed.
+   - [ ] *IF RELEVANT:* If this is a new backwards-compatible release on a single branch (i.e., this is `vx.y.z` where `z>1`), you probably want to examine `git log --stat --no-merges last_release_tag..this_branch_name` to see what source code files have changed (which directly impacts how to increment the `c:r:a` values).
+  - [ ] *IF RELEVANT:* If this is a new release series (e.g., `vx.y.0`), set `r` to 0 and increase `c` values by 10 compared to the first release in the prior series (i.e., `vx.(y-1).0` or `v(x-1).0.0`, as relevant).
+  - [ ] *IF RELEVANT:* If this is a new backwards-compatible release series (i.e., `vx.y.0`, where `y>1`), set`a` values to 10 so that the shared libraries will be ABI compatible with the prior release series.
+  - [ ] *IF RELEVANT:* If this is a new backwards-**IN**compatible release series (i.e., `vx.0.0`), set `a` values to 0 so that there is no possibility of users accidentally mixing shared library versions.
+- Update all documentation files (especially including dates and version numbers), including:
+  - [ ] `README`: all relevant updates, build options, etc.  Be sure to update the date near the top of the file.
+  - [ ] `AUTHORS`: via `contrib/dist/make-authors.pl` (and commit the result)
+  - [ ] `NEWS`: List all user-noticeable changes.  Similar to setting shared library versions (above):
+       - Pro tip: if this is a new release on a single branch (i.e., this is `vx.y.z` where `z>1`), you probably want to examine `git log --stat --no-merges last_release_tag..this_branch_name` to see what has changed.
+       - Pro tip: if this is a new release series (i.e., this is `vx.y.0` where `y>1`, or this is `vx.0.0`), you will need to be more creative in examining the git logs because this release is on a different branch than the prior release (`vx.(y-1).z`).  Hence, `git log ... last_release_tag..this_branch_name` will not necessarily give you need.  You may need to merge what has changed on your branch with what has changed on the prior release branch, depending on when the prior release branched from this branch.  Read the **SPECIFYING RANGES** section `gitrevisions(7)` for more details.
+  - [ ] `LICENSE`: Update the years in the copyright notices
+  - [ ] ...any other doc files that may not be included in this list
+- [ ] Make final `openmpi-x.y.z.tar.gz` and `openmpi-x.y.z.bz2` tarballs using the _make_dist_tarball_ script in contrib/dist. To do this correctly, you'll need to
+      -  log in to aws.open-mpi.org and su mpiteam
+      -  module load (one of the autconf module files)
+      -  git clone the Open MPI, cd into ompi, switch to the right branch, then
+      -  ./contrib/dist/make_dist_tarball
+
+## Ensure that the tarball actually works
+- [ ] Ensure openmpi `make distcheck` passes
+- Build and install Open MPI (from a tarball)
+  - [ ] Check that the version number and release date is correct in the installed man pages
+- [ ] Build and run the examples in the tarball
+  - [ ] MPI / C bindings
+  - [ ] MPI / C++ bindings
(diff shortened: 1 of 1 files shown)


contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
...
+   - **NOTE:** It may be helpful to use a command like `git checkout BRANCH; git pull --rebase; git log --stat --topo-order --decorate TAG_FROM_PREVIOUS_RELEASE..HEAD` to examine the Git logs and see what has changed.
+   - [ ] *IF RELEVANT:* If this is a new backwards-compatible release on a single branch (i.e., this is `vx.y.z` where `z>1`), you probably want to examine `git log --stat --no-merges last_release_tag..this_branch_name` to see what source code files have changed (which directly impacts how to increment the `c:r:a` values).
+  - [ ] *IF RELEVANT:* If this is a new release series (e.g., `vx.y.0`), set `r` to 0 and increase `c` values by 10 compared to the first release in the prior series (i.e., `vx.(y-1).0` or `v(x-1).0.0`, as relevant).
+  - [ ] *IF RELEVANT:* If this is a new backwards-compatible release series (i.e., `vx.y.0`, where `y>1`), set`a` values to 10 so that the shared libraries will be ABI compatible with the prior release series.
+  - [ ] *IF RELEVANT:* If this is a new backwards-**IN**compatible release series (i.e., `vx.0.0`), set `a` values to 0 so that there is no possibility of users accidentally mixing shared library versions.
```
rhc54: Do these instructions match those in the official libtool documentation? I have the following:
```
# These versions are maintained in accordance with the "Library
# Interface Versions" chapter from the GNU Libtool documentation:
#
#
# - If any interfaces have been added, removed, or changed since
#   the last update, increment current, and set revision to 0.
#
# - If any interfaces have been added since the last public release,
#   then increment age.
#
# - If any interfaces have been removed since the last public release,
#   then set age to 0.
#
```

contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
...
+- [ ] Ensure openmpi `make distcheck` passes
+- Build and install Open MPI (from a tarball)
+  - [ ] Check that the version number and release date is correct in the installed man pages
+- [ ] Build and run the examples in the tarball
+  - [ ] MPI / C bindings
+  - [ ] MPI / C++ bindings
+  - [ ] MPI / All three Fortran bindings (you'll need newer compiler, e.g. gcc 4.9 or higher)
+  - [ ] OSHMEM / C bindings (don't run if not using ikrit)
+  - [ ] OSHMEM / Fortran bindings (don't run if not using ikrit)
+- [ ] Build and install the IBM tests from the `ompi-tests` repo
```
rhc54: I think this is overkill in terms of responsibility for the RMs - this is what MTT is for.

contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
...
+  - [ ] Run examples, make sure they still work
+  - [ ] Run IBM tests, make sure they still work
+
+## Test packaging
+- [ ] Ensure that building an Open MPI source RPM works on a RHEL system
+- [ ] Ensure that building the Open MPI binary RPM works on a RHEL system (note had to do hacking on buildrpm.sh script and it doesn't report correct information about where rpms are installed, at least not on my RHEL7.2 system running as root)
+- [ ] Install both RPMs on an RHEL system
+- [ ] Test building and running the IBM tests against the RPM-installed Open MPI
+- [ ] Test building a MPI-based program with the output from `pkg-config` with the RPM-installed `ompi.pc`
+- [ ] Test that uninstall of the rpm works
```
rhc54: Sounds like overkill - we have never supported the RPMs to that extent because nobody uses our RPMs. They get them from the packagers, and they don't use our RPM scripts (or so they say, at least)

contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
...
+    - [ ] Make the tag commit message be "Release vX.Y.Z"
+    - [ ] Double check that you tagged the Right commit, and that it exactly agrees with the hash in `VERSION` in the release tarball
+    - [ ] Triple check that you tagged the Right commit.  **TAGS ARE FOREVER!**
+- [ ] `git push --tags --dry-run` to the ompi repo.  Remove `--dry-run` when you're convinced it is correct.
+- Publish the tarballs on the Open MPI web site
+  - If Z is 0 (i.e., this is the first release in a series):
+    - [ ] `cp -r software/ompi/vCURRENT_RELEASE software/ompi/vRELEASE` (where RELEASE is for the new release X.Y and CURRENT_RELEASE is the X.Y of the current release)
+    - [ ] Edit each file in the new directory to update for the new release
+        - [ ] Edit `timeline-graph.php` to indicate relevant dates, such as branching
+    - [ ] Remove all prior tarballs from the `downloads` subdirectory
```
rhc54: Oh no - definitely don't do that! Remove only the rc's from this release. Leave all prior tarballs alone.

contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
+## Make a perfect tarball
```
jsquyres: Should probably name this file with .md.

Note that I typically update this checklist with every release, too.  Where is the right place to maintain this?  I.e., is it enough to go find the issue corresponding to the last release checklist and use that?  Or should we maintain / update a list here in `contrib/dist`?

contrib/dist/release_checklist_issue_template
```diff
@@ -0,0 +1,102 @@
...
+
+- [ ] Update the major,minor,release,greek version numbers in `VERSION`
+- [ ] Update the `c:r:a` shared library version number(s) in `VERSION` [per the GNU Libtool shared library version number rules](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html)
+   - **NOTE:** It may be helpful to use a command like `git checkout BRANCH; git pull --rebase; git log --stat --topo-order --decorate TAG_FROM_PREVIOUS_RELEASE..HEAD` to examine the Git logs and see what has changed.
+   - [ ] *IF RELEVANT:* If this is a new backwards-compatible release on a single branch (i.e., this is `vx.y.z` where `z>1`), you probably want to examine `git log --stat --no-merges last_release_tag..this_branch_name` to see what source code files have changed (which directly impacts how to increment the `c:r:a` values).
+  - [ ] *IF RELEVANT:* If this is a new release series (e.g., `vx.y.0`), set `r` to 0 and increase `c` values by 10 compared to the first release in the prior series (i.e., `vx.(y-1).0` or `v(x-1).0.0`, as relevant).
```
jsquyres: The indenting changed starting with this bullet.

contrib/dist/release_checklist_issue_template
jsquyres: These instructions complement the GNU Libtool guidelines.  I documented what we do here in OMPI on top of those GNU Libtool guidelines (e.g., bump `a` by 10).

contrib/dist/release_checklist_issue_template
jsquyres: I do this with the final release tarball (vs. the nightly tarballs) to make sure that it was constructed properly / and is ready for final distribution.

contrib/dist/release_checklist_issue_template
jsquyres: If we're going to have an SRPM, we should test it.  Otherwise, we shouldn't bother having the SRPM.

Users have definitely complained before when the SRM didn't work for one reason or another.

If we want to stop distributing an SRPM, that's a conversation we can have.

contrib/dist/release_checklist_issue_template
jsquyres: Yes, that bullet should be worded better: only remove the prior `rc` tarballs (that's the *intent* of this bullet -- I wrote it poorly).  All other official distribution tarballs must be remain.

hppritcha:
For what it's worth, I think we should definitely make sure one can make a source RPM, install it in one's rpm build directory, and hence build the binary rpms.  I do this regularly now for a cluster i maintain and can appreciate the importance of being able to do this.

Issue #3613, opened 2017-05-31
Master compile failure due to O_RSYNC / O_DSYNC in state_base_fns.c


bwbarrett:
Master does not compile on OS X or FreeBSD due to `O_RSYNC` and `O_DSYNC` usage in `orte/mca/state/base/state_base_fns.c`, which were added in https://github.com/open-mpi/ompi/commit/f3ab326b4ae4a7e4b7cb3f000dfee88437018788.

Jenkins failures: [FreeBSD](https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.platforms/Platform=freebsd_11.0/467/console) and [MacOS](https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.platforms/Platform=os-x/467/console).

PR #3614, opened 2017-05-31
Don't sweat the "sync" settings on file descriptors as those flags aren't apparently fully portable

rhc54:


rhc54:
Fixes #3613

PR #3619, opened 2017-05-31
sharedfp: remove the addproc component from the 3.0.x branch

edgargabriel:
the sharedfp/addproc component was not part of the 2.x series, and was not scheduled to be part of the 3.x series either.


bwbarrett:
bot:ompi:retest

Unfortunately, the "default" check was a jenkins misconfiguration and there's no good way to remove it.  It can be ignored (will never get updated).

bwbarrett:
@hppritcha I did the code review, so mind doing the RM review?

hppritcha:
we can ignore the jenkins report from "default".

PR #3631, opened 2017-06-01
WIP: consolidate some version checking code

jsquyres:
Based on conversations week of 29 May 2017:

- we note that there's 2 separate autogen version checking codes: in autogen.pl and make_dist_tarball
- regardless of decisions about what to do w.r.t. minimum enforced versions, it would be good to consolidate these codes together
- autogen.pl currently checks for versions in a "greater than or equal to" manner, make_dist_tarball checks for "exactly equal to".  So the  consolidated code will need to be able to handle both.

This first cut is a WIP.  I'm not 100% happy with it yet, but I'm out of time for this window, and need to save it somewhere to return to it later.

I added --autotools-eq=X and --autotools-ge=X options to autogen.pl, which take an arg of the form "ac:X,am:X,lt:X" where you specify the versions you want checked.  The thought was that make_dist_tarball could sepcify -eq or -ge (depending on whether --highok was specified or not) and specify the exact versions that it was looking for.

This commit does *not* adjust the default versions that autogen.pl is looking for, nor does it change the default "greater than or equal to" behavior.

I put a comment in autogen.pl about what might be a better set of CLI options.  I made an obvious syntax error after the comment so that Future Jeff will be forced to find the comment and pick up from there.  :-)

Note that the current stuff I added in this commit does not check for m4 and flex versions in autogen.pl -- but make_dist_tarball checks for these things (the flex version was quite important to distribution tarballs over time).  Need to expand to include those checks (not sure why we're checking for m4 versions, though...?  That sort of thing should be covered by Automake).

Note that it is a separate question as to whether we want make_dist_tarball and autogen.pl to -- by default -- look for the same versions (i.e., and therefore potentially only have the versions listed in one place, not two).  If the versions in one are updated to match the other, note that it would probably be good to put comments in both places reminding Future Editors to change *both* places. 

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -54,6 +54,8 @@ my $help_arg = 0;
 my $platform_arg = 0;
 my $include_arg = 0;
 my $exclude_arg = 0;
+my $autotools_eq_arg = 0;
+my $autotools_ge_arg = 0;
 my $force_arg = 0;
 
 # Include/exclude lists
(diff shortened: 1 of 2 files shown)


ibm-ompi:

 Gist: https://gist.github.com/efe6173ec3c34d1290a79582271a177f

ibm-ompi:

 Gist: https://gist.github.com/de85a79e1ea0cc9558abc4004b1c3845

ibm-ompi:

 Gist: https://gist.github.com/46a6bc485d1481cd0e91235ded9543e8

bwbarrett:
We should also track (somewhere, so here for now) that we should make the minimum package versions in configure.ac and autogen.pl the same (regardless of whether or not we make the make_dist_tarball and autogen.pl versions the same).

rhc54:
This generally looks good to me - appreciate the effort! I also echo @bwbarrett comment.

I see no reason to be checking m4 versions as automake/autoconf do that for us.

One possible extension I can think of would be to check for "known bad" versions that we know won't work with our code.  IIRC, there was something in the automake 1.14 series that was a problem at some point - I don't know if there might be others. Might be worth adding a little code for "future proofing" as these are discovered.

ibm-ompi:

hppritcha:
closing due to lack of activity.  reopen if there will be more work on this PR.

PR #3634, opened 2017-06-01
predefined MPI object padding: set to fixed number of bytes

### LINKED ISSUE #3610: Mysterious 32 bit MTT build failures on Absort
We're seeing odd MPI Install failures on 32 bit Absoft failures, but the log messages are not showing what the actual error is.

@cagoelz Can you shed any light, perchance, on what is failing in these builds?

https://mtt.open-mpi.org/index.php?do_redir=2446

jsquyres:
Convert the predefined MPI object padding to a fixed number of bytes (vs. a multiple of sizeof(void*)) so that the padding is the same size between 32 and 64 bit builds.  I.e., we won't have a situation where we've run out of padding in 32 bit builds but still have more space available in 64 bit builds.

Fixes #3610

@bwbarrett @hppritcha This is an ABI-changing commit (albeit pretty low risk).  If we want this for v3.0.x, it needs to be in for v3.0.0.

### DIFF
--- a/ompi/communicator/communicator.h
+++ b/ompi/communicator/communicator.h
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2006-2012 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2006-2017 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2006-2010 University of Houston.  All rights reserved.
  * Copyright (c) 2009      Sun Microsystems, Inc. All rights reserved.
  * Copyright (c) 2011-2013 Inria.  All rights reserved.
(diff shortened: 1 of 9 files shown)


jsquyres:
@bosilca I'm not 100% convinced that this is the correct solution.  Yes, it gives us a fixed padding that is the same between 32 and 64 bit builds... but is that important?

I.e., is the real fix to #3610 just to increment communicator padding by N multiples of `sizeof(void*)`?  My *only* (admittedly weak) reason for raising this issue is that by going the route in this PR (converting all the padding sizes to a raw number of bytes) is that we make the padding a bit bigger on 32 bit platforms than we really need to.

Thoughts?

bosilca:
I prefer this solution to having different paddings on 32 and 64 bits architectures to cover all the needs. If someone that cares about 32 bits proposes a better approach, we take it. Meanwhile, this is good enough.

jsquyres:
@bwbarrett @hppritcha Turns out that we don't need this PR (and 037a85a782bcc0bad5d1977384843b4d3f6702c0) in v3.0.0 because we haven't run out of (32-bit) padding yet in v3.0.0.

PR #3500, opened 2017-05-09
Allow MPI_ANY_SOURCE in MPI_Sendrecv_replace.

bosilca:
Fixes #3496.

Signed-off-by: George Bosilca <bosilca@icl.utk.edu>

PR #3635, opened 2017-06-01
MPI_Sendreceive_replace data error with > 2k msg

wlepera:
Signed-off-by: William LePera <lepera@us.ibm.com>

This bug was found during a run of the MPICH test case allpairf, which reported invalid data when using MPI_Sendreceive_replace with a buffer of 666 reals, starting with the 513th index (the first 512 were correct).  The problem occurred because MPI_Sendreceive_replace was setting the length of the receive buffer at 2048, even if it dynamically allocated a larger buffer size.

### DIFF
--- a/ompi/mpi/c/sendrecv_replace.c
+++ b/ompi/mpi/c/sendrecv_replace.c
@@ -12,6 +12,7 @@
  * Copyright (c) 2010-2012 Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


gpaulsen:
@bosilca, can you please review?  Bill Lepera is on our team.

bwbarrett:
ok to test
add to whitelist

jjhursey:
@wlepera Can you PR this into the `v3.0.x` and `v2.x` release branches?

wlepera:
@jjhursey the fix was based on the version of the code in master branch; it does not look like these changes have made their way into the v2.x or v3.0.x release branches yet.

jjhursey:
Thanks for checking. I see that George's commit:
 * https://github.com/open-mpi/ompi/commit/86a7b317a5d9f228342cddf813dffad48e83a7d1
 * https://github.com/open-mpi/ompi/pull/3500

Did not make it into the v3.0 series or any of the other releases yet. I made a note on the Issue requesting that it be PR'ed over to also include this PR.

PR #3637, opened 2017-06-01
osc/sm: fix SEGV in new info usage

hjelmn:
This commit moves the info subscribe for the blocking_fence to after
the global_state is allocated and moves setting win->w_osc_module to
before the info subscribe for alloc_shared_contig. This fixes a SEGV
caught by MTT.


rhc54:
I'm going to bring this in as the segv's are simply trashing the nightly MTT.

Issue #3616, opened 2017-05-31
hang with Put;Accumulate;Barrier osc=rdma


markalle:
I'm using master and running a testcase between two infiniband machines as
% mpirun -mca pml ob1 -mca btl openib,vader,self -mca osc rdma -host mpi03,mpi04 ./x
and can hit the bug using pml yalla as well. It needs to span two hosts to fail.

Here's a gist of the testcase:
https://gist.github.com/markalle/ccbd729df75188378d538767c0321f4e
It boils down to
```
        MPI_Accumulate(sbuf, 100, MPI_INT, to, 0, 100, MPI_INT, MPI_MAX, win);
        MPI_Barrier(MPI_COMM_WORLD);
        MPI_Put(sbuf+2000, 100, mydt, to, 2000, 100, mydt, win);
        MPI_Barrier(MPI_COMM_WORLD);
```
where mydt in the MPI_Put is non-contiguous.

The initiator of the Put and Accumulate ends up going from opal_progress() to handle_wc() where it's handling a completion callback for the Accumulate request, and hangs with the bottom of its stack trace looking like this
```
#0  ompi_osc_rdma_lock_btl_fop (module=0xb07840, peer=0xbd70f0, offset=16)
    at ../../../../../../opensrc/ompi/ompi/mca/osc/rdma/osc_rdma_lock.h:63
#1  ompi_osc_rdma_lock_btl_op (module=0xb07840, peer=0xbd70f0, offset=16)
    at ../../../../../../opensrc/ompi/ompi/mca/osc/rdma/osc_rdma_lock.h:88
#2  ompi_osc_rdma_lock_release_exclusive (module=0xb07840, peer=0xbd70f0, 
    offset=16)
    at ../../../../../../opensrc/ompi/ompi/mca/osc/rdma/osc_rdma_lock.h:342
#3  0x00007fffe95c7805 in ompi_osc_rdma_acc_put_complete (btl=0x74f7b0, 
    endpoint=0x7fffe40125d0, local_address=0x7ffff0020020, 
    local_handle=0xc4e5d8, context=0xc63d90, data=0x0, status=0)
    at ../../../../../../opensrc/ompi/ompi/mca/osc/rdma/osc_rdma_accumulate.c:113
```
But I'm getting lost beyond that.

markalle:
@hjelmn Can you look at this?

markalle:
[...]

PR #3640, opened 2017-06-01
progression discussion [work in progress]

markalle:
We have a progresion issue in openib when we're out of resources and
a completion function wants to send another message.

Currently the completion function blocks (spins calling progression)
expecting its send to happen eventually. But in the out-of-resources case
the send is queued and the needed resources won't be freed until we
return from our completion function which will allow processing to
continue through the rest of the completed work list.

I can only see three categories of fix:
1. don't let completion functions block like that
2. allow recursive handle_wc() while already inside a completion function
3. have some reserved resources for completion functions only so they can always complete

This is an example implementation of #2. This includes a conditional
recursion that gets less frequent the deeper the call stack becomes.

The biggest risk I see for #2 is that as written there is no longer a
strict ordering of completion functions. They would be initiated in the
same order they are seen in ibv_poll_cq() but you can't be sure WC1.cbfunc
has finished before WC2.cbfunc starts. I'm not sure if that's a requirement
for correctness or not.

### DIFF
--- a/opal/mca/btl/openib/btl_openib_component.c
+++ b/opal/mca/btl/openib/btl_openib_component.c
@@ -22,6 +22,7 @@
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2014      Bull SAS.  All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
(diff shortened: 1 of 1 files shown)


markalle:
The issue report around this PR including a testcase and some discussion is here:
https://github.com/open-mpi/ompi/issues/3616

Issue #3639, opened 2017-06-01
FreeBSD build failures on 2.0.x series from git


bwbarrett:
Open MPI 2.0.x is failing all Pull Requests because of a build failure when memory debugging is enabled (as it is by default when building in tree from a git checkout).  The issues is in the Fortran bindings, where there's a -I to `ompi/mpi/fortran/base`, where there is a file `strings.h`.  This file is being included instead of `/usr/include/strings.h`, which is resulting in crazy errors like:

```--- libmpi_fortran_base_la-attr_fn_f.lo ---
In file included from /usr/include/string.h:45:0,
                 from ../../../../opal/include/opal_config_bottom.h:398,
                 from ../../../../opal/include/opal_config.h:2856,
                 from ../../../../ompi/include/ompi_config.h:29,
                 from attr_fn_f.c:20:
./strings.h:42:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
     OMPI_DECLSPEC int ompi_fortran_string_f2c(char *fstr, int len, char **cstr);
                   ^
./strings.h:58:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
     OMPI_DECLSPEC int ompi_fortran_string_c2f(char *cstr, char *fstr, int len);
                   ^
./strings.h:89:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
     OMPI_DECLSPEC int ompi_fortran_argv_f2c(char *farray, int string_len,
                   ^
./strings.h:106:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
     OMPI_DECLSPEC int ompi_fortran_multiple_argvs_f2c(int count, char *array, int len,
                   ^
--- libmpi_fortran_base_la-f90_accessors.lo ---
In file included from /usr/include/string.h:45:0,
                 from ../../../../opal/include/opal_config_bottom.h:398,
                 from ../../../../opal/include/opal_config.h:2856,
                 from ../../../../ompi/include/ompi_config.h:29,
                 from f90_accessors.c:23:
./strings.h:42:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
     OMPI_DECLSPEC int ompi_fortran_string_f2c(char *fstr, int len, char **cstr);
                   ^
./strings.h:58:19: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'int'
[...]

PR #3641, opened 2017-06-02
fortran/base: rename strings.h into fortran_base_strings.h

ggouaillardet:
rename ompi/mpi/fortran/base/strings.h so it does not get pulled
when /usr/include/strings.h is expected.

Refs open-mpi/ompi#3639


ggouaillardet:
@bwbarrett FYI, all CI passed !

bwbarrett:
@ggouaillardet thanks for the patch!  Like I said in the bug report, not sure why this works on master / 3.0.x / 2.x, but not 2.0.x.  But your fix should apply to 2.0.x fairly well and fix the issue.

PR #3655, opened 2017-06-05
VERSION update for 3.0.0

hppritcha:
up the current to 40 and drop age to 0
for all *.so’s

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -82,17 +82,17 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=0:0:0
-libmpi_cxx_so_version=0:0:0
-libmpi_mpifh_so_version=0:0:0
-libmpi_usempi_tkr_so_version=0:0:0
-libmpi_usempi_ignore_tkr_so_version=0:0:0
-libmpi_usempif08_so_version=0:0:0
-libopen_rte_so_version=0:0:0
-libopen_pal_so_version=0:0:0
-libmpi_java_so_version=0:0:0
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -82,32 +82,32 @@ date="Unreleased developer copy"
...
-libmpi_usempi_tkr_so_version=0:0:0
-libmpi_usempi_ignore_tkr_so_version=0:0:0
-libmpi_usempif08_so_version=0:0:0
-libopen_rte_so_version=0:0:0
-libopen_pal_so_version=0:0:0
-libmpi_java_so_version=0:0:0
-liboshmem_so_version=0:0:0
-libompitrace_so_version=0:0:0
+libmpi_so_version=40:0:0
+libmpi_cxx_so_version=40:0:0
```
bwbarrett: dumb question for @jsquyres, do we normally bump all the major version numbers?  Did (say) libmpi_cxx change in some way that should require this?

jsquyres:
@bwbarrett For major releases, we do typically bump all the version numbers to give us runway for the rest of that release series, keeping in mind that we tend to have more than one major release series in flight concurrently.  It also further concretely solidifies (pun somewhat intended) the notion that this release is an ABI break from prior releases.

And yes, we've done it even for libraries like the C++ library (even if it hasn't changed at all) for the same two reasons: 1) give it runway in version number space for the duration of this release series, and 2) solidify the notion that this release is an ABI break from the prior series (e.g., if you have a C++ MPI app compiled against 2.x, it won't run-time link against *any* of Open MPI's libraries, not even `libmpicxx.so`).

PR #3664, opened 2017-06-07
Get the pmix/ext2x component to work. Fix a minor problem in the libevent external component.

rhc54:


rhc54:
@jjhursey @artpol84 I also tested the ext1x component against PMIx v1.2.2. It segfaults in finalize during deregister_nspace in the server as the callback function is no longer valid. My guess is that the "glue" to ORTE has changed sufficiently that the ext1x component will need updating.

artpol84:
@karasevb

PR #3666, opened 2017-06-07
Correct the external pmix configury

rhc54:


rhc54:
This is that random hang again, this time on freebsd where we don't have gstack to help us track it down.

PR #3671, opened 2017-06-07
We cannot use OFI to determine when daemons can finalize as we don't …

rhc54:
…see the "sockets" go away. So always use the OOB for the mgmt conduit - this provides the necessary termination signal AND ensures that IOF and other mgmt messages go solely across TCP.

Cleanup the way we look for matching OFI addresses by using the opal_net_samenetwork helper function. This now works for multi-network environments, but only using the socket provider


rhc54:
@anandhis

rhc54:
sigh - irrelevant Jenkins infrastructure issue.

:bot:mellanox:retest

rhc54:
bot:mellanox:retest

PR #3682, opened 2017-06-08
ompi: add support for new communicator info assertions

hjelmn:
This commit adds code to allow support for the info assertions added
by mpi-forum/mpi-issues#11. The assertions added are:
mpi_assert_no_any_tag, mpi_assert_no_any_source,
mpi_assert_exact_length, and mpi_assert_allow_overtaking.

This commit also adds support for the mpi_assert_no_any_source and
mpi_assert_allow_overtaking info keys to the ob1 pml.

### DIFF
--- a/ompi/communicator/comm_init.c
+++ b/ompi/communicator/comm_init.c
@@ -444,3 +444,40 @@ static void ompi_comm_destruct(ompi_communicator_t* comm)
                                       comm->c_f_to_c_index, NULL);
     }
 }
+
+#define OMPI_COMM_SET_INFO_FN(name, flag)       \
+    static char *ompi_comm_set_ ## name (opal_infosubscriber_t *obj, char *key, char *value) \
+    {                                                                   \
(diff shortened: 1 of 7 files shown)


hjelmn:
@markalle The info stuff seems to be working pretty well. This PR adds support for the communicator assertions that were voted on in March.

jsquyres:
Just to point out: this stuff isn't in the MPI standard yet.  But it's passed 2 votes -- there just hasn't been a new version of the MPI spec published yet.

To be clear: I think that this is totally fine to merge before the next version of the MPI spec is published (since it passed 2 votes).  Just making sure people are aware that this is technically pre-standard behavior.  I think the risk of merging this now is very low (i.e., there is risk that the Forum will change it again before it is formally published, but I think that's very unlikely).  As such, my $0.02 is that including these `mpi_` info keys is safe (i.e., let's not do the "name them `mpix_` for a while and then rename them to `mpi_` when they get officially published dance).

Issue #3683, opened 2017-06-09
OMPI/master failures with set the specifying host


karasevb:
There is failure for the follow cmd: `./mpirun --host cn06 hostname`. This is found only for `master`, it doesn't happen with `v3.x` and `v3.0.x`.

```
$squeue
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
            123332    debug      bash     user  R      56:01      4 cn[06-09]

$./mpirun --host cn06 hostname
[hpchead:02499] *** Process received signal ***
[hpchead:02499] Signal: Segmentation fault (11)
[hpchead:02499] Signal code: Address not mapped (1)
[hpchead:02499] Failing at address: 0x30
[hpchead:02499] [ 0] /usr/lib64/libpthread.so.0(+0xf370)[0x7ffff5ca3370]
[hpchead:02499] [ 1] /home/user/sandbox/install/ompi_master/lib/libopen-rte.so.0(orte_util_encode_nodemap+0x655)[0x7ffff7b0f883]
[hpchead:02499] [ 2] /home/user/sandbox/install/ompi_master/lib/libopen-rte.so.0(orte_odls_base_default_get_add_procs_data+0x1ae)[0x7ffff7b56ff8]
[hpchead:02499] [ 3] /home/user/sandbox/install/ompi_master/lib/libopen-rte.so.0(orte_plm_base_launch_apps+0x4c3)[0x7ffff7b65fcb]
[hpchead:02499] [ 4] /home/user/sandbox/install/libevent/lib/libevent-2.0.so.5(event_base_loop+0x8bc)[0x7ffff60c399c]
[hpchead:02499] [ 5] ./mpirun[0x401746]
[hpchead:02499] [ 6] ./mpirun[0x401140]
[hpchead:02499] [ 7] /usr/lib64/libc.so.6(__libc_start_main+0xf5)[0x7ffff58f4b35]
[hpchead:02499] [ 8] ./mpirun[0x401059]
[hpchead:02499] *** End of error message ***
Segmentation fault (core dumped)
```
```
#0  orte_util_encode_nodemap (buffer=0x7d7bd0)
    at /home/user/sandbox/src/ompi_master/orte/util/nidmap.c:599
#1  0x00007ffff7b56e74 in orte_odls_base_default_get_add_procs_data (buffer=0x7d7bd0, job=476315649)
[...]

PR #3686, opened 2017-06-09
Print a better error message when srun isn't found in the path. 

rhc54:
Ensure we don't segfault if -host specifies a node not included in the allocation


rhc54:
Fixes #3683

PR #3688, opened 2017-06-09
ROMIO 3.1.4 : Add support for missing ops

jjhursey:
* Add work around support for the following missing ops in ROMIO 3.1.4
    - `MPI_File_iread_at_all`
    - `MPI_File_iwrite_at_all`
    - `MPI_File_iread_all`
    - `MPI_File_iwrite_all`

This mimics the fallback mechanism in OMPIO when it doesn't have internal support.

### DIFF
--- a/ompi/mca/io/romio314/src/io_romio314.h
+++ b/ompi/mca/io/romio314/src/io_romio314.h
@@ -129,12 +129,24 @@ int mca_io_romio314_file_iread_at (struct ompi_file_t *fh,
                                 int count,
                                 struct ompi_datatype_t *datatype,
                                 ompi_request_t **request);
+int mca_io_romio314_file_iread_at_all (struct ompi_file_t *fh,
+                                       MPI_Offset offset,
+                                       void *buf,
+                                       int count,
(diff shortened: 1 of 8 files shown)


edgargabriel:
@jjhursey  looks generally good, but I have a comment/suggestion: for the explicit offset operations, I would suggest to just simply call the corresponding non-collective version of that operation, e.g. for MPI_File_iread_at_all just call MPI_File_iread_at, and let that operation handle internally how they deal with explicit offsets. Basically, the pattern that you copied was how OMPIO deals with them, but I am not convinced that this is the ROMIO way. MPI_File_iread_at will however do the *right thing* on your behalf. (t

jjhursey:
@edgargabriel Thanks for the suggestion. That sounds like a better path. I'll update the PR in a moment with those modifications.

jjhursey:
@edgargabriel I just pushed the change to the second commit. Let me know what you think. Thanks!

jjhursey:
@jjhursey Don't forget to PR this to the release branches 😄

PR #3699, opened 2017-06-14
Only set the "bound" flag if we wre actually bound

rhc54:


rhc54:
Sigh...more Mellanox Jenkins infrastructure failures

bot:mellanox:retest

PR #3704, opened 2017-06-15
Control distribution of signals to children vs grandchildren

rhc54:
By default, apply signals to all direct children _and_ any children they might have spawned (so long as they remain in the same process group). Provide an MCA param (odls_base_signal_direct_children_only) to indicate that the signal is to go _only_ to our direct children, and not be delivered to any children spawned by those procs.

Refs https://www.mail-archive.com/users@lists.open-mpi.org/msg31221.html


ibm-ompi:

jjhursey:
bot:ibm:pgi:retest

rhc54:
I added this to the discussion list for the July meeting as this proposal doesn't exactly do what the user requested, and I'm not sure if we should or not.

Issue #3695, opened 2017-06-12
Segmentation fault when opening a file


wkliao:
I am encountering a Segmentation fault when using a communicator created from MPI_Cart_create in 3D, using OpenMPI version 2.1.0 and Intel C compiler 17.0.0 on a Linux Ubuntu machine.

The gdb trace points to a possible cause at line 914 of file io_ompio_file_open.c
```
    int coords_tmp[2] = { 0 };
```
The size of coords_tmp is too small for 3D coordinate communicators while ompio_fh->f_comm->c_topo->mtc.cart->ndims is 3.

Below is the gdb trace and the test program.
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f2e18f87b4c in mca_io_ompio_cart_based_grouping (ompio_fh=0x0)
    at io_ompio_file_open.c:968
968	           if ((coords_tmp[1]/ompio_fh->f_init_procs_per_group) ==
(gdb) where
#0  0x00007f2e18f87b4c in mca_io_ompio_cart_based_grouping (ompio_fh=0x0)
    at io_ompio_file_open.c:968
#1  0x00007f2e18f85da8 in ompio_io_ompio_file_open (comm=0x2051850, 
    filename=0x20584b0 "testfile", amode=9, 
    info=0x601540 <ompi_mpi_info_null>, ompio_fh=0x20588d0, 
    use_sharedfp=1 '\001') at io_ompio_file_open.c:204
#2  0x00007f2e18f8585b in mca_io_ompio_file_open (comm=0x2051850, 
    filename=0x20584b0 "testfile", amode=9, 
    info=0x601540 <ompi_mpi_info_null>, fh=0x20584d0)
    at io_ompio_file_open.c:62
#3  0x00007f2e26a9fd88 in mca_io_base_file_select (file=0x20584d0, 
    preferred=0x0) at base/io_base_file_select.c:457
#4  0x00007f2e2696a40e in ompi_file_open (comm=0x2051850, 
    filename=0x400f54 "testfile", amode=9, info=0x601540 <ompi_mpi_info_null>, 
    fh=0x7ffcffb0abc0) at file/file.c:132
#5  0x00007f2e26a54ffe in PMPI_File_open (comm=0x2051850, 
    filename=0x400f54 "testfile", amode=9, info=0x601540 <ompi_mpi_info_null>, 
    fh=0x7ffcffb0abc0) at pfile_open.c:92
[...]

PR #3706, opened 2017-06-15
common/ompio: remove function call to cart_based_grouping

edgargabriel:
the cart_based_grouping aggregator strategy was not correctly updated
during the last major rewrite of the aggregator selection algorithm.
It is also not supposed to be called from file_open (but from
file_set_view).

fixes issue  #3695


hppritcha:
I approve this but would you mind updating the commit message to indicate this was reported by a user ?  Thanks

edgargabriel:
@hppritcha  I can update the commit message, question maybe is only for the original master commit or the 3.0.x pr as well?

edgargabriel:
commit message has been updated.

PR #3713, opened 2017-06-16
Enable use of OFI fabrics for launch and other collective operations.…

rhc54:
… Update the PMIx repo to the latest master to get the required support for the server to "push" modex info, and to retrieve all its own "modex" values for sending back to mpirun. Have mpirun cache them in its local modex hash as OFI goes point-to-point direct and doesn't route - so the remote daemons don't need a copy of this connection info.

Remove the opal_ignore from the RML/OFI component, but disable that component unless the user specifically requests it via the "rml_ofi_desired=1" MCA param. This will let us test compile in various environments without interfering with operations while we continue to debug


rhc54:
@anandhis I'll probably keep this on "hold" until we figure out the (unrelated to this PR) race condition in the PMIx support.

@hppritcha You wanted to know when this was coming online - it isn't complete yet, but it does work at least with the socket provider (which is all I have here to test). I'm going to put it on Gadget and try it there.

@jsquyres Would be interesting to see if it works on your boxes

jsquyres:
@rhc54 The usnic libfabric provider only provides UD support -- I don't think it's usable by the OFI RML component.

rhc54:
@jsquyres Is it something very difficult to enable? I gather we'd have to open up the protocol type, but is there a lot of support code required?

jsquyres:
@rhc54 yes, there's quite a bit of code you'd have to add - you have to do all the reliability stuff.

PR #3732, opened 2017-06-22
Add prototype support for HWLOC shmem segment

rhc54:
Will use this only when requested to determine best method

### DIFF
--- a/orte/mca/rtc/hwloc/rtc_hwloc.c
+++ b/orte/mca/rtc/hwloc/rtc_hwloc.c
@@ -12,11 +12,16 @@
 #include "orte/constants.h"
 #include "orte/types.h"
 
+#include <stdio.h>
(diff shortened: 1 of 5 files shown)


rhc54:
@bgoglin Here is the prototype per your example code

rhc54:
@bgoglin rats - just realized that the app is checking the hole after fork, but prior to exec - which is too early. I'll adjust and repost. Also, this doesn't apparently work with FreeBSD - it doesn't recognize MAP_NORESERVE - so we may have an issue with portability.

bgoglin:
I used MAP_NORESERVE in my early tests to verify whether mapping 1TB in a hole worked too. It doesn't matter for our actual implementation because we have MAP_SHARED and the shared memory will already be in physical memory.

PR #3737, opened 2017-06-23
Fix pmix.query support

rhc54:
Fix the pmix_query API when it asks for something that returns an array of pmix_info_t. Protect the PMIX_INFO_FREE macro from NULL arrays. Update the mpi_memprobe scaling test

Thanks to Peter Gottesman for the report


rhc54:
@jsquyres per your IM

jsquyres:
Thanks @rhc54!

@petergottesman Can you verify that this fixes the issue?

PeterGottesman:
Yep, this fixed the issue.
Thanks Ralph!

PR #3738, opened 2017-06-23
Fix the TCP performance impact when BTL not used

bosilca:
Based on an idea from Brian move the libevent trigger update to a later
stage instead of the generic add/del procs. So, we are doing the
increment/decrement when we register the recv handler for an endpoint,
so basically when we create and connect a socket to a peer. The benefit
is that as long as TCP is not used, there should be no impact on the
performance of other BTLs. The drawback is that the first TCP connection
will be slightly slower, but then once we have a peer connected over
TCP things go back to normal.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_endpoint.c
+++ b/opal/mca/btl/tcp/btl_tcp_endpoint.c
@@ -464,6 +464,10 @@ static void *mca_btl_tcp_endpoint_complete_accept(int fd, int flags, void *conte
         mca_btl_tcp_endpoint_event_init(btl_endpoint);
         MCA_BTL_TCP_ENDPOINT_DUMP(10, btl_endpoint, true, "event_add(recv) [endpoint_accept]");
         opal_event_add(&btl_endpoint->endpoint_recv_event, 0);
+        if( mca_btl_tcp_event_base == opal_sync_event_base ) {
+            /* If no progress thread then raise the awarness of the default progress engine */
+            opal_progress_event_users_increment();
+        }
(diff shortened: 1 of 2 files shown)


opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -464,6 +464,10 @@ static void *mca_btl_tcp_endpoint_complete_accept(int fd, int flags, void *conte
         mca_btl_tcp_endpoint_event_init(btl_endpoint);
         MCA_BTL_TCP_ENDPOINT_DUMP(10, btl_endpoint, true, "event_add(recv) [endpoint_accept]");
         opal_event_add(&btl_endpoint->endpoint_recv_event, 0);
+        if( mca_btl_tcp_event_base == opal_sync_event_base ) {
```
bwbarrett: Why have the check?  Unless something changed, increment/decrement should work just fine either way?

bwbarrett:
@gabesaba, can you run the same tests you were running agains master yesterday against George's branch and verify the performance looks good?

gabesaba:
@bwbarrett Sure.  Will do that now

opal/mca/btl/tcp/btl_tcp_endpoint.c
bosilca: This code is common to the libevent and asynchronous (aka. with our own thread) progress. If we have our own thread (in which case we have our own eventset) we don't want to mess with the rest of the OMPI infrastructure and should not alter the frequency of the base eventset.

bwbarrett:
@bosilca, this worked for us; want to push?

PR #3739, opened 2017-06-23
Create file for file backed shared memory in process job session dir.

cniethammer:
Prevent file collisions and allow cleaning of temporary files by orte-clean used for shared file pointer management: Create file for file backed shared memory in process job session dir instead in base /tmp.

### DIFF
--- a/ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
+++ b/ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
@@ -130,12 +130,10 @@ int mca_sharedfp_sm_file_open (struct ompi_communicator_t *comm,
     /* the shared memory segment is identified opening a file
     ** and then mapping it to memory
     ** For sharedfp we also want to put the file backed shared memory into the tmp directory
-    ** TODO: properly name the file so that different jobs can run on the same system w/o
-    **      overwriting each other, e.g.  orte_process_info.proc_session_dir
     */
     /*sprintf(sm_filename,"%s%s",filename,".sm");*/
     filename_basename = basename((void *)filename);
-    sm_filename = (char*) malloc( sizeof(char) * (strlen(filename_basename)+64) );
+    sm_filename = (char*) malloc( sizeof(char) * (strlen(filename_basename)+strlen(ompi_process_info.job_session_dir)+64) );
     if (NULL == sm_filename) {
         free(sm_data);
         free(sh);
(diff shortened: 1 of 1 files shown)


hppritcha:
@edgargabriel something you to review when you're back from vacation

hppritcha:
I'll open PRs against the release branches once this is merged in to master.

edgargabriel:
I liked originally the idea behind this commit, and I can confirm that it works as expected. My hesitation comes from a more theoretical scenario, and maybe I don't understand the session directories well enough, so somebody please feel free to correct me.

Basically, the scenario that could cause trouble when using this patch is: a) an MPI job connecting with an MPI job, b) using Intercomm_merge to create an intra communicator that contains processes from the two different jobs (i.e. they will have different session directories if I understand correctly).  

I understand that this is a very theoretical scenario, but it would not work with this patch. Could we bcast the session directory of rank zero, such that everybody uses that one, similarly to how we deal with the jobid?

rhc54:
Your basic understanding of the session directories is correct. I think the critical question here is: what is supposed to happen to this file when the two jobs disconnect? Does the file get removed during the disconnect procedure? If so, then it can exist in one job's session directory since neither job can leave before disconnecting (else it's an error and the file gets dumped anyway).

If one job is supposed to inherit the file, then it gets a little trickier. If we know at the beginning which job is to "own" the file, then we put it there. Otherwise, we'd have to figure out how to deal with it.

If multiple jobs connect, then we get the same question: if one job disconnects, does the shared file go away? Or does it stay so long as any of the jobs remains alive?

edgargabriel:
thanks for the explanations @rhc54 . Yes, the file is supposed to be removed whenever the file has been closed and the communicator that contains the processes from the connected jobs is automatically freed during the MPI_File_close.  Maybe we can talk briefly about that at the devel meeting next week.

rhc54:
I added it to the agenda

edgargabriel:
I think its probably a 5 min conversation between you and me, since I really have mostly some administrative questions on how to approach this.

rhc54:
Agreed - just wanted a reminder 😄

edgargabriel:
We had a conversation with @rhc54  at the OMPI meeting, and I think the consensus is to not worry about dynamically connected jobs. Shared memory components are also are not being used for communication in dynamically connected jobs since they run into exactly the same problems that we would run into here, namely where to place the backend shared memory file, how to ensure that processes from two different jobs have read/write permission to the different session directories (which might have been launched by different user ids) etc. 

I will have one minor change request to this pr, and will make a modification to disable the component for dynamically connected communicators in the component query function.

bwbarrett:
bot:retest

I think this got caught in a different bug in master; the failures are related to using MPI headers in non-MPI code.  Let's make sure...

edgargabriel:
bot:mellanox:retest

PR #3745, opened 2017-06-23
configure: add builtin asm check for s390/s390x

hjelmn:
We accepted a change that enabled CMA on s390 and s390x. This change
had the side-effect that we were no longer using the builtin atomics
for these systems. This is a problem since we do not have ASM for
s390 and s390x. This commit restores the atomics.


hjelmn:
@amckinstry This will go into v2.x.

PR #3748, opened 2017-06-23
v2.x: Improve x86-64 timer performance

matcabral:
porting PR #2926 to v2.x


bosilca:
I found an interesting reading on this issue (http://akaros.cs.berkeley.edu/lxr/akaros/kern/arch/x86/rdtsc_test.c). We might want to have a reference on it somewhere.

bosilca:
The changes in opal/runtime/opal_progress.c are good. I have added a comment regarding the lack of documentation on why lfence+rdtsc is a better approach.

jsquyres:
@bosilca Thanks for the sanity check. 👍

hppritcha:
@hjelmn please double check this PR.

PR #3749, opened 2017-06-23
Improve x86-64 timer performance

matcabral:
porting PR #2926 to v2.0.x


jsquyres:
@matcabral I'm sorry to cause a fuss, but I think you modified this branch / PR incorrectly.  There's now 5 commits here (there should probably only be 2).

PR #3755, opened 2017-06-25
Sort directory listings

bmwiedemann:
to do builds in a more reproducible manner
indepenent of (random) filesystem ordering.

See https://reproducible-builds.org/ for why this matters.

Signed-off-by: Bernhard M. Wiedemann <bwiedemann@suse.de>


rhc54:
Looks reasonable to me - thanks for the patch!

PR #3772, opened 2017-06-27
Enable ORTE to continue running when a node fails 

rhc54:
- user takes responsibility for zombies. Minor cleanup to orte-clean

(cherry picked from commit 8a4565874e1cdedc80ae45d42ba1fdd44e67883f)

### DIFF
--- a/orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
+++ b/orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
@@ -443,12 +443,6 @@ static void proc_errors(int fd, short args, void *cbdata)
                              ORTE_NAME_PRINT(ORTE_PROC_MY_NAME), ORTE_NAME_PRINT(proc)));
         /* record the first one to fail */
         if (!ORTE_FLAG_TEST(jdata, ORTE_JOB_FLAG_ABORTED)) {
-            /* output an error message so the user knows what happened */
-            orte_show_help("help-errmgr-base.txt", "node-died", true,
-                           ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
-                           orte_process_info.nodename,
-                           ORTE_NAME_PRINT(proc),
-                           pptr->node->name);
             /* mark the daemon job as failed */
             jdata->state = ORTE_JOB_STATE_COMM_FAILED;
             /* point to the lowest rank to cause the problem */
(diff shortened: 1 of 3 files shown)


orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
```diff
@@ -443,27 +443,32 @@ static void proc_errors(int fd, short args, void *cbdata)
...
-            ORTE_UPDATE_EXIT_STATUS(ORTE_ERR_COMM_FAILURE);
+            if (!orte_enable_recovery) {
+                /* output an error message so the user knows what happened */
+                orte_show_help("help-errmgr-base.txt", "node-died", true,
+                               ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
+                               orte_process_info.nodename,
+                               ORTE_NAME_PRINT(proc),
+                               pptr->node->name);
+                /* update our exit code */
+                ORTE_UPDATE_EXIT_STATUS(pptr->exit_code);
```
bosilca: Any reason why the exit code should be set twice ?

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
```diff
@@ -443,27 +443,32 @@ static void proc_errors(int fd, short args, void *cbdata)
...
+                /* just in case the exit code hadn't been set, do it here - this
+                 * won't override any reported exit code */
+                ORTE_UPDATE_EXIT_STATUS(ORTE_ERR_COMM_FAILURE);
+            }
+        }
+        /* if recovery is enabled, then we are done - otherwise,
+         * abort the system */
+        if (!orte_enable_recovery) {
```
bosilca: This test is exactly the same as above. merge?

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
rhc54: Urrr...it isn't? We set the status on the proc itself, and then we set the status for mpirun separately.

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
rhc54: It actually isn't quite the same. The prior code block only gets executed if we aren't already aborting the job. We want to protect the call to abort the HNP as that code can be executed even if we aren't already aborting the job. I'll take another look in case I'm missing something and we actually cannot go down the separate paths.

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
bosilca: OK, I don't understand. ORTE_UPDATE_EXIT_STATUS is used twice in this block of code (461 and 464). If I look at the macro definition it is protected by ```0 == orte_exit_status```, so the second invocation is never successful as the first one will already define orte_exit_status.

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
bosilca: OK, I missed the second closing bracket, so we are not in the same code-block.

orte/mca/errmgr/default_hnp/errmgr_default_hnp.c
bosilca: Here I assumed (and this might have been an error on my part) that ```pptr->exit_code```must be non-zero as we are in the error case and that process somehow failed.

rhc54: yeah, i realized after i thought for a minute that this would be confusing. Actually, we wrote it this way because there were use-cases where we wound up with a zero exit status on the process (e.g., when slurm killed it and we don't get an exit status back). So we do it the second time just to be absolutely certain we return a non-zero status out of mpirun.

bosilca: ok, I realize after writing the commit that there was a possible path in which having 2 one after the other make sense. But, I agree with you it was a little confusing.

PR #3775, opened 2017-06-27
opal/mca: Fix mca_base_verbose file suffix processing

jjhursey:
* `-mca mca_base_verbose file:foo` should create an output file with
    the suffix `foo`. But since we free the pointer at the end of this
    function then by the time we use it it is pointing to invalid memory.
 * This commit fixes that corruption
 * This commit also fixes the behavior of `file:` with no suffix.
   Makes it the same as `file` without the colon.

### DIFF
--- a/opal/mca/base/mca_base_open.c
+++ b/opal/mca/base/mca_base_open.c
@@ -230,12 +230,12 @@ static void parse_verbose(char *e, opal_output_stream_t *lds)
             have_output = true;
         }
 
-        else if (strcasecmp(ptr, "file") == 0) {
+        else if (strcasecmp(ptr, "file") == 0 || strcasecmp(ptr, "file:") == 0) {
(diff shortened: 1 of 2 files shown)


ibm-ompi:

jjhursey:
bot:ibm:gnu:retest

jjhursey:
@rhc54 I suspect that we need a similar patch on the PMIx output processing in PMIx `master` and `v2.x`. I'll work on that once I get the 👍  here.

It's not a commonly used feature but is useful if the stdout/stderr channels are disconnected by the resource manager (e.g., LSF) to get debugging information from the `orted`. That's how I found this break.

jjhursey:
@jsquyres I know I don't need a review, but I'd like a second set of eyes before I start pushing this around.

Issue #3759, opened 2017-06-26
builds of openmpi are not yet reproducible


bmwiedemann:
### What version of Open MPI are you using?
1.10.7 on openSUSE Tumbleweed (aka Factory)

## Details of the problem
For https://reproducible-builds.org/ I tried to build packages of openmpi twice and found various diffs in the binaries.

One source of such diffs should be removed by
https://github.com/open-mpi/ompi/pull/3756

But there are still invocations of 

    `date`
    `hostname`

in
openmpi-1.10.7/autogen.pl
openmpi-1.10.7/config/opal_functions.m4
openmpi-1.10.7/config/opal_get_version.m4
openmpi-1.10.7/ompi/tools/ompi_info/Makefile.am
openmpi-1.10.7/orte/tools/orte-info/Makefile.am
openmpi-1.10.7/oshmem/tools/oshmem_info/Makefile.am

master also has
ompi/tools/mpisync/Makefile.am
opal/mca/pmix/pmix2x/pmix/config/pmix_functions.m4
opal/mca/pmix/pmix2x/pmix/config/pmix_get_version.sh

Would be nice if those `date` usages could be either dropped or use the SOURCE_DATE_EPOCH environment var to override the current time (see examples in https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#Bash_.2F_POSIX_shell )

jsquyres:
I think we'd be amenable to `$SOURCE_DATE_EPOCH` in the way it is described in the page you sent (i.e., if it's set, use it, otherwise, use the same date mechanisms we've been using).  Could you post another PR?

(the PMIX stuff will be submitted upstream to PMIX -- but they copy a lot of their build system / methodology from us, so they'll likely be amenable, too)

FWIW: the `hostname` usage (and date usage, for that matter) are mostly an issue of *logging* -- e.g., `ompi_info` shows the build date and hostname.  This can be helpful in debugging, for example, if a user builds on one machine but then tries to run on an different machine with a different / incompatible set of userspace libraries, etc.  I understand why a distro, for example, would want to have the same build date for all of its software; using `$SOURCE_DATE_EPOCH` seems like a fine solution.

Is there a `$SOURCE_DATE_EPOCH` analog for `hostname`?  I.e., does a reproducible build require identical builds on different machines?

rhc54:
Yeah, we'll take it upstream to MPIx. If it doesn't already exist, perhaps something like ```$SOURCE_HOSTNAME```?

[...]

PR #3779, opened 2017-06-28
configury: make build reproducible

ggouaillardet:
use SOURCE_DATE_EPOCH environment variable if defined
in order to make build reproducible

Fixes open-mpi/ompi#3759


rhc54:
I believe you also need to do the ```SOURCE_HOSTNAME``` patch as well to fully fix the issue.

ggouaillardet:
i google'd ```SOURCE_HOSTNAME``` but could not find anything relevant.
is this somehow standard in debian ? or is this something we think it would be nice to have ?

btw, in ```pmix```, it seems ```PMIX_CONFIGURE_DATE``` is not used, and ```PMIX_REPO_REV``` is only used to build a tarball.

rhc54:
There is no standard for it, but we really should add it to be consistent.

PMIx doesn't have an equivalent for "ompi_info" yet - probably something we need to add now that we have all these plugins appearing.

jsquyres:
@bmwiedemann FYI.

ggouaillardet:
@jsquyres i made the requested changes.

jsquyres:
Cool, thanks.  I think we need to wait on #3759, and then update this PR to match whatever the decision is there.

autogen.pl
```diff
@@ -1178,7 +1178,7 @@ sub in_tarball {
 dnl This file is automatically created by autogen.pl; it should not
 dnl be edited by hand!!
 dnl
-dnl Generated by $username at " . localtime(time) . "
+dnl Generated by $username at " . localtime($ENV{SOURCE_DATE_EPOCH} || time) . "
```
jsquyres: I don't think that this does what you expect.  My reading of https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#Bash_.2F_POSIX_shell is that it expects `SOURCE_DATE_EPOCH` to be of the form 'YYYY-MM-DD'.  If you pass a string of the form 'YYYY-MM-DD' to `localtime()`, you get... weirdness:

```
$ perl -e 'print localtime("2010-10-10") . "\n"';
Wed Dec 31 16:33:30 1969
$ perl -e 'print localtime("2018-10-10") . "\n"';
Wed Dec 31 16:33:38 1969
```

Did you mean:

```
... . ($ENV{SOURCE_DATE_EPOCH} || localtime(time)) . ...
```

config/getdate.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# Copyright (c) 2017      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
+
+date --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}"
```
jsquyres: It might be good to put a comment here in this file explaining why it exists.  Otherwise, Future People will wonder why there's a 1-line script to run the `date` command.

It might actually be good to put a 1-line comment in most/all places we do `$SOURCE_DATE_EPOCH`, referring to reproducible builds, just so that no on accidentally removes the `$SOURCE_DATE_EPOCH` because they don't understand what it's for.

jsquyres:
@ggouaillardet Ping.  I'm reminded of this PR (which has a few things that need to be fixed) because of #5653.

autogen.pl
bmwiedemann: no, it is a plain int. In the linked example, you can see `${SOURCE_DATE_EPOCH:-$(date +%s)}`

jsquyres:
@ggouaillardet Re-ping -- just merged #5563, so this PR now has some conflicts.

AboorvaDevarajan:

lanl-ompi:

jsquyres:
@bmwiedemann updated / refreshed this PR in #8136.  Closing this PR in favor of the newer / freshed #8136.

PR #3787, opened 2017-06-28
Update to cross-version for PMIx

rhc54:
[skip ci]
bot:notest

Just for others to debug with.

@artpol84 @karasevb There is a problem in the dstore support. When the fence stores its data and then the client tries to read it, the keys are garbage. It all works if I restrict pmix to using the "hash" GDS component. Can you please take a look?


rhc54:
I think one possible problem may be that we release the pmix_kval_t after we call "store". The "hash" component does a PMIX_RETAIN on it, and so this ensures proper reference counting. However, if the "dstore" component is just pointing at the kval's data, then it will be free'd and you'll point to garbage.

If that's the root cause, you can avoid it by setting the kval's pointers to NULL after you copy them to the shmem location. This will avoid having the pmix_kval_t destructor free the memory.

rhc54:
Note: you have to run this on two nodes to see the problem.

rhc54:
Problems were fixed by latest PMIx master commits

PR #3790, opened 2017-06-29
orte/iof: Address the case when output is a regular file

artpol84:
Regular files are always write-ready, so non-blocking I/O does not
give any benefits for them.
More than that - if libevent is using "epoll" to track fd events,
epoll_ctl will refuse attempt to add an fd pointing to a regular
file descriptor with EPERM.
This fix checks the object referenced by fd and avoids event_add
using event_active instead.

In the original configuration that uncovered this issue "epoll"
was used in libevent, it was triggering the following warning
message:
"[warn] Epoll ADD(1) on fd 0 failed.  Old events were 0; read
change was 1 (add); write change was 0 (none): Operation not
permitted"
And the side effect was accumulation of all output in mpirun
memory and actually writing it only at mpirun exit.

P.S. For the reference:
https://bugzilla.kernel.org/show_bug.cgi?id=15272

### DIFF
--- a/orte/mca/iof/base/base.h
+++ b/orte/mca/iof/base/base.h
@@ -14,6 +14,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2017      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2017      Mellanox Technologies. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -48,6 +49,7 @@
 #include "opal/class/opal_bitmap.h"
 #include "orte/mca/mca.h"
 #include "opal/mca/event/event.h"
+#include "opal/util/fd.h"
 
 #include "orte/mca/iof/iof.h"
 #include "orte/runtime/orte_globals.h"
@@ -84,6 +86,7 @@ ORTE_DECLSPEC OBJ_CLASS_DECLARATION(orte_iof_job_t);
 typedef struct {
     opal_list_item_t super;
     bool pending;
+    bool always_writable;
     opal_event_t *ev;
     int fd;
     opal_list_t outputs;
@@ -157,6 +160,9 @@ typedef struct orte_iof_base_t orte_iof_base_t;
         ep->tag = (tg);                                             \
         if (0 <= (fid)) {                                           \
             ep->wev->fd = (fid);                                    \
+            ep->wev->always_writable = opal_fd_is_regular(fid) ||   \
+                                       opal_fd_is_chardev(fid) ||   \
+                                       opal_fd_is_blkdev(fid);      \
             opal_event_set(orte_event_base,                         \
                            ep->wev->ev, ep->wev->fd,                \
                            OPAL_EV_WRITE,                           \
--- a/orte/mca/iof/base/base.h
+++ b/orte/mca/iof/base/base.h
@@ -55,6 +55,7 @@
 #include "orte/runtime/orte_globals.h"
 #include "orte/mca/rml/rml_types.h"
 #include "orte/util/threads.h"
+#include "orte/mca/errmgr/errmgr.h"
 
 BEGIN_C_DECLS
 
@@ -88,6 +89,7 @@ typedef struct {
     bool pending;
     bool always_writable;
     opal_event_t *ev;
+    struct timeval tv;
     int fd;
     opal_list_t outputs;
 } orte_iof_write_event_t;
@@ -109,9 +111,11 @@ typedef struct {
     opal_object_t super;
     struct orte_iof_proc_t *proc;
     opal_event_t *ev;
+    struct timeval tv;
     int fd;
     orte_iof_tag_t tag;
     bool active;
+    bool always_readable;
     orte_iof_sink_t *sink;
 } orte_iof_read_event_t;
 ORTE_DECLSPEC OBJ_CLASS_DECLARATION(orte_iof_read_event_t);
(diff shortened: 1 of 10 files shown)


orte/mca/iof/base/iof_base_output.c
```diff
@@ -320,6 +331,7 @@ void orte_iof_base_write_handler(int fd, short event, void *cbdata)
             return;
         }
         num_written = write(wev->fd, output->data, output->numbytes);
+        cum_written += num_written;
```
rhc54: Do you really want to do this addition if num_written is negative?

rhc54: I guess it doesn't really matter, does it? If num_written is negative, you're just going to return anyway, and so cum_written doesn't matter.

Other than this one point, which is up to you since it should work okay (just looks a little odd), it's fine with me to commit.

artpol84:
@rhc54 thank you, that is a good point.
I shifted `cum_written` to the more obvious location in the code.
Will wait until CI is finished and commit.

artpol84:
Can we consider this as a bug and port it to all of our branches?
My colleagues caught that when using hpcx based on ompi v2.x. So ideally I'd like to see this in v2.x as well as in v3.x/v3.0.x

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

rhc54:
Thanks! Appreciate your addressing this problem.

jjhursey:
bot:ibm:retest

hppritcha:
Seems like a bug fix to me so should be cherry picked to other branches.  May be too late for 3.0.0 though

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

ibm-ompi:

jjhursey:
bot:ibm:retest

ibm-ompi:

 Gist: https://gist.github.com/88da22bb73e7d54db1d04aed4eef11fd

ibm-ompi:

 Gist: https://gist.github.com/cab11189daad0a887167c4f550a58700

rhc54:
Hmmm...hold on before pushing this to release branches. I see we are now setting timers to cause reads to fire on compute nodes, and that's a little concerning. We might want to step back and look at this again as it has become far more complicated than expected.

artpol84:
activate is a bad approach according to libevent doc:

```C
// Bad Example: making an infinite loop with event_active()
struct event *ev;
static void cb(int sock, short which, void *arg) {
        /* Whoops: Calling event_active on the same event unconditionally
           from within its callback means that no other events might not get
           run! */
        event_active(ev, EV_WRITE, 0);
}

int main(int argc, char **argv) {
        struct event_base *base = event_base_new();
        ev = event_new(base, -1, EV_PERSIST | EV_READ, cb, NULL);
        event_add(ev, NULL);
        event_active(ev, EV_WRITE, 0);
        event_base_loop(base, 0);
        return 0;
}
```

On compute node it won't use timers because orted creates tty device which is ok to do non-blocking,

orte/mca/iof/base/base.h
```diff
@@ -142,61 +149,120 @@ struct orte_iof_base_t {
 };
 typedef struct orte_iof_base_t orte_iof_base_t;
 
+/* Write event macro's */
+
+static inline bool
+orte_iof_base_fd_always_ready(int fd)
+{
+    return opal_fd_is_regular(fd) ||
+           (opal_fd_is_chardev(fd) && !isatty(fd)) ||
```
artpol84: here `isatty` protects application procs stdio from being timer'ed.

artpol84:
Also it is obvious that solution is far from the original, so I won't merge it until we converge.

rhc54:
Yeah, why don't we hold off and discuss this during the upcoming meeting. I can understand the original fix, minus the use of event_active, but I'm not sure I understand the need for all the changes in the read side of the subsystem. It isn't clear how any of that can relate to the original problem, and so maybe we need to separate things out so we address the original problem and then look at whatever changes are being proposed for other reasons.

artpol84:
The reason I started dealing with read is that with libevent/epoll I'm getting warnings about fd 0 as well as for fd 1.
So effectively input from the regular file won't work with epoll. Here is the output of the following mpi program:
```C
#include <stdio.h>
#include "mpi.h"

int main(int argc, char* argv[])
{
    int rank, size, len;

    MPI_Init(&argc, &argv);

char str[1024] = "";
        scanf("%s", str);

        printf("%d: %s\n", rank, str);

    MPI_Finalize();

    return 0;
}
```

```shell
$ cat inpit
hjkhjkhjkkjnkjnjknjk
$ export OMPI_MCA_event_libevent2022_event_include="poll"
$ mpirun ... < input
$ cat slurm-99.out
0: hjkhjkhjkkjnkjnjknjk
1:
```

```
export OMPI_MCA_event_libevent2022_event_include="epoll"
mpirun ... < input
$ cat slurm-100.out
[warn] Epoll ADD(1) on fd 0 failed.  Old events were 0; read change was 1 (add); write change was 0 (none): Operation not permitted
1:
```
And rank 0 is hanging in `scanf()`

With this PR the output is correct.

artpol84:
And according to this https://bugzilla.kernel.org/show_bug.cgi?id=15272 the fact that poll works well doesn't mean that our implementation is correct. It only means that we are *lucky* not hitting this.

rhc54:
@artpol84 I've been digging thru this a bit, and I don't think there is anything bothersome here. I think the protection (as you pointed out) covers my concerns.

Thanks for the explanation!

artpol84:
@rhc54 @hppritcha @jsquyres @bwbarrett 
I listed all branches as targets. Where should I actually PR this?

PR #3796, opened 2017-06-29
Stop all progress threads prior to releasing the peer objects to avoi…

rhc54:
…d a race condition whereby a lost connection could be reported after a peer object was freed and before the threads were stopped.

(cherry picked from commit 85f8eb4c6bcc2b4995901832f566baf3677d023f)


hppritcha:
I did some more checking and it seems that the IBM tests that were doing spawn or connect/accept triggered the heap corruption problem in mpirun.  This PR fixed that problem - at least when using the OB1 and CM PMLs.

PR #3801, opened 2017-07-03
Detect that we have a mix of BE/LE in the system

### LINKED ISSUE #2817: abort job launch in case of mixed BE/LE environment
Open MPI doesn't currently support mixed big endian/little endian jobs.  In conjunction with a fix for issue #2802, a runtime check to detect the situation where some processes in the job are BE while others are  LE would be helpful.  Ideally a show help message would be emitted which explains that mixed BE/LE jobs are not currently supported.

rhc54:
Detect that we have a mix of BE/LE in the system, provide a warning that OMPI doesn't currently support this environment, and error out

Fixes #2817

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -588,7 +588,7 @@ AC_CACHE_SAVE
 opal_show_title "Header file tests"
 
 AC_CHECK_HEADERS([alloca.h aio.h arpa/inet.h dirent.h \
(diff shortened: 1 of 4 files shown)


rhc54:
@hppritcha Does this meet what you were asking for?

PR #3802, opened 2017-07-04
configury: fix gcc builtin atomic detection

ggouaillardet:
test for both 32 and 64 bits.
clang only support 32 bits builtin atomics when -m32 is used

Thanks Paul Hargrove for reporting this.


PHHargrove:
Appears to resolve the problem for at least clang on x86 and "gcc -m32" on x86-64.

PR #3806, opened 2017-07-04
v3.x: mca/sensor: replace OMPI macro with ORTE macro

ggouaillardet:
and keep autogen.pl --no-ompi happy

This is a one-off commit for the v3.x branch

### DIFF
--- a/orte/mca/sensor/Makefile.am
+++ b/orte/mca/sensor/Makefile.am
@@ -2,6 +2,8 @@
 # Copyright (c) 2009-2010 Cisco Systems, Inc.  All rights reserved.
 #
 # Copyright (c) 2017      Intel, Inc.  All rights reserved.
+# Copyright (c) 2017      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 # 
 # Additional copyrights may follow
@@ -21,7 +23,7 @@ libmca_sensor_la_SOURCES += $(headers)
 
 # Conditionally install the header files
 if WANT_INSTALL_HEADERS
-ortedir = $(ompiincludedir)/$(subdir)
+ortedir = $(orteincludedir)/$(subdir)
 nobase_orte_HEADERS = $(headers)
 endif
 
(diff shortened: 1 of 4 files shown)


ggouaillardet:
@rhc54 @bwbarrett i found this issue with ```autogen.pl --no-ompi```
incidentally, i found the ```sensor``` framework has been restored only for the ```v3.x``` branch, and __not__ in the ```master``` branch. is this what was intended ?

rhc54:
It should have been removed as part of the PR I submitted and Brian committed. I wonder if git somehow failed to remove it? Check the tarball and see if it is there. If so, then the entire framework should be removed.

ggouaillardet:
the framework was removed back in April 2014, but you restored it in March 2017
https://github.com/open-mpi/ompi/commits/v3.x/orte/mca/sensor
anyway, if it should be removed, then please close this PR and i will make a new one

rhc54:
I just looked again, and there is no sensor framework in v3.0.x. I believe the problem here is that you are looking at the _stale_ branch (v3.x). We switched to v3.0.x a couple of months ago.

ggouaillardet:
yes, i did look in the ```v3.x``` branch. a PR was merged 20 hours ago, so from my point of view, it is not really a stale branch. btw, if it is a stale branch, why not simply remove it ?
@bwbarrett any thoughts ?

rhc54:
I'll bet that Brian didn't realize the PR was against the old branch and merged it. FWIW: I believe it was removed, but if you merge a PR it gets recreated. We'll have to remove it again and that PR has to be resubmitted.

jsquyres:
Yes, we really should remove the `v3.x` branch to avoid confusion like this.

Oy... looking at the logs in v3.x, it looks like there have been several merges over in the past month.  😦 

The first nightly build email I see for the v3.0.x branch was on May 30.  It looks like I updated Cisco's MTT to point to the v3.0.x download area on May 31.

bwbarrett:
@ggouaillardet, v3.x is dead.  Someone made a mistake in submitting PRs to it, Github for reasons no one understands allowed me to merge into 3.x (despite permissions that should disallow any merging from anyone), etc.  We'll kill off the branch this week.

PR #3813, opened 2017-07-05
Replace syntax with something less strictly C99

rhc54:

### DIFF
--- a/opal/mca/btl/ugni/btl_ugni_add_procs.c
+++ b/opal/mca/btl/ugni/btl_ugni_add_procs.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2011-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2011      UT-Battelle, LLC. All rights reserved.
- * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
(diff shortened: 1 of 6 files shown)


opal/mca/btl/ugni/btl_ugni_add_procs.c
```diff
@@ -35,6 +35,7 @@ int mca_btl_ugni_add_procs (struct mca_btl_base_module_t* btl, size_t nprocs,
     mca_btl_ugni_module_t *ugni_module = (mca_btl_ugni_module_t *) btl;
     int rc;
     void *mmap_start_addr;
+    struct timeval tv = {0, MCA_BTL_UGNI_CONNECT_USEC};
```
PHHargrove: @rhc54 I would advise retaining the named initializer (here and in the other `struct timeval` case) to yield
```
struct timeval tv = {.tv_sec = 0, .tv_usec = MCA_BTL_UGNI_CONNECT_USEC};
```
I say this because POSIX defines the struct with the intentionally vague phrase "shall include at least the following members" used throughout the standard.  By my interpretation this does not guarantee their order or even that they are the first two members.  
I freely acknowledge that I am being paranoid.

rhc54: Good point - will do!

PR #3822, opened 2017-07-06
Fix MPI_SIZEOF for gfortran 4.8

tjcw:
MPI_SIZEOF gets disabled by the configure stage of OMPI when used with gfortran 4.8.5 . This is because the following syntax
    size = storage_size(x)
isn't recognised by the compiler for x which is an array.
Converting this to
    size = storage_size(x(1))
passes the configuration stage, and making a corresponding change to gen-mpi-sizeof.pl causes the generated code to be acceptable to gfortran (and hopefully also to the other fortran compilers which accept the base syntax).

### DIFF
--- a/README
+++ b/README
@@ -19,6 +19,7 @@ Copyright (c) 2013-2015 Intel, Inc. All rights reserved
 Copyright (c) 2015      NVIDIA Corporation.  All rights reserved.
 Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
                         reserved.
+Copyright (c) 2017      IBM Corporation. All rights reserved                        
 
 $COPYRIGHT$
 
@@ -365,8 +366,7 @@ Compiler Notes
   - All Fortran compilers support the mpif.h/shmem.fh-based bindings,
     with one exception: the MPI_SIZEOF interfaces will only be present
     when Open MPI is built with a Fortran compiler that support the
-    INTERFACE keyword and ISO_FORTRAN_ENV.  Most notably, this
-    excludes the GNU Fortran compiler suite before version 4.9.
+    INTERFACE keyword and ISO_FORTRAN_ENV.
 
   - The level of support provided by the mpi module is based on your
     Fortran compiler.
--- a/config/ompi_fortran_check_storage_size.m4
+++ b/config/ompi_fortran_check_storage_size.m4
@@ -11,6 +11,7 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2010-2014 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2017      IBM Corporation. All rights reserved. 
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
@@ -61,7 +62,7 @@ SUBROUTINE storage_size_complex32_r1(x, size)
     COMPLEX(REAL32), DIMENSION(*)::x
     INTEGER, INTENT(OUT) :: size
 
-    size = storage_size(x) / 8
+    size = storage_size(x(1)) / 8
 END SUBROUTINE storage_size_complex32_r1
 
 SUBROUTINE storage_size_int32_scalar(x, size)
@@ -77,7 +78,7 @@ SUBROUTINE storage_size_int32_r1(x, size)
     INTEGER(INT32), DIMENSION(*)::x
     INTEGER, INTENT(OUT) :: size
 
-    size = storage_size(x) / 8
+    size = storage_size(x(1)) / 8
 END SUBROUTINE storage_size_int32_r1
 
 SUBROUTINE storage_size_real32_scalar(x, size)
@@ -93,7 +94,7 @@ SUBROUTINE storage_size_real32_r1(x, size)
     REAL(REAL32), DIMENSION(*)::x
     INTEGER, INTENT(OUT) :: size
 
-    size = storage_size(x) / 8
+    size = storage_size(x(1)) / 8
 END SUBROUTINE storage_size_real32_r1
 ]])],
              [AS_VAR_SET(fortran_storage_size_var, yes)],
--- a/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
@@ -3,6 +3,7 @@
 # Copyright (c) 2014-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2015      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2017      IBM Corporation. All rights reserved.
 # $COPYRIGHT$
 #
 # Script to generate the overloaded MPI_SIZEOF interfaces and
@@ -97,7 +98,7 @@ sub queue_sub {
 ${indent}  INTEGER, INTENT(OUT) :: size
 ${indent}  INTEGER$optional_ierror_param, INTENT(OUT) :: ierror";
     $subr->{start} = $start;
-    $subr->{middle} = "${indent}  size = storage_size(x) / 8
+    $subr->{middle} = "${indent}  size = storage_size(xSUBSCRIPT) / 8
 ${indent}  ${optional_ierror_statement}ierror = 0";
     $subr->{end} = "${indent}END SUBROUTINE ^PREFIX^$sub_name^RANK^";
 
@@ -126,6 +127,7 @@ sub generate {
     if (0 == $rank) {
         $str =~ s/\^RANK\^/_scalar/g;
         $str =~ s/\^DIMENSION\^//;
+        $str =~ s/SUBSCRIPT//;
     } else {
         $str =~ s/\^RANK\^/_r$rank/g;
         my $dim;
@@ -135,6 +137,7 @@ sub generate {
             --$d;
         }
         $str =~ s/\^DIMENSION\^/, DIMENSION($dim*)/;
+        $str =~ s/SUBSCRIPT/($dim 1)/;
     }
 
     # All done


jjhursey:
bot:retest

jjhursey:
@jsquyres I wanted another set of Fortran eyes on this one. Chris was tracking down why the `MPI_SIZEOF` configure check was failing for `gfortran` and this warning message tipped him off:

```shell
configure:58913: checking if Fortran compiler supports STORAGE_SIZE for relevant types
configure:58991: gfortran -o conftest -m64 -O3 -L/home/../lib  conftest.f90  >&5
conftest.f90:32.24:

    size = storage_size(x) / 8
                        1
Error: The upper bound in the last dimension must appear in the reference to the assumed size array 'x' at (1)
conftest.f90:48.24:

    size = storage_size(x) / 8
                        1
Error: The upper bound in the last dimension must appear in the reference to the assumed size array 'x' at (1)
conftest.f90:64.24:

    size = storage_size(x) / 8
                        1
Error: The upper bound in the last dimension must appear in the reference to the assumed size array 'x' at (1)
configure:58991: $? = 1
```

I know there are 2 things to do to this PR before merging - but those are non-functional:
 * squash to 1 with a signoff
 * Add IBM copyright.

jjhursey:
add to whitelist

jsquyres:
In spelunking through history to figure out if this is an actual bug or if there was a reason why Past Jeff used `sizeof(x)` instead of `sizeof(x(1))`, I was reminded that there is a bunch of text in README (on master, v2.0.x, v2.x, and v3.0.x) about MPI_SIZEOF, the `mpi` module, and the `mpi_f08` module -- all in relation to what the compiler supports.  It specifically mentions that all of  this stuff in Open MPI is only active for GFortran >= 4.9.

I haven't checked for sure, but this leads me to believe that `sizeof(x)` failing to compile when `x` is an array was a bug in GFortran that was fixed in 4.9 (alternatively: it could be an extension that GFortran added in 4.9, and all other compilers already supported :man_shrugging:).

Is GFortran 4.8 important to IBM?

And if so, which Fortran interfaces are important with GFortran 4.8 -- just MPI_SIZEOF in the `mpif.h` interface, the entire `mpi` module, and/or the entire `mpi_f08` module?

If we end up supporting GFortran 4.8 for some/all of these as a result of this PR, this PR should be amended to update the README text, too.

@jjhursey already noted, but these commits need to be squashed before merging, too.  And the signed-off-by line needs to be fixed (sorry, for legal reasons, we require a very specific form of Signed-off-by -- you can get it by adding `-s` to your `git commit` command.  E.g. `git commit --amend -s ...`).

tjcw:
Thanks for the comments. I have added a couple of IBM copyrights and fixed the signed-off-by line.
Why gfortran 4.8 ? IBM is building IBM Spectrum MPI https://www-03.ibm.com/systems/spectrum-computing/products/mpi/ on RHEL 7, and that has gfortran 4.8.5 installed.

tjcw:
I've deleted the part in the README about MPI_SIZEOF not being supported by the GNU Fortran 4.8 compiler; I left in the part about the compiler needing to support INTERFACE and ISO_FORTRAN_ENV because this is still true. 
This fix doesn't do anything to change the other Fortran interface support, so I think the rest of the README is still OK.

jsquyres:
Ok, thanks.  I have only one more small request, and I filed a PR against your branch for it.  I'd like to keep *some* mention of GFortran in there, simply because it's so prevalent.  Putting stuff in the README tends to prevent questions from being sent to the mailing list.

If you merge the PR I just filed against your branch, my commit should show up here on this PR, and we should be good to go to merge to OMPI master.

Thanks!

jsquyres:
Thanks @tjcw.  When CI finishes, we'll merge.

jsquyres:
@tjcw @gpaulsen @jjhursey Do you guys need this PR'ed to release branches?  If so, please feel free to do so.  😄

jjhursey:
Yeah we'll want this in the release branches. I'll setup those PRs today.

jjhursey:
MTT Caught a build failure because of this patch while trying to build the 'ibm' test suite:
 * https://mtt.open-mpi.org/index.php?do_redir=2458

```
mpi_sizeof_usempi.o: In function `MAIN__':
/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:24:
undefined reference to `mpi_sizeof_'
/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:25:
undefined reference to `mpi_sizeof_'
/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:26:
undefined reference to
`mpi_sizeof_'/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:27:
undefined reference to `mpi_sizeof_'
/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:28:
undefined reference to `mpi_sizeof_'
mpi_sizeof_usempi.o:/gpfs_dev/mpiczar/mtt/scratch/ompi-master_gcc/installs/TQnA/tests/ibm/ibm/random/mpi_sizeof_usempi.f90:29:
more undefined references to `mpi_sizeof_' follow
```

I put a hold on the release branch PRs until we can sort that out.

jjhursey:
Compare these two builds.

XL compiler:
```
shell$ xlf -qversion
IBM XL Fortran for Linux, V15.1.5 (5725-C75, 5765-J10)
Version: 15.01.0005.0000
shell$ cd ompi-tests/ibm/random
shell$ nm mpi_sizeof_usempi.o  | grep sizeof
                 U mpi_sizeof_complex32_r1
                 U mpi_sizeof_complex32_r2
                 U mpi_sizeof_complex32_scalar
                 U mpi_sizeof_int16_r1
                 U mpi_sizeof_int16_r2
                 U mpi_sizeof_int16_scalar
                 U mpi_sizeof_int32_r1
                 U mpi_sizeof_int32_r2
                 U mpi_sizeof_int32_scalar
                 U mpi_sizeof_int64_r1
                 U mpi_sizeof_int64_r2
                 U mpi_sizeof_int64_scalar
                 U mpi_sizeof_int8_r1
                 U mpi_sizeof_int8_r2
                 U mpi_sizeof_int8_scalar
                 U mpi_sizeof_real32_r1
                 U mpi_sizeof_real32_r2
                 U mpi_sizeof_real32_scalar
                 U mpi_sizeof_real64_r1
                 U mpi_sizeof_real64_r2
                 U mpi_sizeof_real64_scalar
0000000000000000 T mpi_sizeof_test
```

GNU Compiler:
```
shell$ gfortran --version | head -n 1
GNU Fortran (GCC) 4.8.5 20150623 (Red Hat 4.8.5-11)
shell$ cd ompi-tests/ibm/random
shell$ nm mpi_sizeof_usempi.o | grep sizeof
                 U mpi_sizeof_
```

I'm reviewing this commit:
  * https://svn.open-mpi.org/trac/ompi/changeset/32764
  * https://github.com/open-mpi/ompi/commit/d7eaca83fac0d9783d40cac17e71c2b090437a8c
  * This is where it was determined that we need GNU Fortran >= 4.9

I think that this PR fixed one part of the configure logic that was supposed to keep gfortran 4.8 out. However, we will need a stronger configure check for missing compiler support that is causing this test to fail.

I'd like to pin this down a bit better, but reverting this PR might be the right thing to do... 😢

jsquyres:
Oy.  The title of d7eaca8 is quite appropriate: "Fortran: Fix MPI_SIZEOF. What a disaster. :-("

Yay Present Josh for tracking down this old commit; yay Past Jeff for writing a giant, descriptive commit message explaining the issues!

It sounds like:

* IBM may be out of luck for SIZEOF with gfortran 4.8.  😦 
* I think @tjcw's fix isn't necessarily *wrong* (i.e., testing for `sizeof(x(1))` is ok) -- perhaps the original `sizeof(x)` test was fortuitously keeping gfortran 4.8 out, even if it wasn't fully correct.
* I see the following in the original commit:
```
AS_IF([test $OMPI_FORTRAN_HAVE_INTERFACE -eq 1 && \
           test $OMPI_FORTRAN_HAVE_ISO_FORTRAN_ENV -eq 1],
         [OMPI_FORTRAN_BUILD_SIZEOF=1],
         [OMPI_FORTRAN_BUILD_SIZEOF=0])
AC_SUBST(OMPI_FORTRAN_BUILD_SIZEOF)
```
* It looks like that test has been expanded on master to check for even more things.  A good first place to look, then, would be the results of the individual tests that it is checking for SIZEOF support, and why/if all of them are passing.  Do those tests need to be stronger?
* Unless you think you can fix it today, perhaps it is best to revert this commit.  😢  If it's not worth it to fix this properly, it might be ok to revert and just put a comment there explaining what has happened (i.e., we "fixed" it to check `sizeof(x(1))`, but then that unintentionally allowed gfortran 4.8 in, ...yadda yadda yadda.

tjcw:
Let's revert this PR. It looks like something is more broken with gfortran 4.8 than just refusing to accept the sizeof(x) syntax for x being an array. Looks like IBM will have to install gfortran 4.9 or newer on build machines if IBM wants its MPI to allow gfortran users to have access to MPI_SIZEOF.

jsquyres:
I'm imagining this is the room where you were all discussing this Fortran issue inside IBM:

image

😄

jjhursey:
Yeah I think the configure check needs to be stronger. I'm just not exactly sure what that needs to look like. Reverting this and updating the `README` with a comment about this thread is what we should do today (I can do that), then followup with a PR that makes the configure check stronger (if we can devise such a test).

@jsquyres What is the term for what Fortran is doing converting:
```fortran
    integer*8 :: i8, i8_1(2), i8_2(2,2)
    call mpi_sizeof(i8, isize, ierror)
    call mpi_sizeof(i8_1, isize, ierror)
    call mpi_sizeof(i8_2, isize, ierror)
```

to function calls named:
```
                 mpi_sizeof_int64_r1
                 mpi_sizeof_int64_r2
                 mpi_sizeof_int64_scalar
```
I could try to craft a test for just this scenario, but I'd like to see if I can find the gfortran change long item where this feature was introduced.

jjhursey:
Per phone discussion this is the `INTERFACE` functionality in Fortran. We need to make the check stronger in https://github.com/open-mpi/ompi/blob/master/config/ompi_fortran_check_interface.m4 to build an module and then a program that uses the module to check the functionality works. That second part (a program that uses the module) will fail in gfortran 4.8.

The plan for today:
 * (Josh) Revert this commit. Add a comment to the code (not `README`) about this discussion.

The plan for the next week or so:
 * (Josh) Revive the configure fix. Make the `INTERFACE` check stronger to exclude gfortran 4.8. That way gfortran 4.8 will fail the correct configure test.
 * (Josh) Investigate if a gfortran 4.9 is ABI compatible with gfortran 4.8. This might not work anyway because even if we compile MPI with 4.9 the user would be compiling their application with 4.8 and that's where the problem will show up. We'll poke at it a bit though.

PR #2940, opened 2017-02-08
misc fixes for heterogeneous cluster support

ggouaillardet:
Refs. open-mpi/ompi#2838

PR #3828, opened 2017-07-06
Detect that we have a mix of BE/LE in the system, provide a warning that OMPI doesn't currently support this environment, and error out

rhc54:
(cherry picked from commit 2753f53e6d75f6aab863eaa319cea9f9d434d5ef)

### DIFF
--- a/config/opal_configure_options.m4
+++ b/config/opal_configure_options.m4
@@ -286,22 +286,7 @@ fi
 AC_DEFINE_UNQUOTED(OPAL_ENABLE_DLOPEN_SUPPORT, $OPAL_ENABLE_DLOPEN_SUPPORT,
     [Whether we want to enable dlopen support])
 
-#
-# Heterogeneous support
-#
-
-AC_MSG_CHECKING([if want heterogeneous support])
-AC_ARG_ENABLE([heterogeneous],
-    [AC_HELP_STRING([--enable-heterogeneous],
-                    [Enable features required for heterogeneous
-                     platform support (default: disabled)])])
-if test "$enable_heterogeneous" = "yes" ; then
-     AC_MSG_RESULT([yes])
-     opal_want_heterogeneous=1
-else
-     AC_MSG_RESULT([no])
-     opal_want_heterogeneous=0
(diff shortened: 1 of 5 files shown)


ggouaillardet:
@rhc54 you need to cherry-pick open-mpi/ompi@823382f5d7b7477d7c9471e4d4e25db8c77ff8bb

we do support mixing BE/LE, but only when configure'd with ```--enable-heterogeneous```

rhc54:
@ggouaillardet Are you sure that is true?? I looked but that PR you had created a long time ago was never committed, and so far as I can tell, we still don't work correctly in that mode.

ggouaillardet:
i will resume working on that PR.
as far as i am concerned, mixed BE/LE works pretty well in my environment (which might have a few not yet commited patches).
i am fine if ```v3.0.x``` does not support mixed BE/LE, but i will appreciate if ```master``` does.

rhc54:
I'll leave this PR as-is until we see your PR completed. I'm a little disturbed about changing master without clear evidence that hetero is working again, but let's leave that in for now with the proviso that you get that PR completed and committed soon.

hppritcha:
hmm... well I now see something I'd like addressed if we're disabling BE/LE mixed support.  Namely, we should at least for a while, remove the --enable-heterogeneous configure option if we know it doesn't work.

rhc54:
Okay - easy enough to add. Will leave master alone, though, pending the PR from @ggouaillardet

ggouaillardet:
the PR is #2940

rhc54:
@hppritcha I pushed the requested change

bwbarrett:
I'm not sure I like this patch...  Let's push to master and then to v3.0.x.  Yes, Giles is working on fixing heterogenous in master, but he can add a revert with the rest of his patch series.  We need to get in the habit of not special casing all the branches.

rhc54:
no problem - will do

rhc54:
Okay, I replaced the custom commit with the one to master: 8e25733760bccdad434450831f248069191104a5

PR #3833, opened 2017-07-07
pml/yalla: fix getting size of a continuous type.

yosefe:
See #3566

### DIFF
--- a/ompi/mca/pml/yalla/pml_yalla_datatype.h
+++ b/ompi/mca/pml/yalla/pml_yalla_datatype.h
@@ -27,10 +27,11 @@ OBJ_CLASS_DECLARATION(mca_pml_yalla_convertor_t);
 #define PML_YALLA_INIT_MXM_REQ_DATA(_req_base, _buf, _count, _dtype, _stream_type, ...) \
     { \
         size_t size; \
-        ptrdiff_t lb; \
+        ptrdiff_t lb, extent; \
         \
         if (opal_datatype_is_contiguous_memory_layout(&(_dtype)->super, _count)) { \
-            ompi_datatype_get_true_extent(_dtype, &lb, &size); \
+            ompi_datatype_get_true_extent(_dtype, &lb, &extent); \
+            ompi_datatype_type_size(_dtype, &size); \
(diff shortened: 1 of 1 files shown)


markalle:
I left a comment under the issue-report, but I think we should switch to ```opal_datatype_span()``` since it's the most general-purpose function made to identify the byte range of a (dt,count) pair.

markalle:
my proposal would be:
```
        ptrdiff_t len, offset; \
        \
        if (opal_datatype_is_contiguous_memory_layout(&(_dtype)->super, _count)) { \
            len = opal_datatype_span((_dtype), (_count), &offset); \
            (_req_base)->data_type          = MXM_REQ_DATA_BUFFER; \
            (_req_base)->data.buffer.ptr    = (char *)_buf + offset; \
            (_req_base)->data.buffer.length = len; \
```

yosefe:
@markalle fixed
for 2.x same fix would not work, since on v2.x opal_datatype_span() does not always return 0 when dtype->size==0

markalle:
I'm not familiar with the v2.x issue. I'm inclined to call that an opal_datatype_span() bug if it doesn't return a 0 span when dtype->size is 0.  Does it break because of the true extent being identified as 1?  I feel like that itself is a bug, but if not then opal_datatype_span() ought to work around it.

markalle:
Well, looking at the code I guess it wouldn't matter what the true extent is, since "extent" can be anything. So the current line
```    return true_extent + (count - 1) * extent;```
in opal_datatype_span() seems like it could break for both of the degenerate cases size==0 and count==0.

yosefe:
yes. so for "master" we can merge this patch IMHO. for v2.x need to either fix opal_datatype_span(), or use the original patch i've submitted. I prefer the 2nd option.

bosilca:
Why not cherry-picking bf0fc4a84c3 to the v2.x ?

markalle:
I like the cherry-pick idea. Seems like a good fix that keeps opal_datatype_span() always doing the right thing even in the degenerate cases

yosefe:
@bosilca @markalle sounds good.
In the meantime, can we merge this one to master?

bosilca:
👍

PR #3835, opened 2017-07-07
Remove --enable-heterogeneous until fix is ready

rhc54:


rhc54:
Refs #2940 for @ggouaillardet

PR #3836, opened 2017-07-07
Bring the ofi/rml component online

rhc54:
Still only used if rml_ofi_desired=1


rhc54:
@anandhis

anandhis:
Does this fix the orte_no_op over ofi ?

rhc54:
Yes - I've been launching MPI jobs with it just fine. One of the builds in Jenkins is failing, though, so I'm trying to track that down

rhc54:
hmmm...I can't make it fail. I wonder if this is something peculiar to suse?

@bwbarrett Is there any way to get more debug info out of that build? I've tried building with the same config, forcing various components off, etc. - to no avail. However, I don't have a suse build.

opal/mca/pmix/pmix2x/pmix/src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2895,16 +2910,51 @@ static pmix_status_t dstore_store_modex(struct pmix_nspace_t *nspace,
...
         if (PMIX_SUCCESS != rc) {
+>>>>>>> Update PMIx and continue work on RML/OFI component:opal/mca/pmix/pmix2x/pmix/src/mca/gds/ds12/gds_dstore.c
             PMIX_ERROR_LOG(rc);
             bo->bytes = pbkt.base_ptr;
             bo->size = pbkt.bytes_used; // restore the incoming data
             pbkt.base_ptr = NULL;
             PMIX_DESTRUCT(&pbkt);
             return rc;
         }
+<<<<<<< c632784ca34c467055eadcb4efe84a25e5a3911b:opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c
```
artpol84: @rhc54

opal/mca/pmix/pmix2x/pmix/src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2804,6 +2808,16 @@ static pmix_status_t dstore_assign_module(pmix_info_t *info, size_t ninfo,
         }
     }
 
+<<<<<<< c632784ca34c467055eadcb4efe84a25e5a3911b:opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c
```
artpol84: @rhc54 - looks like unresolved merge

rhc54: nah, looks like i just missed removing the last line of the conflict flag - i can clean that up

artpol84: I saw multiple of them and there was clearly an unresolved merges.

rhc54: not sure that last part is true, but no matter - it runs just fine and passed an MTT scan, so I think things resolved correctly. Anyway, I'm updating from tarball now.

artpol84: https://github.com/rhc54/ompi/blob/f7e8780a42eb512ed5c54715471168814ef63383/opal/mca/pmix/pmix2x/pmix/src/mca/gds/ds12/gds_dstore.c#L2811:L2829
I'm 100% that this will not compile!

rhc54: relax dude - do you not see the .pmix_ignore in that component? It is prevented from compiling due to the reported problems in the dstore. I am removing the ignore now as it seems functional, albeit the memory footprint problem remains

artpol84: What memory footprint problem? Is there any issue on that?
I think that  @karasevb have fixed (at least) one of the possible causes he discovered when keys was sent in the message from the server to the client even if the dstore was used.

rhc54: Look at the issue he filed this morning in PMIx

artpol84: Ok, thanks. I thought it's something else.

PR #3837, opened 2017-07-10
master: mtl-portals4: add timeout to rendezvous get fragments

tkordenbrock:
This PR adds a timeout to each fragment of a rendezvous get.  If any fragment times out or fails, the entire receive fails.

Signed-off-by: Todd Kordenbrock <thkgcode@gmail.com>

### DIFF
--- a/ompi/mca/mtl/portals4/mtl_portals4.h
+++ b/ompi/mca/mtl/portals4/mtl_portals4.h
@@ -73,6 +73,7 @@ struct mca_mtl_portals4_module_t {
 
     /* free list of rendezvous get fragments */
     opal_free_list_t fl_rndv_get_frag;
+    int get_retransmit_timeout;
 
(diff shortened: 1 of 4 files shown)


hppritcha:
@regrant is this an enhancement?   Adding a new MCA parameter to releases that are now in bug fix only mode (2.0.x and 2.x) makes accepting this PR into those release problematic.

regrant:
@hppritcha I agree this should go into 3.X not the 2.X series. It's an enhancement to a previous bug fix to make it more efficient and therefore an enhancement.

bwbarrett:
This looks like a PR for master.  You should probably remove the target:3.0 tag and file a new PR for the merge to 3.0 once this is in master (anyone can approve the pull into master, so that's all you...).

regrant:
@bwbarrett thanks I missed the master tag, will do.

PR #3838, opened 2017-07-10
opal/info: fix recursive deadlock in opal_info_dup_mode()

ggouaillardet:
use opal_info_{get,set}_nolock() instead of opal_info_{get,set}()
since the former can be invoked when the info lock is being held.


ibm-ompi:

ggouaillardet:
fwiw, this is solving issues such as https://mtt.open-mpi.org/index.php?do_redir=2457 when ```--mca
btl_tcp_progress_thread 1``` is used

PR #3844, opened 2017-07-11
pml/ob1: do not access fragment after calling btl rget

### LINKED ISSUE #3821: vader fails IMB-EXT Unidir_Get test
Hi, I am testing OMPI with Intel IMB test “**IMB-EXT Unidir_Get**” as following. With “**vader**”, it gave error. Without “vader” or adding openib as “-mca btl **openib,vader,tcp,self**”, the test passed well.   Does “vader” need special configure option to build? Does it need other options to make it work? 

Thanks,

Dahai

```shell
mpirun -n 2 \
 -mca btl **vader**,tcp,self \
 imb_src/IMB-EXT Unidir_Get
```

```shell       
#bytes #repetitions      t[usec]   Mbytes/sec
            0         1000         0.04         0.00
            4         1000         0.72         5.31
            8         1000         0.89         8.54
           16         1000         0.88        17.39
           32         1000         1.10        27.76
           64         1000         1.15        53.29
          128         1000         0.91       134.17
          256         1000         0.94       260.82
 Read -1, expected 18446744073709547536, errno = 22
 Read -1, expected 18446744073709547536, errno = 22
 Read -1, expected 18446744073709550864, errno = 22
          512         1000         0.95       516.29
         1024         1000         0.56      1731.87
         2048         1000         0.68      2892.14
```

hjelmn:
This commit fixes a bug that occurs when the btl callback happens before
the rget returns. In this case the fragment has been returned and is no
longer valid. This commit saves the size before calling rget. This is
valid since the BTL is not allowed to change the read size.

Fixes #3821

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2008      UT-Battelle, LLC. All rights reserved.
  * Copyright (c) 2011      Sandia National Laboratories. All rights reserved.
  * Copyright (c) 2012-2015 NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2011-2016 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2011-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


hjelmn:
Fixes the silent data corruption.

Issue #3845, opened 2017-07-11
Issue when using btl/vader under heavy contention/


hjelmn:
Thank you for taking the time to submit an issue!

## Background information

There is a longstanding bug in the ob1 pml that has bitten us yet again. Under heavy contention it is possible to get into a state where ob1 will not progress its send pending list even though there are network resources available. This happens when the BTL makes no callbacks (because there are none to be made) when there are fragments on the ob1 pending list. The quick fix is to update vader to avoid the issue but the real fix will be to update ob1 to periodically progress the send pending list 

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

All versions.

hjelmn:
Working on the vader workaround now.

jjhursey:
@hjelmn We are just trying to better understand the fix. in PR #3845 the discussion is about fixing a progression issue. How did that lead to a data corruption?
Or is the data corruption handled by #3844 and a hang fixed by PR #3845?

hjelmn:
The data corruption is #3844 and the hang is #3845. Eventually I will make ob1 better with regard to  the pending list but modifying vader works for now.

bwbarrett:
@hjelmn, where are we on this?

hjelmn:
Complete. All branches merged.

PR #3846, opened 2017-07-11
btl/vader: work around ob1 pending fragment bug

hjelmn:
This commit ensures that the pml callback is always made when
sending fragments. This is needed to avoid #3845. Once that is
fixed the #if 0'd code can be restored.


jjhursey:
How does PR #3844 relate to this one?

I'll setup a build and give it a try.

hjelmn:
@jjhursey Both commits are required to fix #3844.

jjhursey:
Tested with both PR #3844 and PR #3846 on `master`. The test passed on our system:
`mpirun -np 8 -mca pml ob1 -mca btl vader,self ./IMB-EXT Unidir_get`

PR #3848, opened 2017-07-12
configury: fix PBS Pro support

ggouaillardet:
recent versions of PBS Pro requires libcrypto.so, but libpbs.so
does not (yet) depend on it, so manually add -lcrypto if -lpbs alone fails.

Thanks Petr Hanousek for bringing this to our attention

Refs PBSPro/pbspro#331

### DIFF
--- a/config/orte_check_tm.m4
+++ b/config/orte_check_tm.m4
@@ -11,7 +11,7 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2015      Research Organization for Information Science
+dnl Copyright (c) 2015-2017 Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@jsquyres any thougts on that PR ?
PBS Pro will likely no more need ```-lcrypto``` in a near future.
this PR tries ```-lpbs``` and then ```-lpbs -lcrypto``` if the first one failed, so Open MPI can support PBS Pro out of the box

ggouaillardet:
fwiw, my PR was merged into PBS Pro, so Open MPI will work out of the box with PBS Pro in a near future.
meanwhile, we can merge this PR, or simply document a workaround

jsquyres:
Do you want to change the order -- i.e., check *without* crypto first, and then *with* it second?  That would probably make us more friendly to upstream improvements coming soon (i.e., just try to link against PBS and if it works, great -- but if not, then try again with crypto).

ggouaillardet:
@jsquyres i double checked the PR again, and we do try in this order
1. ```-lpbs```
2. ```-lpbs -lcrypto```
3. ```-ltorque```

so as far as i am concerned, this PR does what we expect

PR #3851, opened 2017-07-12
Fix v2.0.x ob1 issues

hjelmn:


jjhursey:
@hjelmn The original issue could not be reproduced on this branch. The test passes clean without any additional changes from this PR. Is this PR necessary then?

hppritcha:
@jsquyres not sure we should take this per @jjhursey s comment.

bosilca:
The bug exists even if we are unable to replicate it with a particular test.

PR #3860, opened 2017-07-12
README: say we don't support PGI on OS-X

### LINKED ISSUE #3859: README: state that PGI is not supported on OS-X
Add note that we don't support the PGI compiler on OS-X.

hppritcha:
[skip ci]
fixes #3859

### DIFF
--- a/README
+++ b/README
@@ -177,6 +177,11 @@ Compiler Notes
   source directory path names that was resolved in 9.0-4 (9.0-3 is
   known to be broken in this regard).
 
+- Open MPI does not support the PGI compiler suite on OS X or MacOS.
+  See issues below for more details:
+  https://github.com/open-mpi/ompi/issues/2604
+  https://github.com/open-mpi/ompi/issues/2605
+
 - OpenSHMEM Fortran bindings do not support the `no underscore` Fortran
   symbol convention. IBM's xlf compilers build in that mode by default.
   As such, IBM's xlf compilers cannot build/link the OpenSHMEM Fortran


README
```diff
@@ -185,6 +187,9 @@ Compiler Notes
   for more details:
   https://github.com/open-mpi/ompi/issues/3612
 
+- IBM's xlf compilers: NO known good version that can build/link
+  the MPI f08 bindings or build/link the OpenSHMEM Fortran bindings.
+
```
jjhursey: @hppritcha This line is not correct. We regularly test xlf in this way. It is correct the XL and OpenSHMEM have a problem. See note on line 180/182 about that issue. I think this bullet should be removed.

hppritcha: I think I messed this up with a conflict rebase.  removing.

hppritcha:
@jjhursey check now

README
```diff
@@ -177,6 +177,8 @@ Compiler Notes
   source directory path names that was resolved in 9.0-4 (9.0-3 is
   known to be broken in this regard).
 
+- The PGI compiler is not supported on OS-X.
```
jsquyres: Nit picks:

* The PGI compiler itself may be supported on OS X -- we don't want to say that it isn't.  We want to say that Open MPI doesn't support building with the PGI compiler on OS X.
* It's not "OS-X".  It's "OS X" and "MacOS".

How about:

> Open MPI does not support the PGI compiler suite on OS X or MacOS.

hppritcha:
@jsquyres check now

hppritcha:
I'll cite the issue numbers like we do for other problems further down in the README.

hppritcha:
Rewrote per Jeffs request.  Can't wait for him to come back from vacation and give okay.

Issue #3075, opened 2017-03-01
PGI F08 symbol Issue on POWER


jjhursey:
MTT found this issue with Open MPI using the PGI compiler (16.10 on ppc64le) and the F08 module.

As you can see below the `use mpi` module is fine, but the `use mpi_f08` module is not.
```
shell$  mpirun -np 2 ./ring_usempi
Process 0 sending 10 to  1 tag 201 ( 2 processes in ring)
Process 0 sent to  1
Process 0 decremented value:  9
Process 0 decremented value:  8
Process 0 decremented value:  7
Process 0 decremented value:  6
Process 0 decremented value:  5
Process 0 decremented value:  4
Process 0 decremented value:  3
Process 0 decremented value:  2
Process 0 decremented value:  1
Process 0 decremented value:  0
Process  0 exiting
Process  1 exiting
```

```
shell$ mpirun -np 2 ./ring_usempif08
Process 0 sending 10 to  1 tag 201 ( 2 processes in ring)
[mpi10:137972] *** An error occurred in MPI_Recv
[mpi10:137972] *** reported by process [4003397633,1]
[mpi10:137972] *** on communicator MPI_COMM_WORLD
[mpi10:137972] *** MPI_ERR_TYPE: invalid datatype
[mpi10:137972] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[mpi10:137972] ***    and potentially your MPI job)
[mpi10:137959] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[mpi10:137959] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

Attaching a debugger the datatype being passed into the `MPI_Recv` (and `MPI_Send` - which also triggers an error, but `MPI_Recv` just won the race to issue an error here) is `0` instead `7`.

The `ring_usempif08` program has a weak symbol for the datatype:
```
shell$ nm ring_usempif08 | grep _integer
00000000100200b8 V ompi_f08_mpi_integer
```
Which if not resolved probably defaults to `0` causing the issue.

The issue is in `master` and the release branches. The Fortran marshaling code is _complex_ so I'm filing a bug to get some more visibility on the problem.

jsquyres:
@jjhursey and I chatted about this issue this afternoon; he got a bit of an education on Fortran.  😄 

We narrowed the issue down a bit (the `_integer` symbol is a red herring).  Short version:

[...]

Issue #2606, opened 2016-12-19
PGI Fortran: runtime error when using using mpi_f08 module on OpenPOWER


hppritcha:
@PHHargrove has reported a runtime problem when using the PGI 16.10 fortran compiler on OpenPOWER.  

This posting on devel describes the problem in some more detail:

https://mail-archive.com/devel@lists.open-mpi.org/msg19836.html

PHHargrove:
@jjhursey 
Believe it or not, I see the same failure with the XLC community edition compilers on the same system! 
No issues with gcc/gfortran.

xiachsh:
@PHHargrove 

which version of IBM Compiler did you used ?

PHHargrove:
@xiachsh Here is the full version info for xlc and xlf:
```
phargrov@openpower-6:~$ xlc -qversion
IBM XL C/C++ for Linux, V13.1.4 (Community Edition)
Version: 13.01.0004.0001
/opt/ibm/xlC/13.1.4/bin/.orig/xlc: note: XL C/C++ Community Edition is a no-charge product and does not include official IBM support. You can provide feedback at the XL on POWER C/C++ Community Edition forum (http://ibm.biz/xlcpp-linux-ce). For information about a fully supported XL C/C++ compiler with advanced optimization features, visit XL C/C++ for Linux (http://ibm.biz/xlcpp-linux).
phargrov@openpower-6:~$ xlf -qversion
IBM XL Fortran for Linux, V15.1.5 (Community Edition)
Version: 15.01.0005.0001
/opt/ibm/xlf/15.1.5/bin/.orig/xlf: 1501-303 (I) XL Fortran Community Edition is a no-charge product and does not include official IBM support. You can provide feedback at the XL on POWER Fortran Community Edition forum (http://ibm.biz/xlfortran-linux-ce). For information about a fully supported XL Fortran compiler, visit XL Fortran for Linux (http://ibm.biz/xlfortran-linux).
```

DanielCChen:
@PHHargrove. 
Are you still experiencing problems building openmpi with IBM's community edition compilers?

I tried it using the same IBM XLC and XLF compilers as you mentioned to build Openmpi 2.0.2, and it compiled and ran ring_openmpif08.f90 just fine for me.

[...]

PR #3866, opened 2017-07-12
README: Note about ld issue for XL and PGI on PPC

jjhursey:
* Related to Issue #2606 and Issue #3075
 * The core problem in those two issues is related to a regression in
   ld upstream. Add a note in the README about this issue.

[skip ci] bot:notest


jjhursey:
I'll PR this to all of the release branches since it impacts them all

PR #3867, opened 2017-07-12
test Mlnx jenkins

artpol84:


artpol84:
@jsquyres @hjelmn Jenkins should be fixed,
PRs will need to be retriggered once I'll finish with checking.

PR #3870, opened 2017-07-12
pmix/s2: fix srun native launch for pmi2

hppritcha:
recent changes that broke native launch on cray
using srun or aprun also broke native launch
using pmi2.


rhc54:
Ah...sorry for the problem. I wonder how we made it thru CI? Is it because srun defaults to pmi-1, and the s1 component was updated?

PR #3871, opened 2017-07-13
v3.0.x: misc fixes for heterogeneous clusters

ggouaillardet:


bwbarrett:
Why is the attribute handling patch required for heterogeneous clusters?  That's an awful big change to bring into the v3.0.x series this late in the game.

ggouaillardet:
@bwbarret strictly speaking, the attribute handling patch is not required for heterogeneous clusters.

that being said, from #2940 

> the changes involves the prototype of a function pointer.
> so there are several ways to see this.
> 
>   1.  this breaks ABI
>    2. this does not break ABI (the pointer size did not change)
>    3. current implementation is broken anyway, so this really a bug fix, and since the pointer size did not change, this is an acceptable commit
> 
> i'd rather go with the latter option.

so if you are willing to accept the attribute handling patch for ```v3.0.1``` (e.g. no or acceptable ABI break), then i am fine postponing it. but if this is now or in ```v4```, then i'd rather do it now.

bwbarrett:
If it's not a heterogeneous fix, it shouldn't be in the list of heterogenous fixes.  We can discuss whether the attribute patch is appropriate in a PR that just contains the attribute patch (since it sounds like it stands on its own).

ggouaillardet:
@bwbarrett i removed the attribute handling patch from this PR, and moved it in #3891

Issue #3854, opened 2017-07-12
malloc(0) in MPI_Unpublish_name() with MPI_INFO_NULL


dalcinl:
This is with v2.1.1

```python
>>> from mpi4py import MPI
>>> port = MPI.Open_port()
>>> MPI.Publish_name("service", port)  # info defaults to MPI.INFO_NULL
>>> MPI.Unpublish_name("service", port)  # info defaults to MPI.INFO_NULL
malloc debug: Request for 0 bytes (pmix1_client.c, 849)
>>> MPI.Close_port(port)
>>> 
```

The offending line (in master) is here: https://github.com/open-mpi/ompi/blob/2f11d371cdb2a1fa7e19ca738d3c753eb05bb912/opal/mca/pmix/ext1x/pmix1x_client.c#L862

PR #3873, opened 2017-07-13
pmix: do not invoke PMIX_INFO_CREATE() with a zero size

ggouaillardet:
Thanks Lisandro Dalcin for the report

Fixes open-mpi/ompi#3854


jsquyres:
@ggouaillardet This PR now has conflicts; can you rebase?

rhc54:
I fixed the conflict.

FWIW: the ext2x component is just a carbon copy of the pmix2x component with a few name changes. So if these are required for the ext2x component, why aren't they replicated in pmix2x?

ggouaillardet:
@rhc54 i do not really know what happened.

(almost) everything was ok after open-mpi/ompi@9124afbeae77e88c79a7eb3597a627d7d85395c4 landed into master, but then some changes were (accidentally i guess) reverted in open-mpi/ompi@a1b15c5666593bcd3bf4844bf9fd53a640e442a1.

anyway, made the changes again and squashed the commits.
we should hence merge this PR after it completes CI

ibm-ompi:

 Gist: https://gist.github.com/883a8c0ecb67d290e88e04b78eba2dba

ibm-ompi:

 Gist: https://gist.github.com/80c1619dec1429abfa9d8d7395250471

ibm-ompi:

 Gist: https://gist.github.com/d883e0c0c2e866223956c6a8d63962be

rhc54:
Answer is simple: as I said, the ext2x component is supposed to be a carbon copy of the pmix2x one, with some minor renaming. If you are making changes to ext2x without making them in pmix2x, then they are going to be lost every time we do an update.

If they need to be different, then you need to explain why.

rhc54:
Otherwise, I would strongly suggest you make the changes in pmix2x, and then we copy them over to ext2x so the two are identical again.

ggouaillardet:
in this case, should we instead make something easier to maintain ?

i have two options on top of my head
- one option is to use the same source file for both components, and use macro to generate component based symbols
- an other option would be to generate the ```ext2x``` sources from ```pmix2x``` sources and apply some ```sed``` to them

rhc54:
In this kind of situation, we usually create base files that contain the common functionality between the two components. I'll take a poke at that today

rhc54:
Crud - that won't work here as the component integration functions must include their respective PMIx headers, and we cannot reference component headers in the base as we don't know where they are when the base is being built.

I can't think of anything beyond your two suggestions. I think the second option is likely the easier one to execute as it pretty much is a /pmix2x/ext2x replacement. HOWEVER, there is one major difference that must be preserved. There is a function in pmix2x that deals with OPAL_PREFIX and friends that _must not_ be in ext2x. Still, that would be an easy difference to maintain (could even be in a separate file that we exclude from the sed).

You want to take a crack at it?

ggouaillardet:
@rhc54 i pushed an extra commit that does the job.

that being said, is this the right way of doing things ?
the rationale is that ```pmix2x``` and ```ext2x``` are mutually exclusive (e.g. both __cannot__ be built) so there is no need to worry about two components defining the same symbols.
we could for example have only one component (and in order to ease debugging, the name of the component/library could be ```pmix2x``` or ```ext2x``` depending on how Open MPI was configured)

rhc54:
I actually had committed the one component approach before, but it created an issue (I forget exactly what was wrong - might have been something about static builds). I agree there is no need to rename symbols except for the component name and the version string in the component.c files. We also have to avoid the OPAL_PREFIX thing, but (as I said) we can just move that function to a separate file and not include it in the generator.

One way to do this would be to just do a symlink between the relevant files in the two directories. I'm not sure how folks would feel about that one, but I can raise it at today's telecon.

Alternatively, we could figure out a way to set CPPFLAGS and friends so that we can include the right paths in a Makefile.am in the base. Then we could move those files over there, which is more in keeping with the usual OMPI method. My earlier comment wasn't correct - by the time the base is actually being built, we have completed configure and we _do_ know where the PMIx files are located. We just have to set some defines and capture them as AM_CPPFLAGS.

I'll see if anyone has opinions on today's telecon.

rhc54:
Let's go with this solution - can't see another path forward.

PR #3875, opened 2017-07-13
coll/spacc: Add Alltoallv implementation

mkurnosov:
1. Add pairwise exchange algorithm for Alltoallv. This algorithm takes into account
   zero values in `scounts[]` and `rcouns[]`.

2. Fix: prepare the placeholder for the array of `request*` (for using linear allreduce).

### DIFF
--- a/ompi/mca/coll/spacc/coll_spacc_alltoallv.c
+++ b/ompi/mca/coll/spacc/coll_spacc_alltoallv.c
@@ -1,0 +1,124 @@
+/*
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+#include "coll_spacc.h"
+
+#include "mpi.h"
+#include "ompi/constants.h"
+#include "opal/util/bit_ops.h"
+#include "ompi/datatype/ompi_datatype.h"
+#include "ompi/communicator/communicator.h"
+#include "ompi/mca/coll/coll.h"
+#include "ompi/mca/coll/base/coll_base_functions.h"
+#include "ompi/mca/coll/base/coll_tags.h"
+#include "ompi/mca/coll/base/coll_base_util.h"
+#include "ompi/mca/pml/pml.h"
+
+/*
+ * mca_coll_spacc_alltoallv_intra_block
+ *
+ * Function:  Alltoallv using pairwise exchange algorithm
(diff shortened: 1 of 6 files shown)


bwbarrett:
ok to test

bosilca:
What is the difference between this and ompi_coll_base_alltoallv_intra_pairwise ?

ompi/mca/coll/spacc/coll_spacc_alltoallv.c
```diff
@@ -0,0 +1,124 @@
...
+    for (i = 0; i < comm_size; i += block_size) {
+        nreqs = 0;
+        nops = comm_size - i < block_size ? comm_size - i : block_size;
+        /* Post nops irecvs and isends */
+        for (j = 0; j < nops; j++) {
+            remote = (rank + i + j) % comm_size;
+            if (rcounts[remote] <= 0 || rank == remote)
+                continue;
+            precv = ((char *)rbuf) + (ptrdiff_t)rdisps[remote] * rdtype_ext;
+            err = MCA_PML_CALL(irecv_init(precv, rcounts[remote], rdtype,
```
ggouaillardet: Is there any reason why you prefer `i{send,recv}_init` over `i{send,recv}` ?

If you really want to use `i{send,recv}_init`, then these requests must be properly free'd right after `ompi_request_wait_all()`

ggouaillardet:
@bosilca @mkurnosov 

I think the commit message is misleading since the implemented algorithm is not really a pairwise exchange. (it is more of a multi-pairs wise exchange so to speak).

Also, no message is exchanged when the count is zero.
Note we used to have this implementation (we also handled zero size datatypes) and then we removed it.
Though it might look like a good idea at first glance, it leads to MPI tasks being unsynchronized, and then a lot of unexpected messages that ends up degrading performances. That is why we chose to revert the zero size message optimization.

mkurnosov:
@ggouaillardet @bosilca thank you for your review and comments.

> Note we used to have this implementation (we also handled zero size datatypes) and then we removed it. Though it might look like a good idea at first glance, it leads to MPI tasks being unsynchronized, and then a lot of unexpected messages that ends up degrading performances. That is why we chose to revert the zero size message optimization.

I am agree with your arguments. I think the right solution is to close this PR. Is there list of open tasks (RoadMap) for the ompi/coll/{base,basic,tuned} subsystem?

bosilca:
No at this point there is no common roadmap for any of the collective frameworks.

PR #3886, opened 2017-07-13
README: Note about ld issue for XL and PGI on PPC

jjhursey:
* Related to Issue #2606 and Issue #3075
 * The core problem in those two issues is related to a regression in
   ld upstream. Add a note in the README about this issue.

[skip ci] bot:notest

### DIFF
--- a/README
+++ b/README
@@ -188,6 +188,16 @@ Compiler Notes
   for more details:
   https://github.com/open-mpi/ompi/issues/3612
 
+- Compiling Fortran programs using the mpi_f08 module on PowerPC with
+  the PGI (tested 17.5) or XL (tested v15.1.5) Fortran compilers and GNU
+  linker after 2.25.1 and before 2.28 will likely experience runtime failures.
+  This was noticed on Ubuntu 16.04 which uses the 2.26.1 version of ld by
+  default. However, this issue impacts any OS running the impacted
+  version of ld. This GNU linker regression will be fixed in version 2.28.
+  Below is a link to the GNU bug on this issue:
+    https://sourceware.org/bugzilla/show_bug.cgi?id=21306
+  The XL compiler will have a fix for this issue in their next release.
+
 - On NetBSD-6 (at least AMD64 and i386), and possibly on OpenBSD,
   libtool misidentifies properties of f95/g95, leading to obscure
   compile-time failures if used to build Open MPI.  You can work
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -188,6 +188,16 @@ Compiler Notes
   for more details:
   https://github.com/open-mpi/ompi/issues/3612
 
+- Compiling Fortran programs using the mpi_f08 module on PowerPC with
+  the PGI (tested 17.5) or XL (tested v15.1.5) Fortran compilers and GNU
+  linker after 2.25.1 and before 2.28 will likely experience runtime failures.
```
jsquyres: Per IM discussion with @jjhursey, I think that this above sentence is trying to say that compiling MPI applications that utilize the `mpi_f08` module with the ... compilers on the ... platforms with the GNU linker ... will likely experience runtime failures when running the MPI applications.  Even that language is a little awkward -- the key points here are:

1. compiling Fortran MPI applications that utilize `mpi_f08` (vs. compiling Open MPI itself)
1. problems occur when you run the MPI application (vs. when compiling or linking Open MPI and/or the application)

README
```diff
@@ -188,6 +188,16 @@ Compiler Notes
...
+  This was noticed on Ubuntu 16.04 which uses the 2.26.1 version of ld by
+  default. However, this issue impacts any OS running the impacted
+  version of ld. This GNU linker regression will be fixed in version 2.28.
+  Below is a link to the GNU bug on this issue:
+    https://sourceware.org/bugzilla/show_bug.cgi?id=21306
+  The XL compiler will have a fix for this issue in their next release.
```
jsquyres: * s/fix/workaround/ because you just said the bug is in the GNU ld, so XL isn't *fixing* the bug -- it's *working around* the bug
* s/their/its/ because you're not anthropomorphizing the XL compiler 😄 

Can you cite a specific version number of the next XL release?  "Next" is an relative (i.e., amorphous) word to use in a README file that will continue to live on the web for years.

jjhursey: How about instead of "in their next release" I just say "in a future release"?

jsquyres: 👍  That works -- that's a little more intentionally saying "some future version, but I'm specifically not telling you which version" (Vs. "next", which might force a reader to look at the date on this README/release and try to reverse-map it back to XL versions).

jsquyres:
@jjhursey Any progress, perchance?

jjhursey:
@jsquyres I pushed a new commit to clarify the language. Let me know what you think.

I figure we can iterate on the additional commit until we get the language set to how we like it. Then we can discuss how to update the language in all of the necessary release branches (this was already merged into `master` and `v3.0.x` so we will need new PRs for those).

PR #3887, opened 2017-07-13
Orte/v3.0.x/iof sbatch

artpol84:


artpol84:
@hppritcha For some reasons when I tried to cherry-pick this to the v2.x and I had a bunch of conflicts. I recall that you had only Copyright-related conflicts, but I saw that because the following thread-safety things are not in v2.x the set of conflicts is much more substantial:
master/v3.x: https://github.com/open-mpi/ompi/blob/master/opal/threads/threads.h#L122:L128
v2.x: https://github.com/open-mpi/ompi/blob/v2.x/opal/threads/threads.h

rhc54:
Yeah, I'm afraid you'll need a custom patch for v2.x

hppritcha:
@rhc54 please review when you have a chance.

PR #3888, opened 2017-07-13
configure: remove CR/FT related options

hppritcha:
As part of the process for addressing removal of CR/FT related
code from master (and hence from the 3.0.0 release), it was agreed
at the OMPI devel F2F on 7/13/17 that we'd break this in to two
pieces:

1) remove the configure arguments (fewer changes)
2) remove all the CR/FT code, etc. in a subsequent bigger commit
    that may not make it in to 3.0.0 in time.

By doing 1), the available configure options would not change
in a subsequent 3.0.x release if we end up not being able to do 2)
before 3.0.0 is released.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1077,7 +1077,7 @@ AC_CACHE_SAVE
 # visible again
 #
 ###########################################################
-OPAL_SETUP_FT_OPTIONS
+dnl OPAL_SETUP_FT_OPTIONS
 ###########################################################
 # The following line is always required as it contains the
 # AC_DEFINE and AM_CONDITIONAL calls that set variables used


configure.ac
```diff
@@ -1077,7 +1077,7 @@ AC_CACHE_SAVE
 # visible again
 #
 ###########################################################
-OPAL_SETUP_FT_OPTIONS
+#OPAL_SETUP_FT_OPTIONS
```
bwbarrett: This will not do what you want.  You want dnl (to prevent the m4 macro from expanding) rather than #, which will just make the macros commented out.

But the bigger question is why not just remove the line?

bosilca: We decided the second pass will remove everything (and I suppose the macro was kept to mark it as a starting point for the second effort).

configure.ac
```diff
@@ -1062,28 +1062,6 @@ AC_INCLUDES_DEFAULT
...
-#
-# However, we recognize that there are researchers who use this
-# option on their independent branches. In such cases, simply
-# uncomment the line below to render the FT configure options
-# visible again
-#
-###########################################################
-#OPAL_SETUP_FT_OPTIONS
-###########################################################
-# The following line is always required as it contains the
```
bwbarrett: These comments seem important...

hppritcha:
@bwbarrett check now.

hppritcha:
@bwbarrett please recheck

bosilca:
I am preparing a large PR that removes all CR stuff, including this.

PR #3894, opened 2017-07-14
pml/ob1: fix mca_pml_ob1_progress_needed usage

ggouaillardet:
OPAL_ATOMIC_ADD32() returns the *new* value and *not* the previous one.

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_progress.c
+++ b/ompi/mca/pml/ob1/pml_ob1_progress.c
@@ -10,6 +10,8 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
+ * Copyright (c) 2017      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@bosilca @hjelmn could one of you please review this before i merge and PR vs the release branches ?

ompi/mca/pml/ob1/pml_ob1_progress.c
```diff
@@ -54,8 +56,8 @@ static inline int mca_pml_ob1_process_pending_cuda_async_copies(void)
 static int mca_pml_ob1_progress_needed = 0;
 int mca_pml_ob1_enable_progress(int32_t count)
 {
-    int32_t old = OPAL_ATOMIC_ADD32(&mca_pml_ob1_progress_needed, count);
-    if( 0 != old )
+    int32_t progress_count = OPAL_ATOMIC_ADD32(&mca_pml_ob1_progress_needed, count);
```
ggouaillardet: if mca_pml_ob1_progress_needed is 0 and count is 1, then the current version returns, but the patched one does not
OPAL_ATOMIC_ADD32 returns the __new__ value, and __not__ the old one, so the current version is clearly misleading, and imho, busted too

PR #3900, opened 2017-07-16
pml/yalla: fix getting size of a continuous type.

### LINKED ISSUE #3566: yalla with irregular contig datatypes

yosefe:
picked from #3833
Fixes #3566

### DIFF
--- a/ompi/mca/pml/yalla/pml_yalla_datatype.h
+++ b/ompi/mca/pml/yalla/pml_yalla_datatype.h
@@ -26,14 +26,13 @@ OBJ_CLASS_DECLARATION(mca_pml_yalla_convertor_t);
 
 #define PML_YALLA_INIT_MXM_REQ_DATA(_req_base, _buf, _count, _dtype, _stream_type, ...) \
     { \
-        size_t size; \
-        ptrdiff_t lb; \
+        ptrdiff_t span, gap; \
         \
         if (opal_datatype_is_contiguous_memory_layout(&(_dtype)->super, _count)) { \
-            ompi_datatype_get_true_extent(_dtype, &lb, &size); \
+            span = opal_datatype_span(&(_dtype)->super, (_count), &gap); \
             (_req_base)->data_type          = MXM_REQ_DATA_BUFFER; \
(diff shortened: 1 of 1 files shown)


yosefe:
@alinask @jladd-mlnx @bosilca pls take a look

jladd-mlnx:
@hppritcha @bwbarrett Could someone please merge this? Without this patch our internal MTT is breaking.

bwbarrett:
@jladd-mlnx, @yosefe set the milestone to 3.0.1, so we haven't touched it.  I take it you'd like us to merge before 3.0.0 ships?

yosefe:
@bwbarrett yes

PR #3901, opened 2017-07-17
coll/libnbc: demote progress_lock to regular flag

Zzzoom:
The LOCK COMPXCHG that grabs ```mca_coll_libnbc_component.progress_lock``` (which detects recursion and defines a critical section) has been showing up on my profiles during single-threaded runs, where there shouldn't be any contention. This PR changes the lock to an OPAL_THREAD_LOCKed boolean flag to speed up the single-thread case while keeping the critical section otherwise.

Signed-off-by: Carlos Bederián <bc@famaf.unc.edu.ar>

### DIFF
--- a/ompi/mca/coll/libnbc/coll_libnbc.h
+++ b/ompi/mca/coll/libnbc/coll_libnbc.h
@@ -75,7 +75,6 @@ struct ompi_coll_libnbc_component_t {
     opal_free_list_t requests;
     opal_list_t active_requests;
     int32_t active_comms;
-    opal_atomic_lock_t progress_lock; /* protect from recursive calls */
     opal_mutex_t lock;                /* protect access to the active_requests list */
 };
 typedef struct ompi_coll_libnbc_component_t ompi_coll_libnbc_component_t;
--- a/ompi/mca/coll/libnbc/coll_libnbc_component.c
+++ b/ompi/mca/coll/libnbc/coll_libnbc_component.c
@@ -39,6 +39,7 @@ const char *mca_coll_libnbc_component_version_string =
 
 
 static int libnbc_priority = 10;
+static bool libnbc_in_progress = false;     /* protect from recursive calls */
 bool libnbc_ibcast_skip_dt_decision = true;
 
 
@@ -101,8 +102,6 @@ libnbc_open(void)
     /* note: active comms is the number of communicators who have had
        a non-blocking collective started */
     mca_coll_libnbc_component.active_comms = 0;
-
-    opal_atomic_init(&mca_coll_libnbc_component.progress_lock, OPAL_ATOMIC_UNLOCKED);
 
     return OMPI_SUCCESS;
 }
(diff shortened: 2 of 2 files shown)


Zzzoom:
@ggouaillardet am I missing something here?

ompi/mca/coll/libnbc/coll_libnbc.h
```diff
@@ -75,7 +75,7 @@ struct ompi_coll_libnbc_component_t {
     opal_free_list_t requests;
     opal_list_t active_requests;
     int32_t active_comms;
-    opal_atomic_lock_t progress_lock; /* protect from recursive calls */
+    bool in_progress;                 /* protect from recursive calls */
```
ggouaillardet: imho, ```in_progress``` should rather be a static variable of ```coll_libnbc_component.c``` since it is not used anywhere else

ompi/mca/coll/libnbc/coll_libnbc_component.c
```diff
@@ -263,37 +262,38 @@ ompi_coll_libnbc_progress(void)
...
+                else {
+                    request->super.req_status.MPI_ERROR = res;
+                }
+                ompi_request_complete(&request->super, true);
             }
-            ompi_request_complete(&request->super, true);
+            OPAL_THREAD_LOCK(&mca_coll_libnbc_component.lock);
         }
-        OPAL_THREAD_LOCK(&mca_coll_libnbc_component.lock);
+        mca_coll_libnbc_component.in_progress = false;
```
ggouaillardet: at first glance, this should be in the ```if (! in_progress)``` block.
makes sense ?

Zzzoom: It is, the diff without -w is making it hard to read.

Zzzoom:
> if i understand correctly this PR, you replaced an opal_atomic_lock_t with a boolean that is protected by the mca_coll_libnbc_component.lock, is that right ?
Yes, that's correct.

> imho, the right approach, if feasable of course, would be to get rid of the recursivity.
Sadly I'm not familiar enough with the inner workings of libnbc (or openmpi) to figure that out.

ompi/mca/coll/libnbc/coll_libnbc.h
Zzzoom: Right, I've updated the PR.

Zzzoom:
FWIW if this isn't going to get merged (it slashes around 40% of libnbc time when running single-threaded), the recursive calls are being caused by [this line](https://github.com/open-mpi/ompi/blob/master/ompi/mca/coll/libnbc/nbc.c#L329) which has been in libnbc since the beginning.

Stack trace goes:
```
#0  ompi_coll_libnbc_progress () at coll_libnbc_component.c:295
#1  0x00007ffff7562641 in opal_progress () at runtime/opal_progress.c:226
#2  0x00007ffff7b0daf1 in ompi_request_default_test_all (count=8013936, 
    requests=0x7fffe39d7b60, completed=0x1, statuses=0x2)
    at request/req_test.c:204
#3  0x00007fffe39d87b5 in NBC_Progress (handle=0x7a4870) at nbc.c:329
#4  0x00007fffe39d7bcb in ompi_coll_libnbc_progress ()
    at coll_libnbc_component.c:275
#5  0x00007ffff7562641 in opal_progress () at runtime/opal_progress.c:226
#6  0x00007ffff7b0e499 in ompi_request_wait_completion (req_ptr=0x7a4870, 
    status=0x7fffe39d7b60) at ../ompi/request/request.h:392
#7  ompi_request_default_wait (req_ptr=0x7a4870, status=0x7fffe39d7b60)
    at request/req_wait.c:41
#8  0x00007ffff7b4207e in PMPI_Wait (request=0x7a4870, status=0x7fffe39d7b60)
    at pwait.c:70
```

ggouaillardet:
Could you please squash the two commits into a single one ?
Then i will merge this PR and backport into the release branches

Zzzoom:
There you go, thanks

ggouaillardet:
thanks, i just merged the PR
what is the benchmark you used to evidence the 40% bump in performance ?

Zzzoom:
Quantum Espresso PWscf benchmark runs, libnbc went from taking 2% of walltime to around 1.1%. I should take OMB for a spin.

Issue #3902, opened 2017-07-17
Compilation error: SIGSEGV not declared here. First use in this function


victorsndvg:
### What version of Open MPI are you using?

OpenMPI 3.0.0 rc1 and also OpenMPI 2.1.1 

### Describe how Open MPI was installed

Both versions compiled from tarballs

### Please describe the system on which you are running

* Operating system/version:  RedHat 6 - kernel version 2.6.32-573.12.1.el6.x86_64

## Details of the problem

Error detected during compilation process.

Configuration has been performed with some external libraries:  hwloc, libverbs, pmix and libevent.

**Configuration and compilation commands:**
```
$ ./configure --with-hwloc=... --enable-shared --with-slurm --enable-mpi-thread-multiple --with-verbs-libdir=... --enable-mpirun-prefix-by-default --disable-dlopen --with-pmix=... --with-libevent=... --with-knem
$ make all install
```

**Error message:**
```
...

btl_vader_knem.c: En la función ‘mca_btl_vader_knem_init’:
btl_vader_knem.c:112:5: aviso: declaración implícita de la función ‘signal’ [-Wimplicit-function-declaration]
     signal (SIGSEGV, SIG_DFL);
     ^~~~~~
btl_vader_knem.c:112:13: error: ‘SIGSEGV’ no se declaró aquí (primer uso en esta función)
     signal (SIGSEGV, SIG_DFL);
             ^~~~~~~
btl_vader_knem.c:112:13: nota: cada identificador sin declarar se reporta sólo una vez para cada función en el que aparece
btl_vader_knem.c:112:22: error: ‘SIG_DFL’ no se declaró aquí (primer uso en esta función)
     signal (SIGSEGV, SIG_DFL);
                      ^~~~~~~
make[2]: *** [btl_vader_knem.lo] Error 1
make[2]: se sale del directorio `/mnt/lustre/scratch/home/cesga/vsande/SingularityImages/tests/OMPI_compatibility/openmpi-3.0.0rc1/opal/mca/btl/vader'
make[1]: *** [all-recursive] Error 1
make[1]: se sale del directorio `/mnt/lustre/scratch/home/cesga/vsande/SingularityImages/tests/OMPI_compatibility/openmpi-3.0.0rc1/opal'
make: *** [all-recursive] Error 

```

**Possible fix:**

In both versions, seems that adding `#include <signal.h>` in `btl_vader_knem.c` file avoid the issue in my case.

Thanks in advance.

[...]

PR #3909, opened 2017-07-17
btl/vader: remove debug code that should not be in a release

hjelmn:
References #3902. Close when in master, v3.0.x, and v2.x.


bwbarrett:
@hjelmn, per the meeting last week, please don't add release target labels for a PR into master.  Open an issue with the target labels to track, but the PR itself should not have release targets unless it's against the release branch.

hjelmn:
@bwbarrett Ah yes. Need to get used to the new procedure.

Issue #3903, opened 2017-07-17
switching to a non-default grouping options in ompio leads to segfault


edgargabriel:
Thank you for taking the time to submit an issue!

## Background information

when changing the aggregator selection algorithm from 'simple' to any other value, a segfault occurs in mca_io_ompio_create_groups due to a double-free problem. Fix will be provided immediatly

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

problem is observed on master, and 3.0 

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

any configuration should lead to this problem, does not occur however with default ompio settings, only if user changes the value of the mca_io_ompio_grouping_option parameter.

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

edgargabriel:
target: v3.0.1

bwbarrett:
@edgargabriel, any reason you tagged this for 3.0.1 instead of 3.0.0?

edgargabriel:
@bwbarrett if we can bring this into 3.0.0, that would be great, since this is a regression from 2.1. I just thought that it might be too late for 3.0.0

bwbarrett:
As long as it's in before 3.0.0 releases (which looks likely), I think pulling into 3.0.0 is reasonable.  It's certainly a bugfix.

PR #3910, opened 2017-07-17
io/ompio: fix grouping option

edgargabriel:
changing the value of mca_io_ompio_grouping_option lead to a segfault due to
a double-free problem. Remove the erroneous free statements that have been introduced
and add a note ensuring that we are not re-adding them back at that spot.

fixes issue #3903


hppritcha:
@bwbarrett this one's ready to go

Issue #3918, opened 2017-07-17
Use project name when registering performance variables


hjelmn:
## Background information

The MCA variables system uses the project name to create variable groups. Unfortunately the performance variable system does not. This is a bug in the MCA base.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

This affects Open MPI v2.0.x, v2.x, and v3.0.x.

bwbarrett:
@hjelmn, I see commits for all the tagged targets; can this be closed?

PR #3876, opened 2017-07-13
mtl/psm2: add pvar support for PSM2 MQ stats

hppritcha:
Add pvars for PSM2 MQ stats to help in analyzing performance
of Omnipath.

Tested (modestly) using modified OSU pt2pt benchmarks.

Signed-off-by: Howard Pritchard <howardp@lanl.gov>

PR #3919, opened 2017-07-17
mca/base: use the project name when registering pvars

hjelmn:
References #3918. Close when applied to v2.0.x, v2.x, and v3.0.x.


hjelmn:
Because of this bug the variables in #3876 were not showing up in ompi_info.

PR #3927, opened 2017-07-18
opal/mca: Fix mca_base_verbose file suffix processing

jjhursey:
* `-mca mca_base_verbose file:foo` should create an output file with
    the suffix `foo`. But since we free the pointer at the end of this
    function then by the time we use it it is pointing to invalid memory.
 * This commit fixes that corruption
 * This commit also fixes the behavior of `file:` with no suffix.
   Makes it the same as `file` without the colon.

### DIFF
--- a/opal/mca/base/mca_base_open.c
+++ b/opal/mca/base/mca_base_open.c
@@ -221,12 +221,12 @@ static void parse_verbose(char *e, opal_output_stream_t *lds)
             have_output = true;
         }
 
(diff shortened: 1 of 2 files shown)


jjhursey:
@hjelmn since Jeff is out of town, if you have a minute can you review this PR. It's pretty small.

jjhursey:
bot:ompi:retest

hppritcha:
@hjelmn please review

PR #3930, opened 2017-07-18
Fix signal forwarding on ORTE daemons

rhc54:
Fix signal forwarding on ORTE daemons so that _all_ daemons do it, regardless of environment. Add missing support for SIGTSTP and a few others.

Thanks to Eugene Dedits for reporting the problem.


rhc54:
bot:mellanox:retest

PR #3935, opened 2017-07-19
Fix the isolated pmix component. Cleanup the ess/singleton component …

rhc54:
…- we shouldn't be automatically discovering the local topology as that is now done on-demand.

(cherry picked from commit 543c16b28dc0f51eb504cc0cc377c455e30142b6)


hppritcha:
bot:mellanox:retest

rhc54:
Fixes https://github.com/open-mpi/ompi/issues/3939

hppritcha:
@rhc54 we could use an alternate view since Jeff won't be back till next week.

rhc54:
Anyone can check it, but FWIW the original reporting user tested it and signed off that it fixes the problem.

Issue #3941, opened 2017-07-20
PMIx server direct modex response with error status is broken


artpol84:
In case direct modex response cannot be satisfied `send_error()` is called that packs status and peer id in the reply (https://github.com/open-mpi/ompi/blob/v2.0.x/orte/orted/pmix/pmix_server.c#L385:L397)
At the receiver side a "hotel room number" is expected to proceed correctly:
https://github.com/open-mpi/ompi/blob/v2.0.x/orte/orted/pmix/pmix_server.c#L597:L615.

artpol84:
It seems like this Issue can be closed as all PRs are merged.

PR #3946, opened 2017-07-20
v3.0.x: orte/pmix/server: Fix direct modex response with error status

artpol84:
`send_error()` is only packing status and peer info in the reply.
While remote counterpart in `pmix_server_dmdx_resp()` expects
the "hotel room number" to proceed correctly.

(cherry picked from commit 79c10c884d0a47c6551d246459a5f057ffba283d)


artpol84:
Address #3941

Issue #85, opened 2014-10-01
tcp btl lack of diagnostic output and bad failure modes


ompiteam:
On a very long thread on the users list (started here, http://www.open-mpi.org/community/lists/users/2009/09/10712.php, eventually moved off-list), I ended up digging through the TCP BTL connection logic.  I discovered two things:
1. If you run with --mca btl_base_verbose 100, the TCP BTL reports very little of what it is doing during wireup, making networking misconfiguration and other kinds of problems very difficult to identify.
2. If an error occurs during the TCP BTL wireup (e.g., connect(2) fails), error handling is pretty much non-existent, meaning that the PML isn't notified that anything went wrong, and the job will therefore usually hang.  ''Sometimes'' the TCP BTL prints a diagnostic error message indicating that something went wrong, but usually it does not.

I have an hg branch with a lot more diagnostic TCP BTL wireup output that I plan to put back to the trunk eventually (I actually consider this a defect, not a feature enhancement).  I also am working to make OMPI do "something reasonable" (to include possibly even aborting) if an error occurs during TCP BTL wireup.

Finally, I'll likely be adding a feature in the TCP BTL wireup: a brief handshake right after the initial connect/accept to pass a magic number to ensure that the peer is actually Open MPI (vs. some other random entity that happens to connect to my listening socket).  Ralph suggested today on the engineering call to use the version number info in the modex to check for compatible versions.

jsquyres:
I'm throwing this to IBM @gpaulsen @dsolt.  Do you guys care?

gpaulsen:
did this get into master?
Yes, extra debug info is always a good thing, unless it affects performance. :)

jsquyres:
The bitbucket that the mailing list post refers to is quite stale -- it was never directly brought back into master.

However, I think that various people have put in at least *some* of those features over time (e.g., I think @rhc54 or @bosilca put in some kind of magic number handshake...?).

It might not be a bad idea to compare the content of the commit that was cited (https://bitbucket.org/jsquyres/tcp-debug-printf/commits/9efe756cda30/) and see if similar functionality was put in.  And then make a new PR with whatever functionality was not put in.

[...]

PR #3955, opened 2017-07-24
Btl tcp: Improved diagnostic output and failure mode

mohanasudhan:
This PR handles issue https://github.com/open-mpi/ompi/issues/85 As part of this PR we focus on improving Btl tcp verbose, help and failure mode. 

Overview of changes:
1. Moved non-blocking send/receive function to btl_tcp will help reusing these function wherever needed. In this case we plan to reuse receive function to retrieve magic string to validate established connection is from mpi process.
2. Used magic string to verify mpi connection. In case if there is mismatch or missing magic string we can identify that we are trying to connect with some other process.
3. Moved few BTL_ERROR to show_help and also update the function behaviour of
mca_btl_tcp_endpoint_complete_connect to return SUCCESS and ERROR cases.
4. Improved debugging and verbose around tcp btl.

Code sourced from https://bitbucket.org/jsquyres/tcp-debug-printf/commits/9efe756cda30/

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp.c
+++ b/opal/mca/btl/tcp/btl_tcp.c
@@ -536,9 +536,13 @@ void mca_btl_tcp_dump(struct mca_btl_base_module_t* base_btl,
 
 
 /*
- * A blocking recv on a non-blocking socket. Used to receive the small
- * amount of connection information that identifies the endpoints
- * endpoint.
+ * A blocking recv for both blocking and non-blocking socket. 
+ * Used to receive the small amount of connection information 
+ * that identifies the endpoints
+ * 
+ * when the socket is blocking (the caller introduces timeout) 
+ * which happens during initial handshake otherwise socket is 
+ * non-blocking most of the time.
  */
 
 int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
--- a/opal/mca/btl/tcp/btl_tcp.c
+++ b/opal/mca/btl/tcp/btl_tcp.c
@@ -31,11 +31,13 @@
 #include "opal/mca/mpool/base/base.h"
 #include "opal/mca/mpool/mpool.h"
 #include "opal/mca/btl/base/btl_base_error.h"
+#include "opal/opal_socket_errno.h"
 
 #include "btl_tcp.h"
 #include "btl_tcp_frag.h"
 #include "btl_tcp_proc.h"
 #include "btl_tcp_endpoint.h"
+
 
 mca_btl_tcp_module_t mca_btl_tcp_module = {
     .super = {
@@ -530,4 +532,65 @@ void mca_btl_tcp_dump(struct mca_btl_base_module_t* base_btl,
         OPAL_THREAD_UNLOCK(&btl->tcp_endpoints_mutex);
     }
 #endif /* OPAL_ENABLE_DEBUG && WANT_PEER_DUMP */
+}
+
+
+/*
+ * A blocking recv on a non-blocking socket. Used to receive the small
+ * amount of connection information that identifies the endpoints
+ * endpoint.
+ */
+
+int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
+{
+    unsigned char* ptr = (unsigned char*)data;
+    size_t cnt = 0;
+    while (cnt < size) {
+        int retval = recv(sd, ((char *)ptr) + cnt, size - cnt, 0);
+        /* remote closed connection */
+        if (0 == retval) {
+            BTL_ERROR(("remote peer unexpectedly closed connection while I was waiting for blocking message"));
+            return -1;
+        }
+
+        /* socket is non-blocking so handle errors */
+        if (retval < 0) {
+            if (opal_socket_errno != EINTR &&
+                opal_socket_errno != EAGAIN &&
+                opal_socket_errno != EWOULDBLOCK) {
+                BTL_ERROR(("recv(%d) failed: %s (%d)", sd, strerror(opal_socket_errno), opal_socket_errno));
+                return -1;
+            }
+            continue;
+        }
+        cnt += retval;
+    }
+    return cnt;
+}
+
+
+/*
+ * A blocking send on a non-blocking socket. Used to send the small
+ * amount of connection information that identifies the endpoints
+ * endpoint.
+ */
+
+int mca_btl_tcp_send_blocking(int sd, const void* data, size_t size)
+{
+    unsigned char* ptr = (unsigned char*)data;
+    size_t cnt = 0;
+    while(cnt < size) {
+        int retval = send(sd, ((const char *)ptr) + cnt, size - cnt, 0);
+        if (retval < 0) {
+            if (opal_socket_errno != EINTR &&
+                opal_socket_errno != EAGAIN &&
+                opal_socket_errno != EWOULDBLOCK) {
+                BTL_ERROR(("send() failed: %s (%d)", strerror(opal_socket_errno), opal_socket_errno));
+                return -1;
+            }
+            continue;
+        }
+        cnt += retval;
+    }
+    return cnt;
 }
--- a/opal/mca/btl/tcp/btl_tcp.h
+++ b/opal/mca/btl/tcp/btl_tcp.h
@@ -359,9 +359,13 @@ int mca_btl_tcp_ft_event(int state);
 int mca_btl_tcp_send_blocking(int sd, const void* data, size_t size);
 
 /*
- * A blocking recv on a non-blocking socket. Used to receive the small
- * amount of connection information that identifies the endpoints
- * endpoint.
+ * A blocking recv for both blocking and non-blocking socket.
+ * Used to receive the small amount of connection information
+ * that identifies the endpoints
+ *
+ * when the socket is blocking (the caller introduces timeout)
+ * which happens during initial handshake otherwise socket is
+ * non-blocking most of the time.
  */
 int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size);
 
--- a/opal/mca/btl/tcp/btl_tcp.h
+++ b/opal/mca/btl/tcp/btl_tcp.h
@@ -351,5 +351,19 @@ mca_btl_tcp_dump(struct mca_btl_base_module_t* btl,
   */
 int mca_btl_tcp_ft_event(int state);
 
+/*
+ * A blocking send on a non-blocking socket. Used to send the small
+ * amount of connection information that identifies the endpoints
+ * endpoint.
+ */
+int mca_btl_tcp_send_blocking(int sd, const void* data, size_t size);
+
+/*
+ * A blocking recv on a non-blocking socket. Used to receive the small
+ * amount of connection information that identifies the endpoints
+ * endpoint.
+ */
+int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size);
+
 END_C_DECLS
 #endif
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -858,13 +858,18 @@ static int mca_btl_tcp_component_create_listen(uint16_t af_family)
         freeaddrinfo (res);
 
 #ifdef IPV6_V6ONLY
-        /* in case of AF_INET6, disable v4-mapped addresses */
+        /* If this OS supports the "IPV6_V6ONLY" constant, then set it
+           on this socket.  It specifies that *only* V6 connections
+           should be accepted on this socket (vs. allowing incoming
+           both V4 and V6 connections -- which is actually defined
+           behavior for V6<-->V4 interop stuff).  See
+           https://github.com/open-mpi/ompi/commit/95d7e08a6617530d57b6700c57738b351bfccbf8 for some
+           more details. */
         if (AF_INET6 == af_family) {
             int flg = 1;
             if (setsockopt (sd, IPPROTO_IPV6, IPV6_V6ONLY,
                             (char *) &flg, sizeof (flg)) < 0) {
-                opal_output(0,
-                    "mca_btl_tcp_create_listen: unable to disable v4-mapped addresses\n");
+                BTL_ERROR((0, "mca_btl_tcp_create_listen: unable to set IPV6_V6ONLY\n"));
             }
         }
 #endif /* IPV6_V6ONLY */
@@ -906,6 +911,10 @@ static int mca_btl_tcp_component_create_listen(uint16_t af_family)
 #else
             ((struct sockaddr_in*) &inaddr)->sin_port = htons(port + index);
 #endif  /* OPAL_ENABLE_IPV6 */
+            opal_output_verbose(30, opal_btl_base_framework.framework_output,
+                                "btl:tcp: Attempting to bind to %s port %d",
+                                (AF_INET == af_family) ? "AF_INET" : "AF_INET6",
+                                port + index);
             if(bind(sd, (struct sockaddr*)&inaddr, addrlen) < 0) {
                 if( (EADDRINUSE == opal_socket_errno) || (EADDRNOTAVAIL == opal_socket_errno) ) {
                     continue;
@@ -915,6 +924,10 @@ static int mca_btl_tcp_component_create_listen(uint16_t af_family)
                 CLOSE_THE_SOCKET(sd);
                 return OPAL_ERROR;
             }
+            opal_output_verbose(30, opal_btl_base_framework.framework_output,
+                                "btl:tcp: Successfully bound to %s port %d",
+                                (AF_INET == af_family) ? "AF_INET" : "AF_INET6",
+                                port + index);
             goto socket_binded;
         }
 #if OPAL_ENABLE_IPV6
@@ -945,6 +958,9 @@ static int mca_btl_tcp_component_create_listen(uint16_t af_family)
     if (AF_INET6 == af_family) {
         mca_btl_tcp_component.tcp6_listen_port = ((struct sockaddr_in6*) &inaddr)->sin6_port;
         mca_btl_tcp_component.tcp6_listen_sd = sd;
+        opal_output_verbose(30, opal_btl_base_framework.framework_output,
+                            "btl:tcp: my listening v6 socket port is %d",
+                            ntohs(mca_btl_tcp_component.tcp6_listen_port));
     } else
 #endif
     {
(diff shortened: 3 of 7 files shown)


ibm-ompi:

 Gist: https://gist.github.com/495751746c071046fe95798f15715bc6

ibm-ompi:

 Gist: https://gist.github.com/c3f2c22a652631ddd7b1ae5811ef30ab

ibm-ompi:

 Gist: https://gist.github.com/a6206c49ec3ea5671ec433e5fc477479

mohanasudhan:
Closing this pull request. seems i miss few things

bosilca:
Btw unlike the log message suggests you moved the blocking version of the send and receive (the ones that are only used during the initial handshake).

jsquyres:
@mohanasudhan GitHub pro tip: you don't have to close pull requests if you mean to fix the code and get it merged.  You can just add new commits to your branch and push them to your fork on github, and the pull request will automatically update to reflect the new state of the branch on your fork.

mohanasudhan:
@jsquyres Thanks. Reopening after fixing the issue

mohanasudhan:
do you know where this open-mpi-build-script.sh resides? some variants of autogen configs are failing during pull request validation. i like to run this script locally to make sure i am good before posting in pull request. I searched in ompi git but it's not there

opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -371,31 +375,18 @@ int mca_btl_tcp_endpoint_send(mca_btl_base_endpoint_t* btl_endpoint, mca_btl_tcp
 
 
 /*
- * A blocking send on a non-blocking socket. Used to send the small amount of connection
- * information that identifies the endpoints endpoint.
+ * A blocking send on a non-blocking socket. Used to send the small
+ * amount of connection information that identifies the endpoints endpoint.
  */
 static int
 mca_btl_tcp_endpoint_send_blocking(mca_btl_base_endpoint_t* btl_endpoint,
```
bosilca: Why one extra level of indirection?

opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -406,12 +397,30 @@ mca_btl_tcp_endpoint_send_blocking(mca_btl_base_endpoint_t* btl_endpoint,
 
 static int mca_btl_tcp_endpoint_send_connect_ack(mca_btl_base_endpoint_t* btl_endpoint)
 {
-    /* send process identifier to remote endpoint */
+
     opal_process_name_t guid = opal_proc_local_get()->proc_name;
```
bosilca: I don't see the benefit of having a unique magic_id_string, it can be easily generated. The original idea was to exchange a unique key only known by the processes in the same jobid (or by those where the modex info has been delivered). I don't recall it's name, but @rhc54 can certainly help here.

By doing this with a constant string in 2 steps the overhead of the connection establishment increases unnecessary.

opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -406,12 +397,30 @@ mca_btl_tcp_endpoint_send_blocking(mca_btl_base_endpoint_t* btl_endpoint,
...
 
+    int len = (int) strlen(mca_btl_tcp_magic_id_string);
+
+    /* send magic string to the remote endpoint, identifying me as a
+       fellow Open MPI TCP BTL */
+    if (mca_btl_tcp_endpoint_send_blocking(btl_endpoint,
```
bosilca: why incurring the cost of 2 blocking sends when it could have been done with a single one ?

opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -573,44 +582,51 @@ static void mca_btl_tcp_endpoint_connected(mca_btl_base_endpoint_t* btl_endpoint
...
-    size_t s;
+    size_t s, len = strlen(mca_btl_tcp_magic_id_string);;
     opal_process_name_t guid;
     mca_btl_tcp_proc_t* btl_proc = btl_endpoint->endpoint_proc;
+    char msg[1024];
+
+    /* First get magic string indicating that the connector is an Open MPI TCP BTL */
+    assert(len < sizeof(msg));
+    msg[0] = '\0';
+    s = mca_btl_tcp_endpoint_recv_blocking(btl_endpoint, msg, len);
```
bosilca: Similar to the send side, this 2 steps operation can be done with a single access to the socket.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -1300,44 +1336,133 @@ static void mca_btl_tcp_component_recv_handler(int sd, short flags, void* user)
...
     int retval;
+    char str[128];
+    size_t len = strlen(mca_btl_tcp_magic_id_string);
 
     OBJ_RELEASE(event);
 
+    /* Receive the magic string */
+    assert(len < sizeof(str));
+    str[0] = '\0';
+    /* TODO: recv_blocking() will block forever, without timeout
```
bosilca: Can setting the SO_RCVTIMEO temporary during mca_btl_tcp_recv_blocking be the solution ?

opal/mca/btl/tcp/btl_tcp_endpoint.c
```diff
@@ -62,6 +62,10 @@
 #include "btl_tcp_frag.h"
 #include "btl_tcp_addr.h"
 
+/*
+ * Magic ID string send during connect/accept handshake
+ */
+const char mca_btl_tcp_magic_id_string[] = "OPAL-TCP-BTL";
```
bosilca: We could at least add the version of the BTL TCP component. More about this a little later.

opal/mca/btl/tcp/btl_tcp_component.c
mohanasudhan: SO_RCVTIMEO will help in this case will update

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: Thanks for your comment. After discussing with @bwbarrett, here is my understanding. Adding a unique key to identify connection across processes in same job is a problem which should solve and currently i am not addressing that in this PR. I can iterate over in my next PR for it. Here we are just identifying where established connection is from MPI process.

Ya i agree adding 2 steps (send + recv of const string) is increasing overhead here. Do you have any suggestion here?

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: We are targeting on verifying established connection is from MPI process. We are not verifying its belonging to same version or job here. Can add one new PR for it if it's one of the failure case we face today

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: since i moved mca_btl_tcp_recv_blocking to use it across multiple places. I moved mca_btl_tcp_send_blocking with the intention it will be shared too

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: Haven't thought about it. will update to single call if not an issue. will clarify with @bwbarrett

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: as commented above

opal/mca/btl/tcp/btl_tcp_endpoint.c
bosilca: These are 2 different issues. Here I was suggesting to add the BTL module/component version, to make sure we dont connect incompatible TCP modules.

opal/mca/btl/tcp/btl_tcp_endpoint.c
bosilca: asprintf the entire id and send it at once ? Or extend the blocking API to take an iovec (this second approach seems cleaner).

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: Update to single call

opal/mca/btl/tcp/btl_tcp_component.c
mohanasudhan: Updated

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: done

opal/mca/btl/tcp/btl_tcp_endpoint.c
mohanasudhan: updated

opal/mca/btl/tcp/btl_tcp.c
```diff
@@ -531,3 +533,64 @@ void mca_btl_tcp_dump(struct mca_btl_base_module_t* base_btl,
...
 #endif /* OPAL_ENABLE_DEBUG && WANT_PEER_DUMP */
 }
+
+
+/*
+ * A blocking recv on a non-blocking socket. Used to receive the small
+ * amount of connection information that identifies the endpoints
+ * endpoint.
+ */
+
```
bosilca: This function (and the corresponding send) should now be updated to handle the timeout flags. Reading through the socket(7) man page, the SO_RCVTIMEO and SO_SNDTIMEO would force the return with EAGAIN, EWOULDBLOCK or EINPROGRESS. Currently these return values are not correctly handled (most of the are ignored).

The comment states that the socket is in non-blocking mode. If this was true then SO_RCVTIMEO would have no impact, as the recv would return EWOULDBLOCK as soon as no data is pending on the socket and the timeout would never trigger. Luckily the comment is not accurate as when this mca_btl_tcp_recv_blocking is called, from mca_btl_tcp_component_recv_handler (thus from the accept callback), the socket is still in blocking mode (as we use accept). That being said my comment just above should be addressed.

mohanasudhan: I used fcntl to see whether it is a blocking/non-blocking socket. Here is my finding, Send is always non-blocking call.  In case of recv, which is called from two place 
1. mca_btl_tcp_endpoint_recv_connect_ack (from btl_tcp_endpoint.c) - which is non blocking 
2. mca_btl_tcp_component_recv_handler (from btl_tcp_component.c) - which is blocking where we are using SO_RCVTIMEO. 

I will update comment and handle EINPROGRESS (which i missed in recv)

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -729,7 +732,9 @@ static int mca_btl_tcp_component_create_instances(void)
         char* if_name = *argv;
         int if_index = opal_ifnametokindex(if_name);
         if(if_index < 0) {
-            BTL_ERROR(("invalid interface \"%s\"", if_name));
+            opal_show_help("help-mpi-btl-tcp.txt", "invalid if_inexclude",
```
mohanasudhan: Ya https://github.com/mohanasudhan/ompi/blob/master/opal/mca/btl/tcp/help-mpi-btl-tcp.txt#L18 the help message says "An invalid value was given for btl_tcp_if_%s" refering both [in|ex]clude.

bosilca: Not really. If I correctly remember the matching of opal_show_help the %s will match if_name, so for eth0 you will get 
```
WARNING: An invalid value was given for btl_tcp_if_eth0.  This
value will be ignored.
```
Not really what we want.

opal/mca/btl/tcp/btl_tcp.c
bosilca: Argh, I forgot about the second path from the connect_ack. In the case where the socket is in non-blocking mode the timeout attached to the socket will not be useful. We need to fix this in such a way to cover both cases.

mohanasudhan: Just wondering since EINPROGRESS is not handled in the error cases of mca_btl_tcp_recv_blocking, we will loop through till we get value. so I hope EINPROGRESS is also handled indirectly here. Correct me if i am wrong.  

https://github.com/mohanasudhan/ompi/blob/master/opal/mca/btl/tcp/btl_tcp.c#L544

Definetly comments has to be improved here. Let me know if i missing anything. Thanks for comments.

mohanasudhan: Note: we are setting SO_RCVTIMEO only in mca_btl_tcp_component_recv_handler where it is blocking call. so we are handling both cases. But I feel, I need to change the function comments and add more details to make it more clear (may be change function name here)

bosilca: You are right, the timeout is not set on the connect_ack part. I poorly expressed myself. What I really meant is that we are not getting out of the recv_blocking function with error once the expected timeout passed. SO_RCVTIMEO might solve the pb for the blocking socket, because we will eventually trigger the internal timeout (even if we receive byte by byte). But I do not see anything that would help with the timeout in the case where the socket is in non-blocking mode.

bwbarrett: Yeah, we know that this is still an issue.  I wanted @mohanasudhan to get the first round of changes committed and then tackle that change, just to minimize the number of changes at once.

mohanasudhan: @bosilca  Ya exactly. But I am not clear what's the change proposed here as we use SO_RCVTIMEO for blocking and we don't do anything in non-blocking mode.

bwbarrett:
wow, no idea what happened with the OMPI builder: https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.autogen_options/759/.  Let's do the CS thing and retry.

bot:ompi:retest

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -732,8 +732,8 @@ static int mca_btl_tcp_component_create_instances(void)
         char* if_name = *argv;
         int if_index = opal_ifnametokindex(if_name);
         if(if_index < 0) {
-            opal_show_help("help-mpi-btl-tcp.txt", "invalid if_inexclude",
-                           true, if_name, opal_process_info.nodename,
+            opal_show_help("help-mpi-btl-tcp.txt", "invalid interface",
```
bwbarrett: I liked the old message better.  I just think you wanted to use "include" instead of if_name as the first positional argument.  The problem with this error message is that it won't be clear to the user where they specified the invalid interface name.

opal/mca/btl/tcp/btl_tcp.c
```diff
@@ -531,3 +533,64 @@ void mca_btl_tcp_dump(struct mca_btl_base_module_t* base_btl,
     }
 #endif /* OPAL_ENABLE_DEBUG && WANT_PEER_DUMP */
 }
+
+
+/*
+ * A blocking recv on a non-blocking socket. Used to receive the small
```
bwbarrett: This comment is no longer true.  The socket is blocking (but with a receive timeout) during the initial handshake.   Does it make sense to continue using this wrapper in the initial handshake receive?  Possibly, as you'd need a bit of a loop to handle EINTR/EAGAIN in the case of a signal.  But then you need to update the comment with an explanation that usually (but not always) the socket is in non-blocking.

opal/mca/btl/tcp/btl_tcp_component.c
mohanasudhan: ah.. i got confused. fixed now.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -1299,45 +1338,156 @@ static void mca_btl_tcp_component_recv_handler(int sd, short flags, void* user)
...
+    size_t retval, len = strlen(mca_btl_tcp_magic_id_string);
+    mca_btl_tcp_endpoint_hs_msg_t hs_msg;
+    struct timeval save, tv;
+    socklen_t rcvtimeo_save_len = sizeof(save);
+    char str[128];
+
+    /* Setting SO_RCVTIMEO to say 2 seconds here to avoid chance of spin
+     * forever if it tries to connect to old version
+     * as older version will send just process id which won't be long enough
+     * to cross sizeof(str) length + process id struct*/
```
bwbarrett: Just to hit those of us who aren't that smart over the head a bit more, I'd add to this comment that you leave the socket in blocking mode until the initial receive is complete, so that RCVTIMEO does something sane.  Possibly explain the usage of RCvTIMEO (to handle the case where the remote side isn't OMPI and isn't going to send us any data).  The why is important :).

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -1299,45 +1338,156 @@ static void mca_btl_tcp_component_recv_handler(int sd, short flags, void* user)
...
+                            opal_process_info.nodename,
+                            getpid(), "server", hs_msg.magic_id,
+                            "string value");
+        /* The other side probably isn't OMPI, so just hang up */
         CLOSE_THE_SOCKET(sd);
         return;
     }
+    
+    if (sockopt) {
+       /* return the socket to normal */
```
bwbarrett: it may not be "normal".  I'd say "reset the RECVTIMEO option to its original state".

opal/mca/btl/tcp/btl_tcp.c
mohanasudhan: updated comment

opal/mca/btl/tcp/btl_tcp_component.c
mohanasudhan: updated comment

opal/mca/btl/tcp/btl_tcp_component.c
mohanasudhan: updated comment

mohanasudhan:
Squashed commits.

opal/mca/btl/tcp/btl_tcp.h
```diff
@@ -351,5 +351,19 @@ mca_btl_tcp_dump(struct mca_btl_base_module_t* btl,
   */
 int mca_btl_tcp_ft_event(int state);
 
+/*
+ * A blocking send on a non-blocking socket. Used to send the small
```
bwbarrett: you missed a comment update :).  The socket may or may not be non-blocking...

mohanasudhan: socking is non-blocking while send. I have already updated recv method comments 

+ * A blocking recv for both blocking and non-blocking socket. 
 + * Used to receive the small amount of connection information 
 + * that identifies the endpoints
 + * 
 + * when the socket is blocking (the caller introduces timeout) 
 + * which happens during initial handshake otherwise socket is 
 + * non-blocking most of the time.

mohanasudhan: My bad. I was looking at main PR page and didnt realize its in header file.

mohanasudhan:
bot:ompi:retest

PR #3957, opened 2017-07-25
Add missing export directives

rhc54:


bosilca:
Thanks @rhc54 it does fix the pb.

PR #3958, opened 2017-07-25
v3.0.x: mtl-portals4: add timeout to rendezvous get fragments

tkordenbrock:
This PR adds a timeout to each fragment of a rendezvous get. If any fragment times out or fails, the entire receive fails.

(cherry picked from master commits 06b15ce, 99453e6, 37766d7, 5ecd905)

Signed-off-by: Todd Kordenbrock thkgcode@gmail.com


bwbarrett:
@tkordenbrock / @regrant , are these cherry picks from master?  If so, can you add a note to that effect with the master commit id?

tkordenbrock:
@bwbarrett Sorry about that.  PR updated with master commits.

PR #3961, opened 2017-07-25
Update the tools support so it allows tools to access PMIx

rhc54:


rhc54:
@artpol84 This reported failure makes no sense to me - is something else wrong here? I looked at the test and then tried to run it manually (so far as I could figure out what it is doing):
```shell
$ mpirun -n 1 -host rhc001 -mca mca_base_env_list 'XXX_A=7;XXX_B=8' env
...
XXX_A=7
XXX_B=8
...
```
so it looks like it does the right thing. What am I missing?

rhc54:
bot:mellanox:retest

PR #3971, opened 2017-07-26
fix srun latency, change default yield_when_idle=0

plesn:
This changes the default to 0, to avoid yields during progress in srun.

In mpirun, ompi_mpi_yield_when_idle is set to 1 if oversubscribed
otherwise 0. But the default is 1 though, and it is used in srun.
Now srun and mpirun have the same latency in non-oversubscribed cases.


hjelmn:
You need to add a signed off line. Quickest way is ```git commit --amend -s --no-edit```

jjhursey:
ok to test

plesn:
Added the sign-off

Issue #3977, opened 2017-07-28
v3.0.x: PMIx v1.2 external doesn't work with v3.0.x


karasevb:
The problem reproduced with v3.0.x and PMIx v1.2 as external component and`hello_c` program.

There is a hang when the client is trying to request a key from the server:
```
(gdb) bt
#0  0x00007f8e9378ff3d in nanosleep () at ../sysdeps/unix/syscall-template.S:81
#1  0x00007f8e937c14a4 in usleep (useconds=<optimized out>) at ../sysdeps/unix/sysv/linux/usleep.c:32
#2  0x00007f8e90748997 in PMIx_Get (proc=0x7ffd021e0a10, key=0x7f8e913b21bf "pmix.appnum", info=0x0, ninfo=0, val=0x7ffd021e0a00) at src/client/pmix_client_get.c:104
#3  0x00007f8e909a3cae in pmix1_get () from /home/user/ompi/v3.0.x/v1.2_ext/install/lib/openmpi/mca_pmix_ext1x.so
#4  0x00007f8e913b13c8 in rte_init () from /home/user/ompi/v3.0.x/v1.2_ext/install/lib/openmpi/mca_ess_singleton.so
#5  0x00007f8e93437929 in orte_init () from /home/user/ompi/v3.0.x/v1.2_ext/install/lib/libopen-rte.so.40
#6  0x00007f8e93cffa75 in ompi_mpi_init () from /home/user/ompi/v3.0.x/v1.2_ext/install/lib/libmpi.so.40
#7  0x00007f8e93d21deb in PMPI_Init () from /home/user/ompi/v3.0.x/v1.2_ext/install/lib/libmpi.so.40
#8  0x00000000004009ae in main ()
```

karasevb:
@artpol84

hppritcha:
This is a known limitation.  We'll make sure its documented.

artpol84:
@hppricha, we need 1.2 working in 3.0. It seems that I missed the decision of making this problem just a "known limitation".

bwbarrett:
We won't hold 3.0.0rc2 (which has already been cut, will be released today), but I think it's reasonable to pull these in for 3.0.0, assuming @rhc54 is on board.

artpol84:
Sounds good.

jladd-mlnx:
@karasevb @artpol84 please close the issue once satisfied that it's working.

artpol84:
Yes, it's working now.

PR #3978, opened 2017-07-28
pmix: fixed immediate request

karasevb:
This commit fixes a hang when using external PMIx v1 module

refs #3977


artpol84:
@karasevb
So this is only for a singletons?

karasevb:
@artpol84 Not only for singletons. Also problem has existed for multi ranks.

artpol84:
Need another PR for that?

karasevb:
@artpol84 
No. This fix targeting to both cases. There are two fixes, one of them fixes the optional request for only singleton, and second fixes the info-key for the immediate request that leads to hang for both cases (for single and multi ranks).

bwbarrett:
@jladd-mlnx, please don't add 3.0.0 milestone to a patch on master.  Feel free to add it to the issue (and add the blocker label so we won't release 3.0.0 before the issue is fixed), but this particular PR doesn't impact 3.0.0 directly and adding milestones just makes the release managers' job harder.

jladd-mlnx:
@bwbarrett Roger that.

PR #3985, opened 2017-07-31
Check for OPAL_PREFIX and set corresponding PMIX_PREFIX if found

### LINKED ISSUE #3980: OMPI v3.0: pmix install dirs info is not propagated to orteds
### Open MPI version

gitclone ompi v3.0.x (2f13cce)

## Details of the problem

If OMPI installation is moved to a different location `OPAL_PREFIX` env var is used to identify that and `installdirs/env`  is handling this correctly.
PMIx also has MCA infrastructure and similar installdirs component. in PMIx `PMIX_INSTALL_PREFIX` playing the role of `OPAL_PREFIX`. The important difference is that `OPAL_PREFIX` is an OMPI variable and it gets propagated to orte daemons, i.e.:
```
[cn01:17570] [[10714,0],0] plm:rsh: final template argv:
/usr/bin/ssh <template> \
	OPAL_PREFIX=<ompi-path>/ompi-v3.0.x ; export OPAL_PREFIX; \
	PATH=<ompi-path>/ompi-v3.0.x/bin:$PATH ; export PATH ; \
	LD_LIBRARY_PATH=<ompi-path>/ompi-v3.0.x/lib:$LD_LIBRARY_PATH ; export LD_LIBRARY_PATH ; \
	DYLD_LIBRARY_PATH=<ompi-path>/ompi-v3.0.x/lib:$DYLD_LIBRARY_PATH ; export DYLD_LIBRARY_PATH ;   \
	<ompi-path>/ompi-v3.0.x/bin/orted -mca orte_debug_daemons "1" -mca ess "env" -mca ess_base_jobid "702152704" \
	-mca ess_base_vpid "<template>" -mca ess_base_num_procs "2" -mca orte_node_regex "cn01,cn02@0(2)" \
	-mca orte_hnp_uri "702152704.0;tcp://<IP1>,<IP2>;ud://<UD>" -mca coll_hcoll_enable "1" -mca pml "yalla" \
	--mca plm_base_verbose "100" -mca plm "rsh" -mca rmaps_base_mapping_policy "node" \
	-mca hwloc_base_binding_policy "core" -mca rmaps_base_display_map "1"
```
As it can be seen there `PMIX_INSTALL_PREFIX` is not propagated so orteds are failing with:
```
Sorry!  You were supposed to get help about:
    listener-thread-start
But I couldn't open the help file:
    <OLD-INSTALLATION_PATH>/ompi-v3.0.x/share/pmix/help-pmix-server.txt: No such file or directory.  Sorry!
```
Because they can't find PMIx mca ptl components.

rhc54:
Fixes #3980 

(cherry picked from commit 3744967adb8dcb2fc7b802c46bc6153ecef59ce6)

### DIFF
--- a/orte/mca/ess/base/ess_base_std_prolog.c
+++ b/orte/mca/ess/base/ess_base_std_prolog.c
@@ -9,7 +9,6 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2017      Intel, Inc.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -29,7 +28,6 @@
 #include <unistd.h>
 #endif
 
-#include "opal/util/opal_environ.h"
 #include "orte/mca/errmgr/errmgr.h"
 #include "orte/util/show_help.h"
 #include "orte/runtime/orte_wait.h"
@@ -41,7 +39,6 @@ int orte_ess_base_std_prolog(void)
 {
     int ret;
     char *error = NULL;
-    char *evar;
 
     /* Initialize the ORTE data type support */
     if (ORTE_SUCCESS != (ret = orte_dt_init())) {
@@ -58,10 +55,6 @@ int orte_ess_base_std_prolog(void)
             error = "orte_wait_init";
             goto error;
         }
-    }
-    /* check for OPAL_PREFIX */
-    if (NULL != (evar = getenv("OPAL_PREFIX"))) {
-        opal_setenv("PMIX_PREFIX", evar, false, &environ);
     }
 
     return ORTE_SUCCESS;
--- a/orte/mca/ess/base/ess_base_std_prolog.c
+++ b/orte/mca/ess/base/ess_base_std_prolog.c
@@ -9,6 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
+ * Copyright (c) 2017      Intel, Inc.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -28,6 +29,7 @@
 #include <unistd.h>
 #endif
 
+#include "opal/util/opal_environ.h"
 #include "orte/mca/errmgr/errmgr.h"
 #include "orte/util/show_help.h"
 #include "orte/runtime/orte_wait.h"
@@ -39,6 +41,7 @@ int orte_ess_base_std_prolog(void)
 {
     int ret;
     char *error = NULL;
+    char *evar;
 
     /* Initialize the ORTE data type support */
     if (ORTE_SUCCESS != (ret = orte_dt_init())) {
@@ -55,6 +58,10 @@ int orte_ess_base_std_prolog(void)
             error = "orte_wait_init";
             goto error;
         }
+    }
+    /* check for OPAL_PREFIX */
+    if (NULL != (evar = getenv("OPAL_PREFIX"))) {
+        opal_setenv("PMIX_PREFIX", evar, false, &environ);
     }
 
     return ORTE_SUCCESS;
--- a/opal/mca/pmix/pmix2x/pmix2x_client.c
+++ b/opal/mca/pmix/pmix2x/pmix2x_client.c
@@ -29,6 +29,7 @@
 #include "opal/hash_string.h"
 #include "opal/threads/threads.h"
 #include "opal/util/argv.h"
+#include "opal/util/opal_environ.h"
 #include "opal/util/proc.h"
 
 #include "opal/mca/pmix/base/base.h"
@@ -64,6 +65,7 @@ int pmix2x_client_init(opal_list_t *ilist)
     pmix_info_t *pinfo;
     size_t ninfo, n;
     opal_value_t *ival;
+    char *evar;
 
     opal_output_verbose(1, opal_pmix_base_framework.framework_output,
                         "PMIx_client init");
@@ -74,6 +76,9 @@ int pmix2x_client_init(opal_list_t *ilist)
         if (0 < (dbg = opal_output_get_verbosity(opal_pmix_base_framework.framework_output))) {
             asprintf(&dbgvalue, "PMIX_DEBUG=%d", dbg);
             putenv(dbgvalue);
+        }
+        if (NULL != (evar = getenv("OPAL_PREFIX"))) {
+            opal_setenv("PMIX_PREFIX", evar, false, &environ);
         }
     }
 
--- a/opal/mca/pmix/pmix2x/pmix2x_server_south.c
+++ b/opal/mca/pmix/pmix2x/pmix2x_server_south.c
@@ -36,6 +36,7 @@
 #include "opal/util/argv.h"
 #include "opal/util/error.h"
 #include "opal/util/output.h"
+#include "opal/util/opal_environ.h"
 #include "opal/util/proc.h"
 #include "opal/util/show_help.h"
 #include "opal/mca/pmix/base/base.h"
@@ -99,6 +100,7 @@ int pmix2x_server_init(opal_pmix_server_module_t *module,
     opal_pmix2x_event_t *event;
     opal_pmix2x_jobid_trkr_t *job;
     opal_pmix_lock_t lk;
+    char *evar;
 
     OPAL_PMIX_ACQUIRE_THREAD(&opal_pmix_base.lock);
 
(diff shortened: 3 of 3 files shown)


artpol84:
@rhc54 I'm going to runtime-check it and report back

orte/mca/ess/base/ess_base_std_prolog.c
```diff
@@ -56,6 +59,10 @@ int orte_ess_base_std_prolog(void)
             goto error;
         }
     }
+    /* check for OPAL_PREFIX */
+    if (NULL != (evar = getenv("OPAL_PREFIX"))) {
+        opal_setenv("PMIX_PREFIX", evar, false, &environ);
```
artpol84: @rhc54 while I'm checking if this fixes my particular case I have a coment.
This fix is focused on internal PMIx. This may break external pmix in following cases:
* if `OPAL_PREFIX` is set - we will set `PMIX_PREFIX` and this will mislead external pmix
* if no `OPAL_PREFIX` provided but user still needs to guide external PMIx - this fix will not help.
A;so

rhc54: we can fix the first case by moving this down into opal/mca/pmix/pmix2x itself (note that you can't put it in the ORTE code as that wouldn't help the direct launch situation). There is no way to fix the other case - the user will just have to provide PMIX_PREFIX in that scenario.

artpol84: about the second case: that is fine to provide `PMIX_PREFIX`, the problem is it won't be propagated.

rhc54: The second case will indeed be propagated. Today, mpirun harvests all envars starting with OMPI_ and PMIX_ and forwards them. If someone direct launches, then the launcher is responsible for forwarding the environment. Thus, if someone sets PMIX_PREFIX in their environment, it should get to the backend. There is no way we can determine whether or not PMIX_PREFIX is required for the external PMIx installation.

So the only case we have to deal with is the one where OPAL_PREFIX is set and we need to ensure that PMIX_PREFIX corresponds. As you correctly pointed out, this only happens when we are using the embedded version of PMIx.

artpol84:
But that's what my issue was all about. We are setting PMIX_INSTALL_PREFIX and it is not propagated.

rhc54:
@artpol84 I'm not sure what you mean - in master, we definitely pickup and propagate everything starting with PMIX_. Is your concern that this code hasn't moved to the 3.0 branch? FWIW: It is PMIX_PREFIX, not PMIX_INSTALL_PREFIX - are you setting the right variable?

The problem this PR solves is that someone setting OPAL_PREFIX expects the embedded PMIx to also adjust its plugin location. So we need to detect OPAL_PREFIX and manually force PMIX_PREFIX in that case. Since OPAL_PREFIX is picked up and propagated, we don't have to worry about doing PMIX_PREFIX separately - we can just locally do the right thing.

rhc54:
Just checked and 3.0 is indeed picking up and propagating all PMIX_ envars

artpol84:
As I updated on #3980  (https://github.com/open-mpi/ompi/issues/3980#issuecomment-319413181) I agree that auto-set PMIX search path if OPAL one is used is a good idea, it is related to #3980 but it doesn't fix #3980.

I don't have concerns - I see at runtime that v3.0.x is not able to work if sources are moved to another location and multiple nodes are used (see https://github.com/open-mpi/ompi/issues/3980#issuecomment-319413181). I see that `PMIX_INSTALL_PREFIX` is not propagated although it starts from `PMIX_`.

As far as `PMIX_INSTALL_PREFIX`:
* according to pmix's installdirs component `PMIX_INSTALL_PREFIX` is responsible for install dir prefix ( [see here](https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/pmix2x/pmix/src/mca/pinstalldirs/env/pmix_pinstalldirs_env.c#L65) ). for OMPI `OPAL_PREFIX` plays that role ( [see here](https://github.com/open-mpi/ompi/blob/master/opal/mca/installdirs/env/opal_installdirs_env.c#L64) ). That is the reason I'm using it.
* I verified at runtime that `PMIX_INSTALL_PREFIX` is fixing the problem for hnp-only run (i.e. where `mpirun` is the only orte daemon). This is because I'm setting `PMIX_INSTALL_PREFIX` and `mpirun` sees it. If I go multi-node - this variable is not propagated and `orted`'s are unable to find pmix components.

Have you tried to reproduce the problem in your environment?

artpol84:
`Just checked and 3.0 is indeed picking up and propagating all PMIX_ envars`
I'm going to double-check now. 
Can you tell what was your way of testing that?

artpol84:
Also - propagation to what entity have you tested? To `orted` or to the application process?
Because the problem I see is the result of orted's missing the PMIX_ prefix info.

artpol84:
```
As I updated on #3980  (https://github.com/open-mpi/ompi/issues/3980#issuecomment-319413181) I agree that auto-set PMIX search path if OPAL one is used is a good idea, it is related to #3980 but it doesn't fix #3980.
```
Correction: I expect this PR to fix the case where internal PMIx is used. However I have concerns about external one then.

rhc54:
@artpol84 I honestly don't have any more time to spend on this one. I would suggest you do two things:

* change PMIX_PREFIX in the master commit I did to be PMIX_INSTALL_PREFIX since we changed it to that in PMIx (not sure why, but doesn't matter)

* go into the plm/rsh component and duplicate the code for forwarding OPAL_PREFIX to do the same logic for PMIX_INSTALL_PREFIX. That way if someone sets PMIX_INSTALL_PREFIX  and not OPAL_PREFIX, we'll forward it for the rsh case. If someone sets both to different values, my commit won't override their setting.

Those are the only cases we can impact. We cannot do anything for the cases where we direct launch or launch via the other plm components - the system admin will have to configure their environment to forward the necessary envars.

artpol84:
that sounds ok, @rhc54.
One more question. We have couple of envars in the installdir component:
https://github.com/open-mpi/ompi/blob/master/opal/mca/installdirs/env/opal_installdirs_env.c#L64:L80
I guess we need to forward all of them, please correct me if I am wrong.

artpol84:
I've addressed what was suggested in https://github.com/open-mpi/ompi/pull/3985#issuecomment-319529424. This solves my problem and I believe will solve 99.9% of cases.
I verified that it works both in Slurm-managed allocation and with rsh/hostfile.
What we still may miss (0.1% of cases) is that user may have a finer grained installation:
https://github.com/open-mpi/ompi/blob/master/opal/mca/installdirs/env/opal_installdirs_env.c#L64:L80
and to support this more envars would need to be accounted.
With respect to plm/rsh - only `OPAL_PREFIX` is now propagated. So I can conclude tat rsh do not support this finer granularity even from OMPI perspective.

artpol84:
@bwbarrett @hppritcha 
Once #4004 is merged I'll port it to v3.0.x and I will need this to be there to merge cleanly..

artpol84:
@rhc54 
```
>>>     CID 1416016:  Insecure data handling  (TAINTED_STRING)
>>>     Assigning: "evar" = "getenv("OPAL_PREFIX")", which taints "evar".
81             if (NULL != (evar = getenv("OPAL_PREFIX"))) {
82                 opal_setenv("PMIX_PREFIX", evar, false, &environ);
83             }
84         }
```
this came up recently. I guess what it means is that env var may be constructed in the way that may (un)intentionally break the logic.
I guess thi is not the only place where we use getenv, what is the common policy for that?

rhc54:
We ignore those coverity issues. What they mean is that the envar could have been something that causes you to do evil things, and we don't restrict the potential values to avoid that scenario. In these cases, there is no way to know if a value is right or not as the range is any valid path. Technically, we could check the string to see if it points to a valid path - but then you'd get the same warning because you are using the string in a function without first checking to see if it is valid.

Can't win - we ignore.

artpol84:
@hppritcha @bwbarrett 
Ok, I need this PR merged to port #4004.

rhc54:
You are welcome to push the cherry-pick into my PR - will retain your authorship

hppritcha:
@artpol84 are you going to add #4004 to this PR or would you prefer this be merged in as is?

artpol84:
lets merge as is.

rhc54:
@hppritcha @bwbarrett Just to be clear, the combination of this PR and #4004 does not fully address the problem. You are welcome to piecemeal the commits if you like, or we can roll them all up. Totally up to you.

bwbarrett:
@artpol84 and @rhc54, I have to admit it's pretty unclear where we are.  #4004 is a commit to master; it doesn't look like there's a corresponding commit to 3.0.  I'd really prefer that we have one PR which covers cherry picking whatever we need from master to entirely solve this issue.  Continuing to piece meal is really confusing at this point.

artpol84:
@bwbarrett I guess it is on @rhc54 side now: https://github.com/open-mpi/ompi/pull/4007#issuecomment-319837018.

#4004 depends on this PR, @rhc54 can cherry-pick those commits here if it will be more convenient.

artpol84:
If we will go with one PR - let me know when all the changes are there and I will check at runtime again.

jjhursey:
Ralph is traveling today. I spoke with him this morning briefly. He posted PR #4012 for discussion. I think we should wait until tomorrow to merge any more in on this topic. Maybe @rhc54 @artpol84 and I can get on the phone tomorrow and sort though this a bit.

@bwbarrett I agree that one PR with all of the necessary commits is what we should wait for for `v3.0.x`. It shouldn't take long to sort out, and we'll probably have it ready be COB tomorrow. Just want to wait for Ralph before moving forward too much more. I think watching Issue #3980 is the best at the moment.

rhc54:
Dropped in favor of #4052

PR #3987, opened 2017-08-01
Topic/backport 3930 - Fix signal forwarding on ORTE daemons

hppritcha:
Patch PR #3930 back on to v2.x.  Turns out additional signals added to 3930 are very much needed for at least one customer trying to deploy v2.1.x on their SLURM cluster.

### DIFF
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -12,7 +12,7 @@
  * Copyright (c) 2009      Institut National de Recherche en Informatique
  *                         et Automatique. All rights reserved.
  * Copyright (c) 2011      Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2011-2015 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2011-2017 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2015 Intel, Inc.  All rights reserved.
  * Copyright (c) 2017      IBM Corporation. All rights reserved.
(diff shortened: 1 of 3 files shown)


orte/mca/ess/base/ess_base_std_orted.c
```diff
@@ -111,6 +112,10 @@ int orte_ess_base_orted_setup(char **hosts)
     char *param;
     hwloc_obj_t obj;
     unsigned i, j;
+    orte_topology_t *t;
+    opal_list_t transports;
```
rhc54: Other than generating an unused variable warning, these two lines won't hurt but are not really part of this change. You might want to remove them.

hppritcha:
@rhc54 check now.  thanks.

hppritcha:
@jsquyres please double check - this is needed for LANL

orte/mca/ess/base/ess_base_std_orted.c
```diff
@@ -611,13 +625,22 @@ int orte_ess_base_orted_setup(char **hosts)
...
+    unsigned int i;
+
     if (signals_set) {
-        /* Release all local signal handlers */
         opal_event_del(&epipe_handler);
         opal_event_del(&term_handler);
         opal_event_del(&int_handler);
-        opal_event_signal_del(&sigusr1_handler);
-        opal_event_signal_del(&sigusr2_handler);
+        /** Remove the USR signal handlers */
```
jsquyres: This is a nit pick, but: I don't think this comment matches the code.  The loop below doesn't remove (just) SIGUSR1 and SIGUSR2 like the comment implies.  This comment should likely be fixed or removed (at least on master -- here on the release branch, it's probably not as important) so as not to cause confusion in the future.

hppritcha: hmmm...okay I think that was from the original cherry-pick.  will update on master.

Issue #3998, opened 2017-08-01
ORTE needs to work better with SLURM signal mechanisms


hppritcha:
Our site (and other DOE CTS-1 system owners) are having problems with Open MPI/ORTEs interaction with SLURM's mechanism for propagating signals to processes in a job, in particular when using SLURM scancel of sbatch + --signal option.

The problem has to do with what's going on with the HNP, namely, that MPI processes are currently children of the HNP rather than a slurmd daemon.   SLURM doesn't know about the HNP or any child processes, so they are left out of any SLURM signaling.   Note the Cray SLURM environment is currently being treated differently so it doesn't have this problem.

bwbarrett:
@hppritcha, which branches does this impact?

ggouaillardet:
@hppritcha what is the behavior you would like/expect ?

1. slurm signals the HNP (and probably the tasks running on this node too), and then Open MPI does the signal propagation
2. slurm signals the HNP (but not its children), and then Open MPI does the signal propagation
3. slurm signals all processes (shells, mpirun, orted, mpi tasks, ...)
4. slurm signals mpirun (and some shells) and orted, and it is up to Open MPI to kill the MPI tasks

hppritcha:
@bwbarrett oops forgot about this one.  Since it adds an mca parameter, I didn't think we normally would push this kind of a change back to current release branches.  My site is okay with a patch to solve this issue and we're going that route locally for the 2.1.x.  We are okay with this waiting till the next major release.

rhc54:
Completed

PR #3999, opened 2017-08-01
SLURM: launch all processes via slurmd

hppritcha:
It turns out that the approach of having the HNP do the
fork/exec of MPI ranks on the head node in a SLURM environment
introduces problems when users/sysadmins want to use the SLURM
scancl tool or sbatch --signal option to signal a job.

This commit disables use of the HNP fork/exec procedure when
a job is launched into a SLURM controlled allocation.

related to #3998

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -65,6 +65,11 @@ Master (not on release branches yet)
   via --enable-mpi-cxx.
 - Removed embedded VampirTrace.  It is in maintenance mode since 2013.
   Please consider Score-P (score-p.org) as an external replacement.
+- Add a mca parameter ras_base_launch_orted_on_hn to allow for launching
+  MPI processes on the same node where mpirun is executing using a separate
+  orte daemon, rather than the mpirun process.   This may be useful to set to
+  true when using SLURM, as it improves interoperability with SLURM's signal
+  propagation tools.  By default it is set to false, except for Cray XC systems.
 
 3.0.0 -- July, 2017
 -------------------
--- a/orte/mca/ras/base/base.h
+++ b/orte/mca/ras/base/base.h
@@ -51,6 +51,7 @@ typedef struct orte_ras_base_t {
     orte_ras_base_module_t *active_module;
     int total_slots_alloc;
     int multiplier;
+    bool launch_orted_on_hn;
 } orte_ras_base_t;
 
 ORTE_DECLSPEC extern orte_ras_base_t orte_ras_base;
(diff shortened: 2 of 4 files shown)


rhc54:
I don't personally have an issue with making this change. However, this is a _big_ change in behavior for SLURM users that does have consequences - i.e., there is now another daemon on the head node impacting performance. I would much prefer to make this optional instead of required, with the default being the old behavior until we give users a chance to know this is coming and adapt.

hppritcha:
sounds like a good.  Maybe a configure option? Or should it be an mca parameter for ras framework?

rhc54:
I would do it as an MCA param in the RAS component so it can be overridden (in either direction) by a user, especially if some problem turns up that we missed and/or didn't anticipate.

rhc54:
Just to document the reasons behind the caution. There has been considerable research done regarding the impact of daemons sitting on compute nodes. Even daemons that block still contribute to jitter on the node, thus slowing down the application procs on that node.

mpirun executing on a compute node where application procs are running causes additional impact. Every process that generates stdout/err causes mpirun to wake up - ditto for the passing of stdin. The result is that procs on the head node wind up being measurably slower than their peers.

For embarrassingly parallel applications, this doesn't have too much impact other than slightly lengthening time to solution. Experiments have shown up to a 3% impact in that regard. However, more complex applications, especially those utilizing collectives, see an increased likelihood of application failure as the impacted procs continue to lag behind.

Thus, the direction has been to reduce and/or eliminate daemons from the compute nodes. This is one of the objectives of the PMIx effort - integration with the RM provides access to the info and services that otherwise would require mpirun and its daemons.

Adding another daemon to the head node will increase the impact on the application in both time to solution and probability of failure. This is why it is important to allow users to "opt out".

Hope the explanation helps document the reason for caution.

hppritcha:
@rhc54 check now.  added a ras mca parameter that allows for control of whether or not a separate orted is used on the head node.  in non-cray slurm environments, it is default set to false so we keep current behavior.  on cray systems it is default set to true to handle the rdma cookie problem we know about.  If you think a different name would be better let me know.

ibm-ompi:

 Gist: https://gist.github.com/110f9ce2b92191f0becc3c6a68dd9d24

ibm-ompi:

 Gist: https://gist.github.com/28378baa2ae11866011964686b1025c6

jjhursey:
Note that `master` is broken so all CI is going to break on topo_treematch until it is fixed 😞

hppritcha:
thanks for the heads up @jjhursey

rhc54:
Thanks - looks good

jjhursey:
bot:retest

Issue #3993, opened 2017-08-01
3.0.0rc2:  Borked OPAL_CHECK_GCC_ATOMIC_BUILTINS definitions


hppritcha:
@PHHargrove reports seeing problems using xlc12.1 to build 3.0.0rc2
```
With xlc-12.1 I am seeing spurious failures of configure tests due to multiple definitions of OPAL_ASM_SYNC_HAVE_64BIT:

"conftest.c", line 211.9: 1506-236 (W) Macro name OPAL_ASM_SYNC_HAVE_64BIT has been redefined.
"conftest.c", line 211.9: 1506-358 (I) "OPAL_ASM_SYNC_HAVE_64BIT" is defined on line 209 of conftest.c.

Indeed the generated boilerplate in each conftest.c past a certain point contains two *different* definitions of this macro on lines 209 and 211:

 [...]
 #define OPAL_ASM_SYNC_HAVE_64BIT
 #define OPAL_HAVE_GCC_BUILTIN_CSWAP_INT128 0
 #define OPAL_ASM_SYNC_HAVE_64BIT 1
 [...]
```

reported on devel:
https://www.mail-archive.com/devel@lists.open-mpi.org//msg20323.html

PHHargrove:
It appears that both `OPAL_CHECK_GCC_ATOMIC_BUILTINS` and `OPAL_CHECK_SYNC_BUILTINS` both set `OPAL_ASM_SYNC_HAVE_64BIT`.  That alone may be erroneous.

However, I have tracked the empty (first) definition to the `OPAL_CHECK_GCC_ATOMIC_BUILTINS` macro which is consuming `opal_asm_sync_have_64bit` which has yet to be set:

```
AC_DEFUN([OPAL_CHECK_GCC_ATOMIC_BUILTINS], [
  AC_MSG_CHECKING([for __atomic builtin atomics])

  AC_TRY_LINK([
#include <stdint.h>
uint32_t tmp, old = 0;
uint64_t tmp64, old64 = 0;], [
__atomic_thread_fence(__ATOMIC_SEQ_CST);
__atomic_compare_exchange_n(&tmp, &old, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
__atomic_add_fetch(&tmp, 1, __ATOMIC_RELAXED);
__atomic_compare_exchange_n(&tmp64, &old64, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
[...]

PR #4000, opened 2017-08-01
config: remove erroneous define

hjelmn:
This removes a copy-and-paste error where we were setting the
OPAL_ASM_SYNC_HAVE_64BIT more than once.

References #3993. Close when on master and v3.0.x.


hjelmn:
@PHHargrove This will fix the configure error. Will open a PR to v3.0.x once merged on master.

PR #4003, opened 2017-08-01
opal: remove generated asm code

hjelmn:
Every modern compiler supports either inline assembly or builtin atomic
operations. Because of this it is time to delete all the code associated
with pre-built atomic.


hjelmn:
@jsquyres, @bosilca Any objections?

hjelmn:
@rhc54 I would suggest the same be done in pmix if possible.

ibm-ompi:

 Gist: https://gist.github.com/7da4761c43d92e5465b63bd9bd373acd

ibm-ompi:

 Gist: https://gist.github.com/969be1204291e2f52ba3bf43e45e559c

ibm-ompi:

 Gist: https://gist.github.com/0a2eaa712317fee9afe2511f93ce2111

jsquyres:
I've been out of the ASM for years.  If @bosilca @bwbarrett and @hjelmn come to consensus, 🎉

bwbarrett:
@hjelmn, have you tested with XLC?  I'm positive we don't care about Solaris CC, but XLC was also one of the holdouts to the proper inline assembly support camp...

ibm-ompi:

 Gist: https://gist.github.com/1d48a6dab089bb7515a3ab50febd74d5

bwbarrett:
Ok, there's two things in this patch.  First, it removes generating non-inline atomic functions, which we can't remove because at least PGI still doesn't support inline assembly (see the Gist above).  Second, it removes generating those files for common compilers / platforms during "make dist", which is the part that had the bug in PMIX.  That part we did to avoid a perl requirement in the common case, and can probably die.  But it might also be easier just to fix that bug.

hjelmn:
@bwbarrett PGI has supported inline assembly since 2010. Looks like I screwed up the check. Will update. The question is do we want to continue supporting compilers that do not have inline assembly or builtins? For 3.0.x I think we should just fix the check but for v3.1 I think it would be prudent to rip of the bandaid.

hjelmn:
Ok, should work with PGI. I also cleaned out the DEC and XLC assembly checks. They are not used anywhere in the code base so they did nothing.

ggouaillardet:
let me give this a try on linux/sparc and fujitsu compilers first.
iirc, fujitsu compilers are busted and generate incorrect code for atomics, so the workaround is to use the asm.
@kawashima-fj any thoughts on that ?

hjelmn:
@ggouaillardet Ok, let me know. I hope this isn't too ambitious for v3.1. The hope is to take this as the first step to C11. The compilers that do not have inline asm or builtins more than likely do not have C11 atomics. That said, we are not planning on trimming the compiler list for v3.1 :).

kawashima-fj:
@ggouaillardet The `__sync_bool_compare_and_swap` bug of the Fujitsu compiler was fixed and the fixed version was delivered to our customers. Inline assembly is supported. So I don't care about removing generated asm code in a future version.

ggouaillardet:
i just checked and Open MPI worked out of the box, so i have no more objections to this PR

bosilca:
This looks good. I have a comment on this topic, that I have also raised during the face-to-face meeting: C++ compilers are incompatible with C11 atomics. We might either have to remove all CXX support if we go for C11 atomics, or generate our own indirection layer.

bwbarrett:
IBM says that XLC should be good from 12.1 onward, so I'm good with this patch.

hjelmn:
Great! Thanks for taking a look @bosilca, @bwbarrett. I will merge this now and make a smaller patch to address the issue seen on v3.0.x. In that chase configure should just fail.

rhc54:
I gather that this is still something we need to port to PMIx, yes?

PR #4006, opened 2017-08-02
oshmem: shmem_ptr() implementation

alex-mikheev:
@yosefe @jladd-mlnx please take a look

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx.c
+++ b/oshmem/mca/spml/ucx/spml_ucx.c
@@ -64,6 +64,7 @@ mca_spml_ucx_t mca_spml_ucx = {
                                every spml */
         mca_spml_ucx_rmkey_unpack,
         mca_spml_ucx_rmkey_free,
+        mca_spml_ucx_rmkey_ptr,
         mca_spml_ucx_memuse_hook,
         (void*)&mca_spml_ucx
     },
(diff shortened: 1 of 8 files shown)


oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -353,6 +354,23 @@ void mca_spml_ucx_rmkey_free(sshmem_mkey_t *mkey)
     ucp_rkey_destroy(ucx_mkey->rkey);
 }
 
+void *mca_spml_ucx_rmkey_ptr(const void *dst_addr, sshmem_mkey_t *mkey, int pe)
+{
+#if ((UCP_API_MAJOR >= 1) && (UCP_API_MINOR >= 3))
```
yosefe: what about UCX 2.1?

oshmem/shmem/c/shmem_ptr.c
```diff
@@ -19,18 +19,53 @@
...
+    RUNTIME_CHECK_PE(pe);
+    RUNTIME_CHECK_ADDR(dst_addr);
+
+    /* process can access its own memory */
+    if (pe == oshmem_my_proc_id()) {
+        return (void *)dst_addr;
+    }
+
+    /* The memory must be on the local node */
+    proc = oshmem_proc_group_find(oshmem_group_all, pe);
```
yosefe: don't think we should restrict the access to "local nodes". if the spml has some magic to be able to access to remote nodes as well, need to use it.

jladd-mlnx:
👍

PR #3960, opened 2017-07-25
Topic/treematch

bosilca:
Update the treematch-based topo support, including a newer version of treematch.
Fixes #1183.

PR #4008, opened 2017-08-02
Revert "Topic/treematch"

bwbarrett:
Reverts open-mpi/ompi#3960


bwbarrett:
It looks like the Pull Request Build Checker failed because it doesn't like the branch name with a /.  Have to figure that out, but pushing the revert to unblock development.

PR #4016, opened 2017-08-03
opal_check_libnl: it's not a fatal error if libnl check fails

### LINKED ISSUE #3995: If libfabric is borked, configure unconditionally aborts
As an offshoot out of #3989, it was discovered that if configure determines that libfabric is borked (by linking in both libnl v1 and v3), then configure aborts.

Instead, it should obey `--with[out]-libfabric` CLI conventions (i.e., only abort if `--with-libfabric` was specified; otherwise, just skip libfabric support).

jsquyres:
Per https://github.com/open-mpi/ompi/issues/3995, it should not be a fatal error if the libnl checks fail.  Instead, just fail the check and let the upper layer decide what to do.  In this case, OPAL_CHECK_PACKAGE will mark this library as no good, and then propagate that upward.

E.g., if libfoo fails the libnl check, and the user had specified --with-libfoo, this will eventually cause configure to fail (because the libnl check will fail with libfoo, which will cause OPAL_CHECK_PACKAGE to fail with libfoo, which will ultimately cause some upper-level logic to realize "a human asked for libfoo but we could not provide it -- abort!").

However, if libfoo fails the libnl check and the user did *not* specify --with-libfoo, then this will cause the upper layer to silently skip libfoo (because the libnl check will fail libfoo, which will cause OPAL_CHECK_PACKAGE to fail libfoo, but then the upper-level logic will realize "oh, we can't use libfoo, but a human didn't ask for it -- so just skip libfoo support.").

Fixes #3995

@PHHargrove Can you test that this makes your machine skip libfabric without aborting configure?  It should emit a noisy warning in the middle of configure that your libfabric is borked.  A tarball you can test is here (WARNING TO FUTURE READERS: this link will almost certainly become stale after Paul tests with this tarball): http://jeff.squyres.com/unofficial/openmpi-gitclone-libnl-for-phargrove.tar.bz2

### DIFF
--- a/config/opal_check_libnl.m4
+++ b/config/opal_check_libnl.m4
@@ -2,6 +2,7 @@ dnl -*- shell-script -*-
 dnl
 dnl Copyright (c) 2015-2017 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
+dnl Copyright (c) 2017      Cisco Systems, Inc.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
@@ -18,7 +19,7 @@ dnl libnl-3.so.200 and friends, so if libnl3-devel packages are not
 dnl installed, but libnl-devel are, Open MPI should not try to use libnl.
 dnl
 dnl GROSS: libnl wants us to either use pkg-config (which we
-dnl can't assume is always present) or we need to look in a
+dnl cannot assume is always present) or we need to look in a
 dnl particular directory for the right libnl3 include files.  For
 dnl now, just hard code the special path into this logic.
 dnl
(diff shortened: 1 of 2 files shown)


PHHargrove:
@jsquyres 
I tested the tarball you provided, and see the desired behaviors with both with and without --enable-libfabric.

However, the warning does appear to report the same condition *twice* (and the second says "fabric" where I suspect "libfabric" was probably intended):
```
checking for library containing fi_getinfo... -lfabric
configure: WARNING: Unfortunately, libfabric links to both libnl and libnl-3.
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: libnl version conflict: fabric requires libnl whereas fabric requires libnl-3
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
```

And here is the fatal case when I pass --with-libfabric:
```
checking for library containing fi_getinfo... -lfabric
configure: WARNING: Unfortunately, libfabric links to both libnl and libnl-3.
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: libnl version conflict: fabric requires libnl whereas fabric requires libnl-3
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: OFI libfabric support requested (via --with-ofi or --with-libfabric), but not found.
configure: error: Cannot continue.
```

jsquyres:
@PHHargrove Ah!  Missed that.  I updated this PR and have a new tarball for you to try out -- should only see the message once this time (and I fixed the `fabric` -> `libfabric` in the 2nd message that you won't see 😉 ): http://jeff.squyres.com/unofficial/openmpi-gitclone-libnl-for-phargrove2.tar.bz2

(SAME NOTE TO FUTURE READERS: the above link will almost certainly become stale)

PHHargrove:
@jsquyres 
Sorry to report that the new tarball is a regression to the old behavior.
Configured with only --prefix=... and --enable-debug:
```
--- MCA component common:ofi (m4 configuration macro)
checking for MCA component common:ofi compile mode... dso
checking --with-ofi value... simple ok (unspecified)
checking --with-ofi-libdir value... simple ok (unspecified)
checking looking for OFI libfabric in... ()
checking rdma/fabric.h usability... yes
checking rdma/fabric.h presence... yes
checking for rdma/fabric.h... yes
looking for library without search path
checking for library containing fi_getinfo... -lfabric
configure: WARNING: Unfortunately, libfabric links to both libnl and libnl-3.
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
checking if MCA component common:ofi can compile... no
[...]
*** Final output
checking for libraries that use libnl v1... fabric
checking for libraries that use libnl v3... fabric
configure: WARNING: libnl v1 and libnl v3 have been found as dependent libraries
configure: error: This is a configuration that is known to cause run-time crashes
```
Configure terminates here w/ non-zero exit code

bwbarrett:
@jsquyres, I pulled off the target tags since this was for master, but were you going to do something with this PR?

bwbarrett:
@jsquyres, were you going to push this?

jsquyres:
@PHHargrove I finally got the time to poke into this, and figured out what the problem was.  I re-tooled the code a bit and have a new tarball for you: http://jeff.squyres.com/unofficial/openmpi-gitclone-libnl-for-phargrove.tar.bz2

PHHargrove:
@jsquyres I have a big deadline on 09/29.  Please remind me again after that if you've not heard from me.

PHHargrove:
Here is what I see w/ the latest with only -prefix and --enable-debug passed to configure.  
I see the expected *warning*:
```
--- MCA component common:ofi (m4 configuration macro)
checking for MCA component common:ofi compile mode... dso
checking --with-ofi value... simple ok (unspecified value)
checking --with-ofi-libdir value... simple ok (unspecified value)
checking looking for OFI libfabric in... ()
checking rdma/fabric.h usability... yes
checking rdma/fabric.h presence... yes
checking for rdma/fabric.h... yes
looking for library without search path
checking for library containing fi_getinfo... -lfabric
checking if libfabric requires libnl v1 or v3... v1 v3
configure: WARNING: Unfortunately, libfabric links to both libnl and libnl-3.
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: libnl version conflict: fabric requires libnl whereas libfabric requires libnl-3
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
checking if MCA component common:ofi can compile... no
```
The build completes successfully.

When I additionally pass --with-libfabric is see the expected *failure*.
```
--- MCA component common:ofi (m4 configuration macro)
checking for MCA component common:ofi compile mode... dso
checking --with-ofi value... simple ok (unspecified value)
checking --with-ofi-libdir value... simple ok (unspecified value)
checking looking for OFI libfabric in... (default search paths)
checking rdma/fabric.h usability... yes
checking rdma/fabric.h presence... yes
checking for rdma/fabric.h... yes
looking for library without search path
checking for library containing fi_getinfo... -lfabric
checking if libfabric requires libnl v1 or v3... v1 v3
configure: WARNING: Unfortunately, libfabric links to both libnl and libnl-3.
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: libnl version conflict: fabric requires libnl whereas libfabric requires libnl-3
configure: WARNING: This is a configuration that is *known* to cause run-time crashes.
configure: WARNING: This is an error in libfabric (not Open MPI).
configure: WARNING: Open MPI will therefore skip using libfabric.
configure: WARNING: OFI libfabric support requested (via --with-ofi or --with-libfabric), but not found.
configure: error: Cannot continue.
```

So, other than some repetition in the output, all looks good now.

I did notice libz being checked for which libnl is used - a bit silly but harmless.

jsquyres:
@PHHargrove Thank you for testing!

* Oops on the repetitiveness -- I fixed that.
* Yes, `libz` is checked for libnl Badness.  We've actually been checking *all* libraries for libnl conflicts because configure doesn't know what libraries pull in `libnl` -- so we check them all.  In this PR, I added some verbose checking statements to stdout to make it obvious that that is happening.

Issue #3685, opened 2017-06-09
Correctness failure when using BTL RDMA


vspetrov:
Thank you for taking the time to submit an issue!

## Background information
Possibly related to the "patcher" memory framework

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
v2.0.x
v2.x
master

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
Built from sources, cloned from github

### Please describe the system on which you are running

* Operating system/version: Red Hat Enterprise Linux Server release 7.2 (Maipo)
* Computer hardware: Intel(R) Xeon(R) CPU E5-2680 0 @ 2.70GHz
* Network type: Mellanox infiniband, CIB

-----------------------------

## Details of the problem
Multithreaded correctness test (attached 
[mt_stress.zip](https://github.com/open-mpi/ompi/files/1063859/mt_stress.zip)
) fails with OMPI. 
Reproduced on 2 nodes.
```shell
shell$ mpirun -np 3 --map-by node -mca pml ob1 -mca btl openib,self   `nif 50`    -mca coll ^hcoll ./mt_stress 1497004394

...
Splitting id 124
CORRECTNESS ERROR: id 124, TEST_TYPE 2, pos 3692, value 3, expected 6, dtype MPI_INT, root 0, rank 2, count 15045, comm_size 3, color 1
CORRECTNESS ERROR: id 124, TEST_TYPE 2, pos 3692, value 3, expected 6, dtype MPI_INT, root 0, rank 0, count 15045, comm_size 3, color 1
CORRECTNESS ERROR: id 124, TEST_TYPE 2, pos 3692, value 3, expected 6, dtype MPI_INT, root 0, rank 1, count 15045, comm_size 3, color 1
Splitting id 125
Splitting id 126
...
```
This is an allreduce failure. After some debug i narrowed it down to the single p2p inside allreduce. One ranks sends the data to the other side, but the data is received corrupted for some reason.

The test would pass if "-mca mpi_leave_pinned 0" OR if the ompi is built without memory manager support (--without-memory-manager). This is why my suspicion goes to "patcher" memory framework.

Additionally, the same issues are observed with pml yalla (mellanox mxm based p2p). Again disabling mem notifications (MXM_MEM_ON_DEMAND_MAP=n) helps.

[...]

PR #4026, opened 2017-08-04
Hook madvise when using patcher memory hooks

hjelmn:
It is not possible to use the patcher based memory hooks without
hooking madvise (MADV_DONTNEED). This commit updates the patcher
memory hooks to always hook madvise. This should be safe with recent
rcache updates.
    
References #3685. Close when merged into v2.0.x, v2.x, and v3.0.x.

### DIFF
--- a/opal/mca/rcache/base/rcache_base_vma_tree.c
+++ b/opal/mca/rcache/base/rcache_base_vma_tree.c
@@ -29,12 +29,14 @@
 
 #include "opal/util/output.h"
 #include "rcache_base_vma_tree.h"
+#include "opal/mca/rcache/base/base.h"
 
 OBJ_CLASS_INSTANCE(mca_rcache_base_vma_reg_list_item_t, opal_list_item_t, NULL, NULL);
 
 static void mca_rcache_base_vma_item_construct (mca_rcache_base_vma_item_t *vma_item)
 {
     OBJ_CONSTRUCT(&vma_item->reg_list, opal_list_t);
+    vma_item->in_use = false;
 }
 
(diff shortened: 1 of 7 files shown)


hjelmn:
:bot:retest:

opal/mca/rcache/base/rcache_base_vma_tree.c
```diff
@@ -131,6 +132,17 @@ mca_rcache_base_vma_item_t *mca_rcache_base_vma_new (mca_rcache_base_vma_module_
     return vma_item;
 }
 
+/* NTH: this function MUST not allocate or deallocate memory */
```
bosilca: How do we know thatOBJ_RELEASE is not putting the refcount to 0 (and free the memory) ?

hjelmn: I should make it clear. It is impossible (due to the lock) for the OBJ_RELEASE in that loop to be the last release on an object. I will add a comment. As for the documentation I did it through the commit message. I do no intend to let the VMA tree exist in its current form for much longer. It has serious threading performance issues.

hjelmn: Oh wait. Sorry about that. The comment was on a function that was deleted before the final version of the patch. Will eliminate the incorrect comment.

jsquyres: @hjelmn Does this mean you'll have a follow-on commit here on master, and then add that commit to #4041 and #4042?

hjelmn: @jsquyres Yes. Will open a PR to clear the erroneous comment now.

PR #4033, opened 2017-08-06
oshmem: sshmem sysv: auto huge page alloc can fallback to regular pages.

alex-mikheev:
@yosefe @jladd-mlnx
Fallback to the regular pages if huge page allocation is set to auto
and it was not possible to allocate requested amount of memory with
the hugepages.

### DIFF
--- a/oshmem/mca/sshmem/sysv/sshmem_sysv_module.c
+++ b/oshmem/mca/sshmem/sysv/sshmem_sysv_module.c
@@ -115,6 +115,7 @@ segment_create(map_segment_t *ds_buf,
     void *addr = NULL;
     int shmid = MAP_SEGMENT_SHM_INVALID;
     int flags;
+    int try_hp;
 
     assert(ds_buf);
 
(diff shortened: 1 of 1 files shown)


oshmem/mca/sshmem/sysv/sshmem_sysv_module.c
```diff
@@ -129,14 +130,28 @@ segment_create(map_segment_t *ds_buf,
...
 
     /* Create a new shared memory segment and save the shmid. */
+retry_alloc:
     shmid = shmget(IPC_PRIVATE, size, flags);
     if (shmid == MAP_SEGMENT_SHM_INVALID) {
+        /* hugepage alloc was set to auto. Hopefully it failed because there are no
+         * enough hugepages on the system. Turn it off and retry.
+         */
+        if (-1 == try_hp) {
+            OPAL_OUTPUT_VERBOSE(
```
miked-mellanox: maybe should be warn as taking fallback may have perf implications?

alex-mikheev: It is a normal behavior of this feature so imho no warning is needed. There are similar fallback schemes in both mxm and ucx and they also  do not warn.

miked-mellanox:
+1

PR #4037, opened 2017-08-07
instructions: Update to reflect autogen.pl usage

HaoZeke:
It's a bit of a daft change but it'd help noobs...

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -47,13 +47,15 @@ performance penalty at run-time because of extra debugging overhead.
 User Builds
 ===========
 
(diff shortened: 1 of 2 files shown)


rhc54:
Ummm...those comments are actually incorrect. The INSTALL and README files are intended for typical users who are building from release tarballs. The HACKING file is for those building from GitHub clones. It is only that last case that needs to run autogen.pl, and we mention it there.

What we could do, I suppose, is add a comment pointing those building from GitHub clones to the HACKING file as it provides more detailed instructions. I can do that.

Appreciate the suggestion!

PR #4043, opened 2017-08-07
Fix libpmix linking

rhc54:
When we specify --with-devel-headers, we also emit a copy of libpmix. However, that library was built against the OPAL libevent component, which means all the libevent functions are prefixed with OPAL names. So ensure that the emitted libpmix is linked back against libopen-pal so those symbols will be resolved.

Also, remove the stale --enable-pmix-dstore flag as PMIx v2.1 does not recognize it since the dstore functionality is now in a framework and can be disabled at runtime.


ibm-ompi:

PR #4046, opened 2017-08-07
Fix linking of libpmix when emitted

### LINKED ISSUE #4045: When emitted, libpmix is not linked against libopen-pal
We emit a copy of libpmix based on the internal PMIx code tree when we are configured --with-devel-headers. This is done so that applications can directly reference PMIx calls and ensure they are compatible with the included ORTE support.

However, libpmix in this case is referencing OPAL-prefixed names for libevent and other functions, so we need to link it back to libopen-pal to resolve them.

rhc54:
When we specify --with-devel-headers, we also emit a copy of libpmix. However, that library was built against the OPAL libevent component, which means all the libevent functions are prefixed with OPAL names. So ensure that the emitted libpmix is linked back against libopen-pal so those symbols will be resolved.

Fixes #4045

(cherry picked from commit d593e5a4cee4684e74ecfd758781b77a84caf593)

### DIFF
--- a/opal/mca/pmix/pmix2x/configure.m4
+++ b/opal/mca/pmix/pmix2x/configure.m4
@@ -56,7 +56,7 @@ AC_DEFUN([MCA_opal_pmix_pmix2x_CONFIG],[
           [opal_pmix_pmix2x_args="--disable-debug $opal_pmix_pmix2x_args"
            CFLAGS="$OPAL_CFLAGS_BEFORE_PICKY $OPAL_VISIBILITY_CFLAGS"])
     AS_IF([test "$with_devel_headers" = "yes"],
-          [AS_IF([test "$enable_dlopen" = "yes"],
+          [AS_IF([test "$enable_dlopen" != "no"],
                  [opal_pmix_pmix2x_args="--with-pmix-extra-lib=$OPAL_TOP_BUILDDIR/opal/lib${OPAL_LIB_PREFIX}open-pal.la $opal_pmix_pmix2x_args"])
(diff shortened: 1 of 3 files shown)


rhc54:
@jsquyres I'm not sure this fix is correct - at least, it is breaking --disable-dlopen builds on master. I'm working on a variant of it for there.

rhc54:
@jsquyres I think I have this correct now - please let me know what you think

opal/mca/pmix/pmix2x/configure.m4
```diff
@@ -56,7 +56,9 @@ AC_DEFUN([MCA_opal_pmix_pmix2x_CONFIG],[
           [opal_pmix_pmix2x_args="--disable-debug $opal_pmix_pmix2x_args"
            CFLAGS="$OPAL_CFLAGS_BEFORE_PICKY $OPAL_VISIBILITY_CFLAGS"])
     AS_IF([test "$with_devel_headers" = "yes"],
-          [opal_pmix_pmix2x_args="--with-devel-headers --with-pmix-extra-lib=$OPAL_TOP_BUILDDIR/opal/lib${OPAL_LIB_PREFIX}open-pal.la $opal_pmix_pmix2x_args"])
+          [AS_IF([test "$enable_dlopen" = "yes"],
```
jsquyres: Should this be `"$enable_dlopen" != "no"` to cover the case when `$enable_dlopen` is blank?

rhc54:
@jsquyres Done, as requested

rhc54:
cray test node failed to respond

bot:ompi:retest

PR #4056, opened 2017-08-08
mca/registry: fix problem group_component_register

### LINKED ISSUE #4020: `ompi_info --param` does not work for vader
```
                Open MPI: 2.1.1
  Open MPI repo revision: v2.1.0-100-ga2fdb5b
   Open MPI release date: May 10, 2017
```

### Describe how Open MPI was installed
Open MPI was installed from source:
```
wget https://www.open-mpi.org/software/ompi/v2.1/downloads/openmpi-2.1.1.tar.bz2
tar xjf openmpi-2.1.1.tar.bz2
cd openmpi-2.1.1
./configure --prefix=$HOME/local/openmpi/2.1.1
make -j 4
make install
```

### Describe the system on which you are running
* Operating system/version:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
```

* Computer hardware: Lenovo ThinkPad T430 laptop
```
$ uname -a
Linux t430 3.19.0-80-generic #88~14.04.1-Ubuntu SMP Fri Jan 13 14:54:07 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
-----------------------------

## Details of the problem
`ompi_info` does not list parameters for the vader layer when vader is requested explicitely.
However, it does list vader parameters when `btl all` or `all all` are requested.
I also observe this behaviour on RHEL 7.3 ppc64le.

```
$ ompi_info --param btl vader --level 9
                 MCA btl: vader (MCA v2.1.0, API v3.0.0, Component v2.1.1)
```

```
$ ompi_info --param all all --level 9 | grep vader
                 MCA btl: vader (MCA v2.1.0, API v3.0.0, Component v2.1.1)
           MCA btl vader: ---------------------------------------------------
           MCA btl vader: parameter "btl_vader_free_list_num" (current value: "8", data source: default, level: 9 dev/all, type: int)
           MCA btl vader: parameter "btl_vader_free_list_max" (current value: "4096", data source: default, level: 9 dev/all, type: int)
           MCA btl vader: parameter "btl_vader_free_list_inc" (current value: "64", data source: default, level: 9 dev/all, type: int)
...
           MCA btl vader: parameter "btl_vader_bandwidth" (current value: "40000", data source: default, level: 5 tuner/detail, type: unsigned_int)
```

hppritcha:
Turns out that supplying NULL to group_register in the
mca_base_var_group_component_register is not a good
idea if one wants for ompi_info to work as intended.

The ugni and vader btl's both call this before
registering component variables.  This borks up
the ompi_info works since NULL is supplied as the project
name.  So, now supply the project name rather than
just NULL to group register.

Fixes #4020.

### DIFF
--- a/opal/mca/base/mca_base_var_group.c
+++ b/opal/mca/base/mca_base_var_group.c
@@ -321,8 +321,7 @@ int mca_base_var_group_register (const char *project_name, const char *framework
 int mca_base_var_group_component_register (const mca_base_component_t *component,
                                            const char *description)
 {
-    /* 1.7 components do not store the project */
-    return group_register (NULL, component->mca_type_name,
+    return group_register (component->mca_project_name, component->mca_type_name,
                            component->mca_component_name, description);
 }
 


hjelmn:
Opps. Missed that one.

PR #4059, opened 2017-08-09
NEWS: update news for 2.1.2 release

hppritcha:
Doing the old-fashioned thing with NEWS.
[skip ci]

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -53,13 +53,50 @@ included in the vX.Y.Z section and be denoted as:
    (** also appeared: A.B.C)  -- indicating that this item was previously
                                  included in release version vA.B.C.
 
-2.1.2 -- August, 2017
---------------------
+2.1.2 -- September, 2017
+------------------------
 
 Bug fixes/minor improvements:
-- Remove IB XRC support from the OpenIB BTL due to lack of support.
+
+- Update internal PMIx version to 1.2.3.
+- Fix some problems when using the NAG Fortran compiler to build Open MPI
+  and when using the compiler wrappers.  Thanks to Neil Carlson for reporting.
+- Fix a compilation problem with the SM BTL.  Thanks to Paul Hargrove for
+  reporting.
+- Fix a problem with MPI_IALLTOALLW when using zero-length messages.
+  Thanks to Dahai Guo for reporting.
+- Fix a problem with C11 generic type interface for SHMEM_G.  Thanks
+  to Nick Park for reporting.
+- Switch to using the lustreapi.h include file when building Open MPI
+  with Lustre support.
+- Fix a problem in the OB1 PML that led to hangs with OSU collective tests.
+- Fix a progression issue with MPI_WIN_FLUSH_LOCAL.  Thanks to
+  Joseph Schuchart for reporting.
+- Fix an issue with recent versions of PBSPro requiring libcrypto.
+  Thanks to Petr Hanousek for reporting.
+- Fix a problem when using MPI_ANY_SOURCE with MPI_SENDRECV.
+- Fix an issue that prevented signals from being propagated to ORTE
+  daemons.
+- Ensure that signals are forwarded from ORTE daemons to all processes
+  in the process group created by the daemons.  Thanks to Ted Sussman
+  for reporting.
+- Fix a problem with launching a job under a debugger. Thanks to
+  Greg Lee for reporting.
+- Fix a problem with Open MPI native I/O MPI_FILE_OPEN when using
+  a communicator having an associated topology.  Thanks to
+  Wei-keng Liao for reporting.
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -57,7 +57,39 @@ included in the vX.Y.Z section and be denoted as:
 --------------------
 
 Bug fixes/minor improvements:
+
 - Remove IB XRC support from the OpenIB BTL due to lack of support.
+- Fix a problem with MPI_IALLTOALLW when using zero-length messages.
```
jsquyres: For the XRC bullet, it says "support" twice -- perhaps it can be:

> Remove IB XRC functionality from the OpenIB BTL due to lack of support.

...the more I think about this, though,  are we breaking backwards compatibility?  I.e., is the broken XRC functionality a regression since v2.0.0?  Or has it been 100% broken since v2.0.0?  (i.e., it can't break backwards compatibility because there's no possible chance that anyone has been using it because it has been completely broken)

hppritcha: I'll fix that wording.

@artpol84 could you test 2.0.0 release to see if MLNX sees the XRC issue there as well?

hppritcha: @jsquyres let's clarify.  its not completely broken.  it works for me on my connectx5 MOFED 4.0.0 cluster.  It just doesn't work for Mellanox.

hppritcha:
@artpol84 could you test whether XRC support is broken on your cluster for the 2.0.0 release?

jsquyres:
@hppritcha Careful about your clarification, because it's now not possible to enable XRC (via `configure`).

jsquyres:
@artpol84 Ping 😄

artpol84:
@jsquyres can you explain what is needed? Do you want me to enable XRC for v2.0.0 in Jenkins?
If so - what is the right way to configure as you pointed in https://github.com/open-mpi/ompi/pull/4059#issuecomment-321338933.

artpol84:
@jsquyres @hppritcha I must admit that I can't reproduce manually as well.
I tried 2.0.0 and 2.1.1 tarballs - all OK, then  I checkout v2.x and revert c22a7c7e92b729b2dd762a59e8cf7fc4a93eb1ca - and still all was working OK.
I'm going to enable XRC in jenkins and open a test PR with XRC commit reverted to see if this is still the case on v2.x.

hppritcha:
@jsquyres i removed the XRC support removed sentence since it looks like @artpol84 is busy reverting.

hppritcha:
okay, put the sentence back in.  @artpol84 we're not taking a revert of XRC disable for 2.1.2 release.

artpol84:
@hppritcha it were testing PRs and at some point it seemed that XRC is ok, but more detailed analysis done by @vspetrov (https://github.com/open-mpi/ompi/pull/4082#issuecomment-322709334) indicated that the issue can be reproduced without `hcoll`. So I take it back about reverting, XRC has problems.

jsquyres:
Summary from discussion on Webex 22 Aug 2017: 

* `hcoll` was a red herring.  The issue #4087 is a good summary of that -- the issue is still somewhere in the `openib` BTL.
* @artpol84 tested manually, and was only able to reproduce on the v2.x branch (not v2.0.x and not v3.0.x).  To be clear, we don't know exactly what the issue is yet, so the fact that he wasn't able to reproduce on the other branches does not mean that the problem does not exist on the other branches.

Meaning: 

* There is definitely a problem on v2.x.  
* There is not enough proof that the problem does *not* occur on other branches.
* We still need to decide what to do (e.g., in terms of disabling XRC or fixing it).

hppritcha:
@jsquyres I plan to rebase and merge this PR in before generating rc3 today.

jsquyres:
Ok.  Summary of XRC discussion yesterday:

* We decided that no one seems to care about XRC in the openib BTL at present.
* Decision: we will leave it disabled.  If someone wants to go in and fix it, they can re-enable it and then fix it.
* This affects master, v2.x, and v3.0.x.

jsquyres:
Were there any additional items that needed to be added to NEWS since you created this PR 3 weeks ago?

hppritcha:
@jsquyres  looks like there was a fix for SM BTL/CMA - #4123.  Added that as well as rearranging removing items to be less depressingly located.

NEWS
```diff
@@ -57,9 +57,44 @@ included in the vX.Y.Z section and be denoted as:
 --------------------
 
 Bug fixes/minor improvements:
-- Remove IB XRC support from the OpenIB BTL due to lack of support.
+
+- Update internal PMIx version to 1.2.3
```
jsquyres: Super minor/nit-picky: this bullet has no period (all other bullets have periods).

hppritcha: i'll fix that.  Also added a blurb about NAG (ha ha).

PR #4065, opened 2017-08-09
configure: check for C11 and atomic types

hjelmn:
This commit updates the configure code for Open MPI to check for C11
support. The features requested are: atomics and thread local
storage.

### DIFF
--- a/config/opal_setup_cc.m4
+++ b/config/opal_setup_cc.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2006 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2007-2009 Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2012      Los Alamos National Security, LLC. All rights
+dnl Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
 dnl                         reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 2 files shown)


ibm-ompi:

 Gist: https://gist.github.com/d51f6de012152920af28750bfc54a830

hjelmn:
Hmm, not quite right. IBM CI is using C90 mode.

ibm-ompi:

 Gist: https://gist.github.com/65744c3ad06861538c0f0220c2807ae6

hppritcha:
hmmm, looks like a file system is "full" on one or more of the AWS VMs.  Let's try our luck again.
bot:ompi:retest

hjelmn:
Ok, that should do it. For some reason the C11 tests pass without a flag with gcc 4.9.4 even though the default compiler mode is gnu90. Added a check to ensure C99 is supported at a minimum before assuming no flag is needed.

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,122 @@ dnl
...
+
+                 if test "x$opal_cv_c11_noflag" = "xno" ; then
+                     flags="-std=gnu11 -std=c11 -c11"
+
+
+                     AC_MSG_NOTICE([checking if $CC supports C11 with a flag])
+                     opal_cv_c11_flag=
+                     for flag in $(echo $flags | tr ' ' '\n') ; do
+                         OPAL_PROG_CC_C11_HELPER([$flag],[opal_cv_c11_flag=$flag],[])
+                         if test ! "x$opal_cv_c11_flag" = x ; then
```
jsquyres: `test "x$foo" != "x"` works -- I think that's the style we use most elsewhere.

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,122 @@ dnl
...
+                             CFLAGS="$CFLAGS $opal_cv_c11_flag"
+                             AC_MSG_NOTICE([using $flag to enable C11 support])
+                             opal_cv_c11_supported=yes
+                             break;
```
jsquyres: Extra semi colon.

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,122 @@ dnl
...
+                         fi
+                     done
+
+                     if ! test -z "$opal_cv_c11_flag" ; then
```
jsquyres: You used the `x` test notation above, but used `-z` here.  Please be consistent.

configure.ac
```diff
@@ -427,6 +427,11 @@ AC_CHECK_SIZEOF(wchar_t)
 
 AC_CHECK_SIZEOF(pid_t)
 
+# Check sizes of atomic types so we can define fixed-width types in OPAL
+AC_CHECK_SIZEOF(atomic_short, [],[[#include <stdatomic.h>]])
+AC_CHECK_SIZEOF(atomic_int,[],[[#include <stdatomic.h>]])
+AC_CHECK_SIZEOF(atomic_long,[],[[#include <stdatomic.h>]])
+AC_CHECK_SIZEOF(atomic_llong,[],[[#include <stdatomic.h>]])
```
jsquyres: I don't see the addition of any C code that uses the resulting `#define`s.   What are these for?

configure.ac
```diff
@@ -833,7 +838,7 @@ OPAL_SEARCH_LIBS_CORE([ceil], [m])
 # -lrt might be needed for clock_gettime
 OPAL_SEARCH_LIBS_CORE([clock_gettime], [rt])
 
-AC_CHECK_FUNCS([asprintf snprintf vasprintf vsnprintf openpty isatty getpwuid fork waitpid execve pipe ptsname setsid mmap tcgetpgrp posix_memalign strsignal sysconf syslog vsyslog regcmp regexec regfree _NSGetEnviron socketpair strncpy_s usleep mkfifo dbopen dbm_open statfs statvfs setpgid setenv __malloc_initialize_hook])
+AC_CHECK_FUNCS([asprintf snprintf vasprintf vsnprintf openpty isatty getpwuid fork waitpid execve pipe ptsname setsid mmap tcgetpgrp posix_memalign aligned_alloc strsignal sysconf syslog vsyslog regcmp regexec regfree _NSGetEnviron socketpair strncpy_s usleep mkfifo dbopen dbm_open statfs statvfs setpgid setenv __malloc_initialize_hook])
```
jsquyres: Similar to above, I don't see any new C code that checks for `HAVE_aligned_alloc`.  What is this for?

hjelmn: aligned_alloc will be used instead of posix_memalign. That code is not in place yet.

configure.ac
hjelmn: The code that uses these defines will be added soon. The problem is that C11 didn't define fixed-width atomic type (atomic_uint32_t, etc) so we will have to define them ourselves. The problem is that Intel only implemented half of C11 atomics. They do not support _Atomic. This means we have to define atomic_uint32_t, etc terms of the atomic_int, atomic_long, etc.

config/opal_setup_cc.m4
hjelmn: Will clean up the inconsistencies now.

configure.ac
hjelmn: If as part of requiring C11 we also require a modern glibc on Linux (2.12+) we can eliminate memalign/posix_memalign entirely. That will have to be discussed later.

hjelmn:
@jsquyres Should be better now.

configure.ac
hjelmn: Adding a description of the new standard call: http://en.cppreference.com/w/c/memory/aligned_alloc

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,115 @@ dnl
...
+        if test $opal_cv_c11_supported = no ; then
+            AC_MSG_WARN([Open MPI recommends a C11 compiler])
+            # uncomment the following line and update the warning when
+            # a C11 compiler is required
+            # AC_MSG_ERROR([Aborting.])
+            # From Open MPI 1.7 on we require a C99 compiant compiler
+            AC_PROG_CC_C99
+            # The result of AC_PROG_CC_C99 is stored in ac_cv_prog_cc_c99
+            if test "x$ac_cv_prog_cc_c99" = xno ; then
+                AC_MSG_WARN([Open MPI requires a C99 compiler])
```
jsquyres: This warning message is now no longer strictly accurate -- perhaps it should become:

> Open MPI requires a C compiler that supports C99 or later

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,115 @@ dnl
...
+#endif
+                                           ]],[])],
+                          [opal_cv_c11_flag_required=no])
+
+        AC_MSG_RESULT([$opal_cv_c11_flag_required])
+
+        if test "x$opal_cv_c11_flag_required" = "xno" ; then
+            AC_MSG_NOTICE([verifying $CC supports C11 without a flag])
+            OPAL_PROG_CC_C11_HELPER([], [], [opal_cv_c11_flag_required=yes])
+        fi
```
jsquyres: Any reason not to move the 2 addition lines from 112/113 up into this block (where no flag is required), and then make the next block be an `else`?

hjelmn:
This support is already in master.

PR #4066, opened 2017-08-09
Provide the mapping, ranking, binding patterns

rhc54:
Apps might want to make use of the relative patterns used to place/assign their procs

### DIFF
--- a/orte/orted/pmix/pmix_server_register_fns.c
+++ b/orte/orted/pmix/pmix_server_register_fns.c
@@ -45,7 +45,7 @@
 #include "orte/runtime/orte_globals.h"
 #include "orte/runtime/orte_wait.h"
 #include "orte/mca/errmgr/errmgr.h"
-#include "orte/mca/rmaps/rmaps_types.h"
+#include "orte/mca/rmaps/base/base.h"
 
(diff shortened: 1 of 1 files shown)


rhc54:
@edgargabriel This provides the info you wanted. You can retrieve it by something like the following:

```
opal_proc_t wildcard = {myjobid, OPAL_VPID_WILDCARD};
char *str=NULL;
int rc;

OPAL_MODEX_RECV_VALUE_OPTIONAL(rc, OPAL_PMIX_MAPBY, &wildcard, &str, OPAL_STRING);
```
The string is being generated by the ```orte_rmaps_base_print_mapping``` function down in orte/mca/rmaps/base/rmaps_base_print_fns.c.

I have also included similar values for OPAL_PMIX_RANKBY and OPAL_PMIX_BINDTO, with the latter being printed by the ```opal_hwloc_base_print_binding``` function in opal/mca/hwloc/base/hwloc_base_util.c.

The "optional" just means that we will return the info if it is available. This will protect you against earlier versions of OMPI where we don't provide it. So you should check the rc value for success before using the returned value. Of course, you should also check the string for NULL just for safety.

PR #3617, opened 2017-05-31
fortran2008: fix mpiext example

ggouaillardet:
in order to solve an egg and the chicken problem,  in which mpiext need mpi-f08-types.mod
and/but use-mpi-f08[-desc] needs mpiext, add an extra step
- build fortran 2008 modules only
- build fortran 2008 mpi extensions
- and then build fortran 2008 bindings

Fixes open-mpi/ompi#3605

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>

PR #4068, opened 2017-08-10
ompi/fortran: remove proof-of-concept mpi_f08 module

jsquyres:
This module was always intended to be a proof of concept, and was far from complete.  If/when someone implemented F08 descriptor support for the mpi_f08 module, this commit can either be restored or used as reference material.

@ggouaillardet per #3617, this PR removes the proof of concept mpi_f08 descriptor module.


ggouaillardet:
@jsquyres would you mind adding open-mpi/ompi@344f000841b53e0aca4dce841bf9dc654cd0c761 in a first commit, and then remove everything in a second one ?
that will make things easier if/when this module is restored.

jsquyres:
@ggouaillardet Ah -- 344f000841b53e0aca4dce841bf9dc654cd0c761 is on PR #3617.  You'd like to get that fix in so that at least the mpi_f08 descriptor stuff is fully correct before we remove everything (thereby potentially making it easier to restore someday).  Got it.

I propose an alternative: we merge #3617 and then I update this PR to cover the removal of everything.  Same net effect -- but it keeps the fix and the removal as distinct PRs.

jsquyres:
@ggouaillardet PR updated.

PR #4070, opened 2017-08-10
v3.0.x: remove proof of concept use mpi f08 descriptor module

jsquyres:
There's 4 commits here -- the last one is the one that matters.  The 3 leading up to it are just to make the cherry-pick of the 4th commit (i.e., the one that matters) trivial.

v3.0.x RMs: do you want this?  It's removal of (effectively) dead code.

Could really be done at any time; I just figured I'd PR it to v3.0.x and let you decide the timing.


bwbarrett:
@ggouaillardet, can you review?

PR #4079, opened 2017-08-15
rtc/hwloc: fix usage under valgrind

ggouaillardet:
instead of lookin for a hole in the VM and ask hwloc to mmap into it,
simply mmap() followed by munmap() in order to "create" some mmap'able space.

otherwise, valgrind is unable to mmap to the requested address

### DIFF
--- a/orte/mca/rtc/hwloc/rtc_hwloc.c
+++ b/orte/mca/rtc/hwloc/rtc_hwloc.c
@@ -1,6 +1,8 @@
 /*
  * Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2017      Cisco Systems, Inc.  All rights reserved
+ * Copyright (c) 2017      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -113,6 +115,13 @@ static int init(void)
         /* we couldn't find a hole, so don't use the shmem support */
         return ORTE_SUCCESS;
     }
+    shmemaddr = mmap(NULL, shmemsize, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+    if (MAP_FAILED == shmemaddr) {
+        opal_output(0, "%s could not mmap %ld bytes for shmem",
+                    ORTE_NAME_PRINT(ORTE_PROC_MY_NAME), shmemsize);
+        return ORTE_ERROR;
+    }
+
     /* create the shmem file in our session dir so it
      * will automatically get cleaned up */
     asprintf(&shmemfile, "%s/hwloc.sm", orte_process_info.jobfam_session_dir);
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@rhc54 @bgoglin could you please review this ?
with the latest master, an error occurs in ```rtc/hwloc``` if i run ``mpirun``` under ```valgrind``` (no such error if not under valgrind)

```
$ valgrind mpirun -np 1 ./hello_c
[c7.kmc.kobe.rist.or.jp:24101] WRITE FAILED -5
[c7.kmc.kobe.rist.or.jp:24101] [[4781,0],0] ORTE_ERROR_LOG: Error in file ../../../../../../src/ompi-master/orte/mca/rtc/hwloc/rtc_hwloc.c at line 155
```
(note there are a bunch of __non__ related valgrind errors, and these are fixed in pmix/pmix#471)

currently, we look for a hole in the VM, and ask ```hwloc``` to ```mmap``` to it (and for some reason, ```mmap``` fails)
at first glance, this looks over complicated to me, and i was unable to find a rationale for doing so.

so i simplified things and ended up writing this PR.

an other option would be to reorder things, and allow ```hwloc_shmem_topology_write``` being invoked with ```mmap_address=NULL```

ibm-ompi:

 Gist: https://gist.github.com/0bec490f25e4b751ce88450cb749f0ae

bgoglin:
I don't understand anything in this PR.

ggouaillardet:
ok, i will revamp it to make it clearer.

ggouaillardet:
i updated the PR.
the real thing happens in the first commit open-mpi/ompi@8a4179577b8eb9757fbafe990bb826858b20a888
(the second one remove some dead code and fix a type)

bgoglin:
We need the *exact* same mmap address in all processes, this master and all clients, not just this master (see hwloc/shmem.h in OMPI master for more details). There's no easy way to find one, so we assume all processes have a similar address space layout, and all of them will have available room in the biggest hole of the master.
If mmap fails in hwloc_shmem_topology_write() (or in adopt()), we just disable this shmem code and fallback to the normal non-shmem case. This may occur for processes whose virtual address space is pretty much full, but that's very unlikely on 64bits architectures since we have much more virtual addr space than physical addr space.

ggouaillardet:
i see, that makes this PR moot then ...

so that means that ```mpirun/orted``` looks for a "good" address (via ```find_hole```), and then we hope that the same address will be available in ```a.out```, is this correct ?

fwiw, one way to test this is by using the ```topo/treematch``` module, but it is unavailable since at ```init_query``` time, ```opal_hwloc_topology``` is ```NULL```

bgoglin:
Yes, that's correct.

I don't see how treematch could help there, but I'll have treematch guys in my office when back to work tomorrow so I'll ask them :)

ggouaillardet:
i was just saying ```topo/treematch``` could be used to check the hwloc shared memory is both used and working.

rhc54:
Given that we agree that this PR is moot, I'm going ahead and closing it

Issue #3879, opened 2017-07-13
Update Open MPI to require a C11 compiler


hjelmn:
We are discussing updating the compiler requirement for Open MPI to require C11. The features we want to require are:

 - C11 atomics. These will completely replace the hand-coded assembly in Open MPI.

 - Thread local support. This feature will completely replace the thread local storage code provided by opal.

 - aligned_alloc. Right now we use a memalign, posix_memalign, or malloc + hand aligning. This will bump the minimum glibc up to 2.16. If 2.16 is too new we can wrap this function in opal/util.

Like to haves:

 - Static assert

 - Generics

Compiler status pages:

 - Intel: https://software.intel.com/en-us/articles/c11-support-in-intel-c-compiler . Intel is missing _Atomic even in the latest versions. We will have to use atomic_* to support this compiler.

 - GCC: https://gcc.gnu.org/wiki/C11Status

 - Clang: v3.1+ started adding support in v3.1. C11 support was enabled by default in 2014.

 - PGI: Supported since at least late 2014. Will have to verify older release.

hjelmn:
One comment. LANL tends to support some old compiler versions to support fortran. This can be handled by using a supported compiler and using the older fortran compiler for the bindings. Example: ```./configure CC=gcc-6 FC=ifort```.

hjelmn:
One question is whether we should continue to offer ```opal_atomic_*``` functions or just go straight to using ```atomic_*``` everywhere. There are benefits to both but the later is a little more work and is probably cleaner in the long run.

hjelmn:
Basic configury is complete. Will push the branch to my repository shortly.

```
configure: checking if gcc-7 supports C11 without a flag
checking if gcc-7 supports C11 thread local storage... yes
checking if gcc-7 supports C11 atomic variables... yes
checking if gcc-7 supports C11 _Atomic keyword... yes
checking if gcc-7 supports C11 _Generic keyword... yes
configure: no flag required for C11 support
```

hjelmn:
We may want to keep some of the inline assembly going forward. It looks like gcc v7 removed support for lock-free compare-and-set for 128-bit integers on x86-64. This was done because the load/store of these integers is not atomic unless CAS is used. We do not care about this in the lifo/fifo implementations. A similar problem affects the LL/SC fifo/lifo implementations. There is no C11 atomic support for these atomics.

[...]

PR #4088, opened 2017-08-15
configure: add an option to set wrapper CC

hjelmn:
This commit adds a new configure option: --with-wrapper-cc. This
option can be used to set which C compiler is invoked by mpicc. The
feature is necessary when building for compiler suites that do not
support C11 (Ex: icc). A version of Open MPI that is set up for these
compilers can be built with:

./configure --with-wrapper-cc=icc CC=gcc FC=ifort CXX=icpc

### DIFF
--- a/config/opal_setup_cc.m4
+++ b/config/opal_setup_cc.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2006 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2007-2009 Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2012      Los Alamos National Security, LLC. All rights
+dnl Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
 dnl                         reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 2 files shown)


hjelmn:
@jsquyres Thoughts? I think this is a useful configure option even if we do not chose to require C11. Should I add similar options for mpicxx and mpifort for completeness?

rhc54:
Urgggg....ick. I'm not much of a company man, as you know, but I think I can hear the howl from HQ from here.

hjelmn:
@rhc54 Yeah, needing this would be unfortunate. Intel really dropped the ball on C11. Both clang and gcc are years ahead of them.

bwbarrett:
@rhc54, I think your comments were more about Nathan's comments in #3879 than in this change, right?

rhc54:
@bwbarrett I was actually commenting about both, I guess - it was this configure option thing that caught my immediate eye as it singles out the compilers from one vendor. Won't sit well, and I concur with your statement over there about being careful about what is needed vs what it might cost us.

hjelmn:
@rhc54 I can remove the reference to icc from the commit message. Intel is just the one that motivated this particular PR :).

rhc54:
I don't think that really resolves the problem. As was said on the other issue, this just creates more ways things can go wrong, and actually advocates cross-compiler builds (which we generally have said is "bad").

hjelmn:
This is generally how libraries are compiled. Usually they are built with gcc, clang, etc. The application itself is compiled with pgi, intel, gcc, etc. MPI is slightly different in that we provide wrappers for the users and we have fortran. I don't think an Open MPI compiled with gcc and linked against an application built with icc is a novel or dangerous idea.

hjelmn:
We probably should discuss this in a future call. Can't be next week but the week after should work fine.

hppritcha:
@hjelmn please rebase

jsquyres:
@hjelmn Do you still care about this PR?  If so, it needs to be rebased.

hjelmn:
Yes we do. If we tighten compiler requirements we will need this. Will rebase it today.

awlauria:
Ping. Rebase is needed still if still desired.

lanl-ompi:

jsquyres:
This PR is pretty stale, but could be fairly easily fixed if the functionality is actually still desired.  I moved it to "draft" state.

rhc54:
@hjelmn You have several PRs that date back 3-6 years - would it make sense for you to triage them and close the ones not worth rebasing, fixing, resubmitting for review, and finally committing?

hppritcha:
I think this PR can be closed as never to merged. @jsquyres

jsquyres:
Yes, let's close.  Can be re-opened / re-created if someone wants this functionality.

PR #4090, opened 2017-08-15

### LINKED ISSUE #4053: Add XL compiler version check to configure
Add configure check for XL compiler version support. As indicated by various bug reports, we don't really support older XL compiler versions (<=12.1) so we should error out with an informative message.
  * https://github.com/open-mpi/ompi/issues/3816
  * https://github.com/open-mpi/ompi/issues/3817 (somewhat related)
  * https://github.com/open-mpi/ompi/issues/3811 (somewhat related)

It might also be useful to check for `ppc64` vs `ppc64le` and emit a message.

jjhursey:
 * Fixes #4053

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,7 +14,7 @@ Copyright (c) 2006      Sun Microsystems, Inc.  All rights reserved.
                         Use is subject to license terms.
 Copyright (c) 2006-2017 Los Alamos National Security, LLC.  All rights
                         reserved.
-Copyright (c) 2010-2012 IBM Corporation.  All rights reserved.
+Copyright (c) 2010-2017 IBM Corporation.  All rights reserved.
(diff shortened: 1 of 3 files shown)


jjhursey:
Mellanox failure is not related
```
00:16:43 cp: skipping file ‘/hpc/local/share/automake-1.15/ar-lib’, as it was replaced while being copied
00:16:43 configure.ac:41: error: installing 'confdb/ar-lib'
00:16:43 configure.ac:41:     error while copying
```
Let's try refreshing it.
bot:mellanox:retest

jsquyres:
In talking through this PR, @jjhursey identified a few more places that need to be de-BE-PPC'ed, such as https://github.com/open-mpi/ompi/blob/master/config/opal_config_asm.m4#L1010.

Please at least file a ticket to remove that dead code, too.  Thanks!

PR #4104, opened 2017-08-16

### LINKED ISSUE #4053: Add XL compiler version check to configure

jjhursey:
 * Fixes #4053
 * `NEWS` update in separate commit since it's custom for this branch.


jsquyres:
@hppritcha Good to go.

PR #4105, opened 2017-08-16

jjhursey:
 * Fixes Issue #4053


jsquyres:
@gpaulsen I'm good with this, but can you provide a secondary review from IBM?

jsquyres:
@hppritcha Good to go.

PR #4107, opened 2017-08-16
Remove dead code related to PPC Big Endian

### LINKED ISSUE #4102: Remove PowerPC Big Endian integration
Follow-on to Issue #4053 there are some places in the code that still refer to:
  `powerpc-*|powerpc64-*|ppc-*`

These should be removed since they represent dead code. It may take a while to identify all of the spots, so we are doing so separately.

jjhursey:
* Resolves #4102

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -1007,7 +1007,11 @@ AC_DEFUN([OPAL_CONFIG_ASM],[
             OPAL_GCC_INLINE_ASSIGN='"or %0,[$]0,[$]0" : "=&r"(ret)'
             ;;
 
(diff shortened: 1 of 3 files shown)


rhc54:
@jjhursey Do we need to port the configure changes relating to #4102 to the PMIx code as well?

jjhursey:
@rhc54 Yeah we should. I'll take point on that. It should be a slight variation on the patch in #4053

hjelmn:
I think the consensus is to bring back BE support.

hppritcha:
@jjhursey what's happening with this PR?  Should it be closed?

jsquyres:
Yes, per https://github.com/open-mpi/ompi/issues/4349#issuecomment-374970982, we should close this PR without merging.

PR #4121, opened 2017-08-21
mca: Dynamic components link against project lib

### LINKED ISSUE #3705: Fixing dlopen() the MPI shared library with RTLD_LOCAL
The lack of support for dlopen()ing the MPI shared library within a local namespace has been a recurrent issue for implementors of MPI bindings in dynamic languages like R, Julia, Python. Even the Java JNI bindings you guys distribute have to resort to the hack of re-dlopening the library with `RTLD_GLOBAL` before the invocation of `MPI_Init()`. I complained about this ages ago, it was never fixed, and eventually I rolled my own re-delopen hack for mpi4py. However, I really HATE it because is quite fragile. For example, macOS changed at some point the rules. Also, in Linux distros, if the user does not install the `openmpi-devel` package, the symlink `libmpi.so -> libmpi.so.<version>` is not available, then the code implementing the hackery has to be updated from time to time every time the library version is bumped. BTW, you Java JNI bindings are broken because of this, you should not `dlopen("libmpi.so",...)`, you have to `dlopen("libmpi.so.20",...)` in Linux to not depend on the `openmpi-devel` package at runtime.

As a reproducer, you have this piece of Python code.

```python
import ctypes
libdir = "/home/devel/mpi/openmpi/2.1.1/lib/"

lib = ctypes.CDLL(libdir+"libmpi.so", ctypes.RTLD_LOCAL)
ierr = lib.MPI_Init(None,None)
assert ierr==0

r = ctypes.create_string_buffer(4096)
n = ctypes.c_int()
ierr = lib.MPI_Get_library_version(r, ctypes.byref(n))
assert ierr==0
print(r[0:n.value])

ierr = lib.MPI_Finalize()
assert ierr==0
```

Running it of course fails:
```shell
$ python ompi-dlopen.py 
[kw14821:30046] mca_base_component_repository_open: unable to open mca_patcher_overwrite: /home/devel/mpi/openmpi/2.1.1/lib/openmpi/mca_patcher_overwrite.so: undefined symbol: mca_patcher_base_patch_t_class (ignored)
... lots of output ...
[kw14821:30046] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```

The root problem is that all the `*.so` files in `$prefix/lib/openmpi/` do not explicitly link to the MPI library. The following script hot-fix the issues using `patchelf` (version 0.9 required):

```bash
#!/bin/sh
prefix="/home/devel/mpi/openmpi/2.1.1"
for filename in $(ls $prefix/lib/openmpi/*.so); do
    patchelf --add-needed libmpi.so.20 $filename
    patchelf --set-rpath "\$ORIGIN/.." $filename
done
```
Please note the special rpath I'm adding `$ORIGIN\..`, for macOS it should be`@loader_path/..` and the script should be based in `install_name_tool`.

After running the shell script above and hot-fixing the plugins, the Python script runs just fine:
```shell
$ ./ompi-fix-libs.sh 
$ python ompi-dlopen.py 
Open MPI v2.1.1, package: Open MPI dalcinl@kw14821 Distribution, ident: 2.1.1, repo rev: v2.1.0-100-ga2fdb5b, May 10, 2017
```

In short, I think fixing the issue once and for all (at least for Linux, macOS, and Solaris) is to link the plugins in `$prefix/lib/openmpi` with `-L$BUILDDIR/lib -lmpi -Wl,-rpath,\$ORIGIN/..` in Linux/Solaris and `-L$BUILDDIR/lib -lmpi -Wl,-rpath,@loader_path/..` in macOS.

Unfortunately, I cannot offer a patch, I'm not an expert on autotools and I have no idea how to implement these changes. However, I can offer my help to review changes and test them in Linux and macOS.

jjhursey:
* Resolves #3705
 * Components should link against the project level library to better
   support `dlopen` with `RTLD_LOCAL`.
 * Different project levels link to different sets of libraries by
   using the `mca_FRAMEWORK_COMPONENT_la_LIBADD` in the `Makefile.am`.
```
MCA components in ompi/
       $(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la
MCA components in orte/
       $(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la
MCA components in opal/
```

Note: The changes in this commit were automated. Some components
were not included because they are statically built only.

### DIFF
--- a/contrib/libadd_mca_comp_update.py
+++ b/contrib/libadd_mca_comp_update.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2017      IBM Corporation. All rights reserved.
+# $COPYRIGHT$
+#
+
+import glob, os, re, shutil
+
+projects= {'opal' : ["$(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la"],
+           'orte' : ["$(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la"],
+           'ompi' : ["$(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la"],
+           'oshmem' : ["$(top_builddir)/oshmem/liboshmem.la"],
+          }
+
+no_anchor_file = []
+missing_files = []
+skipped_files = []
+partly_files = []
+updated_files = []
+
+#
+# Check of all of the libadd fields are accounted for in the LIBADD
+# Return a list indicating which are missing (positional)
+#
+def check_libadd(content, libadd_field, project):
+    global projects
+
+    libadd_list = projects[project]
+    libadd_missing = [True] * len(libadd_list)
+
+    on_libadd = False
+    for line in content:
+        # First libadd line
+        if re.search( r"^\s*"+libadd_field, line):
+            # If line continuation, then keep searching after this point
+            if line[-2] == '\\':
+                on_libadd = True
+
+            for idx, lib in enumerate(libadd_list):
+                if True == libadd_missing[idx]:
+                    if 0 <= line.find(lib):
+                        libadd_missing[idx] = False
+
+        # Line continuation
+        elif True == on_libadd:
+            for idx, lib in enumerate(libadd_list):
+                if True == libadd_missing[idx]:
+                    if 0 <= line.find(lib):
+                        libadd_missing[idx] = False
+
+            # No more line continuations, so stop processing
+            if line[-2] != '\\':
+                on_libadd = False
+                break
+
+    return libadd_missing
+
+#
+# Update all of the Makefile.am's with the proper LIBADD additions
+#
+def update_makefile_ams():
+    global projects
+    global no_anchor_file
+    global missing_files
+    global skipped_files
+    global partly_files
+    global updated_files
+
+    for project, libadd_list in projects.items():
+        libadd_str = " \\\n\t".join(libadd_list)
+
+        print("="*40)
+        print("Project: "+project)
+        print("LIBADD:\n"+libadd_str)
+        print("="*40)
+
+        #
+        # Walk the directory structure
+        #
+        for root, dirs, files in os.walk(project+"/mca"):
+            parts = root.split("/")
+            if len(parts) != 4:
+                continue
+            if parts[-1] == ".libs" or parts[-1] == ".deps" or parts[-1] == "base":
+                continue
+            if parts[2] == "common":
+                continue
+
+            print("Processing: "+root)
+
+            #
+            # Find Makefile.am
+            #
+            make_filename = os.path.join(root, "Makefile.am")
+            if False == os.path.isfile( make_filename ):
+                missing_files.append("Missing: "+make_filename)
+                print("  ---> Error: "+make_filename+" is not present in this directory")
+                continue
+
+            #
+            # Stearching for: mca_FRAMEWORK_COMPONENT_la_{LIBADD|LDFLAGS}
+            # First scan file to see if it has an LIBADD / LDFLAGS
+            #
+            libadd_field  = "mca_"+parts[2]+"_"+parts[3]+"_la_LIBADD"
+            ldflags_field = "mca_"+parts[2]+"_"+parts[3]+"_la_LDFLAGS"
+            has_ldflags = False
+            has_libadd  = False
+
+            r_fd = open(make_filename, 'r')
+            orig_content = r_fd.readlines()
+            r_fd.close()
+            libadd_missing = []
+
+            for line in orig_content:
+                if re.search( r"^\s*"+ldflags_field, line):
+                    has_ldflags = True
+                elif re.search( r"^\s*"+libadd_field, line):
+                    has_libadd = True
+
+            if True == has_libadd:
+                libadd_missing = check_libadd(orig_content, libadd_field, project)
+
+            #
+            # Sanity Check: Was there an anchor field.
+            # If not skip, we might need to manually update or it might be a
+            # static component.
+            #
+            if False == has_ldflags and False == has_libadd:
+                no_anchor_file.append("No anchor ("+ldflags_field+"): "+make_filename)
+                print("  ---> Error: Makefile.am does not contain necessary anchor")
+                continue
+
+            #
+            # Sanity Check: This file does not need to be updated.
+            #
+            if True == has_libadd and all(False == v for v in libadd_missing):
+                skipped_files.append("Skip: "+make_filename)
+                print("       Skip: Already updated Makefile.am")
+                continue
+
+            #
+            # Now go though and create a new version of the Makefile.am
+            #
+            r_fd = open(make_filename, 'r')
+            w_fd = open(make_filename+".mod", 'w')
+
+            num_libadds=0
+            for line in r_fd:
+                # LDFLAGS anchor
+                if re.search( r"^\s*"+ldflags_field, line):
+                    w_fd.write(line)
+                    # If there is no LIBADD, then put it after the LDFLAGS
+                    if False == has_libadd:
+                        w_fd.write(libadd_field+" = "+libadd_str+"\n")
+                # Existing LIBADD field to extend
+                elif 0 == num_libadds and re.search( r"^\s*"+libadd_field, line):
+                    parts = line.partition("=")
+                    num_libadds += 1
+
+                    if parts[0][-1] == '+':
+                        w_fd.write(libadd_field+" += ")
+                    else:
+                        w_fd.write(libadd_field+" = ")
+
+                    # If all libs are missing, then add the full string
+                    # Otherwise only add the missing items
+                    if all(True == v for v in libadd_missing):
+                        w_fd.write(libadd_str)
+                        # Only add a continuation if there is something to continue
+                        if 0 != len(parts[2].strip()):
+                            w_fd.write(" \\")
+                        w_fd.write("\n")
+                    else:
+                        partly_files.append("Partly updated: "+make_filename)
+                        for idx, lib in enumerate(libadd_list):
+                            if True == libadd_missing[idx]:
+                                w_fd.write(lib+" \\\n")
+
+                    # Original content (unless it's just a line continuation)
+                    if 0 != len(parts[2].strip()) and parts[2].strip() != "\\":
+                        w_fd.write("\t"+parts[2].lstrip())
+
+                # Non matching line, just echo
+                else:
+                    w_fd.write(line)
+
+            r_fd.close()
+            w_fd.close()
+
+            #
+            # Replace the original with the updated version
+            #
+            shutil.move(make_filename+".mod", make_filename)
+            updated_files.append(make_filename)
+
+
+if __name__ == "__main__":
+
+    update_makefile_ams()
+
+    print("")
+
+    print("="*40);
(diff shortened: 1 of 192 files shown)


ompi/mca/bml/r2/Makefile.am
```diff
@@ -37,6 +38,9 @@ mcacomponentdir = $(ompilibdir)
 mcacomponent_LTLIBRARIES = $(component_install)
 mca_bml_r2_la_SOURCES = $(r2_sources)
 mca_bml_r2_la_LDFLAGS = -module -avoid-version
+mca_bml_r2_la_LIBADD = $(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la \
+	$(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la \
```
bwbarrett: blah; First, we should find a way to macro-ize this, so there's only one thing in LIBADD.

But second, is a dependency on all three libraries the right thing here?  It means that a bump in the major version of any one of the three will result in a load failure, which I guess is fine?

jjhursey:
Note the WIP tag. Initial commit basically reversed [this commit](https://github.com/open-mpi/ompi/commit/213b5d5c6e59387f8b700873484aa83795c0d422) with a script. Initial PR posted to give folks an early chance to test.

This PR needs some cleanup still including:
 * Cleanup Java bindings (dlopen trick no longer needed)
 * Does not include the oshmem project (only ompi/orte/opal)
 * See if we clean generalize the per-component change using project-level rules or additional build system scripts.

If you have more suggestions for changes feel free to keep the conversation going here or on the issue.

dalcinl:
@jjhursey I confirm all of mpi4py testsuite passes. I have two minor regressions, but they seem to be unrelated to this PR (these failures occur even if I dlopen libmpi).

rhc54:
@jjhursey You really cannot do this - building "-no-orte" and using your own RTE will cause all of these components to fail. You can only be certain that the project the component belongs to will have built - you cannot assume any other project did.

bwbarrett:
I'm not sure I completely agree with @rhc54.  I agree that we probably shouldn't explicitly include the liborte / libopal paths in every OMPI makefile (or libopal in every ORTE makefile).  But I think at configure time, you can know which liborte you're going to link with (at least, whether it's a relative path or out-of-tree libopen-rte), since you can know when you link libmpi.  So that's solvable, but with more work than we have today.

It's also worth noting that not all components have a build-time dependency on libopen-rte, so it's not clear that we should be linking libopen-rte into every component.  Probably safest to do that during the initial pass, but definitely not necessary in general.

jjhursey:
I could try a version where each project level only links to it's project level library (and not that of the one(s) beneath it). I suspect that will still work for the driving use case since `libmpi` links against `libopen-rte` and `libopen-pal` (and `libopen-rte` links against `libopen-pal`). That seems a bit cleaner.

rhc54:
@bwbarrett I'm afraid that isn't true as the RTE enters the MPI layer solely thru the ompi/mca/rte framework. There is nothing in the configure line to tell you which RTE component is going to be selected, and only the RTE component actually links in the RTE library.

rhc54:
This is why there are no RTE symbols anywhere in the OMPI layer - not in the base, nor in any component.

bwbarrett:
@rhc54, ah, if we've actually completed that transition, then I agree with you.  We shouldn't link libopen-rte.so into the OMPI modules (nor should we need to).

Still an open question about open-pal.so, but it's not needed, so seems simpler to avoid it and just link in the project library.

jjhursey:
I just re-posted the commit where each component only links against it's project's lib. I'm rebuilding now to confirm it still fixes the problem.

jjhursey:
I confirmed that this linking model works for the python binding example.

jjhursey:
Per teleconf - This is looking good. A few notes:
 * Add the _update_ script to `contrib/` - so we have it for later
 * Note that PMIx and hwloc projects may want to consider a similar patch for their component setup.
 * Add `oshmem` project

I'll try to polish this off today/tomorrow.

dalcinl:
@jjhursey I've rebuilt with your latest changes, mpi4py worked just fine. Do you still need help with the Java bindings? I can contribute a patch, but I have no clue how to test it.

dalcinl:
@jjhursey `autogen.pl` after `git clean -dxf` produces the following warnings. Is that expected?

```shell
...
opal/mca/btl/openib/Makefile.am:121: warning: mca_btl_openib_la_LIBADD was already defined in condition TRUE, which includes condition OPAL_cuda_support ...
opal/mca/btl/openib/Makefile.am:117: ... 'mca_btl_openib_la_LIBADD' previously defined here
configure.ac: installing 'config/ylwrap'
opal/mca/btl/smcuda/Makefile.am:57: warning: mca_btl_smcuda_la_LIBADD was already defined in condition TRUE, which includes condition OPAL_cuda_support ...
opal/mca/btl/smcuda/Makefile.am:52: ... 'mca_btl_smcuda_la_LIBADD' previously defined here
opal/mca/rcache/gpusm/Makefile.am:54: warning: mca_rcache_gpusm_la_LIBADD was already defined in condition TRUE, which includes condition OPAL_cuda_support ...
opal/mca/rcache/gpusm/Makefile.am:51: ... 'mca_rcache_gpusm_la_LIBADD' previously defined here
opal/mca/rcache/grdma/Makefile.am:53: warning: mca_rcache_grdma_la_LIBADD was already defined in condition TRUE, which includes condition OPAL_cuda_support ...
opal/mca/rcache/grdma/Makefile.am:50: ... 'mca_rcache_grdma_la_LIBADD' previously defined here
opal/mca/rcache/rgpusm/Makefile.am:52: warning: mca_rcache_rgpusm_la_LIBADD was already defined in condition TRUE, which includes condition OPAL_cuda_support ...
opal/mca/rcache/rgpusm/Makefile.am:49: ... 'mca_rcache_rgpusm_la_LIBADD' previously defined here
...
```

jjhursey:
@dalcinl Ah good catch. I'll fix that in the next series of pushes (missed the `+=` in the script).
I also have the patch for Java that I'll push up too. I tested it locally with a simple Java MPI program and it worked fine.

jjhursey:
Ok. This should be good to go now. Please have a look. There are 3 individual commits to hopefully make it easier to review.

dalcinl:
One minor detail and a question about Java bindings:

* @jjhursey Remove the line `static void *libmpi = NULL;` from file `ompi/mpi/java/c/mpi_MPI.c`.
* @jsquyres Is the call to `opal_init_psm()` still needed? See 5071602 for reference.

jsquyres:
@dalcinl Unfortunately, I do not think we'll ever be able to remove that call to `opal_init_psm()`.

jjhursey:
I just pushed an update to the Java bindings commit that removed the `static void *libmpi = NULL;`. Thanks for catching that.

For the `opal_init_psm()` I wasn't sure if we needed it still or not. It seems like we still do from Jeff's comment. We should probably get someone to test PSM+Java with this branch to make sure all is ok.

rhc54:
I'll only modify that "never" a bit to be "for the next few years". AFAIK, the problem in the PSM library has been resolved, and so we only need it to protect against earlier versions until "everyone" has upgraded.

dalcinl:
@jsquyres I'm not sure you understood my question. Do you need to call  `opal_init_psm()` right there in the Java bindings bootstrap function before MPI is initialized?

matcabral:
Hi all,
@jsquyres I will test your request. 
@dalcinl , this has to be done here because and old version of PSM/PSM2 was hijacking signal handlers in the library constructor. Which implies this happened whenever it was dlopened.  Details at https://github.com/open-mpi/ompi/pull/1781 . As @rhc54 mentions, this was fixed in newer versions.

matcabral:
@jjhursey I just remembered the JAVA bindings are expected to not work with PSM:
https://www.open-mpi.org/faq/?category=java#java_limitations

dalcinl:
@matcabral Maybe that comment in the FAQ is actually outdated? 

OK, so now I think you should keep the `opal_init_psm()` call in the Java bindings. If the Java bindings are loaded but `MPI_Init()` is never called, you may be in trouble with the PSM signal hijacking. So better  to stay safe and keep the call.

jjhursey:
Once I get the 👍  from the reviewers we can merge, and then consider the change for the release branches.

contrib/libadd_mca_comp_update.py
```diff
@@ -0,0 +1,232 @@
...
+
+import glob, os, re, shutil
+
+projects= {'opal' : ["$(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la"],
+           'orte' : ["$(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la"],
+           'ompi' : ["$(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la"],
+           'oshmem' : ["$(top_builddir)/oshmem/liboshmem.la"],
+          }
+# projects= {'ompi' : ["$(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la"],
+#           }
```
jsquyres: Should probably remove this commented-out code.

contrib/libadd_mca_comp_update.py
```diff
@@ -0,0 +1,232 @@
...
+
+        
+if __name__ == "__main__":
+
+    update_makefile_ams()
+
+    print("")
+
+    print("="*40);
+    print("{:>3} : Files skipped".format(len(skipped_files)))
```
jsquyres: Why `len(..)` here, but `str(len(...))` elsewhere?

dalcinl:
@jsquyres What release is going to get this change? Maybe 3.0.0?
@jjhursey Do you know whether Spectrum MPI will incorporate the changes in this PR?

jsquyres:
@bwbarrett Is this too late for v3.0.0?

contrib/libadd_mca_comp_update.py
jjhursey: Good catch. I'll fix.

bwbarrett:
I think it's probably too late for 3.0.0, but since this doesn't break ABI, I think I'd be ok pulling it in for 3.0.1.  Let's see how rc4 goes; if there are other major issues, we can re-evaluate.

contrib/libadd_mca_comp_update.py
jjhursey: Probably just need the `len(` version here since I'm using the new `.format` mode that doesn't need it to be converted to a string first. I'll update the others.

jjhursey:
@jsquyres I just re-pushed fixing the python items you mentioned.

gpaulsen:
We'd love this to go into v3.0.0 (If there's going to be another RC of substantial content).  If v3.0.0 ships, we're excited about that too. 🥇  v3.0.1 will work in that case.

jjhursey:
@rhc or @bwbarrett Any other comments on this PR before I merge?

rhc54:
ok with me!

bwbarrett:
@jjhursey, Works for me.

rhc54:
Ouch - I have found one downside to this PR. Now every top-level compile will rerun thru every component and relink it, even if nothing changed in it or anywhere in the code base. Significantly slower build times.

Issue #4122, opened 2017-08-22
CMA compilation error in sm BTL


kawashima-fj:
Compilation of sm BTL fails with newer glibc on Linux/SPARC64 (at least).
We need to cherry-pick d984b4b to v2.x and v2.0.x branch. I'll create PRs.

Reported by @PHHargrove.
See https://www.mail-archive.com/devel@lists.open-mpi.org/msg20340.html

PR #4123, opened 2017-08-22
v2.x: CMA: Fixing logic for CMA system call detection

kawashima-fj:
The OPAL_CMA_NEED_SYSCALL_DEFS is always defined/set to 0 or 1. Therefore instead of checking if the macro is defined, we have to look at the value itself.

(cherry picked from commit d984b4b3f9f18a4dfae9a1ef3e9e7f2fbd1f0e33)

Refs: #4122

Though I set milestone to v2.1.2, I don't have strong opinion against deferring it to v2.1.3.


PHHargrove:
I can confirm that cherry-picking this one commit does resolve the build failure with the 2.1.2rc2 tarball, which I reported in https://www.mail-archive.com/devel@lists.open-mpi.org/msg20340.html

jsquyres:
@hppritcha Good to go

PR #4124, opened 2017-08-22
v2.0.x: CMA: Fixing logic for CMA system call detection

kawashima-fj:
The OPAL_CMA_NEED_SYSCALL_DEFS is always defined/set to 0 or 1. Therefore instead of checking if the macro is defined, we have to look at the value itself.

(cherry picked from commit d984b4b3f9f18a4dfae9a1ef3e9e7f2fbd1f0e33)

Refs: #4122

### DIFF
--- a/opal/mca/btl/sm/btl_sm.c
+++ b/opal/mca/btl/sm/btl_sm.c
@@ -19,6 +19,7 @@
  * Copyright (c) 2013-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016      ARM, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


PHHargrove:
I have retested the 2.0.3 release tarball and found that it builds fine on  the Linux/SPARC64 system I have access to, *without* the need for this change (though it also builds fine *with*).  
So, I am uncertain there is a need to cherry-pick this commit as there is for 2.x (where the 2.1.2rc2 tarball builds *only* with the cherry-picked commit).

kawashima-fj:
@PHHargrove Thanks for testing! It is because this feature (CMA) is disabled by default in v2.0.x and auto-detected in v2.1.x and later. You'll see the same error when you specify `--with-cma` to `configure`.

PHHargrove:
@kawashima-fj 
You are exactly correct.  
If I configure 2.0.3 using `--with-cma` the error occurs and the one commit is sufficient to fix it.

jsquyres:
@hppritcha Good to go.

PR #4125, opened 2017-08-22
configury: patch configure in order to correctly support flang compilers

ggouaillardet:

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -979,6 +979,28 @@ sub patch_autotools_output {
 
         # We have to change the search pattern and substitution on each
         # iteration to take into account the tag changing
+        my $search_string = '# icc used to be incompatible with GCC.\n\s+' .
+                            '# ICC 10 doesn\047t accept -KPIC any more.\n.*\n\s+' .
+	                    "lt_prog_compiler_wl${tag}=";
+        my $replace_string = "# Flang compiler
+      *flang)
+	lt_prog_compiler_wl${tag}='-Wl,'
+	lt_prog_compiler_pic${tag}='-fPIC -DPIC'
+	lt_prog_compiler_static${tag}='-static'
+        ;;
+      # icc used to be incompatible with GCC.
+      # ICC 10 doesn't accept -KPIC any more.
+      icc* | ifort*)
+	lt_prog_compiler_wl${tag}=";
+
+        push(@verbose_out, $indent_str . "Patching configure for flang Fortran ($tag)\n");
+        $c =~ s/$search_string/$replace_string/;
+    }
+
+    foreach my $tag (("", "_FC")) {
+
+        # We have to change the search pattern and substitution on each
+        # iteration to take into account the tag changing
         my $search_string = '\052Sun\134 F\052.*\n.*\n\s+' .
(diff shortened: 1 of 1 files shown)


ggouaillardet:

currently, Open MPI cannot be built out of the box with flang or (flang based) ARM compiler.
as described at https://developer.arm.com/products/software-development-tools/hpc/resources/porting-and-tuning/building-openmpi-with-arm-compiler a few extra steps are needed.
the root cause is libtool does not support flang compilers, but since we already hack configure in order to get support for what we need, i just made yet an other hack.

@PHHargrove by any chance, would you be able to test this PR with ARM compilers (i only tested flang on linux/x86_64) ? note you need recent autotools and run ```autogen.pl```

PHHargrove:
@ggouaillardet I don't have flang access on ARM64.  
I don't even have gcc-6 or gcc-7!    
I think @shamisp is likely a better resource for ARM64 testing.

ggouaillardet:
thanks @PHHargrove 

@shamisp could you please give this PR a try and provide some feedback ?

autogen.pl
```diff
@@ -979,6 +979,26 @@ sub patch_autotools_output {
 
         # We have to change the search pattern and substitution on each
         # iteration to take into account the tag changing
+        my $search_string = '# Lahey Fortran 8.1.\n.*\n\s+' .
```
jsquyres: Why is Lahey Fortran mentioned again down around new line 1026?

ggouaillardet: thanks, i will double check that tomorrow.
(new) line 1026 has a different purpose, but my ```$search_string``` might match more expressions than needed

ggouaillardet: i checked and each ```$search_string``` matches a different expression.
fwiw, the new ```$search_string``` is used to insert ```flang``` related stuff, whereas the previous one is used to enhance Lahey Fortran support.~

ggouaillardet:
@jsquyres i made the changes. latest libtool does not support ```flang``` so i submitted a patch upstream. meanwhile, we can hack the generated ```configure``` in ```autogen.pl``` as we do for similar purposes

shamisp:
@pawosm-arm can you please take a look at the flang patch ?

autogen.pl
```diff
@@ -979,6 +979,28 @@ sub patch_autotools_output {
 
         # We have to change the search pattern and substitution on each
         # iteration to take into account the tag changing
+        my $search_string = '# icc used to be incompatible with GCC.\n\s+' .
```
pawosm-arm: did we really have to introduce one more foreach loop? Couldn't the one that looks for Sun's compiler be reused?

autogen.pl
```diff
@@ -998,7 +1020,6 @@ sub patch_autotools_output {
         push(@verbose_out, $indent_str . "Patching configure for Sun Studio Fortran version strings ($tag)\n");
         $c =~ s/$search_string/$replace_string/;
     }
-
```
pawosm-arm: Don't add or remove lines.

autogen.pl
ggouaillardet: this is the way we have always done. it makes it easier to follow the logic of what we are doing.

an empty line was removed by mistake, and i fixed that

Issue #4126, opened 2017-08-22
Regression in MPI_Info_delete() with missing key


dalcinl:
This is with branch `jjhursey:explore/dlopen-local` which should be (very close to) current master plus the dlopen fixes.

```python
>>> from mpi4py import MPI
>>> info = MPI.Info.Create()
>>> info.Delete("key")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mpi4py/MPI/Info.pyx", line 86, in mpi4py.MPI.Info.Delete
mpi4py.MPI.Exception: MPI_ERR_INTERN: internal error
```
Previous versions of Open MPI failed with `MPI_ERR_INFO_NOKEY`.

jjhursey:
Due to change made in PR #3743

jsquyres:
Looks straightforward to fix; I'll take a stab at it.

dalcinl:
@jsquyres After almost every new release of Open MPI, I found regressions by running mpi4py's testsuite. You know, I'm so maniac as to test the error code value after attempting to delete a missing key from an Info object.

Would it be possible to somehow get mpi4py testsuite run as part of the CI infrastructure? I proposed this in the past, and I got comments like *we may not be interested in mpi4py bugs*. I'm not claiming that my codebase (and its very few dependencies) is bug-free, but from experience on the last 10 years, the mpi4py bug rate is almost null. Most of the bug reports and issues we get are always related to build issues because of broken user environment, dynamic linking issues related to the infamous dlopen situation, of bugs in the backend MPI implementation. Most of mpi4py's testsuite is plagued of special-casing to disable tests that are know to fail as Open MPI releases appeared over the years. I'm now at a point were most of mpi4py tests pass, and it would be good to keep things that way for the future to come :-). 

mpi4py is thoroughly tested on Linux and macOS with MPICH and Open MPI, and on Windows with Microsoft MPI, all of this using public CI services (Travis CI, Circle CI, AppVeyor, Bitbucket Pipelines). Yes, sh.. happens, a future bug in mpi4py could eventually slip and disrupt Open MPI testing infrastructure, but I think the risk outweighs the benefits. I'm quite confident that mpi4py should help a lot in preventing regressions like the one reported in this issue.

jsquyres:
@dalcinl BTW, I don't know if you know how much we value your tests and bug reports.  If we haven't said it before, THANK YOU!

[...]

PR #4128, opened 2017-08-22
mpi/info_delete: fix return code

jsquyres:
Per MPI-3.1, ensure to raise an MPI exception with value MPI_ERR_INFO_NOKEY if we try to MPI_INFO_DELETE a key that does not exist.  Thanks to @dalcinl (Lisando Dalcin) for raising the issue.

Refs #4126.  Once this PR is merged, need to PR to v3.0.x branch.


jsquyres:
@dalcinl Can you verify that this patch fixes the regression you mentioned in #4126?

dalcinl:
@jsquyres Yes, your patch fixes #4126.

Issue #4131, opened 2017-08-22
BTL/TCP broken on master


rhc54:
When trying to run a simple ring program, I am getting the following error:
```
[rhc002][[16379,1],2][btl_tcp.c:556:mca_btl_tcp_recv_blocking] remote peer unexpectedly closed connection while I was waiting for blocking message
[rhc002][[16379,1],3][btl_tcp.c:556:mca_btl_tcp_recv_blocking] remote peer unexpectedly closed connection while I was waiting for blocking message
--------------------------------------------------------------------------
WARNING: Open MPI failed to handshake with a connecting peer MPI
process over TCP.  This should not happen.

Your Open MPI job may now fail.

  Local host: rhc002
  PID:        209028
  Message:    did not receive entire connect ACK from peer
--------------------------------------------------------------------------
```
The ring still completes - does anyone know why this started happening?

bwbarrett:
@rhc54, is this two porcesses on the same host or different hosts?

rhc54:
The two processes reporting the warning are on the same host, which is odd because vader is also enabled. However, they are talking to two procs on another host, and that might be the connection they are complaining about.

ggouaillardet:
it looks like there is some inconsistency here.

from ```mca_btl_tcp_recv_blocking()```
```c
int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
{
       int retval = recv(sd, ((char *)ptr) + cnt, size - cnt, 0);
        /* remote closed connection */
        if (0 == retval) {
            BTL_ERROR(("remote peer unexpectedly closed connection while I was waiting for blocking message"));
            return -1;
        }
```
so if zero byte is read
- an error message is displayed
- the function returns ```-1```


but in ```mca_btl_tcp_endpoint_recv_connect_ack()```

```c
static int mca_btl_tcp_endpoint_recv_connect_ack(mca_btl_base_endpoint_t* btl_endpoint)
{
    mca_btl_tcp_endpoint_hs_msg_t hs_msg;
    retval = mca_btl_tcp_endpoint_recv_blocking(btl_endpoint, &hs_msg, sizeof(hs_msg));

    if (sizeof(hs_msg) != retval) {
        if (0 == retval) {
            /* If we get zero bytes, the peer closed the socket. This
               can happen when the two peers started the connection
               protocol simultaneously. Just report the problem
               upstream. */
            return OPAL_ERROR;
        }
[...]

PR #4134, opened 2017-08-23
btl/tcp: fix mca_btl_tcp_recv_blocking() returned value

ggouaillardet:
if a short read occurs, return the number of bytes read,
and only issue an error message if something was read, otherwise
that might be just fine (e.g. this is how we detect a race condition)

Fixes open-mpi/ompi#4131

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp.c
+++ b/opal/mca/btl/tcp/btl_tcp.c
@@ -552,9 +552,12 @@ int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
     while (cnt < size) {
         int retval = recv(sd, ((char *)ptr) + cnt, size - cnt, 0);
         /* remote closed connection */
-        if (0 == retval) {
+        if (0 == retval && 0 == cnt) {
+            /* nothing was read at all, this might be fine, so simply returns */
+            return 0;
+        } else if (0 == retval) {
             BTL_ERROR(("remote peer unexpectedly closed connection while I was waiting for blocking message"));
-            return -1;
+            return cnt;
         }
 
         /* socket is non-blocking so handle errors */


opal/mca/btl/tcp/btl_tcp.c
```diff
@@ -552,9 +552,12 @@ int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
     while (cnt < size) {
         int retval = recv(sd, ((char *)ptr) + cnt, size - cnt, 0);
         /* remote closed connection */
-        if (0 == retval) {
+        if (0 == retval && 0 == cnt) {
+            /* nothing was read at all, this might be find, so simply returns */
```
bosilca: fine not find.

opal/mca/btl/tcp/btl_tcp.c
```diff
@@ -552,9 +552,12 @@ int mca_btl_tcp_recv_blocking(int sd, void* data, size_t size)
...
+            return 0;
+        } else if (0 == retval) {
             BTL_ERROR(("remote peer unexpectedly closed connection while I was waiting for blocking message"));
-            return -1;
+            return cnt;
         }
```
bosilca: why not simply returning ```cnt``` ?

ggouaillardet: i fixed the typo, thanks for noticing it.
i am fine with simply returning ```cnt``. i do not fully understand the rationale for #3955, so an error message might be desired if we read "something but not all" here.
@mohanasudhan can you please comment on that ?

bosilca: This looks like an oversight more than anything else. If you read anything between 0 and the expected size it is clearly a network issue in today's implementation (we drop connections before sending the first byte of the handshake).

mohanasudhan: @ggouaillardet ya you are right if we read "something but not all" then we throw error. In case of cnt being 0 ( where here both the connection are from MPI proc but happens simultaneously) then one of the connection will be dropped because of error. But if we read in between 0 - expected size then we are throwing error as we expect (guid and magic string) to recognize connection is from MPI proc.

mohanasudhan:
bot:ompi:retest

opal/mca/btl/tcp/btl_tcp.c
mohanasudhan: Thanks for making this change. It looks fine for me as we have ERROR when cnt !=0 , I didn't realized simultaneously case is pretty common.  

I haven't looked at git yesterday, will try to follow closed henceforth.

mohanasudhan:
On checking Mellanox failure <http://bgate.mellanox.com/jenkins/job/gh-ompi-master-pr/6248/console>. It is failing with "05:13:37 /bin/sh: ../../.././config/depcomp: No such file or directory
05:13:37 make[5]: *** [keyval_lex.lo] Error 127" seems its not related to this PR. Hence restarting this test 

bot:ompi:retest

mohanasudhan:
bot:ompi:retest

bwbarrett:
bot:mellanox:retest

actually kick off the mellanox tests.

mohanasudhan:
Can we merge this PR ? anyone has objection?

bosilca:
This patch does not fix the problem. I have a different approach (patch to come), one that unifies the different interfaces using blocking operations.

mohanasudhan:
@bosilca Let me know if you have the patch. If not can a elaborate a bit so that i can pick in cleaning this up.

mohanasudhan:
I try to reproduce this issue locally here. I have two ec2 machine, and triggering simple sendReceive program in loop. 

for i in {1..1000}; do mpirun -host machine1:16,machine2:16 --mca btl tcp,self sendReceive.o; done;

Running this command on both those machines simultaneously. But I couldn't catch this racing condition. If you guys have any suggestion to reproduce it, let me know. I am going to try few more methods.

mohanasudhan:
I am able to reproduce the issue locally. Working on fix now.  Just fyi updating command i use. 

[ec2-user@ip-172-31-32-140 ~]$ mpirun -host ip-172-31-44-240,ip-172-31-44-240 --mca btl tcp,self sendReceive.o | grep -v Received
[ip-172-31-44-240][[19259,1],1][btl_tcp.c:556:mca_btl_tcp_recv_blocking] remote peer unexpectedly closed connection while I was waiting for blocking message
WARNING: Open MPI failed to handshake with a connecting peer MPI
process over TCP.  This should not happen.

Your Open MPI job may now fail.

  Local host: ip-172-31-44-240
  PID:        79195
  Message:    did not receive entire connect ACK from peer

`
[ec2-user@ip-172-31-32-140 ~]$ cat sendReceive.c 
#include "mpi.h"
#include <stdio.h>

int main(argc,argv)
int argc;
char *argv[];  {
int numtasks, i, rank, dest, source, rc, count, tag=1;
char inmsg, outmsg='x';
MPI_Status Stat;

MPI_Init(&argc,&argv);
MPI_Comm_size(MPI_COMM_WORLD, &numtasks);
for(i=0; i< 10000; i++) {
  if (rank == 0) {
    dest = 1;
    source = 1;
    rc = MPI_Send(&outmsg, 1, MPI_CHAR, dest, tag, MPI_COMM_WORLD);
    rc = MPI_Recv(&inmsg, 1, MPI_CHAR, source, tag, MPI_COMM_WORLD, &Stat);
    }

  else if (rank == 1) {
    dest = 0;
    source = 0;
    
    rc = MPI_Send(&outmsg, 1, MPI_CHAR, dest, tag, MPI_COMM_WORLD);
    rc = MPI_Recv(&inmsg, 1, MPI_CHAR, source, tag, MPI_COMM_WORLD, &Stat);
  }
rc = MPI_Get_count(&Stat, MPI_CHAR, &count);
printf("Task %d: Received %d char(s) from task %d with tag %d \n",
       rank, count, Stat.MPI_SOURCE, Stat.MPI_TAG);
}


MPI_Finalize();
}
`

mohanasudhan:
This happens when simultaneous connect

```
[ip-172-31-32-140:53497] btl:tcp: connect() to 172.31.44.240:1024 completed (complete_connect), sending connect ACK
[ip-172-31-32-140:53497] btl:tcp: now connected to 172.31.44.240, process [[15401,1],1]
[ip-172-31-44-240:85697] btl:tcp: connect() to 172.31.32.140:1024 completed (complete_connect), sending connect ACK
[ip-172-31-44-240:85697] btl:tcp: now connected to 172.31.32.140, process [[15401,1],0]
```

bosilca:
Yes, the simultaneous handshake is broken. Before the connection initiated by the smallest guid was dropped by both peers. Currently an earlier condition decide to drop the remote connection on each peer (because they already have the one locally initiated). Thus, both decide to drop the peer connection.

bosilca:
Here is a [patch](https://github.com/bosilca/ompi/commit/a8f894185e0b803ed1ddb3421f55bac06d1c3b33). It returns whatever has been read out the socket, and leaves the decision to the upper level. I removed the mca_btl_tcp_endpoint_recv_blocking function, and instead make a call directly to mca_btl_tcp_recv_blocking. I validated all possible execution path from this function, to make sure the return value is used correctly. It also change the output in case of warning.

mohanasudhan:
taken care in this PR https://github.com/open-mpi/ompi/pull/4295

PR #4137, opened 2017-08-24
Fix the orte-dvm operations so that orterun can connect and execute an application. 

rhc54:
There is a lingering problem, though. The first invocation of orterun succeeds every time. However, subsequent invocations have a high probability of hanging in the OOB connection handshake.


rhc54:
@ggouaillardet Would you have a few cycles to dig into this connection issue? I think the disconnect from the first invocation is leaving the OOB in some unsettled state.

rhc54:
@bwbarrett I'm not sure when I'll have time to try and finish the repairs, so this may not make it for 3.0 unless someone else can help.

PR #4140, opened 2017-08-25
mca: Dynamic components link against project lib

### LINKED ISSUE #3705: Fixing dlopen() the MPI shared library with RTLD_LOCAL

jjhursey:
* Resolves #3705
 * Ref PR #4121 for `master`
 * Components should link against the project level library to better
   support `dlopen` with `RTLD_LOCAL`.
 * Different project levels link to different sets of libraries by
   using the `mca_FRAMEWORK_COMPONENT_la_LIBADD` in the `Makefile.am`.
```
MCA components in ompi/
       $(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la
MCA components in orte/
       $(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la
MCA components in opal/
```

Note: The changes in this commit were automated. Some components
were not included because they are statically built only.

### DIFF
--- a/contrib/libadd_mca_comp_update.py
+++ b/contrib/libadd_mca_comp_update.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2017      IBM Corporation. All rights reserved.
+# $COPYRIGHT$
+#
+
+import glob, os, re, shutil
+
+projects= {'opal' : ["$(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la"],
+           'orte' : ["$(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la"],
+           'ompi' : ["$(top_builddir)/ompi/lib@OMPI_LIBMPI_NAME@.la"],
+           'oshmem' : ["$(top_builddir)/oshmem/liboshmem.la"],
+          }
+
+no_anchor_file = []
+missing_files = []
+skipped_files = []
+partly_files = []
+updated_files = []
+
+#
+# Check of all of the libadd fields are accounted for in the LIBADD
+# Return a list indicating which are missing (positional)
+#
+def check_libadd(content, libadd_field, project):
+    global projects
+
+    libadd_list = projects[project]
+    libadd_missing = [True] * len(libadd_list)
+
+    on_libadd = False
+    for line in content:
+        # First libadd line
+        if re.search( r"^\s*"+libadd_field, line):
+            # If line continuation, then keep searching after this point
+            if line[-2] == '\\':
+                on_libadd = True
+
+            for idx, lib in enumerate(libadd_list):
+                if True == libadd_missing[idx]:
+                    if 0 <= line.find(lib):
+                        libadd_missing[idx] = False
+
+        # Line continuation
+        elif True == on_libadd:
+            for idx, lib in enumerate(libadd_list):
+                if True == libadd_missing[idx]:
+                    if 0 <= line.find(lib):
+                        libadd_missing[idx] = False
+
+            # No more line continuations, so stop processing
+            if line[-2] != '\\':
+                on_libadd = False
+                break
+
+    return libadd_missing
+
+#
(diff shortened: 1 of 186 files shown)


jjhursey:
Posting for consideration (if there is another rc) for v3.0.0. Otherwise it can slip to v3.0.1

dalcinl:
@jsquyres I confirm the `dlopen` issues are fixed with `v3.0.0rc5-2-gfc852fe9c2`. However, I'm having now some trouble with `MPI_Comm_spawn()` calls, but they are likely unrelated to this PR. Let me figure out what's wrong and I'll submit a new issue.

PR #4141, opened 2017-08-27
pml_ucx: fix tag/context_id layout and upper bounds.

yosefe:

PR #4142, opened 2017-08-28
pml_ucx: fix tag/context_id layout and upper bounds.

yosefe:
picked from master #4141

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx_request.h
+++ b/ompi/mca/pml/ucx/pml_ucx_request.h
@@ -26,15 +26,15 @@ enum {
 /*
  * UCX tag structure:
  *
- * 01234567 01234567 01234567 01234567 01234567 01234567 01234567 01234567
- *                           |                          |
- *      message tag (24)     |     source rank (24)     |  context id (16)
- *                           |                          |
+ * 01234567 01234567 01234567 01234567 01234567 0123 4567 01234567 01234567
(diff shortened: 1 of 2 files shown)


yosefe:
@jladd-mlnx @miked-mellanox

ompi/mca/pml/ucx/pml_ucx_request.h
```diff
@@ -26,15 +26,15 @@ enum {
...
+ * 01234567 01234567 01234567 01234567 01234567 0123 4567 01234567 01234567
+ *                           |                      |
+ *      message tag (24)     |   source rank (20)   |     context id (20)
+ *                           |                      |
  */
 #define PML_UCX_TAG_BITS                       24
-#define PML_UCX_RANK_BITS                      24
-#define PML_UCX_CONTEXT_BITS                   16
-#define PML_UCX_ANY_SOURCE_MASK                0x800000000000fffful
+#define PML_UCX_RANK_BITS                      20
```
jladd-mlnx: Only 20 bits for ranks? So UCX PML will only support ~1 million ranks?

yosefe: yes

jladd-mlnx: For exascale it's not enough. Can we extend it later?

PR #4143, opened 2017-08-28

aravindksg:

GPU non-contiguous memory.

### DIFF
--- a/ompi/mca/pml/cm/pml_cm_recvreq.h
+++ b/ompi/mca/pml/cm/pml_cm_recvreq.h
@@ -13,6 +13,7 @@
  * Copyright (c) 2012      Sandia National Laboratories.  All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2017      Intel, Inc. All rights reserved
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -92,7 +93,8 @@ do {                                                                           \
                                            src,                         \
                                            datatype,                    \
                                            addr,                        \
-                                           count )                      \
+                                           count,                       \
+					   flags )                      \
 do {                                                                    \
     OMPI_REQUEST_INIT(&(request)->req_base.req_ompi, false);            \
     (request)->req_base.req_ompi.req_mpi_object.comm = comm;            \
@@ -108,12 +110,13 @@ do {                                                                    \
     } else {                                                            \
         ompi_proc = ompi_comm_peer_lookup( comm, src );                 \
     }                                                                   \
+    MCA_PML_CM_SWITCH_CUDA_CONVERTOR_OFF(flags, datatype, count);       \
     opal_convertor_copy_and_prepare_for_recv(                           \
                                   ompi_proc->super.proc_convertor,      \
                                   &(datatype->super),                   \
                                   count,                                \
                                   addr,                                 \
-                                  0,                                    \
+                                  flags,                                    \
                                   &(request)->req_base.req_convertor ); \
 } while(0)
 #else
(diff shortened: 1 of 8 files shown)


ompi/mca/pml/cm/pml_cm_request.h
```diff
@@ -53,4 +54,22 @@ struct mca_pml_cm_request_t {
...
 OBJ_CLASS_DECLARATION(mca_pml_cm_request_t);
 
+/*
+ * Avoid CUDA convertor inits only for contiguous memory and if indicated by
+ * the MTL. For non-contiguous memory, do not skip CUDA convertor init phases.
+ */
+#if OPAL_CUDA_SUPPORT
+#define MCA_PML_CM_SWITCH_CUDA_CONVERTOR_OFF(convertor)                         \
+    {                                                                           \
+        if (opal_datatype_is_contiguous_memory_layout(&datatype->super, count)  \
```
bosilca: This macro is context sensitive. datatype and count should be part of the arguments.

ompi/mca/pml/cm/pml_cm_recvreq.h
```diff
@@ -134,6 +137,7 @@ do {                                                                    \
     OBJ_RETAIN(comm);                                                   \
     OMPI_DATATYPE_RETAIN(datatype);                                     \
                                                                         \
+    MCA_PML_CM_SWITCH_CUDA_CONVERTOR_OFF(ompi_mpi_local_convertor);     \
```
bosilca: I don't think you should meddle with the global convertor, nor in fact with any convertor that does not belong to the MTL requests. I noticed that you also alter the ```ompi_proc->super.proc_convertor```, which is used as a master copy for the convertor and will be copied by all convertors for the peer.

Thus this patch might work for simple blocking applications, but it will not work for any multi-threaded scenarios (as you alter a convertor that will be used by multiple requests in same time instead of altering the convertor that belongs to the request).

matcabral: Hi @bosilca,  just be on the same page, you mean the flags should be set on ```&(request)->req_base.req_convertor``` , right?

bosilca: Yes, ```(request)->req_base.req_convertor``` is the convertor associate with each request.

matcabral: ok, will work with @aravindksg to update the patch. thanks!

aravindksg:
Updated the patch to address @bosilca 's concerns.

ompi/mca/pml/cm/pml_cm_request.h
```diff
@@ -53,4 +54,22 @@ struct mca_pml_cm_request_t {
...
+ * the MTL. For non-contiguous memory, do not skip CUDA convertor init phases.
+ */
+#if OPAL_CUDA_SUPPORT
+#define MCA_PML_CM_SWITCH_CUDA_CONVERTOR_OFF(flags, datatype, count)            \
+    {                                                                           \
+        if (opal_datatype_is_contiguous_memory_layout(&datatype->super, count)  \
+            && (ompi_mtl->mtl_flags & MCA_MTL_BASE_FLAG_CUDA_INIT_DISABLE)) {   \
+            flags |= CONVERTOR_SKIP_CUDA_INIT;                                  \
+        } else {                                                                \
+            flags &= ~CONVERTOR_SKIP_CUDA_INIT;                                 \
```
bosilca: The flag passed to opal_convertor_copy_and_prepare_for_* is added to the flags of the original convertor (via an or operation). Moreover, it seems you initialize ```flags``` to 0 in all cases, so I don't think there is a need to &~CONVERTOR_SKIP_CUDA_INIT here.

rhc54:
Thanks @bosilca - I'm commiting for @aravindksg as he doesn't yet have commit privileges.

PR #4119, opened 2017-08-21
remove nmcheck_prefix.pl test due to false positives

markalle:
This test has proven to produce too many false positives so far. I hope
to re-enable it in the future, but until it has a longer history of not
producing false postivies it doesn't need to produce false nuisance
failures for everybody.

PR #4145, opened 2017-08-29
Removing the symbol_name test from make check until it can be reworked.

gpaulsen:
Reworking this check in a script in this PR:
https://github.com/open-mpi/ompi/pull/4119


gpaulsen:
After 4119 was reworked, that was better than this PR.  So we accepted that, and removing this.

PR #4147, opened 2017-08-29
Update configure logic to PMIx v1.2.4

### LINKED ISSUE #4109: autogen.pl problems with PMIx 1.2.3
I was noticing when running autogen.pl on a system that use to be able to autogen correctly for the 2.1.x now fails in part of the pmix configury.  This seems to be related to the upgrade to 1.2.3. For master and the other branches, autogen.pl still works for me.

Unfortunately the system where i observed this problem is not currently accessible.

rhc54:
Fixes #4109 

@hppritcha Can you give this a try on the problem machine? I don't have something I can test it against, but this should resolve the problem.

### DIFF
--- a/opal/mca/pmix/pmix112/pmix/VERSION
+++ b/opal/mca/pmix/pmix112/pmix/VERSION
@@ -16,7 +16,7 @@
 
 major=1
 minor=2
-release=3
+release=4
 
 # greek is used for alpha or beta release tags.  If it is non-empty,
 # it will be appended to the version number.  It does not have to be
@@ -24,14 +24,14 @@ release=3
 # The only requirement is that it must be entirely printable ASCII
 # characters and have no white space.
 
-greek=rc1
+greek=
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"
 # command, or with the date (if "git describe" fails) in the form of
 # "date<date>".
 
-repo_rev=gitb5387205
+repo_rev=gita3f126f
 
 # If tarball_version is not empty, it is used as the version string in
 # the tarball filename, regardless of all other versions listed in
@@ -45,7 +45,7 @@ tarball_version=
 
 # The date when this release was created
 
-date="Aug 14, 2017"
+date="Sep 06, 2017"
 
 # The shared library version of each of PMIx's public libraries.
 # These versions are maintained in accordance with the "Library
--- a/opal/mca/pmix/pmix112/pmix/config/pmix.m4
+++ b/opal/mca/pmix/pmix112/pmix/config/pmix.m4
@@ -35,7 +35,13 @@ AC_DEFUN([PMIX_SETUP_CORE],[
 
     AC_REQUIRE([AC_USE_SYSTEM_EXTENSIONS])
     AC_REQUIRE([AC_CANONICAL_TARGET])
-    AC_REQUIRE([AC_PROG_CC])
+
+    # AM_PROG_CC_C_O AC_REQUIREs AC_PROG_CC, so we have to be a little
+    # careful about ordering here, and AC_REQUIRE these things so that
+    # they get stamped out in the right order.
+    AC_REQUIRE([_PMIX_START_SETUP_CC])
+    AC_REQUIRE([_PMIX_PROG_CC])
+    AC_REQUIRE([AM_PROG_CC_C_O])
 
     # If no prefix was defined, set a good value
     m4_ifval([$1],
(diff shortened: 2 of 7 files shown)


hppritcha:
This still does not work for me.  The only way I get it to configure is by dnl'ing 
```AC_REQUIRE([AM_PROG_CC_C_O])``` in pmix_setup_cc.m4.

**Comment by jsquyres** [comment]
Defer to 2.1.3 (see #4109 for more details).

**Comment by rhc54** [comment]
@hppritcha Please try this again - I think we have a solution now

**Comment by nmorey** [comment]
This should really go into 2.1.2 as it is currently impossible to build in on SUSE Leap 42.3 without a fix

**Comment by rhc54** [comment]
@nmorey Can you confirm this fixes the problem?

**Comment by jsquyres** [comment]
@nmorey I'm a bit confused -- we thought this issue would only appear if you ran `autogen.pl`, which you shouldn't need to do on a distribution tarball (and therefore the scope of this problem was quite limited, and therefore safe to defer to v2.1.2).  Can you describe the exact scenario which is refusing to build on SUSE Leap 42.3?  Thanks.

**Comment by hppritcha** [comment]
This also seems to be bringing in more than just configury fixes.

**Comment by rhc54** [comment]
It represents the current state of 1.2.4 - and no, I honestly don't have time to separate everything out. Given it was deferred, there seemed no point to do so.

**Comment by nmorey** [comment]
@rhc54 this does not fix the issue.

On a Leap 42.3:
```
=== Processing subdir: /work1/nmorey/workspace/alternates/master/openmpi/opal/mca/pmix/pmix112/pmix
--- Found autogen.sh; running...
Running: ./autogen.sh
==> Running autoreconf
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I ./config
configure.ac:173: error: AC_PROG_CC cannot be called after AM_PROG_CC_C_O
configure.ac:173: the top level
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal: error: echo failed with exit status: 1
autoreconf: aclocal failed with exit status: 1
Command failed: ./autogen.sh
```


@jsquyres: We are working with github tarballs instead of the release ones as it allows more flexibility when trying out new version. Anyway we often have a patches on the build system for flag/path reasons and need to rerun autogen.

**Comment by nmorey** [comment]
```
$ aclocal --version 
aclocal (GNU automake) 1.13.4
```

jsquyres:
@nmorey Open MPI's official distributions are the tarballs on www.open-mpi.org.  If downstream maintainers want to use a different mechanism, then you should be aware that you're outside of our official / tested distribution mechanism.

I don't know if this PR will get into 2.1.2 or not.  I'd encourage you to use/test an official tarball, as that will give you the best results (and will solve this immediate issue).

rhc54:
I honestly don't have time to work on the configure issue right now, so I'd recommend deferring to 2.1.3 unless someone else has available cycles.

jjhursey:
I can work on this, but I won't get to it until mid-next week at this point.

nmorey:
@jsquyres As I said we often need to patch stuff in the build system to fit SUSE packaging guidelines and in these case, we need to rerun autogen as autoreconf usually fails due to different versions being used.

I am willing to switch to release tarballs. But I still need to be able to patch them later on. And if it won't rebuild, it's too great a risk.

nmorey:
@jjhursey If you need help for testing, feel free to contact me

jsquyres:
@nmorey Using the release tarballs is the right choice.  We do some post-processing during `make dist` (e.g., putting release dates in man pages and the like) that you won't get in github tarballs, and will confuse branding/versioning tracking.

`autogen.pl` is in the tarball, too, so it *is* possible to run it (aside from whatever this issue is).  We advise most users *not* to run `autogen.pl`, but it's there precisely for downstream packagers such as yourself that carry patches and need to re-Autotools the package to run configure/yadda/yadda/yadda.

jsquyres:
@hppritcha After talking to @jjhursey, the only reason for this PR is to fix the configury issue.  If this PR doesn't fix the configury issue, we might as well wait for a proper fix.

rhc54:
I loaded AM 1.13.4 and found the remaining problem - there was still an AC_PROG_CC in configure.ac. Removed it and all builds fine, so this is good to go.

jsquyres:
Thanks @rhc54!

@nmorey Can you confirm that this now solves the problem?

nmorey:
@jsquyres @rhc54 I tested 9c31385 both alone (cherry picked on v2.x) and with the pmix update of this PR and both are working on my system.

hppritcha:
I'll open a different PR with just 9c31385

hppritcha:
moving milestone for this PR to 2.1.3.

jjhursey:
So PR #4196 took one portion of this PR. The full version of this PR will be taken into PMIx in https://github.com/pmix/pmix/pull/486
Should we wait and refresh this PR from PMIx 1.2 branch?

jsquyres:
@jjhursey Yes, that would be great, thanks.  We took the minimum-distance fix that was necessary to get v2.1.2 out the door -- the rest can come in for v2.1.3.

jsquyres:
@nmorey Thanks for the testing / confirmation!

Issue #4144, opened 2017-08-29
romio ADIOI_GEN_WriteStrided with mixed contig/non-contig types


markalle:
I'm able to hit the below bug with master, if I run on a filesystem that takes the ADIOI_GEN_WriteStrided path for MPI_File_write().  (I was using gpfs, and offhand the users of that path appear to be gpfs, hfs, ntfs, panfs, pfs, sfs, ufs, xfs.)

Testcase:  https://gist.github.com/markalle/d7da240c19e57f095c5d1b13240dae24
% mpicc -o x romio_write_timing.c
% mpirun -mca io romio314 -np 4 ./x

The test has one of the ranks using a small contiguous datatype, eg
```
// for R0 : [ . . . x x . . . ]
```
while the other ranks use non-contiguous types that surround the above, eg
```
// for R1 : [ . . x . . x . . ]
// for R2 : [ . x . . . . x . ]
// for R3 : [ x . . . . . . x ]
```

The ADIOI_Gen_writestrided() in ad_write_str.c has the contiguous rank fall into this code:
```
            ADIO_WriteContig(fd, buf, count, datatype, ADIO_EXPLICIT_OFFSET,
                             offset, status, error_code);
```
around line 294, at which point it's not holding any ADIO_WRITE_LOCK.

While the non-contiguous ranks end up around line 356 in code that goes
```
/* contiguous in memory, noncontiguous in file. should be the most
   common case. */
...
                    ADIOI_BUFFERED_WRITE
(that macro boils down to ADIOI_WRITE_LOCK and read of a contiguous region like "x . . . . x")
...
        /* write the buffer out finally */
..
            ADIO_WriteContig(...);
            if (!(fd->atomicity))
                ADIOI_UNLOCK(fd, writebuf_off, SEEK_SET, writebuf_len);
```

If all the ranks had non-contiguous data and took the { lock, read, write, unlock } path it would work, but having some ranks do a direct contiguous write with no lock creates a race condition.

I prototyped a fix that added ADIOI_WRITE_LOCK/UNLOCK around the code the contiguous rank uses and that worked.  But I'd like thoughts from someone with romio knowledge about whether there's a better way.  I assume my fix would have performance implications.

roblatham00:
Ugh, wrapping LOCK/UNLOCK around contiguous writes gets you correct behavior but at significant performance cost.  Factor of 5x on our GPFS systems, easy. 

[...]

PR #4148, opened 2017-08-29
switch ADIOI_GEN_WriteStrided to ADIOI_GEN_WriteStrided_naive

markalle:
When ADIOI_GEN_WriteStrided wants to write data into a file with
a non-contiguous view, it does so with the sequence
    lock
    read
    modify
    write
    unlock

This adequately protects against other ranks executing the same
sequence. But as near as I can tell it isn't standard practice
to lock/unlock around all write calls in the entire romio product.

So the result is a very selective protection -- protection against
other ranks who happen to be doing the same operations we're doing,
but no protection against other ranks who could be doing something
else hitting one of the myriad of other write calls in romio.

The _naive version does more of a piecemeal walking of the datatype
writing each chunk. Unless we put a lock/unlock around every write
in romio, I think this is the only safe way to do it.

See bug report:
https://github.com/open-mpi/ompi/issues/4144
which includes testcase
https://gist.github.com/markalle/d7da240c19e57f095c5d1b13240dae24

### DIFF
--- a/ompi/mca/io/romio314/romio/adio/ad_gpfs/ad_gpfs.c
+++ b/ompi/mca/io/romio314/romio/adio/ad_gpfs/ad_gpfs.c
@@ -31,7 +31,7 @@ struct ADIOI_Fns_struct ADIO_GPFS_operations = {
     ADIOI_GEN_SetInfo, /* SetInfo for any platform besides BlueGene or PE */
 #endif
     ADIOI_GEN_ReadStrided, /* ReadStrided */
-    ADIOI_GEN_WriteStrided, /* WriteStrided */
+    ADIOI_GEN_WriteStrided_naive, /* WriteStrided_naive */
     ADIOI_GPFS_Close, /* Close */
 #ifdef ROMIO_HAVE_WORKING_AIO
 #warning Consider BG support for NFS before enabling this.
--- a/ompi/mca/io/romio314/romio/adio/ad_sfs/ad_sfs.c
+++ b/ompi/mca/io/romio314/romio/adio/ad_sfs/ad_sfs.c
@@ -20,7 +20,7 @@ struct ADIOI_Fns_struct ADIO_SFS_operations = {
     ADIOI_SFS_Fcntl, /* Fcntl */
     ADIOI_GEN_SetInfo, /* SetInfo */
     ADIOI_GEN_ReadStrided, /* ReadStrided */
-    ADIOI_GEN_WriteStrided, /* WriteStrided */
+    ADIOI_GEN_WriteStrided_naive, /* WriteStrided_naive */
     ADIOI_GEN_Close, /* Close */
     ADIOI_FAKE_IreadContig, /* IreadContig */
     ADIOI_FAKE_IwriteContig, /* IwriteContig */
--- a/ompi/mca/io/romio314/romio/adio/ad_ufs/ad_ufs.c
+++ b/ompi/mca/io/romio314/romio/adio/ad_ufs/ad_ufs.c
@@ -21,7 +21,7 @@ struct ADIOI_Fns_struct ADIO_UFS_operations = {
     ADIOI_GEN_Fcntl, /* Fcntl */
     ADIOI_GEN_SetInfo, /* SetInfo */
     ADIOI_GEN_ReadStrided, /* ReadStrided */
-    ADIOI_GEN_WriteStrided, /* WriteStrided */
+    ADIOI_GEN_WriteStrided_naive, /* WriteStrided_naive */
     ADIOI_GEN_Close, /* Close */
 #ifdef ROMIO_HAVE_WORKING_AIO
     ADIOI_GEN_IreadContig, /* IreadContig */


markalle:
@jsquyres Any suggestions for a romio person to evaluate if this is the best fix?

rhc54:
@roblatham00 Can you please take a look?

ggouaillardet:
Since `MPI_File_set_view()` is a collective operation, should we check here if we have a mix of contiguous and non-contiguous file type ? that would make the fix less intrusive imho.
note ```romio314``` is now lagging behind the latest mpich romio. by any chance, did you test the latest mpich ?

roblatham00:
wrapping writes with lock/unlock is what ROMIO's NFS driver does.   that's a recipie for awful performance.  (not a theoretical concern --  IBM's MPI for BlueGene does this and we have to say "set this odd-looking environment variable to pretend the file system is PVFS" to get performance back.

markalle:
Thanks for the feedback. I was guessing universal lock/write/unlock would be a big slowdown but it's good to have confirmation.

For a minute there I was liking the idea of detecting mixed-contig/non-contig types in MPI_File_set_view() and making the fix more targeted. But trying to work through that design I'm getting skeptical now.

I wanted to be able to say "all ranks have a non-contiguous view" therefore GEN_WriteStrided()'s use of { lock, read, modify, write, unlock } is safe since all the ranks will be doing it. But that's the same logical bug we started with: assuming all the ranks are doing the same thing. The peer ranks could be in completely different calls where writes occur w/o that protection. Uniformity of a non-contiguous view isn't enough.

I think in theory we could do the reverse and have a legal design: a uniformly contiguous view guarantees no rank will be doing GEN_WriteStrided's { lock, .. unlock } path, and that run-mode would allow the current overall lock-light design to work. But the existence of a single rank with a non-contiguous view could force a universal lock/unlock around all writes to that file, the only way for GEN_WriteStrided()'s current behavior to be safe. I think that could work, but I don't think it would be a good fix. It seems like would force low performance on a lot of users (are non-contig views rare?)

To the mpich question: I didn't evaluate mpich or newer romio.

roblatham00:
Stepping back a bit, the worry is that in a collective call, some other process might be updating the file, right?

We can lean on MPI-IO consistency semantics here pretty heavily.   Updates outside of the communicator are undefined ("All file
accesses in MPI are relative to a specific file handle created from a collective open").  

Could independent operations interfere with collective operations?  Well, years ago I added some logic to keep processes in a collective operation from leaving until everyone was done.  It definitely has a performance implication, but I needed to add that to defend against a mix of independent and collective operations in an HDF5-generated workload. 

https://github.com/open-mpi/ompi/blob/master/ompi/mca/io/romio314/romio/adio/common/ad_write_coll.c#L213

There is the flag MPI_MODE_UNIQUE_OPEN, but that does not defend against the workload given in the test case.  That defends against some external process coming along and manipulating the file while the MPI-IO processes are also operating on the file.  No implementation does anything with this flag. 

Referring back to the test case, one process sees what looks like contiguous data (we might have relaxed the "contiguous" check to be contiguous in more cases, now that I think about it) and interferes with the other processes doing data-sieving.  A type check at file_set_view time isn't enough because we set the file view and then could do any mix of independent or collective operations.  

I hate to throw one more collective call in the collective write path but we might need to allreduce the "contig-ness" of access.

markalle:
I want to clarify one thing though: the data sieving is happening in the individual write not the collective write. The test is just using MPI_File_write(), not the _all() variant.

So to amend what you said, it's not that in a collective call I'm worried some other process might be updating the file, it's that we're in the individual MPI_File_write() with that worry. That makes it worse since we have very little knowledge of what other peers might be doing. But even if it was a collective, it seems there could still be any number of individual MPI_File_iwrite() calls from peers trickling through in the background.

Fwiw I didn't really look at what MPI_File_write_all() does, but if I use that call it passes.

ggouaillardet:
```MPI_File_set_view()``` is a collective operation, so at this stage, i do not think it matters if we use the collective (e.g. ```MPI_File_write_all()```) or the individual (e.g. ```MPI_File_write()```) subroutine

roblatham00:
If you want to see what life is like with lock/unlock around each I/O call, you can prefix the file with 'nfs:' (if you've configured ROMIO to build the NFS driver).  

Collecting information in the set_view call is a good first step.  Maybe that's enough?   Some workloads are going to get flagged as more noncontiguous than they really are, and we might have to skip the pnetcdf-motivated optimizaiton that issues a contiguous write for some patterns.  

@wkliao : got any clever ideas?

markalle:
I didn't know about the history of this. Is that the application that led to the data sieving MPI_File_write() call when it wants to write with a non-contig view? How big was the performance difference for that vs the _naive() version that walks the datatype writing each contiguous chunk instead? And same question on lock/unlock : how bad does it get if that was on more universally? because I don't see how you can legally data-sieve without the mode being universal lock/unlock

I still think the collective-ness of MPI_File_set_view() doesn't help us much. It lets us know the view all our peers have. The MPI_File_write() is still a non-collective call though, so it's trying to data sieve without knowing what its peers are up to. That lack of knowing what the peers are doing is what leads me to say they'd have to be using an NFS-style universal lock/write/unlock

Playing the odds, that might still be a performance win for most runs: if one rank is data sieving they _probably_ all are and the existing code is protected for that case. I'd hate to implement this as "collect File_set_view datatypes, switch to the NFS function pointers if any rank is non-contig" though. Or is that just a suggestion for proof-of-concept on this?

wkliao:
Hi, Rob

For independent MPI I/O, if data sieving is enabled, then all contiguous writes should be protected by a lock, in addition to read-modify-write operations. Look like ROMIO needs to fix this.

As for the approach of checking fileview being whether contiguous or not, the only true contiguous case is when the entire file is visible to all processes. Otherwise, after the call to MPI_File_set_view, some processes may still make a noncontiguous request which triggers data sieving.

Note collective writes also do data sieving. However, the file domains are partitioned as contiguous and exclusive file regions among I/O aggregators. No read-modify-write will overlap with others, including contiguous writes. Thus, Collective I/O is fine in ROMIO.

CC: Rajeev Thakur @rsth

roblatham00:
I am in agreement that locking all writes will give us correct behavior (assuming the file system implements mandatory locking and not advisory locking etc).

I am deeply concerned about the performance cost of doing so, however.  
- we already see lock/unlock serialization contributing to massive overhead on Blue Gene with IBM's default driver.  
- Processing "noncontiguous in file" workloads with 'naiive ' will result in abysmal performance in any workload with more than a handful of chunks.    Every segment will turn into a system call and you will carry out a denial of service attack against your nice shiny parallel file system.   We saw this on Blue Gene too -- our PVFS file system did not support file locking and could only implement _naive for noncontig-in-file . 

I hate to fall back to this non-solution solution, but it might be the case that we will have to release-note this issue?  

    > "if your workload is like this, you should disable data sieving, explicitly request the 'nfs' ROMIO driver, or  use collective I/o"

something along those lines?

wkliao:
Was your significant performance impact on BG happening to independent I/O or collective?
For collective, there is no need to lock, even data sieving is enabled. So it must not be collective.

For independent where the fileviews may be interleaved, I do not know any better alternative solution, other than turning off data sieving. I also think turning off data sieving is better than using lock/unlock in most cases. It has less file system overhead.

roblatham00:
Poor performance with independent I/O of course.   Collective gives us many more options.

There is no one answer to the question "is lock/write/unlock better than naive (piece-wise) i/o".  If the access is very sparse, data sieving will read/modify/write more than needed, but if the access is very fragmented naive will issue many write() system calls.

markalle:
In the cases where sieving is needed for performance I'm guessing there aren't a crazy amount of simultaneous updates going at the same time, so if we only forced lock/unlock onto that particular use-case the performance might remain adequate.

I don't think this can be auto-detected though. We'd be hitting with a very broad brush, perhaps turning any file with a non-contiguous view into lock/write/unlock mode. Is there a logical place within the MPI_File_* API for hints to go, like the info object at Open or set_view time? Data sieving could be an optimization that's only offered if the app specifically gives us a hint, and an app would only choose to do that if they weren't going crazy with simultaneous updates from other ranks

markalle:
Any updates on this? (did the discussion migrate to some romio forum or is this still the main spot)?

hppritcha:
the author of this PR thinks it can be closed.

hppritcha:
reopening because Mark remember this was actually disabling an incorrect optimization

hppritcha:
@markalle and @edgargabriel here's the reopened PR

AboorvaDevarajan:

lanl-ompi:

jsquyres:
@markalle is this PR still relevant?  (it's still open)

markalle:
I think it can be closed, and possibly I'll make a smaller change as a new PR (probably over in mpich though).  In Spectrum I ended up using a more targeted fix for the testcase listed at the top of this PR, and in practice I couldn't actually get things to fail after that, even though I was never 100% convinced there was enough locking to make GEN_WriteStrided safe.

PR #4149, opened 2017-08-30
Move the common pmix2x integration code to the OPAL base, and share it

rhc54:
between the ext2x and pmix2x components.

Refs #3873 


rhc54:
@ggouaillardet Per discussion on #3873, I have refactored the code to push the common code into the pmix/base area. This appears to be working for me, but I'd appreciate a second pair of eyes on it. If it looks okay to you, feel free to "push" your changes into the new code area here - or you can wait for this to be committed and then do it separately.

ibm-ompi:

 Gist: https://gist.github.com/d0998366b77ff280a34f7a6e9c2738e5

ibm-ompi:

 Gist: https://gist.github.com/9029587c45ca1baaaf2d747f5c0a3d3f

ggouaillardet:
the build failure suggest you need to link ```libopen-pal.so``` with the embedded or external ```libpmix.so``` in order to resolve the many ```PMIx``` undefined symbols. having said that, this might never work if the ```ext1x``` component is built. bottom line, i am afraid moving stuff into the ```base``` directory is not a viable option.

rhc54:
I'm not convinced of that yet, so I'll give it a little more work. It builds fine for me, so I guess there is some more configury work to do.

ggouaillardet:
it does not build on my up-to-date CentOS 7 box. i guess it might build on some OS where the linker is not so picky (e.g. unresolved symbols are allowed when building a shared library).
or did you configure with ```--disable-dlopen``` ?
and did you run a simple program ? if i understand correctly, unresolved symbols will prevent the app from starting.

rhc54:
I did all of that on my Mac, so it is likely some configure issue.

rhc54:
Okay, I see the issue here, and why the Mac didn't show it. The problem is that we would have to explicitly link libpmix into the base. When building an external version, that might not be so bad - but it seems rather ugly when building the embedded version. It's fine for the static frameworks like hwloc - but that isn't the case here.

@jsquyres Does this sway your opinion about the symlink answer? Or should we just punt and go with parallel code as we have been doing (perhaps with the generator @ggouaillardet built)?

ggouaillardet:
i am afraid you need to do much more in order to avoid an abstraction violation.
for example ```opal/mca/pmix/base/common/pmix2x_client.c``` invokes ```PMIx_Init()```, which means ```libopen-pal.so``` will have an unresolved symbol ```PMIx_Init()``` (or whatever mangled name it has).
currently, this symbol is (indirectly) provided by a component 
- the embedded ```PMIx``` if ```pmix2x``` is built
- the external ```PMIx``` if ```ext2x``` is built
as far as i understand, you would at least have to move the embedded ```PMIx``` into the ```base```, link ```libopen-pal.so``` with the (embedded or external) ```libpmix.so```, and make sure there are no unresolved symbols if ```ext1x``` is used.

rhc54:
I think we could work around all that, if necessary. However, I think it is an unacceptable solution to link a component-generated library into the base, so it's a moot point.

rhc54:
I've convinced myself that the symlink option isn't a good one either - sigh. So we are left with the options we discussed on the other PR. I'll close this one and let's focus over there

PR #4156, opened 2017-08-31
Revert "* Some recent versions of GCC try very hard to make it imposs…

### LINKED ISSUE #4146: Absoft MTT 32 bit build failure on master
Absoft's MTT is showing a 32 bit build failure on master -- linking `opal_wrapper` seems to be looking for a non-existent `opal_cmpset_64` symbol.

See https://mtt.open-mpi.org/index.php?do_redir=2478.

@cagoelz sent the attached full logs, but they apparently don't show much beyond what we see in MTT generated by:

```
$ ../openmpi-master-201708290239-a84af67/configure FC=af90 FCFLAGS=-Z790 "CFLAGS=-g -pipe" --enable-picky --enable-debug --enable-mpirun-prefix-by-default 2>&1 | tee config_output.txt
$ make 2>&1 | tee make_output.txt
```

[make_output.txt](https://github.com/open-mpi/ompi/files/1261112/make_output.txt)
[config_output.txt](https://github.com/open-mpi/ompi/files/1261111/config_output.txt)

hjelmn:
…ible to"

This reverts commit b5ea5e0994a827915107e03d6744e73156534a04

This commit reverts a change that is hopefully not necessary. If this
is the case this will fix #4146.

### DIFF
--- a/opal/include/opal/sys/ia32/atomic.h
+++ b/opal/include/opal/sys/ia32/atomic.h
@@ -13,7 +13,7 @@
  * Copyright (c) 2007-2010 Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2015-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


hjelmn:
@bwbarrett Should be just go ahead an merge this and see if MTT works? I can build a ia32 virtual machine tomorrow to test it as well.

bwbarrett:
I don't think that you need a 32 bit VM; should fail if you compile in 32 bit mode.  I tried both gcc 4.7.4 and gcc 5.4.0 and both compiled in 32 bit mode without problem.  I looked at the old trac bug database and couldn't find any mention of which gcc version was irritating, but my guess is that it was in the 3.x series, based on timelines.  I vote we commit this tomorrow and see where MTT fails.

PR #4161, opened 2017-09-01
monitoring: fix MPI_T regression

hjelmn:
The monitoring code causes MPI_T based tools to segfault when
monitoring is disabled. This happens because the performance and
control variables remain registered after the common/monitoring
component is dlclosed. This commit is fixes the issue but doesn't
address some underlying design bugs that lead to the segfault.

The issue on github is #4162. Close when applied to master.

### DIFF
--- a/ompi/mca/common/monitoring/common_monitoring.c
+++ b/ompi/mca/common/monitoring/common_monitoring.c
@@ -1,3 +1,4 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2013-2017 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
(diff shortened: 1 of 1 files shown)


hjelmn:
@bosilca This resolves the SEGV but I think ideally the coll variables should be registered in coll/monitoring, pml in pml/monitoring, etc. Not sure how you want to address the issue or if this is fine for the long-term fix.

bosilca:
@hjelmn I was under the impression that MCA parameters were automatically unregistered upon module unloading. In fact I am certain it was a case at some point, because we documented that we had to find a work-around to keep a filename from being automatically freed (common_monitoring.c:290).

Was there a change in behavior ?

hjelmn:
@bosilca True. The big issue is the pvar's were missing the IWG flag. I am probably being overly cautious.

bosilca:
In this case adding the IWG flag should be enough to fix the issue right ? Looking through the OMPI source code it becomes clear that nobody is calling mca_base_var_group_deregister to unregister their MCA params. Isn't this an overkill then ? I'm OK with this patch I'm just not eager to add unnecessary code.

hjelmn:
Ok, I will update the patch to just add IWG. That should be enough.

PR #4165, opened 2017-09-04
SPML_UCX: use ompi_proc_world_size() to set the estimated_num_eps value

alinask:
before this fix, mca_spml_ucx_component_open was using
oshmem_num_procs() to set the value of params.estimated_num_eps for UCX.
The oshmem_num_procs() function uses oshmem_group_all which will be
initialized after the call to mca_spml_ucx_component_open and therefore,
cannot be used there.

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx_component.c
+++ b/oshmem/mca/spml/ucx/spml_ucx_component.c
@@ -128,7 +128,7 @@ static int mca_spml_ucx_component_open(void)
     memset(&params, 0, sizeof(params));
     params.field_mask = UCP_PARAM_FIELD_FEATURES|UCP_PARAM_FIELD_ESTIMATED_NUM_EPS;
     params.features   = UCP_FEATURE_RMA|UCP_FEATURE_AMO32|UCP_FEATURE_AMO64;
-    params.estimated_num_eps = oshmem_num_procs();
+    params.estimated_num_eps = ompi_proc_world_size();
 
     err = ucp_init(&params, ucp_config, &mca_spml_ucx.ucp_context);
     ucp_config_release(ucp_config);


alinask:
Fixes the UCX issue - https://github.com/openucx/ucx/issues/1799

oshmem/mca/spml/ucx/spml_ucx_component.c
```diff
@@ -128,7 +128,7 @@ static int mca_spml_ucx_component_open(void)
     memset(&params, 0, sizeof(params));
     params.field_mask = UCP_PARAM_FIELD_FEATURES|UCP_PARAM_FIELD_ESTIMATED_NUM_EPS;
     params.features   = UCP_FEATURE_RMA|UCP_FEATURE_AMO32|UCP_FEATURE_AMO64;
-    params.estimated_num_eps = oshmem_num_procs();
+    params.estimated_num_eps = ompi_proc_world_size();
```
miked-mellanox: @alinask - could you please add comment here to describe why mpi call used and not shmem

alinask: @miked-mellanox the oshmem_num_procs() function uses an unitialized value to get the number of processess in the job so the value returned by this function isn't correct. 
the ompi context was already initialized so using its function instead.
@alex-mikheev what do you think?

jsquyres: Why is the value in `osmem_num_procs()` uninitialized -- should that be fixed?

jsquyres:
@hppritcha This PR is failing in an NFS test on the Cray -- any idea what is happening there?

```
../../config/test-driver: line 107: 35146 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: opal_path_nfs
```

oshmem/mca/spml/ucx/spml_ucx_component.c
alex-mikheev: oshmem_num_procs() is initialized after spml frameworks are opened.

alinask: The call to `oshmem_proc_group_init()` which initializes that value is done after the opening and selection of all the components for the job - in `_shmem_init`.

alinask:
@yosefe @jladd-mlnx Please review as well.

jladd-mlnx:
@alinask This looks good. 👍

miked-mellanox:
@alinask - please integrate into v3.x and v2.x (?) branches

Issue #4153, opened 2017-08-30
Segmentation fault in simple program using MPI_Comm_accept()/connect()


awlauria:
Updated, new info 9/1/17

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

3.0.0rc4

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

Red Hat 7.3
Power and X86

# Details of the problem

I have a simple test that hits a segmentation fault intermittently in all v3.0 rc's.
The tests seems to pass on the master branch, though it's not apparent as to where the bug got 'fixed'. I attached a sample testcase that will fail running on a single node with two tasks. It may not fail every time, but if you run it in a loop you will hit the segmentation fault after some runs. The location the segmentation fault will change, so below is an example stack-trace.

I can only get it to crash using an optimized build.

Running with valgrind doesn't show any heap corruption, even on the fail case. So it seems to be stack related, unless valgrind is missing something.

```
0  0x000010000050eb18 in raise () from /lib64/libc.so.6
#1  0x0000100000510c9c in abort () from /lib64/libc.so.6
#2  0x0000100000555784 in __libc_message () from /lib64/libc.so.6
#3  0x000010000055f800 in malloc_consolidate () from /lib64/libc.so.6
#4  0x0000100000561be4 in _int_malloc () from /lib64/libc.so.6
#5  0x00001000005645ec in malloc () from /lib64/libc.so.6
#6  0x000010000059fc58 in __alloc_dir () from /lib64/libc.so.6
#7  0x000010000059fdcc in __opendirat () from /lib64/libc.so.6
#8  0x000010000059fe30 in opendir () from /lib64/libc.so.6
#9  0x00001000001f1004 in opal_os_dirpath_is_empty () from /smpi_dev/awlauria/ompi/exports/lib/libopen-pal.so.40
#10 0x00001000000a8d10 in orte_session_dir_cleanup () from /smpi_dev/awlauria/ompi/exports/lib/libopen-rte.so.40
#11 0x00000000100018f4 in ?? ()
#12 0x0000000010001060 in main (argc=-1, argv=0x0) at allgather_inter.c:43
```

sample run:

`mpirun -np 2 ./simple_test

[simple_test.zip](https://github.com/open-mpi/ompi/files/1270788/simple_test.zip)

[...]

PR #4167, opened 2017-09-05
orte_data_server: fix a typo in ORTE_PMIX_PUBLISH_CMD handling

ggouaillardet:
correctly balance some parenthesis ...

Fixes open-mpi/ompi#4153

Thanks Austen Lauria for the report

This is a one-off commit for the v3.0.x branch, master was fixed as part of a larger commit,
and the v2 branches are unaffected.

### DIFF
--- a/orte/runtime/orte_data_server.c
+++ b/orte/runtime/orte_data_server.c
@@ -351,8 +351,8 @@ void orte_data_server(int status, orte_process_name_t* sender,
                     OPAL_OUTPUT_VERBOSE((1, orte_data_server_output,
                                         "%s NOT STORING DATA FROM %s AT INDEX %d",
                                         ORTE_NAME_PRINT(ORTE_PROC_MY_NAME),
-                                        ORTE_NAME_PRINT(&data->owner), data->index);
-                    opal_pointer_array_set_item(&orte_data_server_store, data->index, NULL));
+                                        ORTE_NAME_PRINT(&data->owner), data->index));
+                    opal_pointer_array_set_item(&orte_data_server_store, data->index, NULL);
                     OBJ_RELEASE(data);
                     goto release;
                 }


ggouaillardet:
@bwbarrett this fix is obvious (though finding the root cause was not ...) and though the faulty path is not always taken and does not always cause a crash, i made this PR a blocker so hopefully it can make it for ```v3.0.0```.

rhc54:
The failure is in the opal_path_nfs test, which has nothing whatever to do with this PR.

gpaulsen:
This is important to get into v3.0.0.

bwbarrett:
bot:ompi:retest

bwbarrett:
It looks like something changed on the Cray so that the nfs check now is consistently failing.  I'm sure it doesn't have anything to do with this patch, but it needs to be addressed.  @hppritcha, can you take a look?  The test is only faling on the Cray and is consistently failing there.

hppritcha:
will do.  We can ignore for this PR.

hppritcha:
actually there's nothing screwed up on the crays.

problem is that the entries in /proc/mounts on some cray systems can be enormous, much larger than the buffer being used in the ```get_mounts``` method in opal_patn_nfs.c.  Anyway,  one is suppose to use ```getmntent``` and friends to avoid this problem.  

Do we want to fix the test or remove it from make check?

jjhursey:
We should resolve the CI issues separate from this PR. This PR is an obvious, correct, critical fix. We should get it in the branch soon (today?) - it's absence is causing testing problems with optimized builds the longer we let it sit here.

jsquyres:
@jjhursey #4179 was just merged that should fix the issue.  Let's see if we can get a nice green checkmark now...

bot:ompi:retest

rhc54:
Urrr...that commit went into master. How is that going to fix this PR into v3.0?

bwbarrett:
@jjhursey, the cranky RM for 3.0 has said no green, no merge.  So we need to get CI fixed...

jsquyres:
@rhc54 Errr... right.  Duh.  v3.0.x PR filed here: #4180

PRs for other branches to be filed shortly.

jjhursey:
Looks green to me 😄

PR #4171, opened 2017-09-05

aravindksg:

GPU non-contiguous memory.

(cherry picked from commit 2e83cf15ce790f89c782b6222253ab18252a7d2f)


bwbarrett:
ok to test

PR #3834, opened 2017-07-07
mtl/psm2: create mca variables to shadow PSM2 environment variables

hjelmn:
This commit enables MCA support for the following PSM2 environment
variables: PSM2_DEVICES, PSM2_MEMORY, PSM2_MQ_SENDREQS_MAX,
PSM2_MQ_RECVREQS_MAX, PSM2_MQ_RNDV_HFI_THRESH,
PSM2_MQ_RNDV_SHM_THRESH, PSM2_RCVTHREAD, PSM2_SHAREDCONTEXTS,
PSM2_SHAREDCONTEXTS_MAX, and PSM2_TRACEMASK. These variable can be set
by MCA if they are not already set in the environment.

Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

PR #4172, opened 2017-09-05

aravindksg:

GPU non-contiguous memory.

(cherry picked from commit 2e83cf15ce790f89c782b6222253ab18252a7d2f)

Conflicts:
	opal/datatype/opal_convertor.h

### DIFF
--- a/ompi/mca/mtl/psm2/help-mtl-psm2.txt
+++ b/ompi/mca/mtl/psm2/help-mtl-psm2.txt
@@ -1,7 +1,7 @@
 # -*- text -*-
 #
 # Copyright (C) 2009. QLogic Corporation.  All rights reserved.
-# Copyright (c) 2013-2015 Intel, Inc. All rights reserved.
+# Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -47,5 +47,17 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 Message size %llu bigger than supported by PSM2 API. Max = %llu
 #
 [no psm2 cuda env]
-Using CUDA enabled OpenMPI but PSM2_CUDA environment variable is %s.
-This is not a recommended combination. If the application uses %s.
+Warning: Open MPI has detected that you are running in an environment with CUDA
+devices present and that you are using Intel(r) Ompi-Path networking. However,
+the environment variable PSM2_CUDA was not set, meaning that the PSM2 Omni-Path
+networking library was not told how to handle CUDA support.
+
+If your application uses CUDA buffers, you should set the environment variable
+PSM2_CUDA to 1; otherwise, set it to 0. Setting the variable to the wrong value
+can have performance implications on your application, or even cause it to
+crash.
+
+Since it was not set, Open MPI has defaulted to setting the PSM2_CUDA
+environment variable to 1.
+
+Local hostname: %s
--- a/ompi/mca/mtl/psm2/help-mtl-psm2.txt
+++ b/ompi/mca/mtl/psm2/help-mtl-psm2.txt
@@ -45,3 +45,7 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 #
 [message too big]
 Message size %llu bigger than supported by PSM2 API. Max = %llu
+#
+[no psm2 cuda env]
+Using CUDA enabled OpenMPI but PSM2_CUDA environment variable is %s.
+This is not a recommended combination. If the application uses %s.
--- a/ompi/mca/mtl/psm2/mtl_psm2.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2.c
@@ -98,9 +98,6 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     char *generated_key;
     char env_string[256];
     int rc;
-#if OPAL_CUDA_SUPPORT
-    char *cuda_env;
-#endif
 
     generated_key = getenv("OMPI_MCA_orte_precondition_transports");
     memset(uu, 0, sizeof(psm2_uuid_t));
@@ -176,11 +173,6 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
 
 #if OPAL_CUDA_SUPPORT
     ompi_mtl_psm2.super.mtl_flags |= MCA_MTL_BASE_FLAG_CUDA_INIT_DISABLE;
-
-    cuda_env = getenv("PSM2_CUDA");
-    if (!cuda_env || ( strcmp(cuda_env, "0") == 0) )
-        opal_output(0, "Warning: If running with device buffers, there is a"
-                    " chance the application might fail. Try setting PSM2_CUDA=1.\n");
 #endif
 
     return OMPI_SUCCESS;
--- a/ompi/mca/mtl/psm2/mtl_psm2.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2.c
@@ -11,7 +11,7 @@
  * Copyright (c) 2004-2006 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006      QLogic Corporation. All rights reserved.
- * Copyright (c) 2013-2015 Intel, Inc. All rights reserved
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved
  * Copyright (c) 2014      Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
@@ -98,6 +98,9 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     char *generated_key;
     char env_string[256];
     int rc;
+#if OPAL_CUDA_SUPPORT
+    char *cuda_env;
+#endif
 
     generated_key = getenv("OMPI_MCA_orte_precondition_transports");
     memset(uu, 0, sizeof(psm2_uuid_t));
@@ -170,6 +173,15 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
 
     /* register the psm2 progress function */
     opal_progress_register(ompi_mtl_psm2_progress);
+
+#if OPAL_CUDA_SUPPORT
+    ompi_mtl_psm2.super.mtl_flags |= MCA_MTL_BASE_FLAG_CUDA_INIT_DISABLE;
+
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env || ( strcmp(cuda_env, "0") == 0) )
+        opal_output(0, "Warning: If running with device buffers, there is a"
+                    " chance the application might fail. Try setting PSM2_CUDA=1.\n");
+#endif
 
     return OMPI_SUCCESS;
 }
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -11,9 +11,9 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006-2010 QLogic Corporation. All rights reserved.
- * Copyright (c) 2012-2015 Los Alamos National Security, LLC.
- *                         All rights reserved.
- * Copyright (c) 2013-2016 Intel, Inc. All rights reserved
+ * Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
@@ -28,6 +28,7 @@
 #include "opal/mca/event/event.h"
 #include "opal/util/output.h"
 #include "opal/util/show_help.h"
+#include "opal/util/opal_environ.h"
 #include "ompi/proc/proc.h"
 
 #include "mtl_psm2.h"
@@ -42,6 +43,10 @@
 #include <glob.h>
 
 static int param_priority;
+
+#if OPAL_CUDA_SUPPORT
+static bool cuda_envvar_set = false;
+#endif
 
 static int ompi_mtl_psm2_component_open(void);
 static int ompi_mtl_psm2_component_close(void);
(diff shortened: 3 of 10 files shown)


bosilca:
The PR message mentions a conflict on opal/datatype/opal_convertor.h. I don't see any ...

jsquyres:
@artpol84 @Di0gen This looks like the same periodic failure we see from Mellanox Jenkins: it claims that `depcomp` does not exist.  Have you ever been able to track this down?

bot:mellanox:retest

aravindksg:
@bosilca , I had resolved the conflicts before committing the changes. There were two complaints from git:
1. Regarding one of the copyright lines at the top of the file. (Resolution was to only introduce the relevant copyright line to the patch)
2. Regarding the macro definition-
#define CONVERTOR_HAS_REMOTE_SIZE  0x20000000

I left out this macro definition to resolve the conflict resulting from the cherry-pick.

bwbarrett:
ok to test

artpol84:
@jsquyres unfortunately for now we don't know what causes depcomp to be missing. We'll try to keep an eye on this failure.

artpol84:
But I saw this
```
20:08:57 ompi/Makefile.am: installing 'config/depcomp'
20:08:58 parallel-tests: installing 'config/test-driver'
20:09:17 configure.ac: installing 'config/ylwrap'
20:09:45 cp: skipping file ‘/hpc/local/share/automake-1.15/depcomp’, as it was replaced while being copied
20:09:56 autoreconf: Leaving directory `.'
```
Which may be a clue. Will try to investigate.

jsquyres:
@artpol84 Cool.  I think we saw this same symptom quite a while ago -- the conjecture at the time was that somehow there were directory trees being re-used by different Jenkins builds.  Or somehow the autotools themselves were being re-installed behind the scenes (i.e., outside of Jenkins), which caused this race condition.  Good luck.

artpol84:
@jsquyres this should be fixed now, please let me know if this will be observed again.

jsquyres:
@artpol84 Great, thank you!  Just curious (since this went on for so long) -- what was the issue?

Re-run on the Cray to fix the opal_path_nfs test issue:

bot:ompi:retest

artpol84:
@jsquyres it was infrastructure-related. `depcomp` was accidentally "updated" quite frequently without need. "cp" was confused by `depcomp` file that was "modified" during the copy operation, copy was aborted and later missing `depcomp` was causing a problem.

bwbarrett:
Looks like the Cray went offline during the test...

bot:ompi:retest

aravindksg:
Ping @bosilca : could you please Ack this PR as well? The code has already been reviewed and merged to master and v3.0.x branch (PR's 4143 and 4171)

hppritcha:
The RMs for the 2.1.x release are reluctant to take this feature unless its fixing a correctness bug.

matcabral:
Hi @hppritcha,  this is not a bug fix, but a new feature. Without this patch OMPI does not support the GPU Direct functionality in PSM2 library.  
Does this mean that new features will ONLY be accepted in 3.x ? 
thanks,

matcabral:
Ok, I'm correcting the label to a performance bug and retracting my previous comment about new feature, and here is the rationale:  CUDA support is already part of OMPI. Without this fix OMPI built with CUDA support runs significantly slower on a CUDA aware PSM2 library.

ompi/mca/mtl/psm2/mtl_psm2.c
```diff
@@ -171,6 +174,15 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     /* register the psm2 progress function */
     opal_progress_register(ompi_mtl_psm2_progress);
 
+#if OPAL_CUDA_SUPPORT
+    ompi_mtl_psm2.super.mtl_flags |= MCA_MTL_BASE_FLAG_CUDA_INIT_DISABLE;
+
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env || ( strcmp(cuda_env, "0") == 0) )
+        opal_output(0, "Warning: If running with device buffers, there is a"
+                    " chance the application might fail. Try setting PSM2_CUDA=1.\n");
```
jsquyres: Please use `opal_show_help()` for these kinds of messages (because `opal_show_help()` deduplicates such messages when outputting to `mpirun`).  There's also very little context given in this message to inform the user that the message is being emitted from the Open MPI PSM2 code.

ompi/mca/mtl/psm2/mtl_psm2.c
```diff
@@ -171,6 +174,15 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     /* register the psm2 progress function */
     opal_progress_register(ompi_mtl_psm2_progress);
 
+#if OPAL_CUDA_SUPPORT
+    ompi_mtl_psm2.super.mtl_flags |= MCA_MTL_BASE_FLAG_CUDA_INIT_DISABLE;
+
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env || ( strcmp(cuda_env, "0") == 0) )
```
jsquyres: Is `PSM2_CUDA` an environment variable that is exported by the PSM2 library?  Or is that an env variable that is solely being used by Open MPI?  If it is solely used by the PSM2 support in Open MPI, it should be an MCA variable.

ompi/mca/pml/cm/pml_cm_recvreq.h
```diff
@@ -134,12 +138,13 @@ do {                                                                    \
...
     OBJ_RETAIN(datatype);                                               \
                                                                         \
+    MCA_PML_CM_SWITCH_CUDA_CONVERTOR_OFF(flags, datatype, count);       \
     opal_convertor_copy_and_prepare_for_recv(                           \
         ompi_mpi_local_convertor,                                       \
         &(datatype->super),                                             \
         count,                                                          \
         addr,                                                           \
-        0,                                                              \
+        flags,                                                              \
```
jsquyres: Nit: since the rest of the code here is scrupulously pretty, you might want to re-indent the `\` at the end of this line.

ompi/mca/pml/cm/pml_cm_recvreq.h
```diff
@@ -153,6 +158,7 @@ do {                                                                    \
                                           datatype,                     \
                                           addr,                         \
                                           count,                        \
+					  flags,                        \
```
jsquyres: Ditto with above -- you might want to re-indent `flags` to match the others (beware of tabs vs. spaces -- Open MPI style requires *spaces* and bans *tabs*).  This happens a few other places in this PR, too.

aravindksg: Will fix the indentation. Thought I converted tabs to spaces, but I'll double check.

ompi/mca/pml/cm/pml_cm_recvreq.h
aravindksg: will fix

ompi/mca/mtl/psm2/mtl_psm2.c
aravindksg: ok, will fix this

ompi/mca/mtl/psm2/mtl_psm2.c
aravindksg: It is an envvar exported by PSM2 library.

matcabral: it is actually not "exported", but "read" by PSM2.

aravindksg:
@jsquyres Since master and v3.0.x branch already pulled these changes in, shall I generate separate PR's to fix the warning message and indentation issues there?

jsquyres:
Do a separate PR for master to make the changes.  Once that PR is merged in, you can cherry pick the commit(s) from that master PR to this PR (i.e., just add the commits here -- no need to make a 2nd PR).

Make sense?

aravindksg:
@jsquyres , a patch to address your concerns has been merged to master and cherry-picked here. Could you also please review?

ompi/mca/mtl/psm2/help-mtl-psm2.txt
```diff
@@ -45,3 +45,7 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 #
 [message too big]
 Message size %llu bigger than supported by PSM2 API. Max = %llu
+#
+[no psm2 cuda env]
+Using CUDA enabled OpenMPI but PSM2_CUDA environment variable is %s.
```
jsquyres: Open MPI is two words, not one.

Also, please use full sentences and good grammar in these messages.  These are meant to be formal help messages displayed to the user, not "good enough" messages that are only intended for developers.  Also, the way you are adding an additional lengthy string via the call to `opal_show_help()` is... unconventional.

More below, where you actually call `opal_show_help()`.

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -89,6 +93,30 @@ ompi_mtl_psm2_component_register(void)
...
+     */
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env) {
+        opal_show_help("help-mtl-psm2.txt",
+                       "no psm2 cuda env", true,
+                       "not set",
+                       "Host buffers,\nthere will be a performance penalty"
+                       " due to OMPI force setting this variable now.\n"
+                       "Set environment variable to 0 if using Host buffers" );
+        setenv("PSM2_CUDA", "1", 0);
```
jsquyres: Can't this lead to a segv if you put the "1" string into the environment, but then the PSM2 MTL is `dlclose()`d (because the PSM2 MTL was not selected)?

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -89,6 +93,30 @@ ompi_mtl_psm2_component_register(void)
...
+     * to set it.
+     */
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env) {
+        opal_show_help("help-mtl-psm2.txt",
+                       "no psm2 cuda env", true,
+                       "not set",
+                       "Host buffers,\nthere will be a performance penalty"
+                       " due to OMPI force setting this variable now.\n"
+                       "Set environment variable to 0 if using Host buffers" );
```
jsquyres: Do not use "OMPI" in a show help string -- the product name is "Open MPI" (two words).

I would suggest that you should have 2 different show_help messages here, rather than feeding in a lengthy string to `opal_show_help()`.  

Additionally -- this message is being displayed during the component registration phase.  Isn't that far too early?  I.e., won't it display on systems that don't even have PSM2 and/or CUDA hardware available?

I would think that you should only display these messages if a) the PSM2 MTL is actually selected *and* b) there are CUDA devices in the system.  Otherwise, if this message is emitted if a) the user doesn't have PSM2 hardware or b) doesn't have CUDA devices, the user will (rightfully) be quite confused.  For example, you might replace this message with:

> Warning: Open MPI has detected both PSM2-based hardware and CUDA hardware, but the PSM2_CUDA environment variable was not set.  Open MPI has therefore defaulted to setting PSM2_CUDA=1.
>
> This can leave to a performance penalty, however, if your application uses host buffers.  You should set the PSM2_CUDA environment variable to 0 before invoking mpirun to both silence this warning and provide the hint to Open MPI / PSM2 that your application will mainly be using host buffers.
>
>   Local hostname: %s

(and include the local hostname in the message, just to let the user know which machine Open MPI is talking about)

You can have a similar message for the 2nd `opal_show_help()`, below.

Also, keep in mind that the component registration function is really only intended to be used to register MCA vars -- it is not actually intended to *do* anything in terms of initialization or setup.

If you *really really really* have to set the `PSM2_CUDA` environment variable all the way up here in the component registration function, you should include a lengthy comment about *why* you have to do so.  And then you'll need to set some kind of state variable (on the component?) indicating that you do so so that you can emit an appropriate show_help message later (i.e., if/when the PSM2 MTL is selected to be used).

Make sense?

ompi/mca/mtl/psm2/help-mtl-psm2.txt
```diff
@@ -45,3 +45,7 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 #
 [message too big]
 Message size %llu bigger than supported by PSM2 API. Max = %llu
+#
```
jsquyres: NOTE: I realize the `message too big` help message is not part of this PR, but I can't help seeing it.  You should not use `%llu` in printf-formatted strings (such as show_help messages).  Instead, you should use the appropriate `PRIfoo` escape sequence.  Please introduce another commit to fix the `message too big` show_help message (per my other comments, you might want to make this message a bit more verbose and clear to the user -- the message as it currently stands does not explain anything about what MPI API call caused this error, etc.).  Thanks!

bosilca: As .txt files are not preprocessed using ```PRIfoo``` would not deliver the expected outcome ?!

jsquyres: Er... you're right.  Duh.  Ok, disregard my comment about PRI.  😊

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: @jsquyres , will fix the help text output in separate commit against master and then cherry pick here.

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: It seems to be a harmless setting. I tested this by increasing OFI MTL priority to be higher than PSM2 and I did see the environment variable being set. But the application itself worked fine. Also- PSM2_DEVICES envvar setting is now done in component registration phase (PR #3834).

jsquyres: I'm not 100% sure it's harmless.  I think you may be leaving now-unallocated memory in the `environ` array when the PSM MTL is unloaded.  It *probably* won't cause problems, but if anyone goes to try to read it -- especially if that virtual address is no longer valid -- it may cause problems.  You might be able to test that a little more rigorously (this is one reason that passing values around in the environment is kinda sucky).

If I'm right, it may be best to `strdup()` the `"1"`, even though that technically makes a minor memory leak.

And/or you might defer this setenv until you know that PSM2 is going to be used.  Then you will have a much smaller window for the bogus memory pointer in `environ` to cause a problem (i.e., because it won't become bogus until the PSM MTL is closed during `MPI_FINALIZE`, but that's usually on the way out of the process, anyway).

rhc54:
FWIW: our normal procedure is to use opal_setenv to set it, mark in a flag that we did so, and then use opal_unsetenv to unset it in the close function. This protects the user from what @jsquyres describes.

jsquyres:
Oh, yes, unsetenving it in the component close function is a much better idea. 😄

aravindksg:
@jsquyres : The concerns have been addressed in PR #4323 and cherry-picked here. Please review.

jsquyres:
bot:ompi:retest

aravindksg:
@jsquyres : To allay any fears regarding testing: I have compile-tested this and also ran combinations of both CUDA and non-CUDA workloads against the resulting libmpi. They work fine 👍

jsquyres:
@aravindksg I figured.  I just wanted to qualify my review.  😄

jsquyres:
@hppritcha I'm good with this PR.

PR #4176, opened 2017-09-06
test: add bounds check for nfs test

bwbarrett:
The nfs test builds a list of entries for all mount points on the
system, and then checks each mount point for expected results
against the opal_path_nfs() call.  This test assumed that there
were less than 256 mount points on the system, which apparently
isn't true on some Cray systems.  This patch makes two changes:
add the proper bound check to avoid a segmentation fault and
double the list of mount points we'll return.

### DIFF
--- a/test/util/opal_path_nfs.c
+++ b/test/util/opal_path_nfs.c
@@ -132,7 +132,7 @@ void test(char* file, bool expect)
 
 void get_mounts (int * num_dirs, char ** dirs[], bool * nfs[])
 {
-#define MAX_DIR 256
+#define MAX_DIR 512
 #define SIZE 1024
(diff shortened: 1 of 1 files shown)


ggouaillardet:
i incidentally noticed ```dirs_tmp``` is not correctly allocated, it should be
```
dirs_tmp = (char**) calloc (MAX_DIR, sizeof(char*));
```
(instead of ```sizeof(char **)```)

should we ```realloc()``` instead of using a larger ```MAX_DIR``` ?
so we do not get bitten again by systems with more than 512 mount points

jsquyres:
#4179 was taken over this one.  Closing.

PR #4179, opened 2017-09-06
opal_path_nfs: ensure arrays are always long enough

jsquyres:
This PR is an alternative to #4176.  Both PRs solve the problem -- this one just allows us to check all mount points (even if there are more than 512).  Which may seem overkill.  But what the hell -- isn't overkill the Open MPI Way?

(specifically: only #4716 or this one should be merged -- not both)

-----


bwbarrett:
Let's go with this over #4176.

PR #4180, opened 2017-09-06
v3.0.x: fix opal NFS test

jsquyres:

(cherry picked from commit dee8cfbfd0b75dd2443ade13f5955498eb6e0c5d)


jsquyres:
This PR will fix the false failure CI on #4167.

jjhursey:
@jsquyres Note that the failure in PR #4167 seems to have fixed itself in the mean time.

PR #4181, opened 2017-09-06
pmix: configure.ac: fix setup order

nmorey:
AC_PROG_CC needs to be called before PMIX_SETUP_CORE to avoid fatal
error when using some versions of autools:
  configure.ac:173: error: AC_PROG_CC cannot be called after AM_PROG_CC_C_O

### DIFF
--- a/opal/mca/pmix/pmix112/pmix/configure.ac
+++ b/opal/mca/pmix/pmix112/pmix/configure.ac
@@ -155,16 +155,6 @@ PMIX_BASIC_SETUP
 AS_IF([test "$pmix_debug" = "1"],
       [CFLAGS="$CFLAGS -g"])
 
-############################################################################
-# Setup the core
-############################################################################
-
-# Setup the pmix core
-PMIX_SETUP_CORE([])
-
-# Run the AM_CONDITIONALs
-PMIX_DO_AM_CONDITIONALS
-
 ####################################################################
 # Setup C compiler
 ####################################################################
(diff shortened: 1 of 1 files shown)


rhc54:
I'll pull this down to PMIx so it is captured there, then bring it back to OMPI. Thanks!

jjhursey:
OMPI Bot: test this please

rhc54:
At least in  PMIx master, this really breaks things - tons of warnings and errors emitted. Investigating alternatives

rhc54:
This is the heart of the problem, just so we record it:
```
AM_PROG_CC_C_O
This is an obsolescent macro that checks that the C compiler supports the -c and -o options together. Note that, since Automake 1.14, the AC_PROG_CC is rewritten to implement such checks itself, and thus the explicit use of AM_PROG_CC_C_O should no longer be required.
```

ibm-ompi:

 Gist: https://gist.github.com/e3ec74c0e32f76993e8f1dc3c93044ad

ibm-ompi:

 Gist: https://gist.github.com/732cb7de303657e73705cb48b2a3f356

ibm-ompi:

 Gist: https://gist.github.com/e5184b84332584455ae8d3428292f5e5

rhc54:
I have a fix coming that avoids the issues here - please see https://github.com/pmix/pmix/pull/503

rhc54:
I think the correct solution is in https://github.com/open-mpi/ompi/pull/4147

PR #4183, opened 2017-09-06
v2.x: opal_path_nfs: ensure arrays are always long enough

jsquyres:

(cherry picked from commit dee8cfbfd0b75dd2443ade13f5955498eb6e0c5d)


jsquyres:
@hppritcha This is good to go.  I don't have strong feelings as to whether this goes in v2.1.2 or not, but we should merge it so that we don't get false failures on CI on this branch.

PR #4191, opened 2017-09-07
orte/pmix: Always seed environment with global rank

jjhursey:
* Even if we are only launching one app context, we might call spawn
   later and the remote groups might want their global rank information.

### DIFF
--- a/orte/orted/pmix/pmix_server_register_fns.c
+++ b/orte/orted/pmix/pmix_server_register_fns.c
@@ -18,6 +18,7 @@
  *                         All rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


jjhursey:
We have a pmix client that wants to use the `{OPAL_}PMIX_GLOBAL_RANK` key in a dynamic application and noticed that we don't always set it. Forcing it always to be set fixed it for them. This patch might be more than what is needed, but thought it was a good place to start a discussion.

rhc54:
IIRC, it was only done to reduce the memory footprint when running at scale as we rarely have more than one app_context. It increases the footprint in such cases by 32 bytes/proc, so at large scale it is a significant impact. Mitigated by use of dstore, of course, but still measurable.

Alternatives might be to suggest that:

* they check for number of apps - if only one, then just set global_rank and friends to their app-level values

* they use an "optional" flag when trying to retrieve the global values. If the "get" fails, then set the value to the non-global one

Or I suppose we could do some of this under the covers...or maybe we just eat the footprint?

jjhursey:
I can reduce the footprint here, since they only need the `PMIX_GLOBAL_RANK`. I think the `PMIX_GLOBAL_RANK` should always be present, but the app context info can be optional. Let me draw up a smaller version of this PR with that change. 

The `PMIX_GLOBAL_RANK` is pmix server specific, so we really need it to be set by the server instead of inferred by the client.

The situation is that we do a `mpirun ./my-app` (1 app context) but then `my-app` calls `MPI_Comm_spawn` N times with different subgroups (each with 1 app context). Those subgroups need to determine their global rank in the universe. Since `PMIX_GLOBAL_RANK` is only set when we have more than one app context they get an error when accessing that key. I can see if I can get a simple reproducer if that helps.

rhc54:
I'm a little confused. If you comm_spawn, then the global rank of each proc is solely its rank within that comm_spawn'd job. It definitely doesn't span the different jobs - the global rank is the rank across multiple apps in the _same_ job. So how is it supposed to help this scenario?

jjhursey:
Ok let's try this one. Should increase the memory footprint by one integer value, which is an increase, but seems necessary.

rhc54:
nurts - my bad. Global rank does indeed span the jobs, not the apps, so it should indeed be exposed

PR #4192, opened 2017-09-08
orte/pmix: Always seed environment with global rank

jjhursey:
* Even if we are only launching one app context, we might call spawn
   later and the remote groups might want their global rank information.


jjhursey:
Ref PR 4191

hppritcha:
running a check to see if arm node is working:
bot:ompi:retest

jjhursey:
If the v3.0.0 door opens again (hopefully not), then we should get this in there. Otherwise we are fine waiting until v3.0.1.

PR #4196, opened 2017-09-11
Remove spurious AC_PROG_CC - checked and verified with AM 1.13.4

### LINKED ISSUE #4109: autogen.pl problems with PMIx 1.2.3

hppritcha:
fixes #4109

(cherry picked from commit 9c313859049a2981b7681f426be48184301d0b26)

### DIFF
--- a/opal/mca/pmix/pmix112/pmix/configure.ac
+++ b/opal/mca/pmix/pmix112/pmix/configure.ac
@@ -19,7 +19,7 @@
 # Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
Just to be clear - I have no idea if this by itself is enough to fix the problem. There were other things in the original changes that also touched this code path and may (or may not) have contributed to the fix.

hppritcha:
@rhc54 it fixes the problem

hppritcha:
not waiting for @jsquyres  review, merging.

PR #4197, opened 2017-09-11
osc/sm: do not require 64-bit atomic math

hjelmn:
This commit fixes an issue on 32-bit systems that required atomic
64-bit math in the active target path. This commit updates the code to
work correctly on 32-bit systems.

### DIFF
--- a/ompi/mca/osc/sm/osc_sm.h
+++ b/ompi/mca/osc/sm/osc_sm.h
@@ -18,6 +18,20 @@
 
 #include "opal/mca/shmem/base/base.h"
 
+#if OPAL_HAVE_ATOMIC_MATH_64
+
+typedef uint64_t osc_sm_post_type_t;
+#define OSC_SM_POST_BITS 6
+#define OSC_SM_POST_MASK 0x3f
+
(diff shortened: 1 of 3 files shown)


hjelmn:
@bwbarrett This should get things working again on ia32.

hjelmn:
opps. commit message has the wrong component name.

ompi/mca/osc/sm/osc_sm.h
```diff
@@ -18,6 +18,20 @@
...
+
+typedef uint64_t osc_sm_post_type_t;
+#define OSC_SM_POST_BITS 6
+#define OSC_SM_POST_MASK 0x3f
+
+#else
+
+typedef uint32_t osc_sm_post_type_t;
+#define OSC_SM_POST_MASK 0x1f
+#define OSC_SM_POST_BITS 5
```
bwbarrett: the pedantic part of me is irritate that the order of the constants is different between the two cases...

hjelmn: fixed.

PR #4199, opened 2017-09-12
Miscellaneous show_help updates

jsquyres:
- OOB TCP: be more explicit about version mismatches
- BTL TCP: be more explicit about connection errors
- show_help: improve show_help cascading message effects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -629,6 +629,7 @@ test/datatype/ddt_raw
 test/datatype/opal_datatype_test
 test/datatype/position_noncontig
 test/datatype/unpack_ooo
+test/datatype/unpack_hetero
 
 test/dss/dss_buffer
 test/dss/dss_copy
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2007-2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2007-2017 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2008      Sun Microsystems, Inc.  All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Laboratory
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC.  All rights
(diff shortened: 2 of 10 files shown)


jsquyres:
Yeah, I have no doubt that this won't apply to all situations.  The thought was that it could apply to at least some situations:

For example, for the OOB TCP show_help message I added in this PR, using the "disable"-like functionality is the difference between:

```
$ mpirun --host mpi001,mpi002 --mca btl tcp,vader,self ring_c
Open MPI detected a mismatch in versions between two processes.  This
typically means that you executed "mpirun" (or "mpiexec") from one
version of Open MPI on on node, but your default path on one of the
other nodes upon which you launched found a different version of Open
MPI.

Open MPI only supports running exactly the same version between all
processes in a single job.

This will almost certainly cause unpredictable behavior, and may end
up aborting your job.

  Local host:             savbu-usnic-a
  Local process name:     [[4366,0],0]
  Local Open MPI version: 4.0.0a1
  Peer host:              something
  Peer process name:      [[4366,0],2]
  Peer Open MPI version:  bogus
```

and

```
$ mpirun --host mpi001,mpi002 --mca btl tcp,vader,self ring_c
Open MPI detected a mismatch in versions between two processes.  This
typically means that you executed "mpirun" (or "mpiexec") from one
version of Open MPI on on node, but your default path on one of the
other nodes upon which you launched found a different version of Open
MPI.

Open MPI only supports running exactly the same version between all
processes in a single job.

This will almost certainly cause unpredictable behavior, and may end
up aborting your job.

  Local host:             savbu-usnic-a
  Local process name:     [[4366,0],0]
  Local Open MPI version: 4.0.0a1
  Peer host:              something
  Peer process name:      [[4366,0],2]
  Peer Open MPI version:  bogus

  HNP daemon   : [[15141,0],0] on node savbu-usnic-a
  Remote daemon: [[15141,0],2] on node mpi002

ORTE does not know how to route a message to the specified daemon
located on the indicated node:

  my node:   savbu-usnic-a
  target node:  mpi001

This is usually an internal programming error that should be
reported to the developers. In the meantime, a workaround may
be to set the MCA param routed=direct on the command line or
in your environment. We apologize for the problem.
```

jsquyres:
@rhc54 and I talked on the phone.  Seems like this is generally a good idea, but there's definitely some corner cases that it doesn't cover well (e.g., from the `orted` in a DVM).

Still need to do a little more testing / fixing on this branch; setting `opal_show_help_enable(false)` after the new TCP BTL show_help messages doesn't appear to work properly.  Investigating...

jsquyres:
@rhc54 I finally pushed some updates based on what we talked about before.  I ended up re-working the interplay between OPAL and ORTE show_help systems; I think it's a bit simpler now than it was previously.

@bosilca @bwbarrett @mohanasudhan I updated some TCP BTL user messages a bit.  Could you guys have a look?

rhc54:
I confess I'm still stuck on this one as it will clearly cause problems for the DVM. I just don't know how we can safely turn off all show_help messages just because some developer felt that theirs should be the last one out, and then leave it off from then on. Any kind of "run-thru" operation, not just the DVM, will be impacted.

Although it would be too big a change, you ideally would want something that was at least job-specific - i.e., don't emit any more show_help messages for this specific job. Not entirely sure how to get there without a lot of editing. I'm just not convinced this solution isn't going to cause more trouble than it solves.

At the very least, we need an option to turn it off so those projects that can't operate this way (e.g., the DVM, ULFM) can continue to function.

bosilca:
get_peer_name only works for IPv4. Most changes in the TCP BTL are superficial and might have some potential benefit.

I share @rhc54 concern regarding why are we doing this enable/disable thing?

rhc54:
We had planned to discuss this on today's telecon, but ran out of time. @jsquyres indicated he and I might iterate some more offline before bringing it back to the group.

Just to ensure we don't miss it, @gpaulsen would you please add it to next week's agenda?

bwbarrett:
I'm not sure where I fall on the show help change, but I definitely feel like we're combining too many things in this PR.  I'd much rather see the OOB/BTL messages in one PR and the cascade change in a second.  Otherwise, we lose the OOB/BTL changes on a revert if the cascade change causes impact we didn't expect.

hppritcha:
@jsquyres do you still want to work on this, or should we close?

jsquyres:
As cited above, there are problems with the concept of a "last" error.  I give up -- closing.

The TCP OOB/BTL commits were salvaged to #4942.

PR #4201, opened 2017-09-12
build: Bump VERSION to 3.0.1rc1

bwbarrett:

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -15,7 +15,7 @@
 
 major=3
 minor=0
-release=0
+release=1
 
 # greek is generally used for alpha or beta release tags.  If it is
 # non-empty, it will be appended to the version number.  It does not
@@ -24,7 +24,7 @@ release=0
 # requirement is that it must be entirely printable ASCII characters
 # and have no white space.
 
-greek=rc6
+greek=a1
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"


bwbarrett:
bot:ompi:retest

bwbarrett:
When was the last time we released an a1 tarball?

hppritcha:
@jsquyres we need your input.

hppritcha:
I'm confused.  Are we going to start rc's for 3.0.1 already?  Our convention in the past was to go back to a1 for the next cycle for a given release stream.

bwbarrett:
I don't really care.  It seems silly to set the version to 3.0.1a1, so that as soon as we decide to cut an rc, the first thing we do is set the VERSION file to 3.0.1rc1.  We only seem to release rc tarballs, so what's the point of setting the version to a1, when we know we'll never release such a version?

rhc54:
The nominal idea was that the nightly tarballs and the code branch would be labeled as "a1" to indicate they are not release candidates. We then switch the version to "rcN" when we are in a release candidate phase. It was just a way to ensure that anyone who clones the branch, or downloads the nightly tarball, understands that this isn't code that the community considers as ready for release.

I suppose that if we are truly going to an "always ready for release" mode, then the distinction becomes less clear.

jsquyres:
I confirm @rhc54's recollection of Why Things Are The Way They Were (go to `a` instead of `rc`).

@bwbarrett Don't forget that the version (including the greek) shows up in other places, even if it's not in the filename.  From last night's v3.0 snapshot:

```
$ pwd
/home/jsquyres/openmpi-v3.0.x-201709080343-fc852fe 
$ ompi_info | head
                 Package: Open MPI jsquyres@savbu-usnic-a Distribution
                Open MPI: 3.0.0rc6
  Open MPI repo revision: v3.0.0rc5-2-gfc852fe
   Open MPI release date: Unreleased developer copy
                Open RTE: 3.0.0rc6
  Open RTE repo revision: v3.0.0rc5-2-gfc852fe
   Open RTE release date: Unreleased developer copy
                    OPAL: 3.0.0rc6
      OPAL repo revision: v3.0.0rc5-2-gfc852fe
       OPAL release date: Unreleased developer copy
```

bwbarrett:
Sure, I get all that.  But does it make any sense?  We know we're not going to release an a1, right?  We know we're going to release 3.0.1rc1 as the next release, right?  Also, now that we're tagging releases before we build them and based on the previous night's MTT run, the cost of forgetting is fairly high, for something that I really see no advantage in doing.  But since this is clearly religious, i fixed the PR to be a1 and we'll deal with the inevitable screwup / delays when they come.

jsquyres:
I don't actually have a religious opinion here -- I was just pointing out the history.

FWIW:
* I grok the points @bwbarrett is making.
* But it's still a little weird (to me at least) to have a snapshot tarball with `rc` in the `ompi_info` version.

Should we do something else entirely?  I.e., should we change our scheme?

bwbarrett:
I'll only point out that every nightly snapshot of 3.0.0 since we released rc1 has had an rc? in the ompi_info version (https://github.com/open-mpi/ompi/blob/v3.0.x/VERSION).  As do the current 2.1.x tarballs (https://github.com/open-mpi/ompi/blob/v2.x/VERSION).

jsquyres:
@bwbarrett Yes, this has been a dichotomy in the way we have done this -- that for the duration of the RC process, all nightly snapshots have `rc` in them.  I don't disagree with you.  😄   That's why I'm asking: should we be doing something altogether different?

bwbarrett:
Let's talk about it on the devel call on Tuesday.  We'll do a1 for now so we can make progress.  But it would be helpful for everyone to think about what they want to have happen; discussing end results will be a lot easier than discussing mechanisms...

jsquyres:
👍

Issue #3546, opened 2017-05-18
'./configure --without-lsf' causes build to fail with "no/lib" error


pmillet:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
v2.1.1

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
release tarball: openmpi-2.1.1.tar.gz

### Please describe the system on which you are running

* Operating system/version: 
Scientific Linux 6 
* Computer hardware: 
various
* Network type: 
various
-----------------------------

## Details of the problem

When trying to build OpenMPI without LSF support using --without-lsf in configure the build will eventually fail with the following error:
```shell
make[2]: Entering directory `/build/millet/ext/CMSSW_9_1_X/20170515_1016/a/BUILD/slc6_amd64_gcc530/external/openmpi/2.1.1/openmpi-2.1.1/orte/mca/ess/lsf'
  CC       ess_lsf_component.lo
  CC       ess_lsf_module.lo
  CCLD     mca_ess_lsf.la
../../../../libtool: line 7489: cd: no/lib: No such file or directory
libtool:   error: cannot determine absolute directory name of 'no/lib'
make[2]: *** [mca_ess_lsf.la] Error 1
make[2]: Leaving directory `/build/millet/ext/CMSSW_9_1_X/20170515_1016/a/BUILD/slc6_amd64_gcc530/external/openmpi/2.1.1/openmpi-2.1.1/orte/mca/ess/lsf'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/build/millet/ext/CMSSW_9_1_X/20170515_1016/a/BUILD/slc6_amd64_gcc530/external/openmpi/2.1.1/openmpi-2.1.1/orte'
make: *** [all-recursive] Error 1
error: Bad exit status from /build/millet/ext/CMSSW_9_1_X/20170515_1016/a/tmp/rpm-tmp.Lu2Zvd (%build)
RPM build errors:
    Bad exit status from /build/millet/ext/CMSSW_9_1_X/20170515_1016/a/tmp/rpm-tmp.Lu2Zvd (%build)
```

A similar issue with a different set of configure options has been reported in https://github.com/open-mpi/ompi/issues/3447 (just with 'yes/lib' instead of 'no/lib').

ggouaillardet:
thanks for the report,
can you please give this patch a try ?
[...]

PR #4204, opened 2017-09-13
Fix --without-lsf and LSF in default search path

jjhursey:
* A fix for Issue #3546
 * Include a clarification to `OPAL_CHECK_WITHDIR`

### DIFF
--- a/config/opal_check_withdir.m4
+++ b/config/opal_check_withdir.m4
@@ -8,6 +8,7 @@ dnl                         reserved.
 dnl Copyright (c) 2008-2009 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
+dnl Copyright (c) 2017      IBM Corporation.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


jjhursey:
To all that review this:
>  I did not try to rework the bulk of the logic in the lsf configure check on purpose. I'm trying for a minimal distance change that fixes the problem. The fix is similar to what @ggouaillardet proposed in the original ticket with some minor tweaks.
> I did have to reformat a lot of the file since a good section of it had to be indented. I tried to break out those changes as best as I could.

config/opal_check_withdir.m4
```diff
@@ -19,17 +20,21 @@ dnl
...
+          [AC_MSG_RESULT([simple no (specified --without-$1)])],
+          [AS_IF([test "$2" = "yes" || test "x$2" = "x"],
+                 [AC_MSG_RESULT([simple ok (unspecified value)])],
+                 [AS_IF([test ! -d "$2"],
                         [AC_MSG_RESULT([not found])
-                         AC_MSG_WARN([Expected file $2/$3 not found])
+                         AC_MSG_WARN([Directory $2 not found])
                          AC_MSG_ERROR([Cannot continue])],
-                        [AC_MSG_RESULT([sanity check ok ($2)])]
+                        [AS_IF([test "x`ls $2/$3 2> /dev/null`" = "x"],
```
ggouaillardet: would ```test ! -e "$2/$3"``` do the trick here ?

jjhursey: Possibly. I didn't want to change the prior behavior, so I left it as is. The git diff shows the indentation change in an odd way :/

jsquyres: Can't remember why we initially went with `ls` instead of `test -e`.  Seems like a (slightly) odd choice, but `ls` works just fine.

rhc54:
@jjhursey Do we need to bring the change to opal_check_withdir over to PMIx?

jjhursey:
@rhc54 yeah we will. It's somewhat cosmetic, but does help clarify when looking at a snippet of configure output. I can do that once I merge this PR in and post the PRs to the various release branches - later this evening.

PR #4206, opened 2017-09-13
Add a new launcher "prun" for starting applications against the ORTE DVM.

rhc54:
Unlike "orterun", "prun" is a PMIx-only program that discovers the DVM connection instead of requiring that we explicitly provide it. Only build "prun" if PMIx v2.x is available.

This gets the DVM working again, but still is showing problems for multiple executions. I'll detail those in a separate issue. Thus, the DVM should still be considered "broken".


ibm-ompi:

 Gist: https://gist.github.com/b0d068e41261c43fad65953a01423f21

Issue #4212, opened 2017-09-14
OMPI/master is hanging with external PMIx 1.2


karasevb:
This problem reproduced for OMPI/master with PMIx v1.2 as external and `hello_c` program.
```
#0  0x00007ffff75dcefd in nanosleep () from /usr/lib64/libc.so.6
#1  0x00007ffff760db34 in usleep () from /usr/lib64/libc.so.6
#2  0x00007ffff2d5c656 in PMIx_Get (proc=0x7fffffffc840, key=0x7ffff7046a6f "", info=0x666560, ninfo=1, val=0x7fffffffc830) at src/client/pmix_client_get.c:97
#3  0x00007ffff2fb3d36 in pmix1_get (proc=<optimized out>, key=<optimized out>, info=<optimized out>, val=0x7fffffffc9d0) at pmix1x_client.c:486
#4  0x00007ffff7032020 in opal_obj_run_destructors (object=0x7ffff725ef40 <opal_value_t_class>) at ../../../opal/class/opal_object.h:461
#5  opal_hwloc_base_get_topology () at base/hwloc_base_util.c:347
#6  0x0000000000661150 in ?? ()
#7  0x0000000000602b98 in ?? ()
#8  0x00007ffff72c9cd5 in orte_ess_base_proc_binding () at base/ess_base_fns.c:103
#9  0x00007ffff3bc76af in rte_init () at ess_pmi_module.c:408
#10 0x00007ffff7284529 in orte_init (pargc=pargc@entry=0x0, pargv=pargv@entry=0x0, flags=flags@entry=32) at runtime/orte_init.c:273
#11 0x00007ffff7b46fc1 in ompi_mpi_init (argc=1, argv=0x7fffffffd8a8, requested=0, provided=0x7fffffffd650) at runtime/ompi_mpi_init.c:512
#12 0x00007ffff7b608db in PMPI_Init (argc=0x7fffffffd6ac, argv=0x7fffffffd6a0) at pinit.c:66
#13 0x00000000004009e1 in main ()
```

karasevb:
@artpol84

PR #4214, opened 2017-09-15
pmix: fixed immediate request for PMIx v1.2

karasevb:
This fixes a hang of immediate PMIx request. PMIx v1.2 does not support
the info key `PMIX_IMMEDIATE` that leads to hanging. Instead, the fix
uses the key `PMIX_OPTIONAL` which is supported for all versions of PMIx

Closes: #4212


rhc54:
Ouch - no, this must be _immediate_ or you will break OMPI in a variety of cases. What you need to do instead is fix the ext1x component to look for immediate and not go to the server.

karasevb:
@rhc54 Thanks. Will be updated.

karasevb:
@rhc54
Done. Could you look, please.

PR #4216, opened 2017-09-15
util: move graph solver from usnic to util

bwbarrett:
Cisco wrote a bipartite graph solver to properly solve
interface pair selection for usNIC.  Using the reachable
framework, the TCP BTL (and possibly the runtime network
code can use the graph solver to make more optimal pair
selection.  Jeff was happy to have the code more broadly
used, but didn't have time to do the move, hence this
commit.

There are a couple of minor changes to the code compared
to the usNIC version.  Obviously, the functions have
been renamed to match naming convention for their new
home.  Since it's easier to write unit tests for
util/ code, the unit tests have been made first class
tests run at "make check" time.  This last bit required
moving some of the definitions into a new header,
bipartite_graph_internal.h, so that they could be
included in both the library code and the test code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -681,3 +681,4 @@ test/util/opal_sos
 test/util/opal_path_nfs
 test/util/opal_path_nfs.out
 test/util/opal_bit_ops
+test/util/bipartite_graph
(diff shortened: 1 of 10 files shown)


bwbarrett:
The only change between what Jeff approved and 98c63af is fixing a typo in the commit message (missing a close paren).  I'll wait for CI to finish to push, but not going to wait for another review.

jsquyres:
👍

Issue #4208, opened 2017-09-13
Compile error when using both --disable-mca-dso and --with-devel-headers


nmorey:
On SUSE LEap 42.3, using either a git tarball of v3.0.0 or the packaged release from the site, build fails:
```
[  156s] make[5]: Entering directory '/home/abuild/rpmbuild/BUILD/openmpi-3.0.0/opal/mca/pmix/pmix2x/pmix/src'
[  156s] make[5]: *** No rule to make target '/home/abuild/rpmbuild/BUILD/openmpi-3.0.0/opal/libopen-pal.la', needed by 'libpmix.la'.  Stop.
```

Tried make without -j, same thing.

```
$ cd opal
$ make libopen-pal.la
[some stuff builds ...]
make: *** No rule to make target 'mca/pmix/pmix2x/libmca_pmix_pmix2x.la', needed by 'libopen-pal.la'.  Stop.
$ cd mca/pmix/pmix2x/
$ make libmca_pmix_pmix2x.la
[some stuff builds ...]
make: *** No rule to make target '/home/abuild/rpmbuild/BUILD/openmpi-3.0.0/opal/mca/pmix/pmix2x/pmix/src/libpmix.la', needed by 'libmca_pmix_pmix2x.la'.  Stop.
$ cd pmix/src/
$ make libpmix.la
[some stuff builds ...]
make: *** No rule to make target '/home/abuild/rpmbuild/BUILD/openmpi-3.0.0/opal/libopen-pal.la', needed by 'libpmix.la'.  Stop.
```

Configure is 
```
./configure --host=x86_64-suse-linux-gnu --build=x86_64-suse-linux-gnu \
--program-prefix= --disable-dependency-tracking --prefix=/usr \
--exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc \
--datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/lib \
--localstatedir=/var --sharedstatedir=/usr/com --mandir=/usr/share/man \
--infodir=/usr/share/info --disable-dependency-tracking \
--prefix=/usr/lib64/mpi/gcc/openmpi3 --exec-prefix=/usr/lib64/mpi/gcc/openmpi3 \
--bindir=/usr/lib64/mpi/gcc/openmpi3/bin --datadir=/usr/lib64/mpi/gcc/openmpi3/share \
[...]

PR #4218, opened 2017-09-16
Attempt vpath inside component

rhc54:
Refs #4208 

The current configure logic installs a "native" version of libpmix (and its associated headers) in the OMPI install area when --with-devel-headers is given. The rationale for providing this capability was to allow users to write/test PMIx-based applications and execute them using mpirun, ensuring that the two were compatible. This is viewed as a desirable, but temporary, situation until broader PMIx support is available.

What we would like to do is separate this installation request from --with-devel-headers as the latter really was intended for another purpose. This PR adds a new --enable-install-libpmix option for this purpose. However, the symbols in the resulting libpmix are prefixed so they aren't exposed to the application that just links against OMPI libraries. As a result, we have to link the output libpmix against libopen-pal, which causes problems in certain configurations as noted in the referenced issue.

One way to resolve this, as proposed by @jsquyres, would be to build the internal pmix library twice - once with prefixed symbols (used to construct the component), and once without prefixing the symbols (used for the libpmix installation). This would involve doing VPATH builds of the library included in the component directory.

This PR contains a checkpoint of the work done to-date. I have unfortunately been unable to get the idea to work as the VPATH configure consistently fails. It first appeared that OPAL_CONFIG_SUBDIRS wasn't correctly setting the path to the configure - it would go to the correct subdirectory, but would execute "./configure" instead of "../configure", therefore errorring out due to not finding configure.

I fixed that by creating a new OPAL_CONFIG_VPATH that was a little less generalized. It now executes the correct command, and appears to be in the correct directory - but it still reports failure.

The subdirectory for the VPATH build is being created. If I manually go to that directory and execute "../configure", everything works fine. So I am unclear as to the problem.

@ggouaillardet @jsquyres Additional eyes would be much appreciated!

### DIFF
--- a/config/opal_configure_options.m4
+++ b/config/opal_configure_options.m4
@@ -275,8 +275,8 @@ AC_ARG_ENABLE([dlopen],
                      Disabling dlopen implies --disable-mca-dso.
                      (default: enabled)])])
 if test "$enable_dlopen" = "no" ; then
-    enable_mca_dso="no"
-    enable_mca_static="yes"
+    enable_mca_dso=no
+    enable_mca_static=yes
     OPAL_ENABLE_DLOPEN_SUPPORT=0
     AC_MSG_RESULT([no])
 else
--- a/config/opal_mca.m4
+++ b/config/opal_mca.m4
@@ -11,7 +11,7 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2010-2016 Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2013-2014 Intel, Inc.  All rights reserved.
+dnl Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
@@ -173,6 +173,7 @@ AC_DEFUN([OPAL_MCA],[
     elif test "$enable_mca_dso" = "no"; then
         DSO_all=0
         msg=none
+        enable_dlopen=no
     else
         DSO_all=0
         ifs_save="$IFS"
--- a/contrib/platform/intel/bend/linux
+++ b/contrib/platform/intel/bend/linux
@@ -18,6 +18,7 @@ enable_cxx_exceptions=no
 enable_mpi_java=no
 enable_io_romio=no
 enable_contrib_no_build=libnbc
+enable_install_libpmix=yes
 with_memory_manager=no
 with_tm=no
 with_psm=no
--- a/opal/mca/pmix/pmix2x/configure.m4
+++ b/opal/mca/pmix/pmix2x/configure.m4
@@ -37,6 +37,10 @@ AC_DEFUN([MCA_opal_pmix_pmix2x_CONFIG],[
     opal_pmix_pmix2x_save_LDFLAGS=$LDFLAGS
     opal_pmix_pmix2x_save_LIBS=$LIBS
 
+    AC_ARG_ENABLE([install-libpmix],
+                  [AC_HELP_STRING([--enable-install-libpmix],
+                                  [Enable a native PMIx library and headers in the OMPI install location (default: disabled)])])
+
     AC_ARG_ENABLE([pmix-timing],
                   [AC_HELP_STRING([--enable-pmix-timing],
                                   [Enable PMIx timing measurements (default: disabled)])])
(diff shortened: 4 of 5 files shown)


ibm-ompi:

 Gist: https://gist.github.com/954d4fa3a9f7fdfff4a617917b93326b

ibm-ompi:

 Gist: https://gist.github.com/f5d491045fedcb673c998c07e101d757

ibm-ompi:

 Gist: https://gist.github.com/e6661c80c6f19e79c5117aea5242f933

ggouaillardet:
@rhc54 i will take a crack at it.

at first glance, if you are building two ```PMIx``` libraries, one embedded (with prefixed symbols) and one non-embedded (with non prefixed) symbols, how do you handle the include files (so tests can be built, and they do not use prefixed symbols) ?
also, don't you also need to build non embedded ```libevent``` and possibly ```hwloc``` in order to be able to use ```PMIx``` ?

ggouaillardet:
@rhc54 the issue you are facing is a ```cd``` is failing
also, as written, this cannot work if you ```configure``` in ```VPATH``` in the first place

the patch below should fix that, i am still in the process of testing it

```patch
diff --git a/config/opal_config_subdir.m4 b/config/opal_config_subdir.m4
index 789a7af..d2bf880 100644
--- a/config/opal_config_subdir.m4
+++ b/config/opal_config_subdir.m4
@@ -173,9 +173,9 @@ AC_MSG_NOTICE([OPAL configuring VPATH in $subdir_dir])
 #
 
 subdir_parent=`pwd`
-cd $subdir_dir/..
+cd $srcdir/$subdir_dir/..
 subdir_srcdir=`pwd`
-cd $subdir_dir
+cd $subdir_parent/$subdir_dir
 
 #
 # Construct the --cache-file argument
@@ -189,10 +189,10 @@ subdir_cache_file="/dev/null"
 # Invoke the configure script in the subdirectory
 #
 
-sub_configure="$SHELL '../configure'"
-AC_MSG_NOTICE([running $sub_configure $subdir_args --cache-file=$subdir_cache_file --srcdir=$subdir_srcdir --disable-option-checking])
+sub_configure="$SHELL '$subdir_srcdir/configure'"
+AC_MSG_NOTICE([running $sub_configure $subdir_args --cache-file=$subdir_cache_file --disable-option-checking])
 eval "$sub_configure $subdir_args \
---cache-file=\"\$subdir_cache_file\" --srcdir=\"$subdir_srcdir\" --disable-option-checking"
+--cache-file=\"\$subdir_cache_file\" --disable-option-checking"
 if test "$?" = "0"; then
 eval $subdir_success
 AC_MSG_NOTICE([$sub_configure succeeded for $subdir_dir])
```

rhc54:
@ggouaillardet thanks for looking at this! I agree that the CPPFLAGS are going to have to be adjusted for the embedded build.

I too am unsure how the libevent support is going to work for the non-embedded version. We build libevent as a static framework with prefixed symbols, and the whole thing is simply slurped into libopen-pal. Somehow, the non-embedded version has to link against non-prefixed libevent - maybe it should look for an external version? Probably need some head-scratching here.

ggouaillardet:
@rhc54 i finally revamped a bit my patch and added the necessary stuff so Open MPI can build.

i initially pushed the changes to your branch, but then a conflict with the ```master``` branch popped up, so i rebased and force-pushed the changes.
sorry for the late heads up.
feel free to squash the commits and force-push at any time.

ggouaillardet:
@rhc54 i updated the PR with some more fixes

i just spotted one more thing that needs to be addressed.
if we ```configure --enable-install-libpmix```, then ```PMIx``` plugins of the ```standalone``` version overwrite the ones from the ```embedded``` version.
i can see two options right now
- install the plugins in an other directory
- force ```--disable-dlopen``` for the standalone plugin

ggouaillardet:
@rhc54 @jsquyres there is an other issue with this. when running ```make distdir``` the generated tarball contains twice the ```PMIx``` sources in ```opal/mca/pmix/pmix2x/pmix/{embedded,standalone}``` instead of a single copy in ```opal/mca/pmix/pmix2x/pmix```. currently, i have no idea on how to solve this one.

rhc54:
Hmmm...what if we step back a little? If we simply add the --enable-install-pmix option, we can decouple the build/emit of libpmix from --with-devel-headers. This would avoid the issue in the original report, and make it even more unlikely that someone would select the "wrong" combination of flags. We could then check to see if the problem combination is given (I gather that would be --enable-install-pmix --disable-mca-dso) and error out with a helpful message if we see it.

Make sense?

ggouaillardet:
If we aim at solving the issue that was previously reported, the root cause is we test ```$enable_dlopen``` and ```--disable-dlopen``` implies ```--disable-mca-dso``` but not the other way around.
Unless i am missing a subtle difference, these options should be made equivalent, or we could simply forbid ```--disable-mca-dso```.

rhc54:
@ggouaillardet Please take a look - I believe I addressed your last comment as well as implemented the alternative solution.

rhc54:
@ggouaillardet I'm going to commit as this completely fixes the DVM as well. Please let me know what you think of the configure changes - we can adjust.

PR #4224, opened 2017-09-18
util: Fix graph allocation size

bwbarrett:
Fix an allocation bug that could occur on non-LP64 platforms.
match_edges_out is an array of integers representing the
edges of the graph (where vertices are ints), with two ints
for every edge.  The previous code allocated enough space
for num_dges * sizeof(int*), which happens to be the same
as num_edges * 2 * sizeof(int) on LP64 platforms, but would
be wrong on all other platforms.

Fixes: CID 1417754


bwbarrett:
@jsquyres, not sure if usNIC supports 32-bit builds, but if so, you probably want to back-port this change to the release branches.

PR #4225, opened 2017-09-18
Clean up reachable framework / components

bwbarrett:
The reachable framework was largely some code copied from usNIC by Jeff and Ralph, but never got finished.  This patch series makes the reachable framework usable, at least for the purposes of the TCP BTL.

After lots of discussion with Jeff, it became clear that the primary use for the reachable framework should be as input to the bipartite graph code also recently moved from usNIC to util/, so the interface for the framework has been tweaked slightly to better match the inputs of the graph solver.


opal/test/reachable/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+PROGS = reachable_weighted reachable_netlink
+
+all: $(PROGS)
+
+CC = ortecc
+CFLAGS = -g
+CXX = ortec++
+CXXFLAGS = -g
+FFLAGS = -g
```
jsquyres: Nit: are `FFLAGS` and `CXXFLAGS` necessary

opal/test/reachable/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+clean:
+	rm -f $(PROGS) *~
+
```
jsquyres: Nit: blank line at end of file.

opal/mca/reachable/netlink/reachable_netlink_module.c
```diff
@@ -13,12 +15,27 @@
...
+#include "opal/util/net.h"
 #include "opal/mca/reachable/base/base.h"
 #include "reachable_netlink.h"
+#include "libnl_utils.h"
+
+enum connection_quality {
+    CQ_NO_CONNECTION = 0,
+    CQ_DIFFERENT_NETWORK = 50,
+    CQ_SAME_NETWORK = 100
+};
```
jsquyres: You define the same values in `reachable_weighted.c` (and also the TCP BTL).  Should these be put in `reachable.h` and prefixed with `OPAL_REACHABILITY_` or somesuch?

bwbarrett: I thought about it, but I don't actually want them to be part of the public interface; it means we can't tweak the algorithms and have to live with those choices going forward.  Imagine a netlink version that counted router hops as part of the weight.  Or ping latency, or something like that.  Those would be hard if we baked the constants into the API.

If you look at the usNIC BTL (which is where we should be moving the TCP BTL), there's no need for these constants, because it's the relative weights that matter.  So in time, we'll be removing them from the TCP BTL.

opal/test/reachable/Makefile
bwbarrett: Sigh :)

bwbarrett:
we really need to get warnings to the point where "make > /dev/null" pops the new warnings easily...  Anyway, comments fixed.

PR #4231, opened 2017-09-18
VERSION: stupid mistake with revision vs age

hppritcha:
thank goodness for the release checklist!
Caught this while grinding through the backward compatibility part of the checklist.


jsquyres:
So just to be clear: the source code of all of these libraries changes, but none of the *interfaces* (that we care about) changed?

hppritcha:
that's correct.

PR #4236, opened 2017-09-19
reachable: remove framework

bwbarrett:
The reachable framework is not used in the 3.0.x series and should
have been removed before release.  Due to a bug in the configure
code for the netlink component, if the libraries are installed but
the headers are not, a compile error can result.  Rather than
clean up dead code, we decided on the call to just remove the
entire framework.


bwbarrett:
There's no master commit to match this PR, because we're going to properly fix master with #4225.

bwbarrett:
@hppritcha, can you RM review?

Issue #4211, opened 2017-09-13
Compile failure if libnl-3 is installed but libnl-route-3 is not


bwbarrett:
The netlink reachable component fails to compile if libnl-3 is installed but libnl-route-3 is not.  Ubuntu splits the two into different packages, without a dependency from libnl-3 to libnl-route-3.  It looks like `OPAL_CHECK_LIBNL_V3` fails but `OPAL_CHECK_LIBNL_V1` succeeds and has no further version checking in the configure macro, so we assume success and try to build the component.  But the netlink.h header is for libnl-3 and nothing good happens from there.

ggouaillardet:
#4242 if only half a fix. on ubuntu, ```reachable/netlink``` will not build. the real fix is to have it built since i believe everything needed is already there.

ggouaillardet:
on second thought, this is the real fix.
in order to build ```reachable/netlink```, the ```libnl-route-3-dev``` package has to be installed.

jsquyres:
@ggouaillardet Are you saying that https://github.com/open-mpi/ompi/blob/master/config/opal_check_libnl.m4#L177-L186 somehow isn't working?

ggouaillardet:
the code you pointed to is working as expected.

a few lines earlier, we ```OPAL_CHECK_PACKAGE([$2], ..., [nl_recvmsgs_report], ...)``` and upon success, this sets ```$2_LIBS```.
a few lines later, we invoke the code you pointed to, and it might fail (for example if only half the ```libnl3``` devel headers are installed on debian based distro.
in ```opal/mca/reachable/netlink/configure.m4``` of ```v3.0.x``` __only__ (e.g. not ```master``` and i noticed that a few minutes ago !), we
```
        AS_IF([test "$opal_reachable_netlink_LIBS" = ""],
              [opal_reachable_netlink_happy=0])
```
so if the second test failed, we need to reset ```$opal_reachable_netlink_LIBS``` (otherwise ```reachable/netlink``` component is set to be built, and building will fail.

bottom line, the logic changed on master, and the ```reachable``` framework is scheduled for removal in ```v3.0.x```, so i am fine reverting my commit on ```master``` and simply dropping my ```v3.0.x``` PR.

makes sense ?

jsquyres:
(more followup on #4243)

bwbarrett:
[...]

PR #4242, opened 2017-09-21
configury: do not use libnl-3 when it is half broken

ggouaillardet:
Refs. open-mpi/ompi#4211


bwbarrett:
I don't understand what you're trying to do; that doesn't feel like the right fix to me; resetting the LIBS variable only on success?

PR #4243, opened 2017-09-21
configury: do not use libnl-3 when it is half broken

ggouaillardet:
Fixes open-mpi/ompi#4211

(cherry picked from commit open-mpi/ompi@94747a1d283c249f92c5505076bac35fa5785585)

### DIFF
--- a/config/opal_check_libnl.m4
+++ b/config/opal_check_libnl.m4
@@ -301,7 +301,11 @@ AC_DEFUN([OPAL_CHECK_LIBNL_V3],[
     AS_IF([test $opal_libnlv3_happy -eq 1],
           [$2_LIBS="-lnl-3 -lnl-route-3"
            OPAL_HAVE_LIBNL3=1],
-          [$2_LIBS=""])
+          [# OPAL_CHECK_PACKAGE(...,nl_recvmsgs_report,...) might have set the variables below
+           # so reset them if libnl v3 cannot be used
+           $2_CPPFLAGS=""
+           $2_LDFLAGS=""
+           $2_LIBS=""])
 
    OPAL_VAR_SCOPE_POP
 ])
--- a/config/opal_check_libnl.m4
+++ b/config/opal_check_libnl.m4
@@ -1,6 +1,6 @@
 dnl -*- shell-script -*-
 dnl
-dnl Copyright (c) 2015-2016 Research Organization for Information Science
+dnl Copyright (c) 2015-2017 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl Copyright (c) 2017      Cisco Systems, Inc.  All rights reserved.
 dnl $COPYRIGHT$
@@ -300,7 +300,8 @@ AC_DEFUN([OPAL_CHECK_LIBNL_V3],[
     # If we found everything
     AS_IF([test $opal_libnlv3_happy -eq 1],
           [$2_LIBS="-lnl-3 -lnl-route-3"
-           OPAL_HAVE_LIBNL3=1])
+           OPAL_HAVE_LIBNL3=1],
+          [$2_LIBS=""])
 
    OPAL_VAR_SCOPE_POP
 ])


config/opal_check_libnl.m4
```diff
@@ -217,7 +217,8 @@ AC_DEFUN([OPAL_CHECK_LIBNL_V3],[
     # If we found everything
     AS_IF([test $opal_libnlv3_happy -eq 1],
           [$2_LIBS="-lnl-3 -lnl-route-3"
-           OPAL_HAVE_LIBNL3=1])
+           OPAL_HAVE_LIBNL3=1],
+          [$2_LIBS=""])
```
jsquyres: @ggouaillardet Can you explain how this fixes the problem?  From my reading of the code, `$2_LIBS` is not set before this point, so setting it to an empty string shouldn't do anything...?

bwbarrett: I asked a similar question on the master commit; I don't think this is right.

jsquyres:
@ggouaillardet Ah, per your https://github.com/open-mpi/ompi/issues/4211#issuecomment-331318900 comment, now I understand.  It *is* possible that `$2_LIBS` was set via the call to `OPAL_CHECK_PACKAGE`.

Shouldn't we also reset `$2_LDFLAGS` and `$2_CPPFLAGS` in this case?

And can you put a comment explaining this?  I'm sure that Future Jeff/Brian/Gilles will look at this code someday -- having a comment here explaining why clearing out these variables will be helpful.

config/opal_check_libnl.m4
jsquyres: @ggouaillardet Ah, per your https://github.com/open-mpi/ompi/issues/4211#issuecomment-331318900 comment, now I understand.  It *is* possible that `$2_LIBS` was set via the call to `OPAL_CHECK_PACKAGE`.

Shouldn't we also reset `$2_LDFLAGS` and `$2_CPPFLAGS` in this case?

And can you put a comment explaining this?  I'm sure that Future Jeff/Brian/Gilles will look at this code someday -- having a comment here explaining why clearing out these variables will be helpful.

ggouaillardet: i updated master (in open-mpi/ompi@e28cc8c9869d18c0710fc20c6d6f01611408ae28).
regarding the ```v3.0.x``` branch, shall i update the PR ? or are we simply going to remove the ```reachable``` framework ?

jsquyres: `reachable` has already been removed from v3.0.x.  It looks like `reachable` was the last one to use `OPAL_CHECK_LIBNL_V*`, but it looks like `OPAL_LIBNL_SANITY_*` is still used (e.g., it can detect problems between `libibverbs` and `libfabric`).  Hence, we can't just flat-out remove `opal_check_libnl.m4` from v3.0.x because we're still using part of it.

So I think we have a choice:

1. Update only the relevant parts of `opal_check_libnl.m4` that are used on v3.0.x.
   1. Optional: remove the part of `opal_check_libnl.m4` that are not being used on v3.0.x
1. Update all of `opal_check_libnl.m4` on v3.0.x (even the parts we're not using on v3.0.x).

I think the last option is the best option for long-term maintenance (vs., for example, the next time we need to touch `opal_check_libnl.m4` on v3.0.x, I don't want to have to try to remember/re-create *why* `opal_check_libnl.m4` is different on master than on v3.0.x).  Meaning: I think we should just update `opal_check_libnl.m4` on v3.0.x to match exactly what is on master.  Particularly when involving m4 code, the benefits outweigh the cost of a little dead code on v3.0.x.

@bwbarrett @hppritcha Do you have an opinion here?

bwbarrett: Even if it's some unused code, I'd prefer we make opal_check_libnl.m4 on v3.0.x match master.  Way easier maintenance down the line...

jsquyres:
@ggouaillardet Can you verify that your small patch here, plus e28cc8c makes v3.0 be equivalent to master (w.r.t. `config/opal_check_libnl.m4`)?  If so, I think we're good to go to merge this into v3.0.

ggouaillardet:
@jsquyres i cherry-picked the commit you mentioned, and now ```v3.0.x``` is up to date with ```master```

jsquyres:
Thanks @ggouaillardet

PR #4245, opened 2017-09-21
hwloc: do not build hwloc CUDA support if --without-cuda used (and also always disable hwloc GL and OpenCL support)

### LINKED ISSUE #4219: OpenMPI builds with CUDA and OpenCL dependency even with --with-cuda=no option
## Background information

Tested on version 2.1.1 from source tarball.

-----------------------------

## Details of the problem

I tried to build OpenMPI from source with the following script:

```shell
shell$ ./configure --enable-mpi-thread-multiple CC=clang-3.9 CXX=clang++-3.9 --prefix=/home/user/openmpi2 --without-cuda  --with-cuda=no --disable-vt
shell$ make -j8 all
shell$ make -j8 install
```
Then I checked mpiexec dependecies with ldd command and found, that there are still dependencies on libOpenCL.so and libcudart.so. Of course on the server without these libraries this program do not work.

jsquyres:
Fixes #4219.

@bgoglin FYI

@hppritcha This might be worth putting into v2.0.x...?

### DIFF
--- a/opal/mca/hwloc/hwloc2a/configure.m4
+++ b/opal/mca/hwloc/hwloc2a/configure.m4
@@ -118,6 +118,10 @@ AC_DEFUN([MCA_opal_hwloc_hwloc2a_CONFIG],[
     # ensure we are in "embedded" mode
     hwloc_mode=embedded
 
+    # GL and OpenCL OS devices aren't used in OMPI
+    enable_gl=no
+    enable_opencl=no
+
     # Per https://github.com/open-mpi/ompi/issues/4219, if
     # --without-cuda was specified, be sure to disable it in hwloc,
     # too.  Note that hwloc uses --disable-cuda, so just set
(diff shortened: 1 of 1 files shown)


jsquyres:
@hjelmn @bwbarrett Looks like the OS X CI test failed, but there's no stdout/stderr to indicate why.  It just has:

```
3. Running autotools on top-level tree

==> Remove stale files
==> Writing m4 file with autogen.pl results
==> Running autoreconf
Running: autoreconf -ivf --warnings=all,no-obsolete,no-override -I config
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal -I config --force --warnings=all,no-obsolete,no-override -I ./config
FATAL: command execution failed
...Jenkins backtrace...
```

Any idea what happened here?

jsquyres:
Just for the helluvit, do the CS thing and try to make it fail again, just to see if it was some kind of transient / networking error to the OS testing box...

bot:ompi:retest

bwbarrett:
The root cause was:

```
Caused: java.io.IOException: Backing channel 'JNLP4-connect connection from c-67-160-8-108.hsd1.wa.comcast.net/67.160.8.108:17678' is disconnected.
```

Looks like the connection between the builder and the controller died.  Not sure why this didn't auto-retry; add that to my todo list.

jsquyres:
@bwbarrett Ah, thanks for the explanation -- I missed that in the backtrace.

Issue #4248, opened 2017-09-21
Do not build CUDA support in hwloc when --without-cuda is specified


jsquyres:
* Originally reported in #4219 
* Master PR to fix the issue: #4245 

Needs to be applied to v2.x and v3.0.x.  It *may* be desirable to also apply this to v2.0.x.  @jsquyres and @hppritcha will discuss.

jsquyres:
@Mazay0 Thanks for reporting the original issue in #4219.  This issue is to track applying the fix to all the release branches.

jsquyres:
See #4257 for a follow-on commit that could be added to all these PRs.

hppritcha:
this fix has been back ported to all relevant branches. closing.

PR #4250, opened 2017-09-21
v2.x: disable hwloc CUDA support when relevant

jsquyres:
Refs #4248.

PR #4249, opened 2017-09-21
v2.0.x: disable hwloc CUDA support when relevant

jsquyres:
Refs #4248.

@hppritcha I'm not 100% convinced, but it *may* be desirable to commit this fix to the v2.0.x branch.

PR #4251, opened 2017-09-21
v3.0.x: disable hwloc CUDA support when relevant

jsquyres:
Refs #4248

### DIFF
--- a/opal/mca/hwloc/hwloc1117/configure.m4
+++ b/opal/mca/hwloc/hwloc1117/configure.m4
@@ -117,10 +117,36 @@ AC_DEFUN([MCA_opal_hwloc_hwloc1117_CONFIG],[
 
     # Per https://github.com/open-mpi/ompi/issues/4219, if
     # --without-cuda was specified, be sure to disable it in hwloc,
-    # too.  Note that hwloc uses --disable-cuda, so just set
-    # enable_cuda=$with_cuda to get the same value that was passed in
-    # via the top-level --with-cuda CLI option.
-    enable_cuda=$with_cuda
+    # too.  Note that hwloc uses --disable-cuda (i.e., a yes or no
+    # value), whereas we use --with-cuda here in Open MPI (i.e., a
+    # yes, no, or path value).  Need to translate appropriately.
+    #
+    # Set enable_cuda to yes if:
+    #
+    # 1. --with-cuda was specified (i.e., a human specifically asked
+    # for it)
+    # 2. --with-cuda=blah was specified (i.e., a human specifically
+    # asked for it)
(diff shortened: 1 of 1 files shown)


jsquyres:
@ggouaillardet Good catch on the disparity between `--with-cuda` and `--enable-cuda`.  I think this PR's code is probably *correct* as-is (for the `--with-cuda=/path` case), but probably only by accident.  I'll fix on master and bring over to the relevant PRs.

As for NVML: I don't see any relation from NVML to CUDA in the hwloc configury (in hwloc proper, I don't see "disable CUDA? Ok, we'll also disable NVML" like we have in the OMPI master hwloc 2a component `configure.m4`).  So -- another good catch by @ggouaillardet.  😄   Do you know why that stanza is there in master?  I see 8d7541f766dfa00768a8bf797437fd37fdde9262, but the commit message doesn't say *why* we disable NVML if CUDA support is disabled in OMPI.  ☹️   CUDA != NVML -- they don't share support libraries or something, do they?

ggouaillardet:
@bgoglin can you please enlight us on `CUDA` vs `NVML` and how Open MPI might benefit from `hwloc` being built with ```--enable-nvml``` ?
If i understand/remember correctly, `NVML` is a `CUDA` feature, but Open MPI does not use this info from a `hwloc` topology.

bgoglin:
NVML is not a CUDA feature, it's another NVIDIA library that can be used to query GPU information without actually using GPUs. IIRC, that can be used by RMs for allocating GPUs to jobs, and by administrators for configuring/querying stuff such as temperature, UUID, etc.
https://developer.nvidia.com/nvidia-management-library-nvml

I am pretty sure OMPI doesn't care about NVML at all. So you could disable it even when CUDA is enabled.

As long as OMPI doesn't look at hwloc GPU "osdevs", you can disable CUDA, etc.

Grep tells me that the only users of osdevs in OMPI master seem to be usnic and hwloc_base_util.c. The latter looks for KNC MICs and Network/OFED NICs. Nothing for GPUs.

sjeaugey:
We should disable CUDA and NVML in hwloc since we're not using it. If we want to use it in the future, we'll re-enable parts that are of interest only.

jsquyres:
@sjeaugey Am I understanding you correctly: we're not using CUDA functionality in hwloc at all, even when CUDA support is enabled in the Open MPI build?

jsquyres:
See #4257 for a follow-on commit that could be added to all these PRs.

jsquyres:
@ggouaillardet New commit added to this v3.0.x PR.

jsquyres:
bot:mellanox:retest

jsquyres:
@ggouaillardet Thanks!  Can you also review #4249 and #4250?

PR #4269, opened 2017-09-26
oshmem: introduced the definition `SHMEM_ALLTOALLS_SYNC_SIZE`

karasevb:
In accordance with the OSHMEM spec, this definition must be included in
the code.

Signed-off-by: Boris Karasev <karasev.b@gmail.com>

Closes #4268

PR #4271, opened 2017-09-26
oshmem: refactoring the definition of `SHMEM_ALLTOALLS_SYNC_SIZE`

karasevb:
This PR is refactoring of PR #4269.

Since the value of `SHMEM_ALLTOALLS_SYNC_SIZE` depends on `_SHMEM_BARRIER_SYNC_SIZE`as can be seen here: [scoll_basic_barrier.c](https://github.com/open-mpi/ompi/blob/master/oshmem/mca/scoll/basic/scoll_basic_barrier.c#L187-L189) where `pSync` defined as [`long pSync[SHMEM_ALLTOALL_SYNC_SIZE];`](https://github.com/open-mpi/ompi/blob/master/oshmem/shmem/man/man3/shmem_alltoall32.3in#L184), so `SHMEM_ALLTOALLS_SYNC_SIZE` should be equal with `_SHMEM_BARRIER_SYNC_SIZE`


artpol84:
@miked-mellanox @jladd-mlnx please double check

artpol84:
The main idea here is that for alltoall and alltoalls pSync object is simply forwarded to the barrier primitive https://github.com/open-mpi/ompi/blob/master/oshmem/mca/scoll/basic/scoll_basic_alltoall.c#L116:L118
and so the size of pSync for alltoall and alltoalls should be the same as for barrier.

Issue #4264, opened 2017-09-25
master: getting segv's in MPI_Bcast_c over TCP BTL


jsquyres:
In Cisco MTT runs (and manually, I'm getting a segv in MPI_Finalize with *many* of the Intel tests in Cisco MTT.  I'm honestly not sure where the error is *actually* occurring -- it looks like a double free of a proc in MPI_Finalize.

Here's an example from `MPI_Bcast_c` -- I see it when running on two servers with as low as NP=8/PPN=4:

```
Core was generated by `./MPI_Bcast_c'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000003d3dc7b81c in free () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x2aaaabae0480 (LWP 23389))]
(gdb) bt
#0  0x0000003d3dc7b81c in free () from /lib64/libc.so.6
#1  0x00002aaaaab2d973 in ompi_proc_destruct (proc=0x7131d0)
    at ../../ompi/proc/proc.c:90
#2  0x00002aaaaab2e849 in opal_obj_run_destructors (object=<optimized out>)
    at ../../opal/class/opal_object.h:462
#3  ompi_proc_finalize () at ../../ompi/proc/proc.c:409
#4  0x00002aaaaab331a5 in ompi_mpi_finalize ()
    at ../../ompi/runtime/ompi_mpi_finalize.c:464
#5  0x00000000004022a8 in main (argc=1, argv=0x7fffffffbe98) at MPI_Bcast_c.c:237
(gdb) print *proc
$1 = {super = {super = {super = {obj_class = 0x2aaaaaee2360 <ompi_proc_t_class>, 
        obj_reference_count = 0}, opal_list_next = 0x713a90, 
      opal_list_prev = 0x2aaaaaf08f10 <ompi_proc_list+16>, item_free = 1}, 
    proc_name = {jobid = 1814757377, vpid = 5}, proc_arch = 4291367424, 
    proc_flags = 32768, proc_convertor = 0x6b2a80, 
    proc_hostname = 0x7102a800 <error: Cannot access memory at address 0x7102a800>}, 
  proc_endpoints = {0x0}, 
  padding = "\000 Ra\000\000\000\000\000\001\000\000\000*\000"}
```

Somehow it seems like the `proc_hostname` memory is getting freed to early...?

jsquyres:
[...]

PR #4283, opened 2017-09-29
create the 'pool' framework

ggouaillardet:
Refs open-mpi/ompi#4264

### DIFF
--- a/opal/mca/pool/base/pool_base_open.c
+++ b/opal/mca/pool/base/pool_base_open.c
@@ -1,0 +1,77 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2017      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+
+#include "opal_config.h"
+
+#include "opal/constants.h"
+#include "opal/mca/mca.h"
+#include "opal/mca/base/base.h"
+#include "opal/mca/pool/pool.h"
+#include "opal/mca/pool/base/base.h"
+
+
+/*
+ * The following file was created by configure.  It contains extern
+ * statements and the definition of an array of pointers to each
+ * component's public mca_base_component_t struct.
+ */
+#include "opal/mca/pool/base/static-components.h"
+
+/*
+ * Globals
+ */
+opal_pool_base_component_2_0_0_t *opal_pool = NULL;
+
+
+/*
+ * Function for finding and opening either all MCA components, or the one
+ * that was specifically requested via a MCA parameter.
+ */
+static int opal_pool_base_open(mca_base_open_flag_t flags)
+{
+    mca_base_component_list_item_t *item, *next;
+    opal_pool_base_component_2_0_0_t *tmp;
+    int priority, highest_priority = 0;
+    int ret;
+
+    /* can only be zero or one */
+    OPAL_LIST_FOREACH(item, &opal_pool_base_framework.framework_components, mca_base_component_list_item_t) {
+        tmp = (opal_pool_base_component_2_0_0_t *) item->cli_component;
+        ret = tmp->query (&priority);
+        if (OPAL_SUCCESS != ret || priority < highest_priority) {
+            continue;
+        }
+
+        highest_priority = priority;
+        opal_pool = tmp;
+    }
+
+    OPAL_LIST_FOREACH_SAFE(item, next, &opal_pool_base_framework.framework_components, mca_base_component_list_item_t) {
+        if ((void *) opal_pool != (void *) item->cli_component) {
+            mca_base_component_unload (item->cli_component, opal_pool_base_framework.framework_output);
+            opal_list_remove_item (&opal_pool_base_framework.framework_components, &item->super);
+        }
+    }
+
+    /* open remaining component */
+    ret = mca_base_framework_components_open (&opal_pool_base_framework, flags);
+    if (ret != OPAL_SUCCESS) {
+        return ret;
+    }
+
+    /* All done */
+    return OPAL_SUCCESS;
+}
+
+/* Use default register/close functions */
+MCA_BASE_FRAMEWORK_DECLARE(opal, pool, "pool of strings", NULL, opal_pool_base_open, NULL,
+                           mca_pool_base_static_components, 0);
--- a/ompi/mca/bml/r2/bml_r2.c
+++ b/ompi/mca/bml/r2/bml_r2.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2008-2016 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2013      Intel, Inc. All rights reserved
  * Copyright (c) 2014      NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
+ * Copyright (c) 2014-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2014-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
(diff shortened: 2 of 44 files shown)


ggouaillardet:
:bot:mellanox:retest

ggouaillardet:
:bot:ompi:retest

ggouaillardet:
@jsquyres @rhc54 can you please have a look at this PR ?

opal/mca/pool/base/pool_base_open.c
```diff
@@ -0,0 +1,77 @@
...
+ * Globals
+ */
+opal_pool_base_component_2_0_0_t *opal_pool = NULL;
+
+
+/*
+ * Function for finding and opening either all MCA components, or the one
+ * that was specifically requested via a MCA parameter.
+ */
+static int opal_pool_base_open(mca_base_open_flag_t flags)
```
rhc54: Why did this require a custom component open routine?

ggouaillardet: i did quite some copy/paste from existing frameworks, so this could be unnecessary. i will double check that.

ggouaillardet:
@rhc54 just to be clear, the goal of the ```pool``` framework is not to fix a bug, but to provide a way of reducing the memory footprint.

- The ```trivial``` component mimic the current behavior (the same string is ```strdup```'ed several times).
- The ```strings``` component uses ref-counted ```opal_string_t```, so there is one object per unique string (e.g. hostname) on each MPI task.
- The next step is to implement a ```sm``` component so there is only one instance of a given hostname on a given host (e.g. the instance is in shared memory, and is shared between the MPI tasks and mpirun/orted).


As far as i am concerned, my main concern is that some backport/adjustement will be required when PR from master (with the ```pool``` framework) to a release branch without the ```pool``` framework.

If you do not expect a significant reduction of the memory footprint via this framework, then i am fine dropping it.

rhc54:
@ggouaillardet good point - I stand corrected 😄

bosilca:
I share @rhc54 feeling, frankly it's scary by its own size. We started as a discussion on how to prevent a segfault, and it evolved into a ref counted string framework. Adding the ```sm``` framework is indeed a huge improvement, but then we are basically replicating the key-value shared memory framework from PMIx.

This might not necessarily be a bad thing, if we extend it to encompass all MPI-needed infromation that can be shared among local processes. If PMIx cannot provide us with a sharing mechanism allowing us to limit the memory used per process to store business card information about the other processes, maybe we should have our own...

rhc54:
To be clear, PMIx _does_ provide that sharing mechanism. Others felt it would create too strong a dependency on PMIx, however, and so a duplicate one inside OMPI was desired.

Personally, I'm unconvinced - but I suppose we could add a pmix component to this new framework to cover that option?

bosilca:
Can we at least analyze that framework and if we decide that it match our goals slurp it in OPAL ? I would assume this will not be a problem with PMIx folks.

rhc54:
Let me explain a bit. PMIx doesn't have a similar framework. What it does have is the ability to return a pointer to storage when asked for data instead of a malloc'd copy of it. We can also refcount that storage, if necessary, to protect it - though that won't solve the case of someone calling "free" on a string.

What I had offered to do before was to simply modify the opal/pmix_get function so we fetch pointers to values instead of copies. We would then have to audit OMPI to ensure that those values were no longer "free'd", so it is still a bit of an effort.

bosilca:
@rhc54 this sounds like a lighter approach.

There is a significant amount of information that is duplicated among all local MPI processes. If we are serious about reducing memory requirements for OMPI that information must be stored into a shared location. So we might want to use a sharing framework for more than just the hostname.

ggouaillardet:
folks, let's not get distracted by the ```SIGSEGV``` issue.
it is fixed, and while investigating the issue, i found some room to reduce the memory footprint, and here we are. at this stage, the fixed bug is no more relevant.

note the ```pool/strings``` component reduces the memory footprint by doing what storage folks call deduplication. this is deduplication per process, and the ```sm``` module could do deduplication per node.
does ```PMIx``` do deduplication internally ?
we can uses this framework for virtually any null terminated string. but if we want to use it for any objects, and we still want to perform deduplication, then null terminated strings and binary objects should be handled separately.
```hwloc``` can use a shared memory segment to reduce the memory footprint, but it is mandatory that all processes (mpirun/orted and MPI tasks) ```mmap``` it at the very same address (so the pointers it might contain can be dereferenced). i do not know how big a limitation this is.

now, we cannot assume ```PMIx``` is built with the shared memory datastore.
i do not think this impacts the functionality, but this definitely impacts memory footprint.

If ```PMIx``` is using a shared memory datastore, then the ```pool/pmix``` component could be as simple
as a copy of the ```pool/trivial``` component, but the ```free``` function would be just a no-op.
(e.g. you do not have to worry about someone calling ```free()``` on a string)

my primary concerns with this approach (e.g. relying on ```PMIx```) are
- i do not think PMIx does deduplication
- we cannot use the ```pool``` framework before and after the ```pmix``` framework is opened/closed.

i did not investigate any of these yet, and i could be wrong and/or there could be no issue.

an other approach (but basically limited to ```proc_hostname```) would be to use an index (4 bytes integer seems enough to me) instead of a (likely 8 bytes) pointer, and then add an indirection to access the string hostname.

at this stage, i think the ```pool``` framework give us the flexibility to experiment various techniques in order to reduce the memory footprint. we might end up with one component outperforming the other,
and at that point, we could simply remove the framework and directly use the component.

**Comment by rhc54** [comment]
@ggouaillardet I honestly don't understand why you would say that PMIx doesn't de-duplicate the data. What we do is create one shared memory instance of all data on the node. We offer the option of providing pointers to that data as individual elements. So in the full use-case, there would only be one actual instance of any job-level or modex data on a node, and every process would have pointers to it.

I don't know how you can "de-duplicate" more than that 😄 

For direct-launch on systems that don't offer PMIx, then we have to duplicate the data anyway. Fortunately, the latest information is that these systems are going away in the not-too-distant future, at least on machines of any scale. So it might not be a bad option, though I still acknowledge the validity of concerns over dependencies.

**Comment by ggouaillardet** [comment]
@rhc54 currently, if you are running ```n``` tasks on one node, then on a given MPI task and at the MPI level
```c
OPAL_MODEX_RECV_VALUE_OPTIONAL(ret, OPAL_PMIX_HOSTNAME, &proc->super.proc_name,
                                   &hostname, OPAL_STRING);
```
invoked on all the ```ompi_proc_t``` might lead to the same hostname being ```strdup```'ed ```n``` times.
with the ```pool/strings``` component, there would be only one (refcounted) string (e.g. deduplication)

in this scenario (e.g. ```n``` tasks running on a given node), how many strings are there for the hostname inside (the shared memory datastore of) ```PMIx``` ?
if there is only one, then that could not be better, but if the same hostname is stored ```n``` times (e.g. once per task), then there is room for improvement (e.g. deduplication)

makes sense ?

rhc54:
@ggouaillardet Let me repeat what I said earlier - PMIx _can_ simply return a pointer to the memory instead of copying it. This is _not_ what we currently do. PMIx only stores what OMPI/ORTE tells it to store. We currently do _not_ store a minimal footprint set of data - instead, we push more info than is required because we know we are going to pull it a certain way.

ORTE currently stores a regex for the names of all the nodes in the job. This is expanded inside PMIx to an array of hostnames for the job. We do _not_ store a hostname for every process as that blows out the footprint.

If we truly want to minimize footprint, we could ask PMIx to retain just the regex, and then use the PMIx resolve APIs to ask that it create a hostname string for a particular node and return a pointer to it. Thus, subsequent resolve or modex recv requests would return the pointer to that string.

In that way, we only create strings for the hosts that we need, instead of doing so for every node in the job. Yes, it will involve some code changes on both sides - and we'd have to ensure we retain compatibility with non-PMIx opal/pmix components.

To be clear: I'm not advocating any particular option here. I'm just trying to clear up some apparent confusion over what PMIx does and doesn't do, and explain the PMIx-based option. If people want togo with this pool framework, then that's fine - I'm just clarifying an alternative..

rhc54:
We had planned to discuss this at today's telecon, but ran out of time.. @gpaulsen would you please put this on next week's agenda?

gpaulsen:
I added WIP-DNM until we discuss on next week's telcon

jsquyres:
@bosilca @rhc54 @ggouaillardet Is this PR going anywhere?  It's over a year old.

rhc54:
PMIx is working on returning a pointer, as described. I personally am not a fan of ref-counting strings, but I leave that up to you folks.

ggouaillardet:
This is going nowhere so let's simply drop that PR.

PR #4295, opened 2017-10-03
Btl tcp: Fix racing condition on simultaneous handshake

mohanasudhan:
Pushing George patch (Verified and tested locally)<https://github.com/bosilca/ompi/commit/a8f894185e0b803ed1ddb3421f55bac06d1c3b33>

Their is racing condition in TCP connection establishment
during simultaneous handshake. This PR handles the fix for
it.

Signed-off-by: Mohan Gandhi <mohgan@amazon.com>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -109,3 +109,5 @@ Thomas Naughton <naughtont@ornl.gov> <naughtont@ornl.gov>
 Geoffrey Paulsen <gpaulsen@us.ibm.com> <gpaulsen@users.noreply.github.com>
 
 Anandhi S Jayakumar <anandhi.s.jayakumar@intel.com>
+
+Mohan Gandhi <mohgan@amazon.com>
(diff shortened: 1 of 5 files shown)


mohanasudhan:
bot:ompi:retest

jsquyres:
Try running `make dist` yourself.

I think it's telling you that you need to add an entry for yourself in the `.mailmap` file.

mohanasudhan:
@jsquyres Ya I ran locally it's fine. 

According to the code https://github.com/open-mpi/ompi-scripts/blob/b26aef08e3f0244250f717ab2159d5fd3b8f7ea7/jenkins/open-mpi-build-script.sh#L179 , it won't run any make once it enters diskcheck block. how to I skip it ?

mohanasudhan:
Ya i was using `make distcheck` :)

mohanasudhan:
Ya you are right. Naming was different in one of my previous commit.

bwbarrett:
@mohanasudhan , you should change the commit order in `mohanasudhan:iss4131` so that the .mailmap change gets pushed first...  The Pull Request Checker will pass, but you then can't make distcheck on that commit, which is poor for bisecting.

mohanasudhan:
ah. Ya good catch. will reorder

bwbarrett:
Looks like this got caught in the MacOS failure of 10/2017.  Let's try that again...

bot:ompi:retest

PR #4304, opened 2017-10-04
Fix OFI MTL to recognize correct CQ empty scenario

aravindksg:
Currently, the progress function is incorrectly interpreting any error
value other than a positive value or -FI_EAVAIL to mean CQ is empty. This is
incorrect. CQ is empty only if fi_cq_read() call returned -EAGAIN error
code. Fixing that here

### DIFF
--- a/ompi/mca/mtl/ofi/help-mtl-ofi.txt
+++ b/ompi/mca/mtl/ofi/help-mtl-ofi.txt
@@ -1,6 +1,6 @@
 # -*- text -*-
 #
-# Copyright (c) 2013-2015 Intel, Inc. All rights reserved
+# Copyright (c) 2013-2017 Intel, Inc. All rights reserved
 #
 # $COPYRIGHT$
 #
@@ -8,3 +8,9 @@
 #
 # $HEADER$
 #
+[OFI call fail]
+Open MPI failed an OFI Libfabric library call (%s).This is highly unusual;
+your job may behave unpredictably (and/or abort) after this.
+  Local host: %s
+  Location: %s:%d
+  Error: %s (%zd)
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2016 Intel, Inc. All rights reserved
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved
  *
  * $COPYRIGHT$
  *
@@ -14,6 +14,7 @@
 #include "ompi/mca/mtl/mtl.h"
 #include "ompi/mca/mtl/base/base.h"
 #include "opal/datatype/opal_convertor.h"
+#include "opal/util/show_help.h"
 
 #include <rdma/fabric.h>
 #include <rdma/fi_cm.h>
@@ -79,13 +80,14 @@ ompi_mtl_ofi_progress(void)
                 assert(ofi_req);
                 ret = ofi_req->event_callback(&wc, ofi_req);
                 if (OMPI_SUCCESS != ret) {
-                    opal_output(ompi_mtl_base_framework.framework_output,
-                                "Error returned by request event callback: %zd",
-                                ret);
-                    abort();
+                    opal_output(0, "%s:%d: Error returned by request event callback: %zd.\n"
+                                   "*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).\n",
+                                   __FILE__, __LINE__, ret);
+                    fflush(stderr);
+                    exit(1);
                 }
             }
-        } else if (ret == -FI_EAVAIL) {
+        } else if (OPAL_UNLIKELY(ret == -FI_EAVAIL)) {
             /**
              * An error occured and is being reported via the CQ.
              * Read the error and forward it to the upper layer.
@@ -94,9 +96,11 @@ ompi_mtl_ofi_progress(void)
                                 &error,
                                 0);
             if (0 > ret) {
-                opal_output(ompi_mtl_base_framework.framework_output,
-                            "Error returned from fi_cq_readerr: %zd", ret);
-                abort();
+                opal_output(0, "%s:%d: Error returned from fi_cq_readerr: %s(%zd).\n"
+                               "*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).\n",
+                               __FILE__, __LINE__, fi_strerror(-ret), ret);
+                fflush(stderr);
+                exit(1);
             }
 
             assert(error.op_context);
@@ -104,16 +108,22 @@ ompi_mtl_ofi_progress(void)
             assert(ofi_req);
             ret = ofi_req->error_callback(&error, ofi_req);
             if (OMPI_SUCCESS != ret) {
-                opal_output(ompi_mtl_base_framework.framework_output,
-                        "Error returned by request error callback: %zd",
-                        ret);
-                abort();
+                    opal_output(0, "%s:%d: Error returned by request error callback: %zd.\n"
+                                   "*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).\n",
+                                   __FILE__, __LINE__, ret);
+                fflush(stderr);
+                exit(1);
             }
         } else {
-            /**
-             * The CQ is empty. Return.
-             */
-            break;
+            if (ret == -FI_EAGAIN) {
+                break;
+            } else {
+                opal_output(0, "%s:%d: Error returned from fi_cq_read: %s(%zd).\n"
+                               "*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).\n",
+                               __FILE__, __LINE__, fi_strerror(-ret), ret);
+                fflush(stderr);
+                exit(1);
+            }
         }
     }
     return count;
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2013-2016 Intel, Inc. All rights reserved
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved
  *
  * Copyright (c) 2014-2015 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015-2016 Los Alamos National Security, LLC.  All rights
@@ -14,6 +14,7 @@
 
 #include "mtl_ofi.h"
 #include "opal/util/argv.h"
+#include "opal/util/show_help.h"
 
 static int ompi_mtl_ofi_component_open(void);
 static int ompi_mtl_ofi_component_query(mca_base_module_t **module, int *priority);
@@ -364,9 +365,10 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                      hints,        /* In: Hints to filter providers            */
                      &providers);   /* Out: List of matching providers          */
     if (0 != ret) {
-        opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
-                            "%s:%d: fi_getinfo failed: %s\n",
-                            __FILE__, __LINE__, fi_strerror(-ret));
+        opal_show_help("help-mtl-ofi.txt", "OFI call fail", true,
+                       "fi_getinfo",
+                       ompi_process_info.nodename, __FILE__, __LINE__,
+                       fi_strerror(-ret), ret);
         goto error;
     }
 
@@ -392,9 +394,10 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                     &ompi_mtl_ofi.fabric, /* Out: Fabric handle                 */
                     NULL);                /* Optional context for fabric events */
     if (0 != ret) {
-        opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
-                            "%s:%d: fi_fabric failed: %s\n",
-                            __FILE__, __LINE__, fi_strerror(-ret));
+        opal_show_help("help-mtl-ofi.txt", "OFI call fail", true,
+                       "fi_fabric",
+                       ompi_process_info.nodename, __FILE__, __LINE__,
+                       fi_strerror(-ret), ret);
         goto error;
     }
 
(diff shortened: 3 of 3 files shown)


jsquyres:
ok to test

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -110,10 +110,15 @@ ompi_mtl_ofi_progress(void)
...
-             * The CQ is empty. Return.
-             */
-            break;
+            /* CQ is empty, return */
+            if (ret == -FI_EAGAIN) {
+                break;
+            } else {
+                opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                                    "%s:%d: fi_cq_read failed: %s(%zd)\n",
+                                    __FILE__, __LINE__, fi_strerror(-ret), ret);
```
jsquyres: This message should be output via `opal_show_help()`, not `opal_output_verbose()`.

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -110,10 +110,15 @@ ompi_mtl_ofi_progress(void)
...
+                abort();
```
jsquyres: Doesn't `abort()` drop a corefile?  I don't think you want to do that -- especially at scale.

ompi/mca/mtl/ofi/mtl_ofi.h
aravindksg: Other parts of the code emit error messages in the same manner- either using opal_output_verbose() or opal_output() (in ompi_mtl_ofi_progress()). Would you prefer I convert all these instances to use opal_show_help() as well? Also, general clarification for future- when exactly should we use opal_show_help() vs. opal_output()/_verbose() ?

ompi/mca/mtl/ofi/mtl_ofi.h
aravindksg: Right. Will fix this. But this is again prevalent in other areas of the code as well (ompi_mtl_ofi_finalize()). Shall I fix these as well?

ompi/mca/mtl/ofi/mtl_ofi.h
jsquyres: If you have the cycles to do so, yes -- converting all user-visible errors to use `opal_show_help()` would be great.

In general, we've all be a bit sloppy about using `opal_output()` and `opal_output_verbose()` to show user errors.  I try to catch PRs that add *new* error messages via `opal_output[_verbose]()` and ask them to update to `opal_show_help()`.

In reality, you should classify each message and decide what you want:

1. If it's a warning or an error message that needs to *always* be displayed to the user, then it should use `opal_show_help()`.  Particularly if it can happen on multiple processes in the same job (because `opal_show_help()` de-duplicates and only shows the message once).
1. If it's a behind-the-scenes message that is really for developers (or when you're troubleshooting with a user), then `opal_output_verbose()` is sufficient.  I.e., you can show the message *sometimes*, but basically only when it is specifically requested by the user (by setting a verbosity level).

Does that make sense / help?

ompi/mca/mtl/ofi/mtl_ofi.h
jsquyres: Yes -- dropping corefiles is quite anti-social.  Especially if N Open MPI procs do it all at once -- it'll hammer networked filesystems, and can quickly cause you to run out of disk space.

ompi/mca/mtl/ofi/mtl_ofi.h
aravindksg: Yes, makes sense. Thanks for the explanation. Shall try to keep this rule of thumb in mind when adding more messages.

matcabral: Hi @jsquyres, in this particular case if reading fi_cq fails the mtl progress function will return an error that should be captured by opal_progress. So, wouldn't opal_progress() actually print the user friendly message?  My point is that there is not a clear line (I mean, I don't have it)  on what is important to the user vs developer for this case. It seems to me that the fi_errno is more like a developers thing. Clarification ? thanks!

ompi/mca/mtl/ofi/mtl_ofi.h
aravindksg: @jsquyres  Generic question- If we reach this point, it is to be interpreted as a fatal error and we need to stop execution right away. Is there a way to hint `opal_progress()` (which is higher up call stack) to "abort"? `opal_progress()` itself is void function and doesn't take any parameters. So not sure how to do this.

ompi/mca/mtl/ofi/mtl_ofi.h
jsquyres: Yeah, we're probably not very consistent across the code base in this regard (i.e., what to do if a component `progress()` function fails) -- there's probably no hard-and-fast rule here.

My $0.02: if you know the job is going to abort, `opal_show_help()` something useful in friendly language that a user can understand.

ompi/mca/mtl/ofi/mtl_ofi.h
jsquyres: I don't have the cycles to look at this immediately, but in the OB1 PML, there's a callback function for making the process fail (i.e., the BTL can invoke a callback back into OB1).  I don't know if the same mechanism exists in the CM MTL, but if it does, that's probably what you should use.

aravindksg: I looked at this and found the error callback function. But I didn't find a similar mechanism in CM PML. From looking at `opal_btl_usnic_exit()`, I see that it searches for a valid callback and if present, calls it. If not, we print message using stderr and simply `exit(1)` out of here. This seems a nice approach to stop execution without dumping a corefile. I can follow the same approach here. 

Thoughts?

jsquyres: 👍

aravindksg: Fixed this and used `opal_show_help()` for fail paths of some basic OFI calls.

aravindksg:
ping @jsquyres

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -581,38 +586,35 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
...
-        abort();
-    }
-    if (fi_close((fid_t)ompi_mtl_ofi.fabric)) {
-        opal_output(ompi_mtl_base_framework.framework_output,
-                "fi_close failed: %s", strerror(errno));
-        abort();
-    }
+    /* Close all the OFI objects */
+    if (ret = fi_close((fid_t)ompi_mtl_ofi.ep))
+        goto finalize_err;
```
jsquyres: Please use `{}` even around 1-line blocks.  It's the Open MPI Way.  😄 

(we only have a few style things that we try to be consistent about -- that's one of them)

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -79,13 +80,14 @@ ompi_mtl_ofi_progress(void)
...
                 ret = ofi_req->event_callback(&wc, ofi_req);
                 if (OMPI_SUCCESS != ret) {
-                    opal_output(ompi_mtl_base_framework.framework_output,
-                                "Error returned by request event callback: %zd",
-                                ret);
-                    abort();
+                    fprintf(stderr, "Error returned by request event callback: %zd.\n"
+                                    "*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).\n",
+                                    ret);
+                    fflush(stderr);
```
jsquyres: I would suggest `opal_show_help()` for all of these that you used `fprintf()` for, for 2 reasons:

1. `opal_output()` and `opal_show_help()` use a dedicated IO stream to send back messages to `mpirun`'s stderr.  Hence, they provide a level of atomicity that you don't get with normal `fprintf()` (because `fprintf()` contents are intercepted by the local `orted` via a pipe.  OS/pipe buffering may split single `fprintf()` messages into multiple strings, and therefore can (and do!) get munged together in `mpirun` stderr.  `opal_output()` and `opal_show_help()` guarantee to show their messages atomically (and without munging multiple messages together in `mpirun` stderr).
1. `opal_show_help()` de-duplicates error messages (i.e., only shows it once instead of N times, even if it is emitted multiple times from a single process, or even from multiple processes.

The de-duplication is surprisingly important.  It's the difference between seeing:

```
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
....etc.
```

and something like this:

```
Error returned by request event callback: 3.
*** The Open MPI OFI MTL is aborting the MPI job (via exit(3)).
16 more processes ...
```

ompi/mca/mtl/ofi/mtl_ofi_component.c
aravindksg: Sure, will fix.

ompi/mca/mtl/ofi/mtl_ofi.h
aravindksg: I can switch to `opal_show_help()`. But having an issue when I switch to this- I am not seeing the help text output as it seems the process terminates sooner than the output stream can be flushed. I am a little puzzled by this as when I use `opal_output()` or `opal_output_verbose()`, I can see the output message. Internally they all seem to call `output()` (in opal/util/output.c) but `opal_show_help()` isn't working as expected. Any other internal mechanism to flush the output stream that I am missing?

ompi/mca/mtl/ofi/help-mtl-ofi.txt
```diff
@@ -1,10 +1,15 @@
...
 #
 # Additional copyrights may follow
 #
 # $HEADER$
 #
+[OFI call fail]
+%s:%d: Call to OFI API %s failed.
+  Local host: %s
+  Error: %zd
+  Error Description: %s
```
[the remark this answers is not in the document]
aravindksg: %zd is required to interpret signed values of unsigned type size_t. (i.e), to print values of type ssize_t, we need this. "z" specifier itself seems to be C standard. I found some relevant info here: https://stackoverflow.com/questions/32916575/how-to-use-zd-specifier-with-printf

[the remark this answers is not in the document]
jsquyres: Ah.  We usually use the `PRIfoo` stuff, but I guess that won't work in a `show_help` message... would it...?  Hmm.  I guess I don't know offhand.

Regardless, my only question is: is it portable?  If it is, then we're good.  😄

ompi/mca/mtl/ofi/mtl_ofi.h
jsquyres: Ah; interesting point.  I wonder if we have a race condition here.

@rhc54 Can you shed any light on why `opal_output[_verbose]()` would tend to show up but `opal_show_help()` output would not when they are immediately followed by an `exit()`?  I'm guessing it's just a race condition; there's nothing explicit in `opal_output[_verbose]()` that ensures it is output before everything dies, right?

rhc54: opal_show_help goes thru PMIx log so we avoid making an OOB connection out of the app for aggregating the messages. Means you have to give PMIx time to transmit it to the daemon. 

opal_output just gets written thru a file descriptor and picked up by the daemon. fflush will force it out.

You should never call show_help or opal_output and then call exit! You are supposed to go thru finalize so the app gets an error. You never call exit from within a library (except when they call "MPI_Abort" and want us to do the exit for them).

jsquyres: @rhc54 Thanks for the explanation of outputs.

This is actually a case where the application is totally hosed, and we're down deep in a PML or BTL.  I.e., the network stack is completely hosed and there is nothing rational left to do other than exit.  We even have a few calls to `exit(3)` in the usNIC BTL and elsewhere (they usually end up calling the upper-layer PML abort function, but sometimes not even that is available, and so they call `exit(3)` directly).

ompi/mca/mtl/ofi/help-mtl-ofi.txt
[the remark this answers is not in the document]
aravindksg: "z" specifier is C standard and ssize_t is POSIX defined. That should meet the portability criteria?

ompi/mca/mtl/ofi/mtl_ofi.h
rhc54: Up to you - I only noted it because we have had complaints about doing that in the past. I know _we_ may believe there is nothing rational the user can do, but that assumes we really know that the app couldn't have done _anything_ useful (e.g., checkpoint partial results they have completed before exiting). Not sure the users out there necessarily agree with our assumption.

aravindksg: @jsquyres , @rhc54 : Any further guidance on this?
FWIW, I tried using `opal_show_help()` followed by `return()` error value back to `opal_progress()` . The effect is that we still try to make progress even after this fatal error scenario. Example:

mtl_ofi.h:129: Call to OFI API fi_cq_read failed.
  Local host: arav-rhel
  Error: -258
  Error Description: Operation not permitted in current state
# OSU MPI-CUDA Bandwidth Test v5.3.2
# Send Buffer on HOST (H) and Receive Buffer on HOST (H)
# Size      Bandwidth (MB/s)
1                       1.17
2                       1.45
...
[arav-rhel:47807] 85721 more processes have sent help message help-mtl-ofi.txt / OFI call fail
2097152              1515.20
4194304              2417.16

I don't think we should be trying to make progress here. We should bail out (as Jeff mentioned- there is nothing rational left to do) . Given the absence of a better way to allow app to exit gracefully, it seems `opal_output()` followed by an `exit()` seems the best we have for this scenario. Thoughts?

rhc54: Sounds like our architecture leaves you without a good option here, so I guess we just have to exit. Be prepared for some user complaints, though I'm not familiar enough with this code area to offer an alternative.

aravindksg:
@jsquyres, @rhc54  : Request to Review.

rhc54:
FWIW: calling sleep for a brief time is probably adequate. The transfer from client to server is fast and takes place in another thread, so it should be fine. I agree putting it in a common place would make things simpler and cleaner.

PR #4309, opened 2017-10-06
Coll hcoll dte extensions

vspetrov:
Adds new hcoll type mappings: lb/ub/wchar/bool (for legacy mpi apps).
Small type create optimization: if hcoll type creation fails we set zero dte on the ompi_datatype for subsequent queries to avoid repeating attempts to create hcoll type.


jladd-mlnx:
@vspetrov Please cherry pick into 3.0.1 and 3.1.x branches.

PR #4315, opened 2017-10-10
OSHMEM/CONFIGURE: Check for the presence of ibv_exp_reg_shared_mr.

alinask:
+ The sshmem verbs component will disqualify itself if this verb isn't
present on the build host.
+ In case where support was requested but not found, don't stop the
build - continue without this component.

### DIFF
--- a/oshmem/mca/sshmem/verbs/configure.m4
+++ b/oshmem/mca/sshmem/verbs/configure.m4
@@ -76,6 +76,26 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
     exp_reg_mr_happy=0
     AS_IF([test "$oshmem_have_mpage" = "3"],
             [
+                oshmem_verbs_save_CFLAGS="$CFLAGS"
+                CFLAGS="$CFLAGS -Wno-strict-prototypes -Werror"
+
+                AC_COMPILE_IFELSE(
+                         [AC_LANG_PROGRAM([[#include <infiniband/verbs_exp.h>]],
+                                           [[
+                                             struct ibv_exp_reg_shared_mr_in in_smr;
+                                             uint64_t access_flags = IBV_EXP_ACCESS_SHARED_MR_USER_READ   |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_USER_WRITE  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_WRITE |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_WRITE;
+                                             in_smr.exp_access = access_flags;
+                                             ibv_exp_reg_shared_mr(&in_smr);
+                                          ]])], [],
+                          [oshmem_verbs_sm_build_verbs=0])
+
+                CFLAGS="$oshmem_verbs_save_CFLAGS"
+
               AC_CHECK_MEMBER([struct ibv_exp_reg_shared_mr_in.exp_access],
                 [exp_access_happy=1],
                 [],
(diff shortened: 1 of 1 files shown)


alinask:
@yosefe @miked-mellanox @alex-mikheev @jladd-mlnx Please review.

oshmem/mca/sshmem/verbs/configure.m4
```diff
@@ -37,6 +37,49 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
...
+                                                                     IBV_EXP_ACCESS_SHARED_MR_USER_WRITE  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_WRITE |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_WRITE;
+                                             in_smr.exp_access = access_flags;
+                                             ibv_exp_reg_shared_mr(&in_smr);
+                                         ]])],
+                        [oshmem_have_exp_reg_shared_mr=1],
+                        [oshmem_have_exp_reg_shared_mr=0])
```
yosefe: can we set `oshmem_verbs_sm_build_verbs=0` directly from here?

oshmem/mca/sshmem/verbs/configure.m4
```diff
@@ -37,6 +37,49 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
     AC_SUBST([oshmem_verbs_LDFLAGS])
     AC_SUBST([oshmem_verbs_LIBS])
 
+    oshmem_have_exp_reg_shared_mr=0
+    AS_IF(
+        [test "$oshmem_verbs_sm_build_verbs" = "1"],
+        [
+            OSHMEM_LIBSHMEM_EXTRA_LDFLAGS="$OSHMEM_LIBSHMEM_EXTRA_LDFLAGS $oshmem_verbs_LDFLAGS"
```
yosefe: why is it needed to set these variables?

oshmem/mca/sshmem/verbs/configure.m4
alinask: yes, but this is how it is also done in the next check in this file so i kept the consistency

oshmem/mca/sshmem/verbs/configure.m4
```diff
@@ -37,6 +37,49 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
     AC_SUBST([oshmem_verbs_LDFLAGS])
     AC_SUBST([oshmem_verbs_LIBS])
 
+    oshmem_have_exp_reg_shared_mr=0
```
yosefe: looking at the m4 file, seems like this check(AC_COMPILE_IFELSE) should replace these lines:
```
            AC_CHECK_DECLS([IBV_EXP_ACCESS_ALLOCATE_MR,IBV_EXP_ACCESS_SHARED_MR_USER_READ],
                   [oshmem_have_mpage=3], [],
                   [#include <infiniband/verbs.h>])
```

because it's also possible that the "old" API of shared_mr is supported (oshmem_have_mpage=2) and then no need to disable the verbs component

hppritcha:
had to kill off armv8 testing.
bot:lanl:retest

alinask:
@yosefe updated.

oshmem/mca/sshmem/verbs/configure.m4
```diff
@@ -76,6 +76,27 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
     exp_reg_mr_happy=0
     AS_IF([test "$oshmem_have_mpage" = "3"],
             [
+                oshmem_verbs_save_CFLAGS="$CFLAGS"
+                CFLAGS="$CFLAGS $oshmem_verbs_save_CFLAGS -Wno-strict-prototypes -Werror"
```
yosefe: no need to add `$oshmem_verbs_save_CFLAGS`

oshmem/mca/sshmem/verbs/configure.m4
```diff
@@ -76,6 +76,27 @@ AC_DEFUN([MCA_oshmem_sshmem_verbs_CONFIG],[
...
+                                             uint64_t access_flags = IBV_EXP_ACCESS_SHARED_MR_USER_READ   |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_USER_WRITE  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_GROUP_WRITE |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_READ  |
+                                                                     IBV_EXP_ACCESS_SHARED_MR_OTHER_WRITE;
+                                             in_smr.exp_access = access_flags;
+                                             ibv_exp_reg_shared_mr(&in_smr);
+                                          ]])],
+                          [oshmem_verbs_sm_build_verbs=1],
```
yosefe: no need to set oshmem_verbs_sm_build_verbs=1 here, just set it to 0 in the 'else' case

alinask:
@yosefe updated, please review

PR #4322, opened 2017-10-10
Fix token separator for pmix envar files

npe9:
When pmix calls read_files() on pmix_mca_base_envar_files it attempts to tokenize on ',' when the list passed is separated by ':'. This patch fixes that problem.

### DIFF
--- a/opal/mca/pmix/pmix3x/pmix/src/mca/base/pmix_mca_base_var.c
+++ b/opal/mca/pmix/pmix3x/pmix/src/mca/base/pmix_mca_base_var.c
@@ -563,7 +563,7 @@ int pmix_mca_base_var_cache_files(bool rel_path_search)
     if (NULL != pmix_mca_base_envar_file_prefix) {
        resolve_relative_paths(&pmix_mca_base_envar_file_prefix, pmix_mca_base_param_file_path, rel_path_search, &pmix_mca_base_envar_files, ',');
     }
-    read_files (pmix_mca_base_envar_files, &pmix_mca_base_envar_file_values, ',');
+    read_files (pmix_mca_base_envar_files, &pmix_mca_base_envar_file_values, ':');
 
     if (0 == access(pmix_mca_base_var_override_file, F_OK)) {
         read_files (pmix_mca_base_var_override_file, &pmix_mca_base_var_override_values, PMIX_ENV_SEP);


rhc54:
hmmm...what we have matches the upstream version in opal. @hjelmn do you have any comment?

Here is the OPAL code:
```
    if (NULL != mca_base_var_file_prefix) {
       resolve_relative_paths(&mca_base_var_file_prefix, mca_base_param_file_path, rel_path_search, &mca_base_var_files, OPAL_ENV_SEP);
    }
    read_files (mca_base_var_files, &mca_base_var_file_values, OPAL_ENV_SEP);

    if (NULL != mca_base_envar_file_prefix) {
       resolve_relative_paths(&mca_base_envar_file_prefix, mca_base_param_file_path, rel_path_search, &mca_base_envar_files, ',');
    }
    read_files (mca_base_envar_files, &mca_base_envar_file_values, ',');
```
I see that one line uses OPAL_ENV_SEP, while the other is hard-coded to ','. Should they both be OPAL_ENV_SEP?

npe9:
@rhc54 Something's passing a ':' separated list to it (it might be the list maker that's the problem though). This is leading to a segfault because fopen(...,"r") has undefined behavior if the file doesn't exist. I also propose adding access(.., F_OK) checks to all the fopen(..., "r") calls so ompi fails gracefully as well.

jjhursey:
test this please (this phrase will trigger our CI full testing)

rhc54:
No issues with the proposal - I'd just like to get @hjelmn's input because it sounds like there is a bug upstream (or else this code is actually intended to behave this way, and you are using the option incorrectly)

jjhursey:
bot:ibm:retest

npe9:
@rhc54 no worries there at all. The real culprits are the fopens, but this particular problem was what led me to them.

rhc54:
I looked a little more at this, and I am becoming convinced this is an error in OPAL that has been passed down to us. In both cases, we are asking users for a list of entries - yet in one case, the separate is OPAL_ENV_SEP (which is defined as ':') and a comma in the other. This seems very confusing and probably not what was intended.

@hjelmn any thoughts on this?

jsquyres:
FWIW: the original intent was that `PATH`-like values (i.e., lists of directories) should use `:` so that they would be analogous to `$PATH`.  Other lists should be `,`, because that's similar to natural language.  But  `mca_base_var_file_prefix` and `mca_base_envvar_file_prefix` look like they are lists of *files*, not *directories*.  So they should probably be delimited with `,`, not `:`.

npe9:
@jsquyres If that's the case then PMIX_ENV_SEP should be replaced with "," in ./opal/mca/pmix/pmix3x/pmix/src/mca/base/pmix_mca_base_var.c:439
I'll be happy to submit a pull request for that if that's what you guys think the best solution is.

rhc54:
Makes sense to me - I'm not that strong on what gets used, only that the two be consistent to avoid confusion. I concur with the original intent.

@npe9 It looks to me like we need to make the change in the opposite direction - i.e., we need to fix the list maker so it understands the logic described in the prior comment so it matches what the user might directly provide. We also need to make that change in the corresponding OPAL code.

Are you willing to tackle that? Or is this outside your scope?

npe9:
@rhc54 one sec and I'll submit the pull request for the list maker.

npe9:
@rhc54 alternative fix pushed.

npe9:
Now a duplicate of #4328

PR #4323, opened 2017-10-10
Move help text output regarding PSM2_CUDA environment variable

aravindksg:
The messages should be printed only in the event of CUDA builds and when
PSM2 MTL has actually been selected for use. To this end,
move help text output to component init phase.

Also use opal_setenv/unsetenv() for safer setting, unsetting of the environment
variable and sanitize the help text message.

### DIFF
--- a/ompi/mca/mtl/psm2/help-mtl-psm2.txt
+++ b/ompi/mca/mtl/psm2/help-mtl-psm2.txt
@@ -1,7 +1,7 @@
 # -*- text -*-
 #
 # Copyright (C) 2009. QLogic Corporation.  All rights reserved.
-# Copyright (c) 2013-2015 Intel, Inc. All rights reserved.
+# Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -47,5 +47,17 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 Message size %llu bigger than supported by PSM2 API. Max = %llu
 #
 [no psm2 cuda env]
-Using CUDA enabled OpenMPI but PSM2_CUDA environment variable is %s.
-This is not a recommended combination. If the application uses %s.
+Warning: Open MPI has detected that you are running in an environment with CUDA
+devices present and that you are using Intel(r) Ompi-Path networking. However,
+the environment variable PSM2_CUDA was not set, meaning that the PSM2 Omni-Path
+networking library was not told how to handle CUDA support.
+
+If your application uses CUDA buffers, you should set the environment variable
+PSM2_CUDA to 1; otherwise, set it to 0. Setting the variable to the wrong value
+can have performance implications on your application, or even cause it to
+crash.
+
+Since it was not set, Open MPI has defaulted to setting the PSM2_CUDA
+environment variable to 1.
+
+Local hostname: %s
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2006-2010 QLogic Corporation. All rights reserved.
  * Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
- * Copyright (c) 2013-2015 Intel, Inc. All rights reserved
+ * Copyright (c) 2013-2017 Intel, Inc. All rights reserved
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
@@ -28,6 +28,7 @@
 #include "opal/mca/event/event.h"
 #include "opal/util/output.h"
 #include "opal/util/show_help.h"
+#include "opal/util/opal_environ.h"
 #include "ompi/proc/proc.h"
 
 #include "mtl_psm2.h"
@@ -44,6 +45,10 @@
 static int param_priority;
 /* MPI_THREAD_MULTIPLE_SUPPORT */
 opal_mutex_t mtl_psm2_mq_mutex = OPAL_MUTEX_STATIC_INIT;
+
+#if OPAL_CUDA_SUPPORT
+static bool cuda_envvar_set = false;
+#endif
 
 static int ompi_mtl_psm2_component_open(void);
 static int ompi_mtl_psm2_component_close(void);
@@ -201,9 +206,6 @@ static int
 ompi_mtl_psm2_component_register(void)
 {
     int num_local_procs, num_total_procs;
-#if OPAL_CUDA_SUPPORT
-    char *cuda_env;
-#endif
 
     ompi_mtl_psm2.connect_timeout = 180;
     (void) mca_base_component_var_register(&mca_mtl_psm2_component.super.mtl_version,
@@ -228,30 +230,6 @@ ompi_mtl_psm2_component_register(void)
         param_priority = 40;
     }
 
-#if OPAL_CUDA_SUPPORT
-    /*
-     * If using CUDA enabled OpenMPI, the user likely intends to
-     * run with CUDA buffers. So, force-set the envvar here if user failed
-     * to set it.
-     */
-    cuda_env = getenv("PSM2_CUDA");
-    if (!cuda_env) {
-        opal_show_help("help-mtl-psm2.txt",
-                       "no psm2 cuda env", true,
-                       "not set",
-                       "Host buffers,\nthere will be a performance penalty"
-                       " due to OMPI force setting this variable now.\n"
-                       "Set environment variable to 0 if using Host buffers" );
-        setenv("PSM2_CUDA", "1", 0);
-    } else if (strcmp(cuda_env, "0") == 0) {
-        opal_show_help("help-mtl-psm2.txt",
-                       "no psm2 cuda env", true,
-                       "set to 0",
-                       "CUDA buffers,\nthe execution will SEGFAULT."
-                       " Set environment variable to 1 if using CUDA buffers");
-    }
-#endif
-
     (void) mca_base_component_var_register (&mca_mtl_psm2_component.super.mtl_version,
                                             "priority", "Priority of the PSM2 MTL component",
                                             MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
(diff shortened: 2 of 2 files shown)


jjhursey:
test this please (this phrase will trigger our CI full testing)

ibm-ompi:

jjhursey:
bot:ibm:pgi:retest

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -334,6 +312,10 @@ ompi_mtl_psm2_component_query(mca_base_module_t **module, int *priority)
 static int
 ompi_mtl_psm2_component_close(void)
 {
+#if OPAL_CUDA_SUPPORT
+    if (cuda_envvar_set)
+        opal_unsetenv("PSM2_CUDA", &environ);
```
jsquyres: Minor nit: please use `{}`, even for 1-line blocks.

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -387,6 +372,26 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
...
     }
 
+#if OPAL_CUDA_SUPPORT
+    /*
+     * If using CUDA enabled OpenMPI, the user likely intends to
+     * run with CUDA buffers. So, force-set the envvar here if user failed
+     * to set it.
+     */
+    cuda_env = getenv("PSM2_CUDA");
+    if (!cuda_env) {
```
jsquyres: I'll repeat what I said in the last PR: don't you want to check and see if there are CUDA devices present before you do this?  Just being compiled for CUDA support is not any kind of guarantee that there are CUDA devices present.

Indeed, if someone uses an Open MPI compiled with CUDA support (e.g., via OpenHPC) in an environment without CUDA devices, it will be quite confusing to them as to why they are getting help messages about CUDA support (because it won't matter at all).

Run-time sensing what devices are present is a big theme of Open MPI.

ompi/mca/mtl/psm2/help-mtl-psm2.txt
```diff
@@ -47,5 +47,19 @@ Unknown path record query mechanism %s. Supported mechanisms are %s.
 Message size %llu bigger than supported by PSM2 API. Max = %llu
 #
 [no psm2 cuda env]
-Using CUDA enabled OpenMPI but PSM2_CUDA environment variable is %s.
-This is not a recommended combination. If the application uses %s.
+Warning: Running an Open MPI build with CUDA support over Intel(r) Omni-Path
+without enabling CUDA support on PSM2 library. Open MPI has therefore defaulted
+to setting PSM2_CUDA=1. This may impact performance if NOT running CUDA aware
+applications. Set your environment with variable PSM2_CUDA equal to 1 to clear
+this message, or set it to 0 to hint PSM2 that no CUDA support is needed.
```
jsquyres: Minor tweak suggestions for this message (per below, I'm assuming you'll only emit this message when there are CUDA devices present):

> Warning: Open MPI has detected that you are running in an environment with CUDA devices present and that you are using Intel(r) Ompi-Path networking.  However, the environment variable PSM2_CUDA was not set, meaning that the PSM2 Omni-Path networking library was not told how to handle CUDA support.
> 
> If your application uses CUDA buffers, you should set the environment variable PSM2_CUDA to 1; otherwise, set it to 0.  Setting the variable to the wrong value can have performance implications on your application, or even cause it to crash.
>
> Since it was not set, Open MPI has defaulted to setting the PSM2_CUDA environment variable to 1.

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -387,6 +372,26 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
...
+        opal_show_help("help-mtl-psm2.txt",
+                       "no psm2 cuda env", true,
+                       ompi_process_info.nodename);
+        opal_setenv("PSM2_CUDA", "1", false, &environ);
+        cuda_envvar_set = 1;
+    } else if (strcmp(cuda_env, "0") == 0) {
+        opal_show_help("help-mtl-psm2.txt",
+                       "psm2 cuda env zero", true,
+                       ompi_process_info.nodename);
```
jsquyres: So you're showing a help message:

* If `PSM2_CUDA` was not set
* If `PSM2_CUDA` was set to 0

The 2nd message there is... a bit weird.  So even if I (a user) am doing what I am supposed to be doing (i.e., telling you that I am not using CUDA buffers in my application), you're going to emit a warning message.  If you really want to do that, feel free -- this is Intel's plugin.  But I think that's downright weird.

ompi/mca/mtl/psm2/help-mtl-psm2.txt
aravindksg: Ok, Will fix this

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: Intent was to inform user that _only_ host buffers will work with `PSM2_CUDA=0`. I'll remove it. Should be reasonable to assume user will already know that.

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: Ok, I'll add some CUDA device detection here. Can I simply issue  `lspci -nn | grep -i "10de"`

and check the return value? Any other suggestions?

jsquyres: No, that will not scale (imagine N processes on a server all executing that at once).

jsquyres: We literally *just* disabled hwloc CUDA detection in master (c341b5347509037c145f0c3b8fd990d1511e6908) per https://github.com/open-mpi/ompi/pull/4257#issuecomment-332900393.  @sjeaugey is this a good reason to turn it back on?

aravindksg: Added some CUDA device detection logic in the updated patch.

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: Fixed this

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -362,6 +345,12 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
     int	verno_major = PSM2_VERNO_MAJOR;
     int verno_minor = PSM2_VERNO_MINOR;
     int local_rank = -1, num_local_procs = 0;
+#if OPAL_CUDA_SUPPORT
+    int ret;
+    char *cuda_env;
+    glob_t globbuf;
```
jsquyres: You probably need to `#include <glob.h>` to guarantee that this will work.

Is PSM/PSM2 Linux-only?  I.e., do you need to add a test for `glob()` and/or `<glob.h>` to this component's `configure.ac`?

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -389,6 +378,27 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
...
     }
 
+#if OPAL_CUDA_SUPPORT
+    /*
+     * If using CUDA enabled OpenMPI, the user likely intends to
+     * run with CUDA buffers. So, force-set the envvar here if user failed
+     * to set it.
+     */
+    ret = glob("/sys/module/nvidia", GLOB_DOOFFS, NULL, &globbuf);
+    if (0 == ret ||  GLOB_NOMATCH == ret) {
```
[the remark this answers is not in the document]
aravindksg: Unless we set `GLOB_NOCHECK` flag, I think we can expect `globbuf.gl_pathv` to be empty if the pattern didn't match. (I verified this on a sandbox too). So, I'll modify this to check:
    if (globbuf.gl_argc > 0) {
        globfree(&globbuf)
    }

FYI- The current check I have was basically following the existing usage in lines 259, 264. Which itself seems to have been introduced in commit 1daa80d (plug a memory leak in ompi_mtl_psm2_component_open). I don't think this commit was cherry-picked to other branches though. If this is fixing a memory leak, should this commit be ported to v2.x, v3.0.x?

While at it, shall I modify the above usage to also initialize the entire struct to `{0}` and to `globfree()` only `if (globbuf.gl_argc > 0)`?

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -389,6 +378,27 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
         ompi_mtl_psm2_set_shadow_env (ompi_mtl_psm2_shadow_variables + i);
     }
 
+#if OPAL_CUDA_SUPPORT
+    /*
+     * If using CUDA enabled OpenMPI, the user likely intends to
```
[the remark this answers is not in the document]
aravindksg: sorry about that. I thought I fixed this.

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -362,6 +345,12 @@ ompi_mtl_psm2_component_init(bool enable_progress_threads,
     int	verno_major = PSM2_VERNO_MAJOR;
     int verno_minor = PSM2_VERNO_MINOR;
     int local_rank = -1, num_local_procs = 0;
+#if OPAL_CUDA_SUPPORT
+    int ret;
+    char *cuda_env;
+    glob_t globbuf;
+    globbuf.gl_offs = 0;
```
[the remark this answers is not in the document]
aravindksg: Sure, will do.

ompi/mca/mtl/psm2/mtl_psm2_component.c
aravindksg: `glob.h` is already included. (There is an existing usage of `glob()` in line 279). PSM2 is Linux only and `glob.h` is included in glibc-headers package. So, don't think we need a check for this in `configure.ac`

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -45,6 +46,10 @@ static int param_priority;
 /* MPI_THREAD_MULTIPLE_SUPPORT */
 opal_mutex_t mtl_psm2_mq_mutex = OPAL_MUTEX_STATIC_INIT;
 
+#if OPAL_CUDA_SUPPORT
+static int cuda_envvar_set;
```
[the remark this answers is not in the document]
aravindksg: Ok, will do that.

ompi/mca/mtl/psm2/mtl_psm2_component.c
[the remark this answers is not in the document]
jsquyres: I think the answer is "yes" to all your questions:

* Cherry pick as relevant to v2.x, v3.0.x, and v3.x (since this is fixing a [minor] bug).
    * Skip v2.0.x -- that series is effectively dead.  We're really only taking serious bug fixes to that series now (this minor memory leak is not serious enough).
* I like your idea of initializing the entire struct with `{0}` and only `globfree()`ing if `globbuf.gl_argc > 0`.

ompi/mca/mtl/psm2/mtl_psm2_component.c
jsquyres: Oops -- missed that.  You're right.

ompi/mca/mtl/psm2/mtl_psm2_component.c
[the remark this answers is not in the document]
aravindksg: Fixed this (Y)

jsquyres:
ok to test

aravindksg:
Thanks @rhc54

PR #4326, opened 2017-10-11
V3.0.x coll hcoll dte extensions

vspetrov:
V3.0.x of #4309


jladd-mlnx:
@hppritcha could someone merge this PR so that we can get it covered in our nightly MTT?

PR #4328, opened 2017-10-11
Fix path separator for PMIX config

npe9:
When creating a list of PMIX configuration files the list building code was using PMIX_ENV_SEP (':') instead of the idiomatic ','. This change fixes the list building and path parsing to match each other.

### DIFF
--- a/opal/mca/base/mca_base_var.c
+++ b/opal/mca/base/mca_base_var.c
@@ -401,7 +401,7 @@ int mca_base_var_cache_files(bool rel_path_search)
 #if OPAL_WANT_HOME_CONFIG_FILES
     asprintf(&mca_base_var_files, "%s"OPAL_PATH_SEP".openmpi" OPAL_PATH_SEP
              "mca-params.conf%c%s" OPAL_PATH_SEP "openmpi-mca-params.conf",
-             home, OPAL_ENV_SEP, opal_install_dirs.sysconfdir);
+             home, ',', opal_install_dirs.sysconfdir);
 #else
     asprintf(&mca_base_var_files, "%s" OPAL_PATH_SEP "openmpi-mca-params.conf",
              opal_install_dirs.sysconfdir);
@@ -517,7 +517,7 @@ int mca_base_var_cache_files(bool rel_path_search)
     if (NULL != mca_base_var_file_prefix) {
        resolve_relative_paths(&mca_base_var_file_prefix, mca_base_param_file_path, rel_path_search, &mca_base_var_files, OPAL_ENV_SEP);
     }
-    read_files (mca_base_var_files, &mca_base_var_file_values, OPAL_ENV_SEP);
+    read_files (mca_base_var_files, &mca_base_var_file_values, ',');
 
     if (NULL != mca_base_envar_file_prefix) {
        resolve_relative_paths(&mca_base_envar_file_prefix, mca_base_param_file_path, rel_path_search, &mca_base_envar_files, ',');
--- a/opal/mca/pmix/pmix3x/pmix/src/mca/base/pmix_mca_base_var.c
+++ b/opal/mca/pmix/pmix3x/pmix/src/mca/base/pmix_mca_base_var.c
@@ -436,7 +436,7 @@ int pmix_mca_base_var_cache_files(bool rel_path_search)
 #if PMIX_WANT_HOME_CONFIG_FILES
     ret = asprintf(&pmix_mca_base_var_files, "%s"PMIX_PATH_SEP".pmix" PMIX_PATH_SEP
                    "mca-params.conf%c%s" PMIX_PATH_SEP "pmix-mca-params.conf",
-                   home, PMIX_ENV_SEP, pmix_pinstall_dirs.sysconfdir);
+                   home, ',', pmix_pinstall_dirs.sysconfdir);
 #else
     ret = asprintf(&pmix_mca_base_var_files, "%s" PMIX_PATH_SEP "pmix-mca-params.conf",
                    pmix_pinstall_dirs.sysconfdir);


npe9:
Quick change to fix the path separator. I haven't tested this for the fopen problem but this should make the semantics match.

rhc54:
Looks good to me - can you add the corresponding change to opal/mca/base? I can backport this one to the PMIx master so it doesn't get lost.

npe9:
Sure, I'll do that this afternoon.

jjhursey:
bot:ibm:retest

jsquyres:
@rhc54 What is this going to do for backwards compatibility?

I know it's a slim-to-none-chance that this will affect anyone, but it *is* a change in how some MCA var values are interpreted -- does this need to wait for v4.0?

jsquyres:
Also, the commits are not signed off.

test this please

npe9:
@jsquyres I fixed the sign off on my end and pushed -f my local repo. Will it register on its own?

npe9:
@jsquyres Also to fully test this I will need to fix a couple of fopens (by adding access(..., F_OK)s in the OpenMPI source code. Right now if the list fails to parse or the configuration file does not exist OpenMPI will segfault because fopen(...,"r") assumes that the file it's opening exists.

jsquyres:
@npe9 I'm confused between this PR and #4322 -- are they duplicates of each other?

And yes, if you force push to your github fork branch, the PR will automatically update, and all the CI hooks will run again (including checking for signed off).

If you need a few more calls to `access()` to truly fix the issue, no problem.  This is in a non-performance-critical section of the code, so a few filesystem accesses shouldn't be a problem.

npe9:
@jsquyres They're duplicates. This one just contains a few changes that @rhc54 requested (mostly fixing the same problem outside of pmix). The fopen problem is actually global to OpenMPI and will require 72 changes (according to`grep -r 'fopen(.*"r")' . | wc -l`) as well as a patch submitted to libnl. Libnl tries to `fopen("/proc/net/psched", "r")` which causes it to segfault on some systems where the file doesn't exist (e.g. some Crays).

jsquyres:
@npe9 Ok.  You can close the PR that is now a duplicate.  FWIW: if you have changes to a PR, you can just (force) push to your fork's branch that the existing PR is based from, and the PR will auto-update.  No need to make a new PR.

For the `fopen()` thing, that sounds like an entirely different problem -- that should probably be a separate PR.

Libnl... sigh.  Unfortunately, we don't control that here in this repo.  Have you tried to submit a patch upstream?

npe9:
@jsquyres Not yet, but I will. It's not on my critical path (unlike this problem) because typically the systems I use that tickle the bug should be using '--with-alps' anyway. We should probably merge this PR *after* the fixing the fopen problem though. That way it will fail with "can't open path" rather than just segfaulting making it much easier to test.

jsquyres:
Ok, if they're both symptoms of the same problem, then it probably makes sense to make 2 commits on this single PR.

npe9:
@jsquyres Sorry, they're not. It's just that it's just more easy to fix *this* problem after you fix the fopens (which I'll address in a separate PR tomorrow).

jsquyres:
test this please

hppritcha:
bot:mellanox:retest

rhc54:
bot:mellanox:retest

rhc54:
don't know why this is hung up for so long

bot:mellanox:retest

rhc54:
@npe9 This is a real failure - the env_list mca param uses semi-colon delimiters and this change somehow breaks it. Can you please take a look? We would like to get this into v3.1, but this needs to be fixed first.

npe9:
@rhc54 Isn't this the wrong semantics? There was an earlier discussion with @jsquyres where the agreement was that (IIRC) env_sep is ',' and path_sep is ':'. This fix is trying to make the env_sep's consistent. Could it be that the mellanox test is using the wrong semantics? If you can walk me through where the test is failing I can look at it tonight.

rhc54:
If you click on "Details" on the Mellanox test line, and then "Console" on the very left side of the resulting page, you'll see the complete output from the tester. Here is the failed step:
```
18:54:51 + echo 'mca_base_env_list=XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E'
18:54:51 ++ /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/ompi_install1/bin/mpirun -mca pml ob1 -mca btl self,vader -np 2 -am /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/test_amca.conf /var/lib/jenkins/jobs/gh-ompi-master-pr/workspace/jenkins_scripts/jenkins/ompi/env_mpi
18:54:51 ++ grep '^XXX_'
18:54:51 ++ wc -l
18:54:51 + val=6
18:54:51 + '[' 6 -ne 10 ']'
18:54:51 + exit 1
18:54:51 + jenkins_cleanup
18:54:51 + echo 'Script exited with code = 1'
18:54:51 Script exited with code = 1
```
I'm not entirely sure why it is failing - the env_list code kinda loops thru various existing code snippets, so you may have to track thru it a bit to find where it crosses the lines you changed.

npe9:
@rhc54 is there any way to dump test_amca.conf and env_mpi?

npe9:
Now that I know the executable that's failing and the setup I'll try to reproduce the failure tonight.

npe9:
@rhc54 brief update: I reproduced the Mellanox jenkins command lines and grabbed their environment dumping program. Specifying `mca_base_env_list` manually using `-mca` (partially?) works with the command line:
 `nevans@mutrino:~/src/ompi.sep> mpirun -np 2 -mca mca_base_env_list "XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E" ./env_mpi 2>/dev/null | grep "^XXX"
XXX_A=1
XXX_B=2
XXX_A=1
XXX_B=2`

However using a file does not work:
```
nevans@mutrino:~/src/ompi.sep> cat test_amca.conf
mca_base_env_list=XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E
nevans@mutrino:~/src/ompi.sep> mpirun -np 2 -am test_amca.conf ./env_mpi  | grep "^XXX"
nevans@mutrino:~/src/ompi.sep>
```

I'll need to dig deeper into the parsing machinery to figure out exactly what's going wrong. What's your time frame for the release?

rhc54:
probably at least two weeks out, I'm guessing...thanks for digging into this!

npe9:
Okay, after more digging my initial hypothesis is that there's not actually a bug in my code, it's a problem with the mellanox script using `-am` to specify the conf file. Using `-am` I get the following output:
```
nevans@nid00140:~/src/ompi.sep> mpirun -np 2 -am test_amca.conf .//env_mpi
You're using the --am option. Please be advised that this option is deprecated;
you should use --tune instead. The --tune option allows one to specify mca
parameters as well as environment variables from within a file using the same
command line syntax (e.g. -mca var val -mca var "val" -x var=val -x var).
The --am option will be removed in a future release.
OMPI_MCA_mca_base_param_file_prefix=test_amca.conf
OMPI_MCA_pmix=^s1,s2,cray,isolated
PMIX_SERVER_TMPDIR=/tmp/ompi.nid00140.85819/pid.21678/0/0
```
If I switch the option to using "--tune" the test seems to work with the following output. 

```
nevans@nid00140:~/src/ompi.sep> mpirun -np 2 --tune test_amca.conf .//env_mpi
WARNING: The format of "mca_base_env_list" parameter is a delimited list of VAR=VAL or
VAR instances. By default, the delimiter is a semicolon: VAR1=VAL1;VAR2;VAR3=VAL3;...
You can set other via "mca_base_env_list_delimiter" parameter. If the delimiter is a
semicolon, the value of "mca_base_env_list" variable should be quoted to not interfere
with SHELL command line parsing. In the case where a value is not assigned to variable
VAR, the value will be taken from the current environment.
The following environment variable was not found in the environment:
  Variable:             XXX_C
  MCA variable value:   XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E
OMPI_MCA_mca_base_envar_file_prefix=test_amca.conf
OMPI_MCA_mca_base_env_list=XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E
OMPI_MCA_pmix=^s1,s2,cray,isolated
PMIX_SERVER_TMPDIR=/tmp/ompi.nid00140.85819/pid.21499/0/0
XXX_A=1
XXX_B=2
```
Is it better to fix the mellanox script?

npe9:
I submitted a pull request on the mellanox tests. cf. https://github.com/mellanox-hpc/jenkins_scripts/pull/64

npe9:
bot:mellanox:retest

npe9:
@rhc54 can you try rerunning the mellanox test?

rhc54:
Looks like everything now passes!

npe9:
@rhc54 maybe I'll need to check.

rhc54:
@hjelmn We made the change in PMIx - do you want this change in OPAL? Otherwise, please close.

Issue #4306, opened 2017-10-05
Show load errors is defaulting to true


rhc54:
In v3.x, and on master, the default value for the MCA param "mca_base_component_show_load_errors" is being set to "true". This means that people who install OMPI from a packager get load errors reported for every component whose library isn't installed.

IIRC, our policy was to default this to "false" for just that reason. Did someone change it by mistake?

jjhursey:
In v2.x and v2.0.x it is default to `true` as well. I thought this was always the policy to show the load errors by default - did that decision change recently?

jsquyres:
Yeah, this is a sticky issue -- there are definitely valid arguments for setting the default to both True and False.

Perhaps the least-worst compromise could be:
* Add a `configure` CLI argument that can set the default of `show_load_errors` (if it doesn't exist already)
* We advise downstream packagers that they might want to update their builds to set this value to False for the reasons @rhc54 cited.

This would still seem to give safety to the majority of installations out there (i.e., those who build Open MPI themselves with just the libraries that they want, and want to know immediately if a component fails to load because of a missing library, etc.), but give the downstream packagers (who tend to build Open MPI support with *everything* that they can) a viable option to not have OMPI bombard users with useless warnings.

Thoughts?

bosilca:
+1 on making the default to False. We can assume that whoever carefully build a minimalistic list of component know what she's doing, so it is not unreasonable to ask her (on the FAQ) to manually set the flag to True.

jsquyres:
@bosilca Keep in mind that building a minimalistic set of components usually just entails "./configure && make -j 32 && make install".  I.e., `configure` will automatically make a minimalistic set of components based on what it finds on the system.

My $0.02 is that it's more unusual to build Open MPI with a *maximal* set of components.

However, this is subjective -- I admittedly have no proof either way.

rhc54:
I'm not sure when the change occurred, but my point was only that it _has_ changed. I know because I have it set to true in my default MCA param file (in my platform files) precisely because, as a developer, I was running into problems with new components that weren't being loaded. So I unfortunately never saw when it changed.

I honestly don't believe the statement that "the majority" of installations are built from scratch - I'm not sure we have any data to support it. I was simply contacted by the OpenHPC folks because they are unexpectedly getting reports of the problem, and it is rather annoying to their users.

Perhaps one solution is to make it true by default on debug builds, and false otherwise? It is, after all, a debug output.

bosilca:
[...]

PR #4330, opened 2017-10-11
configure: add --en|disable-show-load-errors-by-default

jsquyres:
Give packagers a configure CLI option to set the value of the MCA variable `mca_base_component_show_load_errors`.

The `--disable` form of this option is intended for Open MPI packagers who tend to enable support for many different types of networks and systems in their packages.  For example, consider a packager who includes support for both the FOO and BAR networks in their Open MPI package, both of which require support libraries (`libFOO.so` and `libBAR.so`).  If an end user only has BAR hardware, they likely only have `libBAR.so` available on their systems -- not `libFOO.so`.  Disabling load errors by default will prevent the user from seeing potentially confusing warnings about the FOO components failing to load because `libFOO.so` is not available on their systems.

Conversely, system administrators tend to build an Open MPI that is targeted at their specific environment, and contains few (if any) components that are not needed.  In such cases, they might want their users to be warned that the FOO network components failed to load (e.g., if `libFOO.so` was mistakenly unavailable), because Open MPI may otherwise silently failover to a slower network path for MPI traffic.

I'm labeling this as both a bug and an enhancement.  Technically, it's an enhancement (because it's a new feature), but it enables functionality that prevents downstream packagers from having to patch Open MPI.  So from that perspective, you could consider it a bug.

And if we consider it a bug, we should probably consider bringing this to all release branches (or perhaps at least to v3.0.x and v3.1.x).

Refs #4306

### DIFF
--- a/README
+++ b/README
@@ -838,6 +838,33 @@ INSTALLATION OPTIONS
   are build as static or dynamic via --enable|disable-static and
   --enable|disable-shared.
 
+--disable-show-load-errors-by-default
+  Set the default value of the mca_base_component_show_load_errors MCA
+  variable: the --enable form of this option sets the MCA variable to
+  true, the --disable form sets the MCA variable to false.  The MCA
+  mca_base_component_show_load_errors variable can still be overridden
+  at run time via the usual MCA-variable-setting mechanisms; this
+  configure option simply sets the default value.
+
+  The --disable form of this option is intended for Open MPI packagers
+  who tend to enable support for many different types of networks and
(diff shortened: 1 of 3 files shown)


jsquyres:
I'm going to go ahead and merge this so that we can make forward progress on the issue, and start bringing it to whatever release branches we choose (v3.0.x and v3.1.x?).  I left the default alone -- it's still enabled by default.

But I agree: we can continue the conversation that if we choose to change the default, that can be a follow-on commit+PR.

PR #4331, opened 2017-10-12
optionally passive wait when the progress loop is idle for a while

ggouaillardet:
Add the new mpi_poll_when_idle and mpi_poll_threhold MCA parameters to
control if and when the progress loop should poll() when idle and
when polling should start.

The default is not to poll when idle.

Thanks Paul Kapinos for bringing this to our attention

### DIFF
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -17,7 +17,7 @@
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016      Research Organization for Information Science
+ * Copyright (c) 2016-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -63,6 +63,7 @@ bool ompi_have_sparse_group_storage = OPAL_INT_TO_BOOL(OMPI_GROUP_SPARSE);
 bool ompi_use_sparse_group_storage = OPAL_INT_TO_BOOL(OMPI_GROUP_SPARSE);
 
 bool ompi_mpi_yield_when_idle = false;
+int ompi_mpi_sleep_when_idle_threshold = -1;
 int ompi_mpi_event_tick_rate = -1;
 char *ompi_mpi_show_mca_params_string = NULL;
 bool ompi_mpi_have_sparse_group_storage = !!(OMPI_GROUP_SPARSE);
@@ -109,11 +110,19 @@ int ompi_mpi_register_params(void)
        exactly/under-subscribed, or 1 when oversubscribed */
     ompi_mpi_yield_when_idle = false;
     (void) mca_base_var_register("ompi", "mpi", NULL, "yield_when_idle",
-                                 "Yield the processor when waiting for MPI communication (for MPI processes, will default to 1 when oversubscribing nodes)",
+                                 "Yield the processor when waiting for communication (for MPI processes, will default to 1 when oversubscribing nodes)",
                                  MCA_BASE_VAR_TYPE_BOOL, NULL, 0, 0,
-                                 OPAL_INFO_LVL_9,
+                                 OPAL_INFO_LVL_6,
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &ompi_mpi_yield_when_idle);
+
+    ompi_mpi_sleep_when_idle_threshold = -1;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "sleep_when_idle_threshold",
+                                 "Sleep after waiting for communication too long",
+                                 MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                 OPAL_INFO_LVL_6,
+                                 MCA_BASE_VAR_SCOPE_READONLY,
+                                 &ompi_mpi_sleep_when_idle_threshold);
 
     ompi_mpi_event_tick_rate = -1;
     (void) mca_base_var_register("ompi", "mpi", NULL, "event_tick_rate",
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2006-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
- * Copyright (c) 2015-2016 Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  *
  * $COPYRIGHT$
@@ -26,6 +26,11 @@
 
 #ifdef HAVE_SCHED_H
 #include <sched.h>
+#endif
+#if HAVE_POLL_H
+#include <poll.h>
+#elif HAVE_SYS_POLL_H
+#include <sys/poll.h>
 #endif
 
 #include "opal/runtime/opal_progress.h"
@@ -66,6 +71,9 @@ static size_t callbacks_lp_size = 0;
 
 /* do we want to call sched_yield() if nothing happened */
 bool opal_progress_yield_when_idle = false;
+/* do we want to sleep if nothing happened for a while */
+int opal_progress_sleep_when_idle_threshold = -1;
+static int yield_count = 0;
 
 #if OPAL_PROGRESS_USE_TIMERS
 static opal_timer_t event_progress_last_time = 0;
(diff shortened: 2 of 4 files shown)


ggouaillardet:

opal/runtime/opal_progress.c
```diff
@@ -229,6 +236,15 @@ opal_progress(void)
         }
     }
 
+    if (opal_progress_poll_when_idle) {
```
jsquyres: Might want to make this an `OPAL_UNLIKELY`, since a) it won't be the common case, and b) it's in an extremely performance-sensitive part of the code.

opal/runtime/opal_progress.c
```diff
@@ -138,6 +143,8 @@ opal_progress_init(void)
                  opal_progress_event_flag));
     OPAL_OUTPUT((debug_output, "progress: initialized yield_when_idle to: %s",
                  opal_progress_yield_when_idle ? "true" : "false"));
+    OPAL_OUTPUT((debug_output, "progress: initialized poll_when_idle to: %s",
```
jsquyres: What does it mean to set both `poll_when_idle` *and* `yield_when_idle` to true?

I'm wondering if we should have a higher-level abstraction name -- i.e., use `yield_when_idle` to mean that the user intends to yield the processor when MPI is idle (regardless of whether we use `yield()` or some other mechanism).  Otherwise, we have 2 mechanisms that do (kinda) the same thing, and one of them may not be supported on all platforms.  I think it would be better to have one MCA param that intelligently tries to give up the processor when MPI is "idle".

opal/runtime/opal_progress.c
```diff
@@ -229,6 +236,15 @@ opal_progress(void)
...
 
+    if (opal_progress_poll_when_idle) {
+        if (events <= 0) {
+            if (yield_count++ > opal_progress_poll_threshold) {
+                poll(NULL, 0, 1);
+            }
+        } else {
+            yield_count = 0;
+        }
+    }
```
jsquyres: If, per the comment above, we end up squashing both mechanisms into the same MCA variable, you should probably squash these two blocks together (i.e., the `yield()` and `poll()` blocks) so that we only have to have a single `if` statement (probably should be with an `OPAL_UNLIKELY`) gating entrance into that block.

opal/runtime/opal_progress.c
```diff
@@ -66,6 +67,10 @@ static size_t callbacks_lp_size = 0;
 
 /* do we want to call sched_yield() if nothing happened */
 bool opal_progress_yield_when_idle = false;
+/* do we want to poll() if nothing happened for a while */
+uint32_t opal_progress_poll_threshold = 1000;
```
jsquyres: I'm guessing you chose this number fairly arbitrarily.

Any idea how it impacts real application performance?

Put differently: any idea how long it takes to go through 1,000 cycles of the progress engine in an "idle" MPI process?

My *guess* is that 1,000 is going to be too low (i.e., we'll go into `poll()` mode too quickly, which could impact latency-sensitive applications).

rhc54:
I'm no expert on ```poll```, but I am curious to know how calling poll with a NULL argument will impact the event library.

ggouaillardet:
@jsquyres i merged both mechanisms per your comment
```
mpirun --mca mpi_yield_when_idle true ...
```
will simply ```sched_yield()```, in order to start ```poll```ing after a while, just do
```
mpirun --mca mpi_yield_when_idle true --mca mpi_poll_when_idle_threshold <value> ...
```
with a positive value.

@rhc54 i am not sure i fully understand your concern.
as far as i understand ```poll(NULL, 0, 1)``` is just a way to sleep for 1 millisecond.
```usleep()``` could be used (```select()``` could be even used here), but i am not sure it is available on all platforms.
or are you saying we should return from ```opal_progress()``` asap and use ```libevent``` timeout instead ?

**Comment by rhc54** [comment]
Your understanding matches my own - it was your comments in this PR that caused my confusion. You seemed to imply that somehow we were using the threshold to start polling file descriptors, but that isn't what you were doing at all - you're just sleeping to cause the scheduler to kick us out. It was very confusing.

**Comment by ggouaillardet** [comment]
ok, i will do some rewording use ```sleep``` instead of ```poll``` and add a note on how ```sleep``` is implemented.

File: ompi/runtime/ompi_mpi_params.c [hunk 5]
```diff
@@ -115,6 +116,14 @@ int ompi_mpi_register_params(void)
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &ompi_mpi_yield_when_idle);
 
+    ompi_mpi_sleep_when_idle_threshold = -1;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "sleep_when_idle_threshold",
+                                 "Sleep after waiting for MPI communication too long",
```
jsquyres: It's weird to put "MPI" in the help message for an OPAL MCA var...  Can this be re-worded?

ompi/runtime/ompi_mpi_params.c
```diff
@@ -115,6 +116,14 @@ int ompi_mpi_register_params(void)
...
+                                 MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                 OPAL_INFO_LVL_9,
```
jsquyres: My $0.02: make this level 6.

ompi/runtime/ompi_mpi_params.c
ggouaillardet: this is basically a copy/paste of ```yield_when_idle```, i fixed both

ompi/runtime/ompi_mpi_params.c
ggouaillardet: done (and i updated ```yield_when_idle```)

ggouaillardet:
@jsquyres i share the same vision.
just to be clear, what should be the default (e.g. ```spin only``` vs ```spin then yield then poll```) ? should it depend whether we oversubscribe or not ?

you have a good point with respect to ```MPI_Test```, and i guess the same applies to ```MPI_Iprobe``` and ```MPI_Improbe``` (are we missing some more ?)

jsquyres:
@ggouaillardet Yes, perhaps the demarkation line should be exiting the MPI library.  I.e., when MPI_TEST (or MPI_IPROBE or MPI_ISEND or ...) returns, the counters -- or whatever measures of "contiguous" we use -- should be reset.

As for what the default should be, I'm not sure.  I have dim recollections of some vendor MPI touting the power efficiencies of doing spin-then-yield by default a while ago (which is a dubious claim at best -- if your program is doing nothing for so long such that you frequently get into "MPI can yield the processor without harming performance" scenarios, then your program is not efficient to begin with, and therefore any power efficiencies gained by spin-then-yield probably mean that you're only wasting *less* energy than you were before).

This is probably a topic best discussed by the community -- others may have direct experience with this kind of thing.  @bosilca @gpapaure @jjhursey @edgargabriel @artpol84 @jladd-mlnx @rhc54 ...etc. -- anyone have an opinion here?

ompi/runtime/ompi_mpi_params.c
```diff
@@ -109,12 +110,20 @@ int ompi_mpi_register_params(void)
        exactly/under-subscribed, or 1 when oversubscribed */
     ompi_mpi_yield_when_idle = false;
     (void) mca_base_var_register("ompi", "mpi", NULL, "yield_when_idle",
-                                 "Yield the processor when waiting for MPI communication (for MPI processes, will default to 1 when oversubscribing nodes)",
+                                 "Yield the processor when waiting for communication (for MPI processes, will default to 1 when oversubscribing nodes)",
```
jsquyres: You still mention MPI in here, and also mention oversubscription (but the code doesn't check for oversubscription (per comments, we're still discussing what the default should be -- I just want to make sure that we don't forget to update the help message when a decision is made).

rhc54:
I can't speak to the performance issue, but I have seen a vendor make that claim. We have repeatedly gotten questions raised on the mailing list when users are "surprised" to see 100% cpu utilization, and several of us have gone to significant lengths to explain why it isn't an issue over the years.

I'd say just make it the default to idle so we quit having to explain it, but I'm not by any means sold on that position.

bosilca:
1. This patch delays the low priority callbacks (which contributes to counting the ```events```) by pausing before giving them a chance to trigger.
2. It also fails to protect itself in multithreaded scenarios, unlike most of the surrounding code.
3. Why are we having the MCA parameters at the MPI level, so that we add accessors to OPAL so that we can force a non-coordinated behavior on the progress. Why not having everything at a consistent level, down in OPAL ?

hjelmn:
@ggouaillardet I have seen better performance when using nanosleep to put the current thread to sleep. Might be worth looking at that vs poll. The Linux implementation seems pretty fast.

artpol84:
@ggouaillardet what was the original issue? Can you provide the reference?
We have seen issues with Slurm daemons starving of CPUs when spinning in the direct modex case. In this case some processes may have all they need and go calculating while the remote procs may be asking for the EP and the local PMIx server is delayed to respond because it is pushed back by the app procs.

ggouaillardet:
@artpol84 please refer to the thread starting at https://www.mail-archive.com/devel@lists.open-mpi.org//msg20407.html

long story short, if we `while (...) sched_yield();`, then `top` reports 100% usage even if the system remains very responsive since the MPI app spends its time yielding. The goal of this PR is to (virtually) stop CPU usage when nothing is happening.
 in the case of MPI paraview (which is an interactive program), MPI tasks spend most of the time in `sched_yield()`, which can be confusing when running `top` output that states 100% usage.

AboorvaDevarajan:

lanl-ompi:

gpaulsen:
@bosilca do we want this for v5.0?

bosilca:
The idea is still of interest, but this PR is stale. If I summarize, we want to have a straightforward approach: after `X` unsuccessful polls we start yielding and then after another `Y` unsuccessful polls we nanosleep for a duration `Z`.

rhc54:
@ggouaillardet You have several PRs that date back 3-6 years - would it make sense for you to triage them and close the ones not worth rebasing, fixing, resubmitting for review, and finally committing?

hppritcha:
can this PR be closed?  this is almost 8 years stale now.

jsquyres:
This PR is old and stale, and would need to be fully refreshed.

Additionally, I'm a little unconvinced by https://github.com/open-mpi/ompi/pull/4331#issuecomment-1232161734 -- there needs to be a distinction between periodically calling `MPI_TEST` N times (for example) and blocking in something like `MPI_WAIT` (per https://github.com/open-mpi/ompi/pull/4331#pullrequestreview-69292966).

Issue #14, opened 2014-10-01
Allow F90 MPI_BUFFER_DETACH to return correct pointer


ompiteam:
The current F77 MPI_BUFFER_DETACH implementation does not return the detached buffer pointer to the caller -- it simply does not make sense to do this in F77 because a) you can't get it, b) pointer implementations between compilers seem to differ, and c) even among the F77 compilers that do support pointers, you can't compare or use the pointer in a meaningful way.  There are two precedents that support this interpretation: LAM/MPI and CT6 both do not return the pointer to F77 callers.

Oh, and users of buffered sends should be punished, anyway.  :-)

However, this is a problem for the F90 bindings, which are [mostly] layered on top of the F77 bindings.  In F90, you can manage memory much like C, so it does make sense to return the detached buffer though the F90 API.  Hence, we need to override the default MPI F90 interface for MPI_DETATCH_BUFFER and have a specific implementation that returns the buffer pointer to the caller.

Here's some nuggets of information that may be helpful from an e-mail exchange from us and a Fortran expert at Sun (Ian B.):

---

Terry's e-mail to Ian:

  In MPI there is a function pair to called MPI_Buffer_attach and MPI_Buffer_detach.  These are used by the application program to give the MPI library some buffer space to use for the buffered communications functions.  

  When you call MPI_Buffer_attach you pass it a pointer to a buffer that you want MPI to use.  In C when you call MPI_Buffer_detach you pass it a pointer to a pointer in which the MPI library returns to you the pointer to the buffer you passed it via the MPI_Buffer_attach.  For C I can see this being used if you don't keep around the pointer to the buffer and you want to free the buffer returned by MPI_Buffer_detach.

  My question is the above applicable to Fortran programs at all?  Could do something similar with Fortran (90-03) pointers?

---

Ian's response:

  Yes, one could do that with f90 pointers.  You would need an interface for the MPI_Buffer_\* routines, or else the pointer arguments won't be passed correctly.  Something like

```
  interface
    subroutine MPI_Buffer_attach(p)
      integer, pointer, intent(in) :: p(:)
    end subroutine
  end interface
  interface
    subroutine MPI_Buffer_detach(p)
      integer, pointer, intent(out) :: p(:)
    end subroutine
  end interface
```

  (I haven't actually tried compiling that, so caveat emptor.)

hppritcha:
closing.  if someone wants to reopen they can.

Issue #13, opened 2014-10-01
Heterogeneous Multi-port OpenIB fails various IMB tests


ompiteam:
Between PPC64 and X86-64 using two ports of Open IB we fail a number of IMB benchmarks. This does not occur on the 1.2 branch with heterogeneous patches. 

```

#----------------------------------------------------------------
# Benchmarking Allreduce 
# #processes = 2 
# ( 2 additional processes waiting in MPI_Barrier)
#----------------------------------------------------------------
       #bytes #repetitions  t_min[usec]  t_max[usec]  t_avg[usec]
            0         1000         0.07         0.17         0.12
            4         1000        17.23        17.24        17.24
            8         1000        18.78        18.79        18.78
           16         1000        19.19        19.21        19.20
           32         1000        21.32        21.34        21.33
           64         1000        25.07        25.09        25.08
          128         1000        34.13        34.15        34.14
          256         1000        49.14        49.17        49.15
          512         1000        71.47        71.52        71.50
         1024         1000       122.10       122.17       122.14
         2048         1000       219.91       220.04       219.97
         4096         1000       417.00       417.24       417.12
m in the input buffer 2 bytes
unpack there is still room in the input buffer 2 bytes
unpack there is still room in the input buffer 2 bytes
unpack there is still room in the input buffer 2 bytes
unpack there is still room in the input buffer 2 bytes

```

gpaulsen:
In current release branches v4.0.x, v4.1.x, the datatype engine has been reimplemented relative to this old issue.  Open-ib no longer exists in the v5.0.x branch.
We should retest IMB using two ports of openib on a v4.0.x.

Issue #12, opened 2014-10-01
Cannot launch from head node on Big Red due to oob issues


ompiteam:
Filing this mainly so we don't forget about it...

On Big Red we cannot always launch jobs from the head node to the remote nodes. This seems to be due to the oob not finding the right communication paths.

The networking on Big Red is a bit confusing. There are 3 networks:[[BR]]

1. A Myrinet network which is global to all compute nodes[[BR]]

2. A GigE network which is global[[BR]]

3. A GigE network which is local to each cabinet.

If the compute nodes are in the same cabinet as the head node, we use the cabinet GigE network and are fine. If we launch on a backend node, we find and use the myrinet network (or force it to use the global GigE network) and are fine.

However, if we launch from the head node to nodes which are not in the same cabinet, we do not automatically find the correct network and simply hang. I can get it to launch correctly if I pass "-mca oob_tcp_include eth3,eth1" (the global GigE interfaces on the head node and the compute nodes, respectively).

This doesn't seem to be an issue for others, and since one normally isn't supposed to launch jobs from the head node of Big Red, I'm putting this to 1.3.

jsquyres:
The Big Red machine no longer exists.

Issue #11, opened 2014-10-01
Better error message propagation and reporting when shared memory backing file does not get created


ompiteam:
When the shared memory backing file does not get created, the error code generated is the out of resource error code, which is not very descriptive of helpful when trying to track down problems.  Need to change the code to  be a bit more descriptive.

gpaulsen:
This is now handled by the PMIx layer in current release branches, and is much more robust.

PR #4335, opened 2017-10-13
Add access() calls to ensure fopen(...,"r") doesn't segfault

npe9:
One of the invariants of the fopen() function called with the "r"
option is that the file exist before the function is called. If
this is not the case behavior is undefined (in the case of glibc
on some Crays and Redhat machines this means you get a segfault
in fileno_unlocked()). This commit is a first attempt at ensuring
this invariant holds when calls to fopen() are made.

### DIFF
--- a/ompi/debuggers/dlopen_test.c
+++ b/ompi/debuggers/dlopen_test.c
@@ -65,12 +65,17 @@ static int do_test(void)
 
     /* Double check that the .la file is there that we expect; if it's
        not, skip this test. */
+    if (!access(full_filename, F_OK)) {
+      fprintf(stderr,
+              "File %s.la doesn't seem to exist; skipping this test\n",
+              full_filename);
+      exit(77);
+    }
     fp = fopen(full_filename, "r");
-    if (NULL == fp) {
-        fprintf(stderr,
-                "File %s.la doesn't seem to exist; skipping this test\n",
-                full_filename);
-        exit(77);
+    if (fp == NULL) {
+      fprintf(stderr, "File %s.la fopen failed; skipping this test\n",
+              full_filename);
+      exit(77);
     }
     /* We know the .la file is there, so read it, looking for the
        dlopen value.  If the dlopen value is '' (i.e., empty), then
--- a/ompi/debuggers/ompi_debuggers.c
+++ b/ompi/debuggers/ompi_debuggers.c
@@ -148,7 +148,7 @@ static void check(char *dir, char *file, char **locations)
         FILE *fp;
 
         /* Just try to open the file */
-        if (NULL != (fp = fopen(str, "r"))) {
+        if (access(str, F_OK) && (NULL != (fp = fopen(str, "r"))) {
             fclose(fp);
             opal_argv_append_nosize(&locations, file);
         }
--- a/ompi/mca/coll/tuned/coll_tuned_dynamic_file.c
+++ b/ompi/mca/coll/tuned/coll_tuned_dynamic_file.c
@@ -88,6 +88,10 @@ int ompi_coll_tuned_read_rules_config_file (char *fname, ompi_coll_alg_rule_t** 
         return (-3);
     }
 
+    if (!access(fname, F_OK)) {
+        OPAL_OUTPUT((ompi_coll_tuned_stream,"rules file does not exist [%s]\n", fname));
+        goto on_file_error;
+    }
     fptr = fopen (fname, "r");
     if (!fptr) {
         OPAL_OUTPUT((ompi_coll_tuned_stream,"cannot read rules file [%s]\n", fname));
--- a/ompi/mca/topo/treematch/treematch/tm_mapping.c
+++ b/ompi/mca/topo/treematch/treematch/tm_mapping.c
@@ -73,6 +73,12 @@ int nb_lines(char *filename)
   char line[LINE_SIZE];
   int N = 0;
 
+  if(!access(filename, F_OK)) {
+    if(tm_get_verbose_level() >= CRITICAL)
+      fprintf(stderr,"%s does not exist\n",filename);
+    exit(-1);
+  }
+
   if(!(pf = fopen(filename,"r"))){
     if(tm_get_verbose_level() >= CRITICAL)
       fprintf(stderr,"Cannot open %s\n",filename);
@@ -97,6 +103,11 @@ void init_mat(char *filename,int N, double **mat, double *sum_row)
   int i,j;
   unsigned int vl = tm_get_verbose_level();
 
+  if(!access(filename, F_OK)) {
+    if(tm_get_verbose_level() >= CRITICAL)
+      fprintf(stderr,"%s does not exist\n",filename);
+    exit(-1);
+  }
 
   if(!(pf=fopen(filename,"r"))){
     if(vl >= CRITICAL)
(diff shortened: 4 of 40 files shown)


npe9:
This code needs some pretty thorough review, both for style and to make sure I haven't forgotten anything.

Also the calls to `access()` depend on unistd.h and I expect some open-mpi environments might not support it.

npe9:
It's also worth noting that I've left hwloc alone. How is hwloc merged from its individual repo to its main tree?

rhc54:
We generally take hwloc release tarballs. However, at the moment, we have an early copy of HWLOC v2.0 from their nightly tarball. Your best bet would be to fork the hwloc repo and generate a pull request against their master.

rhc54:
BTW: are you sure about that existence requirement for fopen? I've been scanning the man pages, and I don't see that - indeed, there is a clear statement that the call will fail with ENOENT set in errno. Where are you seeing this requirement?

npe9:
@rhc54 You're right if gnu followed the posix standard.

From [fopen](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html)

> [CX] [Option Start] The character 'b' shall have no effect, but is allowed for ISO C standard conformance. [Option End] *Opening a file with read mode (r as the first character in the mode argument) shall fail if the file does not exist or cannot be read.*

The [gnu documentation](https://www.gnu.org/software/libc/manual/html_node/Opening-Streams.html#Opening-Streams) however makes this assumption:

> ‘r’ Open an *existing* file for reading only.

This means that any time OpenMPI does an `fopen(...,"r")` of a file that may or may not exist you get undefined behavior.

I've seen segfaults all over the place because of this. In particular when trying to open configuration or help files that don't exist and when specialized parts of the `/proc` file system (e.g. `/proc/net/psched`) are implemented on a particular system.

I'm certainly open to a better solution if you can think of one. In the meantime I'll fix whatever bugs pop up in the automated testing (it looks like, minimum, I broke an NFS test).

jsquyres:
Huh; I've never seen `fopen()` fail that way, but I can't say I've ever gone looking for it, either.

I wonder if we should do what we've done in a few other portability scenarios:

* Make an `opal_fopen()` function that basically wraps calling `access()` and `fopen()`.  It guarantees to return either a valid file pointer or NULL (e.g., if `access()` fails, it returns NULL).
   * Be sure to include a good comment in `opal_fopen()` that describes the need for this wrapper, perhaps even pointing to this PR.
* Strategically `#define fopen opal_fopen` so that the entire code base benefits from this wrapper function.

That way, we don't have to add calls to `access()` everywhere, but we still protect the entire code base properly.

rhc54:
Me too - I've never heard of nor encountered this before, which is why I'm suspicious. I like the suggestion from @jsquyres - not only is it cleaner, but it means we only have one place that static code analyzers are going to complain about. Both Klockworks and Coverity will object that this PR creates a race condition between checking for file existence and then accessing it as there is no guarantee that the file hasn't been removed in the interim.

This is why we studiously cleansed the code base of calls to access (or stat) prior to calling a file access function - we were getting peppered with defect reports, and some of our companies (ahem, mine) will not allow release of code that crosses certain defect density boundaries. I'm a little bothered that we are now going back to reinsert those calls - but if we do the wrapper, then it will only count as one defect, so we might get away with it.

npe9:
@rhc54 @jsquyres Here's a relevant crash if I move my mca-params.conf out of my home directory:

`#0  0x00007ffff6330060 in fileno_unlocked () from /lib64/libc.so.6
#1  0x00007ffff71b2965 in fopen (path=<optimized out>, mode=<optimized out>) at src/syscall.c:60
#2  0x00007fff760d1c59 in pmix_util_keyval_parse (filename=0x6f8c90 "/home/nevans/.pmix/mca-params.conf", callback=0x7fff7615b982 <save_value>)
    at util/keyval_parse.c:75
#3  0x00007fff7615b96e in pmix_mca_base_parse_paramfile (paramfile=0x6f8c90 "/home/nevans/.pmix/mca-params.conf",
    list=0x7fff763a1240 <pmix_mca_base_var_file_values>) at pmix_mca_base_parse_paramfile.c:44
#4  0x00007fff7615e24e in read_files (file_list=0x6f84e0 "/home/nevans/.pmix/mca-params.conf:/home/nevans/etc/pmix-mca-params.conf",
    file_values=0x7fff763a1240 <pmix_mca_base_var_file_values>, sep=58 ':') at pmix_mca_base_var.c:1269
#5  0x00007fff7615cca7 in pmix_mca_base_var_cache_files (rel_path_search=false) at pmix_mca_base_var.c:549
#6  0x00007fff7615c2f7 in pmix_mca_base_var_init () at pmix_mca_base_var.c:279
#7  0x00007fff7615646d in pmix_rte_init (type=PMIX_PROC_SERVER, info=0x6d9da0, ninfo=4, cbfunc=0x0) at runtime/pmix_init.c:134
#8  0x00007fff76103cd1 in OPAL_MCA_PMIX2X_PMIx_server_init (module=0x7fff7639c1c0 <mymodule>, info=0x6d9da0, ninfo=4) at server/pmix_server.c:153
#9  0x00007fff760adb3b in pmix2x_server_init (module=0x7ffff7dd4c40 <pmix_server>, info=0x7fffffffce60) at pmix2x_server_south.c:144
#10 0x00007ffff7b4fba1 in pmix_server_init () at orted/pmix/pmix_server.c:287
#11 0x00007ffff510f943 in rte_init () at ess_hnp_module.c:657
#12 0x00007ffff7b10b7d in orte_init (pargc=0x7fffffffd06c, pargv=0x7fffffffd060, flags=4) at runtime/orte_init.c:273
#13 0x00007ffff7b3db02 in orte_daemon (argc=16, argv=0x7fffffffd438) at orted/orted_main.c:345
#14 0x00000000004008ec in main (argc=16, argv=0x7fffffffd438) at orted.c:60`
(gdb) disassemble
Dump of assembler code for function fileno_unlocked:
=> 0x00007ffff6330060 <+0>:	testl  $0x2000,(%rdi)
   0x00007ffff6330066 <+6>:	je     0x7ffff6330078 <fileno_unlocked+24>
...
(gdb) regs
...
rdi            0x0	0
...
`
So fileno is trying to dereference null pointer, which I believe is the indirection on the `fp->_flags` in the following code.

`int
__fileno (_IO_FILE *fp)
{
  CHECK_FILE (fp, EOF);

  if (!(fp->_flags & _IO_IS_FILEBUF) || _IO_fileno (fp) < 0)
    {
      __set_errno (EBADF);
      return -1;
    }

  return _IO_fileno (fp);
}`

npe9:
I'm compiling a personal glibc to see if I can figure out what the bug is exactly.

rhc54:
Weird - understand, very few users even _have_ a local mca-params.conf in their home directory. I never have had one myself. In 13+ years of production, I have never heard anyone report a segfault because the file didn't exist.

As I said,  maybe you have found some weird configuration that results in problems. We can certainly add the proposed protection (though I'd follow the suggestion from @jsquyres) - we just need to understand that it introduces a race condition that will generate (perfectly valid) complaints from the code checkers.

Might even want to ad a configure option to turn this off to avoid the complaints - like I said, nobody has ever reported such a thing, so maybe it only needs to be there for some configurations.

npe9:
@rhc54 Let me dig a little deeper here. I'll close this pull request for the moment (You guys' solution is better anyway) and see if I can really understand what its problem is. I can use this code to test the environment code now which will let me make sure the earlier pull request works.

PR #4343, opened 2017-10-15
OSHMEM/CONFIGURE: Check for the presence of ibv_exp_reg_shared_mr.

alinask:
+ The sshmem verbs component will disqualify itself if this verb isn't
present on the build host.
+ In case where support was requested but not found, don't stop the
build - continue without this component.

Signed-off-by: Alina Sklarevich <alinas@mellanox.com>
(cherry picked from commit 3008827f834618ad039674a6fa317180300e76a9)

PR #4344, opened 2017-10-16
OSHMEM/CONFIGURE: verbs component - restore the previous build behavior

alinask:
In case where support was requested but not found, stop the build.


alinask:
@yosefe @rhc54 , This is following https://github.com/open-mpi/ompi/pull/4343 , can you please review?

Issue #4347, opened 2017-10-17
pmix: OMPI configure error with external PMIx that installed to `lib64` path


karasevb:
If there is the installed PMIx in `lib64` path then an attempt to configure the OMPI with it looks follow:
```
$ ls /opt/pmix/lib64
/opt/pmix/lib64/libpmix.la
/opt/pmix/lib64/libpmix.so
/opt/pmix/lib64/libpmix.so.2
/opt/pmix/lib64/libpmix.so.2.0.3

$ ./configure –with-pmix=/opt/pmix
...
checking --with-external-pmix value... sanity check ok (/opt/pmix/include)
checking --with-external-libpmix value... not found
configure: WARNING: Directory /opt/pmix/lib not found
configure: error: Cannot continue
```

PR #4348, opened 2017-10-17
configury: enhance external PMIx detection - add the --with-pmix-libd…

ggouaillardet:
…ir=DIR option look for libpmix.* libs in DIR, DIR/lib64 and DIR - if --with-pmix=DIR is given, look for libpmix.so in DIR/lib64 and DIR/lib

Fixes open-mpi/ompi#4347

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -14,7 +14,7 @@
 # Copyright (c) 2011-2014 Los Alamos National Security, LLC. All rights
 #                         reserved.
 # Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
-# Copyright (c) 2014-2016 Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@karasevb can you please give this a try ?
this should work even without using the new ```--with-pmix-libdir=DIR``` option.

karasevb:
:+1:

ggouaillardet:
@hppritcha @bwbarrett i will issue a PR for the release branches.
strictly speaking, ```--with-pmix-libdir``` is a new option, but we can consider this as a bug fix,
otherwise, Open MPI cannot be built with an external 32 bits PMIx on a 64 bits system.
so i'd rather not trim this from the PR to come.

any thoughts ?

rhc54:
@ggouaillardet I would suggest bringing it back to the 2.1 and 3.x - not sure it needs to go any further back as the older branches are closing.

Issue #4336, opened 2017-10-13
Can't open more than 234 files over the course of a program run


william-dawson:
## Background information
I'm sure that this bug is rarely encountered, but I encountered this when testing my code, which involves running it with a large number of parameters.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
2.1.2

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
Installed using homebrew. 

### Please describe the system on which you are running
Mac OS 10.12.6
MacBook Pro (Retina, 13-inch, Early 2015)

-----------------------------

## Details of the problem
You can reproduce the code with the following Fortran program:
```
PROGRAM TestFileIO
  USE MPI
  IMPLICIT NONE
  INTEGER :: counter
  INTEGER :: mpi_err
  INTEGER :: mpi_file_handler
  INTEGER(KIND=MPI_OFFSET_KIND), PARAMETER :: zero_size = 0

  CALL MPI_Init(mpi_err)

  DO counter = 1, 250
    CALL MPI_File_open(MPI_COMM_WORLD, "test.txt", &
        & IOR(MPI_MODE_CREATE,MPI_MODE_WRONLY), MPI_INFO_NULL, &
        & mpi_file_handler, mpi_err)
    CALL MPI_File_close(mpi_file_handler, mpi_err)
  END DO

  CALL MPI_Finalize(mpi_err)
END PROGRAM
```

Compile with: `mpif90 test.f90 -o test`
Run with: `mpirun -np 1 ./test`

```
[redacted] mca_sharedfp_sm_file_open: Error, unable to open file for mmap: /tmp/OMPIO_test.txt_-1648689151_.sm
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[redacted] mca_sharedfp_sm_file_open: Error during file open
[...]

PR #4352, opened 2017-10-17
sharedfp/sm: close the named semaphore

edgargabriel:
in case a named semaphore is used, it is necessary to close the semaphore to remove
all sm segments. sem_unlink just removes the name references once all proceeses have closed
the sem.

Fixes issue: #4336

### DIFF
--- a/ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
+++ b/ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
@@ -231,6 +231,12 @@ int mca_sharedfp_sm_file_open (struct ompi_communicator_t *comm,
 
     comm->c_coll->coll_barrier (comm, comm->c_coll->coll_barrier_module );
 
+#if defined(HAVE_SEM_OPEN)
+    if ( 0 == rank ) {
+        sem_unlink ( sm_data->sem_name);
+    }
+#endif
+
     return err;
 }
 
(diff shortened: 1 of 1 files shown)


edgargabriel:
I can confirm that using this patch a test open/closing an MPI File finishes correctly without error messages on MacOS.

ompi/mca/sharedfp/sm/sharedfp_sm_file_open.c
```diff
@@ -231,6 +231,10 @@ int mca_sharedfp_sm_file_open (struct ompi_communicator_t *comm,
 
     comm->c_coll->coll_barrier (comm, comm->c_coll->coll_barrier_module );
 
+#if defined(HAVE_SEM_OPEN)
+    sem_unlink ( sm_data->sem_name);
```
ggouaillardet: should this only be invoked on (per node ?) rank zero ? that is likely harmless, but this should be invoked only once per node.

edgargabriel:
@ggouaillardet , thank you, I fixed this. Can you please review again?

ggouaillardet:
looks good to me, thanks !
something looks busted with the github review process, so i am now merging this PR.

edgargabriel:
@ggouaillardet thank you. I created to pull requests of this patch for 3.0.1 and 3.1, if necessary, we can also create a patch for the 2.1.x.

PR #4362, opened 2017-10-18
Add file locking support in posix fbtl

edgargabriel:
Based on two bug reports on the Open MPI mailing list, this commit introduces support for locking byte-ranges in a file before reading/writing to better support NFS file systems. The challenge is to find an approach that works without introduces an outrages performance penalty for other scenarios.

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_file_open.c
+++ b/ompi/mca/common/ompio/common_ompio_file_open.c
@@ -90,6 +90,7 @@ int mca_common_ompio_file_open (ompi_communicator_t *comm,
     ompio_fh->f_amode  = amode;
     ompio_fh->f_info   = info;
     ompio_fh->f_atomicity = 0;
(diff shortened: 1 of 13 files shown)


edgargabriel:
Note: not intending to merge yet, have to add one more feature. Just want to see whether jenkins raises some issues that need to be addressed.

breuner:
hi @edgargabriel , 
i saw your comment here (https://github.com/open-mpi/ompi/blame/main/ompi/mca/fs/ufs/fs_ufs_file_open.c#L101) and the team is working on new nfs improvements so that the locking should no longer be necessary when we're done. 
you mentioned in the comment that you ran tests for this. do you still have that test suite or test program so that we can use it to double check our improvements?

edgargabriel:
Just as an update for the community, I provided the tests to @breuner by email.

Issue #4373, opened 2017-10-22
install problem macOS High Sierra


thierrygosselin:
Trying to install Open MPI v.3.0.0 on macOS High Sierra (10.13)

Compiler installed: `gcc version 7.1.0 (GCC) `

The command:
```
cd Downloads
wget https://www.open-mpi.org/software/ompi/v3.0/downloads/openmpi-3.0.0.tar.gz
tar -xvf openmpi-3.0.0.tar.gz
cd openmpi-3.0.0
export CC=/usr/local/bin/gcc
export CXX=/usr/local/bin/g++
export CXX1X=/usr/local/bin/g++
export FC=/usr/local/bin/gfortran
export TMPDIR=/tmp
./configure
```

no error, everything looks fine

then running `make` gives these errors:

```
output.c: In function ‘opal_output_init’:
output.c:152:49: error: ‘LOG_INFO’ undeclared (first use in this function); did you mean ‘SIGINFO’?
             opal_output_redirected_syslog_pri = LOG_INFO;
                                                 ^~~~~~~~
                                                 SIGINFO
output.c:152:49: note: each undeclared identifier is reported only once for each function it appears in
output.c:154:49: error: ‘LOG_ERR’ undeclared (first use in this function); did you mean ‘SIG_ERR’?
             opal_output_redirected_syslog_pri = LOG_ERR;
                                                 ^~~~~~~
                                                 SIG_ERR
output.c:156:49: error: ‘LOG_WARNING’ undeclared (first use in this function)
             opal_output_redirected_syslog_pri = LOG_WARNING;
                                                 ^~~~~~~~~~~
output.c: In function ‘opal_output_close’:
output.c:354:13: warning: implicit declaration of function ‘closelog’; did you mean ‘close’? [-Wimplicit-function-declaration]
             closelog();
             ^~~~~~~~
             close
output.c: In function ‘do_open’:
output.c:621:13: warning: implicit declaration of function ‘openlog’; did you mean ‘openat’? [-Wimplicit-function-declaration]
             openlog(redirect_syslog_ident, LOG_PID, LOG_USER);
             ^~~~~~~
             openat
output.c:621:44: error: ‘LOG_PID’ undeclared (first use in this function); did you mean ‘P_PID’?
             openlog(redirect_syslog_ident, LOG_PID, LOG_USER);
                                            ^~~~~~~
                                            P_PID
[...]

PR #4374, opened 2017-10-22
Topic/osx syslog

bosilca:
gcc 7.1 is unable to correctly parse the syslog.h header file on OSX 10.13. Thus we need to completely disable syslog support.

A discussion about this topic can be found in #4373 .

### DIFF
--- a/orte/mca/notifier/syslog/configure.m4
+++ b/orte/mca/notifier/syslog/configure.m4
@@ -1,0 +1,31 @@
+# -*- shell-script -*-
+#
+# Copyright (c) 2017 Cisco Systems, Inc.  All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+# MCA_notifier_syslog_CONFIG([action-if-found], [action-if-not-found])
+# -----------------------------------------------------------
+AC_DEFUN([MCA_orte_notifier_syslog_CONFIG], [
+    AC_CONFIG_FILES([orte/mca/notifier/syslog/Makefile])
+
+    OPAL_VAR_SCOPE_PUSH([orte_notifier_syslog_happy])
+
+    # Per https://github.com/open-mpi/ompi/issues/4373 and
+    # https://github.com/open-mpi/ompi/pull/4374, we need to check
+    # that syslog.h is compilable.  If syslog.h is not compilable,
+    # disable this component.
+    AC_CHECK_HEADER([syslog.h],
+        [orte_notifier_syslog_happy=1],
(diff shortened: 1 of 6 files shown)


jsquyres:
@rhc54 Heads up -- this contains PMIX changes.  You'll likely want to pull them into PMIx proper.

jsquyres:
@bosilca I just added fe8d4aca5c069b487838181d0d25ad68908539eb to this PR which also disables the ORTE `syslog` `notifier` component if `<syslog.h>` is not compilable.

You probably want to squash that commit into your ORTE commit.

orte/mca/notifier/syslog/configure.m4
```diff
@@ -0,0 +1,33 @@
...
+# -----------------------------------------------------------
+AC_DEFUN([MCA_orte_notifier_syslog_CONFIG], [
+    AC_CONFIG_FILES([orte/mca/notifier/syslog/Makefile])
+
+    OPAL_VAR_SCOPE_PUSH([orte_notifier_syslog_happy])
+
+    # Per https://github.com/open-mpi/ompi/issues/4373 and
+    # https://github.com/open-mpi/ompi/pull/4374, we need to check
+    # that syslog.h is compilable (we already checked this in the main
+    # configure.ac, but the result is cached / invisible to us -- so
```
bwbarrett: This comment is poorly phrased.  The result of the previous test *IS* available to the component.  The right way to get the result is to call AC_CHECK_HEADER() again.  Caching is a wonderful thing...

jsquyres: K.  How about:

```sh
    # Per https://github.com/open-mpi/ompi/issues/4373 and
    # https://github.com/open-mpi/ompi/pull/4374, we need to check
    # that syslog.h is compilable (we already checked this in the main
    # configure.ac, but the result is not visible to us -- so check it
    # again and examine the result this time).  If syslog.h is not
    # compilable, disable this component.
```

bwbarrett: Why not just:

```shell
    # Per https://github.com/open-mpi/ompi/issues/4373 and
    # https://github.com/open-mpi/ompi/pull/4374, we need to check
    # that syslog.h is compilable.  If syslog.h is not
    # compilable, disable this component.
```

Calling AC_CHECK_HEADER and friends again and again is a common pattern in Open MPI's configure script.  It doesn't make sense to call that pattern out as somehow special in this one case.

bosilca:
If we're good with this I will make PR for 2.x and 3.x.

jsquyres:
Thanks @bosilca.  I wouldn't bother with 2.0.x -- just start with 2.1.x.

PR #4377, opened 2017-10-23
Pr/v3.0.x fbtl locking support

edgargabriel:
Based on two bug reports on the Open MPI mailing list, this commit introduces support for locking byte-ranges in a file before reading/writing to better support NFS file systems. The challenge is to find an approach that works without introduces an outrages performance penalty for other scenarios.

Equivalent to pr #4362


bwbarrett:
bot:ompi:retest

PR #4383, opened 2017-10-24
configury: revamp ucx detection

ggouaillardet:
- when --with-ucx=DIR is not set, try the default path and fallback to /opt/ucx
 - when --with-ucx-libdir is not set, try lib64 and then lib directories
 - do not handle --with-ucx-libdir (this is a user mistake, no need to over-complicate our logic)

### DIFF
--- a/config/ompi_check_ucx.m4
+++ b/config/ompi_check_ucx.m4
@@ -20,86 +20,97 @@
 # LDFLAGS, LIBS} as needed and runs action-if-found if there is
 # support, otherwise executes action-if-not-found
 AC_DEFUN([OMPI_CHECK_UCX],[
-    if test -z "$ompi_check_ucx_happy" ; then
-	AC_ARG_WITH([ucx],
-		    [AC_HELP_STRING([--with-ucx(=DIR)],
-				    [Build with Unified Communication X library support])])
-	OPAL_CHECK_WITHDIR([ucx], [$with_ucx], [include/ucp/api/ucp.h])
-	AC_ARG_WITH([ucx-libdir],
-		    [AC_HELP_STRING([--with-ucx-libdir=DIR],
-				    [Search for Unified Communication X libraries in DIR])])
-	OPAL_CHECK_WITHDIR([ucx-libdir], [$with_ucx_libdir], [libucp.*])
+    OPAL_VAR_SCOPE_PUSH([ompi_check_ucx_dir])
 
-	AS_IF([test "$with_ucx" != "no"],
-              [AS_IF([test ! -z "$with_ucx" && test "$with_ucx" != "yes"],
-                     [_x_ac_ucx_dirs="$with_ucx"],
-                     [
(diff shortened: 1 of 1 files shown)


ggouaillardet:
:bot:aws:retest

ggouaillardet:
:bot:aws:retest

bwbarrett:
bot:ompi:retest

@ggouaillardet, the OMPI Jenkins sitting in AWS didn't understand bot:aws:retest (only ompi: or lanl:).  It seemed like a reasonable guess, so I added aws: to the list as well.

ggouaillardet:
Thanks !
so that means there is a bigger mystery here. i issued the aws command on #4384 and it instantly triggered a rebuilt ! then i issued the same aws command twice on this PR, but nothing happened ... and still nothing happened with the ompi command you issued.

bwbarrett:
there was a typo in my update to make `bot:aws:retest` work.  It never would have worked before (but there are some other updates, like a new commit, that can cause a retest.

bot:aws:retest

ggouaillardet:
thanks !

jsquyres:
I'm confused by the commit message on this commit:

```
 - when --with-ucx-libdir is not set, try lib64 and then lib directories
 - do not handle --with-ucx-libdir (this is a user mistake, no need to over-complicate our logic)
```

The 1st line says it will handle `--with-ucx-libdir`.  But the 2nd line says it will *not* handle `--with-ucx-libdir`.  Which is it?

ggouaillardet:
```configure --with-ucx=/foo``` will look for lib in ```/foo/lib64``` and ```/foo/lib```
```configure --with-ucx=/foo --with-ucx-libdir=/foo/bar``` will look for lib in ```/foo/bar```
```configure --with-ucx=/foo --with-ucx-libdir``` will look for lib in ```yes``` (this is a dumb thing to do, and we do not protect end users against themselves any more in this case)

jsquyres:
Ooohhhh... I see.  I'm sorry to be type-A, but can you add something to that 3rd line in the commit to explain the `yes` value?

...and/or, what do we do elsewhere if the user specifies `--with-FOO-libdir` (with no value) -- do we ignore, or do we abort?  We should be consistent.

ggouaillardet:
ok, i ll reword this.
i checked here and there, and in other places, if ``--with-FOO-libdir``` is used, then the libdir is ```yes``` (e.g. we do nothing special), i will check again to be sure.

jsquyres:
Is it worth making a support m4 macro that we can use in multiple places (for `--with-FOO` and `--with-FOO-libdir` handling) to guarantee consistent handling of these CLI args across the code base?

ggouaillardet:
it turns we tread `--with-FOO-libdir=yes` as a no-op almost everywhere. (and a notable exception is `opal_check_pmi.m4` and this is what i used as a reference ...)
fwiw, if we gently ignore `--with-FOO-libdir`, should we also ignore `--without-FOO-libdir`
and yes, a macro could be helpful to achieve consistency

PR #4396, opened 2017-10-25
Alter the PMIx embedded configuration

rhc54:
Alter the PMIx embedded configuration so that we can build static with devel headers - if the builder requests that we install a separate libpmix, then don't prefix the PMIx variables.


rhc54:
sigh - lost osx tester connection

bot:aws:retest

rhc54:
continued loss of connection to os-x tester

Issue #4345, opened 2017-10-16
Configure generates bogus makefile


planetA:
## Background information

When I try to compile OpenMPI together with UCX, I expect the configure system to find UCX libraries. Configure script seems to find them, but fails to generate makefiles which point to the right location. This results in libtool failing to properly link the libraries.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

I use ucx-1.2.1 and OpenMPI 2.1.1 downloaded from GitHub.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

When I compile OpenMPI it fails with following error message:

```
libtool:   error: require no space between '-L' and '-lrt'
```

To ease the reproducibility, I give Dockerfile to compile everything in an isolated environment.

### Please describe the system on which you are running

* Operating system/version: Debian 9 inside docker
* Computer hardware: NA
* Network type: NA

-----------------------------

## Details of the problem

Using this dockerfile, compilation fails

```Dockerfile
FROM Debian:9
MAINTAINER "Nobody <nobody@domain.invalid>"

RUN apt-get update && \
    apt-get install -y flex git libtool build-essential gcc gfortran wget libnuma-dev autoconf automake && \
    wget https://github.com/openucx/ucx/archive/v1.2.1.tar.gz && \
    tar xvf v1.2.1.tar.gz && \
    cd ucx-1.2.1 && ./autogen.sh && ./configure && make && make install && cd .. && \
    rm -rf ucx-1.2.1 v1.2.1.tar.gz && \
    apt-get clean && apt-get purge
RUN wget https://github.com/open-mpi/ompi/archive/v2.1.1.tar.gz && \
    tar xvf v2.1.1.tar.gz && \
    cd ompi-2.1.1 && \
    ./autogen.pl && ./configure --with-ucx && make && make install && \
    cd .. && \
    rm -rf ompi-2.1.1 v2.1.1.tar.gz
```
Changing --with-ucx to --with-ucx=/usr results in successful compilation.

bwbarrett:
@jladd-mlnx, can you find someone from your team to look at this?

ggouaillardet:
@planetA  can you give the inline patch a try ?

fwiw, ```master``` does things a bit differently, but bottom line, we do not need ```$ompi_check_ucx_extra_libs```. also some more things need to be cleaned up
[...]

PR #3962, opened 2017-07-25
configure: detect UCX support by default

karasevb:
Adds detecting UCX from following paths: "/usr /usr/local /opt/ucx"

Signed-off-by: Boris Karasev <karasev.b@gmail.com>

PR #4398, opened 2017-10-25
v3.0: configury: revamp ucx detection

ggouaillardet:
- when --with-ucx=DIR is not set, try the default path and fallback to /opt/ucx
 - when --with-ucx-libdir is not set, try lib64 and then lib directories
 - do not handle --with-ucx-libdir (this is a user mistake, no need to over-complicate our logic)

Refs. open-mpi/ompi#4345

(cherry picked from commit open-mpi/ompi@af03f55aa87b6c97c8a0b3375e97cc8aa7eb3539)

### DIFF
--- a/config/ompi_check_ucx.m4
+++ b/config/ompi_check_ucx.m4
@@ -20,86 +20,97 @@
 # LDFLAGS, LIBS} as needed and runs action-if-found if there is
 # support, otherwise executes action-if-not-found
 AC_DEFUN([OMPI_CHECK_UCX],[
-    if test -z "$ompi_check_ucx_happy" ; then
-	AC_ARG_WITH([ucx],
-		    [AC_HELP_STRING([--with-ucx(=DIR)],
-				    [Build with Unified Communication X library support])])
-	OPAL_CHECK_WITHDIR([ucx], [$with_ucx], [include/ucp/api/ucp.h])
-	AC_ARG_WITH([ucx-libdir],
-		    [AC_HELP_STRING([--with-ucx-libdir=DIR],
-				    [Search for Unified Communication X libraries in DIR])])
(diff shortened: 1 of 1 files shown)


hppritcha:
hmmm, now its OS-X.
bot:ompi:retest

jsquyres:
Per https://github.com/open-mpi/ompi/pull/4383#issuecomment-339683584, I'm confused by the commit message on this PR.

alinask:
@ggouaillardet Can you please comment on why you are setting /opt/ucx as the fallback?

ggouaillardet:
The current version already fallback to ```/opt/ucx```. I did not write this, and i did not change that behavior.

ggouaillardet:
@karasevb since you introduced this in https://github.com/open-mpi/ompi/commit/d917d54ddc56c67a334f0b36e93fcd92a985933e can you please answer @alinask ?

karasevb:
@alinask the main idea of the original PR https://github.com/open-mpi/ompi/pull/3962 was the added UCX autodetection by default (and also for case when the UCX path was not defined when configure) on the specified paths:
1. discovered by `pkg-config`
2. `/usr` 
3. `/usr/local` 
4. `/opt/ucx` this is one of the possible path where UCX can be found by default

@ggouaillardet meant that `configure` will look for UCX in `/usr:/usr/local` and then `/opt/ucx`, please correct me if this is not true.

hppritcha:
@jladd-mlnx please review.  we hit this on 2.1.3 at LANL and its very annoying (issue #4345 I mean).  I had to ask the sysadmin to do a hack LDFLAG workaround on the configure line.

I'd like to patch 2.1.X with this patch if its accepted.

hppritcha:
@xinzhao3 please ask Josh to review this PR.

jladd-mlnx:
@hppritcha good to go.

PR #4403, opened 2017-10-26
fix PML monitoring configury to compile DSOs

### LINKED ISSUE #4402: pml monitoring fails to build

 - OMPI master `05f98d6`
 - Linux x86-64, git clone, but should apply to all

* Operating system/version:  Linux
* Computer hardware: x86-64
* Network type: 

-----------------------------

## Details of the problem

The PML Monitoring framework is not building.  It silently says that it can not compile, e.g., output from config.log

```
   ...<snip>...
configure:5914: --- MCA component pml:monitoring (m4 configuration macro)
configure:325400: checking for MCA component pml:monitoring compile mode
configure:325406: result: dso
configure:325917: checking if MCA component pml:monitoring can compile
configure:325919: result: no
   ...<snip>...
```

Test that **should** work but does not:

```
shell$ cd test/monitoring/
shell$ mpirun -np 2 -mca pml_monitoring_enable 2 -mca pml_monitoring_enable_output 2 ./monitoring_test
```

naughtont3:
Fixes #4402 

### DIFF
--- a/ompi/mca/coll/monitoring/configure.m4
+++ b/ompi/mca/coll/monitoring/configure.m4
@@ -16,7 +16,7 @@
 AC_DEFUN([MCA_ompi_coll_monitoring_CONFIG],[
     AC_CONFIG_FILES([ompi/mca/coll/monitoring/Makefile])
 
-    AS_IF([test MCA_BUILD_ompi_common_monitoring_DSO_TRUE == ''],
+    AS_IF([test "$MCA_BUILD_ompi_common_monitoring_DSO_TRUE" == ''],
           [$1],
           [$2])
 ])dnl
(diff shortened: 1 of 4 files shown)


ompi/mca/coll/monitoring/configure.m4
```diff
@@ -16,7 +16,7 @@
 AC_DEFUN([MCA_ompi_coll_monitoring_CONFIG],[
     AC_CONFIG_FILES([ompi/mca/coll/monitoring/Makefile])
 
-    AS_IF([test MCA_BUILD_ompi_common_monitoring_DSO_TRUE == ''],
+    AS_IF([test "$MCA_BUILD_ompi_common_monitoring_DSO_TRUE" == ''],
```
jsquyres: Wait -- isn't using `==` with `test` non-portable?  I.e., doesn't this get us in trouble with the BSD flavors of `test`?

naughtont3: Good question, I just added quoting and '$' to expand the variable.  It does appear that other instances of `test` with  `AS_IF` and  `MCA_BUILD_`   are using single `=`.  I'll defer to experts.

bosilca: According to the rest of OMPI code Jeff is right we should use a single = instead.

PR #4428, opened 2017-10-31
Revert the MPI_Init fence operations to use volatile bool instead of thread macros

rhc54:
Refs https://github.com/pmix/pmix/issues/580

### DIFF
--- a/ompi/runtime/ompi_mpi_finalize.c
+++ b/ompi/runtime/ompi_mpi_finalize.c
@@ -95,8 +95,10 @@ extern bool ompi_enable_timing;
 
 static void fence_cbfunc(int status, void *cbdata)
 {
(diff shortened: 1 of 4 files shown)


artpol84:
:+1: Fixes pmix/pmix#580

artpol84:
Need this starting from 3.0.x

rhc54:
Not really - it is only in master today

artpol84:
I was using 3.0.x when I caught it.
And to test I cherry picked

rhc54:
Don't know what to say - look at the v3.0.x branch code and you can clearly see that none of those thread lock code lines are present

artpol84:
@rhc54 you are right, looks like I was working with the master instead of v3.0.x as I though. Sorry for the confusion.

PR #4431, opened 2017-10-31
opal: rename opal_atomic_cmpset* to opal_atomic_bool_cmpset*

hjelmn:
This commit renames the atomic compare-and-swap functions to indicate
the return value. This is in preperation for adding support for a
compare-and-swap that returns the old value. At the same time the
return type has been changed to bool.


hjelmn:
@bosilca More atomic stuff.

Issue #4453, opened 2017-11-06
Incorrect file size from MPI_File_write with Type_subarray


edgargabriel:
Thank you for taking the time to submit an issue!

## Background information

This is a bug reported on the mailing list by Christopher Brady, where the overall data written is less than expected when using large datatypes created with MPI_Type_create_subarray.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

Open MPI 2.x and 3.0.0

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

edgargabriel:
Fix has been applied to master, v2.1.x, v3.0.x and v3.1.x

PR #4454, opened 2017-11-06
io/ompio: fix a bug in handling large write/read operations

edgargabriel:
This is a bug fix based on a problem reported on the mailing list.
For very large read/write operations, ompio breaks the operation
down into multiple cycles. The problem was that
one of the variables required to maintain its values
across the different cycles did not do that, and because
of that the calculations of the memory offsets was wrong.

Fixes issue #4453


edgargabriel:
bot:lanl:retest

edgargabriel:
IBM's CI is temporarily down. Merging this PR.

Issue #4443, opened 2017-11-03
fs/ufs does not return proper error codes on file open


edgargabriel:
Thank you for taking the time to submit an issue!

## Background information

the fs/ufs component does not return proper error codes upon a problem to the application.
This issue has been reported on the mailing list.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

all 2.x and 3.x branches, master

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

edgargabriel:
fix has been applied to master, v3.0.x and v3.1.x

PR #4460, opened 2017-11-07
fs/ufs: add some more error codes on file_open

ggouaillardet:
set proper error codes in mca_fs_ufs_file_open by mapping the errno value to
the MPI error code.

Refs. open-mpi/ompi#4443


ggouaillardet:
:bot:aws:retest

PR #4469, opened 2017-11-08
Sync with PMIx master

rhc54:
Implement direct modex protection to turn off PMIx dstore when direct modex scenario is detected

### DIFF
--- a/opal/mca/pmix/pmix3x/pmix3x_client.c
+++ b/opal/mca/pmix/pmix3x/pmix3x_client.c
@@ -99,6 +99,11 @@ int pmix3x_client_init(opal_list_t *ilist)
         ninfo = 0;
     }
 
+    /* check for direct modex use-case */
+    if (opal_pmix_base_async_modex && !opal_pmix_collect_all_data) {
+        opal_setenv("PMIX_MCA_gds", "hash", true, &environ);
+    }
+
     OPAL_PMIX_RELEASE_THREAD(&opal_pmix_base.lock);
     rc = PMIx_Init(&my_proc, pinfo, ninfo);
     if (NULL != pinfo) {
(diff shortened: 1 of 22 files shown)


opal/mca/pmix/pmix3x/pmix3x_client.c
```diff
@@ -99,6 +99,11 @@ int pmix3x_client_init(opal_list_t *ilist)
         ninfo = 0;
     }
 
+    /* check for direct modex use-case */
+    if (opal_pmix_base_async_modex && !opal_pmix_collect_all_data) {
+        opal_setenv("PMIX_MCA_gds", "hash", true, &environ);
```
artpol84: @rhc54 
This line disables disables `dstore`. We accidentally found that. 
Can you remind me why this was done?
Also I think this kind of change should come in a separate PR as it significantly changes the behavior yet quite small to find it between other 400+ changed lines.

@jladd-mlnx FYI

artpol84: @karasevb

artpol84: @rhc54 
I see that this PR was filed on Nov, 7.
And we had a dix for the dstore and local commits on
Dec, 20:
https://github.com/pmix/pmix/pull/613 

Am I right that it supposed to fix this issue and we can remove this?

artpol84: @rhc54 FYI we discovered the issue with pmix/pmix#613.
@karasevb is going to PR the fix today

karasevb: PR: https://github.com/pmix/pmix/pull/697

rhc54:
You are commenting on something committed 5 months ago? IIRC, and it has been a long time, the change you cite above was done after general consultation as a workaround to the issue you cite that was not fixed until today. I believe the change required some corresponding changes in PMIx, which is why it was brought in together.

I have no issue with removing it now that the fix is available, once that fix has transitioned into OMPI.

artpol84:
`You are commenting on something committed 5 months ago?`
Yeah, this change was made here, so I decided to have a discussion in the context. 

`I have no issue with removing it now that the fix is available, once that fix has transitioned into OMPI.`
Thanks!

PR #4476, opened 2017-11-08
ompi/errhandler: make set/get actually thread safe

### LINKED ISSUE #4477: Get/set error handler functions are not actually thread safe

## Background information

We use atomics in the various error handler get/set routines but it is possible to get into a situation where a thread calling an error handler get routine can call OBJ_RETAIN on a freed pointer.

All versions.This is an old old bug.

N/A

* Operating system/version: N/A
* Computer hardware: N/A
* Network type: N/A

-----------------------------

## Details of the problem

The problem path can occur when one thread calls `MPI_*_get_errhandler` and goes to sleep between the `cmpset` loop and the retain. If another thread calls `MPI_*_set_errhandler` then `MPI_Errhandler_free` while the get thread is asleep then the retain call in the get thread is called on an invalid pointer.

hjelmn:
The current versions of these functions have a fatal flaw. If a
errhandler set and free call is made by another thread while the
thread calling get is between the cmpset and retain then we will
retain an invalid object. Fixing this by just using locking. This is
not a critical path so this should be ok.

Fixes #4477.

### DIFF
--- a/ompi/mpi/c/comm_set_errhandler.c
+++ b/ompi/mpi/c/comm_set_errhandler.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2016-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
@@ -69,11 +69,12 @@ int MPI_Comm_set_errhandler(MPI_Comm comm, MPI_Errhandler errhandler)
     /* Prepare the new error handler */
     OBJ_RETAIN(errhandler);
 
-    /* Ditch the old errhandler, and decrement its refcount.  On 64
-       bits environments we have to make sure the reading of the
-       error_handler became atomic. */
-    tmp = OPAL_ATOMIC_SWAP_PTR(&comm->error_handler, errhandler);
+    opal_mutex_lock (&comm->c_lock);
+    /* Ditch the old errhandler, and decrement its refcount. */
+    tmp = comm->error_handler;
+    comm->error_handler = errhandler;
     OBJ_RELEASE(tmp);
+    opal_mutex_unlock (&comm->c_lock);
 
     /* All done */
     return MPI_SUCCESS;
--- a/ompi/mpi/c/file_set_errhandler.c
+++ b/ompi/mpi/c/file_set_errhandler.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2008      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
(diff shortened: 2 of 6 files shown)


hjelmn:
See #4477. This bug was discovered while I was working to eliminate the cmpset routines. I have not seen the bug actually triggered but it is straightforward to see how it can occur.

bosilca:
I'll take a look after SC.

hjelmn:
@bosilca Ping.

hjelmn:
@bosilca Any comments on this? I want to get this in before updating the compare-and-swap update patch.

bosilca:
This cover an extremely unlike bug, one that can only be generated by a non-deterministic application. That being said the proposed fix works and correctly make the example less open to fault.

As the errhandler modification are protected by a mutex I don't understand what is the need for the opal_atomic_swap_ptr ?

ompi/mpi/c/comm_set_errhandler.c
```diff
@@ -69,11 +69,13 @@ int MPI_Comm_set_errhandler(MPI_Comm comm, MPI_Errhandler errhandler)
     /* Prepare the new error handler */
     OBJ_RETAIN(errhandler);
 
+    opal_mutex_lock (&comm->c_lock);
     /* Ditch the old errhandler, and decrement its refcount.  On 64
        bits environments we have to make sure the reading of the
```
bosilca: change this comment.

ompi/mpi/c/file_set_errhandler.c
```diff
@@ -68,11 +68,13 @@ int MPI_File_set_errhandler( MPI_File file, MPI_Errhandler errhandler)
     /* Prepare the new error handler */
     OBJ_RETAIN(errhandler);
 
+    opal_mutex_lock (&file->f_mutex);
     /* Ditch the old errhandler, and decrement its refcount.  On 64
        bits environments we have to make sure the reading of the
```
bosilca: change this comment.

ompi/mpi/c/win_set_errhandler.c
```diff
@@ -63,11 +63,13 @@ int MPI_Win_set_errhandler(MPI_Win win, MPI_Errhandler errhandler)
     /* Prepare the new error handler */
     OBJ_RETAIN(errhandler);
 
+    opal_mutex_lock (&win->w_lock);
     /* Ditch the old errhandler, and decrement its refcount.  On 64
        bits environments we have to make sure the reading of the
```
bosilca: change this comment.

hjelmn:
@bosilca You indicated that we need that for async progress to ensure that the error handler pointer is set atomically.

bosilca:
Yes, but only when set via atomic operations. The proposed infrastructure rely on mutexes to protect the set/get of the errorhandler, so the atomic operation is just overhead.

hjelmn:
@bosilca Ok, will makes the requested changes. More of atomic updates will be incoming shortly.

bosilca:
@hjelmn it took me a little while to figure out what I was referring to with regard to the atomic swap. The scenario I was envisioning was to protect the calls to the errorhandler while the main application is changing them. In other terms if a thread completes with error and needs to call the errhandler associated with an MPI object, while in another thread the user is changing the errorhandler. But, let's assume this is a corner case, and safe programming principles will prevent users from using such convoluted techniques.

PR #4482, opened 2017-11-09
osc/rdma: mca parameter to list MTLs that lower osc rdma priority 

matcabral:
osc/rdma: add mca parameter to list MTLs for which osc pt2pt should have higher priority than rdma and default it to psm2.

Context: the Intel Omni-path driver (hfi1) has verbs support, so the openib
btl is available to use. However, at a bad performance. Without this
change osc rdma using btl openib is the default choice when running on Intel
Omni-path, with a lower performance than osc pt2pt over mtl psm2.


hppritcha:
forgot to install software on one of the aarch64 boxes before bringing online.
bot:ompi:retest

matcabral:
Hi @hppritcha, would it be OK to merge this ?

PR #4459, opened 2017-11-06
Fix OFI MTL to recognize correct CQ empty scenario

aravindksg:
Currently, the progress function is incorrectly interpreting any error
value other than a positive value or -FI_EAVAIL to mean CQ is empty.
CQ is empty only if fi_cq_read() call returned -EAGAIN error
code. Fix that here.

While at it, fix help text output for calls made to OFI API.

Signed-off-by: Aravind Gopalakrishnan <Aravind.Gopalakrishnan@intel.com>
(cherry picked from commit 285fc42b4ea9c16c8e7b9caa1c5def453f80e557)

PR #4494, opened 2017-11-11
MTL OFI updates

jsquyres:
The most important of these is to not show an error if `fi_getinfo()` returns no data.

Need to get these fixes in to release branches before they go out, or usNIC customers will be asking my why they are getting these warnings.  😦   This affects:

* v2.x
* v3.0.x (PR #4458)
* v3.1.x (PR #4459)

These commits should be added to #4458 and #4459 before those PRs are merged.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -594,23 +594,23 @@ ompi_mtl_ofi_finalize(struct mca_mtl_base_module_t *mtl)
     opal_progress_unregister(ompi_mtl_ofi_progress_no_inline);
 
     /* Close all the OFI objects */
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.ep)) {
+    if ((ret = fi_close((fid_t)ompi_mtl_ofi.ep))) {
         goto finalize_err;
     }
 
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.cq)) {
+    if ((ret = fi_close((fid_t)ompi_mtl_ofi.cq))) {
         goto finalize_err;
     }
 
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.av)) {
+    if ((ret = fi_close((fid_t)ompi_mtl_ofi.av))) {
         goto finalize_err;
     }
 
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.domain)) {
+    if ((ret = fi_close((fid_t)ompi_mtl_ofi.domain))) {
         goto finalize_err;
     }
 
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.fabric)) {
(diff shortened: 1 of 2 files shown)


jsquyres:
Note: v2.0.x has a similar issue (i.e., shows a warning if `fi_getinfo()` returns no info), but it's via `opal_output_verbose()`, and therefore is not displayed by default.

The problem with v2.x and beyond is that it uses `opal_show_help()` in this case, which *does* output by default.  That's why this fix really only needs to be applied to v2.x and beyond (and we don't need to immediately issue another v2.0.x release right after releasing v2.0.4 -- especially when we wanted that to be the last v2.0.x release!).

hppritcha:
checking to see if ompi jenkins chattiness has been disabled if tests pass.
bot:ompi:retest

jsquyres:
Thanks @aravindksg

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -362,13 +362,16 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                      NULL,          /* Optional name or fabric to resolve       */
                      NULL,          /* Optional service name or port to request */
                      0ULL,          /* Optional flag                            */
-                     hints,        /* In: Hints to filter providers            */
+                     hints,         /* In: Hints to filter providers            */
                      &providers);   /* Out: List of matching providers          */
-    if (0 != ret) {
+    if (FI_ENODATA == -ret) {
+        // It is not an error if no information is returned.
```
matcabral: minor:  c++ comment style

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -591,23 +594,23 @@ ompi_mtl_ofi_finalize(struct mca_mtl_base_module_t *mtl)
     opal_progress_unregister(ompi_mtl_ofi_progress_no_inline);
 
     /* Close all the OFI objects */
-    if (ret = fi_close((fid_t)ompi_mtl_ofi.ep)) {
```
matcabral: why the extra parenthesis ?

ompi/mca/mtl/ofi/mtl_ofi_component.c
jsquyres: We allow this in Open MPI (since compiling Open MPI *requires* a C99 compiler, which officially standardized //-style comments in C).

jsquyres:
See the commit message on https://github.com/open-mpi/ompi/pull/4494/commits/a8686a681361177da28aac6e61963600fbd30d8a.

ompi/mca/mtl/ofi/mtl_ofi_component.c
jsquyres: See the commit message on https://github.com/open-mpi/ompi/pull/4494/commits/a8686a681361177da28aac6e61963600fbd30d8a.

PR #4500, opened 2017-11-13
pmix: pack pointer to object (vs. pointer to pointer)

jsquyres:


jsquyres:
@jsquyres and @rhc54 found this in some interactive debugging.

PR #4501, opened 2017-11-13
pml/ob1: fix MPI_ANY_SOURCE when built with --enable-heterogeneous

ggouaillardet:
when --enable-heterogeneous is used, we do not know what the remote
architecture will be when MPI_ANY_SOURCE is used. so we have to
delay the construction of the convertor when a message is matched.

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvfrag.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvfrag.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2009-2010 Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC. All rights
  *                         reserved.
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -285,6 +285,9 @@ void mca_pml_ob1_recv_frag_callback_match(mca_btl_base_module_t* btl,
     if(OPAL_LIKELY(match)) {
         bytes_received = segments->seg_len - OMPI_PML_OB1_MATCH_HDR_LEN;
         match->req_recv.req_bytes_packed = bytes_received;
+        for (size_t i=1; i < num_segments; i++) {
+            match->req_recv.req_bytes_packed += segments[i].seg_len;
+        }
 
         MCA_PML_OB1_RECV_REQUEST_MATCHED(match, hdr);
         if(match->req_bytes_expected > 0) {


ggouaillardet:
that fixes the issue reported at https://www.mail-archive.com/users@lists.open-mpi.org//msg31761.html

bosilca:
With SC going on I can't review this now. I need to understand how did we lost the capability of handling ANY_SOURCE in a heterogeneous environment.

ggouaillardet:
the PR is likely wrong.

the test is working fine with ```v1.10```
i checked the code, and the mechanism is there both in ```master``` and ```v1.10```

```c
static inline void recv_req_matched(mca_pml_ob1_recv_request_t *req,
                                    mca_pml_ob1_match_hdr_t *hdr)
{
    // ...
    if(req->req_recv.req_bytes_packed > 0) {
#if OPAL_ENABLE_HETEROGENEOUS_SUPPORT
        if(MPI_ANY_SOURCE == req->req_recv.req_base.req_peer) {
            /* non wildcard prepared during post recv */
            prepare_recv_req_converter(req);
        }
#endif  /* OPAL_ENABLE_HETEROGENEOUS_SUPPORT */
        // ...
}
```

for an (yet) unknown reason, `req->req_recv.req_bytes_packed` is non zero on ```v1.10``` but zero on ```master```

ggouaillardet:
what about this ?

```patch
diff --git a/ompi/mca/pml/ob1/pml_ob1_recvfrag.c b/ompi/mca/pml/ob1/pml_ob1_recvfrag.c
index f85964f..06c27bd 100644
--- a/ompi/mca/pml/ob1/pml_ob1_recvfrag.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvfrag.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2009-2010 Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2012-2015 Los Alamos National Security, LLC. All rights
- * Copyright (c) 2015      Research Organization for Information Science
+ * Copyright (c) 2015-2017 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -285,6 +285,9 @@ void mca_pml_ob1_recv_frag_callback_match(mca_btl_base_module_t* btl,
     if(OPAL_LIKELY(match)) {
         bytes_received = segments->seg_len - OMPI_PML_OB1_MATCH_HDR_LEN;
         match->req_recv.req_bytes_packed = bytes_received;
+        for (size_t i=1; i < num_segments; i++) {
+            match->req_recv.req_bytes_packed += segments[i].seg_len;
+        }
 
         MCA_PML_OB1_RECV_REQUEST_MATCHED(match, hdr);
         if(match->req_bytes_expected > 0) {
```

ggouaillardet:
@bosilca i updated this PR with the patched i previously inline'd.

could you please review it ?

besserox:
Any progress on that issue?

The current Open MPI build in Ubuntu 17.10 is broken/unreliable because of that.
cf https://bugs.launchpad.net/ubuntu/+source/openmpi/+bug/1731938

It would be annoying if this bug was still there in the next Ubuntu 18.04 LTS.

jsquyres:
@ggouaillardet @bosilca Any news on this really old PR?

bosilca:
This patch is not necessary anymore, the underlying issue has been inadvertently fixed back in February by commit 09cba8b. What we currently have inside is based on what I proposed above (adding num_segments-1 to the length).

jsquyres:
Meaning: we should close this PR without merging...?

bosilca:
@besserox I could not find what version of Open MPI the Ubuntu issue refers to. Any idea ?

bosilca:
Yes, but I wanted to give some time to @ggouaillardet to comment

bosilca:
That's a pretty ancient version. I am quite confident that backporting 09cba8b will be a nightmare, so we might need another way to fix it there. In any case if someone volunteers to fix 2.1.1 the patch for this issue is literally what I posted few messages above.

besserox:
From my point of view, the simple fix for OpenMPI 2.1.1 in Ubuntu was to remove the `--enable-heterogeneous` option as I think it was considered as broken as that time. (did that change since then?)

But considering that the package maintainer in Ubuntu didn't even reply to the bug report, even if that solution was initially proposed, I don't think providing a patch for 2.1.1 will change anything.

besserox:
It seems that the next version of Ubuntu (Cosmic 18.10) will use Open MPI 3.1.2 (https://launchpad.net/ubuntu/+source/openmpi). Is the issue fixed in this version of OpenMPI?

bosilca:
The fix is in [3.x](https://github.com/open-mpi/ompi/blob/v3.1.x/ompi/mca/pml/ob1/pml_ob1_recvfrag.c#L425).

jsquyres:
Let's go ahead and close this.  @ggouaillardet can re-open if necessary.

PR #4506, opened 2017-11-15
Make interface's kernel index an int instead of int16_t

wojciechwasko:
Sometimes, the ethernet interfaces can get quite high kernel indices. struct
ifreq (see netdevice(7)) defines ifr_ifindex to be int's. The OOB component
used int16_t internally for matching (in case of -mca oob_tcp_if_[in|ex]clude)
which meant that any interface index > 32767 would never be matched because the
integer would be truncated to int16_t upon return from the function. OOB would
then refuse to work because it didn't find any usable interfaces and MPI job
would abort.

Signed-off-by: Wojtek Wasko <wwasko@nvidia.com>

### DIFF
--- a/opal/util/if.c
+++ b/opal/util/if.c
@@ -133,7 +133,7 @@ int opal_ifnametoindex(const char* if_name)
  *  corresponding kernel index.
  */
 
-int16_t opal_ifnametokindex(const char* if_name)
+int opal_ifnametokindex(const char* if_name)
 {
     opal_if_t* intf;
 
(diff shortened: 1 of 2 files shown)


jsquyres:
ok to test

wojciechwasko:
Sorry, this is my first time contributing to ompi. What's the next step here?

jsquyres:
@wojciechwasko Sorry for the delay -- most of us are at the Supercomputing trade show this week, which led to some backlog in PRs.

wojciechwasko:
@jsquyres ah yes no worries, I was just wondering if any action was required on my part. This can wait, no problem.

PR #4511, opened 2017-11-17
v2.1.x: Make interface's kernel index an int instead of int16_t

jsquyres:
Sometimes, the ethernet interfaces can get quite high kernel indices. struct
ifreq (see netdevice(7)) defines ifr_ifindex to be int's. The OOB component
used int16_t internally for matching (in case of -mca oob_tcp_if_[in|ex]clude)
which meant that any interface index > 32767 would never be matched because the
integer would be truncated to int16_t upon return from the function. OOB would
then refuse to work because it didn't find any usable interfaces and MPI job
would abort.

Signed-off-by: Wojtek Wasko <wwasko@nvidia.com>
(cherry picked from commit 276de13a1e729c104fc564ddcce775608888c666)

Master PR: #4506

### DIFF
--- a/opal/util/if.c
+++ b/opal/util/if.c
@@ -143,7 +143,7 @@ int opal_ifnametoindex(const char* if_name)
  *  corresponding kernel index.
  */
 
-int16_t opal_ifnametokindex(const char* if_name)
+int opal_ifnametokindex(const char* if_name)
 {
     opal_if_t* intf;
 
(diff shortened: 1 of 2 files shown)


jsquyres:
@hppritcha I think this is good to go.  I've reviewed it (this was a user contribution on master; all I did was cherry-pick it over to the release branch), but github won't let me review it since I created the PR.

hppritcha:
I'm adding news here to help me track an item which I think should get a NEWs blurb for next release.

PR #4522, opened 2017-11-21
mpi/c/dims_create.c: avoid using log, sqrt, ceil in getfactors

haraldBraun-atos:
Hi!

We observed that applications linked against OpenMPI (2.x and onward) yield deviating numerical results than OpenMPI 1.x (or any other MPI-implementation tested).
Investigation revealed that the relatively large deviations (10% and more) originate from small deviations in results of calls to 'log', which then grow due to the error propagation behavior of the numerical algorithms utilized.

It appears that building/compiling ompi and the application with different Versions of ifort/icc made the issue visible: libmpi.so exports an implementation of 'log' from the comliler's preferred math-library.
The responsible call to 'log' was found in ompi/mpi/c/dims_create.c - in the function 'getfactors(...)'.

libmpi.so exports the symbol 'log', this leads to strange deviations, if
ompi was linked against another math-library than applications lateron
(in particular when using Intel's ifort/icc, which bring their own
math-library)
This is fixed by avoiding usage of math-library functions, i.e. log and
sqrt in the getfactors fnction in ompi/mpi/c/dims_create.c

Signed-off-by: Harald Braun harald.braun@atos.net

### DIFF
--- a/ompi/mpi/c/dims_create.c
+++ b/ompi/mpi/c/dims_create.c
@@ -233,7 +233,7 @@ upper_ilog2(const int i) {
  *  Returns:	- integer result
  */ 
 static int
-upper_isqrt(const int i; const int ilog2) {
+upper_isqrt(const int i, const int ilog2) {
 	if (i < 2) return i;
 	int a, b, c;
 	a = 1 << (((ilog2+1) >> 1) - 1);
@@ -281,7 +281,7 @@ getfactors(int num, int *nfactors, int **factors) {
         (*factors)[i++] = 2;
     }
 
-    sqrt_num = upper_isqrt(num, size - i);
+    sqrtnum = upper_isqrt(num, size - i);
     /* determine all occurences of uneven prime numbers up to sqrt(num) */
     d = 3;
     for(d = 3; (num > 1) && (d < sqrtnum); d += 2) {
--- a/ompi/mpi/c/dims_create.c
+++ b/ompi/mpi/c/dims_create.c
@@ -207,6 +207,46 @@ assignnodes(int ndim, int nfactor, int *pfacts, int **pdims)
 }
 
 /*
+ *  upper_ilog2
+ *
+ *  Function:   - compute ceiling of log2(i)
+ *  Accepts:	- integer number
+ *  Returns:	- integer result
+ */ 
+static int
+upper_ilog2(const int i) {
+	int ilog2 = 0;
+	int ii = i;
+	while (ii > 1) {
+		ii = (ii + 1) >> 1;
+		ilog2++;
+	}
+	return ilog2;
+}
+
+/*
+ *  upper_isqrt
+ *
+ *  Function:	- compute ceiling of sqrt(i)
+ *  Accepts:	- integer number
+ *  		- integer number (result of upper_ilog2(i))
+ *  Returns:	- integer result
+ */ 
+static int
+upper_isqrt(const int i; const int ilog2) {
+	if (i < 2) return i;
+	int a, b, c;
+	a = 1 << (((ilog2+1) >> 1) - 1);
+	b = 1 <<  ((ilog2+1) >> 1);
+	do {
+		c = (a + b) >> 1;
+		if (c * c >= i) b = c - 1;
+		else            a = c + 1;
+	} while (b > a);
+	return (a * a >= i) ? a : a + 1;
+}
+
+/*
  *  getfactors
  *
  *  Function:   - factorize a number
@@ -221,6 +261,7 @@ getfactors(int num, int *nfactors, int **factors) {
     int d;
     int i;
     int sqrtnum;
+    int factors2;
 
     if(num  < 2) {
         (*nfactors) = 0;
@@ -228,8 +269,9 @@ getfactors(int num, int *nfactors, int **factors) {
         return MPI_SUCCESS;
     }
     /* Allocate the array of prime factors which cannot exceed log_2(num) entries */
-    sqrtnum = ceil(sqrt(num));
-    size = ceil(log(num) / log(2));
+    size = upper_ilog2(num);
+//  sqrtnum = ceil(sqrt(num));
+//  size = ceil(log(num) / log(2));
     *factors = (int *) malloc((unsigned) size * sizeof(int));
 
     i = 0;
@@ -238,6 +280,8 @@ getfactors(int num, int *nfactors, int **factors) {
         num /= 2;
         (*factors)[i++] = 2;
     }
+
+    sqrt_num = upper_isqrt(num, size - i);
     /* determine all occurences of uneven prime numbers up to sqrt(num) */
     d = 3;
     for(d = 3; (num > 1) && (d < sqrtnum); d += 2) {


ibm-ompi:

 Gist: https://gist.github.com/5f3971342dc972dde15ec9b28c4dccd5

edgargabriel:
So is the goal to avoid linking the math library in ompi? If yes, at least the ceil and floor functions are also used in other location, i.e. ompio.

ibm-ompi:

 Gist: https://gist.github.com/810361dd0a410dac52f3aab419e4cc04

bwbarrett:
I have to say, this doesn't feel like the right solution to the problem.  Yes, you removed one instance, but you haven't done anything to make sure that new instances aren't added later (and the community generally has been ok linking in libm and its functions).

DrHaraldBraun:
the goal was and is not to avoid linking to libm (or libimf) in general, but to avoid having symbols from there statically linked into libmpi (only 'log', actually). The others are not necessary. However there is no Need to call an fp-math function where only integer is needed. Another Option would be to allocate sizeof(int)*8 integers to hold the prime factors...

rhc54:
I'm still not sure I understand the root problem here. Statically linking in the symbols doesn't sound like the issue. Are you building OMPI against one compiler, and then compiling your app using a different one? Otherwise, the 'log' symbol will resolve to the same thing, so I fail to see the problem.

rhc54:
Yeah, that is something we absolutely do _not_ support. You cannot build the app with a different compiler as it causes all kinds of problems, not just this one.

jsquyres:
FWIW, a few things:

1. I confirm that `icc` 2017-17.0.0.098 exhibits this behavior.
1. I confirm that `icc` 2016-16.0.4.258 exhibits this behavior.
1. I confirm that `icc` 2015-15.0.3.187 exhibits this behavior.
   * I did not test farther back than this
1. I confirm that `gcc` 5.2 does not create a public `log` symbol in `libmpi.so`.
1. Putting a public `log` symbol in a library does seem like bad behavior.  But given that this behavior stretches back so far, it certainly sounds like intended behavior.  The Intel compiler team should probably be consulted.
1. That being said, what @bwbarrett and @rhc54 said was correct: we don't really support building Open MPI with one compiler and building applications with another.  This *sometimes* works (and Intel actually tends to go to great pains to make this work), but our experience is that it can lead to weird problems like exactly the one you ran in to.  Hence, we don't advise it.  Build Open MPI and your application with the same compiler, and such problems tend to disappear.

All this being said, you did some excellent testing -- this should definitely be reported to the Intel compiler team and see what they say.

bwbarrett:
@jsquyres just to be clear, I didn't say that we shouldn't support building Open MPI with one compiler and the application with another.  I actually think we should support that case.

My problem with this PR is that it 1) likely doesn't properly fix the issue (there are other places we will cause libm to be pulled in) and 2) doesn't do anything to make sure that we don't have the problem again.  It doesn't remove linking against libm.  It doesn't provide compatibility across the entire code base.  It doesn't #define log to ompi_log or add a test that will catch this.  In other words, it's a patch that is only likely to work for one or two weeks, and then be broken again.

jsquyres:
@bwbarrett 👍  on all that you say.  My main point was that I think it's anti-social for `icc` to define a public `log` symbol in our library, and the Intel compiler team should be consulted about this behavior.

bwbarrett:
yes, agree that it seems like icc is doing something unfortunate and should stop :).

DrHaraldBraun:
Yes, I agree, too. My tests also confirm that this appears to be a 'special' feature of icc, which should be addressed. Unfortunately, any fix to icc would usually be applied to its next release only.

And yes, this PR does not prevent other symbols or occurrences of e.g. 'log' to be linked into a shared object statically at all -- it just fixes the very occurrence, I happend to run into.
Toward peventing symbols to be exported by ompi's shared objects that were inadvertently imported from other libs: Would it be possible to build an export-list (libtool -export-symbols / -export-symbols-regex feature)? 

Regarding the concern about mixing (otherwise compatible) compiler versions in mpi and application:
Although following this policy to always have the same compiler in libs and applications, I could still face this issue, if I was to force icc to use libm instead of libifm when building an application.
However, it would still be a nice and convenient feature to be able to mix compiler versions without risking symbols to be selected from strange origins at link- or run-time, since otherwise software-trees would either be bloated or rather minimal by only providing few arbitrarily chosen sets of compilers and libs.

Last but not least, I think invoking 'ceil(log(x)/log(2))' for the purpose of finding the index (+1) of the most significant bit set in an integer value is not actually necessary. An alternative approach would be a stack-allocation of 'sizeof(int)*CHAR_BIT - 1' integers (usually 124 Bytes) to hold the prime factors, which should be perfectly affordable.

hjelmn:
We actually already have a function to find the highest bit set. We probably should be using that. See opal/util/bit_ops.h

jsquyres:
@DrHaraldBraun Have you looked in the rest of the code base to see where we use `log()`, `sqrt()`, and `ceil()`?  I.e., is it a small number of places, or a lot?  Also, have you contacted the Intel compiler people to see if this is a known issue -- e.g., is there a compiler flag that can be used to avoid this issue?

DrHaraldBraun:
@jsquyres I see just a few references I see (in .h and .c files):

1. log(): <10 references in dims_create.c and several libnbc-related. All of which do the 'ceil(log(x)/log(2)' thing. There is another function named 'log' in libevent, but this is obviously not computing a logarithm.
2. sqrt(): <5 references, but only 2 of relevance: dims_create.c and io_ompio_aggregators.c (`P_x = P_y = (int) sqrt(P);`)
3. ceil(): <20 references (however, ceil() appear not to be that problematic)

DrHaraldBraun:
@jsquyres Usually, the icc option '-shared-intel' should already do the job ('-static-intel' is default, but behavior is slightly different, if '-shared' is provided, too).
My link-step command looks like this:

`/bin/sh $HOME/openmpi_build/ompi/libtool  --silent --tag CC --mode=link icc -std=gnu99 -DNDEBUG -g -shared-intel -m64 -xCORE-AVX2 -align -finline-functions -fno-strict-aliasing -restrict -Qoption,cpp,--extended_float_types -pthread -version-info 0:0:0 -g -shared-intel -m64 -xCORE-AVX2 -Wl,--enable-new-dtags -o libmpi.la -rpath $destination <list of .lo and .la files> -lrt -lutil -lz`

Suprisingly, '-shared-intel' is not passed on to the actual call to icc by 'libtool --mode=link', i.e.

`icc -std=gnu99 -shared  -fPIC -DPIC  <list of .o files>  -Wl,--whole-archive <list of .a files> -Wl,--no-whole-archive  -Wl,-rpath -Wl,$HOME/openmpi_build/ompi/orte/.libs -Wl,-rpath -Wl,$HOME/openmpi_build/ompi/opal/.libs -Wl,-rpath -Wl,$HOME/install/openmpi_master_intel18.0.1/lib -L$HOME/openmpi_build/ompi/opal/.libs $HOME/openmpi_build/ompi/orte/.libs/libopen-rte.so $HOME/openmpi_build/ompi/opal/.libs/libopen-pal.so -lm -lrt -lutil -lz  -m64 -pthread -m64 -Wl,--enable-new-dtags   -pthread -Wl,-soname -Wl,libmpi.so.0 -o .libs/libmpi.so.0.0.0`

The ld-call issued by icc reads:

`ld    /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtbeginS.o --eh-frame-hdr --build-id --hash-style=both -shared -m elf_x86_64 -L$HOME/openmpi_build/ompi/opal/.libs -o .libs/libmpi.so.0.0.0 -L/sw/rhel6-x64/intel/intel-18.0.1/mkl/lib/intel64 -L/mnt/lustre01/sw/rhel6-x64/intel/intel-18.0.1/compilers_and_libraries_2018.1.163/linux/compiler/lib/intel64_lin -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/ -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/ -L/lib/../lib64 -L/lib/../lib64/ -L/usr/lib/../lib64 -L/usr/lib/../lib64/ -L/sw/rhel6-x64/intel/intel-18.0.1/mkl/lib/intel64/ -L/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../ -L/lib64 -L/lib/ -L/usr/lib64 -L/usr/lib -rpath /sw/rhel6-x64/intel/intel-18.0.1/lib/intel64 -rpath /sw/rhel6-x64/intel/intel-18.0.1/mkl/lib/intel64 <list of .o files> --whole-archive <list of .a files> --no-whole-archive -rpath $HOME/openmpi_build/ompi/orte/.libs -rpath $HOME/openmpi_build/ompi/opal/.libs -rpath $HOME/install/openmpi_master_intel18.0.1/lib $HOME/openmpi_build/ompi/orte/.libs/libopen-rte.so $HOME/openmpi_build/ompi/opal/.libs/libopen-pal.so -L/mnt/lustre01/sw/rhel6-x64/intel/intel-18.0.1/compilers_and_libraries_2018.1.163/linux/compiler/lib/intel64_lin -Bstatic -limf -Bdynamic -lm -lrt -lutil -lz --enable-new-dtags -soname libmpi.so.0 -Bdynamic -limf -lsvml -lirng -lm -lipgo -ldecimal --as-needed -lcilkrts -lstdc++ --no-as-needed -lgcc -lgcc_s -lintlc -lpthread -lsvml -lc -lgcc -lgcc_s -lirc_s -ldl -lc /usr/lib/gcc/x86_64-redhat-linux/4.4.7/crtendS.o /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../lib64/crtn.o`

**There is a '-Bstatic -limf'.** . Providing '-Wc,-shared-intel' in the libtool call (or at configure-time) does the job. So icc works 'as designed/intended'.

Well, this was not my first suspect...

jsquyres:
I'm afraid that in the typical bustle of November/December (SC, Thanksgiving, MPI Forum meeting, Christmas, etc.), I've lost the context here.

You're saying that "icc works 'as designed/intended'."

Are you saying that there's nothing for us to do here?  Or that Intel is specifically having icc be anti-social (from our point of view -- earlier in this thread)?

I.e., do we need to have `configure` figure out that it needs to pass `-Bstatic -limf` for (at least some versions of) `icc`?

DrHaraldBraun:
I'm afraid this got almost lost due to seasonal bustle and "virus updates".

To my understanding, we are observing an unfortunate interplay between `libtool`, `icc`, and eventually `ld`:

1. `icc` by default links Intel-libraries statically, in case of `icc -shared` at least the math-library is linked statically (i.e. `-Bstatic -limf`  is passed to `ld`)
2. To obtain the desired behavior -- linking dynamically against libimf (i.e. `-Bdynamic -limf` is passed to `ld`,  it's recommended to use the `-shared-intel` option (which I usually provide when building shared libs using icc/ifort).
3. however, libtool does not recognize `-shared-intel` (nor `-static-intel`) as a valid linker-option, and hence does not pass it on to `icc` on linkage, which in consequence leads to an `-Bstatic -limf` in the subsequent `ld` call...

Adding `-Wc,-shared-intel` to compile-/link-flags at `configure` helps.
If we want `libimf` always to be linked in dynamically for `icc`-builds, maybe `configure` could add `-Wc,-shared-intel` to the respective flags, if Intel Compilers supporting  the `-shared-intel` option are detected... or at least to add `-Wc,...`, if the user provides `-shared-intel` in the `CFLAGS` (etc.)

jsquyres:
I just re-found this long-dormant issue.

After re-reading the thread, I'm not sure we want to have `configure` change the default behavior for how icc links Intel compilers.  Is this worth a section in README and/or the FAQ, advising users of the issue and giving an example of how they could pass various flags on the `configure` command line to override icc's default behavior if desired?

PR #4524, opened 2017-11-22
iof: optimize handling of stdin

ggouaillardet:
since some tasks migth end up having /dev/null as their stdin,
simply avoid pipe creation and destruction for these tasks.

From a pragmatic and MPI point of view, and unless explicitly required
otherwise, all MPI tasks but (the first) one end up with /dev/null
as their stdin.


rhc54:
Your base assumption is incorrect - if you check out the cmd line options for mpirun, you will see that we allow the user to specify any number of targets for stdin. There is no rule that only one process can receive it.

ggouaillardet:
i will reword the commit message.
since i test the `connect_stdin` flag on a per process, i believe the PR is still correct.

PR #4534, opened 2017-11-26
pml/cm: check for request comp. before completing bsend

hppritcha:
It turns out that for certain high performance OFI libfabric
providers (e.g. sockets) that support the fi_inject
interface,  they may upon a completion of a send operation invoke
the completer function for the ompi request.  The logic
around the bsend operation in the CM PML nevertheless was
invoking the ompi_request_complete a second time on the request,
causing much much horribleness, since the ompi_request_complete
blows up with seg faults if the  request is already marked
as complete.

Rather than modify the ompi_request_complete function to check
at the beginning whether a request is complete, before diving
into a bunch of atomic ops, just do the check at the PML level.

It might be better from a software design point of view to
redo ompi_request_complete such that even in a multi-threaded
situation, it can properly handle the case where the request
argument is, in fact, already in completed state.

### DIFF
--- a/ompi/mca/pml/cm/pml_cm_sendreq.h
+++ b/ompi/mca/pml/cm/pml_cm_sendreq.h
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2006 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2015-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


ompi/mca/pml/cm/pml_cm_sendreq.h
```diff
@@ -382,28 +382,30 @@ do {                                                                    \
...
+                                  sendreq->req_tag,                                  \
+                                  &sendreq->req_send.req_base.req_convertor,         \
+                                  sendreq->req_send.req_send_mode,                   \
+                                  sendreq->req_blocking,                             \
+                                  &sendreq->req_mtl));                               \
+        if(OMPI_SUCCESS == ret &&                                                    \
+           sendreq->req_send.req_send_mode == MCA_PML_BASE_SEND_BUFFERED) {          \
+            sendreq->req_send.req_base.req_ompi.req_status.MPI_ERROR = 0;            \
+            if(!REQUEST_COMPLETE(&sendreq->req_send.req_base.req_ompi)) {            \
+                ompi_request_complete(&(sendreq)->req_send.req_base.req_ompi, true); \
```
bwbarrett: This needs a comment (something like `/* the isend call may have marked completion */`).

bwbarrett:
Also, the commit message is really confusing to me.  The title should probably mention completion, something like: _pml:cm: Check for completion before completing bsend_.  Then the description should describe the race in general terms, before giving the explicit details about how we found the bug.  It took me a while to figure out that all you were saying is that the OFI MTL can end up completing a buffered send before the isend() call returns.

hppritcha:
updated per PR comments

PR #4517, opened 2017-11-20
ppc/asm: Fix opal_atomic_wmb definition

jjhursey:
* Fix typo in the `opal_atomic_wmb` declaration.
 * Fix lingering `eieio` reference in the XL assembly to be `lwsync`

Signed-off-by: Joshua Hursey <jhursey@us.ibm.com>

PR #4545, opened 2017-11-28
ppc/asm: Fix opal_atomic_wmb definition

jjhursey:
* Fix lingering `eieio` reference in the XL assembly to be `lwsync`


jjhursey:
Ref PR #4517

PR #4548, opened 2017-11-28
Add a new OMPI rte component to support direct-launch using PMIx.

rhc54:
Cleanup several places where abstraction violations crept into OMPI layer (direct reference of ORTE). Add some missing includes that were exposed by this change.

Note that this compiles, but I haven't tested it for execution yet. Handing it over to @npe9 for completion


rhc54:
Should note: this new component uses the opal/pmix framework. Thus, it will support any PMI-based environment, not just PMIx.

rhc54:
looks like aws/jenkins has jammed again

jjhursey:
Very nice! Thanks!

PR #4550, opened 2017-11-30
fix communicator's c_lock usage

ggouaillardet:
- initialize c_lock in the ompi_communicator_t constructor
 - USE_OPAL_THREAD_[UN]LOCK(c_lock)
 - #ifdef USE_MUTEX_FOR_COMMS protect c_lock access

### DIFF
--- a/ompi/communicator/comm.c
+++ b/ompi/communicator/comm.c
@@ -1315,16 +1315,12 @@ int ompi_comm_compare(ompi_communicator_t *comm1, ompi_communicator_t *comm2, in
 int ompi_comm_set_name (ompi_communicator_t *comm, const char *name )
 {
 
-#ifdef USE_MUTEX_FOR_COMMS
     OPAL_THREAD_LOCK(&(comm->c_lock));
-#endif
     memset(comm->c_name, 0, MPI_MAX_OBJECT_NAME);
     strncpy(comm->c_name, name, MPI_MAX_OBJECT_NAME);
     comm->c_name[MPI_MAX_OBJECT_NAME - 1] = 0;
     comm->c_flags |= OMPI_COMM_NAMEISSET;
-#ifdef USE_MUTEX_FOR_COMMS
     OPAL_THREAD_UNLOCK(&(comm->c_lock));
-#endif
 
     return OMPI_SUCCESS;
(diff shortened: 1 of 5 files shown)


ggouaillardet:
@kawashima-fj FYI, i observed a hang on Linux/sparc in `pthread_mutex_lock()`, the root causes are
 1. communicator's `c_lock` is not initialized
 2. `pthread_mutex_lock()` should have been #ifdef'ed out because `USE_MUTEX_FOR_COMMS` is not defined.

@jsquyres @edgargabriel @bwbarrett @bosilca @hjelmn that looks like very old code (which is  present in the `v2.0.x` branch, and probably older branches too). today, it seems there is no way the `USE_MUTEX_FOR_COMMS` macro can be defined, so that means there is a bunch of dead code that could/should be removed. Also, the `c_lock` element from the `ompi_communicator_t` struct could be removed or protected with `#ifdef USE_MUTEX_FOR_COMMS`.

Do you have any objections against me removing the dead code ? shall i also remove the `c_lock` element from the `ompi_communicator_t` struct ?

bosilca:
The c_lock is needed in a multi-threaded case to protect the communicator from user abuses. It is mainly needed for functions that change the user readable state of the communicator, such as error handler and name. I don't see how you can remove it without losing important capabilities.

I don't see a need for USE_MUTEX_FOR_COMMS. This macro can be removed and have the communicator always contain the c_lock field.

ggouaillardet:
That makes sense !
just to be clear, the `USE_MUTEX_FOR_COMMS` macro is currently __not__ defined and `c_lock` is currently __not__ initialized. I understand now this is a bug.
i will remove the `USE_MUTEX_FOR_COMMS` macro so `c_lock` is always used so the multi-threaded case is properly handled.

ibm-ompi:

 Gist: https://gist.github.com/92fb90595bc24f5a88313359d6354ef0

ibm-ompi:

 Gist: https://gist.github.com/0a6d36dda26c23784858b38ee03ac8d6

ibm-ompi:

 Gist: https://gist.github.com/f55ee4412d11c4814520fe5d3f228a06

PR #4551, opened 2017-11-30
Make usage of ompi_communicator_t, ompi_file_t and ompi_win_t mutex consistent

ggouaillardet:

### DIFF
--- a/ompi/communicator/comm_init.c
+++ b/ompi/communicator/comm_init.c
@@ -465,6 +465,8 @@ static void ompi_comm_destruct(ompi_communicator_t* comm)
         opal_pointer_array_set_item ( &ompi_comm_f_to_c_table,
                                       comm->c_f_to_c_index, NULL);
     }
+
+    OBJ_DESTRUCT(&comm->c_lock);
 }
 
 #define OMPI_COMM_SET_INFO_FN(name, flag)       \
--- a/ompi/file/file.c
+++ b/ompi/file/file.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2008-2009 Sun Microsystems, Inc.  All rights reserved.
  * Copyright (c) 2009-2012 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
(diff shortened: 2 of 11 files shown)


bosilca:
I am not sure I follow the logic here. We only need the protection in the multi-threaded case, which would justify the use of OPAL_THREAD_LOCK instead of opal_thread_lock.

ggouaillardet:
@hjelmn can you please comment on that ?

i was mimicking open-mpi/ompi@6b68d1cfc879f38e1a8b2475c4aca2be47202f1b and in this commit you chose to use `opal_mutex_[un]lock()` instead of `OPAL_THREAD_[UN]LOCK()`

ggouaillardet:
incidentally, i noted
 - `ompi_communicator_t` has a `c_lock`
- `ompi_win_t` has a `w_lock`
- `ompi_file_t` has a `f_mutex`

all are `opal_mutex_t`

also, the `OPAL_THREAD_LOCK()` macro invokes `opal_mutex_lock()`

shall i make everything consistent and use only `lock` or `mutex` ?

for example, `OPAL_THREAD_LOCK(&comm->c_lock)` could become `OPAL_MUTEX_LOCK(&comm->c_mutex)`, or `OPAL_THREAD_LOCK(&file->f_mutex)` could become `OPAL_THREAD_LOCK(&file->f_lock)`

hjelmn:
I used the non-conditional lock because the set error handler path is not in the critical path. It could just have easily been OPAL_THREAD_LOCK().

bosilca:
@ggouaillardet makes perfect sense to rename f_mutex into f_lock to match the rest of the components.

ggouaillardet:
@bosilca i updated this PR is order to make everything consistent
 - the `opal_mutex_t` is always called `?_lock`
 - `OPAL_THREAD_[UN]LOCK()` is always used instead of `opal_mutex_[un]lock()`
 - `ompi_communicator_t`'s `c_lock` is now destroyed in the destructor

bosilca:
this looks good 👍 @edgargabriel any feedback about the file changes ?

edgargabriel:
looks good 👍

PR #4552, opened 2017-11-30
Add atomic fetch-and-op and compare-exchange functions

hjelmn:
This PR contains the following updates to atomics in opal:

 - Remove opal_atomic_cmpset_* operations.

 - Add new compare-exchange functions with signature that matches C11. These functions still return a boolean but take a pointer to the compare value. The old value is stored over the compare value.

 - Add atomic fetch-and-op functions. These replace the old op-and-fetch functions in the underlying atomic implementations. The old atomic arithmetic operations have been renamed from opal_atomic_op_* to opal_atomic_op_fetch_*.


hjelmn:
:bot:retest:ompi

PR #4554, opened 2017-11-30
configure: check for C11 and atomic types

hjelmn:
This commit updates the configure code for Open MPI to check for C11
support. The features requested are: atomics and thread local
storage.

References #3879

### DIFF
--- a/config/opal_setup_cc.m4
+++ b/config/opal_setup_cc.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2006 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2007-2009 Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
-dnl Copyright (c) 2012      Los Alamos National Security, LLC. All rights
+dnl Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
 dnl                         reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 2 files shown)


hjelmn:
This PR just adds C11 checks. Open MPI does not currently require C11 but we may want to use some of the features if they are available.

config/opal_setup_cc.m4
```diff
@@ -41,14 +154,7 @@ AC_DEFUN([OPAL_SETUP_CC],[
...
     AC_SUBST([WRAPPER_CC])
 
-    # From Open MPI 1.7 on we require a C99 compiant compiler
-    AC_PROG_CC_C99
-    # The result of AC_PROG_CC_C99 is stored in ac_cv_prog_cc_c99
-    if test "x$ac_cv_prog_cc_c99" = xno ; then
-        AC_MSG_WARN([Open MPI requires a C99 compiler])
-        AC_MSG_ERROR([Aborting.])
-    fi
-
```
jsquyres: I kinda like the C99 check out here, independent of the C11 check.

Yes, it's (minorly) redundant with the C11 checks -- but the code is very clear this way.

Also -- it might be worth updating the `AC_MSG_WARN` message to say something like "Open MPI requires a C99 (or later) compiler".

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,119 @@ dnl
...
+#endif
+                                           ]],[])],
+                          [opal_cv_c11_flag_required=no])
+
+        AC_MSG_RESULT([$opal_cv_c11_flag_required])
+
+        if test "x$opal_cv_c11_flag_required" = "xno" ; then
+            AC_MSG_NOTICE([verifying $CC supports C11 without a flag])
+            OPAL_PROG_CC_C11_HELPER([], [], [opal_cv_c11_flag_required=yes])
+        fi
```
jsquyres: What's this check for?  It seems to be redundant with the above test...?

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,119 @@ dnl
...
+            opal_cv_c11_flag=
+            for flag in $(echo $opal_prog_cc_c11_flags | tr ' ' '\n') ; do
+                OPAL_PROG_CC_C11_HELPER([$flag],[opal_cv_c11_flag=$flag],[])
+                if test "x$opal_cv_c11_flag" != "x" ; then
+                    CFLAGS="$CFLAGS $opal_cv_c11_flag"
+                    AC_MSG_NOTICE([using $flag to enable C11 support])
+                    opal_cv_c11_supported=yes
+                    break
+                fi
+            done
```
jsquyres: Might want to put a comment here that it's (at least for now) not an error if C11 is not supported.

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,119 @@ dnl
...
+    opal_prog_cc_c11_helper_CFLAGS_save=$CFLAGS
+    CFLAGS="$CFLAGS $1"
+
+    OPAL_CC_HELPER([if $CC $1 supports C11 thread local storage], [opal_prog_cc_c11_helper__thread_local_available],
+                   [],[[static _Thread_local int  foo = 1;++foo;]])
+
+    OPAL_CC_HELPER([if $CC $1 supports C11 atomic variables], [opal_prog_cc_c11_helper_atomic_var_available],
+                   [[#include <stdatomic.h>]], [[static atomic_long foo = 1;++foo;]])
+
+    OPAL_CC_HELPER([if $CC $1 supports C11 _Atomic keyword], [opal_prog_cc_c11_helper__atomic_available],
```
jsquyres: Should this be `opal_prog_cc_c11_helper__Atomic_available` (capital A)?  Ditto for Generic, Static.

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,119 @@ dnl
...
+    OPAL_CC_HELPER([if $CC $1 supports C11 _Generic keyword], [opal_prog_cc_c11_helper__generic_available],
+                   [[#define FOO(x) (_Generic (x, int: 1))]], [[static int x, y; y = FOO(x);]])
+
+    OPAL_CC_HELPER([if $CC $1 supports C11 _Static_assert], [opal_prog_cc_c11_helper__static_assert_available],
+                   [[#include <stdint.h>]],[[_Static_assert(sizeof(int64_t) == 8, "WTH");]])
+
+    AS_IF([test "x$opal_prog_cc_c11_helper_atomic_var_available" = "xyes"], [opal_prog_cc_c11_helper_tmp=1],
+          [opal_prog_cc_c11_helper_tmp=0])
+
+    AC_DEFINE_UNQUOTED([OPAL_C_HAVE_C11_ATOMIC], [$opal_prog_cc_c11_helper_tmp], [Whether C compiler supports C11 atomics])
```
jsquyres: Why not put this `AS_IF` and `AC_DEFINE_UNQUOTED` in the `OPAL_CC_HELPER`?

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,119 @@ dnl
...
+          [opal_prog_cc_c11_helper_tmp=0])
+
+    AC_DEFINE_UNQUOTED([OPAL_C_HAVE__ATOMIC], [$opal_prog_cc_c11_helper_tmp], [Whether C compiler supports __Atomic])
+
+    AS_IF([test "x$opal_prog_cc_c11_helper__thread_local_available" = "xyes"], [opal_prog_cc_c11_helper_tmp=1],
+          [opal_prog_cc_c11_helper_tmp=0])
+
+    AC_DEFINE_UNQUOTED([OPAL_C_HAVE__THREAD_LOCAL], [$opal_prog_cc_c11_helper_tmp], [Whether C compiler supports __Thread_local])
+
+    dnl At this time Open MPI only requires thread local and the atomic convenience types
```
jsquyres: Open MPI *does not* require these things (yet).

config/opal_setup_cc.m4
hjelmn: Some compilers (intel) lie about support. They claim to be 201112 but are missing something. In the case of Intel < 17 it is the atomics.

config/opal_setup_cc.m4
hjelmn: Ok, moving it now.

config/opal_setup_cc.m4
hjelmn: Added.

config/opal_setup_cc.m4
hjelmn: Fixed

hjelmn:
Ok, think it should be good to go.

config/opal_setup_cc.m4
jsquyres: Might want to put this in a comment -- lest Future Jeff forget this conversation and ask you about it again someday... 😄

config/opal_setup_cc.m4
```diff
@@ -23,6 +23,105 @@ dnl
...
+    OPAL_VAR_SCOPE_PUSH([opal_prog_cc_c11_flags])
+    if test -z "$opal_cv_c11_supported" ; then
+        opal_cv_c11_supported=no
+        opal_cv_c11_flag_required=yes
+
+        AC_MSG_CHECKING([if $CC requires a flag for C11])
+
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#if __STDC_VERSION__ < 201112L
+#error "Need a flag for C11"
```
jsquyres: This string -- even though it is irrelevant -- isn't quite right.  If you don't have a C11 compiler, then no flag will help you.  Perhaps `#error "Without any CLI flags, this compiler does not support C11"`

config/opal_setup_cc.m4
```diff
@@ -41,14 +140,25 @@ AC_DEFUN([OPAL_SETUP_CC],[
...
     AC_SUBST([WRAPPER_CC])
 
-    # From Open MPI 1.7 on we require a C99 compiant compiler
-    AC_PROG_CC_C99
-    # The result of AC_PROG_CC_C99 is stored in ac_cv_prog_cc_c99
-    if test "x$ac_cv_prog_cc_c99" = xno ; then
-        AC_MSG_WARN([Open MPI requires a C99 compiler])
-        AC_MSG_ERROR([Aborting.])
-    fi
+    OPAL_PROG_CC_C11
```
jsquyres: Is it *guaranteed* that if `$opal_cv_c11_supported==yes`, you won't need to call `AC_PROG_CC_C99`?

hjelmn: Yes. If c11 support is detected then c99 support must exist as well. So there is no need to check c99.

config/opal_setup_cc.m4
hjelmn: OK. Well change the string

hjelmn:
@jsquyres  Updated based on your comments.

hjelmn:
:bot:retest:ompi

hjelmn:
I can't figure out why the gcc49 build is failing. It is failing in the initial compiler check well before any of the changes in this PR come up. I tested locally and both with Linux and macOS gcc-4.9 is configuring correctly. @bwbarrett Is that on an Amazon system? Can you see why it is failing?

hjelmn:
@jsquyres Can we make CI cat config.log if the configure step fails? It would help in these situations.

bwbarrett:
bot:ompi:retest

bwbarrett:
I somehow removed the gcc-4.9 compiler from the ubuntu image.  Builds should be working now.  And to answer your config.log question, we welcome a PR.

hjelmn:
@bwbarrett Forgot that that is in a repo. Will open a PR.

ggouaillardet:
@hjelmn @bwbarrett i just filled open-mpi/ompi-scripts#5 for that purpose

hjelmn:
@ggouaillardet Beat me to it. Thanks for adding that to the scripts. It will help a lot when we are making autotool script changes.

hjelmn:
@jsquyres I made the requested change. Are you ok with merging this now?

PR #4556, opened 2017-12-01
Add a new posix_spawn component to the ODLS framework.

rhc54:
Only selectable when specifically requested via "-mca odls pspawn"

Note that there are several concerns:
  * we aren't getting SIGCHLD calls when the procs terminate
  * we aren't seeing the IO pipes close on termination, though
    we are getting output forwarded to mpirun
  * I haven't found a way to bind the child process prior to exec.
    If we want to use this method, we probably need someone to
    implement a cgroup component for the orte/rtc framework

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -474,6 +474,7 @@ orte/test/system/orte_sensor
 orte/test/system/event-threads
 orte/test/system/test-time
 orte/test/system/psm_keygen
+orte/test/system/pspawn
 orte/test/system/regex
 orte/test/system/orte_errors
 orte/test/system/evthread-test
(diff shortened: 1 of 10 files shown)


rhc54:
@bosilca @abouteiller @jjhursey @ggouaillardet 

This implementation isn't complete - someone needs to at least resolve the SIGCHLD and IO termination issues. In addition, we would need to resolve the binding question.

I will say that I did find it faster to spawn the proc this way, so if someone cares to help resolve the problems, it might be an attractive option. I probably won't have time to continue this for the next few weeks at least.

ggouaillardet:
@rhc54 i pushed open-mpi/ompi@a4755b694bb001d852daf57218ff7941a0592a4f
now i can `mpirun hello_c` and it does not hand

rhc54:
oh, that is embarrassing - thanks!

PR #4559, opened 2017-12-01
io/ompio: introduce a new function to retrieve mca parameter values

edgargabriel:
ompio has the unique problem, that mca parameters set in the io/ompio component
have to be accessible from other frameworks as well. This is mostly done to avoid
a replication in the parameter names and to reduce the number of mca parameters that
and end-user has to worry about.

This commit introduces a generic function to retrieve ompio mca parameters, the function pointer
is stored on the file handle. It replaces two functions that used the same concept already for
one parameter each.


edgargabriel:
Hm, everything finished successfully except for the ARM build, and not sure what went wrong there

open-mpi.build.platforms » ARMv8 completed with result ABORTED

Should I hold back merging because of that?

edgargabriel:
I decided to go ahead an merge it.

PR #4561, opened 2017-12-02
Create a separate VM hole finder utility

rhc54:
Pull the vm hole search routine into a separate function so it can be accessed by multiple parties to avoid conflicting assignments. We currently use this method for HWLOC shared memory interface, and are exploring its use for PMIx as well.

### DIFF
--- a/opal/runtime/opal_finalize.c
+++ b/opal/runtime/opal_finalize.c
@@ -36,6 +36,7 @@
 #include "opal/util/proc.h"
 #include "opal/util/keyval_parse.h"
 #include "opal/util/show_help.h"
+#include "opal/util/vmtracker.h"
 #include "opal/memoryhooks/memory.h"
 #include "opal/mca/base/base.h"
 #include "opal/runtime/opal.h"
@@ -162,6 +163,9 @@ opal_finalize(void)
 
     /* cleanup the main thread specific stuff */
     opal_tsd_keys_destruct();
+
+    /* finalize the vmtracker */
+    opal_vmtracker_finalize();
 
     /* finalize util code */
     opal_finalize_util();
--- a/opal/runtime/opal_init.c
+++ b/opal/runtime/opal_init.c
@@ -40,6 +40,7 @@
 #include "opal/util/output.h"
 #include "opal/util/show_help.h"
 #include "opal/util/proc.h"
+#include "opal/util/vmtracker.h"
 #include "opal/memoryhooks/memory.h"
 #include "opal/mca/base/base.h"
 #include "opal/mca/base/mca_base_var.h"
@@ -636,6 +637,9 @@ opal_init(int* pargc, char*** pargv)
         error = "opal_cr_init";
         goto return_error;
     }
+
+    /* init the vmtracker */
+    opal_vmtracker_init();
 
     return OPAL_SUCCESS;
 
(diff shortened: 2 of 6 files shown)


rhc54:
@bgoglin Would you mind taking a look to see if I still have this right, especially when I shift to avoid overlapping a prior assignment? I tried to maintain alignment, but would appreciate confirmation.

bgoglin:
You need a Copyright line for Inria 2017 since I wrote lots of lines in there :)

The conflicting avoidance is only useful if two components call assign_address() before they both actually mmap() something.
I guess most cases will actually mmap() before the next assignment is requested. When we'll request the second assignment, the conflict will be auto-avoided by parsing /proc/self/mmaps again, since it contains the first assignment.
Also the second mmap() won't be near the first one at middle of the original biggest hole because that biggest hole didn't exist anymore after the first mmap().

I would rather:
* when parsing /proc/self/maps in assign_address(), ignore lines that correspond to an existing assignment
* once you found the biggest hole while ignoring your previous assignments, replace the middle by something above the last assignment

About the alignment:
0) the goal is to avoid the mapping spreading across two 2-level page table entries, because it'd waste page-table entries, and it may prevent hugepages from being used underneath
1) you don't need both POWER and x86 alignment code. The x86 one is useless if POWER already applied a larger alignment. I didn't have the necessary #ifdef in my original program. We just need a single alignment with the right value. But it depends on the architecture, and some archs support multiple page sizes. See below.
2) I am not sure we want to align for all assignments. Once we align the first assignment, I would just place the next assignments right after it. Maybe just keep a single empty page between them in case you want to catch overflows.
3) Given that there's no proper solution to (1) and given that (2) means we're not wasting virtual space between assignments anymore, I think I'd just use 1GB for (1) which will be a multiple of 2-level size in all cases.

In the end, I don't think you even need to remember all assignments. Just remember the beginning of the first one and the end of the last one. Ignore /proc/self/maps lines that are between them.

We don't really care about one assignment being released in the middle of the list, it wouldn't be reused in your current code anyway (because the hole isn't huge).

Sorry there are things that I should have reported earlier :/

rhc54:
I can't legally assign copyright for another organization - the original code didn't include a copyright, and so I can't just transfer it. Can you please push a commit here with your copyright notice?

I'll take a look at cleaning this up some more. I think I'll put a big thread lock around it, and do the mmap within the code as well to avoid the race condition (which is why I added the tracker). Fortunately, we won't call into this more than twice anyway, so the cost of parsing the file each time won't kill us.

hppritcha:
Looks like something got in to the code base (not this PR), that is causing make check to fail -
```
PASS: predefined_pad_test
../../../../config/test-driver: line 107: 17921 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: dlopen_test
```

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

rhc54:
Looks like this is being bit by the dl_open issue. No hurry to commit it, but it would be nice if we could figure out what is going on here.

bwbarrett:
bot:ompi:retest

not sure why all the compiler tests failed; sigh.

rhc54:
bot:ompi:retest

bwbarrett:
it looks like this is causing segfaults in the dss unit tests, which looks to be a real bug with this patch...

PR #4562, opened 2017-12-04
odls/base: fix handling of the odls_base_num_threads MCA param

ggouaillardet:
If a number of odls threads is explicitly required, then use
that number no matter what.

### DIFF
--- a/orte/mca/odls/base/odls_base_frame.c
+++ b/orte/mca/odls/base/odls_base_frame.c
@@ -153,13 +153,11 @@ void orte_odls_base_start_threads(orte_job_t *jdata)
     /* setup the pool of worker threads */
     orte_odls_globals.ev_threads = NULL;
     orte_odls_globals.next_base = 0;
-    if (0 == orte_odls_globals.num_threads ||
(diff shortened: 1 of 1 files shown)


ggouaillardet:

currently, if you `mpirun --mca odls_base_num_threads 1 -np 1 hostname`,
then zero odls threads are used (1 < cutoff), but `orte_odls_base_harvest_threads()` will try to harvest 1 thread anyway, and will crash.

my understanding is user should get the number of odls threads they ask for, regardless the `cutoff` or `max_threads` value.
i also fixed the bozo case where `max_threads` is zero.

rhc54:
Hmmm..everything built fine, but arm timed out.

bot:ompi:retest

ggouaillardet:
@rhc54 i pushed an other fix to this PR

currently, `mpirun` crashes when using `odls` progress threads and if oversubscription is needed but not requested.

ggouaillardet:
:bot:aws:retest

rhc54:
somehow, I very much doubt that this change causes the fortran compilers to be unable to create executables. It looks like Jenkins has gone over the edge again.

Issue #4564, opened 2017-12-04
posix_memalign gets called with zero alignment in some circumstances


benmenadue:
`posix_memalign` requires an alignment that is a power of two and a multiple of `sizeof(void *)`. However, there is a call to `mpool_alloc` (and thus `mca_mpool_default_alloc`) in `mca_mpool_base_alloc` that passes an alignment of 0 -- which isn't a power of two.

I haven't noticed this actually cause a crash under normal circumstances, but TotalView does catch the call and bail. Perhaps the code path is only taken when debugging?

This is present in master.

benmenadue:
I'm just testing a patch for this, will PR shortly if it compiles.

hjelmn:
Close when the commits in #4565 are in v2.x, v3.0.x, and v3.1.x.

jsquyres:
The fix has now been merged everywhere; closing this issue.

PR #4565, opened 2017-12-04
Use malloc instead of posix_memalign for small (<= sizeof(void *)) alignments

benmenadue:
This works around issue #4564.

### DIFF
--- a/opal/mca/mpool/base/mpool_base_alloc.c
+++ b/opal/mca/mpool/base/mpool_base_alloc.c
@@ -74,7 +74,7 @@ void *mca_mpool_base_alloc(size_t size, opal_info_t *info, const char *hints)
 
     mpool = mca_mpool_base_module_lookup (hints);
     if (NULL != mpool) {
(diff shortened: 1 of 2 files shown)


hjelmn:
Though I did not intend this function to called with an alignment smaller than the malloc alignment. Something is not working as intended but this fix is ok with me.

benmenadue:
The call that triggers it is line 77 of opal/mca/mpool/base/mpool_base_alloc.c:
```
mem = mpool->mpool_alloc (mpool, size, 0, 0);
```

hjelmn:
Yeah, that is a typo. Should have align == 8. Either fixing that typo or allowing a 0 alignment work for me.

benmenadue:
Okay, I've changed the alignment request to `sizeof(void *)` in `mca_mpool_base_alloc`. Other mpools already handle to zero-alignment case (e.g. memkind replaces it with the page size), so I'd suggest keeping the other patch as well.

hjelmn:
Ok, once CI completes I will merge this. I will open the PRs to the release branches.

Issue #4260, opened 2017-09-25
IMB tesuite fails when using vader


nmorey:
Using openmpi 2.1.2 and the Intel MPI Benchmark suite (https://software.intel.com/sites/default/files/managed/76/6c/IMB_2017_Update2.tgz) on x86 systems (multiple SUSE versions)

I get this error
```
mpirun -np 2  --mca btl vader,self /usr/lib/mpi/gcc/openmpi2/tests/IMB/IMB-MPI1
[snip...]
#-----------------------------------------------------------------------------
# Benchmarking Sendrecv 
# #processes = 2 
#-----------------------------------------------------------------------------
       #bytes #repetitions  t_min[usec]  t_max[usec]  t_avg[usec]   Mbytes/sec
            0         1000         0.60         0.60         0.60         0.00
            1         1000         0.59         0.59         0.59         3.38
            2         1000         0.55         0.55         0.55         7.31
            4         1000         0.62         0.62         0.62        12.81
            8         1000         0.64         0.64         0.64        24.94
           16         1000         0.76         0.76         0.76        41.94
           32         1000         0.56         0.56         0.56       114.88
           64         1000         0.64         0.64         0.64       200.60
          128         1000         0.65         0.65         0.65       396.01
          256         1000         1.10         1.10         1.10       463.57
          512         1000         1.49         1.50         1.49       684.71
         1024         1000         1.82         1.82         1.82      1122.39
         2048         1000         2.07         2.07         2.07      1979.64
         4096         1000         2.63         2.63         2.63      3113.39
         8192         1000         2.74         2.74         2.74      5986.21
        16384         1000         4.42         4.42         4.42      7410.65
[portia:25305] *** Process received signal ***
[portia:25305] Signal: Segmentation fault (11)
[portia:25305] Signal code: Address not mapped (1)
[portia:25305] Failing at address: 0x56dc0730
[portia:25305] [ 0] linux-gate.so.1(__kernel_rt_sigreturn+0x0)[0xf77bdf70]
[portia:25305] [ 1] /usr/lib/mpi/gcc/openmpi2/lib/openmpi/mca_btl_vader.so(mca_btl_vader_poll_handle_frag+0x133)[0xf5882e33]
[portia:25305] [ 2] /usr/lib/mpi/gcc/openmpi2/lib/openmpi/mca_btl_vader.so(+0x4251)[0xf5883251]
[...]

PR #4569, opened 2017-12-05
btl/vader: change the way fast boxes are used

hjelmn:
There were multiple paths that could lead to a fast box
allocation. One of them made little sense (in-place send) so it has
been removed to allow a rework of the fast-box send function. This
should fix a number of issues with hanging/crashing when using the
vader btl.

References #4260

### DIFF
--- a/opal/mca/btl/vader/btl_vader_fbox.h
+++ b/opal/mca/btl/vader/btl_vader_fbox.h
@@ -22,12 +22,12 @@ typedef union mca_btl_vader_fbox_hdr_t {
          * in multiple instructions. To ensure that seq is never loaded before tag
          * and the tag is never read before seq put them in the same 32-bits of the
          * header. */
+        /** message size */
+        uint32_t  size;
         /** message tag */
         uint16_t  tag;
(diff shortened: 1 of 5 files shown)


hjelmn:
:bot:ompi:retest

opal/mca/btl/vader/btl_vader_fbox.h
```diff
@@ -52,15 +52,19 @@ static inline void mca_btl_vader_fbox_set_header (mca_btl_vader_fbox_hdr_t *hdr,
...
 {
     const unsigned int fbox_size = mca_btl_vader_component.fbox_size;
+    size_t size = header_size + payload_size;
     unsigned int start, end, buffer_free;
     size_t data_size = size;
-    unsigned char *dst;
+    unsigned char *dst, *data;
     bool hbs, hbm;
 
     /* don't try to use the per-peer buffer for messages that will fill up more than 25% of the buffer */
```
derbeyn: @hjelmn  return NULL below (line 72) should be changed by a return false

PR #4572, opened 2017-12-05
v3.0.x: Use malloc instead of posix_memalign for small (<= sizeof(void *)) alignments

hjelmn:
See #4564

FYI @benmenadue


jsquyres:
See https://github.com/open-mpi/ompi/pull/4573#pullrequestreview-81588946

PR #4573, opened 2017-12-05
v3.1.x: Use malloc instead of posix_memalign for small (<= sizeof(void *)) alignments

hjelmn:
See #4564

FYI @benmenadue

### DIFF
--- a/opal/mca/mpool/base/mpool_base_alloc.c
+++ b/opal/mca/mpool/base/mpool_base_alloc.c
@@ -74,7 +74,7 @@ void *mca_mpool_base_alloc(size_t size, opal_info_t *info, const char *hints)
 
     mpool = mca_mpool_base_module_lookup (hints);
     if (NULL != mpool) {
-        mem = mpool->mpool_alloc (mpool, size, 0, 0);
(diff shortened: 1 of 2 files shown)


opal/mca/mpool/base/mpool_base_alloc.c
```diff
@@ -74,7 +74,7 @@ void *mca_mpool_base_alloc(size_t size, opal_info_t *info, const char *hints)
 
     mpool = mca_mpool_base_module_lookup (hints);
     if (NULL != mpool) {
-        mem = mpool->mpool_alloc (mpool, size, 0, 0);
+        mem = mpool->mpool_alloc (mpool, size, sizeof(void *), 0);
```
jsquyres: If `posix_memalign` requires a power of 2 for the alignment argument, what if `sizeof(void*)` == 4?

benmenadue: 4 is a power of two (2^2). But this will take the malloc code path anyway, since malloc is all you need for <= sizeof(void *) alignment.

jsquyres: Um.. duh.  Right.  /me goes to get more coffee.

PR #4574, opened 2017-12-05
New flag for MCA parameters that allows a behaving with a default value of "unset".

matcabral:
New flag for MCA parameters that allows a behaving with a default value of "unset".
mtl/psm2: Update some shadow mca parameters to use the default "unset".
mtl/psm2: Add new shadow parameter to allow specifying the service level.

    (cherry picked from commit 1fad59465f02522ec840c5e533babf2ba9cd9a9b)

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -92,29 +92,33 @@ struct ompi_mtl_psm2_shadow_variable {
     mca_base_var_info_lvl_t info_level;
     const char *mca_name;
     const char *description;
+    mca_base_var_flag_t flags;
 };
 
 struct ompi_mtl_psm2_shadow_variable ompi_mtl_psm2_shadow_variables[] = {
     {MCA_BASE_VAR_TYPE_STRING, &ompi_mtl_psm2.psm2_devices, {.stringval = "self,shm,hfi"}, "PSM2_DEVICES", OPAL_INFO_LVL_3,
-     "devices", "Comma-delimited list of PSM2 devices. Valid values: self, shm, hfi (default: self,shm,hfi)"},
+     "devices",
+     "Comma-delimited list of PSM2 devices. Valid values: self, shm, hfi (default: self,shm,hfi. Reduced to self,shm in single node jobs)",0},
     {MCA_BASE_VAR_TYPE_STRING, &ompi_mtl_psm2.psm2_memory, {.stringval = "normal"}, "PSM2_MEMORY", OPAL_INFO_LVL_9,
-     "memory_model", "PSM2 memory usage mode (default: normal)"},
-    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_sendreqs_max, {.ulval = 1048576}, "PSM2_MQ_SENDREQS_MAX", OPAL_INFO_LVL_3,
-     "mq_sendreqs_max", "PSM2 maximum number of isend requests in flight (default: 1M)"},
-    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_recvreqs_max, {.ulval = 1048576}, "PSM2_MQ_RECVREQS_MAX", OPAL_INFO_LVL_3,
-     "mq_recvreqs_max", "PSM2 maximum number of irecv requests in flight (default: 1M)"},
-    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_rndv_hfi_threshold, {.ulval = 64000}, "PSM2_MQ_RNDV_HFI_THRESH", OPAL_INFO_LVL_3,
-     "hfi_eager_limit", "PSM2 eager to rendezvous threshold (default: 64000)"},
-    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_rndv_shm_threshold, {.ulval = 16000}, "PSM2_MQ_RNDV_SHM_THRESH", OPAL_INFO_LVL_3,
-     "shm_eager_limit", "PSM2 shared memory eager to rendezvous threshold (default: 16000)"},
+     "memory_model", "PSM2 memory usage mode. Valid values: min, normal, large (default: normal)", 0},
+    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_sendreqs_max, {.ulval = 0}, "PSM2_MQ_SENDREQS_MAX", OPAL_INFO_LVL_3,
+     "mq_sendreqs_max", "PSM2 maximum number of isend requests in flight (default: unset, let libpsm2 use its default)", MCA_BASE_VAR_FLAG_DEF_UNSET},
+    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_recvreqs_max, {.ulval = 0}, "PSM2_MQ_RECVREQS_MAX", OPAL_INFO_LVL_3,
+     "mq_recvreqs_max", "PSM2 maximum number of irecv requests in flight (default:  unset, let libpsm2 use its default)", MCA_BASE_VAR_FLAG_DEF_UNSET},
+    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_rndv_hfi_threshold, {.ulval = 0}, "PSM2_MQ_RNDV_HFI_THRESH", OPAL_INFO_LVL_3,
+     "hfi_eager_limit", "PSM2 eager to rendezvous threshold (default: unset, let libpsm2 use its defaults)", MCA_BASE_VAR_FLAG_DEF_UNSET},
+    {MCA_BASE_VAR_TYPE_UNSIGNED_LONG, &ompi_mtl_psm2.psm2_mq_rndv_shm_threshold, {.ulval = 0}, "PSM2_MQ_RNDV_SHM_THRESH", OPAL_INFO_LVL_3,
+     "shm_eager_limit", "PSM2 shared memory eager to rendezvous threshold (default: unset, let libpsm2 use its default)", MCA_BASE_VAR_FLAG_DEF_UNSET},
     {MCA_BASE_VAR_TYPE_BOOL, &ompi_mtl_psm2.psm2_recvthread, {.boolval = true}, "PSM2_RCVTHREAD", OPAL_INFO_LVL_3,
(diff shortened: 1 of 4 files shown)


matcabral:
@jsquyres: 
 - Sentinel vs unset: I actually started with the sentinel approach which upfront looks simpler. However, that implies you'll not only need a different sentinel value of each type in mca_base_var_storage_t (which also leaves out boolval) but also that will leave everybody to use potentially different values for each possible parameter. Moreover, how are these values shown by ompi_info?  I found that using a flag addresses all these issues, with the existence of MCA_BASE_VAR_SOURCE_DEFAULT the implementation was quite straight forward. 

- Splitting the PR: Sure, I admit I bundled too many things together. How to manage it if this commit already made into master ?

jsquyres:
Don't forget that MCA vars can also be queried to see where they were set.  I.e., if they user didn't set any value (via env var, CLI, or config file), you can know that.

I'm mainly saying: there is a bunch of momentum in the code base for the sentinel approach.  The specific value doesn't always matter if you check to see where the MCA var was set (i.e., if it's the default value or not).

matcabral:
>We have a bunch of other MCA params that take a sentinel value (e.g., -1) to indicate that the user has not set them. 

>Don't forget that MCA vars can also be queried to see where they were set. I.e., if they user didn't set any value (via env var, CLI, or config file), you can know that.

That is clear, in fact I do use MCA_BASE_VAR_SOURCE_DEFAULT in conjunction with the new flag to make decision. The one thing would like make sure we are on the same page is with the behavior intended with the flag, which to essentially ignore the default value (and do nothing), not to understand the where it as set (this is because the underlying library has an already defined algorithm to define the value which is not worth replicating in OMPI and maintaining). Therefore, with the usage of the flag the message to the user is quite clear, even in the "current value" shown by ompi_info. Are there other parameters that use this behavior ? 

``` 
 MCA mtl psm2: parameter "mtl_psm2_mq_sendreqs_max" (current
                          value: "unset", data source: default, level: 3
                          user/all, type: unsigned_long)
                          PSM2 maximum number of isend requests in flight
                          (default: unset, let libpsm2 use its default)
```
Without this approach, you would need "at least" N different versions of this line ( opal/mca/base/mca_base_var.c 1901) one for each data type and finding sentinel values that are valid sentinel for *all parameters* that want this behavior. This doesn't seem to be good a approach. 

```  if ((var->mbv_flags & MCA_BASE_VAR_FLAG_DEF_UNSET) && ```


Using the flag seems to be a cleaner approach. My proposal would be to gradually migrate the the parameters needing this behavior to use the flag.

jsquyres:
@matcabral Sorry for the giant delay; this fell off my radar.  ☹️ 

It seems like there are 3 issues on this PR:

1. You need a new MCA param (`mtl_psm_opa_service_level`) for the v3.1.0 release.
   * If nothing else, maybe we should split that part into a separate, small PR in the *immediate* future that I think no one will disagree on.  I.e., let's make sure it gets into v3.1.0.
2. The "unset" strategy for MCA params.
   * I guess I have a few pieces of feedback here:
      * It's weird to introduce this in a point release.
      * It's weird that only MCA PSM will use this unset variable strategy.  I understand that this would be a new capability that other places in OMPI can use, but like I said -- it feels a bit weird to introduce this in a point release (and then have inconsistency between parts of OMPI using "unset" vs. default/sentinel values).
      * I'm not sure I understand your argument that you'd have to have N lines to compare for different sentinel values.  Couldn't the sentinel value (e.g., 0 or INT_MAX) be a member of `struct ompi_mtl_psm2_shadow_variable`?  That way, you can still keep your loop / have a generalized mechanism for checking/setting underlying env variables.
3. The "shadow parameter" concept.
    * It took me a while to understand this.  Now I understand that you're setting env variables for the underlying PSM library.  I.e., you're effectively exposing underlying env variables as MCA parameters.  That feels... weird (IMHO).  
    * We actually previously disallowed a similar-but-different scheme (of forwarding non-Open MPI MCA params to back-end processes) because the community decided that setting underlying library parameters wasn't a business that Open MPI should be in.
    * However, it is something that the runtime might be able to take care of.  Can we just have some kind of registration system that allows us to forward more env variable prefixes (i.e., in addition to `OMPI_*` and friends)?  E.g., register to forward `PSM_*` and `PSM2_*`?  I realize we'd have to figure out some things about how to get that registered via `mpirun`, since it doesn't check MPI frameworks, but that feels solvable.

All that being said, the most important question is: what does this PR do in terms of backwards compatibility to v3.0.x?  For example:

* What happens if I have a parameter file that sets one of the MTL PSM MCA params to one of its v3.0.0 default values?
* For example, if this config file sets `mtl_psm_mq_sendreqs_max` to 1048576 (which is the default value in 3.0.0), what happens in OMPI 3.0.x vs. OMPI 3.1.x?
   * When running with OMPI 3.1.x, you'll clearly get a value of 1048576.
   * But what happens with OMPI 3.0.x?  Do you *always* get a value of 1048576, or will you get whatever libpsm's default value is (regardless of whether it is 1048576 or not)?
* I.e., what happens to Open MPI's backwards compatibility guarantees for the PSM MTL between 3.0.x and 3.1.x?

Finally: A minor issue: you introduced `mtl_psm_opal_service_level` as a level 9 param.  Did you mean to do that?  See the guidelines MCA param levels here: https://github.com/open-mpi/ompi/wiki/MCAParamLevels (you may well have intended level 9 -- I don't know, but it may also have been a copy-n-paste, so I figured I'd ask).

rhc54:
Please see https://github.com/pmix/RFCs/pull/29 for an alternative method of forwarding envars.

matcabral:
Delay forgiven, lets now get this done  ;)

There is one item not mentioned below that for me has the highest priority:  Without this patch 3.1.0rc1 is defaulting some PSM2 setup parameters that will negatively impact performance. The most significant one is PSM2_SHAREDCONTEXTS_MAX to 8. This will be very impactful any time running more than 8 ranks per node. I admit I oversaw this when this was committed and I reviewed. On the similar line, the RNDV thresholds change according to the arch. As I mention, I don’t think it is worth to re implement PSM logic setting this values here (and later maintaining). So, my highest priority request would be to remove this parameters if we cannot make the “unset” behavior go in (either with the flag proposal or the sentinel value) 

Some answers to your questions:

>	1 You need a new MCA param (mtl_psm_opa_service_level) for the v3.1.0 release.

Yes.

>	I'm not sure I understand your argument that you'd have to have N lines to compare for different sentinel values

There are variables of different types (mca_base_var_storage_t) . You’ll need a different comparison for each type. 

>	 That way, you can still keep your loop

I’m also proposing a change for ompi_info to also show the actual “unset” state of the variable. 

>	I.e., you're effectively exposing underlying env variables as MCA parameters. That feels... weird (IMHO).

Agreed. There is a legacy thing on how psm2 receives config parameters. I’m pushing for API to pass config paramters. 

>	We actually previously disallowed a similar-but-different scheme (of forwarding non-Open MPI MCA params to back-end processes) because the community decided that setting underlying library parameters wasn't a business that Open MPI should be in.

@hjelmn wrote the original commit. Would you please elaborate on what your needs where. In my particular case, I just need the SL (customer specific details)

>	Can we just have some kind of registration system that allows us to forward more env variable prefixes (i.e., in addition to OMPI_* and friends)? 

This could be an alternative, but may actually be an overkill. I don’t expect this to be used in the most common cases. 

>	What happens if I have a parameter file that sets one of the MTL PSM MCA params to one of its v3.0.0 default values?

The “shadow parameters” are being added to 3.1.x (not in 3.0.x). So, older parameters file will not have any of these parameters. Besides, there is also no change in the behavior when setting a parameter (file, env, command line). The change is actually when the user does not set it. OMPI should intentionally let PSM2 decide. 

>	mtl_psm_opal_service_level as a level 9 param.

Thanks for the reference. I see it should be 6.   There is always something else to learn in this OMPI journey …

rhc54:
I believe what @jsquyres and I are trying to convey is that we find this whole "shadow param" approach unsettling and very fragile as it is a one-value-at-a-time solution. It would appear to be complicated from a vendor's perspective as they would now need to educate users on the "native" parameter, and then explain the OMPI-specific "shadow" version. It also involves messing around with the MCA system, which then impacts everyone. Bottom line: it seems like a roundabout way of addressing the root problem - which is how to forward envars that don't start with "OMPI".

What we are proposing is that we don't introduce these "shadow" params in 3.1, and instead use some simpler method as a temporary workaround. I actually already have such a thing implemented in the schizo framework (we forward PMIx envars as well as OMPI ones, so all we'd have to do is add 1-2 lines of code there). Not ideal from an architecture standpoint, but good enough for the immediate need.

We can then revisit this problem in OMPI master using the PMIx extension I mentioned above as a generalized approach to the root problem. There are others needing a similar capability, so this then becomes a more generally usable system.

matcabral:
I see like separated topics (with my corresponding questions below) in your comment:

 - "shadow variables" to be or not to be: I do understand your concerns and how this could be confusing in some cases. However, I think the idea behind this is flexibility. Maybe if the MCA parameters help string says that is setting variable PSM2_*.  My point here is to understand if there is value in the shadow variables independent of the fact that they are env vars. Are there other transports that pass parameters to a lower level library through API ? (that also have default values) 

- Exporting non OMPI variables to processes: I'm not actually convinced this is a must, at least for my case. Given that starting in 2.x (I have not tried on 3.x) you can add to the mca-params.conf tings like:  
``` -x PSM2_MQ_SENDREQS_MAX=1048576 ```

rhc54:
I'm afraid I'm not seeing the flexibility - it seems awfully fragile to me as you have to specify every single variable name, which means you have to somehow track all changes made to psm2 going forward. Getting modifications into OMPI releases means that you will always be out-of-sync.

I personally don't see the value in shadow variables. The only possible reasons for wrapping native envars in MCA params would be to exploit the precedence algorithm used by MCA, and the fact that we forward them. It doesn't sound like precedence is desirable here as the underlying library already has a default value (and the user is simply specifying their desired replacement), which means we are left with forwarding.

Exporting values using -x is certainly possible, but clunky as you again have to list every one of them. If it works for you in this case, then why bother with shadow variables anyway? Just document where/how to specify the values.

jsquyres:
First: I agree that we should split off and get into v3.1 the new MCA param(s) that you need for performance.  I guessed that it was `mtl_psm_opa_service_level`, but perhaps I wasn't entirely right.  Regardless, whichever one(s) it was, let's separate that from the rest of this discussion.

Second: there's been lots of detailed posts going back and forth -- it feels like it's time to just get on the phone and figure this out.  I'll follow up with the existing email thread for times, etc.

matcabral:
As per agreed on a phone call: 
 - I will close this PR without merging it. 
 - I will create a different PR to completely remove the shadow variables from 3.1.x (no PR needed for 3.0.x). 
 - Will leave master as is (with the shadow variables and the MCA_BASE_VAR_FLAG_DEF_UNSET flag) until we find a more robust solution.
 - @jsquyres, @rhc54 and me will start thinking on an approach that allows "registering" other prefixes than OMPI_* to be propagated.

PR #4575, opened 2017-12-06
osc/rdma: add an mca parameter to list MTLs for which osc pt2pt should have higher priority than rdma and default to psm2

matcabral:
osc/rdma: add an mca parameter to list MTLs for which osc pt2pt should have higher priority than rdma and default to psm2

Context: the Intel Omni-path driver (hfi1) has verbs support, so the openib
btl is available to use. However, at a bad performance. Without this
change osc rdma using btl openib is the default choice when running on Intel
Omni-path, with a lower performance than osc pt2pt over mtl psm2.

(cherry picked from commit 80c8858c5a5025c005d4dbe7a56ca52ad2322b16)

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2012-2015 Sandia National Laboratories.  All rights reserved.
  * Copyright (c) 2015      NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2015      Intel, Inc. All rights reserved.
+ * Copyright (c) 2015-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2016-2017 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -223,6 +226,13 @@ static int ompi_osc_rdma_component_register (void)
                                             MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0, OPAL_INFO_LVL_3,
                                             MCA_BASE_VAR_SCOPE_GROUP, &ompi_osc_rdma_btl_names);
 
+    ompi_osc_rdma_mtl_names = "psm2";
+    (void) mca_base_component_var_register (&mca_osc_rdma_component.super.osc_version, "mtls",
+                                            "Comma-delimited list of MTL component names to lower the priority of rdma "
+                                            "osc component favoring pt2pt osc (default: psm2)",
```
jsquyres: You should probably `asprintf()` the help string to show the the default value as whatever is in `ompi_osc_rdma_mtl_names` (because someone will inevitably modify that value someday and forget to update this help string).

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -709,6 +723,21 @@ static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, s
     return ret;
 }
 
+static int ompi_osc_rdma_query_mtls (void)
+{
+    char **mtls_to_use;
+
+    mtls_to_use = opal_argv_split (ompi_osc_rdma_mtl_names, ',');
```
jsquyres: I think you need to remember to `opal_argv_free(mtls_to_use)`.

matcabral:
gentle review request reminder ;)

jsquyres:
@matcabral It doesn't look like anything has changed since my review on Dec 6 2017...?

matcabral:
@jsquyres I addressed the comments in #4580

jsquyres:
Ok, commented over in #4580.

PR #4579, opened 2017-12-06
build: Clean up flag handling for examples

bwbarrett:
Fix ability to build examples from 32-bit builds.  Remove the implicit
rule usage, so that we know what flags are being used.  Make the override
of the FLAGS variables additive so that we don't wipe out FLAGS variables
set in the environment.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -22,15 +22,12 @@
 # $HEADER$
 #
 
-# Use the Open MPI-provided wrapper compilers.  Note that gmake
-# requires the CXX macro, while other versions of make (such as Sun's
-# make) require the CCC macro.
+# Use the Open MPI-provided wrapper compilers.
 
-CC = mpicc
-CXX = mpic++
-CCC = mpic++
-FC = mpifort
-JAVAC = mpijavac
+MPICC = mpicc
+MPICXX = mpic++
+MPIFC = mpifort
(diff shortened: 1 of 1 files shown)


examples/Makefile
```diff
@@ -129,49 +126,56 @@ clean:
 
 # Don't rely on default rules for the Fortran and Java examples
 
+hello_c: hello_c.c
+	$(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
+ring_c: ring_c.c
+	$(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
+connectivity_c: connectivity_c.c
+	$(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
```
jsquyres: I think you missed the C++ targets.

From CI:

```
g++ -g    hello_cxx.cc   -o hello_cxx
hello_cxx.cc:17:17: fatal error: mpi.h: No such file or directory
 #include "mpi.h"
                 ^
compilation terminated.
<builtin>: recipe for target 'hello_cxx' failed
```

bwbarrett: a good reminder of why we have CI :).  I forgot to enable the C++ bindings when I was testing locally...

PR #4580, opened 2017-12-06
osc/rmda: fix missing opal_argv_free in mtls search. Use asprintf in description message

matcabral:
osc/rmda: fix missing opal_argv_free in mtls search. Use asprintf in description message to avoid missing default values description message to avoid missing default values


matcabral:
@jsquyres  addressing the comment of #4575 in master.

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -170,70 +170,81 @@ static int ompi_osc_rdma_pvar_read (const struct mca_base_pvar_t *pvar, void *va
...
 {
+    char *description_str;
     mca_osc_rdma_component.no_locks = false;
+    asprintf(&description_str, "%s%s%s%s", "Enable optimizations available only if MPI_LOCK is ",
+                                  "not used. Info key of same name overrides this value (default: ",
+                                   mca_osc_rdma_component.no_locks  ? "true" : "false",")");
     (void) mca_base_component_var_register(&mca_osc_rdma_component.super.osc_version,
-                                           "no_locks", "Enable optimizations available only if MPI_LOCK is "
-                                           "not used. Info key of same name overrides this value (default: false)",
+                                           "no_locks", description_str,
```
jsquyres: Don't forget to `free(description_str)` after each time you create/use it.

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -170,70 +170,81 @@ static int ompi_osc_rdma_pvar_read (const struct mca_base_pvar_t *pvar, void *va
 
 static int ompi_osc_rdma_component_register (void)
 {
+    char *description_str;
     mca_osc_rdma_component.no_locks = false;
+    asprintf(&description_str, "%s%s%s%s", "Enable optimizations available only if MPI_LOCK is ",
```
jsquyres: Minor style thing -- you really only need one `%s`, even if your string constant spills over multiple lines.  Perhaps something like this:

```c
asprintf(&description_str, "Enable optimizations available only if MPI_LOCK is "
    "not used. Info key of same name overrides this value (default: %s)",
    (mca_osc_rdma_component.no_locks ? "true" : "false"));
```

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -726,16 +737,20 @@ static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, s
...
+            if (0 == strcmp (mtls_to_use[i], ompi_mtl_base_selected_component->mtl_version.mca_component_name)) {
+                mtl_match = true;
+                break;
+            }
+        }
     }
-    return -1;
+
+    opal_argv_free (mtls_to_use);
+    return mtl_match ? OMPI_SUCCESS : -1;
```
jsquyres: Should probably return `OMPI_ERR_NOT_FOUND` or somesuch; not just a straight `-1`.  That way the caller can still check for `OMPI_SUCCESS` or not -- they don't necessarily have to look for a specific return code, but they *could* still look for `<0` if they wanted to check for an error.

PR #4590, opened 2017-12-08
Ensure we don't send a kill signal to pid=0 as that hits ourselves and initiates an infinite loop.

rhc54:
Thanks to Michael Fenn for the report.

(cherry picked from commit ee2a93cb2ea59e9d658762118f8c317fdf7ddef0)


rhc54:
perhaps not, but it is the release the user reported against

jsquyres:
RM approved by the v2.0.x RM's.  This won't force another release, but at least it will end up in the v2.0.x snapshot tarball.

Issue #4578, opened 2017-12-06
Hanging with more than 64 hosts


ceandrade:
## Problem

I can not run jobs in parallel in a cluster with more than 64 machines.

-----------------------------

## System

Using openmpi-3.0.0.tar.bz2, Sep 12, 2017, compiled from source/distribution tarball.

Linux version 2.6.32-573.3.1.el6.x86_64 (mockbuild@c6b8.bsys.dev.centos.org) (gcc version 4.4.7 20120313 (Red Hat 4.4.7-16) (GCC) ) #1 SMP Thu Aug 13 22:55:16 UTC 2015

-----------------------------

## Details of the problem

I'm trying to use OpenMPI in a cluster with 200 machines. Unfortunately, `mpirun` looks like be frozen when I submit the jobs:
```shell
$ mpirun -np 1 -hostfile hosts.txt -pernode hostname
```
where `hosts.txt` has 200 machines. 

If the number of machines is at most 64, I can get the answer pretty fast:
```shell
$ mpirun -np 64 -hostfile hosts_64.txt -pernode hostname
machine1
machine2
....
machine64
```

This will be blocked forever:
```shell
$ mpirun -np 65 -hostfile hosts_65.txt -pernode hostname
<no response for 10min...>
```

rhc54:
If you add "-mca routed_radix 300" to your command line, does it make a difference?

ceandrade:
It got better but didn't scale to more than 128 nodes, even "-mca routed_radix 600". I also tried "routed_debruijn" and "routed_binomial", but no success.

ggouaillardet:
@ceandrade can you please confirm the node on which you run `mpirun` is not part of your `hosts_64.txt` hostfile ?

ceandrade:
@ggouaillardet I tried both actually. I launched from a machine outside the cluster, and from machines within the cluster. This is the line:

$ mpirun -v -display-devel-map -display-allocation -mca routed_radix 600 --bind-to none -np 1 -hostfile hosts_266.txt -pernode hostname

but no answer.  Note that I have 266 machines in my cluster.

Is there a way to create some more detailed logging info that I can share with you guys?

ggouaillardet:
I can reproduce the issue with 65 tasks and with the default radix, but only when `mpirun` is invoked from a node not in the hostfile. I will resume investigations tomorrow. Stay tuned !

rhc54:
[...]

Issue #4465, opened 2017-11-07
RSH launcher hangs on more than 65 nodes


jladd-mlnx:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
v3.0.x
master

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
git clone 

### Please describe the system on which you are running

* Operating system/version: 
RedHat 7.3
* Computer hardware: 
Intel dual socket Broadwell

* Network type: 
IB
-----------------------------

## Details of the problem

Open MPI 3.0.x and master, the following command hangs when the number of nodes in the hostfile is 66 or more (NOTE: this was originally reported to us by a customer; on their system, the hangs happens on at 65 nodes or more):

```
$mpirun --npernode 1 --hostfile mfile -mca plm rsh --debug-daemons -x LD_LIBRARY_PATH hostname
```
Debug output looks as follows:

```
Daemon [[50860,0],46] checking in as pid 24964 on host clx-hercules-096
[clx-hercules-096:24964] [[50860,0],46] orted: up and running - waiting for commands!
[clx-hercules-096:24964] [[50860,0],46] orted_cmd: received tree_spawn
Daemon [[50860,0],6] checking in as pid 4077 on host clx-hercules-008
[clx-hercules-008:04077] [[50860,0],6] orted: up and running - waiting for commands!
[clx-hercules-008:04077] [[50860,0],6] orted_cmd: received tree_spawn
Daemon [[50860,0],41] checking in as pid 30550 on host clx-hercules-060
[clx-hercules-060:30550] [[50860,0],41] orted: up and running - waiting for commands!
[clx-hercules-060:30550] [[50860,0],41] orted_cmd: received tree_spawn
Daemon [[50860,0],22] checking in as pid 11025 on host clx-hercules-025
[clx-hercules-025:11025] [[50860,0],22] orted: up and running - waiting for commands!
[clx-hercules-025:11025] [[50860,0],22] orted_cmd: received tree_spawn
Daemon [[50860,0],30] checking in as pid 8031 on host clx-hercules-034
[clx-hercules-034:08031] [[50860,0],30] orted: up and running - waiting for commands!
[clx-hercules-034:08031] [[50860,0],30] orted_cmd: received tree_spawn
```

We found that disabling the tree based RSH launch works around the issue, i.e.

```
[...]

PR #4583, opened 2017-12-07
v3.0: mpool/base: plug a memory leak

ggouaillardet:
set the key of all mpool_tree_item objects, so they can be retrieved
in mpool_base_free and then returned back to the
mca_mpool_base_tree_item_free_list free list.

Refs. open-mpi/ompi#4567

Thanks Philip Blakely for the bug report.

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>

(cherry picked from commit open-mpi/ompi@11e5f86bf89d17f50e00ee8c07d9a336d99b67d2)

PR #4591, opened 2017-12-08
ess/base: be sure to update the routing tree

ggouaillardet:
so any tree spawn operation properly gets the number of children underneath us.

This commit is a tiny subset of open-mpi/ompi@347ca412b0fd3063d67f254d9470f66fb7e6d955
that should have been back-ported into the v3.0.x branch

Fixes open-mpi/ompi#4578

Thanks to Carlos Eduardo de Andrade for reporting.

### DIFF
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -533,6 +533,9 @@ int orte_ess_base_orted_setup(void)
             error = "construct nidmap";
             goto error;
         }
+        /* be sure to update the routing tree so any tree spawn operation
+         * properly gets the number of children underneath us */
+        orte_routed.update_routing_plan(NULL);
     }
 
     if (orte_static_ports) {


ggouaillardet:
@karasevb @jladd-mlnx can you give this a try and check whether this fixes open-mpi/ompi#4465 too ?

ggouaillardet:
@hppritcha i will update `NEWS` once #4583 gets merged (otherwise there will be a conflict to be resolved in a very near future ...)

rhc54:
Thanks @ggouaillardet !

karasevb:
I'm going to check this fix for #4465

karasevb:
The #4465 issue still remains. I have it reproduced with latest OMPI@master and v3.0.x

artpol84:
@karasevb can you also test v3.1 as well?

artpol84:
@karasevb 
Also please try with `oob/tcp`. By default if IB fabric is present `oob/ud` supposed to be used so we want to try with `oob/tcp` to see if it's an OOB issue.

karasevb:
@artpol84 v3.1.x has the same hang.
I tried launch with `-mca oob tcp` but it not affected, the hang is still observed.

The hanging does not reproduce when try to add option ` --mca routed_radix 100`. I tried to run on 100 nodes:
```
mpirun -npernode 1 --hostfile mfile --mca routed_radix 100 -mca plm rsh -x LD_LIBRARY_PATH hostname
```
That fixes the hang with master/v3.0.x/v3.1x. And also v3.0.x without this PR works well.

rhc54:
Only thing I can suggest is adding ```-mca plm_base_verbose 5 -mca routed_base_verbose 5``` and see if anything pops out. It bother me that none of us can seem to reproduce this behavior.

@ggouaillardet Do you have any luck reproducing it? I can't.

ggouaillardet:
i will give it a try.
@karasevb @artpol84 is the node running `mpirun` part of the `mfile` ?
can you always force `--mca oob tcp` in order to make sure the issue is independent of the `oob` component.
last but not least, can you try to reproduce the issue with a minimal number of nodes ? for example by forcing a smaller radix `--mca routed_radix 2`

karasevb:
@ggouaillardet the node running `mpirun` is not included to the `mfile`
I found out the `routed_radix` value should be greater or equal the number of nodes for the successful execution (`mfile_2` has two nodes):
```
$mpirun -npernode 1 -hostfile mfile_2 -mca plm rsh -mca oob tcp  --mca routed_radix 2 hostname
node-002
node-003

$mpirun -npernode 1 -hostfile mfile_2 -mca plm rsh -mca oob tcp  --mca routed_radix 1 hostname
<hanging>
```
I  have checked the default `routed_radix ` value:
```
$./ompi_info -a
		...
        MCA routed radix: parameter "routed_radix" (current value: "64", data source: default, level: 9 dev/all, type: int)
                          Radix to be used for routed radix tree
		...
```
That is my case why I get a hang when try to run more than 64 nodes.
@ggouaillardet  Is there expected behavior?

ggouaillardet:
@karasevb what is the exact version (e.g. commit id) you are running ? i used to reproduce this only in `v3.0.x` __before__ this PR was merged.

can you please post the output of a hang with  `-mca plm_base_verbose 10` ?

karasevb:
Here is the verbose output during the hang:
https://github.com/karasevb/test/blob/master/mpirun_rsh.log

karasevb:
@ggouaillardet I use this commit 037795911919a4f0dcb32707858ea1b197d79569

ggouaillardet:
@karasevb i observed the same symptom (No children + hang) __before__ this PR was merged.
i rebuilt from the same commit, and am unable to reproduce the issue.

can you please double check you built from the unmodified commit you mentionned and post your `configure` command line ?

karasevb:
@ggouaillardet I had rechecked v3.0.x with the latest commit  a69a84e867431770871fb40f38539705c5e1a15f
Here is updated output with `-mca plm_base_verbose 10`: https://github.com/karasevb/test/blob/master/mpirun_rsh_2.log

Config:
```
Configure command line: '--prefix=<path to work dir>/install/ompi_v3.0.x' '--enable-debug' '--without-ucx' '--enable-orterun-prefix-by-default'
```

ggouaillardet:
@karasevb thanks, can you also please post your `mfile_2` ?

karasevb:
@ggouaillardet `mfile_2` is pretty simple:
```
node-001
node-002
```

karasevb:
@ggouaillardet 
I running `mpirun` into the Slurm allocation and I have an interesting fact: when I use the Slurm allocation of 100 nodes, then a hang is reproduced. But when I tried reduce the Slurm allocation till 4 nodes I see a hang only before this PR was merged (as you pointed above). 
Perhaps this is an our system issue. 
Could you try to reproduce my experience?

ggouaillardet:
This is making less and less sense ...
`front-end` ssh to `node-002` but according to the hostfile, it should really ssh to `node-001`
Are you running this from `salloc` ?

When running on 100 nodes under slurm, are you using `sbatch` (and hence `mpirun` runs on `node-xxx`) or `salloc` (and hence `mpirun` is running on `front-end`)  ?

I am not sure I can spawn 100 vm to test that ... are you able to reproduce the hang with less nodes ?

karasevb:
I use `salloc`, and` mpirun` is running on `front-end`.
I reproduced the hang on two different clusters. The threshold for the number of nodes for reproduce was different for each cluster. One of them is 59 nodes, and second is 72.

rhc54:
@karasevb Your log clearly shows the problems (there are two), but I'm not sure I understand the cause.

First, as @ggouaillardet pointed out, when we read the hostfile we are not finding node-001 in your allocation. Remember, when you are in a _managed_ system, the hostfile acts solely as a filter on the nodes allocated by SLURM. So if node-001 isn't included in your allocation, then we will ignore it. I suspect that is what is happening here.

Second, it looks like the OMPI libraries on node-002 aren't correct. With the updated v3.0.1 code, there should have been a call to compute the routing tree when the daemon starts up. You can see the output from that call on your frontend when mpirun starts up:
```
[frontend:03834] [[57145,0],0]: parent -1 num_children 1
[frontend:03834] [[57145,0],0]:    child 1
[frontend:03834] [[57145,0],0]:            relation 2
[frontend:03834] [[57145,0],0] routed:binomial rank 0 parent 0 me 0 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial 0 found child 1
[frontend:03834] [[57145,0],0] routed:binomial rank 0 parent 0 me 1 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial find children of rank 0
[frontend:03834] [[57145,0],0] routed:binomial find children checking peer 1
[frontend:03834] [[57145,0],0] routed:binomial find children computing tree
[frontend:03834] [[57145,0],0] routed:binomial rank 1 parent 0 me 1 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial find children returning found value 0
[frontend:03834] [[57145,0],0] routed:binomial 0 found child 2
[frontend:03834] [[57145,0],0] routed:binomial rank 0 parent 0 me 2 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial find children of rank 0
[frontend:03834] [[57145,0],0] routed:binomial find children checking peer 1
[frontend:03834] [[57145,0],0] routed:binomial find children computing tree
[frontend:03834] [[57145,0],0] routed:binomial rank 1 parent 0 me 2 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial find children of rank 1
[frontend:03834] [[57145,0],0] routed:binomial find children checking peer 3
[frontend:03834] [[57145,0],0] routed:binomial find children checking peer 2
[frontend:03834] [[57145,0],0] routed:binomial find children computing tree
[frontend:03834] [[57145,0],0] routed:binomial rank 2 parent 0 me 2 num_procs 3
[frontend:03834] [[57145,0],0] routed:binomial find children returning found value 0
[frontend:03834] [[57145,0],0]: parent 0 num_children 2
[frontend:03834] [[57145,0],0]:    child 1
[frontend:03834] [[57145,0],0]:    child 2
[frontend:03834] [[57145,0],0] routed:direct: update routing plan
```
This output should have been seen on node-002, but it isn't - which makes me believe that the OMPI libraries on that node were not updated.

Another problem I see is:
```
[frontend:03834] [[57145,0],0] plm:base:setup_vm assigning new daemon [[57145,0],1] to node node-002
[frontend:03834] [[57145,0],0] plm:base:setup_vm add new daemon [[57145,0],2]
[frontend:03834] [[57145,0],0] plm:base:setup_vm assigning new daemon [[57145,0],2] to node node-003
```
Note that we assigned a daemon to node-003, which isn't in the mfile you provided. So now I am totally confused - where is node-003 coming from? And why is the newly launched orted not updating its routes per the path @ggouaillardet committed?

rhc54:
I believe I see the problems. First, I believe you gave us the incorrect mfile - it actually includes node-002 and node-003, which is why we first launch to node-002.

Second, your opal_prefix is incorrectly set. Look at the ssh line we are trying to execute:
```
[frontend:03834] [[57145,0],0] plm:rsh: final template argv:

        /bin/ssh <template>     PATH=<path to workdir>/install/ompi_v3.0.x/bin:$PATH ; export PATH ; LD_LIBRARY_PATH=<path to workdir>/install/ompi_v3.0.x/lib:$LD_LIBRARY_PATH ; export LD_LIBRARY_PATH ; DYLD_LIBRARY_PATH=<path to workdir>/install/ompi_v3.0.x/lib:$DYLD_LIBRARY_PATH ; export DYLD_LIBRARY_PATH ;   <path to workdir>/install/ompi_v3.0.x/bin/orted -mca ess "env" -mca ess_base_jobid "3745054720" -mca ess_base_vpid "<template>" -mca ess_base_num_procs "3" -mca orte_hnp_uri "3745054720.0;tcp://10.131.5.129,10.131.200.231,172.17.10.240:37506" -mca oob "tcp" --mca routed_radix "1" -mca plm_base_verbose "10" -mca routed_base_verbose "10" -mca plm "rsh" -mca rmaps_ppr_n_pernode "1" -mca pmix "^s1,s2,cray,isolated"
```
See the ```<path to workdir>``` argument? That should have been the actual path, not some odd string. You can see in the actual ssh command that nothing is substituted for it:
```
[frontend:03834] [[57145,0],0] plm:rsh: executing: (/bin/ssh) [/bin/ssh node-002     PATH=<path to workdir>/install/ompi_v3.0.x/bin:$PATH ; export PATH ; LD_LIBRARY_PATH=<path to workdir>/install/ompi_v3.0.x/lib:$LD_LIBRARY_PATH ; export LD_LIBRARY_PATH ; DYLD_LIBRARY_PATH=<path to workdir>/install/ompi_v3.0.x/lib:$DYLD_LIBRARY_PATH ; export DYLD_LIBRARY_PATH ;   <path to workdir>/install/ompi_v3.0.x/bin/orted -mca ess "env" -mca ess_base_jobid "3745054720" -mca ess_base_vpid 1 -mca ess_base_num_procs "3" -mca orte_hnp_uri "3745054720.0;tcp://10.131.5.129,10.131.200.231,172.17.10.240:37506" -mca oob "tcp" --mca routed_radix "1" -mca plm_base_verbose "10" -mca routed_base_verbose "10" -mca plm "rsh" -mca rmaps_ppr_n_pernode "1" -mca pmix "^s1,s2,cray,isolated"]
```
So the path to your installation is botched and you are picking up whatever default OMPI libraries exist on the backend.

Someone might look at how that weird string got in there. Meantime, try adding ```--enable-orterun-prefix-by-default``` to your configure line, drop the -x from your mpirun cmd line, and I expect things will run just fine.

karasevb:
@rhc54 Sorry for the confusion, I attached not appropriate hosts file. In the host file actually contains: 
```
node-002
node-003
```

artpol84:
@rhc54
Regarding path to workdir - this is anonymization of the log, there is a real path there

rhc54:
@karasevb No problem - just let us know when you edit the output (e.g., to remove the actual path) so we don't waste time wondering what it means.

I can only assume that the library path is incorrect, or that the backend libraries have not been updated. Either way, the output from the backend nodes doesn't match the patch @ggouaillardet committed.

artpol84:
Me and @karasevb did some more debugging now. I guess we hit the case that disables @ggouaillardet patch. I think that this codepath is not tested.

It starts here:
https://github.com/open-mpi/ompi/blob/v3.0.x/orte/mca/plm/base/plm_base_launch_support.c#L1559
In our case where we have large number of nodes our node list is not appended to the ssh cmdline (not `-mca orte_node_regex <..>` argument. It can be seen in the log that @karasevb posted above. 
However if the node number is small this argument is appended and the list of nodes is available.

Not having node regex doesn't trigger @ggouaillardet addition here:
https://github.com/open-mpi/ompi/pull/4591/files#diff-90f3ebd36767e677e669d28e2977b492L530

For debugging purposes it is enough to change `ORTE_MAX_REGEX_CMD_LENGTH` in https://github.com/open-mpi/ompi/blob/v3.0.x/orte/mca/plm/base/plm_base_launch_support.c#L1559 and you will be able to trigger the same code path.

rhc54:
Odd - I didn't think we had that switch any more since we can express it as a regex. I can take a look since @ggouaillardet is probably offline by now.

rhc54:
Oh...the light begins to dawn. You have a ```-``` in your node name! It is screwing up the regex processor as we think that stands for a range.

No wonder nobody can reproduce your problem!

artpol84:
Yes we do.

artpol84:
Slurm regex works fine with it btw.

rhc54:
That's fine - we don't. Will try to take a look at how to get around it, but not exactly at the top of the priority list. First time we've seen someone do that, so it is far from normal practice. Still, not saying it shouldn't be allowed.

artpol84:
@rhc54 thank you

PR #4592, opened 2017-12-10
ompi: pml ucx: allow setting of the number of endpoints from command …

alex-mikheev:
…line

Applications that have a sparse connectivity will be able to optimize
UCX resource usage.
@yosefe @jladd-mlnx


yosefe:
@alex-mikheev i thought we wanted it to be a parameter for UCX (env var), not in OMPI...

alex-mikheev:
so that ucx env should override ucx_init settings ?

yosefe:
yes

PR #4597, opened 2017-12-11
request/greq: fix fortran generalized request

AboorvaDevarajan:
Copy the return error from the Fortran query function into the C status


bwbarrett:
ok to test

bosilca:
I took that comment as applying only to the C part. The problem with the Fortran query, is that Fortran does not know the real ompi_status_t structure, and therefore cannot set the MPI_ERROR field.

jsquyres:
That makes a pretty weird disparity between the handling of C and Fortran.

At a minimum, this behavior should be documented on the man page (nothing fancy, just at least add a sentence or three about what we're doing).  Better would be to preserve the same behavior between C and Fortran.  IMHO.  😄

AboorvaDevarajan:
I'm closing this since I accidentally deleted the fork. I will open a new PR.

PR #4599, opened 2017-12-11
v3.0.x: btl/vader: change the way fast boxes are used

hjelmn:
There were multiple paths that could lead to a fast box
allocation. One of them made little sense (in-place send) so it has
been removed to allow a rework of the fast-box send function. This
should fix a number of issues with hanging/crashing when using the
vader btl.

References #4260

(cherry picked from commit a82f761a4a6670ff83d3c47601eeb26120d83be5)


bwbarrett:
bot:ompi:retest

Looks like one of the OS X builders was having issues; I've removed it from the testing fleet until we sort that out.

PR #4603, opened 2017-12-11
v2.x: build: Clean up flag handling for examples

bwbarrett:
Fix ability to build examples from 32-bit builds.  Remove the implicit
rule usage, so that we know what flags are being used.  Make the override
of the FLAGS variables additive so that we don't wipe out FLAGS variables
set in the environment.

(cherry picked from commit 97294dca4a0993c79344293ec90d8949c5ba372e)

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -20,15 +20,12 @@
 # $HEADER$
 #
 
-# Use the Open MPI-provided wrapper compilers.  Note that gmake
-# requires the CXX macro, while other versions of make (such as Sun's
-# make) require the CCC macro.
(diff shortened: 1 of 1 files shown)


jsquyres:
OS X CI failed because autoconf was not found.  Hopefully that was a transient error.

bot:ompi:retest

jsquyres:
@hppritcha Good to go.

bwbarrett:
bot:ompi:retest

Looks like the ARM-hosted ThunderX board got this test and was having issues with UCX.  It's out of service now, so let's try again (sigh).

PR #4609, opened 2017-12-12
build: Update NEWS in preperation for 3.1.0

bwbarrett:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -53,9 +53,31 @@ included in the vX.Y.Z section and be denoted as:
    (** also appeared: A.B.C)  -- indicating that this item was previously
                                  included in release version vA.B.C.
 
-Master (not on release branches yet)
-------------------------------------
+3.1.0 -- January, 2018
+----------------------
 
+- Add monitoring component for PML, OSC, and COLL to track data
+  movement of MPI applications.  See
+  ompi/mca/commmon/monitoring/HowTo_pml_monitoring.tex for more
+  information about the monitoring framework.
+- Add support for communicator assertions: mpi_assert_no_any_tag,
+  mpi_assert_no_any_source, mpi_assert_exact_length, and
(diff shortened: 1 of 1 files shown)


bwbarrett:
This isn't perfect; we need to sync master against the release branches and then pull that back into the release branches (so this reflects 3.0.0 and possibly 3.0.1), but this should be the list that's new for 3.1.0.

NEWS
```diff
@@ -53,9 +53,29 @@ included in the vX.Y.Z section and be denoted as:
    (** also appeared: A.B.C)  -- indicating that this item was previously
                                  included in release version vA.B.C.
 
-Master (not on release branches yet)
-------------------------------------
-
+3.1.0 -- January, 2017
```
jsquyres: Man, I read this too quickly this morning: shouldn't that be 2018?

NEWS
```diff
@@ -53,9 +53,29 @@ included in the vX.Y.Z section and be denoted as:
...
+----------------------
+
+- Add monitoring component for PML, OSC, and IO to track data movement
+  of MPI applications.
+- Add support for communicator assertions: mpi_assert_no_any_tag,
+  mpi_assert_no_any_source, mpi_assert_exact_length, and
+  mpi_assert_allow_overtaking.
+- Update PMIx to version 2.1.0.
+- Update hwloc to 1.11.6.
```
jsquyres: Per Peter K's comments, this should be 1.11.7.

NEWS
bwbarrett: I figured inventing time travel was easier than getting an OMPI release out.  Didn't I tell you that's why I've been so behind in pull requests? :)

Yeah, fixed.

bwbarrett:
@jsquyres can you re-review?  Thanks!

Issue #4509, opened 2017-11-16
sporadic fatal error messages due to critical bug in `madvise()` hook with OpenIB


bingmann:
I believe there is a critical bug in the (new?) `madvise()` hook:

In a program that does lots of Isend()/Irecv() with Wait()/Test(), I
sporadically see something like the following fatal error message:

```
--------------------------------------------------------------------------
Open MPI intercepted a call to free memory that is still being used by
an ongoing MPI communication.  This usually reflects an error in the
MPI application; it may signify memory corruption.  Open MPI will now
abort your job.

  rcache name:    grdma
  Local host:     fh1n076
  Buffer address: 0x2af75c4fc000
  Buffer size:    2568192
--------------------------------------------------------------------------
```

The error message is bogus and suspicious, especially since the program never allocates a buffer of that size. Nor does it use that address as a pointer, but the address _is_ contained in some memory area used.

OpenMPI Version: 3.0.0, installed from the source tarball with debug output. Probably affects all versions with the `madvise()` commits backported.

Running on a Linux HPC cluster, Kernel 3.10.0-693.2.2.el7.x86_64, with an InfiniBand 4X FDR Interconnect, glibc 2.17, gcc 5.2.0.

The error **only occurs with openib BTL**, with TCP it apparently never occurs, because the grdma rcache module is not used.

I believe the bug affects all programs using asynchronous communication, openib, and varying buffer sizes. It occurs naturally after running the program for some time. I have added a test program triggering the error artificially.

### Backtrace and Autopsy

Lots of debugging leads me to believe there is a bug in the way the interception of `madvise()` clears memory from the rcache grdma, which frees RDMA memory regions.

The fatal error message occurs when `_intercept_madvise()` is called,
which calls `opal_mem_hooks_release_hook()`,
which calls `mca_rcache_base_mem_cb()`,
which contain the fatal error message.

`mca_rcache_base_mem_cb()` is supposed to free rcache allocations and prints the message when `mca_rcache_grdma_invalidate_range()` fails.
The deallocation of areas happens by iterating over the memory area tree using `mca_rcache_base_vma_iterate()`, and calling `gc_add()` for areas to invalidate.

`gc_add()` fails if the invalidated area still has reference counts.

The issues is that `madvise()` is called in my program by the libc's malloc implementation with `MADV_DONTNEED` to free up regions no longer needed.  This occurs at unpredictable times, probably when malloc decides to consolidate free space in the heap.

[...]

PR #4611, opened 2017-12-12
v2.1.x: rcache/grdma: try to prevent erroneous free error messages

hjelmn:
It is possible to have parts of an in-use registered region be passed
to munmap or madvise. This does not necessarily mean the user has made
an error but does mean the entire region should be invalidated. This
commit checks that the munmap or madvise base matches the beginning of
the cached region. If it does and the region is in-use then we print
an error. There will certainly be false-negatives where a user
unmaps something that really is in-use but that is preferrable to a
false-positive.

References #4509

(cherry picked from commit d3fa1bbbb0ecd3428e52300752ea6869a1ffac46)


jsquyres:
@hppritcha good to go

Issue #4621, opened 2017-12-13
ORTE regex fails when nodenames include hyphens


rhc54:
If a cluster is configured with ```-``` in the names of their nodes, then the ORTE regex parser gets confused and thinks the hyphen is part of a range designation. This leads to unusually large regex results and can cause odd behavior (e.g., unable to tree spawn).

Fortunately, we haven't seen this in 14 years, but Mellanox now has a cluster that hits the problem. No rules against such configuration, so probably something that should be addressed in case someone else does it.

rhc54:
Just to further clarify: this problem will occur regardless of the plm launcher being used (e.g., when using mpirun under slurm) as we always try to pass the regex to the backend daemons on launch, and there is no guarantee that the length restriction will save us like it did in the referenced case.

ggouaillardet:
@rhc54 i am a bit puzzled with this ...

i renamed my virtual cluster `n-0`, `n-1` and so on, but was unable to reproduce the hang (this is not a large cluster, so the regex length is small enough.

there is a more obvious error when the regex is too long, and this does not involve `-` in the regex.

that can be simply reproduced with the inline patch below, and i can reproduce the issue when running
```
mpirun -npernode 1 -hostfile m -mca plm rsh -mca oob tcp -mca routed_radix 1 --mca plm_base_verbose 10 hostname
```
from `n0` and my hostfile `m` contains
```
n1
n2
```

at first glance, i did not see `n1` perform the phone home operation, but if you can point me to there, i will be happy to debug it.

```patch
diff --git a/orte/mca/plm/base/plm_base_launch_support.c b/orte/mca/plm/base/plm_base_launch_support.c
index 118813e..54930db 100644
--- a/orte/mca/plm/base/plm_base_launch_support.c
+++ b/orte/mca/plm/base/plm_base_launch_support.c
@@ -1573,6 +1573,7 @@ int orte_plm_base_orted_append_basic_args(int *argc, char ***argv,
         ORTE_ERROR_LOG(rc);
         return rc;
     }
+#if 0
     /* if this is too long, then we'll have to do it with
      * a phone home operation instead */
     if (strlen(param) < ORTE_MAX_REGEX_CMD_LENGTH) {
@@ -1582,6 +1583,7 @@ int orte_plm_base_orted_append_basic_args(int *argc, char ***argv,
         /* mark that the nidmap has been communicated */
         orte_nidmap_communicated = true;
     }
+#endif
     free(param);
 
[...]

PR #4627, opened 2017-12-15
orte/nidmap: correctly handle '-' as a valid hostname character

ggouaillardet:
'-' is not an alpha character nor a digit, but it is a valid hostname
character and should be handled as an alpha character, otherwise, nodes
such as node-001 do not get "compressed" in the regex.

Refs open-mpi/ompi#4621

### DIFF
--- a/orte/util/nidmap.c
+++ b/orte/util/nidmap.c
@@ -270,22 +270,15 @@ int orte_util_nidmap_create(opal_pointer_array_t *pool, char **regex)
             }
         }
         node = nptr->name;
-        /* determine this node's prefix by looking for first non-alpha char */
+        /* determine this node's prefix by looking for first digit char */
         fullname = false;
         len = strlen(node);
         startnum = -1;
         memset(prefix, 0, ORTE_MAX_NODE_PREFIX);
         numdigits = 0;
         for (i=0, j=0; i < len; i++) {
-            if (!isalpha(node[i])) {
-                /* found a non-alpha char */
-                if (!isdigit(node[i])) {
-                    /* if it is anything but a digit, we just use
-                     * the entire name
-                     */
-                    fullname = true;
-                    break;
-                }
+            /* valid hostname characters are ascii letters, digits and the '-' character. */
+            if (isdigit(node[i])) {
                 /* count the size of the numeric field - but don't
                  * add the digits to the prefix
                  */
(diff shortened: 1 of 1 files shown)


orte/util/nidmap.c
```diff
@@ -296,6 +289,13 @@ int orte_util_nidmap_create(opal_pointer_array_t *pool, char **regex)
                 }
                 continue;
             }
+            if ('.' == node[i]) {
+                /* just use the entire name */
+                fullname = true;
+                break;
+            }
+            /* this is either an alpha or '-' */
+            assert(isalpha(node[i]) || '-' == node[i]);
```
rhc54: Something doesn't look right here - why would you want to assert if the character is illegal? I should think we wouldn't want the daemon to simply crash. Let's instead do a show_help reporting the unsupported name and  return an error code.

orte/util/nidmap.c
```diff
@@ -296,6 +289,13 @@ int orte_util_nidmap_create(opal_pointer_array_t *pool, char **regex)
                 }
                 continue;
             }
+            if ('.' == node[i]) {
```
rhc54: Not sure I understand why a dot character mandates use of the entire name - could you add a comment to explain that option?

ggouaillardet: well, i did not fully understand the initial logic, so i just preserved it.
the initial test was 
```c
if (!isalpha(node[i]) && !isdigit(node[i])) {
    fullname=true;
    break;
}
```
because `node` is a valid hostname, `node[i]` is either
- an `alpha`
- a `digit`
- the `-` character
- the `.` character

i honestly do not see any reason why `.` should not be treated as a digit, so that means `fullname` is always `false` and hence i can remove some dead code. do you remember the rationale for the `fullname` test ?

at that stage, `node` should always be a valid hostname, so i did not feel the need to validate it (and hence the `assert()`). that being said, this is not a strong opinion, and i am fine 
 - checking `node` is a valid hostname and error otherwise
- remove the dead code

rhc54: IIRC, the rationale was that I didn't want to assume validity. The user can input a hostname in either a hostfile or a -host option, and we never check it for validity. Thus, the characters could be anything.

I suppose a better option would be to add a validity check in the hostfile and -host parsers, and then have a show_help and error out if a name fails. We could cite https://en.wikipedia.org/wiki/Hostname to explain the error.

Then I'd be comfortable just taking out the assert and assuming validity here, and the need for fullname goes away. Make sense?

Issue #4629, opened 2017-12-16
Incorrect communication to client proc


rhc54:
Strictly in the v2.x series. The daemon immediately attempts to send a message to its lowest-ranked child. The child process is required to initiate communications and will connect back to the daemon when ready. The unexpected message results in a SIGPIPE as the child hasn't yet setup a connection and causes the pipe to fail when two simultaneous connections are in progress.

If we are _not_ running under a debugger, then the daemon traps the SIGPIPE and ignores it. However, when under a debugger, the debugger traps the SIGPIPE and reacts to it, causing the entire procedure to potentially collapse. Unfortunately, this specifically impacts a particular debugger vendor.

rhc54:
@dirk-schubert-arm (Dirk Schubert, DDT developer) I have shifted this problem from #3660

dirk-schubert-arm:
Sorry for the late reply, I was on holiday. @xavier1arm has applied your fix in #4630 and can confirm it works.  Thanks @rhc54. Is there anything I can help with to move #4630 forward. I feel not familiar enough with the code base to do the review, but on the other hand it's just a small and local change.

rhc54:
Thanks for the update. I think we are stuck until @jsquyres returns from vacation, but the release isn't going anywhere until then anyway.

hppritcha:
merged fix #4630 into v2.x.  closing this issue.

jsquyres:
@rhc54 Is this a request that I test this (on v2.1.x) under DDT on my cluster?

If so, I can confirm that with the head of each of the following branches as of 3 Jan 2018:

1. 2.0.x: DDT attachment works
1. 2.1.x: DDT attachment works
1. 3.0.x: DDT attachment works
1. 3.1.x: DDT attachment works

jsquyres:
I neglected to mention -- I tested with DDT Allinea Forge v7.0.2

rhc54:
Sorry - wasn't necessary as Allinea as already confirmed it works 😢

PR #4630, opened 2017-12-16
Ensure that daemons NEVER initiate connections to childs

rhc54:
A debugger can generate a release message prior to the proc having fully initialized and established a connection to the daemon. This causes the daemon to initiate the connection, which ultimately collides with the proc's connection and causes very bad things to happen.


rhc54:
Fixes https://github.com/open-mpi/ompi/issues/4629

PR #4637, opened 2017-12-19
orted: fix tree-spawn when the node regex is too long

ggouaillardet:
When the node regex is too long to be sent on the command line,
retrieve  it first from the parent, and then spawn the remote orted

### DIFF
--- a/orte/mca/odls/odls_types.h
+++ b/orte/mca/odls/odls_types.h
@@ -13,6 +13,8 @@
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2018      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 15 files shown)


ggouaillardet:
note to myself:
 - do we also need to retrieve the prefix when receiving the regex ?
 - it seems `remote_spawn()` is always invoked with a `NULL` parameter, is there some dead code to be removed ?

rhc54:
Looks okay to me - thanks!

You shouldn't need to retrieve the prefix because you were already given it on the cmd line that started the orted.

There is indeed some dead code to remove. Now that the daemon calls remote_spawn itself, there is no longer a need for the "tree_spawn" cmd (odls/odls_types.h) or the associated cmd processing code in orted/orted_comm.c as the HNP is no longer sending a tree-spawn message to the orted.

jsquyres:
@ggouaillardet @rhc54 Is this one ready to merge?

orte/mca/odls/odls_types.h
```diff
@@ -44,7 +46,7 @@ typedef uint8_t orte_daemon_cmd_flag_t;
 #define ORTE_DAEMON_KILL_LOCAL_PROCS        (orte_daemon_cmd_flag_t) 2
 #define ORTE_DAEMON_SIGNAL_LOCAL_PROCS      (orte_daemon_cmd_flag_t) 3
 #define ORTE_DAEMON_ADD_LOCAL_PROCS         (orte_daemon_cmd_flag_t) 4
-#define ORTE_DAEMON_TREE_SPAWN              (orte_daemon_cmd_flag_t) 5
+#define ORTE_DAEMON_TREE_SPAWN_UNUSED       (orte_daemon_cmd_flag_t) 5
```
rhc54: I think you can just remove this cmd - no need to renumber the rest.

rhc54:
I think it is, yes - but I defer to @ggouaillardet

ggouaillardet:
ready for primetime, merging now

PR #4639, opened 2017-12-19
Disable the LSF plm if CSM is detected

jjhursey:
LSF running on top of CSM does not provide LSF daemons on the compute nodes.
Release branch version of PR #4439


gpaulsen:
I'd argue that this is a bug, not an enhancement if CSM (an IBM process starter) is present.

Issue #4563, opened 2017-12-04
opal_fifo check hanging on aarch64


hppritcha:
At least on master at  b160cf63 opal_fifo appears to be regularly hanging on aarch64.  I'm using gcc 4.8.5.  This test had been passing regularly with jenkins CI PR testing until sometime in the last several days/week.

The test does not appear to hang when Open MPI is configured with --enable-debug.

kawashima-fj:
@hppritcha I cannot observe the hang in my environment. Is there more information to reproduce it?

My environment:

- CPU: Thunder X
- OS: CentOS 7.2.1603
- GCC: 4.8.5 20150623 (CentOS rpm)
- Open MPI: 2c86b87 (latest master)
- configure: both `--disable-debug` and `--enable-debug`
- test: `make -C test check`

kawashima-fj:
@hppritcha I can observe the hang in Jenkins.
https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.platforms/Platform=ARMv8/1849/consoleFull
But I still cannot observe the hang in my environment. (I ran `make check` 100 times.)

This may related to an old issue #3697 (#3988).

hppritcha:
@kawashima-fj how many cpus does your aarch64 system have?

shamisp:
Lock implementation can depend on compiler and uarch.

shamisp:
Does it use gcc AMOs implementation or OMPI's ?

hppritcha:
Hmm...If I take the patch from #4566 and configure with --disable-builtin-atomics, the opal_fifo doesn't hang on my aarch64 system. gcc 4.8.5,  cpuinfo:
```
processor	: 0
BogoMIPS	: 500.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x1
CPU part	: 0xd07
CPU revision	: 2
```
I'm running opal_fifo in a loop, letting it go for hundreds of iterations.  using default configure options, the test typically hangs after a few iterations.

hjelmn:
@shamisp When using gcc builtins we end up using a bad compare-exchange 128 implementation (lock-based). I think we need to improve the configury and make ompi use the ompi atomic implementations for aarch64.

shamisp:
What version of GCC ?

hjelmn:
[...]

Issue #3697, opened 2017-06-14
Hang in mca_mpool_hugepage_module_init() on ARM64


yosefe:
## Background information

```
               Package: Open MPI mtt@hpc-arm-01 Distribution
                Open MPI: 2.1.2a1
  Open MPI repo revision: v2.1.1-55-g4d82554
   Open MPI release date: Unreleased developer copy
                Open RTE: 2.1.2a1
  Open RTE repo revision: v2.1.1-55-g4d82554
   Open RTE release date: Unreleased developer copy
                    OPAL: 2.1.2a1
      OPAL repo revision: v2.1.1-55-g4d82554
       OPAL release date: Unreleased developer copy
```

* Operating system/version: **RedHat 7.2**
* Computer hardware: **aarch64 (ARM 64-bit)**
* Network type:  **Infiniband**

-----------------------------

## Details of the problem

A hang in ctxalloc test during MPI_Init.
Similar hang is observed in many other tests.
ctxalloc can be found [here](http://trac.mpich.org/projects/mpich/browser/test/mpi/comm/ctxalloc.c)

```
mpirun -np 192 -mca btl_openib_warn_default_gid_prefix 0 \
  --bind-to core -mca pml ucx \
  -x UCX_NET_DEVICES=mlx5_0:1 -mca btl_openib_if_include mlx5_0:1 \
  -x UCX_TLS=rc,sm -mca opal_pmix_base_async_modex 0 \
   -mca mpi_add_procs_cutoff 100000 --map-by node \
ctxalloc 2 1500 100
```
stack trace:
```
Thread 1 (Thread 0x3ffb800a480 (LWP 27563)):
#0  0x000003ffb7ccfa74 in nanosleep () from /usr/lib64/libpthread.so.0
#1  0x000003ffb7911e0c in _opal_lifo_release_cpu () at ../opal/class/opal_lifo.h:195
#2  0x000003ffb7911e40 in opal_lifo_pop_atomic (lifo=0x620520) at ../opal/class/opal_lifo.h:210
#3  0x000003ffb7911fd0 in opal_free_list_get_st (flist=0x620520) at ../opal/class/opal_free_list.h:213
#4  0x000003ffb7911ff4 in opal_free_list_get (flist=0x620520) at ../opal/class/opal_free_list.h:225
[...]

PR #4641, opened 2017-12-19
doc: Add README note about ARM/POWER hangs

bwbarrett:
As documented in #4563 and #3697, there is an issue on ARM and
POWER platforms when the atomic fifo assembly isn't inlined,
which manifests as a hang.  Document the issue and the
work-around until a proper fix is committed.


bwbarrett:
bot:mellanox:retest

PR #4642, opened 2017-12-19
Detect/warn of illegal node names

rhc54:
If we detect that someone has given us an incorrect node name, provide a helpful message telling them as it is almost certainly a typo.


rhc54:
bot:mellanox:retest

Issue #3954, opened 2017-07-24
Fortran/C PMPI profiling wrappers


markalle:
I found an old discussion here
https://www.open-mpi.org/community/lists/devel/2015/08/17842.php
that led to the current situation where libmpi_mpifh.so defines Fortran entrypoints like mpi_send() with PMPI_Send() as the C call doing the main work.

The topic is whether wrapping the C MPI calls should lead to automagic wrapping of the Fortran MPI calls. The standard allows but doesn't require it (the requirement is just that an MPI implementation document whether it happens or not). I'll cut-and-paste that part:
```
3. document the implementation of different language bindings of the MPI interface if
they are layered on top of each other, so that the profiler developer knows whether
she must implement the profile interface for each binding, or can economize by implementing
it only for the lowest level routines.
4. where the implementation of different language bindings is done through a layered
approach (e.g., the Fortran binding is a set of “wrapper” functions that call the C
implementation), ensure that these wrapper functions are separable from the rest of
the library.
```
As far as I know, item 4 is the only reason it's traditional to separate Fortran mpi_send() etc into their own libmpi_mpifh library instead of the main libmpi.

So currently if a libwrap.so just defines its own C entrypoints like MPI_Send() calling PMPI_Send(), that libwrap.so won't intercept a Fortran mpi_send().  (App -> mpi_send -> libmpi_mpifh.so -> PMPI_Send -> libmpi.so).

The former behavior where libmpi_mpifh.so defined mpi_send() and called MPI_Send() would have caused Fortran MPI to be wrapped for free (App -> mpi_send -> libmpi_mpifh.so -> MPI_Send -> libwrap.so -> PMPI_Send -> libmpi.so).

And to summarize the IPM-related issue from the old discussion, libipm.so wraps both C and Fortran MPI calls for itself:
IPM defines MPI_Send() etc calling PMPI_Send()
IPM defines mpi_send() etc calling pmpi_send()
so that led to Fortran MPI calls being accounted for twice (App -> mpi_send -> libipm.so -> pmpi_send -> libmpi_mpifh.so -> MPI_Send -> libipm.so -> PMPI_Send -> libmpi.so).

I agree we have to support IPM without the double accounting, but I also think it's more user friendly to allow the automagic wrapping of Fortran MPI entrypoints.

[...]

PR #4651, opened 2017-12-20
Pmpi fortran

markalle:
This is a convenience feature for PMPI wrapper library writers, that allows them to have F77 language wrapping for free (something allowed but not required by the standard).

Eg if a wrapper library redefines
    int MPI_Send(...args...) { ... record stuff... ; rv = PMPI_Send; return(rv); }
and a Fortran app calls mpi_send() it's possible to have the mpi_send() definition in fortran/mpif-h/send_f.c call MPI_Send which would resolve back out into the wrapper library.

A longer discussion about this feature is here:
    https://github.com/open-mpi/ompi/issues/3954

The main argument against this is it's not required by the standard, and hence there's no good way for wrapper library writers to use this functionality: If they want to support all MPI vendors they can't rely on having access to the above. If they want to be complete, they have to write their own mpi_send() as well, and call either pmpi_send or PMPI_Send from it.

The argument for this feature is that pragmatically these C-only wrapper tools already exist, and customers try to use them with their Fortran apps, so it's a nice convenience feature to offer (and it's not random, it's an optional part of the standard).

This PR changes the fortran/mpif-h/send_f.c (etc) code from a direct PMPI_Send() call to a function pointer where the function pointer defaults to PMPI_Send(), and gives runtime control:
    -mca mpi_fortcall PMPI    :    the default behavior (mpi_send calls PMPI_Send)
    -mca mpi_fortcall MPI     :    make mpi_send call MPI_Send

The code's got an extra
#if OMPI_FORTRAN_USE_FPTR
that isn't being used right now, but is a stub for the idea of making this configurable. Eg if someone wanted no trace of the above changes and just wanted the original direct PMPI_Send() calls.

### DIFF
--- a/ompi/mpi/fortran/base/Makefile.am
+++ b/ompi/mpi/fortran/base/Makefile.am
@@ -50,5 +50,17 @@ libmpi_fortran_base_la_SOURCES = \
         conversion_fn_null_f.c \
         f90_accessors.c \
         strings.c \
-        test_constants_f.c
+        test_constants_f.c \
+        constructed_code_done.h
+
+BUILT_SOURCES = constructed_code_done.h
+
+here_src = $(top_srcdir)/ompi/mpi/fortran/mpif-h
+constructed_code_done.h: $(here_src)/normalize_mpih.pl $(here_src)/mkcode.pl $(top_srcdir)/ompi/include/mpi.h.in
+	rm -f flist && \
+	perl $(here_src)/normalize_mpih.pl $(top_srcdir)/ompi/include/mpi.h.in > flist && \
+	perl $(here_src)/mkcode.pl
+
+clean-local:
+	rm -f constructed_code_*.h
 endif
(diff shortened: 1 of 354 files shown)


ggouaillardet:
@markalle what happens if a `PMPI_*` subroutine is invoked from Fortean ?
Does this PR  guarantee the corresponding `MPI_*` C subroutine will never be invoked ?

markalle:
Oh no, I think you're right. The way this is done there's only one send_f.o and it uses pragma weak to create both mpi_send and pmpi_send which aren't distinguished from each other. So yeah, a fortran app calling pmpi_send() would go back out into MPI_Send() and through any wrapper that might exist there.

I don't see how to fix that without a more significant change to separate the behavior of mpi_send() from pmpi_send().

Hmm, the current PR is within my threshold of "triviality" for adding convenience features like this, but I'm not so sure about the next step of doubling up on the mpi_send()/pmpi_send() definitions... I may have to retract this PR then and go back to our layered entrypoint feature which is a different solution to the same problem.

jsquyres:
@markalle Given your last reply, I marked this PR as "WIP DNM" (work in progress / do not merge).  Feel free to update this PR or close it.

AboorvaDevarajan:

awlauria:
@markalle did you still want this PR or are you planning to pick it up again?

lanl-ompi:

hppritcha:
closing no intent to merge

Issue #4643, opened 2017-12-19
libfabric sockets provider failure on master


bwbarrett:
From https://github.com/open-mpi/ompi/commit/285fc42b4ea9c16c8e7b9caa1c5def453f80e557, we introduced a regression on master with the sockets provider in libfabric.  This change was merged into v3.0.x and v3.1.x, so tagged those targets as well.

Since this appears to be a regression, I've tagged it as a blocker.  It would be good if someone who knows libfabric better than I do can chime in on release blocker or not.

aravindksg:
As @chrisdolan mentioned, the sockets provider could be a little smart in catching `EINTR` (can either retry or map it to `-FI_EAGAIN` and return to user). But for now, a quick and dirty fix in OFI MTL could be to check for `-EINTR` and return from `ompi_mtl_ofi_progress()`. @bwbarrett : Would this be acceptable as a temporary fix while we ask OFI folks regarding adding error handling to sockets provider?

bwbarrett:
@aravindksg, yeah, I think that would be a reasonable fix (to handle it in ompi_mtl_ofi_progress()).

aravindksg:
Cherry picked the above patch to release branches as well. PR's #4656, #4657, #4658

aravindksg:
The fix has been merged into master and release branches. This issue can be closed.

PR #4653, opened 2017-12-20
MTL OFI: Allow retries in MTL progress for interrupted syscalls

aravindksg:
This fixes a regression in sockets provider which could return -EINTR value
from fi_cq_read() due to a syscall being interrupted. The error value is
currently interpreted as fatal condition. Relax the rule so that we can retry
fi_cq_read() operation.


aravindksg:
@bwbarrett : Fixes issue #4643

PR #4654, opened 2017-12-21
dist: Update library versions for 3.0.1

bwbarrett:
Bump release versions based on 'git diff --name-only v3.0.0 HEAD'.
A convertor initialization option change effectively adds an
interface in libopal (although linking against an old version
of libopal will work with the new library) and an interface change
in ompio_common means that it's not backwards compatible.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -82,16 +82,16 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=40:0:0
+libmpi_so_version=40:1:0
 libmpi_cxx_so_version=40:0:0
 libmpi_mpifh_so_version=40:0:0
 libmpi_usempi_tkr_so_version=40:0:0
 libmpi_usempi_ignore_tkr_so_version=40:0:0
 libmpi_usempif08_so_version=40:0:0
-libopen_rte_so_version=40:0:0
-libopen_pal_so_version=40:0:0
+libopen_rte_so_version=40:1:0
+libopen_pal_so_version=41:0:1
 libmpi_java_so_version=40:0:0
-liboshmem_so_version=40:0:0
+liboshmem_so_version=41:0:1
 libompitrace_so_version=40:0:0
 
 # "Common" components install standalone libraries that are run-time
@@ -100,7 +100,7 @@ libompitrace_so_version=40:0:0
 # components-don't-affect-the-build-system abstraction.
 
 # OMPI layer
-libmca_ompi_common_ompio_so_version=40:0:0
+libmca_ompi_common_ompio_so_version=41:0:0
 
 # ORTE layer
 libmca_orte_common_alps_so_version=40:0:0


VERSION
```diff
@@ -100,7 +100,7 @@ libompitrace_so_version=40:0:0
 # components-don't-affect-the-build-system abstraction.
 
 # OMPI layer
-libmca_ompi_common_ompio_so_version=40:0:0
+libmca_ompi_common_ompio_so_version=41:0:0
```
hppritcha: are you sure you don't want to change the age here as well?

bwbarrett: I don't think so?  As I mentioned in the commit log, there was a non-backward compatible change in the interface into the library.

rhc54: I think @hppritcha is correct - here are the libtool rules:
```
#
# - If any interfaces have been added, removed, or changed since
#   the last update, increment current, and set revision to 0.
#
# - If any interfaces have been added since the last public release,
#   then increment age.
#
# - If any interfaces have been removed since the last public release,
#   then set age to 0.
```
So an interface was added, but nothing removed - which means you increment the first number, set the second to 0, and increment the third.

bwbarrett: An interface was "changed", not "added".  See mca_common_ompio_build_io_array() in https://github.com/open-mpi/ompi/commit/0e0d45f733544476cdd22c4a1a9d765ed2df42a9#diff-498be05eb92cb6fc482cd0125a5f2021.  It added a new parameter, which is an interface change.  So I incremented current, set revision to 0, and didn't change age (which was already 0).

rhc54: Sorry, I thought I saw somewhere above that it was added - yes, your version looks correct to me.

rhc54:
Just as point of confusion for myself: I guess we define ABI as meaning you can compile against v3.0.0, and then still execute if your LD_LIBRARY_PATH is set to v3.0.1 - correct? Specifically, we don't support linking against a v3.0.0 libmpi, but using v3.0.1 components - yes?

I'm probably wrong, but it seems to me that the ompio change would preclude someone using a v3.0.0 libmpi (which includes the older version of common/ompio) and then calling an v3.0.1 ompio framework component that calls the new common function - won't that segfault? Ditto for the other way around?

hppritcha:
Yes to your first and second questions.  

I think in the case of these common libs we are okay wrt to the answer of yes to the first question since an application itself doesn't not directly depend on a common *.so.x.y.z lib.

bwbarrett:
@rhc54, my understanding is that we (traditionally) have considered the ABI to include the interfaces between the big three libraries (libmpi, libopen-rte, and libopen-pal), as well as the tooling interfaces and the "functionality" interfaces (for example, removing mxm doesn't change the ABI, but clearly changes expected behavior for MXM users).

The common libraries aren't part of that interface, so they're not part of our ABI stability guarantees.    Frankly, there's no useful way for a customer to run into a problem, either.  The customer would have to do something to their LD_LIBRARY_PATHs to cause the linker to cause problems here (although you're absolutely correct that if you managed to get a 3.0.1 libmpi.so linking against a 3.0.0 common_ompio would, best case, segfault).  But to do that, you'd almost have to explicitly link against 3.0.1's libmpi.so by full path and set LD_LIBRARY_PATH to 3.0.0.  But in that case, with my version change, you'd get a load error, so it would be obvious what happened.

Man, I dislike libraries :)

jsquyres:
Just for clarification: the ABI guarantees only apply to the MPI and OSHMEM APIs (not ORTE or OPAL or any of the components): https://github.com/open-mpi/ompi/blob/master/README#L1482-L1640, in particular the very end of that section: https://github.com/open-mpi/ompi/blob/master/README#L1623-L1631

rhc54:
I think maybe we need to define a new term - Application Binary Compatibility (ABC) - to reflect the broader issue we are facing with containers entering the picture. I fear that ABI is no longer sufficient in all situations, and my question was really more motivated by the ABC issue. Still, the answers were sufficient for my question.

PR #4659, opened 2017-12-22
Add missing Fortran 2008 binding subroutines

yanagibashi:
added missing Fortran 2008 binding pmpi_{*} subroutines to Open MPI.

Signed-off-by: Tsubasa Yanagibashi <fj2505dt@aa.jp.fujitsu.com>


jsquyres:
@kawashima-fj In the future, please don't merge without CI.  You can reply "ok to test" when someone outside the org submits a PR that needs to go through CI (that's why the bot asked "Can one of the admins verify this patch?" -- because @yanagibashi is not part of the Open MPI github org)

kawashima-fj:
@jsquyres Sorry, I didn't know the difference in CI. Now I get it.

PR #4661, opened 2017-12-22
opal/asm: add atomic min/max convenience functions

hjelmn:
This commit adds atomic functions for min/max.


hjelmn:
Will be using these in vader to support MCA_BTL_ATOMIC_MIN and MCA_BTL_ATOMIC_MAX so figured should put them in opal.

bosilca:
For consistency reasons I would add the 2 flavors fetch_and_op and op_and_fetch. Otherwise the patch looks good.

hjelmn:
Ok, done. Both fetch-op and op-fetch are in this PR.

PR #4662, opened 2017-12-22
v2.x: rcache/grdma: fix crash when part of a registration is unmapped

hjelmn:
This commit fixes an issue when a registration is created for a large
region and then invalidated while part of it is in use.

References #4509

(cherry picked from commit 39d598899bcd1dff2bbbe6f6eff9aa09ccb810b4)


jsquyres:
@hppritcha Good to go.

PR #4667, opened 2017-12-26
fortran: Fix PMPI interface bugs in mpi_f08 module

kawashima-fj:
This PR brings bug fixes which were cared in `MPI_` interface but were not cared in `PMPI_` interfaces of the Fortran `mpi_f08` module.

As the result, `ompi/mpi/fortran/use-mpi-f08/mod/pmpi-f08-interfaces.F90` with `sed -e s/PMPI_/MPI_/g -e s/pmpi_/mpi_/g` becomes same as `ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-interfaces.F90` except copyright statements.

@jsquyres, you changed `PMPI_Aint_add_f08` and `PMPI_Aint_diff_f08` to subroutines from functions in your 258d1aa1607. I supposed it was an unintentional change and I reverted them to functions in my d4fc404dc67. Could you confirm it just in case?

Better fix is generating `pmpi-*` files from `mpi-*` files. I remain it as a future work.

I'll create PRs for release branches to handle this PR and #4659.

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-interfaces.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-interfaces.F90
@@ -9,6 +9,7 @@
 ! Copyright (c) 2012      Inria.  All rights reserved.
 ! Copyright (c) 2015-2017 Research Organization for Information Science
 !                         and Technology (RIST). All rights reserved.
+! Copyright (c) 2017      FUJITSU LIMITED.  All rights reserved.
 ! $COPYRIGHT$
 !
 ! This file provides the interface specifications for the MPI Fortran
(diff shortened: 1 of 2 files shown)


jsquyres:
@bwbarrett It breaks ABI for `PMPI_Aint_{add|diff}` (by changing them from subroutines to functions).  But the OMPI was *wrong* (i.e., the MPI spec says they must be functions).  Therefore, no user applications could have been using these functions.

This is worth a note in NEWS, though.

kawashima-fj:
@bwbarrett @jsquyres My commit message d4fc404 explains it.

> Strictly speaking, this change involves binary incompatibility.
> But nobody used these subroutines and nobody will be affected because
> these subroutines were useless (didn't return a calculated value).

Is it OK to merge this change to release branches (v2.0, v2.1, v3.0, v3.1)?

The NEWS item will be something like this:

> - Correct PMPI_Aint_{add|diff} to be functions (not subroutines) in the Fortran mpi_f08 bindings.

bwbarrett:
@kawashima-fj, yes, I think we should merge to the impacted release branches (although v2.0.x is dead, so don't worry about that branch).

PR #4668, opened 2017-12-26
fortran: Call PMPI from PMPI_Status_set_cancelled_f08

kawashima-fj:
One more fix for Fortran PMPI with `mpi_f08`.

This is a bug which was forgotten to change in c08f97b0304.


kawashima-fj:
bot:mellanox:retest

PR #4670, opened 2017-12-27
opal/bitmap: fix opal_bitmap_set_bit()

ggouaillardet:
Correctly reallocate the bitmap when needed


ggouaillardet:

also, when `new_size > bm->max_size`, we do not `realloc` all the bytes, and setting the requested bit will cause an array overflow. should we error or simply do nothing in this case ?

bosilca:
I quickly check the usages of the bitmap and we in the current code we never go outside the original size. But having checks for corner cases is good practice.

jsquyres:
@bosilca This still has the corner case checks, right?  It looks like a minor optimization to me.

bosilca:
@jsquyres you're right. This is good to go.

PR #4674, opened 2017-12-28
use PMIx to manage CIDs

ggouaillardet:
Optimize CID assignment by using PMIx

This is currently at very early prototype stage.

Use PMIx server so an unused CID can be retrieved in one shot for all the tasks running on the same node when working on an intra communicator.

when invoking `ompi_comm_nextcid_nb()`, first gather the list of participating ranks on the local task with the lower rank. Then this "master" task will ask PMIx a CID unused by all the local participants and then "broadcast" it to the non master tasks.

Several things have to be fixed yet
 - return the CID to PMIx when no more used
 - finalize the PMIx interface (`PMIx_CID` is not a great name, right)
 - support external PMIx that to not implement `PMIx_CID`

Currently, the optimized path is only taken with intra communicators, and is a three step tango
 - gather the participants on the "master" task (`MPI`)
 - get a CID for the participants (master makes a single `PMIx` call)
 - broadcast the assigned CID by the master to the other tasks (`MPI`)

There are two ways of improving this on top of my head
 - perform the broadcast/gather at the `MPI` level on inter communicators too
 - all tasks call `PMIx` and `PMIx` performs a (kind of) collective operation (that would work regardless whether the communicator is intra or inter)


rhc54:
Just a suggestion: instead of creating a new API, just use the PMIx_Job_control_nb function and define new attributes. You could have one (or perhaps all?) proc call it to "init a counter" (and giving it a name), maybe allowing passage of another attribute for "scope" (using the PMIx scope values). Subsequent calls might use PMIx_Get to "get" the counter, adding an attribute to "increment the counter after get by all procs in scope".

Needs refinement, of course, but might simplify things.

rhc54:
Small self-debate over using job control. This isn't directly a "control", and so one could argue for a new API. On the other hand, one can envision using the counter(s) for control purposes, perhaps even using attributes to request that the host RM take certain actions upon reaching a specified counter level.

Up to you - just passing along some thoughts.

ggouaillardet:
Thanks @rhc54 for the insight !

From a design point of view, i am still on the fence regarding whether the collective part should happen at the `MPI` or `PMIx` level.
If `PMIx` happens to be the way to go, then i can definitely use `PMIx_Job_control_nb` that could return a local key name, and then `PMIx_Get` this key. That would allow to reuse some code (e.g. notify all the peers when the key has been set), and also spare the need to define a new API.

rhc54:
Personally, I'd opt for having PMIx (and thus, the host RM) perform the collective so it would be more broadly applicable to other programming paradigms. I see a common need for counters (both global and local) in various models. The only caveat is the need for the host RM to provide a fast collective, but we are seeing that coming into play (e.g., what Mellanox did for SLURM in the PMIx plugin).

ibm-ompi:

 Gist: https://gist.github.com/14cd34cd2f6c3307ae70927dbdddcc48

ibm-ompi:

 Gist: https://gist.github.com/bd7b783159cf566fa66c619b8937a009

artpol84:
bot:mellanox:retest

artpol84:
@ggouaillardet what is the main purpose of this PR? Performance optimization of Comm create operations?
If so can you share performance evaluation results?

ggouaillardet:
@artpol84 the goal is to improve CID allocation performance.

long story short, with `n` MPI tasks on `N` nodes 
 - all MPI tasks have a agree on the common lowest available CID
 - in the worst cases, that can take up to `n` iterations if the CID spaces are very fragmented
 - with this PR, the worst case would take "only" `N` iterations
 - by using `PMIx`, the optimization also works with direct launch
 - at this stage, this is work in progress, and when the code gets stable, i will run some benchmarks
   (or i will likely ask some help since i currently do not have access to a large cluster)

artpol84:
```
in the worst cases, that can take up to n iterations if the CID spaces are very fragmented
```
I was touching the current algorithm some time ago and may not recall all of the details. But  I think that it's not only `n` or `N`, fragmentation also has its say.
Consider 2 ranks with following used CIDs:
`rank=0: 0, 2, 4, 6, 8`
`rank=1: 1, 3, 5, 7, 9`
Lowest common CID will be 10 and I believe it will take current algorithm 5 steps:
* step1: r0:1, r1:2
* step2: r0:3, r1:4
* ...

ggouaillardet:
you are absolutely right !
that being said, by using PMIx, a MPI tasks on a local node can agree upon the lowest common available CID in one shot, so that should decrease the total number of iterations.

artpol84:
bot:mellanox:retest

artpol84:
Can you briefly explain how you use PMIx? Fence-like exchange?

ggouaillardet:
i added the new `PMIx_CID` API. as of now, the master process (e.g. the local task with the lowest rank) ask the server a CID for the list of local tasks. the list of local tasks is "locally gathered" at the MPI level, and then, once the algorithm completes, the result is "locally broadcasted" at the MPI level too.

non master MPI tasks still perform a bunch of `allreduce` but with a neutral argument.

i am still on the fence whether the MPI stuff should be done at the PMIx level or not.

artpol84:
And how server knows the CID? Is it stored in the Key?

artpol84:
I’m referring to this:
‘master process (e.g. the local task with the lowest rank) ask the server a CID for the list of local tasks.’

ggouaillardet:
`pmix_rank_info_t` was added a `pmix_bitmap_` of available CIDs

ggouaillardet:
Via the newly introduced `PMIx_CID()` API.
The client basically forwards the request to the server, that replies back the new CID.

rhc54:
Just to be clear: I believe we had converged on not introducing a new PMIx API, but instead to use the PMIx_Job_control_nb API to create/increment generic counters (using PMIx scope values to indicate their range) that OMPI could use as CIDs. It would then be up to the RM to determine the method of ensuring uniqueness across the indicated scope.

bosilca:
@artpol84 I don't agree with your description of the algorithm. If it is then the CID algorithm is wrong.
Starting with:
```rank=0: 0, 2, 4, 6, 8```
```rank=1: 1, 3, 5, 7, 9```
the first step should be instead ```P0:1, P1:2```, but then the second proposal is ```P0:3, P1:2``` (and not ```P0:3, P1:4``` as suggested). The reason is that at that point there is not enough information exchanged between processes to decide that 2 is not a good candidate (as an example P1 does not know if P0 has key 2 reserved).

**Comment by bosilca** [comment]
@ggouaillardet, there is a very good reason to do the CID allocation the way we do it today.

Your approach leaves gaps in the CID space seen by each process, as a CID allocated by whatever centralized entity will be globally uniquefor the entire job, independent on the processes participating in each communicator. Thus, non-overlapping communicators will not be able to have the same CID (unlike today). In other words you are converting a locally-dense CID list into a globally-dense CID list, and this is bad for memory consumption and/or performance. The locally-dense CID guarantees that from each process perspective the CID space is as compact as possible. This compactness ensure optimality in terms of memory needed to store the array, and in terms of access time when doing the cid-to-comm translation we need for each incoming MPI message.

If we decide to relax the performance friendly array translation, then a CID selection becomes a single broadcast, as once the participants decide for a leader (operation local to each process), this leader can pick a CID, add/shift it's daemon ID to make it unique (as an example), and broadcast it to all participants. This will however require to support a sparse CID space, and thus use a hash table to translate CID into comm pointer on each process. The impact on injection rate will certainly be significant, especially in multi-threaded scenarios.

**Comment by artpol84** [comment]
@bosilca let me explain scription.

* At the first step both ranks select the lowest available CID, rank0 - 1, rank1 - 2.
* They do [allreduce with MAX operation](https://github.com/open-mpi/ompi/blob/master/ompi/communicator/comm_cid.c#L338) that I believe will result in CID=2 from rank1.
* On the next stage each rank checks if this MAX value is available locally, rank0 will find it busy and set his flag to 0, rank1 obviously will be ok with this CID.
* Next they do [another allreduce with MIN operation](https://github.com/open-mpi/ompi/blob/master/ompi/communicator/comm_cid.c#L391) to converge on the flag. As everybody will see that flag is 0 (rank0 wil determine this) they [will go ahead with (current_cid + 1)](https://github.com/open-mpi/ompi/blob/master/ompi/communicator/comm_cid.c#L448) which is going to be 3.
* On the second step both ranks will search starting from CID=3 and this will result in rank0 - 3, rank1 - 4.

**Comment by rhc54** [comment]
@bosilca I'm not sure I fully understand your argument. Surely this is simply a question of the algorithm used to identify the desired counter value, isn't it? For example, why can't we have a counter that tracks non-overlapping groups of participants? This would seem to result in the same values you desire, but without multiple steps.

As for whether you have some "root" proc compute it or not - I have no preference either way. I do believe we should implement counter support in PMIx regardless as I can see uses for it beyond CID assignments.

**Comment by artpol84** [comment]
@bosilca 
```
Your approach leaves gaps in the CID space seen by each process, as a CID allocated by whatever centralized entity will be globally uniquefor the entire job, independent on the processes participating in each communicator.
```

If I understood @ggouaillardet correctly this is not what he is proposing. He actually tries to reduce number of participants in the next CID calculation down to 1 proc per node.
In his proposal:
* every process will publish his bitmap of available CIDs to PMIx in the format that is know by PMIx server.
* During next CID calculation process will determine a master for each node. This master is going to request next available CID on this node from the PMIx server. And since PMIx server will have the full bitmaps it will be able to calculate node-local next CID.
* Then they perform an inter-node communication using MPI.

I have concerns about performance of this operation for the following reasons:
* It involves PMIx server on the "critical path". Server that is not bound to any core and most probably will not be running at the time of invocation and this will introduce delays. If this approach would assume (and maybe it is) fetching data directly from the shared memory key storage it could be better, but still key fetch overhead may be relatively high.
* Second - as I noted this is not just a problem of `n` processes or `N` nodes. The main issue here is the fragmentation of the CID space. In the example that I provided for 2 procs you will need 5 steps even though you have only 2 procs. If I am missing something and @bosilca version is correct - it will be even more steps.

**Comment by rhc54** [comment]
Urrr...no, I don't think that is what @ggouaillardet is proposing. We propose that:

* some proc at the beginning of time initializes a counter using PMIx_Job_control_nb, possible requesting the "non-overlapping participants" algorithm for it, and specifying its scope (e.g., node-local vs nspace-wide).

* when a group of procs wants a new CID, they request one, specifying the counter to be used to generate it. The request flows up to the RM (if the counter isn't node-local - otherwise, the PMIx server simply handles it), which computes/retrieves the next value. This value is communicated down to all participants.

I'm not sure if it will be faster or slower than what we have today. There are certainly fewer required steps, but the overall time may be longer - need to test it to see. Probably will depend significantly on the inter-node transport underlying the RM for non-node-local counters.

**Comment by artpol84** [comment]
@rhc54 it seems that I misunderstood the concept. I was confused by the description [here](https://github.com/open-mpi/ompi/pull/4674#issuecomment-356534193):
```
i added the new PMIx_CID API. as of now, the master process (e.g. the local task with the lowest
rank) ask the server a CID for the list of local tasks. the list of local tasks is "locally gathered"
at the MPI level, and then, once the algorithm completes, the result is "locally broadcasted"
at the MPI level too.
```

Specifically by the phrase ` master process ask the server a CID for the list of local tasks`. This seemed to be a node-local related operation.

**Comment by artpol84** [comment]
If CID agreement is going to be handled by PMIx I'm very concerned about performance. Currently we have a relatively slow but generic algorithm compared to MPICH. We can't make it even slower.

@ggouaillardet from @rhc54 description it seems like you are proposing an algorithm similar to [what MPICH implements](https://github.com/open-mpi/ompi/issues/3368#issuecomment-294561519) at MPI level but based on PMIx level.

@ggouaillardet was there any particular use case that you are optimizing for?

**Comment by rhc54** [comment]
Keep in mind that @ggouaillardet's original statement was focused solely on the node-local case, not application-wide. I think this thread may have expanded well beyond his target.

As for performance - let's at least prototype something and see before we declare it to be slow. We may find it faster than what we currently do, especially when large numbers of communicators are being declared. Won't know until someone explores it.

**Comment by artpol84** [comment]
@rhc54 same here. I'd be curious to see and I will be happy to have it if it performs better. My point was that we need to thoroughly analyze the performance before merging.

**Comment by ggouaillardet** [comment]
Folks,

let me try to answer all the questions
 - everything started from an enduser worried about `MPI_Comm_dup()` and `MPI_Comm_split()` [performances at scale](https://mail-archive.com/devel@lists.open-mpi.org/msg20432.html) on [K computer](https://en.wikipedia.org/wiki/K_computer) (80,000 nodes and 8 cores per node)
- there was a consensus the bottleneck is CID allocation can take some time when the CID space is heavily fragmented (fwiw, some high level numerical libraries do create a lot of subcommunicators internally)
 - new method will mainly benefit many core systems (KNL, post-K, ...) 
 - the goal of this PR is to correctly implement a new method, collect performance numbers, and decide whether this should be merged or updated or discarded
 - i did read (and i think i understood too) [Trade-offs in Context Identifier Allocation in MPI](www.icl.utk.edu/files/publications/2016/icl-utk-988-2016.pdf) @bosilca (co)authored
- in Open MPI, CID to communicator is as simple as accessing the `ompi_mpi_communicators` `opal_pointer_t` array, and in order to minimize memory consumption, we need to use the lowest CIDs
- allocating a CID basically means all (participating) MPI tasks must agree on the lowest available CID, this is an iterative algorithm that involves per each iteration
  - selecting a low and available CID per MPI task
  - a first `MPI_Allreduce(..., MPI_MAX, ...)` to get the global lowest CID participating MPI tasks __might__ agree on
  - a second `MPI_Allreduce(..., MPI_MIN, ...)` to check whether the participating MPI tasks could agree or not

  if the CID spaces are heavily fragmented, this algorithm might take a bunch of iterations in order to complete, and my gut feeling is that most of the time is spent in `MPI_Allreduce()`.
 - reducing the number of iterations will hence improve performances.
 - in order to reduce the number of iterations, my simple idea is to use a divide and conquer approach 
   1. have all the participating MPI tasks running on a given node agree __locally__ on the lowest available CID
   2. and then try to agree globally
 - though the first step could be an interative task running at the MPI level, i think it would be faster to involve `PMIx` that can achieve that in one step. the key point is everything occurs here at the __node__ level (e.g. no inter-node PMIx to PMIx communication is involved)
 - from an implementation point of view, each MPI tasks still has its `ompi_mpi_communicators` array, and this is "mirrored" in the local PMIx server in a `pmix_bitmap_t` per `pmix_rank_info_t`.
  - requesting the common lowest available CID for a set of __local__ tasks is a trivial operation, and i do not see any value in having both an `orted` and `SLURM` (or whatever RM that supports PMIx) implementation, so for now, i chose to start with a new API (and hence both `mpirun` and direct launch can be supported), 
 - if we need support for more than one bitmap and/or in the global scope (vs local only for now), then i will be happy to move to the existing `PMIx_Job_control()` API (and RMs will need to be enhanced in order to support direct launch)

Bottom line
 - first, let's gather performance numbers, and then decide if/how to move forward
 - we still allocate the __common lowest CID__, and only the internal implementation differs
 - PMIx do __not__ perform any remote communications
 - there is no __global centralized__ CID whatsoever
 - this PR currently uses the new `PMIx_CID()` API, but at this early stage, this is really an implementation detail that can be changed later.
 - the `ompi_communicator_t` destructor will definitely become slower since it has to notifiy the local `PMIx` server that the CID should be released (that being said, this does not have to be a blocking operation).
 - allocating a CID with only one task per node will also be slower, but my gut feeling is that local CID allocation is still order(s) of magnitude faster than two `MPI_Allreduce()` (and yes, `MPI_Comm_dup(MPI_COMM_SELF)` will always be slower.

I welcome any comments on that, regardless they are based on numbers, past experience and/or gut feeling.

Though the following is unrelated to this PR, here is a new and independent idea to reduce the number of iterations.
Let's say we `MPI_Comm_split(MPI_COMM_WORLD)` between odd and even MPI ranks.
Do we __need__ the same CID on both sub-communicators ?
i do not think so, and in this case, we could update the algo in order to allocate two CIDs on two distinct subsets of `MPI_COMM_WORLD`. basically, we would `MPI_Allreduce()` two `MPI_INT` at a time, which should not be (significantly) more expensive than the same operation with one `MPI_INT`. Best case scenario is this will reduce the number of iterations *and* CID fragmentation.

**Comment by rhc54** [comment]
I agree with taking those steps. I've convinced myself that having a new PMIx API (should this prove useful) would be a good idea, but we cannot isolate it to just MPI CIDs, so we'll need to generalize...but we can cross that bridge if/when this proves worthwhile.

**Comment by artpol84** [comment]
Thanks @ggouaillardet. So my original interpretation of this PR was correct. 
Let me confirm one more thing: am I right that if global consensus wasn’t reached then master will have to request next node-local lowest CID?
We obviously need this as multiple independent comm creations may progress in parallel.
Thinking further about performance evaluation of this we will need to check the case where we check this case where PMIx server may become a bottleneck.

**Comment by ggouaillardet** [comment]
@artpol84 yes, this is still an iterative algorithm, so master might have to request a next node-local (well, restricted to the participating local MPI tasks) lowest CID.
note since we request (and release the previous one in one shot) yet an other CID, that means two more `MPI_Allreduce()` will be performed, and i am still under the assumption that a local PMIx request (that does not involve any blocking synchronization on the server) is order(s) of magnitude faster than `MPI_Allreduce()`.

**Comment by rhc54** [comment]
I'll just add this point and then shut up as I honestly don't care about communicator creation 😄 

Back in the dim, dark past, OMPI created new CIDs by having the root process get the ID from mpirun, and then MPI-broadcast it to the rest of the participants...and the loop_spawn test ran roughly 10x faster than it does today. I base that performance estimate on the fact that we had to up the MTT timeouts for that test by 10x when we moved to the current approach.

As @bosilca points out, the concern at that time was with the memory footprint caused by gaps the algo left in the communicator array of the individual procs as they participated in various different communicator creation operations. The algo was also slower for comm_dup, which could otherwise be done strictly locally, though alternative solutions were offered at the time. Hence the move to today's iterative approach, which trades speed for memory footprint.

I honestly don't know if that is the right trade-off, or if there is some middle ground. Just please stop saying that the old approach "doesn't scale" - it clearly scaled (from a time consideration) quite well. The issue wasn't in time to create the CID - it was with memory footprint.

**Comment by artpol84** [comment]
Need to test some changes in Mellanox jenkins.

**Comment by artpol84** [comment]
bot:mellanox:retest

**Comment by artpol84** [comment]
looks like here Jenkins detects some issues with this PR

**Comment by ggouaillardet** [comment]
@artpol84 there is an error that is evidenced by this error, but i am afraid i lack the skills to investigate it further.

I added some traces, and here is what happens. the issue can be evidenced with `ucx` and only 2 processes on the same node.
 - `shmem_init()` is invoked
 - it ends up calling `PMPI_Comm_dup(MPI_COMM_WORLD, ...)`
 - it ends up calling `ompi_comm_next_cid_nb()`
 - an intra node is performed. the local task with the lowest rank `irecv()` an int from the others (that `isend()` an int.
 - the request completes with an error (UCX message truncated)

by default, the status of the requests are not checked, so the crash that might occur later is really a side effect.

i do not know how to troubleshoot that, nor why the issue does not pop up without this PR.
(by default, we do not `isend/ircev` directly, but we `iallgather` and that does involve `isend/irecv` under the hood).

could the issue be `ucx` is not ready to send/recv at that time ?

**Comment by rhc54** [comment]
This is the error that is the subject of #4686, I believe. The UCX connection would explain why the other CI checkers and @jsquyres have been unable to replicate it.

**Comment by artpol84** [comment]
@rhc54 #4686 is about vader hang. This error is different

**Comment by artpol84** [comment]
@ggouaillardet how close is this work to a merge? I guess it's not there yet. I will investigate once there will be intent to merge this.

**Comment by rhc54** [comment]
Afraid you missed the Tues telecon where we reviewed a variety of errors that showed up recently in the MTT results. All, including the vader hang and this one, showed different signatures, but a likely common root cause of the atomics. The question was how best to reproduce it. One solid signature was in the dynamics (comm_spawn and friends) that was showing on one cluster - the other was one when UCX was used (not saying that UCX is the root cause - only that the code path seemed more likely to trip the error).

**Comment by artpol84** [comment]
@bosilca @ggouaillardet @jladd-mlnx 
I was thinking about alternative solutions to this issue and I wonder if we can just increase CID to be 64 bits.
This will open a huge space of possible CIDs and we will be able to use Allreduce with MAX operation and always converge in one iteration.
I guess that this approach may possibly introduce the issue on the comm fastpath, but this a kind of solution that I'd better invest my time as it seems very beneficial if efficient implementation in OMPI can be found. 
I also believe that this will outperform even centralized approach that @rhc54 was referring as a previous implementation, as there mpirun seems a bottleneck + OOB-based communication.

**Comment by artpol84** [comment]
```
One solid signature was in the dynamics (comm_spawn and friends) that was showing on one cluster
```
Is this also related to Mellanox verification or unrelated cluster?

**Comment by rhc54** [comment]
Unrelated - Cisco cluster

**Comment by artpol84** [comment]
I see, thank you

**Comment by ggouaillardet** [comment]
@artpol84 this PR will not be merged in an immediate future.

When you get some time, i invite you to read  Trade-offs in Context Identifier Allocation in MPI at http://www.icl.utk.edu/files/publications/2016/icl-utk-988-2016.pdf

Long story short, in order to convert a CID (e.g an `int`) to an `ompi_communicator_t *`, we simply `opal_pointer_array_get_item()` which is basically accessing an array.
The pro is it is very fast. The cons is we want to always allocate the lowest CID in order to keep the size of this array low. @bosilca discusses other approaches in this paper, and this PR is a different (and hopefully more efficient) way to allocate the __lowest__ CID.

**Comment by artpol84** [comment]
```
Long story short, in order to convert a CID (e.g an int) to an ompi_communicator_t *, we simply opal_pointer_array_get_item() which is basically accessing an array.
The pro is it is very fast. The cons is we want to always allocate the lowest CID in order to keep the size of this array low. @bosilca discusses other approaches in this paper, and this PR is a different (and hopefully more efficient) way to allocate the lowest CID.
```
This is what I meant by issues on the comm fastpath.
I also noticed that OB1 only supports only 16bit CIDs. And I the same applies to UCX.
I don't know if 64K of communicators is sufficient, but if it is I think that current algo and it's descendants a too complex because they are trying to solve this for 32bit while this CID space can never be used because of PML limitations.
In this circumstances a better I believe that even naive Reduce of a bitmap (MPICH way) could be better suited as it has a fixed complexity. For apps that create just a few comms it won't be a fastpath, but for those who do that frequently - this might be an opt.

ggouaillardet:
Should we then create a new framework for CID management first ?
And then provide various components optimized for various use cases ?

@bosilca any thoughts ?

artpol84:
+1 I was about to propose that.

ggouaillardet:
:bot:mellanox:retest

ggouaillardet:
@artpol84 i fixed the crash ...
the root cause was `isend(MPI_INT)` but `irecv(MPI_BYTE)`

i have no idea why only SHMEM+UCX was the only one crashing ...

ggouaillardet:
:bot:mellanox:retest

artpol84:
:+1:

ggouaillardet:
@bosilca @artpol84 the second commit creates the `cid` framework in order to allocate CIDs.

it has both blocking (e.g. `nextcid`) and non blocking (e.g. `nextcid_nb`) variants.
the rationale is only `MPI_Comm_idup()` requires a non blocking allocation mechanism, and other subroutines (that are likely to be used more frequently) so they could be optimized by using blocking collectives that have less overhead.

When moving `comm_cid` into the `cid/basic` component, i noted the `activate()` function shares some structures with `getnextcid()`, so i created `activate` and `activate_nb` callbacks. They might not be specific to a given component, so the implementation is in `cid/base`. I am not sure that was the right move, I could have left everything in `ompi/communicator`, or directly call `cid/base` and do not add yet an other callback. Anyway, I am very open to reconsider this.

Last but not least, the `query()` callback has the `ompi_mpi_thread_multiple` parameter, so some components can be optimized for non `MPI_THREAD_MULTIPLE` apps.

AboorvaDevarajan:

gpaulsen:
@ggouaillardet This looks like a good direction.  Any idea where this PR stands?  Could it be rebased and potentially applied to master?

ibm-ompi:

 Gist: https://gist.github.com/9d84cc55f41f04fe20d100ec76c474af

hjelmn:
Should really take a look at what we are doing in the sessions prototype. There we break the need for most CID agreement.

rhc54:
Let's just close this one as it is terribly stale and would need to be redone anyway.

PR #4677, opened 2018-01-04
hwloc: updates for hwloc 2.0.x API

jsquyres:

@rhc54 @bgoglin Please review.  These are a handful of changes that I needed to make the OMPI master branch compile with the (external) hwloc 2.0b1 tarball.  Please especially note the inclusion of `"hwloc/shmem.h"` -- is that right?  I know it's right outside of an OMPI environment, but is it right *insight* the OMPI build system?  I ask because we use `#include MCA_hwloc_IMPLEMENTATION_HEADER` to include `hwloc.h` -- do we need to do a similar thing for `"hwloc/shmem.h"`, or will `"hwloc/shmem.h"` *always* be correct?


rhc54:
Huh - looks like there were some additional changes to hwloc v2.0 since we last updated things. It looks okay to me. Might be a good idea to refresh the master's embedded hwloc code so it matches (as I believe you were only testing this against an external installation?).

We might need another definition for the shmem header as the path to that location may differ. Probably safest to do so.

ibm-ompi:

 Gist: https://gist.github.com/715ae6c3d09ec5ef2b3441b548fd93ca

ibm-ompi:

 Gist: https://gist.github.com/bd534a64a9df3d63788631a1a4e45db9

rhc54:
Yeah, you're getting failures because the embedded HWLOC needs to be updated

ibm-ompi:

 Gist: https://gist.github.com/690dc50804509d963adcc5f6f285a1ee

bgoglin:
Looks good to me. Thanks for doing this, I was planning to work on this next week.

I see that hwloc/shmem was included by opal/mca/hwloc/hwloc2a/hwloc2a.h in the embedded case. Should you should change that to work in both embedded/external hwloc (instead of hardwiring hwloc/shmem.h in other components, they don't even seem to #include hwloc.h)?

Can you wait for rc1 (within ~10 days) before merging this and updating the embedded hwloc? In case I have some additional API changes to push by then.

jsquyres:
@bgoglin No problem -- we can wait.  We can/should update the embedded hwloc at the same time as including this PR (i.e., updating the embedded hwloc might become another commit on this PR).

bgoglin:
Actually, don't bother with online and allowed sets anymore since OMPI doesn't set HWLOC_TOPOLOGY_FLAG_WHOLE_SYSTEM anymore.

jsquyres:
@bgoglin Good point; I removed that stuff.

I'm still curious about `"hwloc/shmem.h"`, but that might become more clear when we replace the embedded hwloc 2.0a with the 2.0.0rc1.

ibm-ompi:

 Gist: https://gist.github.com/f169364a16b6520cf4650077d69f272a

ibm-ompi:

 Gist: https://gist.github.com/f72844f3b9093a7fd9d1bc3a6b11c455

ibm-ompi:

 Gist: https://gist.github.com/c01cb653a2fc93dcf064bb2dee19912b

bgoglin:
@jsquyres Now that hwloc 2rc1 is out, you can update the internal hwloc and merge this. API changes in rc1 are unrelated to your patch as far as I see.

bgoglin:
If OMPI 3.1 is the first one with hwloc 2 support, it must get this PR.

jsquyres:
I have tested with hwloc-2.0.0rc1.  It looks like there is a new behavior, which causes a new problem.

## Short version

When I run:

```
$ mpirun -np 3 hostname
No objects of the specified type were found on at least one node:

  Type: NUMANode
  Node: savbu-usnic-a

The map cannot be done as specified.
```

Running with np=2 is fine.

## More detail

https://github.com/open-mpi/ompi/blob/master/orte/mca/rmaps/base/rmaps_base_binding.c#L922 is calling:

```c
bind_depth = hwloc_get_type_depth(node->topology->topo, hwb);
```

and then we check for failure by checking if `bind_depth` is less than 0.  However, according to the docs in hwloc 2.0.0 hwloc.h (https://github.com/open-mpi/hwloc/blob/master/include/hwloc.h#L730-L753):

```
 * If a NUMA node, I/O or Misc object type is given, the function returns a virtual
 * value because these objects are stored in special levels that are not CPU-related.
```

When I run with `rmaps_base_verbose=100`, it looks like it *is*, indeed, mapping by NUMA, which I double checked by adding a few calls to `opal_output()`:

```
...
[savbu-usnic-a:27072] [[14534,0],0]: Final mapper priorities
[savbu-usnic-a:27072]   Mapper: ppr Priority: 90
[savbu-usnic-a:27072]   Mapper: seq Priority: 60
[savbu-usnic-a:27072]   Mapper: resilient Priority: 40
[savbu-usnic-a:27072]   Mapper: mindist Priority: 20
[savbu-usnic-a:27072]   Mapper: round_robin Priority: 10
[savbu-usnic-a:27072]   Mapper: rank_file Priority: 0
[savbu-usnic-a:27072] mca:rmaps: mapping job [14534,1]
[savbu-usnic-a:27072] mca:rmaps: setting mapping policies for job [14534,1] nprocs 3
[savbu-usnic-a:27072] mca:rmaps[177] mapping not set by user - using bynuma
[savbu-usnic-a:27072] mca:rmaps[317] binding not given - using bynuma
[savbu-usnic-a:27072] mca:rmaps:ppr: job [14534,1] not using ppr mapper PPR NULL policy PPR NOTSET
[savbu-usnic-a:27072] mca:rmaps:seq: job [14534,1] not using seq mapper
[savbu-usnic-a:27072] mca:rmaps:resilient: cannot perform initial map of job [14534,1] - no fault groups
[savbu-usnic-a:27072] mca:rmaps:mindist: job [14534,1] not using mindist mapper
[savbu-usnic-a:27072] mca:rmaps:rr: mapping job [14534,1]
[savbu-usnic-a:27072] AVAILABLE NODES FOR MAPPING:
[savbu-usnic-a:27072]     node: savbu-usnic-a daemon: 0
[savbu-usnic-a:27072] mca:rmaps:rr: mapping no-span by NUMANode for job [14534,1] slots 16 num_procs 3
[savbu-usnic-a:27072] mca:rmaps:rr: mapping by slot for job [14534,1] slots 16 num_procs 3
[savbu-usnic-a:27072] mca:rmaps:rr:slot working node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps:rr:slot assigning 16 procs to node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps: assigning locations for job [14534,1]
[savbu-usnic-a:27072] mca:rmaps:ppr: job [14534,1] not using ppr assign: round_robin
[savbu-usnic-a:27072] mca:rmaps:resilient: job [14534,1] not using resilient assign: round_robin
[savbu-usnic-a:27072] mca:rmaps:mindist: job [14534,1] not using mindist mapper
[savbu-usnic-a:27072] mca:rmaps:rr: assign locations for job [14534,1]
[savbu-usnic-a:27072] mca:rmaps:rr: assigning procs to root level for job [14534,1]
[savbu-usnic-a:27072] mca:rmaps:rr:slot working node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps:base: computing vpids by slot for job [14534,1]
[savbu-usnic-a:27072] mca:rmaps:base: assigning rank 0 to node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps:base: assigning rank 1 to node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps:base: assigning rank 2 to node savbu-usnic-a
[savbu-usnic-a:27072] mca:rmaps: compute bindings for job [14534,1] with policy NUMA:IF-SUPPORTED[1003]
[savbu-usnic-a:27072] mca:rmaps: computing bindings for job [14534,1]
[savbu-usnic-a:27072] [[14534,0],0] reset_usage: node savbu-usnic-a has 3 procs on it
[savbu-usnic-a:27072] [[14534,0],0] reset_usage: ignoring proc [[14534,1],0]
[savbu-usnic-a:27072] [[14534,0],0] reset_usage: ignoring proc [[14534,1],1]
[savbu-usnic-a:27072] [[14534,0],0] reset_usage: ignoring proc [[14534,1],2]
[savbu-usnic-a:27072] JMS hwb=13 NUMANODE=13
[savbu-usnic-a:27072] JMS bind_depth=-3, DEPTH_NUMANODE=-3
No objects of the specified type were found on at least one node:

  Type: NUMANode
  Node: savbu-usnic-a

The map cannot be done as specified.
```

Here is a minimal patch to fix the issue, but it sure makes the code ugly / even more hard to follow.  Suggestions for a better patch are welcome:

```diff
diff --git a/orte/mca/rmaps/base/rmaps_base_binding.c b/orte/mca/rmaps/base/rmap
index df37999475..e2ce17d8f7 100644
--- a/orte/mca/rmaps/base/rmaps_base_binding.c
+++ b/orte/mca/rmaps/base/rmaps_base_binding.c
@@ -920,7 +920,12 @@ int orte_rmaps_base_compute_bindings(orte_job_t *jdata)
             } else
 #endif
                 bind_depth = hwloc_get_type_depth(node->topology->topo, hwb);
-            if (0 > bind_depth) {
+#if HWLOC_API_VERSION < 0x20000
+            if (0 > bind_depth)
+#else
+            if (0 > bind_depth && HWLOC_TYPE_DEPTH_NUMANODE != bind_depth)
+#endif
+            {
                 /* didn't find such an object */
                 orte_show_help("help-orte-rmaps-base.txt", "orte-rmaps-base:no-objects",
                                true, hwloc_obj_type_string(hwb), node->name);
```

ibm-ompi:

 Gist: https://gist.github.com/2611d15962b46cc9d8a7d3bc488f2fe1

ibm-ompi:

 Gist: https://gist.github.com/51001a6e4ea05abb5530fd315d1ec7da

rhc54:
The placement procedure is broken into two relevant parts for this discussion. The first pass assigns a locale to each process based on the user's map-by setting. If you like, this defines the envelope of where the process shall execute.

The second pass actually binds the process to a set of cpus based on the user's bind-to setting. There are three possible scenarios:

* bind-to and map-by match, meaning we are simply binding the process to the envelope to which it was assigned

* bind-to a level lower than what was assigned - e.g., binding to a single core after assigning the process to a socket. This is the bind-downward function.

* bind-to a level higher than what was assigned - e.g., binding to a NUMA region after assigning the process to an L2 cache. This is the bind-upward function.

Note that the entire procedure is based on the idea of a tree-like hierarchy - i.e., every element at a given level has a unique parent of a different type. Thus, we expect to find cpus assigned to a NUMANODE - if that is no longer true, then bind-to numa cannot work and we'll have to come up with an alternative procedure.

bgoglin:
There are still CPUs assigned to NUMA nodes, but NUMA nodes are not parents of PU anymore. NUMA nodes, PUs and other objects still have cpusets that can be used to compare localities.

So you want to find an object of a given type/depth that is either included (downwards) or larger (upwards) that locale. You just need to iterate over that level and find the first obj->cpuset that is included-in or contains locale->cpuset (this still works with NUMA nodes). hwloc_get_next_obj_covering_cpuset_by_depth() does the job for upwards, hwloc_get_next_obj_inside_cpuset_by_depth() does the job for downwards.
All of this works with v1.x and v2.x.

However in v2.x you need to change the way to decide between upwards and downwards. In rc1, I added hwloc_get_memory_parents_depth() which does what you want (return a normal level depth that has same locality as NUMA nodes and is part of the main tree, for instance return L3 depth if NUMA node are identical to L3 from a locality point of view). In theory, this function may return HWLOC_TYPE_DEPTH_MULTIPLE on future platforms if all NUMA nodes are not attached to same things (e.g. HBM on die, DRAM on package, NVRAM to the entire machine) but you can fail in that case for now.

I think we could even merge upwards and downwards (and ignore the above corner case) by just returning the first object that intersects locale->cpuset:
```
   hwloc_obj_t obj = NULL;
   while ((obj = hwloc_get_next_obj_by_depth(topology, targetdepth, obj)) != NULL) {
       if (hwloc_bitmap_intersects(locale->cpuset, obj->cpuset))
         return obj;
   }
   return NULL;
```
I can't tell for sure that these 6 lines can remove the current 200 lines of code :)

bgoglin:
@jsquyres I think I fixed the including of hwloc/shmem.h in the external case in https://github.com/bgoglin/ompi/commit/fecf4f8d2773ae33724a077c626f97155157c685
There's an update of the embedded hwloc to 2.0 almost-rc2.
I'll look at the binding issue now.

bgoglin:
My last commit is the implementation of the binding logic explained 9 days ago. Only compile tested because I don't know how/when this code is used. If you give me some command-lines to test things, I'll do that over the week-end.

ibm-ompi:

 Gist: https://gist.github.com/2b8e0acec305211106d98547af5a876d

ibm-ompi:

 Gist: https://gist.github.com/d6b9f67523c7a259e84defbad8605f33

ibm-ompi:

 Gist: https://gist.github.com/d41509b1949dd840184cb2752443a043

ibm-ompi:

 Gist: https://gist.github.com/e9706421a4ba1d3db07a8b037db33d42

ibm-ompi:

 Gist: https://gist.github.com/a0e687436a0a2b6cd6543a5cfed5d033

ibm-ompi:

 Gist: https://gist.github.com/3c57d243d185efd2f02702cda16fafe7

opal/mca/hwloc/external/external.h
```diff
@@ -48,6 +48,15 @@ BEGIN_C_DECLS
 #define HWLOC_OBJ_PACKAGE HWLOC_OBJ_SOCKET
 #endif
 
+#if defined(OPAL_HWLOC_WANT_SHMEM) && OPAL_HWLOC_WANT_SHMEM
+#    if HWLOC_API_VERSION >= 0x20000
+#        include MCA_hwloc_external_shmem_header
+#    else
+// JMS This isn't going to work - it will be an error if we use hwloc >= 2.0
+#        error Tried to include hwloc shmem header, but hwloc < 2.0 found
```
bgoglin: Maybe just do nothing in that case (no #include, no #error). All calls to what's defined in hwloc/shmem.h are protected by #if HWLOC_API_VERSION.
For the record, hwloc/shmem.h isn't included in hwloc.h because it was sort of experimental when we developed it 6 months ago.

jsquyres:
@bgoglin I finally got back to this.  I pushed a few more commits, such as taking your shmem fix, and a few fumbling (but not correct) updates to the rmaps base code.  I see the new diagram on page 49 of the hwloc 2.0rc2 letter PDF.  I'm struggling to reconcile that with what you say above.

The code currently figured out a mapping depth and a binding depth, and uses that to determine whether we want to bind upward or downward, and then calls `bind_upward()` or `bind_downward()`, as appropriate.

I'm confused in your explanations as to whether the NUMA node needs to be treated separately or not in your unified search, and whether the search you propose is for mapping/binding depth determination, or something else.

Can you explain?

If you're looking for some simple command lines to try:

```
$  mpirun -np 2 --mca rmaps_base_verbose 100 --report-bindings --map-by node --bind-to numa --debug-daemons hostname
```

You can change the `-bind-to` to be `numa`, `core`, `node`, `l[123]cache`, `hwthread`.  The `--report-bindings` option will tell you what OMPI did.

bgoglin:
My search replaces everything. map_depth isn't needed anymore (I forgot to remove its declaration in my last patch). No need for special cases for NUMA nodes (except when checking whether bind_depth<0). And no upwards/downwards search anymore. Only one generic search that starts from the map "locale", iterate over the level at bind_depth (without caring whether it above or below) and takes the first object whose bitmap intersects. Everything else isn't needed anymore.

bgoglin:
I pushed more commits to https://github.com/bgoglin/ompi/commits/master

All combination of --map-to --bind-to now work, even when using NUMA. I had to fix df_search_cores() and df_search() because the recursive search was hard to fix and can be implemented very easily a basic traversal of the level.

@rhc54 I need help understanding what OPAL_HWLOC_AVAILABLE means compared to OPAL_HWLOC_PHYSICAL/LOGICAL. Now that we don't have offline/disallowed PU anymore, the difference isn't clear. Does AVAILABLE include LOGICAL? Are PHYSICAL/LOGICAL allowed on searches that don't start at the root object?

df_search_min_bound() may need some similar updates but I don't know yet when it's used.

Overall, this seems to remove quite a lot of code specific to caches. If we use depth everywhere instead of type+cache_level, things are much simpler. I wonder if we really needed to replace OBJ_CACHE with L1..3CACHE in 2.0 now...

rhc54:
@bgoglin I'm not sure I grok your question about AVAILABLE etc. I thought PHYSICAL and LOGICAL were just alternative numbering methods to refer to the same cpu, yes? So AVAILABLE just consists of the cpus that this proc is allowed to use, regardless of whether you refer to it by its physical or logical identifier. The searches should only cover the AVAILABLE cpus.

IIRC, we didn't use depth because the depth of a given object type isn't consistent across sockets in hetero systems. So if we want to map procs to L3, for example, then we'd have to check the depth of L3 on each socket and map accordingly. Doable - just not sure it is much easier?

Totally up to you guys - you are welcome to modify!

bgoglin:
@rhc54 I see two cases where df_search() is called with AVAILABLE for getting a specific object by index. One with a core_id in socket_core_to_cpu_set(), and one btl smcuda where opal_hwloc_base_get_obj_by_type() ends up calling df_search(). So those are buggy?

Regarding depth, indeed two hosts could have different depths. I'll look at that later. I think I'll get more involved in that code in the future, we'll talk about this in Dallas.

bgoglin:
@rhc54 While you're here, can you explain what df_search_min_bound() does?

rhc54:
@bgoglin We decided a while back to only support logical cpu identifiers. I don't think those codes are "buggy" - they are intended to find the nth logical cpu within the available ones. I suspect we pass "available" just for historical reasons.

As for df_search_min_bound: it searches downward for the object of a given type that has the lowest number of procs bound to it, starting with the given starting object (i.e., it only considers object underneath the given one). It only considers objects that have available cpus - again, this was written when we had the entire topology to consider, including cpus and objects that were not available to us.

ggouaillardet:
i just pushed a new commit to replace `hwloc2a` with the newly released `hwloc 2.0.0`

bgoglin:
@rhc54 Turns out df_search_min_bound() is only used in opal_hwloc_base_find_min_bound_target_under_obj() which isn't used anywhere. I pushed the removal to https://github.com/bgoglin/ompi/commits/master together with my other patches and I will now do some testing.

rhc54:
@bgoglin Great - always nice to find and remove stale code!

rhc54:
Hmmm...FWIW: it _was_ used in rmaps_base_binding.c, but I guess your changes eliminated that use. So it wasn't stale, but the use was replaced by something else.

rhc54:
Absorbed into #5213

jsquyres:
Thanks @rhc54

Issue #4665, opened 2017-12-22
Dynamic host management/communicators still broken 


CalugaruVaxile:
Hi Dev team,

With OpenMPI 3.0.0 the code crashes when I try two MPI_Comm_spawn successive directives, while one single spawn works perfectly. I have downloaded and built the head master OpenMPI, same issue. OS = Debian 7. 

The full code is intended to dynamically add/remove hosts to/from a "uber" intracomm (which will be used by the main computation). Therefore I am trying to spawn+merge new communicators and separate/detach nodes in various sequences which will hopefully cover all the situations. 

"Prototype" Master code:

`
#include <iostream>
#include <cstring>
#include <mpi.h>
#include <unistd.h>
#include <limits.h>

using namespace std;

int main() {
  char slavejobtospawn[500];
  strcpy(slavejobtospawn, "./bug_slave");
  char localhost[HOST_NAME_MAX];
  gethostname(localhost, HOST_NAME_MAX);
  
  // == comms
  MPI_Comm worker_comm1_;
  MPI_Comm worker_comm2_;
  
  // == init MPI
  int provided;
  MPI_Init_thread(0, 0, MPI_THREAD_MULTIPLE, &provided);
  if (provided < MPI_THREAD_MULTIPLE) {
    cout << "ERROR: The MPI library does not have full thread support" << endl;
    MPI_Abort(MPI_COMM_WORLD, 1);
  };
  
  // == MPI_info obj
  MPI_Info minfo;
  MPI_Info_create(&minfo);
  
  // == first spawn
  MPI_Info_set(minfo, "add-host", "houprg118070,houprg118071");
  MPI_Comm_spawn(slavejobtospawn, MPI_ARGV_NULL, 
                 2, minfo, 0, MPI_COMM_WORLD, &worker_comm1_, MPI_ERRCODES_IGNORE);
                 
  // == second spawn
  MPI_Info_set(minfo, "add-host", "houprg118072");
  MPI_Comm_spawn(slavejobtospawn, MPI_ARGV_NULL, 
                 1, minfo, 0, MPI_COMM_WORLD, &worker_comm2_, MPI_ERRCODES_IGNORE);

  
  
  usleep(5000000);
  
  // == stop MPI
  MPI_Finalize();
  cout << "MASTER " << localhost << "  SHUTTING DOWN" << endl;        
  return 0;
}
`

"Prototype" Slave code
`
#include <iostream>
#include <cstring>
#include <mpi.h>
[...]

PR #4680, opened 2018-01-05
Continue resolving add_host behavior

rhc54:
Fix a problem in packing/unpacking job updates. There remains a race condition that causes messages to attempt to be sent to the second new daemon before it is completely ready. Not entirely sure where it is coming from.

Refs #4665

### DIFF
--- a/orte/mca/odls/base/odls_base_default_fns.c
+++ b/orte/mca/odls/base/odls_base_default_fns.c
@@ -109,9 +109,9 @@ int orte_odls_base_default_get_add_procs_data(opal_buffer_t *buffer,
     int rc, v;
     orte_job_t *jdata=NULL, *jptr;
     orte_job_map_t *map=NULL;
-    opal_buffer_t *wireup, jobdata;
+    opal_buffer_t *wireup, jobdata, priorjob;
     opal_byte_object_t bo, *boptr;
-    int32_t numbytes, numjobs;
+    int32_t numbytes;
     int8_t flag;
     void *nptr;
     uint32_t key;
@@ -270,16 +270,17 @@ int orte_odls_base_default_get_add_procs_data(opal_buffer_t *buffer,
             flag = 1;
             opal_dss.pack(buffer, &flag, 1, OPAL_INT8);
             OBJ_CONSTRUCT(&jobdata, opal_buffer_t);
-            numjobs = 0;
             rc = opal_hash_table_get_first_key_uint32(orte_job_data, &key, (void **)&jptr, &nptr);
             while (OPAL_SUCCESS == rc) {
                 /* skip the one we are launching now */
                 if (NULL != jptr && jptr != jdata &&
                     ORTE_PROC_MY_NAME->jobid != jptr->jobid) {
+                    OBJ_CONSTRUCT(&priorjob, opal_buffer_t);
                     /* pack the job struct */
-                    if (ORTE_SUCCESS != (rc = opal_dss.pack(&jobdata, &jptr, 1, ORTE_JOB))) {
+                    if (ORTE_SUCCESS != (rc = opal_dss.pack(&priorjob, &jptr, 1, ORTE_JOB))) {
                         ORTE_ERROR_LOG(rc);
                         OBJ_DESTRUCT(&jobdata);
+                        OBJ_DESTRUCT(&priorjob);
                         return rc;
                     }
                     /* pack the location of each proc */
@@ -287,32 +288,33 @@ int orte_odls_base_default_get_add_procs_data(opal_buffer_t *buffer,
                         if (NULL == (proc = (orte_proc_t*)opal_pointer_array_get_item(jptr->procs, n))) {
                             continue;
                         }
-                        if (ORTE_SUCCESS != (rc = opal_dss.pack(&jobdata, &proc->parent, 1, ORTE_VPID))) {
+                        if (ORTE_SUCCESS != (rc = opal_dss.pack(&priorjob, &proc->parent, 1, ORTE_VPID))) {
                             ORTE_ERROR_LOG(rc);
                             OBJ_DESTRUCT(&jobdata);
+                            OBJ_DESTRUCT(&priorjob);
                             return rc;
                         }
                     }
-                    ++numjobs;
+                    /* pack the jobdata buffer */
+                    wireup = &priorjob;
+                    if (ORTE_SUCCESS != (rc = opal_dss.pack(&jobdata, &wireup, 1, OPAL_BUFFER))) {
+                        ORTE_ERROR_LOG(rc);
+                        OBJ_DESTRUCT(&jobdata);
+                        OBJ_DESTRUCT(&priorjob);
+                        return rc;
+                    }
+                    OBJ_DESTRUCT(&priorjob);
                 }
                 rc = opal_hash_table_get_next_key_uint32(orte_job_data, &key, (void **)&jptr, nptr, &nptr);
             }
-            /* pack the number of jobs */
-            if (ORTE_SUCCESS != (rc = opal_dss.pack(buffer, &numjobs, 1, OPAL_INT32))) {
+            /* pack the jobdata buffer */
+            wireup = &jobdata;
+            if (ORTE_SUCCESS != (rc = opal_dss.pack(buffer, &wireup, 1, OPAL_BUFFER))) {
                 ORTE_ERROR_LOG(rc);
                 OBJ_DESTRUCT(&jobdata);
                 return rc;
             }
-            if (0 < numjobs) {
-                /* pack the jobdata buffer */
-                wireup = &jobdata;
-                if (ORTE_SUCCESS != (rc = opal_dss.pack(buffer, &wireup, 1, OPAL_BUFFER))) {
-                    ORTE_ERROR_LOG(rc);
-                    OBJ_DESTRUCT(&jobdata);
-                    return rc;
-                }
-                OBJ_DESTRUCT(&jobdata);
-            }
+            OBJ_DESTRUCT(&jobdata);
         } else {
             flag = 0;
             opal_dss.pack(buffer, &flag, 1, OPAL_INT8);
@@ -367,8 +369,8 @@ int orte_odls_base_default_construct_child_list(opal_buffer_t *buffer,
     orte_job_t *jdata=NULL, *daemons;
     orte_node_t *node;
     orte_vpid_t dmnvpid, v;
-    int32_t n, k;
-    opal_buffer_t *bptr;
+    int32_t n;
+    opal_buffer_t *bptr, *jptr;
     orte_proc_t *pptr, *dmn;
     orte_app_context_t *app;
     int8_t flag;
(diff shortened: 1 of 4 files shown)


rhc54:
@ggouaillardet Would you have a chance to try and track down the remaining race condition? I've done about as much as I can on this issue.

ggouaillardet:
Sure, I will give it a try.
What is the minimal number of hosts required to reproduce the issue ?

rhc54:
You just need three. Start with only one in your "allocation", and then change the names in the orte/test/mpi/add_host.c code so they match what you have available.

ggouaillardet:
@jladd-mlnx can you please double check this build failure ?
I saw a similar one in an unrelated PR and I am tempted to think the root cause is somewhere else
(E.g. from a PR point of view, these are false positive)

ggouaillardet:
I just started troubleshooting this issue.

it seems `mpirun` send an `xcast` message to the second `orted` before it is initialized.
at this stage, a theory i have to investigate is that the `xcast` is here to notify all the `orted` that the first 3 spawned tasks have exited, and this message is not meant to be sent to the second `orted`.

from a MPI point of view, the first task simply `MPI_Comm_disconnect()` the first three spawned tasks.
(e.g. it does __not__ wait for these task to exit), so a possible scenario is
 - `MPI_Comm_disconnect()` by MPI task `0`
 - `mpirun` launch `orted` on the second node
 - `mpirun` receives the notification the first 3 tasks have exited from the first `orted`
 - the second `orted` is still starting and has not yet notified `mpirun`
 - `mpirun` xcast to all `orted` (including the second one that is not yet ready) and that causes the abort.

i noted that
 - increasing `rml_base_max_retries` to a very high value does not help
 - inserting a `sleep(1)` before the second `MPI_Comm_spawn()` is enough to get rid of the error.

rhc54:
Ah, yes - that would explain it. One possibility would be to not increase num_procs or update the routing plan until _after_ the new daemons completely report back. That would prevent the xcast from going to an unstarted location. Currently, we update those in orte/mca/plm/base/plm_base_support_fns.c when we setup the vm - probably too early.

The catch would be that the HNP needs the new routing plan for the initial launch so it knows who its children are, and thus which daemons it should be directly starting when in a tree-spawn mode. Since we know when a job is a dynamic launch, perhaps the better answer is to (a) compute the num_procs and update the routing plan for non-dynamic launches, (b) do not use tree-spawn for dynamic launches (just have the HNP do it directly), and then (c) not update num_procs or routing plan until after daemons report back for dynamic launches.

Make sense?

artpol84:
bot:mellanox:retest

ggouaillardet:
@rhc54 i kept investigating this issue and found several issues
 1. per the various comments in open-mpi/ompi@92c996487c589ef8558a087ce2a9923dacdf0b99 (`mpirun` is invoked on `n0`, that will `add_host` `n1` and then `n2`) the `nidmap` is not updated on `n1` and hence `n1` is not made aware of `n2`. That can be fixed by resetting `orte_nidmap_communicated=false` after `mpirun` has launched the initial `orted` (well, that is zero in this case)
 2. there is also an issue when `no_tree_spawn` is `false`. For example, with `--mca routed_radix 1`, `orted` is __not__ spawned on `n2`. Only `n0` tries to spawn it, but since `n2` is a child of `n1`, then nothing happens. i guess the right fix here is to `xcast` the new `nidmap` and ask all `mpirun` and `orted` to spawn their children.
 3. i think there is an issue with `create_dmns()`. For example, if `add_host` is replaced with `host`, and `mpirun --host n0:1,n1:3,n2:3 --mca routed_radix 1 -np 1 ./host`, `mpirun` hangs with the latest log
`[n0:00119] [[11063,0],0] grpcomm:base: routed is NOT NULL, nexpected = 0`. the signature is `{job=2, vpid=ANY}`. `n2` is not a child of `n0` but (a child of `n1`, and hence) a grand child of `n0`, so i think `nexpected` should be `1` here.
 4. i incidentally noted that if `MPI_Comm_spawn()` is invoked with `add_host` and the host is already part of the initial allocation, (e.g. `mpirun --host n0:1,n1:3,n2:3 -np 1 ./add_host`), then `orted` is spawn'ed anyway (e.g. we end up with two `orted` per host). is this the intended behavior ? on one hand, the user gets what was asked for, so this is kind of an Open MPI standard behavior, but on the other hand, that might over complicate things from our point of view.

The inline patch below is a workaround that improves things.

It does not address the initial issue i reported (e.g. `xcast()` to a not-yet ready node). A cheap workaround would be to simply drop the message instead of returning an `ORTE_ERR_UNREACH` error.

```patch
diff --git a/orte/mca/plm/rsh/plm_rsh_module.c b/orte/mca/plm/rsh/plm_rsh_module.c
index 9f56be9..517e438 100644
--- a/orte/mca/plm/rsh/plm_rsh_module.c
+++ b/orte/mca/plm/rsh/plm_rsh_module.c
@@ -1281,6 +1281,8 @@ static void launch_daemons(int fd, short args, void *cbdata)
         OBJ_RETAIN(caddy->daemon);
         opal_list_append(&launch_list, &caddy->super);
     }
+    mca_plm_rsh_component.no_tree_spawn = true;
+    orte_nidmap_communicated = false;
 
     /* set the job state to indicate the daemons are launched */
     state->jdata->state = ORTE_JOB_STATE_DAEMONS_LAUNCHED;
```

rhc54:
```per the various comments in 92c9964 (mpirun is invoked on n0, that will add_host n1 and then n2) the nidmap is not updated on n1 and hence n1 is not made aware of n2. That can be fixed by resetting orte_nidmap_communicated=false after mpirun has launched the initial orted (well, that is zero in this case)
```
You don't want to do that. Instead, we should reset it to false only when we know we are adding a host.
```
there is also an issue when no_tree_spawn is false. For example, with --mca routed_radix 1, orted is not spawned on n2. Only n0 tries to spawn it, but since n2 is a child of n1, then nothing happens. i guess the right fix here is to xcast the new nidmap and ask all mpirun and orted to spawn their children.
```
Yes, I think what is missing is to pass the new nidmap to all currently-running orteds.
```
i think there is an issue with create_dmns(). For example, if add_host is replaced with host, and mpirun --host n0:1,n1:3,n2:3 --mca routed_radix 1 -np 1 ./host, mpirun hangs with the latest log
[n0:00119] [[11063,0],0] grpcomm:base: routed is NOT NULL, nexpected = 0. the signature is {job=2, vpid=ANY}. n2 is not a child of n0 but (a child of n1, and hence) a grand child of n0, so i think nexpected should be 1 here.
```
Sounds like a bug to me - not sure I fully understand the scenario.
```
i incidentally noted that if MPI_Comm_spawn() is invoked with add_host and the host is already part of the initial allocation, (e.g. mpirun --host n0:1,n1:3,n2:3 -np 1 ./add_host), then orted is spawn'ed anyway (e.g. we end up with two orted per host). is this the intended behavior ? on one hand, the user gets what was asked for, so this is kind of an Open MPI standard behavior, but on the other hand, that might over complicate things from our point of view.
```
We absolutely cannot wind up with more than one daemon on a host as that will mess up quite a few things. I would have thought that the add_host code would recognize that the host already exists and avoid duplicating it - if not, then that's a bug.

ggouaillardet:
__[EDIT]__ i edited the third point of my previous comment

ggouaillardet:
OK, let me rewrite (again) my previous 3rd point. (with `--mca routed_radix 1`)

 1. During the second `MPI_Comm_spawn()`, when starting the 3 MPI tasks on `n2`, the first `PMIx_fence()` leads to `orted` on `n2` calls `orte_grpcomm_API_allgather(sig={jobid=3, vpid=ANY})`
 2. the previously generated message is received by `orted` on `n1` by `allgather_recv()`, and is _forwarded_ to `n0`
 3. the previously message is received by `mpirun` on `n0` in `allgather_recv()`, that calls `orte_grpcomm_base_get_tracker(sig={jobid=3, vpid=ANY}, true)` that creates a new `orte_grpcomm_coll_t` with one participant that has index `2` (e.g. `n2`, so far so good)
 4. `n2` is __not__ a child of `n0` (but it is a grand-child from `n1`), and `n0` is __not__ a participant, so `coll->nexpected` is untouched and its value is `0`
 5. then `allgather_recv()` increments `coll->nreported` whose value is now `1`, and the condition `coll->nreported == coll->nexpected` can never be true.

at this stage, i think the bug is in `orte_grpcomm_base_get_tracker()`, `coll->nexpected` should be incremented each time a child __or its descendant__ is  a participant.

makes sense ?

ggouaillardet:
the inline patch below is an implementation of my previous comment.
i will do some more testing tomorrow.

```patch
diff --git a/orte/mca/grpcomm/base/grpcomm_base_stubs.c b/orte/mca/grpcomm/base/grpcomm_base_stubs.c
index d0d9757..b495093 100644
--- a/orte/mca/grpcomm/base/grpcomm_base_stubs.c
+++ b/orte/mca/grpcomm/base/grpcomm_base_stubs.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2011-2016 Los Alamos National Security, LLC. All rights
  * Copyright (c) 2016-2018 Intel, Inc.  All rights reserved.
- * Copyright (c) 2017      Research Organization for Information Science
+ * Copyright (c) 2017-2018 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -289,11 +289,15 @@ orte_grpcomm_coll_t* orte_grpcomm_base_get_tracker(orte_grpcomm_signature_t *sig
         /* cycle thru the array of daemons and compare them to our
          * children in the routing tree, counting the ones that match
          * so we know how many daemons we should receive contributions from */
+        orte_process_name_t child;
+        child.jobid = orte_process_info.my_name.jobid;
         OBJ_CONSTRUCT(&children, opal_list_t);
         orte_routed.get_routing_list(routed, &children);
         while (NULL != (nm = (orte_namelist_t*)opal_list_remove_first(&children))) {
             for (n=0; n < coll->ndmns; n++) {
-                if (nm->name.vpid == coll->dmns[n]) {
+                child.vpid = coll->dmns[n];
+                orte_process_name_t hop = orte_routed.get_route(routed, &child);
+                if (nm->name.vpid == hop.vpid) {
                     coll->nexpected++;
                     break;
                 }
```

rhc54:
Makes sense to me - thx!

ggouaillardet:
well, strictly speaking we are testing the __first__ hop __to__ a child, and what we really want to test is the __last__ hop __from__ a child. that is good enough for `radix` since there is a unique path between two peers regardless the direction, but that would not work for more complex algorithm such as fat-tree or routing with load balancing.

i will take a crack at `add_host` __not__ spawning a second `orted` per node tomorrow

rhc54:
After sleeping on it, I'm still confused by your third point. The current algorithm only counts the number of direct children in the routing tree. Since the operation is a collective, we must receive a message from each of those children. Any contributions from grandchildren, or daemons of any level beneath our direct children, will be "rolled up" in the message from our direct child and therefore don't get included in nexpected.

Perhaps we should split that one off into a separate issue until we can better understand it? Given that we haven't seen problems in that area, it may just be an artifact of other things we are playing with now.

For now, what if we modified things to the following:

* have the plm/rsh component set the no_tree_spawn flag to true after launch of the initial set of daemons so that all subsequent daemon launches are done directly from the HNP. This is the unusual case and so we don't have to worry about scaling.

* whenever we get add_host requests, reset the orte_nidmap_communicated flag to false - can do that in the orte/mca/ras/base code that processes add_host. That code should check to ensure the host isn't already known - if it is, then the add_host request can be ignored. This should resolve the multiple orted problem.

Make sense? I think that (plus the changes in my original PR) would solve the immediate problems.

ggouaillardet:
i'll (literally) sleep on that too shortly.

the current algorithm in `allgather_recv()` counts (when computing `nexpected`) the number of direct children __that are participants__ (plus itself if the current daemon is a participant).
at this stage, i believe the reason we stayed out of trouble so far is we do not test several `MPI_Comm_spawn()` when the total number of hosts is larger than the `routed_radix`.

My understanding is that `allgather` in `grpcomm` is a collective __for its participants only__.

For example, with `--mca routed_radix 2` and 4 daemons, i guess the topology is
```
0 --- 1 --- 3  
   +- 2
```

so if vpid 3 initiates an `allgather` with only its local processes, `0` is not a participant, it should expect nothing from `2` and one message from `3` that is received via `1`. In this case, from the `0` point of view, none of its (direct) children are participants, but one message should be expected from a child, so `nexpected` should be `1`.

> Since the operation is a collective, we must receive a message from each of those children.

in the case i just described, `2` that is a child of `0` will not send nor relay any message.
`0` is not a participant, it has 2 direct children, but `nexpected` should be `1`


makes sense ?

i will double check tomorrow this case hangs without my patch, and works with it.


thanks for the guidance in using `ras/base`, i will try that tomorrow.

i already set `no_tree_spawn` to `true` after the initial launch, and i agree scalability should not be an issue here.

i do not think `MPI` can issue a `del_host` kind of command. but is that something that should be supported by the `DVM` ? if yes, there could be some additional race conditions in that area too.

rhc54:
Hmmm...I think there is a misunderstanding here about how grpcomm works. _Every_ daemon is required to participate in a grpcomm operation, even if that participation is just "nothing to contribute". You can never have a situation where some daemon doesn't send a message. If you are encountering such a situation, then it is due to an error in the add_host operation, and we should fix that one.

We don't support delete host, even in the DVM, because of all the inflight issues it would create. I see no reason to spend effort on it.

rhc54:
I updated the PR, rebasing to resolve the conflict with master and addressing the two points I listed. See what you think.

NOTE: I added your copyright since we worked together on these changes. Please feel free to remove if you'd rather we didn't.

ggouaillardet:
Yep, there is a clear misunderstanding here.

note `add_host` is not involved here.

let's instead consider the following program with only one `MPI_Comm_spawn()` and the `host` info.

```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/types.h>
#include <unistd.h>
#include <sys/param.h>

#include <mpi.h>

int main(int argc, char* argv[])
{
    int msg, rc;
    MPI_Comm parent, child;
    int rank, size;
    char hostname[MAXHOSTNAMELEN];
    pid_t pid;
    char *env_rank,*env_nspace;
    MPI_Info info;

    env_rank = getenv("PMIX_RANK");
    env_nspace = getenv("PMIX_NAMESPACE");
    pid = getpid();

    printf("[%s:%s pid %ld] starting up on node %s!\n", env_nspace, env_rank, (long)pid, hostname);

    MPI_Init(NULL, NULL);
    printf("%d completed MPI_Init\n", rank);
    MPI_Comm_get_parent(&parent);
    /* If we get COMM_NULL back, then we're the parent */
    if (MPI_COMM_NULL == parent) {
        pid = getpid();
        printf("Parent [pid %ld] about to spawn!\n", (long)pid);
        MPI_Info_create(&info);
        MPI_Info_set(info, "host", "n2");
        if (MPI_SUCCESS != (rc = MPI_Comm_spawn(argv[0], MPI_ARGV_NULL, 3, info,
                                                0, MPI_COMM_WORLD, &child, MPI_ERRCODES_IGNORE))) {
            printf("Child failed to spawn\n");
            return rc;
        }
        printf("Parent done with spawn\n");
        if (0 == rank) {
            msg = 38;
            printf("Parent sending message to child\n");
            MPI_Send(&msg, 1, MPI_INT, 0, 1, child);
        }
        MPI_Comm_disconnect(&child);
        printf("Parent disconnected\n");
    } else {
        pid = getpid();
        printf("Hello from the child %d of %d on host %s pid %ld\n", rank, 3, hostname, (long)pid);
        if (0 == rank) {
            MPI_Recv(&msg, 1, MPI_INT, 0, 1, parent, MPI_STATUS_IGNORE);
            printf("Child %d received msg: %d\n", rank, msg);
        }
        MPI_Comm_disconnect(&parent);
        printf("Child %d disconnected\n", rank);
    }

    MPI_Finalize();
    fprintf(stderr, "%d: exiting\n", pid);
    return 0;
}
```

You can reproduce the issue with `mpirun --mca routed_radix 1 --host n0:1,n1:3,n2:3 -np 1 ./host`

from `orte_grpcomm_base_get_tracker()`
```c
        /* cycle thru the array of daemons and compare them to our
         * children in the routing tree, counting the ones that match
         * so we know how many daemons we should receive contributions from */
        OBJ_CONSTRUCT(&children, opal_list_t);
        orte_routed.get_routing_list(routed, &children);
        while (NULL != (nm = (orte_namelist_t*)opal_list_remove_first(&children))) {
            for (n=0; n < coll->ndmns; n++) {
                if (nm->name.vpid == coll->dmns[n]) {
                    coll->nexpected++;
                    break;
                }
            }
            OBJ_RELEASE(nm);
        }
```
to me, that strongly suggests that not all children are expected to send a message.
(otherwise, we would just `coll->nexpected = opal_list_get_size(&children)`, right ?)


i can observe
 - the 3 spawned MPI tasks call `PMIx_Fence()`
 - on `n2`'s `orted`, `PMIx` calls back `orte_grpcomm_API_allgather(sig={jobid=3, vpid=ANY})`
 - `create_dmns` sets only one participant (`n2`) since the newly spawned tasks are only on this node.
 - the message "flows" to `n1` and then `n0`
 - `n0` nor `n1` invokes `orte_grpcomm_API_alltather()`. we could add some logic to direct these daemons to issue an `allgather()` and do nothing. As far as i am concerned, i do not see any rationale for that.
- `allgather_recv` is invoked on `n0` when the message is "forwarded" from `n1`. currently `n0` nor `n1` invoke `orte_grpcomm_API_allgather()`, so `n0` should set `coll->nexpected=1` solely because `n2` is the only participant and is a grandchild of `n0` via `n1`.

currently, we only consider the children (and do not include other descendants), so `n0` hangs after
```
[n0] [[16396,0],0] grpcomm:direct allgather recv nexpected 0 nrep 1
```

a similar hang can be observed with
```
mpirun --mca routed_radix 2 --host n0:1,n1:1,n3:1,n2:3 -np 1 ./host2
```
same message before hanging
```
[n0] [[24838,0],0] grpcomm:direct allgather recv nexpected 0 nrep 1
```


both cases can be fixed with the patch i posted earlier
```patch
diff --git a/orte/mca/grpcomm/base/grpcomm_base_stubs.c b/orte/mca/grpcomm/base/grpcomm_base_stubs.c
index d0d9757..c49963e 100644
--- a/orte/mca/grpcomm/base/grpcomm_base_stubs.c
+++ b/orte/mca/grpcomm/base/grpcomm_base_stubs.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2011-2016 Los Alamos National Security, LLC. All rights
  * Copyright (c) 2016-2018 Intel, Inc.  All rights reserved.
- * Copyright (c) 2017      Research Organization for Information Science
+ * Copyright (c) 2017-2018 Research Organization for Information Science
  * $COPYRIGHT$
  *
@@ -289,11 +289,15 @@ orte_grpcomm_coll_t* orte_grpcomm_base_get_tracker(orte_grpcomm_signature_t *sig
         /* cycle thru the array of daemons and compare them to our
          * children in the routing tree, counting the ones that match
          * so we know how many daemons we should receive contributions from */
+        orte_process_name_t child;
+        child.jobid = orte_process_info.my_name.jobid;
         OBJ_CONSTRUCT(&children, opal_list_t);
         orte_routed.get_routing_list(routed, &children);
         while (NULL != (nm = (orte_namelist_t*)opal_list_remove_first(&children))) {
             for (n=0; n < coll->ndmns; n++) {
-                if (nm->name.vpid == coll->dmns[n]) {
+                child.vpid = coll->dmns[n];
+                orte_process_name_t hop = orte_routed.get_route(routed, &child);
+                if (nm->name.vpid == hop.vpid) {
                     coll->nexpected++;
                     break;
                 }

```

rhc54:
As I said, let's separate these things a bit.

First, are the add_host changes complete and correct? If so, then let's commit this PR.

Second, something sounds like it has bit-rotted in grpcomm. My prior comment was incomplete. All daemons must participate in some operations. However, it is true that we cannot always guarantee it. For example, we do require that if some procs register (i.e., call MPI_Init), then all procs must register. However, if none of the procs on a given node do so, then there is no way that daemon can know of the failure - and we would hang.

Your patch would _appear_ to address a possible issue. However, I would need time to refresh my memory of what I did back then so I can understand why there is now a problem, given that we have not heard of any issues to-date. I'm disturbed by your analysis as, even though only one node has spawned procs on it, all the nodes are supposed to respond to the launch cmd.

So let's separate that issue into its own tracker, and I'll try to find some time to look at it.

ggouaillardet:
Sure, let's do that !
this PR fixes some issues, so feel free to merge any time (you might want to finalize the commit message)
i will open a new issue from now.

PR #4683, opened 2018-01-08
Fix DIR, DIR/include search for --with-pmix

pkovacs:
This PR allows ompi to find external pmix headers that were installed to a `DIR/include/pmix` subdirectory. 

When packages contain multiple header files, it is common and desirable to be able to place the headers in a project-named subdirectory.   Doing so with external pmix, e.g. installing headers to `/usr/include/pmix`, renders the headers unusable to ompi. 

To address the issue, an optional new argument (action-if-failed) was added to `OPAL_CHECK_WITHDIR` so that, instead of halting the configure immediately, the caller can, if desired, capture the failure and try a different directory.  New code was added to `opal_check_pmi.m4` to do just that: first try DIR/include and then try DIR/include/pmix.

**./configure --with-pmix=/usr --with-libevent=/usr ...**
```
checking --with-external-pmix value... not found (/usr/include)
checking --with-external-pmix value... sanity check ok (/usr/include/pmix)
```

### DIFF
--- a/config/opal_check_package.m4
+++ b/config/opal_check_package.m4
@@ -52,6 +52,15 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_HEADER], [
 
     AS_IF([test "$opal_check_package_header_happy" = "no"],
           [AS_IF([test "$dir_prefix" != ""],
+                 [$1_CPPFLAGS="$$1_CPPFLAGS -I$dir_prefix"
+                  CPPFLAGS="$CPPFLAGS -I$dir_prefix"])
+          AC_CHECK_HEADERS([$2], [opal_check_package_header_happy="yes"], [], [$6])
+          AS_IF([test "$opal_check_package_header_happy" = "no"],
+                [# no go on the as is - reset the cache and try again
+                 unset opal_Header])])
+
+    AS_IF([test "$opal_check_package_header_happy" = "no"],
+          [AS_IF([test "$dir_prefix" != ""],
                  [$1_CPPFLAGS="$$1_CPPFLAGS -I$dir_prefix/include"
                   CPPFLAGS="$CPPFLAGS -I$dir_prefix/include"])
           AC_CHECK_HEADERS([$2], [opal_check_package_header_happy="yes"], [], [$6])
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -257,36 +257,18 @@ AC_DEFUN([OPAL_CHECK_PMIX],[
                  [pmix_ext_install_dir=/usr],
                  [pmix_ext_install_dir=$with_pmix])
 
-           # Make sure we have the headers and libs in the correct location
-           OPAL_CHECK_WITHDIR([external-pmix], [$pmix_ext_install_dir/include], [pmix.h])
+           AS_IF([test ! -z "$with_pmix_libdir" && test "$with_pmix_libdir" != "yes"],
+                 [pmix_ext_install_libdir="$with_pmix_libdir"])
 
-           AS_IF([test -n "$with_pmix_libdir"],
-                 [AC_MSG_CHECKING([libpmix.* in $with_pmix_libdir])
-                  files=`ls $with_pmix_libdir/libpmix.* 2> /dev/null | wc -l`
-                  AS_IF([test "$files" -gt 0],
-                        [pmix_ext_install_libdir=$with_pmix_libdir],
-                        [AC_MSG_CHECKING([libpmix.* in $with_pmix_libdir/lib64])
-                         files=`ls $with_pmix_libdir/lib64/libpmix.* 2> /dev/null | wc -l`
-                         AS_IF([test "$files" -gt 0],
-                               [pmix_ext_install_libdir=$with_pmix_libdir/lib64],
-                               [AC_MSG_CHECKING([libpmix.* in $with_pmix_libdir/lib])
-                                files=`ls $with_pmix_libdir/lib/libpmix.* 2> /dev/null | wc -l`
-                                AS_IF([test "$files" -gt 0],
-                                      [pmix_ext_install_libdir=$with_pmix_libdir/lib],
-                                      [AC_MSG_RESULT([not found])
-                                       AC_MSG_ERROR([Cannot continue])])])])],
-                 [# check for presence of lib64 directory - if found, see if the
-                  # desired library is present and matches our build requirements
-                  AC_MSG_CHECKING([libpmix.* in $pmix_ext_install_dir/lib64])
-                  files=`ls $pmix_ext_install_dir/lib64/libpmix.* 2> /dev/null | wc -l`
-                  AS_IF([test "$files" -gt 0],
-                        [pmix_ext_install_libdir=$pmix_ext_install_dir/lib64],
-                        [AC_MSG_CHECKING([libpmix.* in $pmix_ext_install_dir/lib])
-                         files=`ls $pmix_ext_install_dir/lib/libpmix.* 2> /dev/null | wc -l`
-                         AS_IF([test "$files" -gt 0],
-                               [pmix_ext_install_libdir=$pmix_ext_install_dir/lib],
-                               [AC_MSG_RESULT([not found])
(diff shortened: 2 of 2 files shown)


rhc54:
ok to test

config/opal_check_pmi.m4
```diff
@@ -358,9 +363,9 @@ AC_DEFUN([OPAL_CHECK_PMIX],[
            LDFLAGS=$opal_external_pmix_save_LDFLAGS
            LIBS=$opal_external_pmix_save_LIBS
 
+           opal_external_pmix_CPPFLAGS="-I$pmix_ext_install_include_dir"
```
ggouaillardet: we do __not__ want `-I/usr/include` in `CPPFLAGS`

pkovacs: @ggouaillardet I can filter that case, sure.  When the prefix is `/usr`, that var might be `/usr/include` or `/usr/include/pmix`.

pkovacs:
@ggouaillardet PMIx installs its multiple headers to `/usr/include` by default.  Goal here was not disallow the possibility that distros might want to install to a subdirectory corresponding to a package name., e.g. `/usr/include/pmix` via:

```
# pmix configure
./configure --includedir=/usr/include/pmix ,,,
```

The above provides better include organization.   I'm not changing any default behavior in any case, just supporting the above possibility.  I'll make the change you suggested above, filtering `/usr/include`.

rhc54:
I'm wondering if this is simply exposing a small problem in the pmix configure logic itself. We already have the ability to specify the include location separate from the lib directory. It sounds, though, like we aren't first testing the exact path given to us before appending "include" or other subdirectories.

If that is true, then we should probably just fix that problem.

pkovacs:
@rhc54 If you're suggesting that pmix should install headers to `%{prefix}/include/pmix` by default I'm quite happy with that, although ompi would still not be able to find them without this PR.

jsquyres:
FWIW, I think the way that we have handled this before is to do something like:

```
$ export FOO_DIRr=/path/to/foo
$ ./configure \
    --with-FOO=$FOO_DIR/include \
    --with-FOO-libdir=$FOO_DIR/lib ...
```

I.e., specify the includedir as the `--with-FOO` clause and then specify the libdir via `--with-FOO-libdir`.  We debated this back in the beginning (when creating the `--with-FOO[-libdir]` options) and decided on this approach instead of:

1. User specifies `--with-FOO`, or
2. User specifies `--with-FOO-incdir` and `--with-FOO-libdir`

The rationale was that checking for the all the corner cases and error cases for the above scenario was a royal pain, and just created yet another CLI option (i.e., the `--with-FOO-incdir` option).

Is using `--with-pmix=/usr/include/pmix --with-pmix-libdir=/usr/lib` not sufficient?

rhc54:
I fear you misunderstood me - let me clarify. We are happy with how/where PMIx installs. Per @jsquyres, we do have the two options. Is the problem simply that the configure logic isn't checking the --with-pmix=path-to-include-files argument before trying it with "include" appended?

artpol84:
bot:mellanox:retest

ggouaillardet:
@jsquyres if that works, that would be best for me !
i bet `-L/usr/lib` would end up in the `LDFLAGS`, but this is a bug that can be easily fixed.

@pkovacs `SLURM` is known to install its header in `/usr/include/slurm`, so it makes sense for Open MPI to try that automatically. If `PMIx` is not known to install its headers in `/usr/include/pmix`, i do not see any rationale on why Open MPI would search there automatically. This is only my view, so let's say this will become the de facto standard in a near future. In this case, i'd rather have `SLURM` and `PMIx` use the same logic (and i have no preference) in order to keep the code consistent and easier to maintain.

pkovacs:
Below is exactly the problem and my motivation here.

> Is using --with-pmix=/usr/include/pmix --with-pmix-libdir=/usr/lib not sufficient?

Does not work given the putative example of pmix headers installed to `/usr/include/pmix`.

ompi silently appends /include, looks in `/usr/include/pmix/include` and throws up its hands.

rhc54:
Understood - as I said, that is a bug in the OMPI configure logic. It is supposed to first check in the exact given path, and only if not found should it append "include"

pkovacs:
OK, now we're on the same page!  

My code needs to be revised.  Instead of attempting `DIR/include` and then `DIR/include/pmix`, I should search `DIR` then `DIR/include` to be consistent with the rest of the universe.

Got it!  Gimme a bit to correct the code here.

ggouaillardet:
Note `OPAL_CHECK_PACKAGE` is not used here, so that could explain the "missing logic"

jsquyres:
@ggouaillardet Excellent point that I totally missed.  Is there a reason `OPAL_CHECK_PACKAGE` is not used here?

ggouaillardet:
Could this be because of the many `PMIx` dependencies that make linking a dummy test program non trivial ?

rhc54:
Errr...."many dependencies"? I'm only aware of libevent

pkovacs:
There's extra code in the check to discern the pmix version (1x, 2x, 3x) from the headers.    Presumably
 `OPAL_CHECK_PACKAGE` can be rolled in front of that section.

artpol84:
bot:mellanox:retest

artpol84:
looks like leftover in `/dev/shm` after some failed runs has caused this failure.

artpol84:
bot:mellanox:retest

artpol84:
@jladd-mlnx @vspetrov looks like disabling hcoll helps.
Please let me know when the issue is fixed and we can turn it on again.
I also suggest to have some test running with hcoll disabled so it will be easier to identify issues in the future.

pkovacs:
Changes pushed.  The `OPAL_CHECK_PMIX` code is simplified  by using `OPAL_CHECK_PACKAGE`.  The check for `pmix_version.h` now uses the more robust `AC_CHECK_HEADER` with the CPPFLAGS returned from check package, instead of the manual `ls` command.   

Note that `OPAL_CHECK_PACKAGE` did not actually search `DIR` then `DIR/include` when using `--with=foo=<dir>` (to my surprise).   I added a clause there to perform the missing `DIR` check.

rhc54:
I'm not seeing the test for ```DIR/include``` - can you point it out to me?

pkovacs:
The `DIR/include` check was already there on lines lines 62-69 under my addition to `opal_check_package.m4`.

rhc54:
Got it - thanks!

config/opal_check_package.m4
```diff
@@ -50,6 +50,15 @@ AC_DEFUN([_OPAL_CHECK_PACKAGE_HEADER], [
                   [# no go on the as is - reset the cache and try again
                    unset opal_Header])])
 
+    AS_IF([test "$opal_check_package_header_happy" = "no"],
+          [AS_IF([test "$dir_prefix" != ""],
+                 [$1_CPPFLAGS="$$1_CPPFLAGS -I$dir_prefix"
+                  CPPFLAGS="$CPPFLAGS -I$dir_prefix"])
+          AC_CHECK_HEADERS([$2], [opal_check_package_header_happy="yes"], [], [$6])
```
alex-mikheev: @pkovacs @jsquyres @rhc54 @ggouaillardet @yosefe @jladd-mlnx This check is wrong when the package headers are installed in /usr/include  but configure was called with the specific package dir. The check will pickup h file from the /usr/include and add -I$dir_prefix to the CPPFLAGS which is wrong.

pkovacs: OK, so we can add `/usr/include` and `/usr/local/include` filters to the to code block above that (lines 44, 45).  That will handle the `--with-foo=/usr/include` and ``--with-foo='/usr/include/local` cases without adding unecessary CPPFLAGS
,
```
    AS_IF([test "$dir_prefix" = "/usr" || \
           test "$dir_prefix" = "/usr/local || \
           test "$dir_prefix" = "/usr/include || \
           test "$dir_prefix" = "/usr/local/include"],
```

pkovacs: Added https://github.com/open-mpi/ompi/pull/4722 to correct the concern above.

PR #4684, opened 2018-01-09
MPI_File_set_view(): check datatypes are monotonic

ggouaillardet:


ggouaillardet:

yes, there are 16 bits for an `opal_datatype_t` `flags`, but though `opal` uses bits `0-8`, bits `9,12-15` are used by `ompi`. i thought using available bits `10-11` in `opal` would be inelegant, and renumbering the flags would break backward compatibility. For that reason, i decided `opal_datatype_is_monotonic()` only performs a check, and `ompi_datatype_is_monotonic()` caches the results. as a consequence, bits `10-11` are now used by `ompi` (and there are no more available bits for future usage).

opal/datatype/opal_datatype_monotonic.c
```diff
@@ -0,0 +1,55 @@
...
+        OBJ_RELEASE(pConv);
+        return 0;
+    }
+
+    do {
+        iov_count = 5;
+        rc = opal_convertor_raw( pConv, iov, &iov_count, &max_data);
+        for (uint32_t i=0; i<iov_count; i++) {
+            if ((long)iov[i].iov_base < prev) {
+                monotonic = false;
```
bosilca: The monotonic is now forever false, you might just want to speed things up and jump out of both loops.

ggouaillardet: sure, can we do that in `C` ? if yes, then i forgot the syntax (!)
an other option is to `rc= 1; break;`, that will exit the `for` loop and terminate the `do while` loop.
of course, the good old `goto` is still around.
any recommendation ?

bosilca: that's when ```goto``` comes in handy ;)

PR #4698, opened 2018-01-11
Ensure the epilog gets executed in PMIx server

rhc54:
If we abnormally terminate, then we still want any cleanups to be
executed.


rhc54:
Before someone comments - yes, I know there are some debugs in here. Just want to see the output to ensure things are working okay in other environments.

rhc54:
Debugs removed

PR #4699, opened 2018-01-11
Convert nidmap to regx framework

rhc54:
Handle the need for different regex generator/parsers by moving the
orte/util/nidmap and orte/util/regex code into a new "regx" framework.
Use the original code to complete a "fwd" component, and create a
scaffold for IBM's "reverse" component.


rhc54:
Refs #4689 - actually, replaces it

PR #4704, opened 2018-01-12
orte/regx: fix, revamp and enhancement 

ggouaillardet:
Closes open-mpi/ompi#4689


ggouaillardet:
@jjhursey FYI

ggouaillardet:
note i do not build the `test/util/orte_nidmap` test by default.
otherwise, `make check` fails unless `make install` is performed (or at least one `regx` component is built statically)
and there is no `make install` on AWS.

an alternative is to build the `regx/fwd` component by default.

PR #4706, opened 2018-01-12
osc/rdma: add an mca parameter to list MTLs for which osc pt2pt should have higher priority than rdma and default to psm2 

matcabral:
porting #4575 to V2.x


hppritcha:
@jsquyres do you think we should merge this in to 2.x?

jsquyres:
@hppritcha I have no problem with this in v2.1.x.

PR #4711, opened 2018-01-14
coll/libnbc: do not recursively call opal_progress()

ggouaillardet:
instead of invoking ompi_request_test_all(), that will end up
calling opal_progress() recursively, manually check the status
of the requests.

the same method is used in ompi_comm_request_progress()

Refs open-mpi/ompi#3901

### DIFF
--- a/ompi/mca/coll/libnbc/nbc.c
+++ b/ompi/mca/coll/libnbc/nbc.c
@@ -10,7 +10,7 @@
  *                         rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC.  All rights
  *                         reserved.
- * Copyright (c) 2015-2017 Research Organization for Information Science
+ * Copyright (c) 2015-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  *
  * Author(s): Torsten Hoefler <htor@cs.indiana.edu>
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@bosilca per my comments on #3901, i had the feeling that `ompi_coll_libnbc_progress()` being called recursively was either sub-optimal or a bad design.
This PR should fix that.

Generally speaking, is there any legitimate use case in which `ompi_coll_libnbc_progress()` is called recursively ? an other way to put it, is do we still need `libnbc_in_progress` ?

bosilca:
Your change to the code looks legit. I don't have enough knowledge about libNBC to answer your question, but keep in mind that the progress can be also called every time we reach a resource exhaustion (such as fragments for BTL/PML). So, if you post new communications from a progress function, there is an opportunity to be called recursively (if new communications were posted from a progress function).

artpol84:
bot:mellanox:retest

artpol84:
vader issue, try again
bot:mellanox:retest

hjelmn:
Looks good to me.

ggouaillardet:
@kawashima-fj FYI

as pointed by @bosilca, `MPI_Isend()` might invoke `opal_progress()`, so there is still the possibility the `coll/libnbc` progress function is invoked recursively.

PR #4717, opened 2018-01-16
We now add all nodes to the job data object when we map, so don't do it twice - for v3.1.x

yosefe:
Cherry pick a fix by @rhc54 from master to v3.1.x


yosefe:
adding @karasevb and @artpol84

yosefe:
@rhc54 can you pls take a look?

jladd-mlnx:
@bwbarrett @rhc54 Could someone merge this, please? We need to get it covered under our MTT.

PR #4723, opened 2018-01-17
opal_progress: check timer only once per 8 calls

yosefe:
Reading the system clock on every call to opal_progress() is an
expensive operation on most architectures, and it can negatively affect
the performance, for example of message rate benchmarks.

We change opal_progress() to read the clock once per 8 calls, unless
there are active users of the event mechanism.

(cherry picked from commit 7cee60346e45f5fd22cf9a9e4c8655152ffb6c75)

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -168,23 +168,8 @@ opal_progress_finalize(void)
     return OPAL_SUCCESS;
 }
 
-
-/*
- * Progress the event library and any functions that have registered to
- * be called.  We don't propogate errors from the progress functions,
- * so no action is taken if they return failures.  The functions are
- * expected to return the number of events progressed, to determine
- * whether or not we should call sched_yield() during MPI progress.
- * This is only losely tracked, as an error return can cause the number
- * of progressed events to appear lower than it actually is.  We don't
- * care, as the cost of that happening is far outweighed by the cost
- * of the if checks (they were resulting in bad pipe stalling behavior)
- */
-void
-opal_progress(void)
+static int opal_progress_events()
 {
-    static volatile uint32_t num_calls = 0;
-    size_t i;
     int events = 0;
 
     if( opal_progress_event_flag != 0 ) {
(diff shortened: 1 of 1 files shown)


yosefe:
adding @bwbarrett

bwbarrett:
@rhc54, since I did the code review, can you do the RM approval blessing?

rhc54:
Done!

jladd-mlnx:
@bwbarrett Can you merge this, please, so that we can get this covered under our MTT?

bosilca:
Just to be sure we all understand that in a multi-threaded environment ```((num_calls++) & 0x7) == 0``` might seldomly, or never, be true.

bwbarrett:
@bosilca: that's a good point, although probably wouldn't happen in practice.

Any reason we shouldn't just make num_calls per-thread storage and avoid that problem?

bosilca:
Your proposal makes more sense than hoping that in practice things will work well.

Moreover, I haven't seen a discussion about the PR implications in a multi-threaded scenario. The "low priority" callbacks will be triggered too often overall, and more concerning they will be triggered concurrently by multiple threads. Thus, in addition to making sure the current low priority callbacks are are well protected, which means each component behind them has it's own thread-safety mechanism, we need to evaluate what impact this protection mechanism has on the overall performance.

This PR lacks any analysis outside the single threaded trivial benchmark case.

bwbarrett:
I agree we missed the subtle implication of being able to never see `% 8 == 0` be true, but I disagree that we changed anything about the low priority calling rate with this patch.  In both the previous case and this case, which ever thread saw `% 8 == 0` was going to call the lp callbacks.  There was never any check to make sure that was thread 0 or any specific thread.  There may be problems there, but that's outside of any expectation on this patch.

bosilca:
Ignoring the potential mishap with %, you are right regarding the overall number of calls to opal_progress. But the calling pattern drastically changed, with the current pattern having a higher chance of synchronizing the progress threads (and preventing any progress because they all block in a protected lp callback).

Anyway, from a practical perspective I don't think it really matters as the usage of lp queue in OMPI is extremely seldom. The function itself is used only once, in BML r2 to register BTL progress callbacks. From a logical perspective however, the lp queue is only used for BTLs that have no potential peers during the initial add_proc. Thus, for most static MPI applications the resulting callback has in fact no reason to be ever called.

This usage pattern hint us at the real fix: never register a lp callback until a spawn has been initiated with processes capable of connecting through the BTL corresponding to the progress we are registering.

bwbarrett:
George, I'll admit I'm being dense, but I'm not sure how the call pattern changed significantly.  At this point, I'm less worried about this patch and more worried that I'm missing something fundamental...

rhc54:
Just FYI:
```
runtime/opal_progress.c:172:12: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
 static int opal_progress_events()
```

PR #4724, opened 2018-01-17
ompi/oshmem: ucx is selected over yalla/ikrit by default

alex-mikheev:

@yosefe please take a look


yosefe:
@alex-mikheev pls pick to v3.1.x

PR #4728, opened 2018-01-17
OMPI/OSC/UCX: adding atomic lock for fetch_and_op and compare_and_swap

xinzhao3:


hjelmn:
This looks fine to me. It is a shame to loose that optimization though. We should probably figure out a common info key for all osc components to re-enable this optimization. That can wait for a 3.1.1 though.

PR #4735, opened 2018-01-21
debug oshmem failures

ggouaillardet:


ggouaillardet:
the initial issue had nothing to do with UCX nor SHMEM

PR #4736, opened 2018-01-22
oshmem: debug short irecv

ggouaillardet:


ggouaillardet:
@bosilca @yosefe and I reached the conclusion there is nothing wrong here.

when receiving too much data, for example `isend(MPI_INT`) but `irecv(MPI_BYTE)`, the underlying `btl` or `mtl` is free to decide what should be written to the receive buffer (in this case, most would write one byte, but `UCX` is allowed not to write anything)

PR #4716, opened 2018-01-15
Add __optnone__ attribute for Clang

rhc54:
Add the optnone attribute to help avoid optimizing out MPIR_Breakpoint

Signed-off-by: Ralph Castain rhc@open-mpi.org

PR #4715, opened 2018-01-15
Add __optnone__ attribute for Clang

rhc54:
Add the __optnone__ attribute to help avoid optimizing out MPIR_Breakpoint

Signed-off-by: Ralph Castain <rhc@open-mpi.org>

PR #4743, opened 2018-01-23
opal_check_attributes: fix __extension__ test

jsquyres:
Per
https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#index-_005f_005fextension_005f_005f,
use __extension__ in a C statement that will actually verify if the
compiler supports it or not.

### DIFF
--- a/config/opal_check_attributes.m4
+++ b/config/opal_check_attributes.m4
@@ -11,7 +11,7 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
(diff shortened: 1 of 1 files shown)


jsquyres:
Once this is merged in, this commit should be added to #4715 and #4716.

jsquyres:
@bwbarrett I don't see a failure in https://jenkins.open-mpi.org/jenkins/job/open-mpi.pull_request/1981/, but the PRBC CI here at github is claiming failure.  Do you know what happened?  Is it ok to merge?

bwbarrett:
@jsquyres for some reason, the Cray build aborted (not failed as in detected an error in the build, but aborted as in something happened to the builder or the build was killed by Jenkins).  I can't find any reason for the abort, so it was probably builder going offline for some reason.

Looks like it got far enough in the build that this was a success; I'd merge.

PR #4749, opened 2018-01-24
fs/ufs and fs/lustre: remove erroneous return statement

### LINKED ISSUE #4745: Failure using hdf5 with ompio component of v3.x
[iotest.cpp.gz](https://github.com/open-mpi/ompi/files/1658194/iotest.cpp.gz)

## Background information

A failure is observed using hdf5 with the ompio component of openmpi v3.x.  The example works with '--mca io romio314', with mvapich2, and when collective IO is disabled for HDF5.  I am only assuming the bug is with the ompio component.

Latest test with openmpi-v3.0.x-201801180304-86f3448.tar.bz2

Above tarball: 
CC="gcc" CXX="g++" F77="gfortran" FC="gfortran" \
./configure \
        --prefix=${DEST} \
        --enable-mpirun-prefix-by-default \
        --with-pmi=/usr \
        --with-pmi-libdir=/usr/lib/x86_64-linux-gnu \
        --with-verbs=/usr \
        --with-verbs-libdir=/usr/lib

* Operating system/version: 

cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"

uname -a
Linux 4.9.0-4-amd64 #1 SMP Debian 4.9.65-3+deb9u1 (2017-12-23) x86_64 GNU/Linux

gcc --version
gcc (Debian 6.3.0-18) 6.3.0 20170516

* Computer hardware: 

Intel(R) Xeon(R) CPU E5-2670 v2
NFS file system

* Network type: 

Infiniband/Mellanox ConnectX-3

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

The attached program illustrates a simple hdf5 example with the failure.  There should be no output to terminal if working correctly.  With, 'mpirun -np 4 ./iotest', I see:
BAD write!
BAD write!

Using the romio314 component or commenting out USECOLLECTIVE in the code works.

Note: hdf5 1.10.1 was built with the following (szip probably makes no difference):
./configure --enable-parallel --enable-shared --prefix=${HOME}/local --enable-build-mode=debug --with-szlib=/usr/local/szip/2.1.1
make
make install

See top of attached code for remaining build instructions that I used.

edgargabriel:
an erroneous return statement has creeped in commit 1885d99
which leads to some processes not resetting stripe_size
and stripe_count correctly. This can lead in 3.0.x to different
fcoll modules being selected. The impact is not that dramatic on
master and 3.1.x, but could lead to problems as well.

Fixes #4745

### DIFF
--- a/ompi/mca/fs/lustre/fs_lustre_file_open.c
+++ b/ompi/mca/fs/lustre/fs_lustre_file_open.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2008-2017 University of Houston. All rights reserved.
+ * Copyright (c) 2008-2018 University of Houston. All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016-2017 IBM Corporation. All rights reserved.
@@ -192,7 +192,6 @@ mca_fs_lustre_file_open (struct ompi_communicator_t *comm,
                 ret = MPI_ERR_OTHER;
             }
         }
-        return ret;
     }
 
 
--- a/ompi/mca/fs/ufs/fs_ufs_file_open.c
+++ b/ompi/mca/fs/ufs/fs_ufs_file_open.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2008-2017 University of Houston. All rights reserved.
+ * Copyright (c) 2008-2018 University of Houston. All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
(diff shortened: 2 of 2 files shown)


bwbarrett:
@edgargabriel, it looks like this patch doesn't want to merge into master; do you mind updating your local branch?

edgargabriel:
done

edgargabriel:
Not sure what to make of this error message:
open-mpi.build.platforms » os-x appears to be cancelled
open-mpi.build.platforms » os-x completed with result ABORTED

edgargabriel:
bot:retest

bwbarrett:
@edgargabriel: I think one of the OS X builders is tripping offline for some reason, which Jenkins takes as an aborted build.  It looks like everything but the results push succeeded, so I'd push.

edgargabriel:
ok, then I'll go ahead and merge, just to make sure we have one night of MTT results before we transfer this patch to the 3.0.x branch.

PR #4764, opened 2018-01-29
request/state: update state for canceled request

hoopoepg:
- fixed issue in set state for canceled request

fixes https://github.com/openucx/ucx/issues/2223 and https://github.com/openucx/ucx/issues/1928


jsquyres:
ok to test

yosefe:
@hoopoepg pls pick to v3.1.x

PR #4765, opened 2018-01-29
OMPI/OSC/UCX: move memory hooks init in osc to win creation.

xinzhao3:
Move memory hooks init (for request based operation) in osc ucx to window
creation time, to avoid performance issue in MPI initialization.

Signed-off-by: Xin Zhao <xinz@mellanox.com>

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_component.c
+++ b/ompi/mca/osc/ucx/osc_ucx_component.c
@@ -113,63 +113,15 @@ static int progress_callback(void) {
 }
 
 static int component_init(bool enable_progress_threads, bool enable_mpi_threads) {
-    ucp_config_t *config = NULL;
-    ucp_params_t context_params;
-    bool requests_created = false;
     int ret = OMPI_SUCCESS;
-    ucs_status_t status;
 
     mca_osc_ucx_component.ucp_context = NULL;
     mca_osc_ucx_component.ucp_worker = NULL;
     mca_osc_ucx_component.enable_mpi_threads = enable_mpi_threads;
-
-    status = ucp_config_read("MPI", NULL, &config);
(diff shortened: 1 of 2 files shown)


jladd-mlnx:
@xinzhao3 Once this goes into master, please PR into 3.1.x branch.

xinzhao3:
@jladd-mlnx will do

ompi/mca/osc/ucx/osc_ucx_component.c
```diff
@@ -361,6 +350,24 @@ static int component_select(struct ompi_win_t *win, void **base, size_t size, in
         worker_created = true;
     }
 
+    /* init request pool for request based operations */
+    if (mca_osc_ucx_component.requests_initialized == false) {
+        OBJ_CONSTRUCT(&mca_osc_ucx_component.requests, opal_free_list_t);
+        ret = opal_free_list_init (&mca_osc_ucx_component.requests,
```
yosefe: how does it make a difference? memory hooks are initialized during ucp_init()..

jladd-mlnx: Any reason we can't move the UCP context initialization from component open to here on first window create?  That's the intent of the change, but you're right, `ucp_init` is still being called in component open.

xinzhao3: Yossi is right, we need to move ucp_init to window creation. I am working on it now. Will push soon.

jladd-mlnx:
@vspetrov can you check and see if this fixes the issue with mixing Yalla and UCX OSC.

hppritcha:
@xinzhao3 if you want this in 4.0 release, you need to rebase very soon - like this week or next.

PR #4767, opened 2018-01-30
btl/vader: make the backing file job specific

ggouaillardet:
Since open-mpi/ompi@47fd2313abc1cd78f42488ac04384c16ad5703bc
the backing file is now in /dev/shm by default. As a consequence,
the backing file name has to include the jobid so more than one job
can run at a time.


ggouaillardet:
:bot:mellanox:retest

ggouaillardet:
@jsquyres FYI, that could explain the numerous crashes reported by MTT on the master branch
 when `btl/vader` is used and two jobs (two `mpirun` or `MPI_Comm_spawn()`) with two tasks or more each run on the same node at a given time.

jsquyres:
@ggouaillardet I was wondering exactly that question.  Sweet; thanks for following up.

PR #4781, opened 2018-02-01
rmaps: fixed the ordering of `mpirun` target nodes

karasevb:
Fixed the desync of job-nodelists between mpirun and orted
daemons. The issue was observed when using RSH launching because user
can provide arbitrary order of nodes regarding HNP placement.
The mpirun process propagate the daemon's nodelist order to nodes.
The problem was that HNP itself is assembling the nodelist based on
user provided order. As the result ranks assignment was calculated
differently on orted and mpirun.

Consider following example:
* User launches mpirun on node cn2.
* Hostlist is cn1,cn2,cn3,cn4; ppn=1
* mpirun is passing hostlist cn[2:2,1,3-4]@0(4) to orteds
So as result mpirun will assing rank 0 on cn1 while orted will assign
rank 0 on cn2 (because orted sees cn2 as the first element in the node
list)

Signed-off-by: Boris Karasev <karasev.b@gmail.com>

PR #4787, opened 2018-02-05
v3.1: rmaps: fixed the ordering of `mpirun` target nodes

### LINKED ISSUE #4516: Failure while mpirun job depends on the order of the hosts
v3.0.x
master

git clone

```shell
./configure --prefix=`pwd`/install --enable-orterun-prefix-by-default --with-slurm --with-pmi --with-ucx
```

* Operating system/version: 
RedHat 7.2
* Computer hardware: 
Intel dual socket Broadwell
* Network type: 
IB

-----------------------------

## Details of the problem
Running on nodes `node1,node2` works well, but if change the order of the nodes to `node2,node1` this will result to failure:

```shell
ssh node1
mpirun --bind-to core --map-by node -H node2,node1 -np 2  $HPCX_MPI_DIR/tests/osu-micro-benchmarks-5.3.2/osu_allreduce
[node2:13941] Error: pml_yalla.c:95 - recv_ep_address() Failed to receive EP address
It looks like MPI_INIT failed for some reason; your parallel process is
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):

  PML add procs failed
  --> Returned "Not found" (-13) instead of "Success" (0)
```

karasevb:
Fixes #4516 
Corresponds to #4781 and #4785

### DIFF
--- a/orte/mca/rmaps/base/rmaps_base_support_fns.c
+++ b/orte/mca/rmaps/base/rmaps_base_support_fns.c
@@ -12,7 +12,7 @@
  * Copyright (c) 2011      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC.
  *                         All rights reserved.
- * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * Copyright (c) 2018      Mellanox Technologies, Inc.
  *                         All rights reserved.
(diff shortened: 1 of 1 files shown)


karasevb:
@jjhursey 
In fact this issue is not reproduced on v3.0.x. This branch has difference from v3.1.x and master in the follow:
v3.0.x fills the `jdata->map->nodes` here 
https://github.com/open-mpi/ompi/blob/v3.0.x/orte/mca/rmaps/base/rmaps_base_map_job.c#L427 relatively `orte_node_pool` where is the correct ranks order, that avoids the problem.

jjhursey:
@karasevb Thanks for checking!

Issue #4788, opened 2018-02-05
MPI_Comm_spawn_multiple Fortran bindings


ggouaillardet:
The `spawn_argvsnull_mpifh` test from the `ibm` test suite occasionally crashes.

The test program declares
```fortran
    integer, parameter :: count = 2
    character(512) :: commands(2)
```
and then
```fortran
    call mpi_comm_spawn_multiple(count, commands, ...)
```

so far, that is pretty much like the example 10.2 from MPI 3.1 standard at page 387.

The standard says
> MPI_COMM_SPAWN_MULTIPLE(count, array_of_commands, ...)
IN count number of commands (positive integer, significant to MPI only at root | see advice to users)
IN array_of_commands programs to be executed (array of strings, significant only at root)

> Advice to users. The argument count is interpreted by MPI only at the root, as is
array_of_argv. Since the leading dimension of array_of_argv is count, a non-positive
value of count at a non-root node could theoretically cause a runtime bounds check
error, even though array_of_argv should be ignored by the subroutine. If this happens,
you should explicitly supply a reasonable value of count on the non-root nodes. (End
of advice to users.)

```fortran
MPI_COMM_SPAWN_MULTIPLE(COUNT, ARRAY_OF_COMMANDS, ...)
INTEGER COUNT
CHARACTER*(*) ARRAY_OF_COMMANDS(*)
```

note Fortran 2008 bindings are consistent with older Fortran bindings.

Unless I overlooked it, the standard does __not__ states that `array_of_commands` is an array of `count` "strings", but instead
```fortran
CHARACTER*(*) ARRAY_OF_COMMANDS(*)
```

in `ompi_comm_spawn_multiple_f()` we follow the standard literally, and end up with
```c
    ompi_fortran_argv_f2c(array_commands, cmd_string_len,
                          cmd_string_len, &c_array_commands);
```

So in this case,Open MPI requires the test program declares
```fortran
    character(512) :: commands(3)
    commands(3) = ' '
```
(or any string, empty or with white spaces only).
FWIW, we are lucky most of the time, and that likely explains why the issue never popped up before.

To me, there is something fishy here ...

I would have naively expected the standard states

```fortran
CHARACTER*(*) ARRAY_OF_COMMANDS(COUNT)
```
and we handle `array_of_commands` accordingly in Open MPI.

There might be a rationale for the current standard, but
 - I am surprised it is not described
 - what should happen if `array_of_commands` has less or more that `count` commands ?
[...]

PR #4789, opened 2018-02-05
fortran: fix MPI_Comm_spawn_multiple() parameter handling.

ggouaillardet:
internally handle 'array_of_commands' as an array with size 'count'

Refs open-mpi/ompi#4788


ggouaillardet:
@jsquyres the initial feedback from the Fortran WG is that `array_of_commands` is indeed an array of `count` _strings_.
Shall i merge this PR (and PR to all the release branches) now ?
If the prototype is updated in the upcoming MPI 3.2 standard, I will issue an other PR for that.

ggouaillardet:
replaced by #5016

PR #4796, opened 2018-02-07
opal/btl: add support for flushing RDMA/atomic operations

hjelmn:
This commit adds a new optional function to the BTL module:
btl_flush. This function takes an optional BTL endpoint. When called
this function completes all outstanding RDMA and atomic operations
started prior to the call to btl_flush.

### DIFF
--- a/opal/mca/btl/base/btl_base_frame.c
+++ b/opal/mca/btl/base/btl_base_frame.c
@@ -14,7 +14,7 @@
  * Copyright (c) 2008-2013 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2016-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
@@ -53,6 +53,7 @@ mca_base_var_enum_value_flag_t mca_btl_base_flag_enum_flags[] = {
     {MCA_BTL_FLAGS_NEED_ACK, "need-ack", 0},
     {MCA_BTL_FLAGS_NEED_CSUM, "need-csum", 0},
     {MCA_BTL_FLAGS_HETEROGENEOUS_RDMA, "hetero-rdma", 0},
+    {MCA_BTL_FLAGS_RDMA_FLUSH, "rdma-flush", 0},
     {0, NULL, 0}
 };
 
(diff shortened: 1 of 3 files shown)


hjelmn:
This PR is in preparation for a UCX RDMA-only BTL. This provides functionality that can improve the performance of MPI one-sided.

jsquyres:
@hjelmn What's the purpose of a UCX RDMA-only BTL?  Will that subvert the UCX PML -- e.g., move a user from the UCX PML to OB1?  Or will this UCX BTL only be used for OSC?

hjelmn:
Only for osc. Its an alternate path to osc/ucx.

hjelmn:
Since UCX provides matching I see no reason to add send/recv support at this time.

hjelmn:
I should also mention that there is a benefit to btl_flush for other btls as well. With btl/ugni I see a 4x improvement of small message injection with the RMA-MT benchmarks.

miked-mellanox:
can you please attach comparison with osu atomics between atomics/ucx + pml/ucx and osc/atomics + btl/ucx to clarify performance and duplication concerns?

hjelmn:
No such things were required for osc/ucx which duplicates functionality of osc/rdma. I will be the owner of btl/ucx anyway.

miked-mellanox:
atomics/ucx is not duplicate for osc/rdma, also improves performance for osc/rdma. Adding btl/ucx duplicates ucx logic already existing in ompi and unclear what performance impact it brings (comparing to atomics/ucx counterpart).
since ob1/btl is out-of-the-box default - it would be essential to know the performance impact on other ucx citizens in ompi. (new ucx instance, new progress thread, etc...)

hjelmn:
pml/ob1 will not use an rdma-only btl. They are ignored :).

hjelmn:
Anyway, we can have that discussion with the btl/ucx PR. This PR is just to add a function that can improve osc/rdma's performance even with existing btls. I mention btl/ucx because this is necessary to get peak performance there since ucx doesn't have direct support for tracking individual rdma requests.

miked-mellanox:
will it call btl/ucx directly? if so - the point is the same - impact on other ucx citizens when ucx is compiled and used from pml/atomics layers

hjelmn:
I don't quite follow. The pml won't use it as it is RDMA only (no send/recv). The atomics layer doesn't support btls. The only difference is there will be an additional ucx worker iff osc/rdma is in use.

hjelmn:
I think there is a fundamental lack of understanding of the Open MPI infrastructure WRT to btls. As I said we can discuss when the PR for the btl comes up. For now, this is irrelevant to ucx.

jsquyres:
Agreed.  Sorry for sidetracking this PR -- this PR is just about adding the RDMA flush on a BTL.

hjelmn:
I need to add a to the function documentation then I will commit this. The btl_flush function only guarantees that all previous put, get, and amos are complete. It does not guarantee that all the callbacks have been made.

PR #4798, opened 2018-02-07
pml/ob1: ignore the eager limit of RDMA-only btls

hjelmn:
This commit fixes a flaw in the eager limit check in pml/ob1. The
check was incorrectly checking if RDMA-only BTLs (BTLs without the
send flag) has a valid eager limit. This commit fixes the check by
adding an additional check for the send flag on the BTL module.


hjelmn:
@bwbarrett It says failed but I don't see a failure on Jenkins. What is going on?

hppritcha:
bot:ompi:retest

Issue #4806, opened 2018-02-13
Relative path in output redirection


devreal:
## Background information

I am using the `--output-filename` parameter to redirect the output of my application processes into individual files in a relative output directory. The working directory is a subdirectory of my home directory mounted via NFS. Output redirection seems to work correctly for all processes running on the first node (the node that runs rank 0) while all other processes create output directories in $HOME (instead of the subdirectory).

### What version of Open MPI are you using?
OpenMPI 3.0.0 installed from release tarball (also reproducible with 2.1.1)

-----------------------------

## How to reproduce:
```
# two procs, one per node
~/test $ mpirun -n 2 -N 1 --output-filename output.log ls
output.log
output.log
~/test $ ls output.log/*
rank.0

# two procs, single node
~/test $ mpirun -n 2 -N 2 --output-filename output.log ls
output.log
output.log
~/test $ ls output.log/*
rank.0  rank.1
```
Notice that in the first case the output from rank 1 is missing in the folder `output.log`.

I should note that I sometimes observe the following error when running the application directly in the $HOME directory: 
```
$ mpirun -n 2 -N 1 --output-filename output.log ls
[n121301:25598] [[60171,0],1] ORTE_ERROR_LOG: File open failure in file
/path/to/mpi/openmpi/3.0.0-intel-17.0.4/openmpi-3.0.0/orte/mca/iof/base/iof_
base_setup.c at line 314
[n121301:25598] [[60171,0],1] ORTE_ERROR_LOG: File open failure in file
/path/to/mpi/openmpi/3.0.0-intel-17.0.4/openmpi-3.0.0/orte/mca/iof/orted/iof
_orted.c at line 184
[n121301:25598] [[60171,0],1] ORTE_ERROR_LOG: File open failure in file
/path/to/mpi/openmpi/3.0.0-intel-17.0.4/openmpi-3.0.0/orte/mca/iof/base/iof_
base_setup.c at line 237
[n121301:25598] [[60171,0],1] ORTE_ERROR_LOG: File open failure in file
/path/to/mpi/openmpi/3.0.0-intel-17.0.4/openmpi-3.0.0/orte/mca/odls/base/odl
s_base_default_fns.c at line 1147
```

A workaround is to specify an absolute path to `--output-filename`.

hppritcha:
forgot to close when #4809 was merged

PR #4809, opened 2018-02-13
 Ensure that output-filename is passed as an absolute path

rhc54:


rhc54:
Fixes #4806

orte/orted/orted_submit.c
```diff
@@ -813,7 +813,19 @@ int orte_submit_job(char *argv[], int *index,
     }
     /* if we were asked to output to files, pass it along */
     if (NULL != orte_cmd_options.output_filename) {
-        orte_set_attribute(&jdata->attributes, ORTE_JOB_OUTPUT_TO_FILE, ORTE_ATTR_GLOBAL, orte_cmd_options.output_filename, OPAL_STRING);
+        /* if the given filename isn't an absolute path, then
+         * convert it to one so the name will be relative to
+         * the directory where prun was given as that is what
+         * the user will have seen */
```
jsquyres: Does this assume that the filesystem is the same on all remote nodes?  If so, do we want to make that assumption?

rhc54: If you don't, then the file is going to go somewhere very unexpected - as reported by the user. It will wind up relative to wherever the orted defaults to going, which will vary by system/launcher.

This way, it goes relative to where the user types "mpirun", which is what the user sees and knows.

jsquyres: Just to clarify: is it relative to the PWD of `mpirun`, or is it relative to the app working directory on each node?

jsquyres: @rhc54 and I talked on the phone: we rationalized that this solution is "good enough" -- we're just documenting in the mpirun help output and the man page exactly what it does.

ggouaillardet:
@rhc54 do we also need to something in `plm/base` ?

i initially crafted the inline patch below, but now I am wondering
 - is this needed ?
 - should we simply drop passing `-mca orte_output_filename ...` to `orted` ?
 - should we only interpret this to fail nicely if the directory does not exist ?

```patch
diff --git a/orte/mca/plm/base/plm_base_launch_support.c b/orte/mca/plm/base/plm_base_launch_support.c
index 159c005..8fcd6fd 100644
--- a/orte/mca/plm/base/plm_base_launch_support.c
+++ b/orte/mca/plm/base/plm_base_launch_support.c
@@ -14,7 +14,7 @@
  *                         et Automatique. All rights reserved.
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC.
  * Copyright (c) 2013-2018 Intel, Inc.  All rights reserved.
- * Copyright (c) 2014-2017 Research Organization for Information Science
+ * Copyright (c) 2014-2018 Research Organization for Information Science
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
@@ -39,6 +39,8 @@
 #include "opal/hash_string.h"
 #include "opal/util/argv.h"
 #include "opal/util/opal_environ.h"
+#include "opal/util/os_path.h"
+#include "opal/util/path.h"
 #include "opal/class/opal_pointer_array.h"
 #include "opal/dss/dss.h"
 #include "opal/mca/hwloc/hwloc-internal.h"
@@ -1534,9 +1536,16 @@ int orte_plm_base_orted_append_basic_args(int *argc, char ***argv,

     /* if output-filename was specified, pass that along */
     if (NULL != orte_output_filename) {
+        char *path, cwd[OPAL_PATH_MAX];
         opal_argv_append(argc, argv, "-"OPAL_MCA_CMD_LINE_ID);
         opal_argv_append(argc, argv, "orte_output_filename");
-        opal_argv_append(argc, argv, orte_output_filename);
+        if (opal_path_is_absolute(orte_output_filename) || NULL == getcwd(cwd, OPAL_PATH_MAX)) {
+            opal_argv_append(argc, argv, orte_output_filename);
+        } else {
+            path = opal_os_path( false, cwd, orte_output_filename, NULL );
+            opal_argv_append(argc, argv, path);
+            free(path);
+        }
     }

     /* if --xterm was specified, pass that along */
```

rhc54:
It shouldn't be needed - mpirun set the attribute in the job data object, and that gets forwarded to the orteds. There is no need to put output filename on the cmd line, and that piece of code in the plm/base should be removed.

ggouaillardet:
ok, I will fix that tomorrow and update the existing PRs for the release branches.

FWIW, `orted` uses this in `orte_iof_base_open` and I am not sure of how things will end up if the absolute path cannot be created on the remote nodes (quite unlikely imho though)

rhc54:
You'll just get the same error originally reported by the user: "File cannot be created"

rhc54:
In fact, if you look in iof_base_setup.c, you'll see that we create the dirpath for the given output file (as provided by the attribute) on line 329 - so the code in iof_base_open should also be removed as it is duplicative (and stale, for that matter).

ggouaillardet:
I initially thought we did that earlier in order to report an error at the very early stage when needed.

Thanks for the explanation, I will do that too.

rhc54:
Yeah, I think you are correct, but it doesn't work any more since we moved to attributes. Reason for move was the DVM - we needed to get away from setting output_filename as a global setting and allow it instead to be set per-job. So we can now only report the error when a specific job attempts to setup the path.

PR #4818, opened 2018-02-15
Update check package

rhc54:
When checking headers, always check the given location first before appending "/include".

When checking libs and a location was not specified, then start by always looking at the standard locations without first modifying LDFLAGS. This avoids incorrectly adding an unnecessary -L to the LDFLAGS.

Remove any extra slashes in libevent path

Clear the ac_check_headers cache

### DIFF
--- a/config/opal_check_package.m4
+++ b/config/opal_check_package.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2012-2017 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
-dnl Copyright (c) 2014      Intel, Inc. All rights reserved.
+dnl Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
 dnl Copyright (c) 2015-2016 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl $COPYRIGHT$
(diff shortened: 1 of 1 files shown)


ibm-ompi:

 Gist: https://gist.github.com/95ebe885fef35ef217357fa0ffdacf62

ibm-ompi:

 Gist: https://gist.github.com/8b5af16318d19ebc5b7fb3993f83aae2

ibm-ompi:

 Gist: https://gist.github.com/2677bad77b5e953f8df2c79d575abb99

rhc54:
Refs #4722 

@ggouaillardet @jsquyres Posting this as a result of having to fix check_package in PMIx - we were consistently hitting use-cases where it failed. Specifically, a particular software vendor consistently releases software that installs in the following pattern:

* the include file goes into /usr/include/foo/foo.h
* the library goes into /usr/lib/libfoo.so

In order to find the header, you must configure ```--with-foo=/usr/include/foo```, but check package fails because it _always_ adds an ```/include``` to whatever you pass. I had to fix that first.

Second problem was that it immediately fails to find the lib because the library checker _always_ prepends the ```---with-foo``` value to the libdir. So I had to fix that too.

You are welcome to use or ignore.

rhc54:
Ah, I should note an error in the commit message. I did not backport the fix to the libevent configure logic - if the user ends their ```--with-libevent``` with a ```/```, we fail to remove it later because it gets embedded in the libevent prefix variables by the libevent configure code prior to being passed to check_package. I had to fix that in PMIx as well.

rhc54:
I'll take a look - sounds more like a simple bug. The goal is to resolve the specific problem I'm up against, not the more generalized issue in #4722. This problem should be resolvable by simply getting check_package to do the right thing.

PR #4820, opened 2018-02-15
osc/rdma: remove erroneous check

hjelmn:
This was caught by some users at LANL. It was fixed on master as part
of 7a83fdb9bb3aa13aad10a0f5a5695942a37c8bb1.


hppritcha:
It doesn't look like the correct commit is referenced in the commit message above.

hppritcha:
Sorry take that back, but why in the world was this file touched as part of an hwloc update?

hppritcha:
@jsquyres let's merge this in and do an rc2.  Not clear that "LANL users" was actually anything I'd call user reported, so no NEWS update.

PR #4822, opened 2018-02-16
configury: define C11 macros once

ggouaillardet:
Revamp OPAL_PROG_CC_C11 macro in order to define macros only once.
Otherwise, macros get redefined during the configure process and
issue a bunch of warning in config.log

### DIFF
--- a/config/opal_setup_cc.m4
+++ b/config/opal_setup_cc.m4
@@ -14,8 +14,8 @@ dnl Copyright (c) 2007-2009 Sun Microsystems, Inc.  All rights reserved.
 dnl Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2012-2017 Los Alamos National Security, LLC. All rights
 dnl                         reserved.
-dnl Copyright (c) 2015      Research Organization for Information Science
-dnl                         and Technology (RIST). All rights reserved.
+dnl Copyright (c) 2015-2018 Research Organization for Information Science
+dnl                         and Technology (RIST).  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


hjelmn:
Looks good to me.

config/opal_setup_cc.m4
```diff
@@ -47,30 +41,42 @@ AC_DEFUN([OPAL_PROG_CC_C11_HELPER],[
...
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__ATOMIC], [$opal_prog_cc_c11_helper__Atomic_available],
+                              [Whether C compiler supports __Atomic keyword])
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__GENERIC], [$opal_prog_cc_c11_helper__Generic_available],
+                              [Whether C compiler supports __Generic keyword])
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__STATIC_ASSERT], [$opal_prog_cc_c11_helper__static_assert_available],
+                              [Whether C compiler support _Static_assert keyword])])
+
```
jsquyres: I'm not sure I understand this block -- don't we *always* want to AC_DEFINE this values (and make sure that they are 0 or 1)?

hjelmn: Hmm, indeed. We do. I haven't tested this with a non-C11 compiler. @ggouaillardet Might be worth doing. Anything older than gcc 4.9.0 will work. I have 4.8.x installed specifically to test with.

ggouaillardet: i run a CentOS 7 box with gcc 4.8.x, which is a non-C11 compiler.

In this environment and on `master`, `OPAL_PROG_CC_C11_HELPER` is called several times in `OPAL_PROG_CC_C11` (test the C compiler without flag and then with all flags since all these test fail). Finally, `OPAL_PROG_CC_C11_HELPER` is invoked a last time in `OPAL_SETUP_CC` since `$opal_cv_c11_supported = no`.

With a C11 compiler, then `OPAL_PROG_CC_C11_HELPER` is invoked until it success.

I did not change this behavior in this PR.
In order to __always__ `AC_DEFINE` the macros, I added a forth argument to force the macros to be `AC_DEFINE`'ed, and this argument is only `yes` in `OPAL_SETUP_CC` since we know at this stage we do not have a C11 compiler.

Bottom line, I think this PR is correct.
And that being said, we might want to take this opportunity to revisit the existing logic.
Feel free to suggest anything and I will be happy to update this PR.

benmenadue:
@ggouaillardet Okay, thanks, it's autogen-ing at the moment. I also just pushed my own similar change into my fork ([benmenadue:rework-c11-detection](/benmenadue/ompi/tree/rework-c11-detection)), which only tests for the features once it works out if and how C11 is supported. Couldn't PR it because I forgot to sign off the commits. But if your PR works, then it's all good 👍  .

benmenadue:
@ggouaillardet `configure` finished fine, and I only got one copy of those macros :+1:

config/opal_setup_cc.m4
```diff
@@ -47,29 +45,42 @@ AC_DEFUN([OPAL_PROG_CC_C11_HELPER],[
...
+                              [Whether C compiler support atomic convenience variables in stdatomic.h])
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__ATOMIC], [$opal_prog_cc_c11_helper__Atomic_available],
+                              [Whether C compiler supports __Atomic keyword])
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__GENERIC], [$opal_prog_cc_c11_helper__Generic_available],
+                              [Whether C compiler supports __Generic keyword])
+
+           AC_DEFINE_UNQUOTED([OPAL_C_HAVE__STATIC_ASSERT], [$opal_prog_cc_c11_helper__static_assert_available],
+                              [Whether C compiler support _Static_assert keyword])])
```
jsquyres: Why not move these AC_DEFINEs down into OPAL_SETUP_CC?  That way, you don't need a 4th argument and a bogus extra call into OPAL_PROG_CC_C11_HELPER.

ggouaillardet:
@jsquyres I made the changes (and I should have done this in the first place ...)

can you please double check them ?

config/opal_setup_cc.m4
```diff
@@ -157,9 +151,26 @@ AC_DEFUN([OPAL_SETUP_CC],[
 
         # Get the correct result for C11 support flags now that the compiler flags have
         # changed
-        OPAL_PROG_CC_C11_HELPER([],[],[])
+        OPAL_PROG_CC_C11_HELPER([], [], [])
```
jsquyres: Do you know what this is for?  We only get here if we have already determined that there is no C11 support, so we check for C99 support.  ...but then we check for C11 support again, with no CLI options.

Are we just checking to see if the C99 flags also enable *some* C11 functionality?  If so, the comment above this line is misleading, at best.

ggouaillardet: My guess is we want to make sure all C11 macros are set to zero.

@hjelmn can you please comment on that ?

jsquyres:
bot:ompi:retest

config/opal_setup_cc.m4
hjelmn: That check is there because the C99 support check may add additional flags (for example ```--std=gnu99```) which may change the results. This is entirely defensive. Example (have not seen this happen), without a flag: _Generic = 1, with ```--std=gn99``` _Generic = 0.

hjelmn: It may never happen but that was the original reason. The line dates to the earliest versions of the C11 checks.

thananon:
Seems like I broke mergability with #5272 . Sorry.

ggouaillardet:
No problem, I will fix that later today

ibm-ompi:

 Gist: https://gist.github.com/23b302387a2a091a4836f4f60b699dce

ibm-ompi:

 Gist: https://gist.github.com/ea85c619ac6e84e4940b2d77f0de89c8

ibm-ompi:

 Gist: https://gist.github.com/c1e83ed663612112f4f1437f4725f5d2

ibm-ompi:

 Gist: https://gist.github.com/ac59e4dadff1a4c942a828a7ac55696a

ibm-ompi:

 Gist: https://gist.github.com/804bd915ea7a9ca464fb717ee68e8aca

PR #4825, opened 2018-02-16
btl/usnic: update BTL_VERSION handling

jsquyres:
Follow-on to 8097d09858: now that BTL_VERSION is defined in btl.h, be
a little smarter about whether we define it or not.


opal/mca/btl/usnic/btl_usnic_compat.h
```diff
@@ -335,9 +343,7 @@ opal_btl_usnic_put(
 /* BTL 3.0 (i.e., >=v1.9, but listed separately because these are
    really BTL API issues) */
 
-#elif BTL_VERSION == 310
-
-#include "opal/mca/btl/btl.h"
+#elif BTL_VERSION > 300
```
hjelmn: Should this be >=?

jsquyres:
opensuse_leap_42 test looks like a transient curl / web error.

bot:ompi:retest

PR #4827, opened 2018-02-17
mca/base: add support for long variable values

hjelmn:

### DIFF
--- a/ompi/mpi/tool/mpit_common.c
+++ b/ompi/mpi/tool/mpit_common.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2012-2015 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2012-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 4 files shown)


opal/mca/base/mca_base_var.c
```diff
@@ -100,7 +101,8 @@ const size_t ompi_var_type_sizes[] = {
     sizeof (char),
     sizeof (char),
     sizeof (bool),
-    sizeof (double)
+    sizeof (double),
+    sizeof (long),
```
jsquyres: Please remove the trailing comma -- it generates compiler warnings.

hjelmn: Odd that is complains about that. The trailing comma is allowed by C99 and it is required that the compiler support it. Do you have the warning message handy?

hjelmn: See C99 6.7.8.

hjelmn:
My use case is I have a variable that needs more than 32-bits but also needs -1.

rhc54:
out of curiosity - why not just go with int64?

rhc54:
I ask because of the indeterminate size of long:
```
Long signed integer type. Capable of containing at least the [−2,147,483,647, +2,147,483,647] range;[3][4] thus, it is at least 32 bits in size.
```
If you are on a system that uses the minimum size, then ```long``` won't meet your described use case.

opal/mca/base/mca_base_var.c
jsquyres: Ah, if this is allowable via C99, then I retract my objection.

hjelmn:
I am fine with int64_t. Long just fit into the current type list. I can update this to add all the fixed width types.

opal/mca/base/mca_base_var.c
hjelmn: I will still look into the message as I am curious what is going on. We use trailing commas all over the place.

rhc54:
fwiw: I avoid leaving trailing commas in the code base because of variation across compilers - some don't bark, others definitely do when using pedantic flags. And yes, I know that isn't strictly what C99 saids - but we all know how closely every compiler adheres to every little detail in that document. Seemed like a trivial thing to simply avoid.

opal/mca/base/mca_base_var.c
jsquyres: Don't spend any time on it.  I am probably remembering something with an old compiler and/or before we enforced C99-specific compiler options.

jsquyres:
If Ralph is seeing those warnings recently, then I would agree with what he said.  I know I have seen these messages before, so I, too, have simply avoided trailing commas.  But I don't have any specific data points about which compiler / platform / etc.

hjelmn:
You know my personal opinion on poorly behaved compilers so I won't repeat it here. I will say I never avoid them but I have also not run into a compiler that c complains about trailing commas.

rhc54:
I wouldn't worry about it - if we see them when building, we can simply submit a PR to remove the comma. I've done similar minor fixes in the past as we all use different compilers, each with their own quirks.

hjelmn:
Added fixed width integer types (both signed and unsigned).

ibm-ompi:

 Gist: https://gist.github.com/24bd28ed4fe46bcc54b44293ef17b6a1

hjelmn:
@jsquyres This look ok to you? I simplified the MCA type -> MPI type code to make it easier to add new types. I doubt we will need to after this PR.

ompi/mpi/tool/mpit_common.c
```diff
@@ -28,53 +28,41 @@ void ompi_mpit_unlock (void)
...
+    [MCA_BASE_VAR_TYPE_UNSIGNED_LONG] = MPI_UNSIGNED_LONG,
+    [MCA_BASE_VAR_TYPE_UNSIGNED_LONG_LONG] = MPI_UNSIGNED_LONG_LONG,
+
+#if SIZEOF_SIZE_T == SIZEOF_UNSIGNED_INT
+    [MCA_BASE_VAR_TYPE_SIZE_T] = MPI_UNSIGNED,
+#elif SIZEOF_SIZE_T == SIZEOF_UNSIGNED_LONG
+    [MCA_BASE_VAR_TYPE_SIZE_T] = MPI_UNSIGNED_LONG,
+#elif SIZEOF_SIZE_T == SIZEOF_LONG_LONG
+    [MCA_BASE_VAR_TYPE_SIZE_T] = MPI_UNSIGNED_LONG_LONG,
+#else
```
rhc54: @hjelmn do you want to add the 64-bit and 32-bit fixed size options here?

rhc54: Other than that possible change, looks good. Thanks!

Issue #4829, opened 2018-02-17
issues with a very large derived data type


rabauke:
Using OpenMPI 2.1.1 as distributed by Ubuntu 17.10 x86_64 I get the following error message when dealing with a very large derived data type:

`Read 2147479552, expected 2147483748, errno = 38`

Below, I attached a small program to reproduce the issue. I build a derived vector-like data type with possibly more than `INT_MAX` elements using `MPI_Type_struct`.

```C
#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <stddef.h>
#include <limits.h>
#include "mpi.h"

// build a vector-like data type with more than INT_MAX elements
MPI_Datatype build(size_t count,  MPI_Datatype old_type) {
  MPI_Datatype new_type;
  if (count<=INT_MAX) {
    MPI_Type_contiguous((int)count, old_type, &new_type);
  } else {
    const size_t modulus=INT_MAX;
    const size_t count1=count/modulus;
    const size_t count0=count-count1*modulus;
    MPI_Count lb, extent;
    MPI_Type_get_extent_x(old_type, &lb, &extent);
    MPI_Datatype type_modulus;
    MPI_Type_contiguous((int)modulus, old_type, &type_modulus);
    MPI_Type_commit(&type_modulus);
    int block_lengths[2]={ (int)count0, (int)count1 };
    MPI_Aint displacements[2]={ 0, (MPI_Aint)(count0*extent) };
    MPI_Datatype types[2]={ old_type, type_modulus };
    MPI_Type_create_struct(2, block_lengths, displacements, types, &new_type);
    MPI_Type_commit(&new_type);
    MPI_Type_free(&type_modulus);
  }
  return new_type;
}

int main(int argc, char *argv[]) {
  MPI_Init(&argc, &argv);
  int C_size, C_rank;
  MPI_Comm_size(MPI_COMM_WORLD, &C_size);
  MPI_Comm_rank(MPI_COMM_WORLD, &C_rank);
  if (C_size!=2)
    MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
  const size_t size=(1ull<<31) + 100;
  char *p=malloc(size);
  MPI_Datatype vector_type=build(size, MPI_CHAR);
  if (C_rank==0) {
    for (size_t i=0; i<size; ++i)
      p[i]=i;
    MPI_Send(p, 1, vector_type, 1, 0, MPI_COMM_WORLD);
  } else {
[...]

PR #4832, opened 2018-02-19
btl/vader: handle unexpected short read/write in process_vm_{read,write}v

ggouaillardet:
Thanks Heiko Bauke for the bug report.

Refs. open-mpi/ompi#4829

### DIFF
--- a/opal/mca/btl/vader/btl_vader_get.c
+++ b/opal/mca/btl/vader/btl_vader_get.c
@@ -2,6 +2,8 @@
 /*
  * Copyright (c) 2010-2014 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2018      Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -22,6 +24,7 @@
 #if OPAL_CMA_NEED_SYSCALL_DEFS
 #include "opal/sys/cma.h"
 #endif /* OPAL_CMA_NEED_SYSCALL_DEFS */
+
 
 #endif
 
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@bosilca I also read

> (Partial transfers apply at the granularity of iovec elements.  These system calls won't perform a partial transfer that splits a single iovec element.)

Since our `iovec` has only one element, does this behavior points to a bug in `glibc`/kernel ?
(partial transfer is not permitted, and hence the returned value should be `0` or the full buffer)

bosilca:
I concur. I run on 2 kernel versions and it is the same behavior, and I did not had the time to look further.

jsquyres:
If this behavior is contrary to system man pages, I think you should definitely add a comment in both locations explaining why this code is necessary (and cite the specific system/OS/distro/version/whatever where the non-manpage-conforming behavior was seen).  This won't matter in the next few months, but 2 years from now, someone will look at this code and say "That code is unnecessary because the man page guarantees that the unit of granularity of transfer is a single iovec!" -- opening the door to a possible regression if they delete this extra logic.

jsquyres:
Per 2018-02-20 webex:

1. We agree that the comments should be added to this PR (as cited above).
1. We should probably investigate why the man page digresses from the observed behavior.  E.g., is this a glibc issue?  Is it a kernel issue?  Is this a *regression* (e.g., it *used* to obey iovec boundaries)?  ...etc.
1. This is an important issue and we should fix it, but given that it has existed in all of our release branches for years and this is the first time we're hearing about it (@bosilca says it only happens when you try to send >2GB in a single message), it doesn't feel like a blocker for any just-about-to-be-released version.

hjelmn:
@ggouaillardet Can you add the comments that @jsquyres mentioned. I would like to get this into master.

ggouaillardet:
@hjelmn I made the changes and will merge when CI completes

PR #4837, opened 2018-02-19
ompi: ompi_mpi_init(): do not export threading level to modex.

artpol84:
For some of our configuration this flag increases per-process contribution
by ~20% while it is not being used currently.

The consumer of this flag was communicator ID calculation logic, but it was
changed in 0bf06de3f1444f469303e47752430ec9b423b33f.

(cherry picked from commit b601dd504af3ff97e37781056bc4ce05e7d42260)

### DIFF
--- a/ompi/runtime/ompi_mpi_init.c
+++ b/ompi/runtime/ompi_mpi_init.c
@@ -383,10 +383,6 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided)
 
     OMPI_TIMING_INIT(32);
 
-    /* bitflag of the thread level support provided. To be used
-     * for the modex in order to work in heterogeneous environments. */
-    uint8_t threadlevel_bf;
-
     ompi_hook_base_mpi_init_top(argc, argv, requested, provided);
 
     /* Ensure that we were not already initialized or finalized.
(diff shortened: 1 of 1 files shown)


artpol84:
@rhc54 unfortunately I don't have a spare cycle to investigate that. And I don't want to change it blindly. 
With this thing we are getting some improvement because modex size is reduced at least by 20B.
With env - I don't think we are getting any perf improvement.
I think it should be a separate PR.

rhc54:
@bwbarrett @jjhursey I'll have to defer to someone else to investigate, then - I don't know if somehow these things are tied together and your tests simply aren't tripping over it. Could be someone out there has an idea?

artpol84:
@rhc54 if you can't find modex recv, I think this means that presence of this information in the modex doesn't play a role. Isn't it?

bwbarrett:
bot:ompi:retest

For some reason, the OMPI tests tried to pull a commit that doesn't exist.  Very weird.

artpol84:
I mean that rank don't need to care about someone else's threading level, but it still can be concerned about its own threading level which might be exported through env.
So my believe that those things are unrelated. If you think that they are - I can double check that as it is related to this PR.

rhc54:
@bosilca Appreciate the explanation - I couldn't see where it was being set, nor if it was somehow related to the exchange. Sounds like it is unrelated.

artpol84:
@bosilca indeed thanks for the explanation!

PR #4838, opened 2018-02-20
dist: Update .so version numbers for v3.1.0

bwbarrett:
Update the shared library version numbers for v3.1.0.  As a
ABI compatible version with the v3.0.x series, follow our
convention of bumping the current number by 10 and setting the
age to v3.0.0 age + 10.  There is one exception to that rule,
as libmca_ompi_common_ompio was changed in an incompatible
way in v3.0.1 (v3.1.x includes that change), so we are only
backwards compatible to that version, not all the way to
v3.0.0.  Since ompi_common isn't part of the stable ABI and
users would have to do something really weird to even hit
that case, this isn't too big of a deal.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -84,17 +84,17 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=0:0:0
-libmpi_cxx_so_version=0:0:0
-libmpi_mpifh_so_version=0:0:0
-libmpi_usempi_tkr_so_version=0:0:0
-libmpi_usempi_ignore_tkr_so_version=0:0:0
-libmpi_usempif08_so_version=0:0:0
-libopen_rte_so_version=0:0:0
-libopen_pal_so_version=0:0:0
-libmpi_java_so_version=0:0:0
-liboshmem_so_version=0:0:0
-libompitrace_so_version=0:0:0
+libmpi_so_version=50:0:10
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -84,32 +84,32 @@ date="Unreleased developer copy"
...
+libompitrace_so_version=50:0:10
 
 # "Common" components install standalone libraries that are run-time
 # linked by one or more components.  So they need to be versioned as
 # well.  Yuck; this somewhat breaks the
 # components-don't-affect-the-build-system abstraction.
 
 # OMPI layer
-libmca_ompi_common_ompio_so_version=0:0:0
+libmca_ompi_common_ompio_so_version=50:0:9
```
jsquyres: @bwbarrett Why on earth did we make OMPI's C value be 41 in v3.0.0?  😱 (this is just a rhetorical question)

Also, I notice that `ompi/mca/common/monitoring` has no version number here.  Indeed, it doesn't look like it is versioned at all, which seems problematic.

This PR should still go through, but I think we need another PR that adds a version number for the common monitoring library.

bwbarrett: We didn't; see the commit comment.  There was a backwards-incompatible change in common_ompio in 3.0.1.  3.1 includes that change, so it should only be backwards compatible to 3.0.1, not 3.0.0 (well, the shared libraries built by those versions, but you know what I mean).

Good catch on the monitoring change; thanks for making that.

PR #4842, opened 2018-02-20
ompi/monitoring: add .sh versionig to common monitoring lib

jsquyres:

Refs #4838 

FYI: @bosilca @clementFoyer 

This should be PR'ed over to v3.1.x as well.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -103,6 +103,7 @@ libompitrace_so_version=0:0:0
 
 # OMPI layer
 libmca_ompi_common_ompio_so_version=0:0:0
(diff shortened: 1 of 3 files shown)


bosilca:
I don't see how this version is updated. Theoretically the monitoring library should be in sync with the MPI library version.

jsquyres:
@bosilca It gets updated when the top-level `VERSION` file is updated (e.g., in #4838).

bosilca:
How do we ensure it is always in sync with the MPI library version ?

PR #4853, opened 2018-02-23
configury: fix PMI detection

ggouaillardet:

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -14,8 +14,8 @@
 # Copyright (c) 2011-2014 Los Alamos National Security, LLC. All rights
 #                         reserved.
 # Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
-# Copyright (c) 2014-2017 Research Organization for Information Science
-#                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2014-2018 Research Organization for Information Science
+#                         and Technology (RIST).  All rights reserved.
 # Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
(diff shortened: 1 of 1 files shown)


config/opal_check_pmi.m4
```diff
@@ -39,18 +39,30 @@ AC_DEFUN([OPAL_CHECK_PMI_LIB],
...
-    AC_MSG_CHECKING([for $3.h in $1/include])
-    AS_IF([test -f $1/include/$3.h],
-          [AC_MSG_RESULT([found])
-           opal_check_$3_mycppflags="-I$1/include"],
-          [AC_MSG_RESULT([not found])
-           AC_MSG_CHECKING([for $3.h in $1/include/slurm])
-           AS_IF([test -f $1/include/slurm/$3.h],
+    AS_IF([test -n "$1"],
+          [AC_MSG_CHECKING([for $3.h in $1/include])
+           AS_IF([test -f $1/include/$3.h],
```
rhc54: @ggouaillardet Not sure how important this is, but the boost folks had cited that we should use ```-e``` instead of ```-f``` as the file might be a symlink, and that would yield a false error with ```-f```

ggouaillardet:
@rhc54 I already pushed the requested changes, then I had some second thoughts and ran a few test
```
$ touch file
$ mkdir dir
$ ln -s file lfile
$ ln -s dir ldir
$ test -f file && echo yes
yes
$ test -f lfile && echo yes
yes
$ test -f dir && echo yes 
$ test -f ldir && echo yes
$ test -e file && echo yes
yes
$ test -e lfile && echo yes
yes
$ test -e dir && echo yes 
yes
$ test -e ldir && echo yes
yes
```

It appears `test -f` does follow the symbolic link, and `test -e` is successful when invoked against a (symlink to a) directory.


So it seems `test -f` is indeed what we want to use here.

from the man page
>  -f file       True if file exists and is a regular file.

it does not explicitly mention symbolic link, and fwiw 
 - the observed behavior is intuitive for me.
 - if `test -f lfile` returned true, it would be a mess to test whether `lfile` is either a regular file or a symlink to a regular file
 - `test -f file && ! test -L file` can be used to check `file` is a genuine file.


@jsquyres any thoughts on that ?

rhc54:
Sorry, I guess I wasn't clear. I don't see any reason we would care if the file is regular of a symlink. Either way should be fine for our purposes. If we are looking for a directory, then it sounds like -e is the correct option to use. If we are looking for a file, then -f sounds fine.

ggouaillardet:
ok then, since we are explicitly looking for a (symlink to a) file here and not a (symlink to a) directory, then `test -f` should be used. I reverted the changes and will commit and PR to the release branches once CI completes.

PR #4860, opened 2018-02-25
configury: fix PMI detection

ggouaillardet:

(cherry picked from commit open-mpi/ompi@b86e0f04bfca1d6c14177e00544f5e0a500ee02c)

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -14,8 +14,8 @@
 # Copyright (c) 2011-2014 Los Alamos National Security, LLC. All rights
 #                         reserved.
 # Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
-# Copyright (c) 2014-2017 Research Organization for Information Science
-#                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2014-2018 Research Organization for Information Science
+#                         and Technology (RIST).  All rights reserved.
 # Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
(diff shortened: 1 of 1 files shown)


config/opal_check_pmi.m4
```diff
@@ -39,18 +39,30 @@ AC_DEFUN([OPAL_CHECK_PMI_LIB],
...
 
     # check for the header
-    AC_MSG_CHECKING([for $3.h in $1/include])
-    AS_IF([test -f $1/include/$3.h],
-          [AC_MSG_RESULT([found])
-           opal_check_$3_mycppflags="-I$1/include"],
-          [AC_MSG_RESULT([not found])
-           AC_MSG_CHECKING([for $3.h in $1/include/slurm])
-           AS_IF([test -f $1/include/slurm/$3.h],
+    AS_IF([test -n "$1"],
```
rhc54: Urrr...what does ```test -n``` do?

config/opal_check_pmi.m4
```diff
@@ -39,18 +39,30 @@ AC_DEFUN([OPAL_CHECK_PMI_LIB],
...
+          [AC_MSG_CHECKING([for $3.h in $1/include])
```
rhc54: I note that we never just check to see if the header is present in where they told us to look - everything immediately puts an ```/include``` after it. I believe people have been complaining about that - shouldn't this check pattern start by checking where they told us it is?

config/opal_check_pmi.m4
ggouaillardet: `test -n string` is equivalent to `test ! -z string`. In this case, `$1` is empty when configure'd with `--with-pmix` or `--with-pmix=yes`, and not empty when configure'd with `--with-pmix=DIR`

config/opal_check_pmi.m4
ggouaillardet: well, the goal of this PR is to stop ending up with `-L/usr/lib[64]` in the `LDFLAGS`, and I kept everything else unmodified.
That being said, this is a fair point and I will address this in a new commit and append it to this PR and its siblings for the other release branches.
Shall I stop searching in `$1/include/slurm` when `$1` is not empty then ?

rhc54: I would continue that search as slurm does that strange thing, and users have come to expect it.

PR #4862, opened 2018-02-25
configury: fix PMI detection

ggouaillardet:

(back-ported from commit open-mpi/ompi@b86e0f04bfca1d6c14177e00544f5e0a500ee02c)


jsquyres:
@hppritcha I think we should not take this one as a[nother] way of encouraging users to move up to (at least) 2.x.

ggouaillardet:
@jsquyres this PR is for the `v2.x` series (e.g. __not__ the no more supported `v2.0.x` series). Did you mean you want to encourage users to move the the `v3` series instead ?

PR #4864, opened 2018-02-26
configury: look for PMI header in DIR provided by --with-pmi=DIR

ggouaillardet:

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -40,24 +40,29 @@ AC_DEFUN([OPAL_CHECK_PMI_LIB],
 
     # check for the header
     AS_IF([test -n "$1"],
-          [AC_MSG_CHECKING([for $3.h in $1/include])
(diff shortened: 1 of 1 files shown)


ggouaillardet:
I think this is a `PMI` specific issue caused by the snippet below, and not a generic one.

```
AS_IF([test ! -z "$with_pmi" && test "$with_pmi" != "yes"],
                 [check_pmi_install_dir=$with_pmi],
                 [check_pmi_install_dir=/usr])
```

The same pattern might be present in other parts of the code, but that has not been reported (yet).
I'd rather fix this issue that was reported via the ML, complete the `OPAL_CHECK_PACKAGE` revamp, and then track this kind of issue (if any left)

rhc54:
Agreed - I approved it because I wasn't suggesting we fix everything here. I only wanted to note that we probably do need to consider solving the entire issue, and shouldn't lose track of that problem

ggouaillardet:
thanks, I updated #4860, #4861 and #4862

PR #4878, opened 2018-03-02
Make more robust in finding NAG Fortran Compiler

ThemosTsikas:
the NAG Fortran check only matched nagfor exactly, and failed if a path to nagfor was provided.

### DIFF
--- a/config/ltmain_nag_pthread.diff
+++ b/config/ltmain_nag_pthread.diff
@@ -8,7 +8,7 @@
  	if test -n "$inherited_linker_flags"; then
 -	  tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g'`
 +          case "$CC" in
-+            nagfor*)
++            *nagfor*)
 +	      tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g' | $SED 's/-pthread/-Wl,-pthread/g'`;;
 +            *)
 +	      tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g'`;;


jsquyres:
ok to test

jsquyres:
@ThemosTsikas Can you explain why the `*` is necessary before `nagfor`?  (I have zero experience with the NAG compiler)

ThemosTsikas:
Hello,

You can get a trial licence https://www.nag.co.uk/content/nag-software-trials

I tried an OpenMPI build by setting FC to a full pathname ending in nagfor and the make failed because that line did not get picked up. Although it says CC, it's actually the fortran compiler that is tested. There may be other places where this happens but it did not affect the build.  Also, there is an issue on MacOS that I would like to fix: NAG Compiler needs -fpp option alongside -mismatch because .F90 files are not recognised as needing a preprocessor on MacOS.

ThemosTsikas:
By the way, I work at NAG on our Compiler Support Team!

jsquyres:
@ThemosTsikas Thanks for submitting and explaining!

I think you didn't push your new commit to github -- it looks like the same one, and the signed-off-by checker is still showing failure.  If you `git commit --amend -s`, you can fix up your commit message, but then you'll have to force push to your github fork branch (which is just about the *only* acceptable time to force push 😄).  Then the new commit should be seen by the PR, and CI will kick off again, yadda yadda yadda.

ThemosTsikas:
I am only using the github interface, here... Not sure what's needed, perhaps another person can push the change?

jsquyres:
@ThemosTsikas Heh, this is one of those catch-22 open source situations: you've submitted a patch, but we need a "Signed off" line *on the commit* (not the PR description).  If we just discard your patch and make another identical patch, then we're stealing from you.  😦 

I think all you want to do is:

```sh
# This will add the signed off line to your commit message
$ git commit --amend -s

# This will force push the new commit up to your branch on your fork
$ push push git@github.com:ThemosTsikas/ompi.git --force --dry-run
```

If that looks like it works properly, then remove the `--dry-run` and do the `git push` again.  That should push the commit up to github, and all should be well.

ThemosTsikas:
Please, please, please steal from me!

jsquyres:
LOL!  Ok, how about this...

/me does some magic

Ok, done.  Please check https://github.com/open-mpi/ompi/pull/4878/commits/4098409670e5fbdd3c8abdd2823e9ab0065f7e92 -- I signed off both you and me in the commit message.  Is that acceptable to you?  If you say that's ok here on this PR, that'll be good enough for us.

ThemosTsikas:
Thank you Jeff, that is fine by me.

jsquyres:
Oops -- I lied; I forgot to update the commit message to be more descriptive, too (i.e., that hash I just referenced is now invalid, sorry). Just check the commit on this PR and see both your and my signoffs, and then give us an ok here on the PR.

jsquyres:
Great!  We'll wait for CI to finish and then merge.

Thanks!

jsquyres:
@bwbarrett CI has finished.  Could you review again?

PR #4882, opened 2018-03-06
Fix MPIR Attach Mode for mpirun

jjhursey:
MPIR Attach Mode requires that the MPIR_proctable be available in the
starter process so that a tool can later attach to the running processes.

This can be limiting to scalability since the MPIR_PROCDESC data structure
requires knowledge of the process pid, which has to be reported back to
the starter. For this scalability reason we do not create the proctable
unless the user asks to do so (or the sysadmin sets this MCA parameter
globally.

### DIFF
--- a/orte/mca/plm/base/plm_base_launch_support.c
+++ b/orte/mca/plm/base/plm_base_launch_support.c
@@ -16,7 +16,7 @@
  * Copyright (c) 2013-2018 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014-2018 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2018 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -772,7 +772,7 @@ void orte_plm_base_registered(int fd, short args, void *cbdata)
     jdata->state = caddy->job_state;
 
    /* if this wasn't a debugger job, then need to init_after_spawn for debuggers */
-    if (!ORTE_FLAG_TEST(jdata, ORTE_JOB_FLAG_DEBUGGER_DAEMON)) {
+    if (orte_debugger_always_create_proctable || !ORTE_FLAG_TEST(jdata, ORTE_JOB_FLAG_DEBUGGER_DAEMON)) {
         ORTE_ACTIVATE_JOB_STATE(jdata, ORTE_JOB_STATE_READY_FOR_DEBUGGERS);
     }
 
--- a/orte/mca/state/base/state_base_fns.c
+++ b/orte/mca/state/base/state_base_fns.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2011-2012 Los Alamos National Security, LLC.
  * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2018      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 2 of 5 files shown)


jjhursey:
For the reviewers - I added you two since you are best able to review the expectation of the MPIR interface. The ORTE changes are pretty straightforward. (FYI @rhc54 - not a reviewer, but might be interested since I spoke with him about it last week).

Consider an MPI application that was launched and is thought to be "hung". The MPIR Attach Mode allows us to attach to the `mpirun` process read the `MPIR_proctable` and then launch out to attach to the remote processes.

ORTE currently only fills in the `MPIR_proctable` if the `--debug` command line option is specified by the user when the launch their application. In the scenario above, they didn't expect to hang so did not specify the flag. Further, specifying `--debug` activates a MPIR Job Launch Mode which requires that we launch the debugger with the application - so not really the MPIR Attach Mode.


This PR adds an MCA parameter `orte_always_create_proctable` that forces ORTE to create the `MPIR_proctable` on every launch. Since this has scalability concerns it default to `false`. I am on the fence here about this default value. I lead towards setting it to `true` by default so the user doesn't have to do anything special to get this support, and then if they have scalability concerns they can always turn it off. What do you think?

jjhursey:
Examples:

```
shell$ mpirun --host n02:2,n03:2 --np 4 -mca orte_output_debugger_proctable t  sleep 30
shell$ mpirun --host n02:2n03:2 --np 4 -mca orte_output_debugger_proctable t -mca orte_always_create_proctable t sleep 30
MPIR Proctable for job [4483,1]
0: Host n02 Exe /home/test/sleep Pid 146454
1: Host n02 Exe /home/test/sleep Pid 146455
2: Host n03 Exe /home/test/sleep Pid 83571
3: Host n03 Exe /home/test/sleep Pid 83572
```

```
shell$ cat gdb-test-cmd.txt 
echo MPIR_proctable_size\n
p MPIR_proctable_size

echo MPIR_proctable (pointer)\n
p MPIR_proctable

echo MPIR_proctable (contents)\n
p *MPIR_proctable@MPIR_proctable_size
shell$ gdb --batch -x gdb-test-cmd.txt --pid `ps aux | grep mpirun | grep -v grep | awk '{print $2}'`
[New LWP 100862]
[New LWP 100861]
[New LWP 100860]
[New LWP 100859]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
0x00003fffac42d5d8 in poll () from /lib64/libc.so.6
MPIR_proctable_size
$1 = 4
MPIR_proctable (pointer)
$2 = (struct MPIR_PROCDESC *) 0x1000d3cbd10
MPIR_proctable (contents)
$3 = {{
    host_name = 0x1000d3fcff0 "n02", 
    executable_name = 0x1000d3e2b20 "/home/test/sleep", 
    pid = 146454
  }, {
    host_name = 0x1000d3fbce0 "n02", 
    executable_name = 0x1000d3e80b0 "/home/test/sleep", 
    pid = 146455
  }, {
    host_name = 0x1000d3fbd50 "n03", 
    executable_name = 0x1000d3e83c0 "/home/test/sleep", 
    pid = 83571
  }, {
    host_name = 0x1000d3e7e00 "n03", 
    executable_name = 0x1000d3e7e20 "/home/test/sleep", 
    pid = 83572
  }}
```

jjhursey:
It looks like if you run an MPI application then the proctable is set, but if it not an MPI application then it's not. I'll need to investigate some more. Putting the WIP tag on this for the moment.

rhc54:
Just to clarify: the scalability issue isn't really about the pid. The ORTE daemon always reports the pid today so that the proctable can be constructed, if needed. The issue is with the memory footprint of mpirun as the proctable is actually rather large at scale.

rhc54:
@jjhursey The proctable only gets set for MPI jobs by design - construction gets triggered once all procs in the job have called PMIx_Init via MPI_Init. You can see that here:

https://github.com/open-mpi/ompi/blob/04ec013da99d305d1c5bed02a98c3a21aacb5d78/orte/mca/plm/base/plm_base_launch_support.c#L776

rhc54:
Reason: the MPIR debuggers out there don't know how to debug anything except MPI jobs 😄

jjhursey:
@rhc54 Thanks for the explanation. I should have noticed that the PLM change was on registration - that should have been my tip off that it would activate this path for MPI application.
I'll work with the original reporter to get more clarification. I think this PR is invalid then.

PR #4883, opened 2018-03-07
Topic/get element fix

bosilca:
Fix the calculation of the number of elements in the datatype. As the loop variable has been set outside the internal loop, we were not correctly updating the loop stats, and we were counting more elements than actually existing in the datatype representation.

### DIFF
--- a/opal/datatype/opal_datatype_get_count.c
+++ b/opal/datatype/opal_datatype_get_count.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; -*- */
 /*
- * Copyright (c) 2004-2017 The University of Tennessee and The University
+ * Copyright (c) 2004-2018 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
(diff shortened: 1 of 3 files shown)


ibm-ompi:

 Gist: https://gist.github.com/2f5b1e23de66e06b8982acd00e2c336f

artpol84:
@jladd-mlnx @Di0gen @yosefe 
@jsquyres just pinged me about this Jenkins failure.
It looks like ompi can’t find UCX. Can someone have a look?

jsquyres:
I pushed a missing include file fix to this PR.  If it passes CI, I'll squash it back down to George's original commit and then we can merge.

bosilca:
I compile in debug mode, and the output was available there. I'll squash.

jsquyres:
@bosilca Cool -- thanks!

alinask:
bot:mellanox:retest

artpol84:
@alinask thanks!

edgargabriel:
Are we good to merge this pr, i.e. has the squash been already done?

jsquyres:
@edgargabriel Yes, squash has been done -- we're good to go.

jsquyres:
@bosilca This PR changed the default value of the MCA param `opal_cuda_verbose` from 0 to -1.

Was that intended?  That doesn't seem to have anything to do with this GET_ELEMENT fix.

jsquyres:
@bosilca Is this fix also applicable to v2.1.x?  The cherry picks did not apply anywhere close to clean for me.

bosilca:
@jsquyres the opal_cuda_verbose was not supposed to be part of this. Fortunately, it has no impact on the code. I will change back in master, I have another cleanup pending.

The bug is present in 2.1,e and as you noticed this patch will not apply as is. One will have to search for the lines in this patch and apply them by hand. I am on not available for the next week, but I'll try to send you a quick patch to test.

PR #4885, opened 2018-03-07
Initial Software-based Performance Counters PR

davideberius:
Added Software-based Performance Counters driver code along with several counters.
    
> This code is the implementation of Software-base Performance Counters as described in the paper 'Using Software-Base Performance Counters to Expose Low-Level Open MPI Performance Information' in EuroMPI/USA '17 (http://icl.cs.utk.edu/news_pub/submissions/software-performance-counters.pdf).  More practical usage information can be found here: https://github.com/davideberius/ompi/wiki/How-to-Use-Software-Based-Performance-Counters-(SPCs)-in-Open-MPI.
> 
> All software events functions are put in macros that become no-ops when SOFTWARE_EVENTS_ENABLE is not defined.  The internal timer units have been changed to cycles to avoid division operations which was a large source of overhead as discussed in the paper.  Added a --with-spc configure option to enable SPCs in the Open MPI build.  This defines SOFTWARE_EVENTS_ENABLE.  Added an MCA parameter, mpi_spc_enable, for turning on specific counters.  Added an MCA parameter, mpi_spc_dump_enabled, for turning on and off dumping SPC counters in MPI_Finalize.  Added an SPC test and example.

### DIFF
--- a/examples/spc_example.c
+++ b/examples/spc_example.c
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2018      The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ *
+ * Simple example usage of SPCs through MPI_T.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "mpi.h"
+
+/* Sends 'num_messages' messages of 'message_size' bytes from rank 0 to rank 1.
+ * All messages are send synchronously and with the same tag in MPI_COMM_WORLD.
+ */
+void message_exchange(int num_messages, int message_size)
+{
+    int i, rank;
+    /* Use calloc to initialize data to 0's */
+    char *data = (char*)calloc(message_size, sizeof(char));
+    MPI_Status status;
+
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    if(rank == 0) {
+        for(i = 0; i < num_messages; i++)
+            MPI_Send(data, message_size, MPI_BYTE, 1, 123, MPI_COMM_WORLD);
+    } else if(rank == 1) {
+        for(i = 0; i < num_messages; i++)
+            MPI_Recv(data, message_size, MPI_BYTE, 0, 123, MPI_COMM_WORLD, &status);
+    }
+
+    free(data);
+}
+
+int main(int argc, char **argv)
+{
+    int num_messages, message_size;
+
+    if(argc < 3) {
+        printf("Usage: mpirun -np 2 --mca mpi_spc_attach all --mca mpi_spc_dump_enabled true ./spc_example [num_messages] [message_size]\n");
+        return -1;
+    } else {
+        num_messages = atoi(argv[1]);
+        message_size = atoi(argv[2]);
+    }
+
+    int i, rank, size, provided, num, name_len, desc_len, verbosity, bind, var_class, readonly, continuous, atomic, count, index;
+    MPI_Datatype datatype;
+    MPI_T_enum enumtype;
+    MPI_Comm comm;
+    char name[256], description[256];
+
+    /* Counter names to be read by ranks 0 and 1 */
+    char *counter_names[] = {"runtime_spc_OMPI_BYTES_SENT_USER",
+                             "runtime_spc_OMPI_BYTES_RECEIVED_USER" };
+
+    MPI_Init(NULL, NULL);
+    MPI_T_init_thread(MPI_THREAD_SINGLE, &provided);
+
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &size);
+    if(size != 2) {
+        fprintf(stderr, "ERROR: This test should be run with two MPI processes.\n");
+        MPI_Abort(MPI_COMM_WORLD, -1);
+    }
+
+    /* Determine the MPI_T pvar indices for the OMPI_BYTES_SENT/RECIEVED_USER SPCs */
+    index = -1;
+    MPI_T_pvar_get_num(&num);
+    for(i = 0; i < num; i++) {
+        name_len = desc_len = 256;
+        PMPI_T_pvar_get_info(i, name, &name_len, &verbosity,
+                             &var_class, &datatype, &enumtype, description, &desc_len, &bind,
+                             &readonly, &continuous, &atomic);
+        if(strcmp(name, counter_names[rank]) == 0) {
+            index = i;
+            printf("[%d] %s -> %s\n", rank, name, description);
+        }
+    }
+
+    /* Make sure we found the counters */
+    if(index == -1) {
+        fprintf(stderr, "ERROR: Couldn't find the appropriate SPC counter in the MPI_T pvars.\n");
+        MPI_Abort(MPI_COMM_WORLD, -1);
+    }
+
+    int ret;
+    long long value;
+
+    MPI_T_pvar_session session;
+    MPI_T_pvar_handle handle;
+    /* Create the MPI_T sessions/handles for the counters and start the counters */
+    ret = MPI_T_pvar_session_create(&session);
+    ret = MPI_T_pvar_handle_alloc(session, index, NULL, &handle, &count);
+    ret = MPI_T_pvar_start(session, handle);
+
+    message_exchange(num_messages, message_size);
+
+    ret = MPI_T_pvar_read(session, handle, &value);
+    /* Print the counter values in order by rank */
+    for(i = 0; i < 2; i++) {
+        if(i == rank) {
+            printf("[%d] Value Read: %lld\n", rank, value);
+            fflush(stdout);
+        }
+        MPI_Barrier(MPI_COMM_WORLD);
+    }
+    /* Stop the MPI_T session, free the handle, and then free the session */
+    ret = MPI_T_pvar_stop(session, handle);
+    ret = MPI_T_pvar_handle_free(session, &handle);
+    ret = MPI_T_pvar_session_free(&session);
+
+    MPI_T_finalize();
+    MPI_Finalize();
+
+    return 0;
+}
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2006 The University of Tennessee and The University
+ * Copyright (c) 2004-2018 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
@@ -74,6 +74,9 @@ bool ompi_async_mpi_finalize = false;
 #define OMPI_ADD_PROCS_CUTOFF_DEFAULT 0
 uint32_t ompi_add_procs_cutoff = OMPI_ADD_PROCS_CUTOFF_DEFAULT;
 bool ompi_mpi_dynamics_enabled = true;
+
+char *ompi_mpi_spc_attach_string = NULL;
+bool ompi_mpi_spc_dump_enabled = false;
 
 static bool show_default_mca_params = false;
 static bool show_file_mca_params = false;
@@ -314,6 +317,22 @@ int ompi_mpi_register_params(void)
         (void) mca_base_var_register_synonym(value, "ompi", "mpi", NULL, "abort_print_stack",
                                       MCA_BASE_VAR_SYN_FLAG_DEPRECATED);
     }
+
+    ompi_mpi_spc_attach_string = NULL;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "spc_attach",
+                                 "A comma delimeted string listing the software-based performance counters (SPCs) to enable.",
+                                 MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
+                                 OPAL_INFO_LVL_4,
+                                 MCA_BASE_VAR_SCOPE_READONLY,
+                                 &ompi_mpi_spc_attach_string);
+
+    ompi_mpi_spc_dump_enabled = false;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "spc_dump_enabled",
+                                 "A boolean value for whether (true) or not (false) to enable dumping SPC counters in MPI_Finalize.",
+                                 MCA_BASE_VAR_TYPE_BOOL, NULL, 0, 0,
+                                 OPAL_INFO_LVL_4,
+                                 MCA_BASE_VAR_SCOPE_READONLY,
+                                 &ompi_mpi_spc_dump_enabled);
 
     return OMPI_SUCCESS;
 }
--- a/ompi/runtime/ompi_spc.c
+++ b/ompi/runtime/ompi_spc.c
@@ -1,0 +1,458 @@
+/*
+ * Copyright (c) 2018      The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ *
+ * Copyright (c) 2018      Cisco Systems, Inc.  All rights reserved
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_spc.h"
+
+opal_timer_t sys_clock_freq_mhz = 0;
+
+static void ompi_spc_dump(void);
+
+/* Array for converting from SPC indices to MPI_T indices */
+OMPI_DECLSPEC int mpi_t_offset = -1;
+OMPI_DECLSPEC bool mpi_t_enabled = false;
+
+OPAL_DECLSPEC ompi_communicator_t *comm = NULL;
+
+typedef struct ompi_spc_event_t {
+    const char* counter_name;
+    const char* counter_description;
+} ompi_spc_event_t;
+
+#define SET_COUNTER_ARRAY(NAME, DESC)   [NAME] = { .counter_name = #NAME, .counter_description = DESC }
+
+static ompi_spc_event_t ompi_spc_events_names[OMPI_SPC_NUM_COUNTERS] = {
+    SET_COUNTER_ARRAY(OMPI_SPC_SEND, "The number of times MPI_Send was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BSEND, "The number of times MPI_Bsend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_RSEND, "The number of times MPI_Rsend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SSEND, "The number of times MPI_Ssend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_RECV, "The number of times MPI_Recv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_MRECV, "The number of times MPI_Mrecv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ISEND, "The number of times MPI_Isend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IBSEND, "The number of times MPI_Ibsend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IRSEND, "The number of times MPI_Irsend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ISSEND, "The number of times MPI_Issend was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IRECV, "The number of times MPI_Irecv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SENDRECV, "The number of times MPI_Sendrecv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SENDRECV_REPLACE, "The number of times MPI_Sendrecv_replace was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_PUT, "The number of times MPI_Put was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_RPUT, "The number of times MPI_Rput was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_GET, "The number of times MPI_Get was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_RGET, "The number of times MPI_Rget was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_PROBE, "The number of times MPI_Probe was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IPROBE, "The number of times MPI_Iprobe was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BCAST, "The number of times MPI_Bcast was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IBCAST, "The number of times MPI_Ibcast was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_REDUCE, "The number of times MPI_Reduce was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IREDUCE, "The number of times MPI_Ireduce was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IREDUCE_SCATTER, "The number of times MPI_Ireduce_scatter was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IREDUCE_SCATTER_BLOCK, "The number of times MPI_Ireduce_scatter_block was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ALLREDUCE, "The number of times MPI_Allreduce was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLREDUCE, "The number of times MPI_Iallreduce was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SCAN, "The number of times MPI_Scan was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ISCAN, "The number of times MPI_Iscan was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SCATTER, "The number of times MPI_Scatter was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_SCATTERV, "The number of times MPI_Scatterv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ISCATTER, "The number of times MPI_Iscatter was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ISCATTERV, "The number of times MPI_Iscatterv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_GATHER, "The number of times MPI_Gather was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_GATHERV, "The number of times MPI_Gatherv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IGATHER, "The number of times MPI_Igather was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ALLTOALL, "The number of times MPI_Alltoall was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLTOALL, "The number of times MPI_Ialltoall was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLTOALLV, "The number of times MPI_Ialltoallv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLTOALLW, "The number of times MPI_Ialltoallw was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_NEIGHBOR_ALLTOALL, "The number of times MPI_Neighbor_alltoall was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_NEIGHBOR_ALLTOALLV, "The number of times MPI_Neighbor_alltoallv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_NEIGHBOR_ALLTOALLW, "The number of times MPI_Neighbor_alltoallw was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ALLGATHER, "The number of times MPI_Allgather was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_ALLGATHERV, "The number of times MPI_Allgatherv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLGATHER, "The number of times MPI_Iallgather was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IALLGATHERV, "The number of times MPI_Iallgatherv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_NEIGHBOR_ALLGATHER, "The number of times MPI_Neighbor_allgather was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_NEIGHBOR_ALLGATHERV, "The number of times MPI_Neighbor_allgatherv was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_TEST, "The number of times MPI_Test was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_TESTALL, "The number of times MPI_Testall was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_TESTANY, "The number of times MPI_Testany was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_TESTSOME, "The number of times MPI_Testsome was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_WAIT, "The number of times MPI_Wait was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_WAITALL, "The number of times MPI_Waitall was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_WAITANY, "The number of times MPI_Waitany was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_WAITSOME, "The number of times MPI_Waitsome was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BARRIER, "The number of times MPI_Barrier was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_IBARRIER, "The number of times MPI_Ibarrier was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_WTIME, "The number of times MPI_Wtime was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_CANCEL, "The number of times MPI_Cancel was called."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_RECEIVED_USER, "The number of bytes received by the user through point-to-point communications. Note: Includes bytes transferred using internal RMA operations."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_RECEIVED_MPI, "The number of bytes received by MPI through collective, control, or other internal communications."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_SENT_USER, "The number of bytes sent by the user through point-to-point communications.  Note: Includes bytes transferred using internal RMA operations."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_SENT_MPI, "The number of bytes sent by MPI through collective, control, or other internal communications."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_PUT, "The number of bytes sent/received using RMA Put operations both through user-level Put functions and internal Put functions."),
+    SET_COUNTER_ARRAY(OMPI_SPC_BYTES_GET, "The number of bytes sent/received using RMA Get operations both through user-level Get functions and internal Get functions."),
+    SET_COUNTER_ARRAY(OMPI_SPC_UNEXPECTED, "The number of messages that arrived as unexpected messages."),
+    SET_COUNTER_ARRAY(OMPI_SPC_OUT_OF_SEQUENCE, "The number of messages that arrived out of the proper sequence."),
+    SET_COUNTER_ARRAY(OMPI_SPC_MATCH_TIME, "The number of microseconds spent matching unexpected messages.  Note: The timer used on the back end is in cycles, which could potentially be problematic on a system where the clock frequency can change.  On such a system, this counter could be inaccurate since we assume a fixed clock rate."),
+    SET_COUNTER_ARRAY(OMPI_SPC_UNEXPECTED_IN_QUEUE, "The number of messages that are currently in the unexpected message queue(s) of an MPI process."),
+    SET_COUNTER_ARRAY(OMPI_SPC_OOS_IN_QUEUE, "The number of messages that are currently in the out of sequence message queue(s) of an MPI process."),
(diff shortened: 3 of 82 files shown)


thananon:
Reference [paper](http://icl.cs.utk.edu/news_pub/submissions/software-performance-counters.pdf) for this PR.

jsquyres:
ok to test

ibm-ompi:

 Gist: https://gist.github.com/fc3f553cdcceef1058554ddce9862137

ibm-ompi:

 Gist: https://gist.github.com/e0e2bd8f20c25081c15e358cdd70da84

ggouaillardet:
@davideberius a bit is missing in `example/Makefile`

here is a patch that should fix it

```patch
commit df715db8ab2ac50e2851f700c59707ea200589d7
Date:   Thu Mar 8 14:52:15 2018 +0900

    examples: fix Makefile for spc_example
    
diff --git a/examples/Makefile b/examples/Makefile
index f7d7687..e77e783 100644
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -13,8 +13,8 @@
 # Copyright (c) 2011-2016 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2012      Los Alamos National Security, Inc.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.  All rights reserved.
-# Copyright (c) 2017      Research Organization for Information Science
-#                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2017-2018 Research Organization for Information Science
+#                         and Technology (RIST).  All rights reserved.
 # $COPYRIGHT$
 #
@@ -133,6 +133,8 @@ ring_c: ring_c.c
        $(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
 connectivity_c: connectivity_c.c
        $(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
+spc_example: spc_example.c
+       $(MPICC) $(CFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
 
 hello_cxx: hello_cxx.cc
        $(MPICXX) $(CXXFLAGS) $(LDFLAGS) $? $(LDLIBS) -o $@
```

ibm-ompi:

 Gist: https://gist.github.com/c5876e6706ec5e36ae6922d74a9b71d2

thananon:
Hmm, I wonder if it's something GNU specific in the code that breaks XLC?

```
  CC       pml_ob1_isend.lo
  CC       pml_ob1_rdma.lo
  CC       pml_ob1_rdmafrag.lo
  CC       pml_ob1_recvfrag.lo
  CC       pml_ob1_recvreq.lo
  CC       pml_ob1_sendreq.lo
  CC       pml_ob1_start.lo
1586-494 (U) INTERNAL COMPILER ERROR: Signal 11.
Calling signal handler...
/opt/ibm/xlC/13.1.5/bin/.orig/xlc_r: error: 1501-230 Internal compiler error; please contact your Service Representative. For more information visit:
http://www.ibm.com/support/docview.wss?uid=swg21110810
make[2]: *** [pml_ob1_isend.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory `/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/ompi/mca/pml/ob1'
make[1]: Leaving directory `/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/ompi'
```

jjhursey:
Here is the verbose make at that point, in case it helps:

```shell
shell$ make V=1
depbase=`echo pml_ob1_isend.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/sh ../../../../libtool  --tag=CC   --mode=compile xlc_r -DHAVE_CONFIG_H -I. -I../../../../opal/include -I../../../../ompi/include -I../../../../oshmem/include -I../../../../opal/mca/hwloc/hwloc2a/hwloc/include/private/autogen -I../../../../opal/mca/hwloc/hwloc2a/hwloc/include/hwloc/autogen -I../../../../ompi/mpiext/cuda/c   -I../../../.. -I../../../../orte/include -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/event/libevent2022/libevent -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/event/libevent2022/libevent/include -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/hwloc/hwloc2a/hwloc/include    -O3 -DNDEBUG -finline-functions -fno-strict-aliasing  -MT pml_ob1_isend.lo -MD -MP -MF $depbase.Tpo -c -o pml_ob1_isend.lo pml_ob1_isend.c &&\
mv -f $depbase.Tpo $depbase.Plo
libtool: compile:  xlc_r -DHAVE_CONFIG_H -I. -I../../../../opal/include -I../../../../ompi/include -I../../../../oshmem/include -I../../../../opal/mca/hwloc/hwloc2a/hwloc/include/private/autogen -I../../../../opal/mca/hwloc/hwloc2a/hwloc/include/hwloc/autogen -I../../../../ompi/mpiext/cuda/c -I../../../.. -I../../../../orte/include -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/event/libevent2022/libevent -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/event/libevent2022/libevent/include -I/home/mpiczar/jenkins/workspace/ompi_public_pr_master_xl/ompi-src/opal/mca/hwloc/hwloc2a/hwloc/include -O3 -DNDEBUG -finline-functions -fno-strict-aliasing -MT pml_ob1_isend.lo -MD -MP -MF .deps/pml_ob1_isend.Tpo -c pml_ob1_isend.c  -fPIC -DPIC -o .libs/pml_ob1_isend.o
Calling signal handler...
1586-494 (U) INTERNAL COMPILER ERROR: Signal 11.
/opt/ibm/xlC/13.1.5/bin/.orig/xlc_r: error: 1501-230 Internal compiler error; please contact your Service Representative. For more information visit:
http://www.ibm.com/support/docview.wss?uid=swg21110810
make: *** [pml_ob1_isend.lo] Error 1
```

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+}
+
+/* Initializes the events data structure and allocates memory for it if needed. */
+void events_init()
+{
+    int i;
+
+    /* If the events data structure hasn't been allocated yet, allocate memory for it */
+    if(events == NULL) {
+        events = (ompi_event_t*)malloc(OMPI_NUM_COUNTERS * sizeof(ompi_event_t));
```
jsquyres: Please check for malloc failure.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+        *counter_value /= sys_clock_freq_mhz;
+    /* If this is a high watermark counter, reset it after it has been read */
+    if(index == OMPI_MAX_UNEXPECTED_IN_QUEUE || index == OMPI_MAX_OOS_IN_QUEUE)
+        events[index].value = 0;
+
+    return MPI_SUCCESS;
+}
+
+/* Initializes the events data structure and allocates memory for it if needed. */
+void events_init()
```
jsquyres: Please declare static and use `(void)` for the parameter list (same for other instances below).

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+
+/* Initializes the SPC data structures and registers all counters as MPI_T pvars.
+ * Turns on only the counters that were specified in the mpi_spc_attach MCA parameter.  
+ */
+void ompi_spc_init()
+{
+    int i, j, ret, found = 0, all_on = 0;
+
+    /* Initialize the clock frequency variable as the CPU's frequency in MHz */
+    sys_clock_freq_mhz = opal_timer_base_get_freq() / 1000000;
```
jsquyres: What happens if the system clock frequency is variable?

Should we just be using the high-resolution clock, instead?

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+
+    events_init();
+
+    /* Get the MCA params string of counters to turn on */
+    char **arg_strings = opal_argv_split(ompi_mpi_spc_attach_string, ',');
```
jsquyres: Don't forget to `opal_argv_free(arg_strings)` when you leave this function.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+    /* Turn on only the counters that were specified in the MCA parameter */
+    for(i = 0; i < OMPI_NUM_COUNTERS; i++) {
+        if(all_on)
+            attached_event[i] = 1;
+        else {
+            /* Note: If no arguments were given, this will be skipped */
+            for(j = 0; j < num_args && found < num_args; j++) {
+                if(strcmp(counter_names[i], arg_strings[j]) == 0) {
+                    attached_event[i] = 1;
+                    found++;
```
jsquyres: Is this used anywhere?  If so, do you need to also increment it in the `all_on` conditional block?

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+
+        /* ########################################################################
+         * ################## Add Timer-Based Counter Enums Here ##################
+         * ########################################################################
+         */
+        /* If this is a timer event, sent the corresponding timer_event entry to 1 */
+        if(i == OMPI_MATCH_TIME)
+            timer_event[i] = 1;
+        else
+            timer_event[i] = 0;
```
jsquyres: Similar to high watermark enums, it would be a good idea if there was an attribute to indicate timer items.

Overall -- with the previous comments I made (static initialization, attributes), it might be better to have an array of structs with name, description, enum, attributes, etc.).

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+         */
+        if(ret != OPAL_ERROR) {
+            mpi_t_indices[i] = ret;
+
+            if(mpi_t_offset == -1) {
+                mpi_t_offset = ret;
+                prev = ret;
+            } else if(ret != prev + 1) {
+                mpi_t_disabled = 1;
+                fprintf(stderr, "There was an error registering SPCs as MPI_T pvars.  SPCs will be disabled for MPI_T.\n");
```
jsquyres: This should be an `opal_show_help()`.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+                break;
+            } else {
+                prev = ret;
+            }
+        } else {
+            mpi_t_indices[i] = -1;
+            mpi_t_disabled = 1;
+            fprintf(stderr, "There was an error registering SPCs as MPI_T pvars.  SPCs will be disabled for MPI_T.\n");
```
jsquyres: This should also be `opal_show_help()`.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+                                     OPAL_INFO_LVL_4, MPI_T_PVAR_CLASS_SIZE,
+                                     MCA_BASE_VAR_TYPE_UNSIGNED_LONG_LONG, NULL, MPI_T_BIND_NO_OBJECT,
+                                     MCA_BASE_PVAR_FLAG_READONLY | MCA_BASE_PVAR_FLAG_CONTINUOUS,
+                                     ompi_spc_get_count, NULL, ompi_spc_notify, NULL);
+
+        /* Initialize the mpi_t_indices array with the MPI_T indices.
+         * The array index indicates the SPC index, while the value indicates
+         * the MPI_T index.
+         */
+        if(ret != OPAL_ERROR) {
```
jsquyres: Should this actually be checking for `OPAL_SUCCESS`?

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+        }
+    }
+}
+
+/* Frees any dynamically alocated OMPI SPC data structures */
+void ompi_spc_fini()
+{
+#if SPC_ENABLE == 1
+    if(!ompi_mpi_spc_dump_enabled)
+        goto skip_dump;
```
jsquyres: Might be a little more pleasant for this `fini` function to just do the cleanup.  And if `ompi_mpi_spc_dump_enabled`, call a separate `ompi_spc_dump()` function to do the dump.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+
+    int i, j, rank, world_size, offset, err;
+    long long *recv_buffer, *send_buffer;
+
+    ompi_communicator_t *comm = &ompi_mpi_comm_world.comm;
```
jsquyres: It may not be safe to use MPI_COMM_WORLD here.  I think you might want to dup COMM_WORLD during your startup and then use that here in the dump (and then be sure to free that communicator when you're done with it).

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+    world_size = ompi_comm_size(comm);
+
+    /* Convert from cycles to usecs before sending */
+    for(i = 0; i < OMPI_NUM_COUNTERS; i++) {
+        if(timer_event[i])
+            SPC_CYCLES_TO_USECS(&events[i].value);
+    }
+
+    /* Aggregate all of the information on rank 0 using MPI_Gather on MPI_COMM_WORLD */
+    if(rank == 0) {
```
jsquyres: The only difference between these two blocks is the allocation of `recv_buffer`, right?  Might want to just make that the only thing dependent upon `rank == 0`.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+        }
+        err = comm->c_coll->coll_gather(send_buffer, OMPI_NUM_COUNTERS, MPI_LONG_LONG,
+                                        recv_buffer, OMPI_NUM_COUNTERS, MPI_LONG_LONG,
+                                        0, comm,
+                                        comm->c_coll->coll_gather_module);
+    }
+
+    /* Once rank 0 has all of the information, print the aggregated counter values for each rank in order */
+    if(rank == 0) {
+        fprintf(stdout, "OMPI Software Counters:\n");
```
jsquyres: Should we use `fprintf` or `opal_output` here?

Also, we use the name "Open MPI" publicly; we rarely use the name `OMPI` in help/output messages.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+        offset = 0; /* Offset into the recv_buffer for each rank */
+        for(j = 0; j < world_size; j++) {
+            fprintf(stdout, "World Rank %d:\n", j);
```
jsquyres: I think you mean "MPI_COMM_WORLD rank".

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+ * Note: This assumes that the 'cycles' argument is initialized to 0 if the timer
+ *       hasn't been started yet.
+ */
+void ompi_spc_timer_start(unsigned int event_id, opal_timer_t *cycles)
+{
+    /* Check whether cycles == 0.0 to make sure the timer hasn't started yet.
+     * This is denoted unlikely because the counters will often be turned off.
+     */
+    if(OPAL_UNLIKELY(attached_event[event_id] == 1 && *cycles == 0)) {
+        *cycles = opal_timer_base_get_cycles();
```
jsquyres: See above -- should this just use the high-precision timer.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+    "The maximum number of messages that the unexpected message queue(s) within an MPI process contained at once since the last reset of this counter.  \
+Note: This counter is reset each time it is read.",
+    "The maximum number of messages that the out of sequence message queue(s) within an MPI process contained at once since the last reset of this counter.  \
+Note: This counter is reset each time it is read."
+};
+
+/* An array of integer values to denote whether an event is activated (1) or not (0) */
+OMPI_DECLSPEC unsigned int attached_event[OMPI_NUM_COUNTERS] = { 0 };
+/* An array of integer values to denote whether an event is timer-based (1) or not (0) */
+OMPI_DECLSPEC unsigned int timer_event[OMPI_NUM_COUNTERS] = { 0 };
```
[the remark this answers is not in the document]
jsquyres: Are these DECLSPEC'ed for a reason?

thananon:
Per discussion at face to face meeting on 03/20/18:
- @davideberius will have to follow up with the reviews.
- SPC is bound for 4.0.0 release. No need to PR over to 3.0.x or 3.1.x.
- Have to get in touch with compiler guy from IBM to see whats wrong with XLC.
- "counter_names" and "counter_descriptions" might not be a good name for a symbol due to possible name collision. We might have to raise this problem with PAPI team.

ibm-ompi:

 Gist: https://gist.github.com/154b75812fe732a7906766782994adb0

ibm-ompi:

 Gist: https://gist.github.com/9c0c0618be5243ba277701cda384f8ef

ibm-ompi:

 Gist: https://gist.github.com/065cc2e477d9417e7d6e1987ee5dc903

bosilca:
The XLC compiler seems puzzled by the existence of the following line of code "do {} while (0);".

ompi/runtime/ompi_mpi_params.c
```diff
@@ -315,6 +318,22 @@ int ompi_mpi_register_params(void)
                                       MCA_BASE_VAR_SYN_FLAG_DEPRECATED);
     }
 
+    ompi_mpi_spc_attach_string = NULL;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "spc_attach",
+                                 "A comma delimeted string listing the SPC counters to enable.",
```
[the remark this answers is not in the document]
bosilca: Why do you think the "SPC counter" is redundant ?

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+/* An array of event structures to store the event data (name and value) */
+OMPI_DECLSPEC ompi_event_t *events = NULL;
+
+/* ##############################################################
+ * ################# Begin MPI_T Functions ######################
+ * ##############################################################
+ */
+static int ompi_spc_notify(mca_base_pvar_t *pvar, mca_base_pvar_event_t event, void *obj_handle, int *count)
+{
+    (void)obj_handle;
```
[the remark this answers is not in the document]
bosilca: cheap man solution because in this particular implementation we do not care about the handle.

ompi/runtime/ompi_spc.c
bosilca: These timers are short lived, they basically measure the cost of a match. So we prefer something fast to something slower but more accurate.

ompi/runtime/ompi_spc.c
bosilca: It has been fixed.

ibm-ompi:

 Gist: https://gist.github.com/504e194c3b351871519716e8e85541b6

ompi/runtime/ompi_mpi_params.c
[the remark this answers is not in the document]
thananon: He meant SPC is for "Software Performance Counters". 
So `SPC counters` is `software performance counters counters`?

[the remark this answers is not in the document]
davideberius: Yes, this is redundant similar to ATM Machine. It's actually Software-based Performance Counters, but I'm assuming most people will remember it without the '-based'.

ibm-ompi:

 Gist: https://gist.github.com/f6d621757ccb5ea57cfe94d7251b8f93

ompi/runtime/ompi_spc.h
```diff
@@ -0,0 +1,210 @@
...
+#define SPC_CYCLES_TO_USECS(cycles) \
+    ompi_spc_cycles_to_usecs(cycles)
+
+#define SPC_UPDATE_WATERMARK(watermark_enum, value_enum) \
+    ompi_spc_update_watermark(watermark_enum, value_enum)
+
+#else /* SPCs are not enabled */
+
+#define SPC_INIT()  \
+    do {} while (0)
```
jjhursey: _If_ this is the problem with XL, why don't you just have it:
```
#define SPC_INIT() ;
```

bosilca: I was under the impression that some compilers do not like empty statements such as ";;".

thananon: hopefully this line is our problem.
We can test this hypothesis by putting something random there right?

ibm-ompi:

 Gist: https://gist.github.com/fc7047f0071fd59b8deb07430b64b532

thananon:
Seems like we have a winner!

ompi/runtime/ompi_mpi_params.c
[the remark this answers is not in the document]
jsquyres: Yes -- as Arm said, it's "Software Performance Counter Counters".  My $0.02: Either say

> A comma-delimited string listing the SPCs to enable"

or

> A comma-delimited string listing the software performance counters (SPCs) to enable)

ompi/runtime/ompi_spc.c
jsquyres: I recently had a very similar discussion with a system administrator, but on the subject of continuous integration testing on github.

In general, the CI tests were great (just like they are here on the OMPI project): you can get cheap/free testing across a wide variety of platforms just by pushing a git commit.  The developers came to rely on these tests to exercise platforms that they did not have easy access to.

Over the course of a week, dozens of PRs were merged.

Suddenly, QA reported back that the software was horribly broken on several platforms.

When the developers looked into what happened, they were very confused: all the problems *should* have been caught by CI.

It turns out that the system administrators were focused on delivering *extremely fast* continuous integration tests.  Meaning: if a test was going to take too long, it was killed, and the test simply reported success.  This led to a lot of tests not actually being run, and therefore a lot of bugs slipped through.

In short: developers were relying on behavior that was not actually happening.  They got their CI results *fast*, but they weren't *accurate*.

A similar situation exists here: there are many reasons that the CPU frequency can change -- even a little bit -- and therefore these counters won't actually reflect reality.  But in the reported results, there will be no way to tell that.

What use is that?

ompi/runtime/ompi_spc.c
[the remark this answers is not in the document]
jsquyres: How about just using `__opal_attribute_unused__`?  It does the same thing without looking weird in the code.

ompi/runtime/ompi_mpi_params.c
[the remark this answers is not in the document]
davideberius: Whoops, I forgot to change that.  Commit incoming...

ompi/runtime/ompi_spc.c
jsquyres: This should still really be an `opal_show_help()`.

ompi/runtime/ompi_spc.c
```diff
@@ -106,12 +106,30 @@ static ompi_spc_event_t ompi_spc_events_names[OMPI_SPC_NUM_COUNTERS] = {
...
 /* An array of integer values to denote whether an event is activated (1) or not (0) */
-static unsigned int ompi_spc_attached_event[OMPI_SPC_NUM_COUNTERS] = { 0 };
+static uint32_t ompi_spc_attached_event[OMPI_SPC_NUM_COUNTERS / sizeof(uint32_t)] = { 0 };
 /* An array of integer values to denote whether an event is timer-based (1) or not (0) */
-static unsigned int ompi_spc_timer_event[OMPI_SPC_NUM_COUNTERS] = { 0 };
+static uint32_t ompi_spc_timer_event[OMPI_SPC_NUM_COUNTERS / sizeof(uint32_t)] = { 0 };
 /* An array of event structures to store the event data (name and value) */
 static ompi_event_t *ompi_spc_events = NULL;
 
+static inline void SET_SPC_BIT(uint32_t* array, int32_t pos)
```
jsquyres: Why not use `opal_bitmap_t`?

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,510 @@
...
+
+    /* Convert from MPI_T pvar index to SPC index */
+    int index = pvar->pvar_index - mpi_t_offset;
+    /* Set the counter value to the current SPC value */
+    *counter_value = events[index].value;
+    /* If this is a timer-based counter, convert from cycles to microseconds */
+    if(timer_event[index])
+        *counter_value /= sys_clock_freq_mhz;
+    /* If this is a high watermark counter, reset it after it has been read */
+    if(index == OMPI_MAX_UNEXPECTED_IN_QUEUE || index == OMPI_MAX_OOS_IN_QUEUE)
```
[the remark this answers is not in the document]
jsquyres: This comment is also still true.

jsquyres:
Hmm.  There's quite a few of my comments that are hidden by github in "outdated" blocks, but they weren't discussed / addressed.  E.g., why invent your own bitmap handling instead of using `opal_bitmap_t`?

bosilca:
opal_bitmap_t is an overkill for our needs and in this context the emphasis is on performance. This answer also applies to your question about frequencies, we cannot afford to call an expensive timer as we are in the critical path of the MPI library (as an example in the matching).

ompi/runtime/ompi_spc.c
davideberius: To be more specific, this should be >=0 because the return value of the mca_base_pvar_register function should be a valid array index on success.

thananon:
I just want to follow up on this PR. It has been a while.

Right now the code is functional. We just have some suggestions from @jsquyres that @bosilca does not agree with. (Regarding the CPU frequency changing dynamically, etc) So I request @ggouaillardet to pitch in.

If there is no discussion anymore, I suggest that we should merge this before it goes to the void.

thananon:
Seems like no response will be made. @jsquyres  can we agree and merge this?

examples/spc_example.c
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    int i, rank, size, provided, num, name_len, desc_len, verbosity, bind, var_class, readonly, continuous, atomic, count, index;
+    MPI_Datatype datatype;
+    MPI_T_enum enumtype;
+    MPI_Comm comm;
+    char name[256], description[256];
+
+    /* Counter names to be read by ranks 0 and 1 */
+    char counter_names[2][40] = { "runtime_spc_OMPI_BYTES_SENT_USER",
```
jsquyres: Should probably be `char *counter_names[] = { "...`

test/spc/spc_test.c
```diff
@@ -0,0 +1,174 @@
...
+
+int main(int argc, char **argv)
+{
+    int i, rank, size, provided, num, name_len, desc_len, verbosity, bind, var_class, readonly, continuous, atomic, count, index, MPI_result;
+    MPI_Datatype datatype;
+    MPI_T_enum enumtype;
+    char name[256], description[256];
+
+    /* Counter names to be read by ranks 0 and 1 */
+    char counter_names[2][40] = { "runtime_spc_OMPI_BYTES_SENT_USER",
```
jsquyres: Same as before: `char *counter_names[] = ...`

ompi/runtime/ompi_spc.h
```diff
@@ -0,0 +1,212 @@
...
+ * 4.) Instrument the Open MPI code base where it makes sense for
+ *     your counter to be modified using the SPC_RECORD macro.
+ *     Note: If your counter is timer-based you should use the
+ *     SPC_TIMER_START and SPC_TIMER_STOP macros to record
+ *     the time in cycles to then be converted to microseconds later
+ *     in the ompi_spc_get_count function when requested by MPI_T
+ */
+
+/* This enumeration serves as event ids for the various events */
+enum ompi_spc_counters_t {
```
jsquyres: My suggestion earlier was in shorthand.  This should probably be:

`typedef enum ompi_spc_counters { ... } ompi_spec_counters_t`

I think this will make it like many other declarations in OMPI.

ompi/runtime/ompi_spc.h
```diff
@@ -0,0 +1,212 @@
...
+    OMPI_BYTES_PUT,
+    OMPI_BYTES_GET,
+    OMPI_UNEXPECTED,
+    OMPI_OUT_OF_SEQUENCE,
+    OMPI_MATCH_TIME,
+    OMPI_UNEXPECTED_IN_QUEUE,
+    OMPI_OOS_IN_QUEUE,
+    OMPI_MAX_UNEXPECTED_IN_QUEUE,
+    OMPI_MAX_OOS_IN_QUEUE,
+    OMPI_SPC_NUM_COUNTERS /* This serves as the number of counters.  It must be last. */
```
jsquyres: My prior comment actually intended you to put `SPC_` in *all* of these enums.  `OMPI_SPC_NUM_COUNTERS` was a single example, but it also applies to `OMPI_SPC_MAX_OOS_IN_QUEUE`, `OMPI_SPC_MAX_UNEXPECTED_IN_QUEUE`, ...etc.  I know this is a PITA, but this is a limitation of C -- we need to have decent prefixes to our values as a primitive namespace.  Sorry.  😦

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,437 @@
...
+ */
+
+/* This function returns the current count of an SPC counter that has been retistered
+ * as an MPI_T pvar.  The MPI_T index is not necessarily the same as the SPC index,
+ * so we need to convert from MPI_T index to SPC index and then set the 'value' argument
+ * to the correct value for this pvar.
+ */
+static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle)
+{
+    (void) obj_handle;
```
jsquyres: As I mentioned twice, please use `__opal_attribute_unused__`.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,437 @@
...
+    /* Convert from MPI_T pvar index to SPC index */
+    int index = pvar->pvar_index - mpi_t_offset;
+    /* Set the counter value to the current SPC value */
+    *counter_value = (long long)ompi_spc_events[index].value;
+    /* If this is a timer-based counter, convert from cycles to microseconds */
+    if( IS_SPC_BIT_SET(ompi_spc_timer_event, index) ) {
+        *counter_value /= sys_clock_freq_mhz;
+    }
+    /* If this is a high watermark counter, reset it after it has been read */
+    if(index == OMPI_MAX_UNEXPECTED_IN_QUEUE || index == OMPI_MAX_OOS_IN_QUEUE) {
```
jsquyres: Per my previous comment:

> It might be better to have some kind of attribute somewhere that indicates that a particular SPC counter needs to be reset after being read (Vs. having a hard-coded list of specific enums here in the code).

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,437 @@
...
+
+    /* Convert from cycles to usecs before sending */
+    for(i = 0; i < OMPI_SPC_NUM_COUNTERS; i++) {
+        if( IS_SPC_BIT_SET(ompi_spc_timer_event, i) ) {
+            SPC_CYCLES_TO_USECS(&ompi_spc_events[i].value);
+        }
+    }
+
+    /* Aggregate all of the information on rank 0 using MPI_Gather on MPI_COMM_WORLD */
+    send_buffer = (long long*)malloc(OMPI_SPC_NUM_COUNTERS * sizeof(long long));
```
jsquyres: Per my previous comment, please check for `malloc` failure.

ompi/runtime/ompi_spc.h
```diff
@@ -0,0 +1,212 @@
...
+typedef size_t ompi_spc_value_t;
+
+/* A structure for storing the event data */
+typedef struct ompi_spc_s{
+    char *name;
+    ompi_spc_value_t value;
+} ompi_spc_t;
+
+/* Events data structure initialization function */
+void events_init(void);
```
jsquyres: This function name needs to be prefixed with `ompi_spc_`.

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,437 @@
...
+
+/* Gathers all of the SPC data onto rank 0 of MPI_COMM_WORLD and prints out all
+ * of the counter values to stdout.
+ */
+void ompi_spc_dump(void)
+{
+    int i, j, world_size, offset;
+    long long *recv_buffer = NULL, *send_buffer;
+
+    ompi_communicator_t *comm = &ompi_mpi_comm_world.comm;
```
jsquyres: Per my prior comment:

> It may not be safe to use MPI_COMM_WORLD here. I think you might want to dup COMM_WORLD during your startup and then use that here in the dump (and then be sure to free that communicator when you're done with it).

bosilca:
@jsquyres what we are measuring are very short time intervals (such as the matching time). The change for a clock frequency change during such measurements is minimal (frankly non-existent) so I would not be concerned about the lack of accuracy. Moreover, we never return to the user absolute timing, so even if there is a drift in frequency during the execution the impact on the value returned will be minimal.

jsquyres:
@bosilca I hear what you're saying.  And I agree: it'll be pretty rare for the frequency to change while you're making the measurement.  But:

1. It can still happen (e.g., if an app idles for a while before a message is received, e.g., `while (!done) { MPI_Test(...); if (flag) break; sleep(1); }`.  Yes, we've all seen apps that do this.  😉 ).
1. The frequency is measured at the beginning of the application -- not during each measurement.  The chance of the frequency being different at the beginning of the application vs. the time of measurement is actually not small.

Here's an example: 

* An always-full HPC cluster finishes one job.  
* The job scheduler does all of its cleanup stuff at the end of the job, and resets the nodes for the next job.  
* During this time -- which may take a minute or three -- all but one of the cores spin down to a lower frequency.  
* The job scheduler finally gets the next job and starts it up, one process per core.
* Each process calls MPI_INIT right away, which samples the frequency.
* The process on the first core may will get a high frequency (because it's where the job scheduler was running and has been busy the whole time).  But all the other cores may be in a lower frequency because they were more-or-less idle for a little while.  
* After a short while running the MPI apps, all the cores are back up to their full frequency, but the damage was already done: they measured their frequency during MPI_INIT as X, but now they're running at frequency Y (where X<Y).

That can definitely happen.

I'm also curious: what is the usefulness of having fast-but-inaccurate timings?  If x% of your timings could be incorrect, how is that data useful?

Maybe it would be enough to have 2 different methods of timing: one based on your current frequency method, and another based on slower/more-accurate methods (e.g., `clock_gettime()`), and have an MCA param to switch between the two.  There are tradeoffs to both methods, of course:
* Fast but possibly inaccurate.
* Slow by possibly affects the overall runtime (because it's adding "slow" clock checks in performance-critical code paths).

jsquyres:
Right after I hit submit, I thought of a better example where the frequency can change: while blocking for disk I/O.  E.g.:

```c
while (!done) {
    MPI_Test(...);
    if (flag) break;
    read(fd, large_chunk_of_disk_data, ...);
}
```

Given that the `read()` is blocking, it could actually swap out for a little while (especially if it's waiting on a device or the network), and could actually affect the core frequency.

(this is somewhat moot because the frequency is only measured at the beginning of the app, but it's just better example than the `sleep()` example I listed above)

davideberius:
I checked to make sure that we are verifying that the timers are monotonic.  The following code shows how this is done right now.  Should I add a message specific warning message here?
`int opal_timer_linux_open(void)
{
    int ret = OPAL_SUCCESS;

    if (mca_timer_base_monotonic && !opal_sys_timer_is_monotonic ()) {
#if OPAL_HAVE_CLOCK_GETTIME && (0 == OPAL_TIMER_MONOTONIC)
        struct timespec res;
        if( 0 == clock_getres(CLOCK_MONOTONIC, &res)) {
            opal_timer_linux_freq = 1.e3;
            opal_timer_base_get_cycles = opal_timer_linux_get_cycles_clock_gettime;
            opal_timer_base_get_usec = opal_timer_linux_get_usec_clock_gettime;
            return ret;
        }
#else
        /* Monotonic time requested but cannot be found. Complain! */
        opal_show_help("help-opal-timer-linux.txt", "monotonic not supported", true);
#endif  /* OPAL_HAVE_CLOCK_GETTIME && (0 == OPAL_TIMER_MONOTONIC) */
    }
    ret = opal_timer_linux_find_freq();
    opal_timer_base_get_cycles = opal_timer_linux_get_cycles_sys_timer;
    opal_timer_base_get_usec = opal_timer_linux_get_usec_sys_timer;
    return ret;
}`

thananon:
We would like to get this into 4.0. The fork is this Friday IIRC. We should reach agreement on that timer/tick rate or actually remove it completely. The new matching is fast anyway.

bosilca:
@jsquyres poke

ompi/runtime/ompi_spc.c
```diff
@@ -0,0 +1,447 @@
...
+    free(send_buffer);
+
+    comm->c_coll->coll_barrier(comm, comm->c_coll->coll_barrier_module);
+}
+
+/* Frees any dynamically alocated OMPI SPC data structures */
+void ompi_spc_fini(void)
+{
+#if SPC_ENABLE == 1
+    ompi_spc_dump();
```
davideberius: It looks like I forgot to add the check for whether dumping is enabled.  I'll add this when I get home this evening.

ompi/runtime/ompi_spc.c
davideberius: This will be fixed this evening.

ompi/runtime/ompi_spc.c
davideberius: I'll fix this this evening.

ompi/runtime/ompi_spc.c
davideberius: Actually, this was already fixed, but this line didn't change so the comment still shows.

davideberius:
Ok, I've made all of the changes requested.  What is the status of the frequency issue?

bosilca:
I don't think there is an issue anymore. We discussed it with Jeff privately, and I think we agreed that as long as there is a warning we are good to go. But I hope @jsquyres will confirm and then approve this PR so that we can merge.

jsquyres:
I just filed a quick PR (https://github.com/davideberius/ompi/pull/1 -- I couldn't seem to push to @davideberius's PR branch for some reason) with some minor suggestions.

In terms of the frequency issue, there's two points:

1. George pointed out that we're using the invariant frequency, so the change-the-frequency issue shouldn't be an issue.  Ok, I'm good with that.
2. However, that's still only relevant to an individual core.  I still wouldn't mind a warning of some kind if we *know* that you may well get some level of non-determinism in the results (e.g., if your process is not bound to core, and therefore you may see the cycle count from different cores upon subsequent sampling).

bosilca:
@davideberius highlighted that we have the same level of protection for timers as MPI_Wtime. And we inherit the same level of warning.

davideberius:
Ok, I'll squash this when I get back home tonight.  Should this all be one commit?

bosilca:
Technically there is no strict rule on the number of commits (the details are left to the PR author(s)). In this particular case, due to a significant number of updates, I would indeed squash down to 1.

davideberius:
I have squashed all of the commits into a single commit.

jsquyres:
bot:mellanox:retest
bot:ompi:retest

Looks like the Jenkins/Github git clone bug again. 😦

bosilca:
More than what @davideberius pinpointed to few comments above ?
```
opal_show_help("help-opal-timer-linux.txt", "monotonic not supported", true);
```

jsquyres:
@bosilca Yes, more than that (but perhaps similar?) -- `opal_sys_timer_is_monotonic()` only returns if the system clock supports monotonic time or not.  It does not take into effect process binding (which will be the more likely case: your system supports monotonic time, but your process isn't bound to where it will consistently get the same clock).

Dumb question: is the OMPI time-invariant clock-reading-code getting its values from a *core* or from a *package*?  (I assume core, but I don't know that for a fact)  I.e., if a process is bound to a socket/package, is it still going to always get consistent time?

bosilca:
My understanding is that it depends on the processor, but I have no idea how to check what processor has what type of monotonic hwclock. I agree that from a user perspective it could be nice to be extremely attentive, but realistically it's a lot of work for little benefit.

PR #4888, opened 2018-03-08
io/romio314: mark datatypes of size 0 as contiguous

edgargabriel:
this commit fixes an issue observed with romio314 and the hdf5 1.10.x testsuite.
The ADIOI_Datatype_iscontig() routine in romio314/src/io_romio314_module.c
will now return for a datatype of size 0 that it is contiguous, even if the extent
of the datatype is non-zero. This avoids a segmentation fault observed in the
ADIOI_Flatten routine, and fixes a particular bug with the hdf5 1.10.x testsuite in
OpenMPI with romio314.

### DIFF
--- a/ompi/mca/io/romio314/src/io_romio314_module.c
+++ b/ompi/mca/io/romio314/src/io_romio314_module.c
@@ -137,6 +137,14 @@ void ADIOI_Datatype_iscontig(MPI_Datatype datatype, int *flag)
      * In addition, if the data is contiguous but true_lb differes
      * from zero, ROMIO will ignore the displacement. Thus, lie!
      */
+
+    size_t size;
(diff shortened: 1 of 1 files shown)


jsquyres:
@edgargabriel We do want this on 2.1.x (but we don't care about 2.0.x).

edgargabriel:
not sure what to make of the failure in the pull request build checker. The only errors that I see in the output is something like

Ubuntu 16.04 (sir-fh7r9yan) is offline

Is it ok to go ahead and merge anyway?

jsquyres:
@edgargabriel Let's run it again.  It looks like false Jenkins failures (i.e., the errors had to do with trying to check out the PR from github.com).

bot:ompi:retest

PR #4891, opened 2018-03-08
Fix MPIR_proctable structure visibility

jjhursey:
* The `MPIR_PROCDESC` structure needs to be visible even in optimized
   builds so that debuggers can attach to `mpirun` and properly read the
   `MPIR_proctable`.
 * In the v2.0.x and v2.x series this structure resided in the `orterun`
   directory and included the `CFLAGS` fix included here. This code
   moved in the v3.x series and the `CFLAGS` did not move causing this
   issue.

### DIFF
--- a/orte/orted/Makefile.am
+++ b/orte/orted/Makefile.am
@@ -12,6 +12,7 @@
 #                         All rights reserved.
 # Copyright (c) 2014      Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2015      Intel, Inc. All rights reserved.
+# Copyright (c) 2018      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -29,7 +30,17 @@ headers += \
 
 lib@ORTE_LIB_PREFIX@open_rte_la_SOURCES += \
         orted/orted_main.c \
-        orted/orted_comm.c \
-        orted/orted_submit.c
+        orted/orted_comm.c
+
+# The MPIR portion of the library must be built with -g, even if
+# the rest of the library has other optimization flags.
+# Use an intermediate library to isolate the debug object.
+noinst_LTLIBRARIES += liborted_mpir.la
+liborted_mpir_la_SOURCES = \
+	orted/orted_submit.c
+liborted_mpir_la_CFLAGS = $(CFLAGS_WITHOUT_OPTFLAGS) $(DEBUGGER_CFLAGS)
+
+lib@ORTE_LIB_PREFIX@open_rte_la_LIBADD += liborted_mpir.la
+
 
 include orted/pmix/Makefile.am
--- a/orte/tools/orterun/Makefile.am
+++ b/orte/tools/orterun/Makefile.am
@@ -12,21 +12,13 @@
 # Copyright (c) 2008-2014 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2008      Sun Microsystems, Inc.  All rights reserved.
 # Copyright (c) 2015      Intel, Inc.  All rights reserved.
+# Copyright (c) 2018      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
 #
 # $HEADER$
 #
-
-# This is not quite in the Automake spirit, but we have to do it.
-# Since the totalview portion of the library must be built with -g, we
-# must eliminate the CFLAGS that are passed in here by default (which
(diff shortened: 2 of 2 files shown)


jjhursey:
@jsquyres Tagging you since you made the change 13 years ago to orterun (https://github.com/open-mpi/ompi/commit/4c590580537d487acaae580900d198b5b7e3ed23).
FYI @rhc54 this was the actual MPIR issue that we were seeing.

jjhursey:
Ref https://github.com/open-mpi/ompi/issues/4892

rhc54:
looks ok to me - thx!

orte/orted/Makefile.am
```diff
@@ -21,6 +22,15 @@
...
 # This makefile.am does not stand on its own - it is included from orte/Makefile.am
 
+# This is not quite in the Automake spirit, but we have to do it.
+# Since the totalview portion of the library must be built with -g, we
+# must eliminate the CFLAGS that are passed in here by default (which
+# may already have debugging and/or optimization flags).  We use
+# post-processed forms of the CFLAGS in the library targets down
+# below.
+
+CFLAGS = $(CFLAGS_WITHOUT_OPTFLAGS) $(DEBUGGER_CFLAGS)
```
jsquyres: Hmm.  It strikes me a bit of a bummer that we have to compile the entire `orted` without optimization flags.

Is there a way to segregate the needed symbols into a single, small `.c` file and compile *only that file* without optimization flags?

rhc54: one option you could consider - if you get people to use "prun" for executing jobs under the DVM, then the debugger code and orted_submit stuff could go back into orterun. It would take a little work to move it, but not too horrible.

ggouaillardet: fwiw, I made a proof of concept, and the `-g` flag only has to be passed to `orte/orted/orted_submit.lo`, `orte/libopen-rte.la` and `orte/tools/orted/orted`

the inline patch below addresses the last two ones, but I could not figure out how to pass that flag to `orted_submit.lo` (`orte/orted/Makefile.am` is included in `orte/Makefile.am`, so the "object" file is `orted/orted_submit.lo` and I do not know how to handle the `/` in `Makefile.am`

```patch
diff --git a/orte/Makefile.am b/orte/Makefile.am
index 6af81a2..c9d27bc 100644
--- a/orte/Makefile.am
+++ b/orte/Makefile.am
@@ -41,7 +41,7 @@ lib@ORTE_LIB_PREFIX@open_rte_la_LIBADD = \
        $(MCA_orte_FRAMEWORK_LIBS) \
        $(ORTE_TOP_BUILDDIR)/opal/lib@OPAL_LIB_PREFIX@open-pal.la
 lib@ORTE_LIB_PREFIX@open_rte_la_DEPENDENCIES = $(libopen_rte_la_LIBADD)
-lib@ORTE_LIB_PREFIX@open_rte_la_LDFLAGS = -version-info $(libopen_rte_so_version)
+lib@ORTE_LIB_PREFIX@open_rte_la_LDFLAGS = -version-info $(libopen_rte_so_version) $(DEBUGGER_CFLAGS)
 
 # included subdirectory Makefile.am's and appended-to variables
 headers =
diff --git a/orte/tools/orted/Makefile.am b/orte/tools/orted/Makefile.am
index 92211a3..32fe2bc 100644
--- a/orte/tools/orted/Makefile.am
+++ b/orte/tools/orted/Makefile.am
@@ -43,7 +43,7 @@ orted_SOURCES = orted.c
 #  nothing to -all-static in the Makefile.in
 #  nice for systems that don't have all the shared
 #  libraries on the computes
-orted_LDFLAGS =
+orted_LDFLAGS = $(DEBUGGER_CFLAGS)
 orted_LDADD = \
        $(top_builddir)/orte/lib@ORTE_LIB_PREFIX@open-rte.la \
```

jjhursey: We can use [this trick](https://www.gnu.org/software/automake/manual/html_node/Per_002dObject-Flags.html) to compile just `orted_submit.c` with debug symbols. Let me give that a try.

If we want to further narrow down how much of `orted_submit.c` is compiled this way, we can do that in another effort.

jjhursey:
@ggouaillardet @jsquyres I changed the PR to isolate the debug symbols to just the compilation of `orted_submit.c` which contains the MPIR symbols. This patch works, and isolates the applications of these `CFLAGS`. Let me know what you think.

I confirmed that this patch does fix the MPIR test case.

ggouaillardet:
@jjhursey very clever !

PR #4895, opened 2018-03-08
io/romio314: mark datatypes of size 0 as contiguous

edgargabriel:
this commit fixes an issue observed with romio314 and the hdf5 1.10.x testsuite.
The ADIOI_Datatype_iscontig() routine in romio314/src/io_romio314_module.c
will now return for a datatype of size 0 that it is contiguous, even if the extent
of the datatype is non-zero. This avoids a segmentation fault observed in the
ADIOI_Flatten routine, and fixes this particular with the hdf5 1.10.x testsuite in
OpenMPI with romio314.

(cherry picked from commit c83b47c266c4eac23e352f134e660dd8c6d532b1)


edgargabriel:
the comit email checker seems to be stuck

edgargabriel:
bot:retest

PR #4897, opened 2018-03-09
Be conservative with the array_of_indices

### LINKED ISSUE #4896: Possible illegal memory access in ompi_request_default_wait_some
```
pn1707994:scratch agaspar$ port info openmpi
openmpi @3.0.0 (science, parallel, net)
```

Installed via MacPorts

* Operating system/version: macOS Sierra
* Computer hardware: MacBook Pro (2017)
* Network type: running only on host (shared memory?)

-----------------------------

## Details of the problem
`ompi_request_default_wait_some` is writing to `array_of_indices` past `outcount`. See [this loop](https://github.com/open-mpi/ompi/blob/master/ompi/request/req_wait.c#L422) in req_wait.c.

The MPI v3.1 spec [seems to indicate](https://github.com/mpi-forum/mpi-issues/issues/86#issuecomment-371645898) that `array_of_indices` should not be written to past the value of `outcount`, and does not explicitly state that `array_of_indices` is of size `incount`. The user could reasonably expect that they need not pass an `array_of_indices` larger than the number of _active_ requests in `array_of_requests`.

bosilca:
We were assuming that the array_of_indices has the same size as the
number of requests (incount), instead of the numberr of actually
active requests. While the patch is trivial, the question of the
size of the array_of_indices should be clarified in the MPI Forum.

Fixes #4896

### DIFF
--- a/ompi/request/req_wait.c
+++ b/ompi/request/req_wait.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2010 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2016 The University of Tennessee and The University
+ * Copyright (c) 2004-2018 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2008 High Performance Computing Center Stuttgart,
@@ -388,13 +388,13 @@ int ompi_request_default_wait_some(size_t count,
                                    int * indices,
                                    ompi_status_public_t * statuses)
 {
-    size_t num_requests_null_inactive=0, num_requests_done=0;
+    size_t num_requests_null_inactive, num_requests_done, num_active_reqs;
     int rc = MPI_SUCCESS;
     ompi_request_t **rptr = NULL;
     ompi_request_t *request = NULL;
     ompi_wait_sync_t sync;
     size_t sync_sets = 0, sync_unsets = 0;
-    
+
     if (OPAL_UNLIKELY(0 == count)) {
         *outcount = MPI_UNDEFINED;
         return OMPI_SUCCESS;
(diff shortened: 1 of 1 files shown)


AndrewGaspar:
Thank you for taking a look at this George!

hjelmn:
@bosilca Looks like the forum is going with they must be the same size. No reason not to merge this. Up to you if you still want to.

bosilca:
@hjelmn my understanding of the forum ticket is that they go with the original way we had, where we assumed all 3 arrays (requests, indices and statuses) have the same size, equal to the max possible number of requests (aka. ```count```). This patch enables a slightly different approach where we reduce the sizes of the indices array to the number of active requests, which is potentially smaller. Other than this is has no impact (performance or memory saving) on the code.

jsquyres:
@AndrewGaspar @bosilca @hjelmn Where did the MPI Forum end up on this one -- should this PR be merged?  Are we waiting on the resolution of https://github.com/mpi-forum/mpi-issues/issues/86?

bosilca:
The forum made an unfortunate decision: in order to allow an intricate use of the status and indices arrays (that could have been simply replaced by an Waitall) the forum instead decided to force the MPI library to do a linear search on the entire array of requests (aka count elements) on every call, for the small benefit of using smaller than count arrays of indices and statuses. In conclusion we should merge this PR and move forward.

jsquyres:
Ok.

What is the consequence of this PR -- does it change user-noticeable behavior?  Should it be cherry-picked to release branches?

bosilca:
Depends. Applications using similar approaches to the code posted on the forum discussion will break with our previous implementation. For anybody else there will be no difference. But we should cherry-pick it on all branches to maintain compatibility with the MPI standard.

jsquyres:
@bosilca This cherry-picked cleanly to v4.0.x, but not to any other release branch.  Can you make a back port, perchance?

bosilca:
In order to do it in a timely way @thananon agree to give me a hand here.

Issue #4675, opened 2017-12-29
MPI_PROC_NULL in Topology Creation


omor1:
I'm trying to create a binary tree topology using `MPI_Dist_graph_create_adjacent()`, simplifying the graph boundaries by using `MPI_PROC_NULL`. This allows specifying all nodes in a consistent way.
I'm not sure this is allowed by the specification; I could find no information either way.
However, the neighborhood collectives specify that the borders of a cartesian topology act as though they send and receive from `MPI_PROC_NULL`. It could be useful to be able to obtain similar behavior in a generic graph.

Example code is given below:
```C
int world_rank;
int world_size;
int neighbor[3];

MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);
MPI_Comm_size(MPI_COMM_WORLD, &world_size);

neighbor[0] = world_rank > 0 ? (world_rank-1)/2 : MPI_PROC_NULL;
neighbor[1] = 2*world_rank+1 < world_size ? 2*world_rank+1 : MPI_PROC_NULL;
neighbor[2] = 2*world_rank+2 < world_size ? 2*world_rank+2 : MPI_PROC_NULL;

MPI_Dist_graph_create_adjacent(MPI_COMM_WORLD,
                               3, neighbor, MPI_UNWEIGHTED,
                               3, neighbor, MPI_UNWEIGHTED,
                               MPI_INFO_NULL, true, &CommTree);
```

Currently this code results in the following error:
```
*** An error occurred in MPI_Dist_graph_create_adjacent invalid sources
*** reported by process [3896508417,2]
*** on communicator MPI_COMM_WORLD
*** MPI_ERR_ARG: invalid argument of some other kind
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```

bosilca:
My understanding of the MPI standard description of the topology creation functions is that the list of neighbors is rank based, and should only contains the meaningful neighbors. Thus, you cannot have MPI_PROC_NULL as a neighbor in any type of topology.

omor1:
It *is* true that the distributed graph constructors specify `sources` and `destinations` as "array[s] of non-negative integers" describing "ranks of processes for which the calling process is a destination / source".

[...]

PR #4898, opened 2018-03-09
Allow MPI_PROC_NULL as neighbor.

bosilca:
Allowing MPI_PROC_NULL as a neighbor in any topology allows us to add
gaps on the send and recv buffers. This does make the traditional
neighbor collective have a similar behavior as the V version, but in
same time it allows the users to skip the step where they prepare the
counts and the displacement array.

For more info please take a look at issue #4675.


hjelmn:
@bosilca I don't think we should allow MPI_PROC_NULL for dist graph but it is clear the forum intended it to be allowed. This needs to be PR'd to v3.0.x and v3.1.x. See mpi-forum/mpi-issues#87

PR #4901, opened 2018-03-11
fcoll/two_phase: data sieving has to occur at offset 0 as well

edgargabriel:
data sieving has to occur for any offset provided that is larger
or equal zero for this implementation to work correctly.

(cherry picked from commit da640f98df993ce35811f41be52a878095526908)

### DIFF
--- a/ompi/mca/fcoll/two_phase/fcoll_two_phase_file_write_all.c
+++ b/ompi/mca/fcoll/two_phase/fcoll_two_phase_file_write_all.c
@@ -1053,7 +1053,7 @@ static int two_phase_exchage_data(mca_io_ompio_file_t *fh,
 
     if (nprocs_recv){
 	if (*hole){
-	    if (off > 0){
+	    if (off >= 0){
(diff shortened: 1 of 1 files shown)


jsquyres:
Per 2018-03-13 webex:

We talked about the series-of-3 PRs from Edgar/George today (for 2.x., 3.0.x, 3.1.x):

- George: datatype fix
- Edgar: ROMIO fix
- Edgar: OMPI fix (sieving thing)

All 3 are necessary to make the latest HDF5 test suite pass.

Edgar's 2 fixes are small, self-contained, low risk.  You merged the ROMIO fix on v3.0.x and v3.1.x yesterday.

The datatype fix seems higher risk.

We decided:

- Let's merge Edgar's 2 fixes across all of 2.x, 3.0.x, 3.1.x.  Simple, low risk. Fixes some things.
- Let's wait on the datatype fix (but not hold up releases).  Let's get more testing.  If people need these fixes, they can get them in the nightlies after the release.

Only after *all 3* fixes are in will the HDF5 test suite work.  But we don't think that's a blocker for all of our immanent releases.

Issue #4892, opened 2018-03-08
MPIR_PROCDESC is not visible in non-debug builds


jjhursey:
* Impacts `master`, `v3.0.x`, and `v3.1.x`. The `v2.x` and `v2.0.x` series do not have this problem.
 * Must be built with `--disable-debug`.

Consider the following `gdb` script that we will run against an `mpirun` launching an MPI application.

```shell
shell$ cat gdb-test-cmd.txt 
echo MPIR_debug_state\n
p MPIR_debug_state

echo MPIR_proctable_size\n
p MPIR_proctable_size

echo MPIR_proctable (pointer)\n
p MPIR_proctable

echo MPIR_proctable (type info)\n
ptype MPIR_proctable

echo MPIR_proctable (contents)\n
p *MPIR_proctable@MPIR_proctable_size
```

On `master` with a `--enable-debug` build you can see the expected:
```shell
shell$ gdb --batch -x gdb-test-cmd.txt --pid `ps aux | grep mpirun | grep -v grep | awk '{print $2}'`[New LWP 92532]
[New LWP 92531]
[New LWP 92530]
[New LWP 92529]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
0x00003fff953dd5d8 in poll () from /lib64/libc.so.6
MPIR_debug_state
$1 = 1
MPIR_proctable_size
$2 = 2
MPIR_proctable (pointer)
$3 = (struct MPIR_PROCDESC *) 0x100081b59d0
MPIR_proctable (type info)
type = struct MPIR_PROCDESC {
    char *host_name;
    char *executable_name;
    int pid;
} *
MPIR_proctable (contents)
$4 = {{
    host_name = 0x100081b6a40 "n01", 
    executable_name = 0x100081b45b0 "/home/jjhursey/./hello_c", 
    pid = 92533
  }, {
    host_name = 0x100081b7bd0 "n01", 
    executable_name = 0x100080107f0 "/home/jjhursey/./hello_c", 
    pid = 92534
  }}

```

On `master` with a `--disable-debug` build the `MPIR_PROCDESC` structure is obscured so we cannot read the `MPIR_proctable`
```shell
shell$ gdb --batch -x gdb-test-cmd.txt --pid `ps aux | grep mpirun | grep -v grep | awk '{print $2}'`[New LWP 92409]
[New LWP 92408]
[New LWP 92407]
[New LWP 92406]
[...]

PR #4904, opened 2018-03-13
Fix MPIR_proctable structure visibility

jjhursey:
* The `MPIR_PROCDESC` structure needs to be visible even in optimized
   builds so that debuggers can attach to `mpirun` and properly read the
   `MPIR_proctable`.
 * In the v2.0.x and v2.x series this structure resided in the `orterun`
   directory and included the `CFLAGS` fix included here. This code
   moved in the v3.x series and the `CFLAGS` did not move causing this
   issue.
   - Instead of applying the debug `CFLAGS` globally to libopen-rte,
     only apply them to the `orted_submit.c` compile which contains the
     MPIR symbols.


jjhursey:
Ref #4892

PR #4909, opened 2018-03-13
v2.x: Datatype fix: Update the loop stats.

jsquyres:
The loop should be updated on each internal iteration.

This is a manually-created patch for the v2.x branch, derived from
78480351955b8a27d44221090736df1716f3c384 on master.

Patch created by @bosilca in https://github.com/open-mpi/ompi/pull/4883#issuecomment-371607852.

Per discussion on the 2018-03-13 webex, putting this on the v2.1.4 milestone (because v2.1.3 is too close and we want this datatype change to be tested a bit more).


jsquyres:
Per 2018-03-12 webex, we decided to hold this PR until after the v2.1.3 release (just to let it cook / be tested a bit more).

hppritcha:
@edgargabriel please check when you have a chance

Issue #2530, opened 2016-12-06
osc rdma hang with multiple windows


markalle:
Using the testcase mt_1sided.c (and its support files 1sided.c mt_1sided_td1.c mt_1sided_td2.c) from this test harness pull request:
https://github.com/open-mpi/ompi-tests/pull/25
I get a hang from the osc rdma component:

```
% mpicc -o x mt_1sided.c mt_1sided_td1.c mt_1sided_td2.c
% mpirun -host hostA,hostB -mca osc rdma -mca pml ob1 -mca btl openib,self,vader ./x
```

This was from a vanilla build of openmpi-master-201612022109-0366f3a.

The single threaded "1sided.c" passes. And I think the key difference in the multi-threaded version is that there are two windows active (one for each thread).

1sided.c is a fairly broad test, looping over several synchronization types as well as contiguous and non-contiguous datatypes. I could probably whittle the test to be a bit more targeted for this particular hang if needed.

jsquyres:
@hjelmn Can you have a look?

rkowalewski:
@markalle Your link url does not work. We have probably the same issue regarding communication progress with multiple windows in our project. Could you please fix your url since I want to verify my assumption.

jsquyres:
@rkowalewski Our ompi-test repository is not public; that's why the link is broken for you.

@markalle Are you amenable to sharing your new tests with @rkowalewski?  You could post them in a gist, or attach them here on this issue.

rkowalewski:
@jsquyres @markalle I spent more time with this. Finally, I have pinpointed the error and created another issue: #2538 
Maybe there is some relation between these two issues.

markalle:
@jsquyres @rkowalewski Yeah, since we put it in the ompi-tests it's considered public-ish at least. Even the stuff I've checked in there is a whittled down version of our original tests, as giving out these smaller somewhat whittled down tests is easier to get management approval for.

markalle:
Here, see if I made this gist right:
https://gist.github.com/markalle/56f739c3c907e53960d22d5de2fe4bc9
It should have 1sided.c, mt_1sided.c, mt_1sided_td1.c and mt_1sided_td2.c.

jsquyres:
Perfect -- thanks @markalle!

hjelmn:
Please test with the latest 2.0.2 release candidate.

markalle:
[...]

PR #4918, opened 2018-03-15
osc/rdma: performance improvments and bug fixes

hjelmn:
This commit is a large update to the osc/rdma component. Included in
this commit:

 - Add support for using hardware atomics for fetch-and-op and single
   count accumulate  when using the accumulate lock. This will improve
   the performance of these operations even when not setting the
   single intrinsic info key.

 - Rework how large accumulates are done. They now block on the get
   operation to fix some bugs discovered by an IBM one-sided test. I
   may roll back some of the changes if the underlying bug in the
   original design is discovered. There appear to be no real
   difference (on the hardware this was tested with) in performance so
   its probably a non-issue. References #2530.

 - Add support for an additional lock-all algorithm: on-demand. The
   on-demand algorithm will attempt to acquire the peer lock when
   starting an RMA operation. The lock algorithm default has not
   changed. The algorithm can be selected by setting the
   osc_rdma_locking_mode MCA variable. The valid values are two_level
   and on_demand.

 - Make use of the btl_flush function if available. This can improve
   performance with some btls.

 - When using btl_flush do not keep track of the number of put
   operations. This reduces the number of atomic operations in the
   critical path.

 - Make the window buffers more friendly to multi-threaded
   applications. This was done by dropping support for multiple
   buffers per MPI window. I intend to re-add that support once the
   underlying performance bug under the old buffering scheme is
   fixed.

 - Fix a bug in request completion in the accumulate, get, and put
   paths. This also helps with #2530.

 - General code cleanup and fixes.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma.h
+++ b/ompi/mca/osc/rdma/osc_rdma.h
@@ -8,7 +8,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2007-2017 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2007-2018 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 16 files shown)


hjelmn:
opps, need to fix one small issue with the merge.

PR #4924, opened 2018-03-19
Sync to PMIx master PR pmix/pmix#697

karasevb:
This commit fixes the case when local client asks for the key from the
process on the remote node. The local server don't have commit count for
remote ranks, it is maintained by another PMIx server, so commit count
should be ignored for remote requests.

Refs https://github.com/pmix/pmix/pull/697


karasevb:
@rhc54 
We have failures with testing. Can we synchronize a particular commit directly?

karasevb:
@jladd-mlnx

rhc54:
Sure - so long as it is upstream in PMIx master, we won't lose it on the next update.

karasevb:
@rhc54 Thank you!

PR #4934, opened 2018-03-20
hwloc: bring back the configure error when external hwloc is >= 2.0

bgoglin:
@jsquyres Here's the patch to abort configure in 3.1 if hwloc external is 2.0.

@ggouaillardet added that code a while ago (it's still in 3.0.x, no need to backport there). Then he removed it from master/3.1.x when porting OMPI to the non-final hwloc 2.0 API (which isn't done yet). I am putting it back in 3.1.x but not in master.


jsquyres:
#4936 was accidentally a dup of this one.  So let's merge this one.

PR #4935, opened 2018-03-21
v3.1.x: Sync to PMIx v2.1 (PR pmix/pmix#698)

karasevb:
This commit fixes the case when local client asks for the key from the
process on the remote node. The local server don't have commit count for
remote ranks, it is maintained by another PMIx server, so commit count
should be ignored for remote requests.


karasevb:
Corresponds to https://github.com/pmix/pmix/pull/698

bwbarrett:
@rhc54, @karasevb is this 3.1.0 or 3.1.1?

PR #4936, opened 2018-03-21
v3.1.x: hwloc/external: disallow hwloc >=v2.0.0

jsquyres:

Turns out that this code already exists in the v3.0.x branch.

FYI: @bgoglin @rhc54

### DIFF
--- a/opal/mca/hwloc/external/configure.m4
+++ b/opal/mca/hwloc/external/configure.m4
@@ -184,6 +184,21 @@ AC_DEFUN([MCA_opal_hwloc_external_CONFIG],[
                [AC_MSG_RESULT([yes])],
                [AC_MSG_RESULT([no])
                 AC_MSG_ERROR([Cannot continue])])
+           AC_MSG_CHECKING([if external hwloc version is lower than 2.0])
+           AS_IF([test "$opal_hwloc_dir" != ""],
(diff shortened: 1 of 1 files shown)


bgoglin:
It's identical to my #4934 (except for one blank line) so I am going to assume it's OK :)

ggouaillardet:
note there is still a bunch of `#if HWLOC_API_VERSION < 0x20000` is `opal/mca/hwloc/base` and you might want to remove this dead code as well.

rhc54:
@ggouaillardet We don't want to remove the hwloc v2.0 integration code at this time. This patch is required because your "glue" was based on a pre-final version of hwloc v2.0, and there are changes which break it. So the plan is to first protect this release so someone using an external hwloc doesn't hit the problem, and then fix the integration to match the final hwloc v2 API.

ggouaillardet:
@rhc54 got it !

jsquyres:
@bgoglin Somehow I missed your #4934.  Doh!  Let's close this one and take yours, since you were first.

@rhc54 I think you mis-read the patch.  It emits `checking if external hwloc version is lower than 2.0` to stdout, and then emits `OMPI does not currently support hwloc v2 API / Cannot continue` if hwloc is too high.  The `#error` is inside the internal configure test.

PR #4938, opened 2018-03-21
fcoll/vulcan: add new fcoll component

edgargabriel:
import of the new vulcan component. It is an enhanced version
of the two_phase component, which uses however the ompio internal
codes/loops to assemble the data arrays. It is therefore more inline
with the dynamic and dynamic_gen2 component, and will be easier to
maintain.


edgargabriel:
:bot:retest

PR #4943, opened 2018-03-21
opal/util/net.c: fix compiler warning

jsquyres:
This variable is only used when IPv6 support is enabled; ensure to
protect it in an #if OPAL_ENABLE_IPv6 check.

This commit pulled out from #4199.


bosilca:
something is weird, I only see a copyright update.

jsquyres:
LOL.  I just rebased/cherry-picked, and the fix was otherwise applied to master long ago.

Closing without merging.

PR #4955, opened 2018-03-22
vader fix for 4937 (PPC wmb location)

### LINKED ISSUE #4937: v3.0.x vader on PPC (wmb moved to end of set_header)
This is for the v3.0.x branch.

I have some tests that fail with vader on PPC (pass on x86 due to more generous memory ordering rules there). It looks to me like one of the wmb calls has been moved. I don't have much knowledge of what vader's doing, but I'm guessing the use of the function mca_btl_vader_fbox_set_header() should boil down to
```
    set data
    wmb
    set header that says the data is there
```
but the fbox_set_header function has its wmb() call at the bottom so I think it's probably ending up as
```
    set data
    set header that says the data is there
    wmb
```
which wouldn't ensure the data is visible to the reader.

I can hit the problem using the below "maxsoak.c" testcase as
    mpicc -o x maxsoak.c
    mpirun -np 6 -mca pml ob1 -mca btl vader,self ./x
and the testcase will detect corruption.

For me the failure message from the testcase ends up something like
> 4: Invalid data: Act:525138 Exp:850 Peer:2 Datasize:32 Mult:50
I don't know the maxsoak.c testcase well, it's just something I know we didn't write so I don't have to go through any special approval process to share that code:
https://gist.github.com/markalle/a1c203297cb6af22a3fb5c24e62b2ba3

markalle:
I have some tests that failed on PPC with the recent vader changes.
I'm suspicious of the set_header() function where there's a wmb()
call that looks like it boils down to
    set some data
    set the header to indicate the data is available
    wmb
and I think the wmb needs to go up a line.

More details here:
    https://github.com/open-mpi/ompi/issues/4937
with a copy of the "maxsoak.c" testcase at
    https://gist.github.com/markalle/a1c203297cb6af22a3fb5c24e62b2ba3

Fixes #4937

### DIFF
--- a/opal/mca/btl/vader/btl_vader_fbox.h
+++ b/opal/mca/btl/vader/btl_vader_fbox.h
@@ -51,8 +51,8 @@ static inline void mca_btl_vader_fbox_set_header (mca_btl_vader_fbox_hdr_t *hdr,
                                                   uint16_t seq, uint32_t size)
 {
     mca_btl_vader_fbox_hdr_t tmp = {.data = {.tag = tag, .seq = seq, .size = size}};
-    hdr->ival = tmp.ival;
     opal_atomic_wmb ();
+    hdr->ival = tmp.ival;
 }
 
 /* attempt to reserve a contiguous segment from the remote ep */


hjelmn:
Has this been put in master yet?

hjelmn:
Can lead to incorrect answers. Clear bug.

markalle:
I don't think master has the bug. The function names are a bit different in that branch, but I think it's right in master (didn't hit any bugs on PPC anyway with master).

My change in this PR was partially based on just the logic behind what a function called set_header() is probably doing, but also partially based on comparing vs what master has.

hjelmn:
@markalle Ok. You are correct that this was the intended order. Its a typo that isn't caught on x86.

hjelmn:
@bwbarrett Is this something small enough that it can be included in v3.0.1 without another RC?

jsquyres:
Is this needed on v2.x or v3.1.x?

markalle:
I don't think it's needed anywhere else, it's a pretty recent regression. I browsed in branch "v3.1.x" to make sure and doesn't look like it needs to go there.

markalle:
Oh shoot, re-pushing with the commit signed

PR #4962, opened 2018-03-24
A bunch of CID fixes

jsquyres:
Fix a variety of low-impact Coverity issues.

### DIFF
--- a/ompi/mca/fcoll/static/fcoll_static_file_write_all.c
+++ b/ompi/mca/fcoll/static/fcoll_static_file_write_all.c
@@ -14,6 +14,7 @@
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2017      IBM Corporation. All rights reserved.
+ * Copyright (c) 2018      Cisco Systems, Inc.  All rights reserved
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -947,9 +948,7 @@ mca_fcoll_static_file_write_all (mca_io_ompio_file_t *fh,
 #endif
 
         }
-
-        if (my_aggregator == fh->f_rank) {
-        }    }
+    }
 
 #if OMPIO_FCOLL_WANT_TIME_BREAKDOWN
     end_exch = MPI_Wtime();
--- a/ompi/mca/fcoll/static/fcoll_static_file_write_all.c
+++ b/ompi/mca/fcoll/static/fcoll_static_file_write_all.c
@@ -975,12 +975,12 @@ exit:
         decoded_iov = NULL;
     }
 
-    if (my_aggregator == fh->f_rank) {
+    if (NULL != local_iov_array){
+        free(local_iov_array);
+        local_iov_array = NULL;
+    }
 
-        if (NULL != local_iov_array){
-            free(local_iov_array);
-            local_iov_array = NULL;
-        }
+    if (my_aggregator == fh->f_rank) {
         for(l=0;l<fh->f_procs_per_group;l++){
             if (NULL != blocklen_per_process[l]){
                 free(blocklen_per_process[l]);
--- a/ompi/mca/topo/treematch/treematch/tm_tree.c
+++ b/ompi/mca/topo/treematch/treematch/tm_tree.c
@@ -132,8 +132,8 @@ void free_non_constraint_tree(tm_tree_t *tree)
 {
   int d = tree->dumb;
 
-  free_list_child(tree);
   free_tab_child(tree);
+  free_list_child(tree);
   if(!d)
     FREE(tree);
 }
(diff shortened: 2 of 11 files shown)


opal/mca/btl/tcp/btl_tcp_proc.c
```diff
@@ -452,6 +452,11 @@ int mca_btl_tcp_proc_insert( mca_btl_tcp_proc_t* btl_proc,
     max_peer_interfaces = proc_data->max_local_interfaces;
     peer_interfaces = (mca_btl_tcp_interface_t**)calloc( max_peer_interfaces, sizeof(mca_btl_tcp_interface_t*) );
     assert(NULL != peer_interfaces);
+    if (NULL == peer_interfaces) {
```
ggouaillardet: shouldn't you remove the `assert()` ?

ompi/mca/topo/treematch/treematch/tm_tree.c
```diff
@@ -132,8 +132,8 @@ void free_non_constraint_tree(tm_tree_t *tree)
 {
   int d = tree->dumb;
 
-  free_list_child(tree);
   free_tab_child(tree);
+  free_list_child(tree);
```
ggouaillardet: @bosilca can you please make sure this is pushed upstream ?

opal/mca/btl/tcp/btl_tcp_proc.c
jsquyres: I thought about removing the `assert()` and at the time, decided against it.  But it's probably 60/40 better to remove it and just rely on the returning of `OPAL_ERR_OUT_OF_RESOURCE`.  I'll update.

ompi/communicator/comm_cid.c
```diff
@@ -1026,7 +1026,7 @@ static int ompi_comm_allreduce_group_broadcast (ompi_comm_request_t *request)
 {
     ompi_comm_allreduce_context_t *context = (ompi_comm_allreduce_context_t *) request->context;
     ompi_comm_cid_context_t *cid_context = context->cid_context;
-    ompi_request_t *subreq[2];
```
bosilca: This is not necessary as subreq is a temporary array and the requests inside are copied using the subreq_count. Moreover, from a technical perspective these are supposed to be set to MPI_REQUEST_NULL and not to 0. This comment apply everywhere where such array is used.

ompi/mca/fcoll/static/fcoll_static_file_write_all.c
```diff
@@ -976,12 +975,12 @@ mca_fcoll_static_file_write_all (mca_io_ompio_file_t *fh,
...
+    if (NULL != local_iov_array){
+        free(local_iov_array);
+        local_iov_array = NULL;
+    }
 
-        if (NULL != local_iov_array){
-            free(local_iov_array);
-            local_iov_array = NULL;
-        }
+    if (my_aggregator == fh->f_rank) {
```
bosilca: Reading through this function hurts ;)

ompi/mca/topo/treematch/treematch/tm_tree.c
bosilca: Yep, I forwarded to the treematch maintainers.

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -1183,8 +1183,8 @@ static int mca_btl_tcp_component_exchange(void)
                  if ((AF_INET == my_ss.ss_family) &&
                      (4 != mca_btl_tcp_component.tcp_disable_family)) {
                      memcpy(&addrs[current_addr].addr_inet,
-                             &((struct sockaddr_in*)&my_ss)->sin_addr,
-                             sizeof(addrs[0].addr_inet));
+                            &((struct sockaddr_in*)&my_ss)->sin_addr,
+                            sizeof(((struct sockaddr_in*)&my_ss)->sin_addr));
```
bosilca: How is that not the same ?

ompi/mca/fcoll/static/fcoll_static_file_write_all.c
edgargabriel: @bosilca don't worry too much about this function. The static component might be axed before the branch to 4.0

opal/mca/btl/tcp/btl_tcp_component.c
jsquyres: sizeof(addrs[0].addr_inet) == 16.
sizeof(((struct sockaddr_in*)&my_ss)->sin_addr) == 4

bosilca: Right. How about sizeof(struct in_addr) to make out intentions clear.

jsquyres: I admit that I kinda like the style of `memcpy(dest, &src, sizeof(src))`.  Do you think that's not clear?

bosilca: In other places we use "sizeof(struct in_addr)", which I think is a good indicator that we are dealing with an IPv4 address.

As a special mention, the original code was correct, we are moving 16 bytes around, but only the sin_addr part is initialized.

jsquyres: Updated to `sizeof(struct in_addr)`.

ompi/communicator/comm_cid.c
jsquyres: Ok, removed the 2 commits with these comm_cid fixes.

PR #4968, opened 2018-03-26
opal_free_list: use lifo atomic functions in opal_free_list_wait_mt

hjelmn:
This commit fixes a multi-threading bug when using the thread-safe
free list functions. opal_free_list_wait_mt() was using the
conditional version of opal_lifo_pop() and not the thread-safe call.

(cherry picked from commit 7f761d843498d96ebda408be03778d68ead4b8d9)


nysal:
The patch looks good. I'm curious what bug were you hitting. The conditional code should have called the right function, although it's less efficient.

PR #4980, opened 2018-03-28
ompi/oshmem: fix bug in shmem_finalize.

xinzhao3:
Originally there is a "oshmem_shmem_globalexit_status" variable to track if current shmem_finalize is the implicit one at end of main, and shmem_finalize does real work when it is, otherwise it does nothing.
This is not correct, if shmem_finalize is the explicit one in the program, it should still guarantee the completion of pending communications, according to specification 1.3 page 10. 

This PR deleted "oshmem_shmem_globalexit_status" variable and let shmem_finalize always does real work.

Signed-off-by: xinz@mellanox.com <root@ibm-p9-005.mtx.labs.mlnx>

### DIFF
--- a/oshmem/shmem/c/shmem_finalize.c
+++ b/oshmem/shmem/c/shmem_finalize.c
@@ -22,15 +22,9 @@
 #include "oshmem/shmem/c/profile/defines.h"
 #endif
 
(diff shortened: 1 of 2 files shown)


ibm-ompi:

PR #4983, opened 2018-03-28
Always set opal_cache_line_size in the app

ggouaillardet:


rhc54:
I'm not convinced that is the best thing to do, but I don't really care. However, I am not sure the second commit here is correct, especially in the schizo/ompi component, so please don't do that until some more thought can be given to it as you might break (for example) the ability of the pmix reference server to support an OMPI job (since the two OPAL_MCA_PREFIX values won't match).

orte/mca/ess/base/ess_base_fns.c
```diff
@@ -43,6 +43,13 @@
 
 #include "orte/mca/ess/base/base.h"
 
+static void set_cache_line_size() {
+    char * env;
+    if (NULL != (env=getenv(OPAL_MCA_PREFIX"opal_cache_line_size"))) {
```
jsquyres: Should that be `OPAL_MCA_PREFIX "_opal_cache_line_size"`?

orte/mca/schizo/ompi/schizo_ompi.c
```diff
@@ -979,6 +980,10 @@ static int setup_fork(orte_job_t *jdata,
         opal_setenv("OMPI_MCA_orte_bound_at_launch", "1", true, &app->env);
     }
 
+    char cache_line_size[16];
+    snprintf(cache_line_size, 16, "%d", opal_cache_line_size);
+    opal_setenv("OMPI_MCA_opal_cache_line_size", cache_line_size, true, &app->env);
```
jsquyres: If you use `OPAL_MCA_PREFIX` to read the env variable, you should use it to set the variable, too.

rhc54:
Looked at it a bit more, and we do indeed have a problem here that bears some thought. Please don't merge this right now until we can think thru how we operate with the PMIx ref server. The ```OPAL_MCA_PREFIX``` stuff was all well-and-good when everything ran within its own project - maybe not so good when two projects need to interact.

jsquyres:
bot:ompi:retest

bwbarrett:
bot:ompi:retest

AboorvaDevarajan:

awlauria:
Since orte is gone, should we just close this?

jsquyres:
I think there are conflicts in ORTE files (because they're now gone from master), but there's a lot of non-ORTE stuff on this PR.

awlauria:
It's 2 files worth of changes that I can see: `pmix_cray.c` and `info.c`. Unless those files just got moved somewhere else?

Also it looks like `pmix_cray.c` no longer exists on master.

rhc54:
Correct - none of the files altered in this PR exist any more. Question is: does something like this need to go into the OPAL or OMPI area for when a process is direct launched?

jsquyres:
My mistake; I just looked closer, and this PR looks 100% stale / no longer valid.

PR #4987, opened 2018-03-28
fcoll/dynamic_gen2: use hindexed constructor on the sender side

raafatfeki:
instead of using a temporary buffer and copy data into the temp buffer before sending, use a derived datatype to describe the data that needs to be sent during a cycle in the collective I/O operation.

Signed-off-by: raafatfeki <fekiraafat@gmail.com>


jsquyres:
ok to test

PR #4989, opened 2018-03-29
v3.0.x: Update to PMIx 2.0.3

jjhursey:


rhc54:
whistle...wow, that's a big amount of change, dude! Are the RMs going to accept this?

opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c
```diff
@@ -824,68 +824,99 @@ pmix_status_t pmix_server_unpublish(pmix_peer_t *peer,
...
+    if (PMIX_SUCCESS != (rc = pmix_bfrop.unpack(buf, &cd->ninfo, &cnt, PMIX_SIZE))) {
         PMIX_ERROR_LOG(rc);
         return rc;
     }
     /* unpack the array of directives */
-    if (0 < ninfo) {
-        PMIX_INFO_CREATE(info, ninfo);
-        cnt=ninfo;
-        if (PMIX_SUCCESS != (rc = pmix_bfrop.unpack(buf, info, &cnt, PMIX_INFO))) {
+    if (0 < cd->ninfo) {
```
rhc54: @jjhursey I recvd a request that we always provide the PMIX_PARENT_ID and PMIX_SPAWNED attributes when a client proc (but not a tool) dynamically starts another job. See the code now in master - might be an idea to backport it here for consistency.

jjhursey:
They are using PMIx 2.0.1rc1 right now. If they want cross version support in v3.0.1 then it should be considered. I'll leave it to the RMs.
It's worth considering, thus the PR. But it might be wise to push to v3.0.2 (with a quick turn around) so that it gets some testing.

rhc54:
@hppritcha @jsquyres Just to clarify (since I commented about the size of this PR) - much of the changes consist of mods to copyrights and removal of some files. Of the remaining changes, they fall into three categories:

1. bug fixes that impact correctness and cross-version support
1. better isolation of the PMI-1/2 backward compatibility libraries, which OMPI doesn't build
1. changes to non-executing files such as the rpm spec file

The first category is the smallest of the group, and so I'd recommend moving ahead with it (precise timing is up to you).

opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c
rhc54: NOTE: my "changes request" relates solely to the above comment. Otherwise, things look fine to me, so I'll change my review to "approved" as the above is optional and up to @jjhursey .

bwbarrett:
After some discussion, since Open MPI 3.0.1 is all but released (all that's left is the web page update), we're going to release 3.0.1 without this patch and ship it with 3.0.2, targetted for 1 month out.

PR #4998, opened 2018-03-31
dist: Don't build case-conflicting html man pages

bwbarrett:
The html man page builder creates a web page for every
man page generated by our releases.  We then check all
those pages into the ompi-www repo.  Force the build
to skip case-conflicting pages (ie, mpiCC because it
conflicts with mpicc), even if building on Linux, so that
ompi-www continues to be sane when cloned onto MacOS.

(cherry picked from commit fe2b6cf1d63d352aca247c9d026bb7038cb412c6)


jsquyres:
bot:ompi:retest

bwbarrett:
bot:ompi:retest

Issue #5000, opened 2018-04-01
Open MPI Fails to Build Under OS X v10.11.6 'El Capitan' Due to the Absence of `javah` in Java 10


RandomDSdevel:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The version of Open MPI that I was attempting to build from source and install was v3.0.1 (I currently have v3.0.0 installed.)  

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I initially installed Open MPI via [the Homebrew package manager for OS X/macOS](https://brew.sh/) using the following command:  

```
HOMEBREW_BUILD_FROM_SOURCE=1 brew install -vd --build-from-source open-mpi --with-java --with-mpi-thread-multiple
```

I keep Open MPI up to date by running the following command when upgrading only it:  

```
HOMEBREW_BUILD_FROM_SOURCE=1 brew upgrade -vd --build-from-source open-mpi --with-java --with-mpi-thread-multiple
```

(Please note, however, that setting `HOMEBREW_BUILD_FROM_SOURCE=1` is not standard practice for users of Homebrew, so individuals attempting to reproduce this issue should omit it if they either already have all Homebrew dependencies for Open MPI installed from source or binary or don't mind having Homebrew install 'bottles' pre-built using the project's CI.)  

### Please describe the system on which you are running

* Operating system/version:  Apple OS X v10.11.6 'El Capitan,' build 15G20015
* Computer hardware:  Apple Mid-2007 Aluminum iMac (`iMac7,1`) (More details available on request.)  
* Network type:  Cable broadband over local wireless (More details available on request.)  

-----------------------------

## Details of the problem

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Open MPI fails to build its Java bindings from source under Homebrew on systems which only have Java 10 installed due to how its build system looks for and subsequently cannot find `javah`, which was deprecated in Java 9 in removed entirely in Java 10.  I also reported this problem downstream to Homebrew as Homebrew/homebrew-core#26009 before opening this issue here.

RandomDSdevel:
[...]

PR #5001, opened 2018-04-02
    configury: use javac vs javah whenever possible.

ggouaillardet:
javah has been removed from Java 10, so use javac -h instead
(that is already usable in Java 8)

Refs. open-mpi/ompi#5000

### DIFF
--- a/config/opal_setup_java.m4
+++ b/config/opal_setup_java.m4
@@ -15,7 +15,7 @@ dnl                         reserved.
 dnl Copyright (c) 2007-2012 Oracle and/or its affiliates.  All rights reserved.
 dnl Copyright (c) 2008-2013 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2013      Intel, Inc.  All rights reserved.
-dnl Copyright (c) 2015      Research Organization for Information Science
+dnl Copyright (c) 2015-2018 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl Copyright (c) 2017      FUJITSU LIMITED.  All rights reserved.
 dnl $COPYRIGHT$
@@ -37,7 +37,7 @@ AC_DEFUN([OPAL_SETUP_JAVA_BANNER],[
 AC_DEFUN([OPAL_SETUP_JAVA],[
     AC_REQUIRE([OPAL_SETUP_JAVA_BANNER])
 
-    OPAL_VAR_SCOPE_PUSH([opal_java_bad opal_java_found opal_java_dir opal_java_jnih opal_java_PATH_save opal_java_CPPFLAGS_save])
+    OPAL_VAR_SCOPE_PUSH([opal_java_bad opal_javah_happy opal_java_found opal_java_dir opal_java_jnih opal_java_PATH_save opal_java_CPPFLAGS_save])
     AC_ARG_ENABLE(java,
                   AC_HELP_STRING([--enable-java],
                                  [Enable Java-based support in the system - use this option to disable all Java-based compiler tests (default: enabled)]))
@@ -161,13 +161,13 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
             AS_IF([test -n "$with_jdk_bindir" && test "$with_jdk_bindir" != "yes" && test "$with_jdk_bindir" != "no"],
                   [PATH="$with_jdk_bindir:$PATH"])
             AC_PATH_PROG(JAVAC, javac)
-            AC_PATH_PROG(JAVAH, javah)
             AC_PATH_PROG(JAR, jar)
             AC_PATH_PROG(JAVADOC, javadoc)
+            AC_PATH_PROG(JAVAH, javah)
             PATH=$opal_java_PATH_save
 
-            # Check to see if we have all 4 programs.
-            AS_IF([test -z "$JAVAC" || test -z "$JAVAH" || test -z "$JAR" || test -z "$JAVADOC"],
+            # Check to see if we have all 3 programs.
+            AS_IF([test -z "$JAVAC" || test -z "$JAR" || test -z "$JAVADOC"],
                   [opal_java_happy=no
                    HAVE_JAVA_SUPPORT=0],
                   [opal_java_happy=yes
@@ -178,6 +178,21 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
                   [opal_java_CPPFLAGS_save=$CPPFLAGS
                    # silence a stupid Mac warning
                    CPPFLAGS="$CPPFLAGS -DTARGET_RT_MAC_CFM=0"
+                   AC_MSG_CHECKING([javac -h])
+                   cat > Conftest.java << EOF
+public final class Conftest {
+    public native void conftest();
+}
+EOF
+                   AS_IF([$JAVAC -d . -h . Conftest.java > /dev/null 2>&1],
+                         [AC_MSG_RESULT([yes])],
+                         [AC_MSG_RESULT([no])
+                          AS_IF([test -n "$JAVAH"],
+                                [opal_javah_happy=yes],
+                                [opal_java_happy=no])])
+                   rm -f Conftest.java Conftest.class Conftest.h
+
+
                    AS_IF([test -n "$with_jdk_headers" && test "$with_jdk_headers" != "yes" && test "$with_jdk_headers" != "no"],
                          [OPAL_JDK_CPPFLAGS="-I$with_jdk_headers"
                           # Some flavors of JDK also require -I<blah>/linux.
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@hppritcha @bwbarrett FYI, this PR might break JDK older than 8.

kawashima-fj:
@hppritcha @bwbarrett @jsquyres This change is required to support recently-released OpenJDK 10. But it causes a `make`-time error with OpenJDK 7 and older. [JDK 7 was released in 2011](http://openjdk.java.net/projects/jdk7/). This problem appears only in the non-default `--enable-mpi-java` build. Supported JDK versions are described in `README.JAVA.txt` Can I merge it to master (and release branches)?

hppritcha:
looks good to me.  jenkins 32bit builds still not happy but that has nothing to do with this PR.

ibm-ompi:

jjhursey:
bot:ibm:gnu:retest

bwbarrett:
bot:ompi:retest

hppritcha:
@kawashima-fj have you tried to build the java bindings with java 7?  I vaguely recall that some more recent additions to the bindings prevent compiling with java7.  Can you check this?

kawashima-fj:
@hppritcha Yes, I tried it.

Without this PR:

- OpenJDK 7 on CentOS 6.9: can build and run

With this PR:

- OpenJDK 7 on CentOS 6.9: *cannot* build Open MPI (`make`-time error)
- OpenJDK 8 on CentOS 7.4: can build and run
- OpenJDK 9 on CentOS 7.4: can build and run
- OpenJDK 10 on CentOS 7.4: can build and run

I confirmed the following things on x86_64.

- Open MPI can be built with `--enable-mpi-java`
- `examples/Ring.java` can be built by `mpijavac`
- `examples/Ring.java` can be run by `mpiexec`

A possible related issue is #3705/#4121. I will recall it...

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;As I mentioned [here](https://www.mail-archive.com/devel@lists.open-mpi.org//msg20551.html) on the developer mailing list, I'll rebase an attempt to fix how Open mPI's build process handles finding JDK tools on OS X/macOS that I'm working on (still only locally at the moment, but it will become a PR soon) against this after it's merged.

ggouaillardet:
let me know if we want to support JDK < 8.
In this case, I will also check `javah -version` is working and use some macros to handle the legacy case

bwbarrett:
For 3.0 and 3.1, we definitely want to keep supporting Java 7, because otherwise we'd have a fairly major backwards compatibility problem.  I think 4.0 (aka master) is an open question.

ggouaillardet:
@bwbarrett I reworked this PR so tries `javah` first, and then fallback to `javac -h`.
Not ideal imho, but much simpler than doing it the other way around.

kawashima-fj:
@ggouaillardet If you have JDK 8 in `/usr/bin` and JDK 10 in another directory specified in `--with-jdk-dir`, Open MPI `make` fails with `Could not find class file for 'mpi.MPI'` error.

In this case, `$JDK10/bin/javac` is set in `JAVAC` and `/usr/bin/javah` is set in `JAVAH`. JDK 10 `javac` generates `*.class`. JDK 8 `javah` tries to generates `*.h` from `*.class` but it cannot because JDK 8 cannot read `*.class` generated by JDK 10 (class file version incompatibility).

ggouaillardet:
Good point, I will think about a solution.

ggouaillardet:
@kawashima-fj it should be good now
 - first, try `javac -h`
 - if fails, then ensure `javah` is available

kawashima-fj:
@ggouaillardet I confirmed *almost* successful build & run with OpenJDK 7/8/9/10 and `javah` is used only with OpenJDK 7. But I sometimes see an error when I use parallel make.

```
$ cd build/ompi/mpi/java/java
$ make -j32
  JAVAC    mpi_MPI.h
  JAVAC    mpi_Comm.h
  JAVAC    mpi_Count.h
  JAVAC    mpi_CartParms.h
  JAVAC    mpi_Datatype.h
  JAVAC    mpi_CartComm.h
  JAVAC    mpi_Constant.h
  JAVAC    mpi_Errhandler.h
  JAVAC    mpi_File.h
  JAVAC    mpi_GraphComm.h
  JAVAC    mpi_GraphParms.h
  JAVAC    mpi_Group.h
  JAVAC    mpi_Info.h
  JAVAC    mpi_Op.h
  JAVAC    mpi_Intercomm.h
  JAVAC    mpi_Intracomm.h
  JAVAC    mpi_Message.h
  JAVAC    mpi_Prequest.h
  JAVAC    mpi_ShiftParms.h
  JAVAC    mpi_Request.h
  JAVAC    mpi_Status.h
  JAVAC    mpi_Version.h
JAVADOC    doc
  JAVAC    mpi_Win.h
../../../../../ompi/mpi/java/java/Intracomm.java:134: error: cannot access Info
        @Override public Intracomm dupWithInfo(Info info) throws MPIException
                                               ^
  bad class file: ./mpi/Info.class
    class file contains wrong class: mpi.MPIException
    Please remove or make sure it appears in the correct subdirectory of the classpath.
1 error
make: *** [mpi_Intracomm.h] Error 1
make: *** Waiting for unfinished jobs....
```

I finally find that `javac -h . -d . Foo.java` creates `Foo.h` *and* `Foo.class` at the same time and `javac -h . -d . Bar.java` reads `Foo.class` if `Bar.java` uses `Foo` class. It causes a race. One possible solution will be adding `-h .` to the rule for `mpi/MPI.class` and delete the rule for `$(JAVA_H)` if `OPAL_HAVE_JAVAH_SUPPORT` is true.

ggouaillardet:
@kawashima-fj thanks for the insight ! I suspected such an issue might occur, but was unable to evidence it. I will fully review the process and update the PR

ggouaillardet:
@kawashima-fj we should be good now, can you please give this PR a final review ?

ggouaillardet:
:bot:mellanox:retest

kawashima-fj:
@ggouaillardet I'll review and test later today or tomorrow.

kawashima-fj:
@ggouaillardet I reviewed and tested it on OpenJDK 7/8/9/10. Perfect.

For our record:
This change affects Open MPI only when the non-default `--enable-mpi-java` option (Open MPI Java binding) is specified to `configure`.
Without this change, building Open MPI with OpenJDK 10 (and future versions) fails because OpenJDK 10 dropped `javah` command.
With this change, all recent OpenJDK versions (at least 7, 8, 9, and 10) are supported.
I don't know other vendor's JDK. But if Open MPI can be built without this change, it can also be built with this change.
The combination of OpenJDK 10 and OS X/macOS has still the issue #5000. The PR #5015 is proposed for it.

bosilca:
This broke OMPI on OSX systems without Java SDK. The configure exits with the following:
```
*** Java compiler
Unable to find any JVMs matching version "(null)".
No Java runtime present, try --request to install.
checking OS X/macOS locations... found ()
configure: WARNING: No recognized directory structure found under
```

RandomDSdevel:
@bosilca:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Are you sure that _this_ PR is to blame?  That looks like the part of the build infrastructure that goes through the logic _I_ changed in #5015.  I'm not sure how what I did could have caused this to break, though, at least at first glance; I'm pretty sure I didn't touch anything having to do with whether the Java bindings are enabled or not, and I would suspect that those _always_ required you to have a JDK installed…

bosilca:
I am not 100% sure this PR was to be blamed, but I thought it was the last thing that touched the java support.

If I don't want java bindings, and especially if I did not requested them specifically, why should I have an SDK installed ?

jsquyres:
George is right -- we do not require a Java SDK to build/run OMPI.  Regardless of whether it was this PR or #5001, both of them touched the Java configury recently, and we apparently got it wrong.  😦

RandomDSdevel:
@jsquyres:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Well, back to the drawing board it is, then.  Upon further investigation, I believe my inserting [this line](https://github.com/open-mpi/ompi/blob/master/config/opal_setup_java.m4#L117) into that file as a blind copy of the suggestion that you requested near the bottom of the code sketch you put in list item 2 of [this review comment of yours](https://github.com/open-mpi/ompi/pull/5015#discussion_r184112902) is to blame; one of us should have noticed that this would cause this part of the build scripting to abort and thus make the fallback logic below unreachable.  Removing that line should fix the resulting problem that @bosilca reported; I'll file a new PR to do just that against `master` shortly, and then that can get backported to the other branches (unless you want me to muck with Git history, although _that_…doesn't sound like a very good idea.)

RandomDSdevel:
@jsquyres:  Submitted as #5160.

jsquyres:
@RandomDSdevel Perfect -- thanks!  (and correct: mucking with public git history in a large multi-developer project -- just say no!)

PR #5003, opened 2018-04-02
ompi/oshmem: fix shmem_free to perform no-op on null ptr

sam6258:
According to the shmem specification, performing a free on a NULL pointer should have no action. However, because we are checking to see if the pointer value is within a shmem region, we hit an assertion when freeing a NULL. I've added an explicit check to see if the pointer is NULL, so that shmem_free is a no-op on a NULL pointer.

### DIFF
--- a/oshmem/shmem/c/shmem_free.c
+++ b/oshmem/shmem/c/shmem_free.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2013-2015 Mellanox Technologies, Inc.
  *                         All rights reserved.
+ * Copyright (c) 2018      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


oshmem/shmem/c/shmem_free.c
```diff
@@ -41,7 +42,12 @@ static inline void _shfree(void* ptr)
...
 
-    RUNTIME_CHECK_INIT(); RUNTIME_CHECK_ADDR(ptr);
+    RUNTIME_CHECK_INIT();
+    if (NULL == ptr) {
+        return OSHMEM_SUCCESS;
+    }
+
+    RUNTIME_CHECK_ADDR(ptr);
 
 #if OSHMEM_SPEC_COMPAT == 1
```
sam6258: Not sure if this compatibility check is needed before returning success on shmem_free(nil). Any input on that?

bwbarrett:
bot:ompi:retest

bwbarrett:
https://jenkins.open-mpi.org/jenkins/job/open-mpi.pull_request/2368/ is a real failure.  you're returning a value from a function with a void return type.

sam6258:
Ah, I see that. This compiled with a different compiler and probably just threw a warning I didn't see. Will make that change. Thanks for pointing that out.

oshmem/shmem/c/shmem_free.c
gpaulsen: I think it's okay, but you're right we don't want null pointers going through RUNTIME_CHECK_ADDR().

PR #5010, opened 2018-04-04
Move past the '.' when getting jobstepid

### LINKED ISSUE #5008: Multiple active srun invocations fail in an allocation on PSM system

v2.1.3 and v 3.0.1

github releases
./configure --with-slurm --with-psm --prefix=$PWD/install --with-slurm --with-pmi

* Operating system/version: RHEL7
* Computer hardware: x86_64
* Network type: PSM

-----------------------------

## Details of the problem
When running multiple srun invocations within an allocation on a PSM system, I noticed that the first invocation would run my application OK, but trying to simultaneously run another job would fail with these errors:

catalyst285: Unable to allocate shared memory for intra-node messaging.
catalyst285: Delete stale shared memory files in /dev/shm

After doing some debugging, I noticed that both srun invocations were trying to create a file with the same name in /dev/shm, i.e.:

bash-4.2$ ls -l /dev/shm/psm*
-rwx------ 1 lee218 lee218 6352896 Apr  4 08:12 /dev/shm/psm_shm.1a000000-ed6e-0000-1a00-00001a000000

On PSM 2 systems, the same OpenMPI installation generates unique hashes:

bash-4.2$ ls -l /dev/shm/psm2_shm.5264903ffffff00bd0*
-rw------- 1 lee218 lee218 4337664 Apr  4 08:12 /dev/shm/psm2_shm.5264903ffffff00bd03010
-rw------- 1 lee218 lee218 4337664 Apr  4 08:12 /dev/shm/psm2_shm.5264903ffffff00bd04010

I was able to track this issue down to https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/s2/pmix_s2.c#L233. The jobid and stepid are used on PSM systems to generate the hash (I believe some other mechanism is used on PSM2 systems). The problem with this attempt to add the stepid to the jobid is that the stepid will always return 0. When run under srun, the pmix_kvs_name is of the form "jobid.jobstep". The first strtoul will set str to point to the "." in that string and thus calling strtoul(str,...) will always return 0 since the "." character is not a digit. I believe that the correct thing to do would be to look at the following character (i.e., call strtoul(str+1,...)). This is probably a problem in https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/s1/pmix_s1.c#L226 too.

I will also note that when running on the PSM system using orterun instead of srun, a unique hash is generated for multiple orterun invocations:

bash-4.2$ ls -l /dev/shm/psm*
-rwx------ 1 lee218 lee218 12566528 Apr  4 08:28 /dev/shm/psm_shm.452df885-6aec-13fc-52f9-2dc5ff846503
-rwx------ 1 lee218 lee218 12566528 Apr  4 08:28 /dev/shm/psm_shm.ba118e7f-7761-16d9-8a95-6f1e3de17946

rhc54:
The strtoul function returns the pointer to the first non-digit character, which is a '.' in this case. Calling strtoul at that point will always yield a zero - you have to move past it to get the remaining number

Thanks to Greg Lee for the detailed analysis of the problem.

Fixes #5008 

### DIFF
--- a/opal/mca/pmix/s1/pmix_s1.c
+++ b/opal/mca/pmix/s1/pmix_s1.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
+ * Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016 Cisco Systems, Inc.  All rights reserved.
@@ -223,7 +223,7 @@ static int s1_init(opal_list_t *ilist)
     s1_pname.jobid = strtoul(pmix_id, &str, 10);
     s1_pname.jobid = (s1_pname.jobid << 16) & 0xffff0000;
     if (NULL != str) {
-        ui32 = strtoul(str, NULL, 10);
+        ui32 = strtoul(str+1, NULL, 10);
         s1_pname.jobid |= (ui32 & 0x0000ffff);
     }
     ldr.jobid = s1_pname.jobid;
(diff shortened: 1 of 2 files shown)


rhc54:
@jsquyres @bwbarrett Someone will need to PR this to your release branches.

lee218llnl:
looks good to me

jsquyres:
@rhc54 Done -- v2.x, v3.0.x, and v3.1.x PR's ref'ed this PR.

PR #5026, opened 2018-04-05
v3.1.x: MPI_Comm_spawn_multiple.3in: update Fortran string array notes

jsquyres:
Per 0ab6b201fed, note in the MPI_Comm_spawn_multiple.3in man page that
the array_of_commands does not need to be terminated -- it just need
to have exactly "count" entries.  In the Fortran binding, at least,
this is different than in prior released versions of Open MPI (it's
not a backwards incompatibility, since prior versions of Open MPI
required array_of_commands to be blank-string-terminated in Fortran --
this change makes Open MPI be *less* restrictive, and therefore still
backwards compatible).

(cherry picked from commit fc8ebbb0e09b83bfa19ad68dff4f07cb21c619cf)


jsquyres:
@ggouaillardet Yeah, I'm aware.  But you had already approved those PRs (and github doesn't remove the approval if commits are added) -- I didn't want to add commits to that which was already approved.

PR #5035, opened 2018-04-06
v2.x: jobstepid parsing fix

jsquyres:
The strtoul function returns the pointer to the first non-digit character, which is a '.' in this case. Calling strtoul at that point will always yield a zero - you have to move past it to get the remaining number

Thanks to Greg Lee for the detailed analysis of the problem.

(cherry picked from commit cd52ccdb6833ac5e7300d9d96d0ddd4bf9d4dce2)

Master PR: #5010


jsquyres:
@hppritcha Good to go.

PR #5044, opened 2018-04-09
orterun.1in: fix typo

jsquyres:
Found via https://github.com/open-mpi/ompi-www/pull/61.

### DIFF
--- a/orte/tools/orterun/orterun.1in
+++ b/orte/tools/orterun/orterun.1in
@@ -1,5 +1,5 @@
 .\" -*- nroff -*-
-.\" Copyright (c) 2009-2016 Cisco Systems, Inc.  All rights reserved.
+.\" Copyright (c) 2009-2018 Cisco Systems, Inc.  All rights reserved.
 .\" Copyright (c) 2008-2009 Sun Microsystems, Inc.  All rights reserved.
 .\" Copyright (c) 2017-2018 Intel, Inc. All rights reserved.
 .\" Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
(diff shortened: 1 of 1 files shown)


jsquyres:
False failure in PRBC (github failure).

BTW, I have some insight on false failures that look exactly like what just happened in the OMPI Jenkins -- we have run seemingly-similar issues with Cisco's internal github.

The issue is that sometimes github is a little slow creating the internal test merge git hash.  I.e., sometimes the JSON blob that you get from either the initial github webhook or even if you make a followup API call to get the PR JSON blob, it may have either the `merge_commit_sha` field as `null` or a stale git hash (e.g., if someone previously created the PR and has updated it).  Specifically: the `merge_commit_sha` field should only be considered valid if the `mergeable` field is *also* `True`.  If `mergeable` is `null`, then the `merge_commit_sha` is invalid (and potentially stale).

Meaning: you must wait for `mergeable` to become `True` before trying to use the value of `merge_commit_sha`.  If you simply wait for `merge_commit_sha` to be non-`null`, you can get errors like what just happened in Jenkins:

```
...
Command "git checkout -f ef5cdb903bceb114542a4c5ec9091aa9e2661da1" returned status code 128:
stdout: 
stderr: fatal: reference is not a tree: ef5cdb903bceb114542a4c5ec9091aa9e2661da1
```

I don't know exactly what the Jenkins plugin is doing -- e.g., if it's checking `mergeable` or not.  But I thought I'd share this hard-won knowledge.

FWIW: we've never seen `mergeable` be `True` and the `merge_commit_sha` be invalid.  I suppose that's also possible (given the distributed nature of how github works behind the scenes), but that would kinda seem to defeat the point of having `mergeable`.

jsquyres:
bot:ompi:retest

jsquyres:
FWIW, Brian and I looked at the jenkins failure in a bit more detail.

It *looks* like github sent `merge_commit_sha = ef5cdb903bceb114542a4c5ec9091aa9e2661da1` and `mergeable = True` (!) in the first JSON blob.  Which apparently is a lie, because jenkins failed to find that SHA.  Then when I did the retest, Jenkins got a *different* `merge_commit_sha` value from Github (7b4286da00e29ddbdc91edf6f4cde7257f7fac23), which worked fine.

Smells like a github bug, actually.  Shrug.

jsquyres:
This is not a problem in the v2.x branch -- removed that label.

PR #5049, opened 2018-04-10
pmix/base: set PMIx to follow OPAL's mca_component_show_load_errors

jsquyres:
Have Open MPI's PMIx component to set PMIx's "show_load_errors" to do
the same thing that Open MPI's "show_load_errors" does.

### DIFF
--- a/opal/mca/pmix/base/pmix_base_select.c
+++ b/opal/mca/pmix/base/pmix_base_select.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
- * Copyright (c) 2016 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2016-2018 Cisco Systems, Inc.  All rights reserved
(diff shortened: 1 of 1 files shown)


opal/mca/pmix/base/pmix_base_select.c
```diff
@@ -29,6 +29,16 @@ int opal_pmix_base_select(void)
     opal_pmix_base_component_t *best_component = NULL;
     opal_pmix_base_module_t *best_module = NULL;
 
+    // Set the PMIx component_show_load_errors MCA param based on the
+    // value of OPAL's mca_base_component_show_load_errors (i.e., the
+    // bool variable behind Open MPI's mca_component_show_load_errors
+    // MCA param).
+    char *pmix_show_load_errors_env = NULL;
+    asprintf(&pmix_show_load_errors_env,
+             "PMIX_MCA_mca_component_show_load_errors=%d",
```
jjhursey: Should that be `PMIX_MCA_mca_base_component_show_load_errors` I think the `base` is needed because without it there is a deprecated warning, right?

jsquyres: Fixed.

PR #5034, opened 2018-04-06
configure: remove POWER 7/BE block

jsquyres:
Do not merge until #4563 is fixed.

Fixes #5032.

This will need to be cherry-picked to v2.x, v3.0.x, and v3.1.x.  Let's hope this his master before we branch for v4.0.x, but if it doesn't, it will need to be merged there, too.

Note that there's 2 commits on this PR: one that removes the block (and is easily cherry-picked) and another that contains the NEWS update (that almost certainly will not cherry-pick to the release branches properly).

Issue #4349, opened 2017-10-17
Why is Big-endian powerpc no longer supported ? 


amckinstry:
Sorry about the "non-bug" nature of this issue, but I'm working on OpenMPI 3.0.0 for Debian, and see BE powerpc is no longer supported.
Given that other powerpc and other BE systems _are_ supported, I'm curious about this, and wondering whats going on.

gpaulsen:
Thanks for asking your question.  I'm sorry that dropping support for Big-endian powerpc has affected you.  The community makes every effort to support as wide of a selection of platforms as possible, with our limited volunteered resources.

  During the development of Open MPI v3.0.0, during our Open MPI supported platforms discussion, we discussed who might be willing to work on and support Open MPI on Big-Endian Power PC platforms.  IBM committed to support the ppc64le (Little Endian) platform on Linux, however all IBM roadmaps only have Little Endian products.  Due to perceived lack of interest, IBM chose not to support Big Endian Platforms.

  Another aspect to the removal of support is due to our new improved continuous integration testing automation.  Whenever anyone creates a pull request to either master branch or a release branch, the automation automatically begins a build and test of that pull request on all supported platforms (existing across various different member organizations).  This automation has drastically improved quality to the point where we now define our support statement for Open MPI such that we only support platforms that we have added to this infrastructure and test regularly.

  The Open MPI community is interested in supporting as many platforms as practical, and would be interested in working with anyone who is interested in supporting a platform, provided we can obtain a testing platform, and a contact person or organization.

   If you are still interested, please reply, and we can continue discussing.

amckinstry:
Thanks for this reply.

My concern is that, from a Debian Linux perspective, we support and test multiple architectures that are not part of the CI infrastructure upstream (see https://buildd.debian.org/status/package.php?p=openmpi); we then run our own tests on these.

PPC on Big-endian systems is arguably one of the better supported of the lesser archs. It seems strange that we will support alpha, hppa, sparc64, etc (without the latest fabrics, of course) but not powerpc.

This testing on other archs has two main benefits beyond supporting the tiny number of users doing HPC on minority archs. It shakes out the bugs latent in the codebase, and then in doing so, it makes future ports to new archs (arm64, etc) possible as developers know the codebase actually supports the standards rather than having many assumptions ("All the worlds a Vax", as they used to say).

Can we downgrade the failure to build on BE PPC from an hard error to  a warning - "WARNING: this architecture is not tested and officially supported" ?

[...]

PR #5051, opened 2018-04-10
v3.0.x: remove power7be configure block

jsquyres:
Per #5034 and larger discussion on #4349.

Still waiting for #4563 to be fixed (to resolve hangs on some ARM/POWER platforms), but that's a different issue.


jsquyres:
@jjhursey No; that blurb already exists here in the v3.0.x README.

PR #5055, opened 2018-04-10
plm/base: fix typo in variable name

hjelmn:
An incorrectly named variable caused all pml variables to disappear
from ompi_info. This commit fixes the typo. We may add some logic into
the MCA base to catch these sorts of things in the future.


hjelmn:
@ggouaillardet This crazy bug caused pml variables to disappear from ompi_info :)

hjelmn:
@bwbarrett This affects v3.1.x. Will get a PR open as soon as this is in master.

PR #5057, opened 2018-04-11
Quick test to make sure builder still works

bwbarrett:


bwbarrett:
maybe the even older freebsd ami?  if only these were config files in git.

bot:ompi:retest

PR #5060, opened 2018-04-12
configure: Fix typo in `configure --help`

kawashima-fj:
This affects only output of `configure --help`.
v3.1.x branch has same typo.

```
  --enable--wrapper-runpath
                          enable runpath in the wrapper compilers if linker
                          supports it (default: enabled, unless wrapper-rpath
                          is disabled).
```


bwbarrett:
FreeBSD 11 was having problems last night due to a bad patch being applied to the AMI; retesting

bot:ompi:retest

Issue #4801, opened 2018-02-09
btl-portals4 segfaults when running osu_latency


tkordenbrock:
On the OMPI users mailing list, Brian Larkins reported a segfault in btl-portals4 when running osu_latency.

The problem is a series of add_procs()/del_procs() calls that cause btl-portals4 to free top level Portals4 resources.  Any additional add_procs() calls will trigger the segfault.

Brian reported this problem using v3.0.0, but it also exists on master and v3.1.x.

awlauria:
PR's merged. Closing.

PR #5062, opened 2018-04-12
v3.0.x: btl-portals4: don't free module resources when proc count goes to zero

tkordenbrock:
This PR fixes a segfault in btl-portals4 add_procs(). The segfault
occurs if add_procs() is called after a del_procs() call that reduces
the proc count to zero which would cause PT and NI resources to be
freed. This commit resolves the segfault by using a common
initiailization boolean and only freeing module resources in
finalize().

Reported in #4801.

Signed-off-by: Todd Kordenbrock (thkgcode@gmail.com)
(cherry picked from commit b569633ddfdfcebc06bdf2f3acfb36d88b61334a)


hppritcha:
@bwbarrett good to go

Issue #5009, opened 2018-04-04
io/romio314 does not work with minloc/maxloc datatypes


hjelmn:
Thank you for taking the time to submit an issue!

## Background information

Now, I don't know if these should work at all. I really want these datatypes killed in MPI-4.0.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

master

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Built from source.

### Please describe the system on which you are running

* Operating system/version: macOS 10.13
* Computer hardware: MacBookPro14,3
* Network type: Shared memory

-----------------------------

## Details of the problem

When using the romio314 component the MPI_File_write* functions fail because they call ```MPI_Type_get_contents()``` on our non-contiguous predefined datatypes (MPI_DOUBLE_INT, MPI_SHORT_INT, etc). If these types are allowed for files (I REALLY HOPE NOT) then this is a bug.

Reproducer:

```c
#include <stdlib.h>
#include <stdio.h>

#include <mpi.h>

struct we_should {
  double kill;
  int minmax;
};

int main (int argc, char *argv[]) {
  MPI_File mfh;
  struct we_should foo = {.kill = 1.0, .minmax = 1};

  MPI_Init (NULL, NULL);

  MPI_File_open (MPI_COMM_WORLD, "/tmp/typetest.out", MPI_MODE_CREATE | MPI_MODE_WRONLY,
                 MPI_INFO_NULL, &mfh);

  MPI_File_write_all (mfh, &foo, 1, MPI_DOUBLE_INT, MPI_STATUS_IGNORE);
  MPI_File_close (&mfh);


  MPI_Finalize ();
  return 0;
}
```

Expected result: success

Actual result:

```
[mordor:83551] *** An error occurred in MPI_Type_get_contents
[mordor:83551] *** reported by process [3613327361,0]
[mordor:83551] *** on communicator MPI_COMM_WORLD
[mordor:83551] *** MPI_ERR_INTERN: internal error
[mordor:83551] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[mordor:83551] ***    and potentially your MPI job)
```

hjelmn:
@bosilca @roblatham00 @edgargabriel Thoughts. This works fine with ompio.

edgargabriel:
[...]

PR #5068, opened 2018-04-13
io/romio314: fix minmax datatypes

hjelmn:
romio assumes that all predefined datatypes are contiguous. Because of
the (terribly named) composed datatypes MPI_SHORT_INT, MPI_DOUBLE_INT,
MPI_LONG_INT, etc this is an incorrect assumption. The simplest way to
fix this is to override the MPI_Type_get_envelope and
MPI_Type_get_contents calls with calls that will work on these
datatypes. Note that not all calls to these MPI functions are
replaced, only the ones used when flattening a non-contiguous
datatype.

References #5009


hjelmn:
I will go ahead and commit this. @roblatham00 Please let me know when you fix this upstream and I will apply that fix to ompi.

PR #5077, opened 2018-04-16
fortran: remove intent(in) from asynchronous arguments for accumulate

hjelmn:
The asynchronous and intent(in) keywords are not compatible
(*&#$). Remove the intent(in).

Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

PR #5079, opened 2018-04-17
status_set_cancelled: fix F08 binding

jsquyres:
The F08 bindings for MPI_STATUS_SET_CANCELLED incorrectly had the
"flag" dummy parameter set as INTENT(OUT) when it really should be
INTENT(IN).

On the one hand, this is technically an ABI change.  On the other
hand, this is an incorrect MPI binding.  On the other hand (that's 3
hands for you fans counting at home), this is such a rarely-used API
-- even in the C bindings -- that I'm guessing no one is using this
API, and therefore no one has noticed this error and it isn't worth
porting back to the release branches.

I found this while reviewing #5077, so @hjelmn gets to review this one.  😄

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-interfaces.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-interfaces.F90
@@ -1,6 +1,6 @@
 ! -*- f90 -*-
 !
-! Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
+! Copyright (c) 2009-2018 Cisco Systems, Inc.  All rights reserved
 ! Copyright (c) 2009-2015 Los Alamos National Security, LLC.
 !                         All rights reserved.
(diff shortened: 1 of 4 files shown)


jsquyres:
Per https://lists.mpi-forum.org/pipermail/mpiwg-fortran/2018-April/003545.html, there is actually some ambiguity here.  It seems like the *right* thing to do is obvious (i.e., it should be `IN`).  But MPI-3.1 specifically says that the F08 binding should be `OUT` (even though the language neutral and the C bindings are both clearly `IN`).  Hrm.

jsquyres:
I filed https://github.com/mpi-forum/mpi-issues/issues/91 and https://github.com/mpi-forum/mpi-standard/pull/36.  This should be read as an errata item in the June 2018 MPI Forum meeting in Austin, TX.  Let's see what happens.

hppritcha:
This errata item was passed at the last MPI Forum.  Time to merge?

hjelmn:
I say yes.

Issue #5048, opened 2018-04-10
FORCE-TERMINATE AT base/plm_base_launch_support.c


angainor:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
  Open MPI repo revision: v3.1.0rc3

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
      Configure command line:'--with-knem=/cluster/software/hpcx/2.0/knem' '--with-mxm=/cluster/software/hpcx/2.0/mxm' '--with-hcoll=/cluster/software/hpcx/2.0/hcoll' '--with-ucx=/cluster/software/hpcx/2.0/ucx' '--with-platform=contrib/platform/mellanox/optimized' '--with-pmix=/usr' '--with-hwloc=/usr' '--with-libevent=/usr'

      C compiler: gcc
      C compiler version: 7.2.0
      pmix-1.2.3 
      hwloc-libs-1.11.2
      libevent-2.0.21

### Please describe the system on which you are running

* Operating system/version: CentOS Linux release 7.4.1708 (Core) 

* Computer hardware: Intel(R) Xeon(R) CPU E5-2683 v4 @ 2.10GHz

* Network type: Infiniband controller: Mellanox Technologies MT27700 Family [ConnectX-4]

-----------------------------

## Details of the problem
`mpirun` seems to have problems starting workers. I run `mpirun ls`  from within a SLURM allocation: 
 
```shell
shell$ mpirun ls
[c11-1:139811] PMIX ERROR: BAD-PARAM in file src/dstore/pmix_esh.c at line 1185
[c11-1:139811] [[5046,0],0] ORTE_ERROR_LOG: Not found in file base/odls_base_default_fns.c at line 172
[c11-1:139811] [[5046,0],0] ORTE_ERROR_LOG: Not found in file base/plm_base_launch_support.c at line 550
--------------------------------------------------------------------------
An internal error has occurred in ORTE:

[[5046,0],0] FORCE-TERMINATE AT (null):1 - error base/plm_base_launch_support.c(551)

This is something that should be reported to the developers.
--------------------------------------------------------------------------

```
The above works for v3.0.1. Note that I'm compiling against a locally installed PMIx 1.2.3. Is this the problem?

ggouaillardet:
note you should (that is unlikely to cause the issue though)
```
[...]

PR #5081, opened 2018-04-17
Fix support for external PMIx v2.0 with legacy PMIx_Get()

ggouaillardet:


jjhursey:
Ref Issue #5048
I would suggest getting this into v3.1.0 if the window is still open.

artpol84:
@ggouaillardet Can you explain more about the problem?
We are using OMPI v3.1 with PMIx 1.2.5 without problems.

artpol84:
Ok, I didn't noticed the link to the issue, will go through.

artpol84:
Probably for us it was Slurm direct launch and we were not using PLM component.
@ggouaillardet @kawashima-fj @jjhursey is it correct to say that if using direct launch this problem is not exist?

hjelmn:
@ggouaillardet This doesn't seem to quite fix the issue. I patched master with the parts of this that are for some reason not in master and it still fails with a not found error. This appears to be happening because the daemon namespace is unknown when storing locally. I can get it to run on a single node by making pmix1_store_local always return success but that probably isn't the right fix.

ggouaillardet:
@hjelmn i fixed the title which was misleading.

This fix is necessary for PMIx < v2.1, and though it is sufficient for PMIx v2.0, some more changes are needed for PMIx v1.2.
This is discussed in #5048 

Open MPI v3.0.x does things differently and hence support PMIx v1.2

> I patched master with the parts of this that are for some reason not in master and it still fails with a not found error.

As far as I am concerned, `master` and `v3.1.x` with this PR do things similar, so I am puzzled with your previous statement. Did it apply to `v3.1.x` or `v3.0.x` ?
In the latter case, some code was removed from `master`, and that was backported to `v3.1.x` but not `v3.0.x`. In the former case, can you please share your patch ?

hjelmn:
@ggouaillardet I guess my patch just pulled in what was removed from master. The only other change was to always return OPAL_SUCCESS. That somehow worked. Probably because it was on a single node.

From what I can tell the only thing that isn't working with PMIx v1.2.3-v1.2.5 (after the changes you have already made to master) is storing the daemon uri. This is due to the non-existent namespace for the daemon jobid.

hjelmn:
@bwbarrett This plus the fix headed to master in #5103 should be all we need to get v3.1.0 out the door.

PR #5088, opened 2018-04-24
orte_dt_print_node: correctly handle nodes with no alias(es)

ggouaillardet:
Thanks Erico for reporting this issue

### DIFF
--- a/orte/runtime/data_type_support/orte_dt_print_fns.c
+++ b/orte/runtime/data_type_support/orte_dt_print_fns.c
@@ -14,6 +14,8 @@
  * Copyright (c) 2011-2013 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


ibm-ompi:

rhc54:
Ummm...something doesn't look right here. The only way you can enter that code block is if the ```ORTE_NODE_ALIAS``` was set, which only occurs if there was indeed an alias. Thus, there is no way that the "alias" variable should be NULL. I would instead suspect either an error somewhere in the code that sets the alias, or someplace where the alias is being incorrectly free'd.

ggouaillardet:
@rhc54 note the `ORTE_NODE_ALIAS` is not `NULL` but is set to an empty string (e.g. `""`). `NULL` occurs when we `opal_argv_split("", ',')`. But I get your point, this is likely something we should have avoided in the first place and I will dig that.

Thanks !

ggouaillardet:
@rhc54 i fixed that in #5095.
Could you please review it before I merge and PR to all the release branches ?

Issue #5067, opened 2018-04-12
without-lustre build issue


mdidomenico4:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

3.0.1

### Please describe the system on which you are running
* Operating system/version: rhel 7.4 
* Computer hardware: x86_64

## Details of the problem

./configure --prefix=/openmpi --disable-oshmem --without-slurm --without-lustre

make -j 4 && make install
...snipped...
CCLD libmca_fs_lustre.la
../../../../libtool: line 7489: cd: no/lib: No such file or directory
libtool: error: cannot determine absolute directory name of 'no/lib'

jsquyres:
Do you have a space in your source or build tree directory name?

Open MPI's configure/build system probably does not handle spaces well.

mdidomenico4:
there are no spaces

the resulting Makefiles seemed to have this added

fs_lustre_LDFLAGS = -Lno/lib
OMPI_PKG_CONFIG_LDFLAGS = -Lno/lib -Wl,-rpath -Wl,no/lib -Wl,-rpath -Wl,${libdir} -Wl,--enable-new-dtags
same for OMPI_WRAPPER_EXTRA_FC_LDFLAGS and OMPI_WRAPPER_EXTRA_LDFLAGS

jsquyres:
Oh, I see -- it added `$with_luster`, which is `no`.  Got it.  Yeah, that's a bug.

jsquyres:
This was fixed on master long ago (and automatically became part of the v4.0.x branch).  But the fixes were not cherry picked to the release branches.

Just created PRs for v3.0.x and v3.1.x.

jsquyres:
This has now been merged to all relevant release branches.  Closing.

PR #5093, opened 2018-04-24
Topic/io configury

edgargabriel:
@jsquyres would you mind reviewing this pr?

There are two things really in here:
1. fix for issue #5067 to handle --without-lustre correctly in our configure script (and same for the pvfs2 file system).
2. provide a configure level summary for the file systems detected/configured


config/ompi_check_lustre.m4
```diff
@@ -82,7 +85,7 @@ OPAL_LOG_COMMAND(
 )
     rm -f conftest.c conftest.o
     AC_MSG_RESULT([$ompi_check_lustre_struct_happy])
-
+])
```
jsquyres: I think that this is indented at the wrong level -- it confusingly makes it look like you are terminating the `OMPI_CHECK_LUSTRE` macro rather than the `AS_IF`.

PR #5096, opened 2018-04-25
orte: only set the ORTE_NODE_ALIAS attribute when required

ggouaillardet:
When there is no alias for a given node, do not set the
ORTE_NODE_ALIAS attribute to an empty string any more.

Thanks Erico for reporting this issue.
Thanks Ralph for the guidance.

(back-ported from commit open-mpi/ompi@a05456ab5ed4f415798b202abf251e28afa5c397)


jsquyres:
See https://github.com/open-mpi/ompi/pull/5097#pullrequestreview-115209435.

PR #5097, opened 2018-04-25
orte: only set the ORTE_NODE_ALIAS attribute when required

ggouaillardet:
When there is no alias for a given node, do not set the
ORTE_NODE_ALIAS attribute to an empty string any more.

Thanks Erico for reporting this issue.
Thanks Ralph for the guidance.

(back-ported from commit open-mpi/ompi@a05456ab5ed4f415798b202abf251e28afa5c397)

### DIFF
--- a/orte/mca/ess/hnp/ess_hnp_module.c
+++ b/orte/mca/ess/hnp/ess_hnp_module.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2011-2017 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
- * Copyright (c) 2017      Research Organization for Information Science
+ * Copyright (c) 2017-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
@@ -462,12 +462,14 @@ static int rte_init(void)
     if (orte_retain_aliases) {
         aliases = NULL;
         opal_ifgetaliases(&aliases);
-        /* add our own local name to it */
-        opal_argv_append_nosize(&aliases, orte_process_info.nodename);
-        aptr = opal_argv_join(aliases, ',');
+        if (0 < opal_argv_count(aliases)) {
+            /* add our own local name to it */
+            opal_argv_append_nosize(&aliases, orte_process_info.nodename);
+            aptr = opal_argv_join(aliases, ',');
+            orte_set_attribute(&node->attributes, ORTE_NODE_ALIAS, ORTE_ATTR_LOCAL, aptr, OPAL_STRING);
+            free(aptr);
+        }
         opal_argv_free(aliases);
-        orte_set_attribute(&node->attributes, ORTE_NODE_ALIAS, ORTE_ATTR_LOCAL, aptr, OPAL_STRING);
-        free(aptr);
     }
     /* record that the daemon job is running */
     jdata->num_procs = 1;
--- a/orte/mca/ras/base/ras_base_node.c
+++ b/orte/mca/ras/base/ras_base_node.c
@@ -12,7 +12,7 @@
  * Copyright (c) 2011-2017 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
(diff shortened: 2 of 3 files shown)


orte/mca/ess/hnp/ess_hnp_module.c
```diff
@@ -462,12 +462,14 @@ static int rte_init(void)
...
-        /* add our own local name to it */
-        opal_argv_append_nosize(&aliases, orte_process_info.nodename);
-        aptr = opal_argv_join(aliases, ',');
+        if (0 < opal_argv_count(aliases)) {
+            /* add our own local name to it */
+            opal_argv_append_nosize(&aliases, orte_process_info.nodename);
+            aptr = opal_argv_join(aliases, ',');
+            orte_set_attribute(&node->attributes, ORTE_NODE_ALIAS, ORTE_ATTR_LOCAL, aptr, OPAL_STRING);
+            free(aptr);
+        }
```
jsquyres: I'm a little confused by this logic:

* If there are aliases, then set `ORTE_NODE_ALIAS` to a comma-delimited list of aliases *plus* our local nodename.
* If there are *no* aliases, then do not set `ORTE_NODE_ALIAS` at all -- not even to our local nodename.

Is the behavior w.r.t. the local nodename intended?  I.e., is the local nodename only relevant if there are aliases?

orte/mca/ras/base/ras_base_node.c
```diff
@@ -157,10 +157,12 @@ int orte_ras_base_node_insert(opal_list_t* nodes, orte_job_t *jdata)
                     opal_argv_free(nalias);
                 }
                 /* and store the result */
-                ptr = opal_argv_join(alias, ',');
+                if (0 < opal_argv_count(alias)) {
+                    ptr = opal_argv_join(alias, ',');
+                    orte_set_attribute(&hnp_node->attributes, ORTE_NODE_ALIAS, ORTE_ATTR_LOCAL, ptr, OPAL_STRING);
+                    free(ptr);
+                }
```
jsquyres: Why split the original value that you just obtained from `orte_get_attribute(..., ORTE_NODE_ALIAS, ...)`?  If you get a non-NULL/non-empty value back, why not just directly `orte_set_attribute(...)`?

I ask because you're not saving the value you get back from `orte_get_attribute()` -- you're getting the value from `orte_get_attribute()`, transforming it, then transforming it back, and then using the double-transformed value to call `orte_set_attribute()`.  Am I missing something?

orte/mca/ess/hnp/ess_hnp_module.c
rhc54: yes - that is the intended behavior. Note that this also only happens if ```orte_retain_aliases``` is set. Otherwise, we just ignore the aliases.

orte/mca/ras/base/ras_base_node.c
rhc54: We have to add the new alias to the end of the existing comma-delimited list, so we have to first fetch what we may already have, split it to get an argv array, then add only if unique, recombine and re-store. If he gets nothing back from the original "get", then he still has to store the new one.

Note that the code block 151-158 should be removed - looks like a copy/paste error.

orte/mca/ess/hnp/ess_hnp_module.c
jsquyres: Ok.  👍

orte/mca/ras/base/ras_base_node.c
jsquyres: Ok. 👍

ggouaillardet: @rhc54 at line 143, we get the alias(es) for `&hnp_node->attributes` but at line 151, we get the alias(es) for `&node->attributes`, so are you sure this block should be removed ?

note I will add a missing `free(ptr)` at line 153

also, what is the rationale for testing `ptr != NULL` at line 144 when we check the return of `orte_get_attribute()` everywhere else.

rhc54: @ggouaillardet You are right about the code block - I forgot that "node" may be carrying its own list of aliases.

Technically, it would probably be better to check the return code instead of checking ptr for NULL. Ultimately, it doesn't matter as ptr will obviously remain NULL if the attribute isn't found.

PR #5103, opened 2018-04-25
Fix external PMIx v1.2.5 support

rhc54:
As @hjelmn and I discussed, this is a little hacky. However, it is the only solution that can be done solely from the OMPI side.


rhc54:
@hjelmn I believe this fixes the problem we discussed - at least, it worked when I tested it here. If it works for you, I leave it to you to setup PRs as required.

hjelmn:
@bwbarrett This should get v3.1.0 back on track.

Issue #5085, opened 2018-04-20
mpi_sizeof missing from Fortran module mpi


ThemosTsikas:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

openmpi-v3.0.x-201804200242-e327deb

built with GCC 6.4.1 and NAG Fortran Compiler 6.2 (Build 6210)

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

from overnight tarball

### Please describe the system on which you are running

* Operating system/version: 
Fedora Release 25, Linux 4.13.16-100.fc25.x86_64, glibc-2.24-10.fc25.x86_64

* Computer hardware: 
x86_64, Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz

* Network type: 
single-node machine

-----------------------------

## Details of the problem

There is no symbol mpi_sizeof in the mpi module.

```
Use mpi, Only: mpi_sizeof
```
results in compile-time error.

jsquyres:
`mpi_sizeof` should be there if your compiler supports it.  And I'd be very surprised if the NAG compiler doesn't support what Open MPI needs.

Can you submit configure stdout and config.log so that we can see what happened?

ggouaillardet:
@ThemosTsikas could you please also upload your `ompi/mpi/fortran/use-mpi-tkr/mpi-tkr-sizeof.h` and `ompi/mpi/fortran/use-mpi-tkr/mpi-tkr-sizeof.f90` files ?

jsquyres:
A few more questions...

It looks like Open MPI decided to build the sizeof interfaces -- I see `mpi-tkr-sizeof.f90`.  But I *don't* see `mpi-tkr-sizeof.h` -- which it should have built if it decided to build the sizeof interfaces.

That being said, I'm not overly surprised that you're having a problem -- we've probably not strenuously tested the TKR interfaces in quite a while, and only with older gfortran that doesn't support enough to do the sizeof interfaces.

I see from your configure stdout:

```
checking Fortran compiler ignore TKR syntax... not cached; checking variants
checking for Fortran compiler support of TYPE(*), DIMENSION(*)... no
checking for Fortran compiler support of !GCC$ ATTRIBUTES NO_ARG_CHECK... no
[...]

PR #5104, opened 2018-04-25
fortran/tkr: Fix Makefile.am warnings

jsquyres:
Set mpi.lo dependencies outside of AM conditionals.  Per
https://github.com/open-mpi/ompi/issues/5085, make mpi.lo depend on
whatever we decide the sizeof source files are (which may be empty, if
this compiler does not support the Right Stuff for MPI_SIZEOF, or it
may be mpi-tkr-sizeof.[h|f90]).

This is a follow-on to #5085.  It fixes the autogen warnings that @rhc54 cited in https://github.com/open-mpi/ompi/issues/5085#issuecomment-384333070.  If this is this correct solution, it should be added to the existing PRs for the #5085 function on the v2.x, v3.0.x, and v3.1.x PRs.


jsquyres:
@ggouaillardet Can you add this to your existing TKR PRs?

PR #5108, opened 2018-04-26
Fix rank-by option and improve npernode/skt

rhc54:
This fixes a problem reported by @bgoglin where rank-by was incorrectly generating values when ranking by a type of object (e.g., socket). It also corrects the handling of the pernode, npernode, and npersocket options - these should only set the #procs and the default mapping pattern. They specifically should not prohibit the user from requesting a different mapping.

Thus, the following should be valid:

mpirun -npernode 2 --map-by socket ...

should put 2 procs on each node, mapping them by-socket on each node.

Fixes https://github.com/pmix/prrte/pull/46


bgoglin:
Thanks, all combinations of socket/core/numa ranking/binding seem to work here (tested on dual-socket E5v3 with cluster-on-die = 2 packages with 2 NUMA each with 6 cores each).

PR #5109, opened 2018-04-26
fortran/tkr: Fix Makefile.am warnings

ggouaillardet:
Set mpi.lo dependencies outside of AM conditionals.  Per
https://github.com/open-mpi/ompi/issues/5085, make mpi.lo depend on
whatever we decide the sizeof source files are (which may be empty, if
this compiler does not support the Right Stuff for MPI_SIZEOF, or it
may be mpi-tkr-sizeof.[h|f90]).

(cherry picked from commit open-mpi/ompi@5bd02f6649ac40b43094846799799e32f0a88b4a)


jsquyres:
@hppritcha Good to go.

hppritcha:
NEWS check #5085

PR #5113, opened 2018-04-27
Use standard-conforming dummy names for Fortran bindings

ThemosTsikas:

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
+++ b/ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
@@ -7,8 +7,8 @@
 !                         of Tennessee Research Foundation.  All rights
 !                         reserved.
 ! Copyright (c) 2012      Inria.  All rights reserved.
-! Copyright (c) 2015-2017 Research Organization for Information Science
-!                         and Technology (RIST). All rights reserved.
+! Copyright (c) 2015-2018 Research Organization for Information Science
+!                         and Technology (RIST).  All rights reserved.
 ! $COPYRIGHT$
 !
 ! This file provides the interface specifications for the MPI Fortran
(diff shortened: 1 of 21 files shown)


jsquyres:
ok to test

jsquyres:
Thank you, @ThemosTsikas!

jsquyres:
Actually, I just pushed an update to this PR.  @ggouaillardet: can you review?

I included the corresponding TKR PMPI updates, and also the corresponding MPI/PMPI updates for the ignore-TKR module.  I also updated the commit message to be a bit more descriptive, because the majority of the Open MPI dev community are not wise in the ways of Fortran.

jsquyres:
Oh, and @ThemosTsikas can you verify that I didn't break anything for you?

ThemosTsikas:
Instead of your first paragraph, I would say:

Fortrtan allows calls to procedures to be written with keyword arguments, e.g.
           Call mpi_sizeof(x=x,size=rsize,ierror=ier)
An "explicit interface" for the procedure must be in scope for this to be allowed in a program unit. Therefore, the explicit interface blocks we provide must match the MPI published standard, including the names of the dummy variables, as that is how Fortran programs may call them.

ggouaillardet:
@jsquyres I will fix that right away, sorry for that unfortunate mistake.

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
```diff
@@ -655,10 +655,10 @@ subroutine ompi_type_create_subarray_f(ndims,array_of_sizes, &
    INTEGER, INTENT(OUT) :: ierror
 end subroutine ompi_type_create_subarray_f
 
-subroutine ompi_type_dup_f(type,newtype,ierror) &
+subroutine ompi_type_dup_f(oldtype,newtype,ierror) &
```
jsquyres: FWIW, these `[p]ompi_*_f` interfaces don't matter -- they're the behind-the-scenes interfaces that are not user-facing.  It doesn't hurt to update them as well, but it also doesn't matter.

PR #5116, opened 2018-04-30
ucx: improve error messages during connection establishment

yosefe:


yosefe:
@alinask @karasevb pls take a look

yosefe:
tested with mpi_test_suite on 16nodes, 28ppn, with direct modex enabled

PR #5131, opened 2018-05-02
v2.x: mpi/fortran: use conformant dummy names for Fortran bindings

jsquyres:
The MPI spec defines that the "mpi" and "mpi_f08" module Fortran
bindings support passing by parameters by name.  Hence, we need to use
the MPI-spec-defined parameter names ("dummy variables", in Fortran
parlance) for the "mpi" and "mpi_f08" modules.

Specifically, Fortran allows calls to procedures to be written with
keyword arguments, e.g., "call mpi_sizeof(x=x,size=rsize,ierror=ier)"
An "explicit interface" for the procedure must be in scope for this to
be allowed in a Fortran program unit.  Therefore, the explicit
interface blocks we provide in the "mpi" and "mpi_f08" modules must
match the MPI published standard, including the names of the dummy
variables (i.e., parameter names), as that is how Fortran programs may
call them.

Note that we didn't find this issue previously because even though the
MPI spec *allows* for name-based parameter passing, not many people
actually use it.  I suspect that we might have some more incorrect
parameter names -- we should probably do a full "mpi" / "mpi_f08"
module parameter name audit someday.

Thanks to Themos Tsikas for reporting the issue and supplying the
initial fix.

(cherry picked from commit 4d126c16fa82c64a9a4184bc77e967a502684f02)

@hppritcha This has NEWS on it because we changed the Fortran interface parameter names. This is an ABI-breaking change, but honestly, no one could have been using these names before (because they were wrong).

@ThemosTsikas FYI

### DIFF
--- a/ompi/mpi/fortran/base/attr-fn-int-callback-interfaces.h
+++ b/ompi/mpi/fortran/base/attr-fn-int-callback-interfaces.h
@@ -4,8 +4,8 @@
 ! Copyright (c) 2006-2014 Cisco Systems, Inc.  All rights reserved.
 ! Copyright (c) 2013      Los Alamos National Security, LLC. All rights
 !                         reserved.
-! Copyright (c) 2015      Research Organization for Information Science
-!                         and Technology (RIST). All rights reserved.
+! Copyright (c) 2015-2018 Research Organization for Information Science
+!                         and Technology (RIST).  All rights reserved.
 ! $COPYRIGHT$
 !
 ! Additional copyrights may follow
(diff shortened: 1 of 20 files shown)


ggouaillardet:
@jsquyres In my view, the MPI forum would ideally provide `mpi.h` and Fortran modules (interfaces only) so we could automate conformance to the standard instead of manually checking our code vs a pdf file.

Any idea on why the MPI forum only provides a pdf file ?

jsquyres:
@ggouaillardet Probably because the Forum has only ever provided a document -- not code...?  Keep in mind that there's wide latitude in what needs to be in `mpif.h`...

PR #5134, opened 2018-05-02
OMPI/OSC/UCX: set priority to 0. 

xinzhao3:


bwbarrett:
I disagree with @hjelmn on this one.  I'd rather see it in master and then move to 3.1.  That way, we're positive we've fixed the core issues before we change priority in master.  Less chance of screwing up 4.0 the same way.

xinzhao3:
@hjelmn @bwbarrett could you review again? I can create another PR on 3.1.x once this one looks good.

hjelmn:
@xinzhao3 Thank you, looks good now. Once this passes CI please merge and open the v3.1.x PR.

PR #5136, opened 2018-05-02
mtl: reset ompi_mtl_base_selected_component on framework close

hjelmn:


hjelmn:
@angainor @jladd-mlnx  This should fix the crash when MXM is installed but UCX is used instead.

hjelmn:
@angainor Can you verify this fixes your issue?

angainor:
@hjelmn With this patch I can't reproduce the segfault.

PR #5140, opened 2018-05-03
dist: Update version to 3.1.0rc7

bwbarrett:
We almost had it.  I have great faith we won't need an rc7.


bwbarrett:
@angainor can you file an issue rather than comment on this PR?  This PR doesn't really have anything to do with release status, so your comment is likely to get lost.

PR #5146, opened 2018-05-04
Ensure pmix_cleanup finalizes the class system

rhc54:


rhc54:
@jjhursey Needs to be ported to release branches

PR #5147, opened 2018-05-04
btl/openib: add conditional around an assert

hppritcha:
A user trying to build Open MPI with explicit use
of CFLAGS on the make command line hit problems.

This fixes one of the problems.

https://www.mail-archive.com/users@lists.open-mpi.org//msg32241.html


rhc54:
@hppritcha I'm a little puzzled by this one. We have asserts all over the code base - obviously, if you disable debug but set optimizations to 0, they won't be compiled out. So won't you need to protect them everywhere? I'm asking because it sounds instead like there is an actual bug that is being covered up here.

jsquyres:
This one does seem to be correct -- that field only exists (via `opal_list.h`) if `OPAL_ENABLE_DEBUG` is 1.

rhc54:
Ah, ok - that's the clarification I was missing. Perhaps a comment would help future-person?

PR #5150, opened 2018-05-07
coll: use the basic algo for reduce_scatter_block by default

ggouaillardet:


jsquyres:
For cross reference: this PR is addressing the problem discussed in #5150.

PR #5015, opened 2018-04-04
config/opal_setup_java.m4:  Improve JDK Tool Path Resolution on OS X/macOS

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;This is just a first pass at this, as I expect my changes to be thoroughly reviewed.  I will, of course, address feedback as it comes in.

PR #5160, opened 2018-05-09
config/opal_setup_java.m4:  Fix #5015.  

RandomDSdevel:
Resolves https://github.com/open-mpi/ompi/pull/5001#issuecomment-387803012.

### DIFF
--- a/config/opal_setup_java.m4
+++ b/config/opal_setup_java.m4
@@ -113,8 +113,8 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
                           with_jdk_headers=$opal_java_dir/include
                           with_jdk_bindir=$opal_java_dir/bin
                       else
-                          AC_MSG_WARN([No recognized directory structure found under $opal_java_dir])
-                          AC_MSG_ERROR([Cannot continue])
+                          AC_MSG_WARN([No recognized OS X/macOS JDK directory structure found under $opal_java_dir])
+                          opal_java_found=0
                       fi],
                      [AC_MSG_RESULT([not found])])
 


rhc54:
Just glancing at this, and I'm puzzled - as the person who originally added Java support to OMPI, when did somebody change the default to enabling Java bindings?? The requirement for adding this was that it _never_ be enabled by default.

RandomDSdevel:
@rhc54:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I was wondering that, too, but thought I might best leave figuring that out to somebody else for whom it wasn't 'above their pay grade,' so to speak (this _is_ only my second Open MPI contribution after all.)  I'm leave it be for now just to keep the scope of this PR narrow and preserve its initial intent.  By all means, though, that's a fair question to ask and answer.

jsquyres:
ok to test

config/opal_setup_java.m4
```diff
@@ -113,8 +113,9 @@ AC_DEFUN([OPAL_SETUP_JAVA],[
                           with_jdk_headers=$opal_java_dir/include
                           with_jdk_bindir=$opal_java_dir/bin
                       else
-                          AC_MSG_WARN([No recognized directory structure found under $opal_java_dir])
-                          AC_MSG_ERROR([Cannot continue])
+                          AC_MSG_WARN([No recognized OS X/macOS JDK directory structure found under $opal_java_dir])
+                          opal_java_found=0
+                          AC_MSG_RESULT([Falling back to other platforms' JDK checks])
```
jsquyres: Good catch -- this does seem like it was the error (from my original pseudocode in a prior github comment).

I think you can get rid of the AC_MSG_RESULT -- it'll be obvious in the configure output that we're trying other places.

RandomDSdevel: @jsquyres:  Just thought I'd be consistent with there being a message there before (albeit the faulty one,) but I'll change it.

jsquyres:
@rhc54 The default is still not to enable Java; we *check* for Java compilers and whatnot, and then later determine whether to enable the support or not.  We could restructure that a bit to not even bother checking for Java if it wasn't requested, if that is desired (I don't think that particular logic has changed in quite a while -- it's in a different area than this particular PR addresses: see 31beff63622, back from 2012).

RandomDSdevel:
@jsquyres:  I've addressed your review comment.

jsquyres:
@RandomDSdevel Thank you!  Will merge after CI completes.

@bosilca This fixes your Java issue.

bosilca:
Magnificent !

ggouaillardet:
@jsquyres @rhc54 if I read correctly, the default is to check for Java support (e.g. `$enable_java != "no"`) but not build Java bindings (e.g. `$enable_mpi_java != "yes").

Is there a rationale having these two distinct `configure` command line options ?

rhc54:
Not that I can see - there is no reason to be checking for Java support if we aren't building the Java bindings. There was a time when we needed them separate because we were supporting non-MPI Java programs (e.g., map-reduce) and ORTE needed to know it could support Java, but that time has passed.

jsquyres:
@ggouaillardet Yeah, I noticed that yesterday afternoon, too: the variable we're checking is wrong.  But even fixing that, and fixing the condition to be `!= "yes"`, you get somewhat misleading configure output:

```
*** Java compiler
checking OS X/macOS locations... not found
checking Linux locations not found

*** Java MPI bindings
checking if want Java bindings... no
```

which is not quite accurate, because it's not that the Java compiler isn't found -- it's that *we didn't even look* if `--enable-mpi-java` wasn't specified.  I meant to do up a PR last night to make this a little better, but didn't get to it.

So just to be clear, currently:

* We always check for the Java compilers (regardless of whether you specified `--enable|disable-mpi-java`)
* We do not build the Java MPI bindings unless you specify `--enable-mpi-java`

We could change the first one to be a little more intelligent to not even search for the Java compiler unless `--enable-mpi-java` was specified, but I think to do it Right, we should do a little more than just change [that one incorrect `if` statement](https://github.com/open-mpi/ompi/blob/4a05c7e29fbe5e20b2b6eb52db529cf19979f8f7/config/opal_setup_java.m4#L55) that exists today.

PR #5162, opened 2018-05-11
patch/overwrite: add support for aarch64

hjelmn:
This commit adds patcher support of aarch64. The current
implementation uses mov, movk, and br to perform the jump. This uses 5
instructions.

### DIFF
--- a/opal/mca/patcher/base/patcher_base_patch.c
+++ b/opal/mca/patcher/base/patcher_base_patch.c
@@ -106,6 +106,8 @@ static void flush_and_invalidate_cache (unsigned long a)
     __asm__ volatile("mfence;clflush %0;mfence" : :"m" (*(char*)a));
 #elif OPAL_ASSEMBLY_ARCH == OPAL_IA64
     __asm__ volatile ("fc %0;; sync.i;; srlz.i;;" : : "r"(a) : "memory");
+#elif OPAL_ASSEMBLY_ARCH == OPAL_ARM64
+    __asm__ volatile ("dsb sy");
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


hjelmn:
@hppritcha FYI. Tested on darwin with the patcher check from IBM. Confirm the correct override is being called with the correct arguments.

@bwbarrett This might be considered a bug fix for v3.1.x. But only if we consider full aarch64 support something we want for v3.1.x. Otherwise this will be for v4.0.x.

hjelmn:
@shamisp FYI. This will have no affect on UCX + ARM + MLX until the UCT btl has been committed.

hjelmn:
Hmm,  re-running the checks. :bot:retest

hjelmn:
Very interesting. I am using this on aarch64 + mlx5 and don't see that issue. Will dig deeper and see if I can reproduce the issue.

shamisp:
It is tx1 with mlx5.

opal/mca/patcher/base/patcher_base_patch.c
```diff
@@ -106,6 +106,8 @@ static void flush_and_invalidate_cache (unsigned long a)
     __asm__ volatile("mfence;clflush %0;mfence" : :"m" (*(char*)a));
 #elif OPAL_ASSEMBLY_ARCH == OPAL_IA64
     __asm__ volatile ("fc %0;; sync.i;; srlz.i;;" : : "r"(a) : "memory");
+#elif OPAL_ASSEMBLY_ARCH == OPAL_ARM64
+    __asm__ volatile ("ic ialluis");
 #endif
```
shamisp: @hjelmn  If this is a self modifying code, I would expect to see ISB (or some stronger barrier) somewhere

hjelmn: Ah, ok. Thats simple enough.

One of the checks failed for illegal instruction. I don't know where though. Is ```ic alluis``` ever considered privileged?

shamisp: SCTLR_EL1.UCI has to be set to 1 in order to enable cache maintenance operations in EL0 (user space). More generic question - what we are trying to do here ? Can you just use DSB ?

hjelmn: DSB is probably all I need. Was just trying to replicate what was done on x86_64 which is most likely overkill.

hjelmn: Will update to use DSB now.

PR #5168, opened 2018-05-14
dist: Update shared library versions for v3.0.2

bwbarrett:
All the FORTRAN interfaces saw changes for COMM_SPAWN_MULTIPLE
handling.  We're calling the named parameter fixes an interface
addition (and not removal) under the assumption that no one was
using the old (wrong, hard to discover) names, but could use the
new names.

The verbose interface add was the define for the HDR link speeds.

All other changes were code changes identified through git log.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -82,16 +82,16 @@ date="Unreleased developer copy"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libmpi_so_version=40:1:0
+libmpi_so_version=40:2:0
 libmpi_cxx_so_version=40:0:0
-libmpi_mpifh_so_version=40:0:0
-libmpi_usempi_tkr_so_version=40:0:0
-libmpi_usempi_ignore_tkr_so_version=40:0:0
-libmpi_usempif08_so_version=40:0:0
-libopen_rte_so_version=40:1:0
-libopen_pal_so_version=41:0:1
+libmpi_mpifh_so_version=40:1:0
+libmpi_usempi_tkr_so_version=40:1:0
+libmpi_usempi_ignore_tkr_so_version=40:1:0
+libmpi_usempif08_so_version=40:1:0
+libopen_rte_so_version=40:2:0
+libopen_pal_so_version=41:1:1
 libmpi_java_so_version=40:0:0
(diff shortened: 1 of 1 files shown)


bwbarrett:
@jsquyres will you double check that my thoughts on the Fortran named parameter changes match your expectations?  Thanks!

VERSION
```diff
@@ -82,16 +82,16 @@ date="Unreleased developer copy"
...
-libmpi_mpifh_so_version=40:0:0
-libmpi_usempi_tkr_so_version=40:0:0
-libmpi_usempi_ignore_tkr_so_version=40:0:0
-libmpi_usempif08_so_version=40:0:0
-libopen_rte_so_version=40:1:0
-libopen_pal_so_version=41:0:1
+libmpi_mpifh_so_version=40:1:0
+libmpi_usempi_tkr_so_version=41:0:1
+libmpi_usempi_ignore_tkr_so_version=41:0:1
+libmpi_usempif08_so_version=41:0:1
```
jsquyres: @bwbarrett I confirm: changing a named parameter in a Fortran modulefile inflicts changes at *compile-time*, but does not effect changes at *link time*.

Specifically this works

1. Make a module `blah` with interface `foo` containing a named dummy parameter `ppp`
1. Compile a library with the subroutine that implements the interface `foo` with dummy parameter `ppp`
1. Compile an application that uses module `blah` and calls `foo(ppp=3)`
1. Run the program -- it works (`ppp` is passed into the subroutine properly)
1. Change the module to take named param `qqq` instead of `ppp`
1. Re-create the module
1. Re-compile the library as well
1. Run the program (*without* re-compiling it): the `ppp` value is still passed in to `foo()` properly, but now the `foo` interface implementation sees it as `qqq`
1. Attempting to re-compile the program (that calls `foo(ppp=3)`) predictably (and correctly) fail to compile.

Put simply: changing the dummy parameter name in the module interface does not seem to change link compatibility.  So I think we could just bump `r` in all the Fortran library names, and leave `c` and `a` alone.

bwbarrett:
@jsquyres bumped the Fortran changes to code changes instead of interface additions.  Let me know if this looks right to you.

PR #5171, opened 2018-05-15
Fix merge conflict related to function renaming.

bosilca:


ggouaillardet:
Thanks George !

I wonder how CI was unable to catch it though ...

PR #5173, opened 2018-05-16
PML/UCX: added processing of unordered messages to generic datatype

hoopoepg:
this is fix for https://github.com/openucx/ucx/issues/2568

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx_datatype.c
+++ b/ompi/mca/pml/ucx/pml_ucx_datatype.c
@@ -13,6 +13,12 @@
 
 #include <inttypes.h>
 
+typedef struct pml_ucx_unodered_item {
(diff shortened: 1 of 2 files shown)


ibm-ompi:

bosilca:
This method delays the unpacking of arriving messages until unpacking can be done in order. Another possible approach (used in OB1 PML) detects out-of-sequence messages and for each one clone the receiving convertor, set the clone to the right offset and do the unpacking. Of course this requires the completion to be done outside the convertor, as the clone does not report it's position back to the original convertor.

hoopoepg:
@yosefe look please

hoopoepg:
@bosilca I can't find where OB1 cloned receiver convertor... could you point me to code where it is implemented?
thank you

PR #5176, opened 2018-05-17
Disable OSHMEM layer when no SPMLs available

bwbarrett:
Implement the action item from last devel meeting to not build the OSHMEM layer when there are no SPMLs available (ie, the shmem layer can't possibly work).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,6 +23,8 @@
 # Copyright (c) 2014-2017 Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 # Copyright (c) 2016-2017 IBM Corporation.  All rights reserved.
+# Copyright (c) 2018      Amazon.com, Inc. or its affiliates.
+#                         All Rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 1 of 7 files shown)


bwbarrett:
Doh!  I forgot about making a tarball in this path.  a good reminder of why we have CI testing :).

configure.ac
```diff
@@ -1106,6 +1106,23 @@ OPAL_MCA
 
 m4_ifdef([project_ompi], [OMPI_REQUIRE_ENDPOINT_TAG_FINI])
 
+# Last minute disable of Open SHMEM if we didn't find any oshmem SPMLs
+if test "$project_oshmem_amc" = "true" -a $OSHMEM_FOUND_WORKING_SPML -eq 0 ; then
```
jsquyres: Should this (and the other lines like this in this commit) be `if test ... && test ...`?  I think there's some BSD `test` that doesn't like `-a`/`-o` -- I remember we've gotten in portability kerfuffles with this before.

configure.ac
```diff
@@ -1106,6 +1106,23 @@ OPAL_MCA
...
+    # We don't have an spml that will work, so oshmem wouldn't be able
+    # to run an application.  Therefore, don't build the oshmem layer.
+    if test "$enable_oshmem" != "no" -a -n "$enable_oshmem"; then
+	AC_MSG_WARN([No spml found, so Open SHMEM layer will be non functional.])
```
jsquyres: Minor nit: unlike Open MPI, the name is actually "OpenSHMEM" (i.e., no space).

bwbarrett: I saw Open SHMEM more than OpenSHMEM in our build code; will fix.

jsquyres:
@bwbarrett It just occurs to me: this PR probably means that we need to install UCX in our CI environments so that OSHMEM gets at least compile tested (e.g., PR #5092 nominally touches the OSHMEM layer, but OSHMEM will no longer compile by default on my platforms because I have no UCX-based networks).

Alternatively, should `--enable-oshmem` enable the OSHMEM layer to be compiled even if there are no viable SPMLs?  This feature might be for development/testing/CI only, perhaps...?

bwbarrett:
I'm pretty sure Mellanox CI tests UCX, so I wasn't worried about that.  I'd rather not add compile-only testing to the EC2 CI and when I tried last, the TCP driver in UCX didn't work.

PR #5186, opened 2018-05-22
MCA/UCX: fixed error messages for incorrect msg size

hoopoepg:
- supported 4 or 8 bytes only


hoopoepg:
@yosefe is it correct?

yosefe:
pls pick to v3.1.x

Issue #5190, opened 2018-05-23
master: configure failure due to redefinition of OPAL_C_HAVE__STATIC_ASSERT (among others)


benmenadue:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

master@385d91b

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

* Operating system/version: CentOS 6.9 + custom kernel (based on CentOS 7.4 kernel)
* Computer hardware: Fujitsu PRIMERGY
* Network type: Mellanox InfiniBand FDR

-----------------------------

## Details of the problem

Configuring master is failing in hwloc:hwloc2a because OPAL_C_HAVE__STATIC_ASSERT (among others) is being appended to `confdefs.h` by each test for that language feature. If a second or subsequent test succeeds, you end up with a redefinition of that macro, which then causes a test in hwloc:hwloc2a to fail because it uses `-Werror`.

For example, both of these add `#define OPAL_C_HAVE__STATIC_ASSERT 0`
```
configure:15649: checking if gcc -std=gnu11 supports C11 _Static_assert
configure:15649: checking if gcc -std=c11 supports C11 _Static_assert
```
but then this test (erroneously, see #5189) adds `#define OPAL_C_HAVE__STATIC_ASSERT 1`
```
configure:15665: gcc -c -DNDEBUG -g -march=native -c11  conftest.c >&5
```

Here's the failure in hwloc:hwloc2a (tidied up for readability):
```
configure:81547: checking if gcc -std=gnu99 supports -fvisibility=hidden
configure:81563: gcc -std=gnu99 -o conftest -DNDEBUG -g -march=native -fvisibility=hidden -Werror    -I/short/z00/bjm900/build/openmpi-mofed4.2/master/build/gcc/debug-0/opal/mca/hwloc/external/hwloc/include -I/apps/cuda/9.0/include    conftest.c -lrt -lm -lutil  -lz 
 -lhwloc  >&5
conftest.c:117:1: error: "OPAL_C_HAVE__STATIC_ASSERT" redefined
conftest.c:112:1: error: this is the location of the previous definition
configure:81563: $? = 1
configure: failed program was:
...
| #define OPAL_C_HAVE__THREAD_LOCAL 0
[...]

Issue #5189, opened 2018-05-23
master: false positive in configure for C11 _Static_assert with GCC 4.4.7


benmenadue:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

master@385d91b

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

* Operating system/version: CentOS 6.9 + custom kernel (based on CentOS 7.4 kernel)
* Computer hardware: Fujitsu PRIMERGY
* Network type: Mellanox InfiniBand FDR

-----------------------------

## Details of the problem

This test erroneously concludes that adding the `-c11` flag makes the compiler support `_Static_assert`. GCC is producing a warning about the unrecognised option, but still successfully compiles the file because it treats `_Static_assert(...)` as a implicitly-defined function. If it were to try linking the file it would probably fail with an undefined symbol.
```
configure:15649: checking if gcc -c11 supports C11 _Static_assert
configure:15665: gcc -c -DNDEBUG -g -march=native -c11  conftest.c >&5
gcc: unrecognized option '-c11'
configure:15665: $? = 0
configure:15680: result: yes
```

This is using
```
> gcc --version
gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-18)
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

jsquyres:
Fixed by #5191.

PR #5191, opened 2018-05-23
configure: use AC_LINK_IFELSE instead of AC_COMPILE_IFELSE for C11 tests

benmenadue:
This addresses #5189. It's also a workaround for #5190 in my particular case, but doesn't address the underlying problem there of the macros being defined multiple times.


jsquyres:
ok to test

PR #5193, opened 2018-05-24
Use /dev/shm for shared memory files in osc components

hjelmn:


bosilca:
The idea is good, and it push toward a convergence of all our backends (vader already had the same decision btp_vader_component.c:214). I think it makes sense to move this selection logic outside each individual component, so that all OMPI HPC-focused needs for shared memory regions benefit from the same logic.

jsquyres:
bot:ompi:retest

jsquyres:
bot:ompi:retest

jsquyres:
bot:ompi:retest

hjelmn:
This PR is probably the way to go for v3.0.x and v3.1.x. For v4.0.x I plan implement the decision logic in opal/shmem. This will change the MCA variables for vader, rdma, and sm.

PR #5194, opened 2018-05-24
io/ompio: fix an erroneous condition when selecting aggregator selection algorithm

edgargabriel:
fix the logic in the decision which aggregator selection algorithm
to use.. Note, that this lead to a suboptimal decision, but was not necessarily a problem.


edgargabriel:
bot:retest

edgargabriel:
not sure why the pull request build checker is hanging, but I am fairly confident that it has nothing to do with the one-line change of this commit. Will because of that go ahead and commit.

PR #5198, opened 2018-05-25
OSHMEM/SMPL/UCX: Add real fence support

brminich:
+ Add quiet method to SPML, so it can have different implementation with fence.
+ Use ucp_worker_fence for spml_fence method of UCX SPML


jsquyres:
ok to test

brminich:
@yosefe, please review

PR #5199, opened 2018-05-28
OSHMEM/SMPL/UCX: Add real fence support - 3.1.x

brminich:
+ Add quiet method to SPML, so it can have different implementation with fence.
+ Use ucp_worker_fence for spml_fence method of UCX SPML.


yosefe:
ok to test

hppritcha:
bot:lanl:retest

bwbarrett:
Does someone from Mellanox want to review this patch?

brminich:
@yosefe, plz review

yosefe:
let's keep it in master-only for now

PR #5204, opened 2018-05-30
fix info-subscribe to use snprintf() and warn on long key

### LINKED ISSUE #4959: opal/util/info.c can write keys that are too long
In 482d84b6e54, `opal/util/info.c` was heavily revamped.  Part of its commit message says:

> Inside info_subscribe_subscribe() the specified callback() is called with whatever matching k/v is in the object's info, or with the default. The return string from the callback goes into the new k/v stored in info, and the input k/v is saved as `__IN_<key>/<val>`. It's saved the same way whether the input came from info or whether it was a default. A null return from the callback indicates an ignored key/val, and no k/v is stored for  it, but an `__IN_<key>/<val>` is still kept so we still have access to the original.

However, when compiling this code with a modern gcc (e.g., v7.3.0), one gets the following warning message:

```
info.c: In function ‘opal_info_dup_mode’:
info.c:209:42: warning: ‘%s’ directive writing up to 36 bytes into a region of size 31 [-Wformat-overflow=]
                  sprintf(savedkey, "__IN_%s", iterator->ie_key);
                                          ^~
info.c:209:18: note: ‘sprintf’ output between 6 and 42 bytes into a destination of size 36
                  sprintf(savedkey, "__IN_%s", iterator->ie_key);
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

A few things:

1. We should be using `snprintf()`, not `sprintf()`.
1. The keys are strings of fixed length OPAL_MAX_INFO_KEY.  If you prepend `__IN_`, you're potentially writing up to 5 characters past the end of the array (i.e., if the key length is already >= OPAL_MAX_INFO_KEY-strlen("__IN_")).  The code needs to be adjusted to either take this additional prefix into account or have an additional attribute stashed somewhere that indicates that it's one of these special "in" keys.

markalle:
This checkin mainly concerns our internal info keys that are registering
for callbacks via opal_infosubscribe_subscribe(). Those keys need to have
an extra __IN_<key>/val stored to preserve their pre-callback value. So
that means our internal keys are limited to 5 chars shorter than the usual
key length limit.

The code previously would have been silently inactive if a large key happened
to come in, now it warns and also uses snprintf() to avoid compiler warnings.

I'm also making the top-level MPI_Info_set warn if the user uses our reserved
"__IN_" prefix. I had wanted the feature to be more invisible than that, but
it would require a more sophisticated approach to change that.

Fixes #4959

### DIFF
--- a/ompi/mpi/c/info_set.c
+++ b/ompi/mpi/c/info_set.c
@@ -12,6 +12,7 @@
  * Copyright (c) 2012-2013 Inria.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2018 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -26,6 +27,7 @@
 #include "ompi/communicator/communicator.h"
 #include "ompi/errhandler/errhandler.h"
 #include "ompi/info/info.h"
+#include "opal/util/show_help.h"
 #include <stdlib.h>
 #include <string.h>
 
@@ -95,6 +97,17 @@ int MPI_Info_set(MPI_Info info, const char *key, const char *value)
             return OMPI_ERRHANDLER_INVOKE (MPI_COMM_WORLD, MPI_ERR_INFO_VALUE,
                                            FUNC_NAME);
         }
+    }
+
+// An extra warning condition is a key that uses our reserved prefix "__IN_".
+// That one is used internally to deal with the dynamic nature the key/val
+// pairs where we have callbacks that modify the val, and the MPI standard
+// wants the get_info call to give back the original setting rather than
+// the callback-modified setting. So if a user directly used a key __IN_foo
+// it would confuse our accounting slightly.
+    if (0 == strncmp(key, OPAL_INFO_SAVE_PREFIX, strlen(OPAL_INFO_SAVE_PREFIX))) {
+        opal_show_help("help-mpi-api.txt", "info-set-with-reserved-prefix", true,
+            key, OPAL_INFO_SAVE_PREFIX);
     }
 
     OPAL_CR_ENTER_LIBRARY();
--- a/opal/util/info_subscriber.c
+++ b/opal/util/info_subscriber.c
@@ -16,7 +16,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016-2017 IBM Corporation. All rights reserved.
+ * Copyright (c) 2016-2018 IBM Corporation. All rights reserved.
  * Copyright (c) 2017      Intel, Inc. All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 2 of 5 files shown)


jsquyres:
@markalle Github pro tip: I edited your description to say "Fixes #4959" so that it will auto-close that issue when this PR is merged.

ompi/mpi/c/info_set.c
```diff
@@ -97,6 +97,17 @@ int MPI_Info_set(MPI_Info info, const char *key, const char *value)
         }
     }
 
+// An extra warning condition is a key that uses our reserved prefix "__IN_".
+// That one is used internally to deal with the dynamic nature the key/val
+// pairs where we have callbacks that modify the val, and the MPI standard
+// wants the get_info call to give back the original setting rather than
+// the callback-modified setting. So if a user directly used a key __IN_foo
+// it would confuse our accounting slightly.
+    if (0 == strncmp(key, "__IN_", 5)) {
```
jsquyres: Given that you use the string `"__IN_"` in multiple places in the code base, it might be good to either `#define` it, or have a global const string somewhere, and have all places in the code where you currently have `"__IN_"` use that global reference, instead.

Should we have the word `OMPI` in there somewhere?  E.g., `_OMPI_IN_`?  I only ask because it's a little weird to warn users about not naming info keys `__IN_`.

ompi/mpi/c/info_set.c
```diff
@@ -97,6 +97,17 @@ int MPI_Info_set(MPI_Info info, const char *key, const char *value)
...
+        fprintf(stderr, "MPI_Info_set warning on key %s: "
+            "the __IN_ prefix is reserved for internal use\n", key);
```
jsquyres: This should probably be an `opal_show_help()`.

opal/util/info_subscriber.c
```diff
@@ -348,6 +348,12 @@ int opal_infosubscribe_subscribe(opal_infosubscriber_t *object, char *key, char
     opal_hash_table_t *table = &object->s_subscriber_table;
     opal_callback_list_item_t *callback_list_item;
 
+    if (strlen(key) > OPAL_MAX_INFO_KEY-strlen("__IN_")) {
```
jsquyres: Is this a `opal_show_help()`-worthy error, or should it be an abort/die-worthy error (because it's an OMPI developer programmer error)?

markalle:
I agree with the above, will repush soon

opal/util/info_subscriber.c
markalle: Not sure, all I did so far for this one was make it say "DEVELOPER WARNING"

markalle:
@jsquyres, I made fixes for your comments except for the remaining
```
+    if (strlen(key) > OPAL_MAX_INFO_KEY-strlen("OPAL_INFO_SAVE_PREFIX")) {
+        fprintf(stderr, "DEVELOPER WARNING: Unexpected key length [%s]: "
+            "OMPI internal callback keys are limited to %d chars\n",
+            key, OPAL_MAX_INFO_KEY-strlen("OPAL_INFO_SAVE_PREFIX"));
+    }
```
I'm reluctant to make it abort, although I'm not sure why.

opal/util/info_subscriber.c
```diff
@@ -348,6 +349,12 @@ int opal_infosubscribe_subscribe(opal_infosubscriber_t *object, char *key, char
     opal_hash_table_t *table = &object->s_subscriber_table;
     opal_callback_list_item_t *callback_list_item;
 
+    if (strlen(key) > OPAL_MAX_INFO_KEY-strlen("OPAL_INFO_SAVE_PREFIX")) {
```
jsquyres: I think your strlen (here and a few lines down) is incorrect -- shouldn't have the quotes.

How about having it abort, but only if it's a debug build?  That should be pretty safe -- normal users don't make debug builds.

markalle: sounds good, and wow, what an error

markalle: Is that #if OPAL_ENABLE_DEBUG ?

jsquyres: Yep!

jsquyres:
Sorry -- I didn't see that there was a new commit on here (Github doesn't send emails when force pushes occur).

jsquyres:
This should probably get cherry picked to the relevant release branches.  This new info stuff made it into v3.0.0, right?  (and therefore also v3.1.x?)

Issue #5195, opened 2018-05-24
osc/rdma does not keep track of outstanding atomic op.


thananon:
**branch:** master;
**module:** osc/rdma
**btl**: ofi (in development) / sockets provider.

`ompi_osc_rdma_free()` does not enforce the remaining atomic operation completion. (It has the counter for outstanding rdma but not for atomic op) This leads to target window being freed before atomic operation completion and leads to segfault. This only happens to sockets provider as it is slow.

I verified this by forcing completion of atomic operation in btl_progress() but that is not ideal. I think the osc/rdma should handle this.

thananon:
@hppritcha , this is the cause of sockets provider failing/segfaulting in osu tests.

hjelmn:
Ah. Thanks for reporting this. Should be easy to fix

hjelmn:
I think I have a fix. Testing it now.

PR #5205, opened 2018-05-30
new btl/ofi: RDMA only btl using libfabric.

thananon:
This commit added new transport layer to be used with osc rdma module.
This BTL provides put, get, atomic and fetch atomic operations. It can
be used with multiple hardware vendors as long as they have their
provider under Libfabric and have the right capabilities.

Signed-off-by: Thananon Patinyasakdikul <thananon.patinyasakdikul@intel.com>

AND WE SHOULD TOTALLY CALL IT KYLO.

### DIFF
--- a/opal/mca/btl/ofi/README
+++ b/opal/mca/btl/ofi/README
@@ -1,0 +1,88 @@
+========================================
+Design notes on BTL/OFI
+========================================
+
+This is the RDMA only btl based on OFI Libfabric. The goal is to enable RDMA
+with multiple vendor hardware through one interface. Most of the operations are
+managed by upper layer (osc/rdma). This BTL is mostly doing the low level work.
+
+Tested providers: sockets,psm2,ugni
+
+========================================
+
+Component
+
+This BTL is requesting libfabric version 1.5 API and will not support older versions.
+
+The required capabilities of this BTL is FI_ATOMIC and FI_RMA with the endpoint type
+of FI_EP_RDM only. This BTL does NOT support libfabric provider that requires local
+memory registration (FI_MR_LOCAL).
+
+BTL/OFI will initialize a module with ONLY the first compatible info returned from OFI.
+This means it will rely on OFI provider to do load balancing. The support for multiple
+device might be added later.
+
+The BTL creates only one endpoint and one CQ.
+
+========================================
+
+Memory Registration
+
+Open MPI has a system in place to exchange remote address and always use the remote
+virtual address to refer to a piece of memory. However, some libfabric providers might
+not support the use of virtual address and instead will use zero-based offset addressing.
+
+FI_MR_VIRT_ADDR is the flag that determine this behavior. mca_btl_ofi_reg_mem() handles
+this by storing the base address in registration handle in case of the provider does not
+support FI_MR_VIRT_ADDR. This base address will be used to calculate the offset later in
+RDMA/Atomic operations.
+
+The BTL will try to use the address of registration handle as the key. However, if the
+provider supports FI_MR_PROV_KEY, it will use provider provided key. Simply does not care.
+
+The BTL does not register local operand or compare. This is why this BTL does not support
+FI_MR_LOCAL and will allocate every buffer before registering. This means FI_MR_ALLOCATED
+is supported. So to be explicit.
+
+Supported MR mode bits (will work with or without):
+    enum:
+    - FI_MR_BASIC
+    - FI_MR_SCALABLE
+
+    mode bits:
+    - FI_MR_VIRT_ADDR
+    - FI_MR_ALLOCATED
+    - FI_MR_PROV_KEY
+
+The BTL does NOT support (will not work with):
+    - FI_MR_LOCAL
+    - FI_MR_MMU_NOTIFY
+    - FI_MR_RMA_EVENT
+    - FI_MR_ENDPOINT
+
+Just a reminder, in libfabric API 1.5...
+FI_MR_BASIC == (FI_MR_PROV_KEY | FI_MR_ALLOCATED | FI_MR_VIRT_ADDR)
+
+========================================
+
+Completions
+
+Every operation in this BTL is asynchronous. The completion handling will occur in
+mca_btl_ofi_component_progress() where we read the CQ with the completion context and
+execute the callback functions. The completions are local. No remote completion event is
+generated as local completion already guarantee global completion.
+
+The BTL keep tracks of number of outstanding operations and provide flush interface.
+
+========================================
+
+Sockets Provider
+
+Sockets provider is the proof of concept provider for libfabric. It is supposed to support
+all the OFI API with emulations. This provider is considered very slow and bound to raise
+problems that we might not see from other faster providers.
+
+Known Problems:
+    - sockets provider uses progress thread and can cause segfault in finalize as we free
+      the resources while progress thread is still using it. sleep(1) was put in
+      mca_btl_ofi_componenet_close() for this reason.
--- a/opal/mca/btl/ofi/btl_ofi_atomics.c
+++ b/opal/mca/btl/ofi/btl_ofi_atomics.c
@@ -1,0 +1,180 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2014-2018 Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2018      Intel, Inc, All rights reserved
+ *
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include <rdma/fi_atomic.h>
+#include "btl_ofi_rdma.h"
+
+static inline int to_fi_op(mca_btl_base_atomic_op_t op)
+{
+    switch (op) {
+    case MCA_BTL_ATOMIC_ADD:
+        return FI_SUM;
+    case MCA_BTL_ATOMIC_SWAP:
+        return FI_ATOMIC_WRITE;
+    default:
+        BTL_ERROR(("Unknown or unsupported atomic op."));
+        MCA_BTL_OFI_ABORT();
+
+        /* just to squash the warning */
+        return OPAL_ERROR;
+    }
+}
+
+int mca_btl_ofi_afop (struct mca_btl_base_module_t *btl, struct mca_btl_base_endpoint_t *endpoint,
+                      void *local_address, uint64_t remote_address, mca_btl_base_registration_handle_t *local_handle,
+                      mca_btl_base_registration_handle_t *remote_handle, mca_btl_base_atomic_op_t op,
+                      uint64_t operand, int flags, int order, mca_btl_base_rdma_completion_fn_t cbfunc,
+                      void *cbcontext, void *cbdata)
+{
+    int rc;
+    int fi_datatype = FI_UINT64;
+    int fi_op;
+
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+    mca_btl_ofi_endpoint_t *btl_endpoint = (mca_btl_ofi_endpoint_t*) endpoint;
+    mca_btl_ofi_completion_t *comp = NULL;
+
+    if (flags & MCA_BTL_ATOMIC_FLAG_32BIT) {
+        fi_datatype = FI_UINT32;
+    }
+
+    fi_op = to_fi_op(op);
+
+    comp = mca_btl_ofi_completion_alloc(btl, endpoint,
+                                        local_address,
+                                        local_handle,
+                                        cbfunc, cbcontext, cbdata,
+                                        MCA_BTL_OFI_TYPE_AFOP);
+
+    /* copy the operand because it might get freed from upper layer */
+    comp->operand = (uint64_t) operand;
+
+    remote_address = (remote_address - (uint64_t) remote_handle->base_addr);
+
+    rc = fi_fetch_atomic(ofi_btl->ofi_endpoint,
+                         (void*) &comp->operand, 1, NULL,       /* operand */
+                         local_address, local_handle->desc,     /* results */
+                         btl_endpoint->peer_addr,               /* remote addr */
+                         remote_address, remote_handle->rkey,   /* remote buffer */
+                         fi_datatype, fi_op, comp);
+
+    if (rc == -FI_EAGAIN) {
+        return OPAL_ERR_OUT_OF_RESOURCE;
+    } else if (rc < 0) {
+        BTL_ERROR(("fi_fetch_atomic failed with rc=%d (%s)", rc, fi_strerror(-rc)));
+        MCA_BTL_OFI_ABORT();
+    }
+
+    MCA_BTL_OFI_NUM_RDMA_INC(ofi_btl);
+
+    return OPAL_SUCCESS;
+}
+
+int mca_btl_ofi_aop (struct mca_btl_base_module_t *btl, mca_btl_base_endpoint_t *endpoint,
+                     uint64_t remote_address, mca_btl_base_registration_handle_t *remote_handle,
+                     mca_btl_base_atomic_op_t op, uint64_t operand, int flags, int order,
+                     mca_btl_base_rdma_completion_fn_t cbfunc, void *cbcontext, void *cbdata)
+{
+    int rc;
+    int fi_datatype = FI_UINT64;
+    int fi_op;
+
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+    mca_btl_ofi_endpoint_t *btl_endpoint = (mca_btl_ofi_endpoint_t*) endpoint;
+    mca_btl_ofi_completion_t *comp = NULL;
+
+    if (flags & MCA_BTL_ATOMIC_FLAG_32BIT) {
+        fi_datatype = FI_UINT32;
+    }
+
+    fi_op = to_fi_op(op);
+
+    comp = mca_btl_ofi_completion_alloc(btl, endpoint,
+                                        NULL,
+                                        NULL,
+                                        cbfunc, cbcontext, cbdata,
+                                        MCA_BTL_OFI_TYPE_AOP);
+
+    /* copy the operand because it might get freed from upper layer */
+    comp->operand = (uint64_t) operand;
+
+    remote_address = (remote_address - (uint64_t) remote_handle->base_addr);
+
+    rc = fi_atomic(ofi_btl->ofi_endpoint,
+                   (void*) &comp->operand, 1, NULL,       /* operand */
+                   btl_endpoint->peer_addr,               /* remote addr */
+                   remote_address, remote_handle->rkey,   /* remote buffer */
+                   fi_datatype, fi_op, comp);
+
+    if (rc == -FI_EAGAIN) {
+        return OPAL_ERR_OUT_OF_RESOURCE;
+    } else if (rc < 0) {
+        BTL_ERROR(("fi_atomic failed with rc=%d (%s)", rc, fi_strerror(-rc)));
+        MCA_BTL_OFI_ABORT();
+    }
+
+    MCA_BTL_OFI_NUM_RDMA_INC(ofi_btl);
+
+    return OPAL_SUCCESS;
+}
+
+int mca_btl_ofi_acswap (struct mca_btl_base_module_t *btl, struct mca_btl_base_endpoint_t *endpoint,
+                        void *local_address, uint64_t remote_address, mca_btl_base_registration_handle_t *local_handle,
+                        mca_btl_base_registration_handle_t *remote_handle, uint64_t compare, uint64_t value, int flags,
+                        int order, mca_btl_base_rdma_completion_fn_t cbfunc, void *cbcontext, void *cbdata)
+{
+    int rc;
+    int fi_datatype = FI_UINT64;
+
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+    mca_btl_ofi_endpoint_t *btl_endpoint = (mca_btl_ofi_endpoint_t*) endpoint;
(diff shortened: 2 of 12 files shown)


opal/mca/btl/ofi/README
```diff
@@ -0,0 +1,69 @@
...
+
+========================================
+
+Completions
+
+Every operation in this BTL is asynchronous. The completion handling will occur in
+mca_btl_ofi_component_progress() where we read the CQ with the completion context and
+execute the callback functions. The completions are local. No remote completion event is
+generated.
+
```
hjelmn: By local you mean that the event is local (you can't raise a remote event)? Are completions global in the sense that the amo or put is visible on the remote system?

opal/mca/btl/ofi/btl_ofi_module.c
```diff
@@ -0,0 +1,300 @@
...
+    int rc;
+    int count;
+    char *ep_name = NULL;
+    size_t namelen = mca_btl_ofi_component.namelen;
+
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+
+    for (size_t i = 0 ; i < nprocs ; ++i) {
+        /* all endpoints are reachable for uct :p */
+        peers[i] = mca_btl_ofi_endpoint_create (opal_procs[i], ofi_btl->ofi_endpoint);
```
hjelmn: ofi, not uct :)

thananon:
This PR is test against sockets/psm2/ugni provider with ibm/1sided, osu 1 sided.
Note: There is still SEGV in some cases caused by #5195 . This BTL will crash with np > 2 without the fix.

opal/mca/btl/ofi/README
thananon: We can generate remote event but we choose not to as the local completion event guarantees  remote completion. So, yes, it is visible on remote system.

opal/mca/btl/ofi/btl_ofi_module.c
thananon: woops.

ibm-ompi:

 Gist: https://gist.github.com/c94ee163a102d8930df40a408122c887

ibm-ompi:

 Gist: https://gist.github.com/9a8c864feb6a535c44060e5e25831d67

hjelmn:
Sure, will fix #5195 tomorrow.

ibm-ompi:

 Gist: https://gist.github.com/fd90e9c0237afbed81c5e995305a9e06

thananon:
Hmm, so the build failed because there is no libfabric or am I doing something wrong?

bwbarrett:
@thananon most of the builders do not have libfabric, and that's a good thing.  The thing missing from your patch is to evaluate either $2 or $3 based on whether or not your package can build.  You might want to look at `ompi/mca/mtl/ofi/configure.m4` for an example.

opal/mca/btl/ofi/README
```diff
@@ -0,0 +1,69 @@
...
+Sockets Provider
+
+Sockets provider is the proof of concept provider for libfabric. It is supposed to support
+all the OFI API with emulations. This provider is considered very slow and bound to raise
+problems that we might not see from other faster providers.
+
+Known Problems:
+    - sockets provider uses progress thread and can cause segfault in finalize as we free
+      the resources while progress thread is still using it. sleep(1) was put in
+      mca_btl_ofi_componenet_close() for this reason.
```
jsquyres: Hmm.  Seems sketchy.  This sounds like a bug either in libfabric or in this BTL.

opal/mca/btl/ofi/btl_ofi_atomics.c
```diff
@@ -0,0 +1,177 @@
...
+                         local_address, local_handle->desc,     /* results */
+                         btl_endpoint->peer_addr,               /* remote addr */
+                         remote_address, remote_handle->rkey,   /* remote buffer */
+                         fi_datatype, fi_op, comp);
+
+    if (rc == -FI_EAGAIN) {
+        return OPAL_ERR_OUT_OF_RESOURCE;
+    } else if (rc < 0) {
+        BTL_ERROR(("fi_fetch_atomic failed with rc=%d (%s)", rc, fi_strerror(-rc)));
+        abort();
```
jsquyres: You probably don't want an `abort()` -- you probably just want `exit(1)`, or, better yet, something like `opal_btl_usnic_exit()`.  `abort()` will drop a corefile, which, if you're running under MTT and you hit one of these code paths, can leave terrabytes of corefiles lying around.

I've been meaning to get rid of all the `abort()`s in the usNIC btl for quite a while for exactly this reason.  😲

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,538 @@
...
+     * Since now we are using the first valid info returned, I'm not sure
+     * if we need to provide the support for comma limited provider list. */
+    prov_include = NULL;
+    (void) mca_base_component_var_register(&mca_btl_ofi_component.super.btl_version,
+                                          "provider_include",
+                                          "OFI provider that are considered for use "
+                                          "(e.g., \"psm,psm2\"; an empty value means that all providers will "
+                                          "be considered.",
+                                          MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
+                                          OPAL_INFO_LVL_1,
```
jsquyres: These should not be level 1.  See https://github.com/open-mpi/ompi/wiki/MCAParamLevels.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,538 @@
...
+                                          " be considered). "
+                                          "Mutually exclusive with btl_ofi_provider_include.",
+                                          MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
+                                          OPAL_INFO_LVL_1,
+                                          MCA_BASE_VAR_SCOPE_READONLY,
+                                          &prov_exclude);
+
+    /* Note: better leave it at 1 for now. osc rdma module is designed for 1 completion
+     * at a time. Dealing with more than 1 completion in 1 read will confuse the osc rdma.
+     * source: 8 hours of debugging. :(*/
```
jsquyres: Did you talk to @hjelmn about this?

opal/mca/btl/ofi/btl_ofi_providers.h
```diff
@@ -0,0 +1,98 @@
...
+        return OPAL_ERROR;
+    }
+
+    /* provider specifics */
+    if (!strcmp(prov_name, "sockets")) {
+        return socks_validate_info(info);
+    } else if (!strcmp(prov_name, "psm2")) {
+        return psm2_validate_info(info);
+    } else if (!strcmp(prov_name, "gni")) {
+        return gni_validate_info(info);
```
jsquyres: Why are you hard-coding provider names / logic?

opal/mca/btl/ofi/btl_ofi_providers.h
```diff
@@ -0,0 +1,98 @@
...
+    char *prov_name;
+
+    assert(hints->fabric_attr);
+    assert(hints->domain_attr);
+    assert(hints->fabric_attr->prov_name);
+
+    prov_name = hints->fabric_attr->prov_name;
+
+    if (!strcmp(prov_name, "gni")) {
+        hints->domain_attr->mr_mode = FI_MR_BASIC;
```
jsquyres: This seems dangerous -- if you *have* to hard-code hints on a per-provider basis, you should probably have MCA vars to override them, too (because they will change over time).

opal/mca/btl/ofi/configure.m4
```diff
@@ -0,0 +1,38 @@
...
+# support, otherwise executes action-if-not-found
+
+AC_DEFUN([MCA_opal_btl_ofi_CONFIG],[
+    AC_CONFIG_FILES([opal/mca/btl/ofi/Makefile])
+
+    # substitute in the things needed to build ucx
+    AC_SUBST([btl_ofi_CPPFLAGS])
+    AC_SUBST([btl_ofi_LDFLAGS])
+    AC_SUBST([btl_ofi_LIBS])
+    CPPFLAGS="$opal_common_ofi_CPPFLAGS $CPPFLAGS"
```
jsquyres: What's the purpose of setting the OFI CPPFLAGS in the global `CPPFLAGS` here?

opal/mca/btl/ofi/btl_ofi_component.c
thananon: I talked to @hppritcha who talked to @hjelmn and they agreed that this is a bug. I haven't file the issue upstream.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: I don't like it neither but what would be a better way to do it? I took a look at MTL/OFI and they are more or less the same by converting string into some enum. So I think the hard-coding has to be there somewhere right?

opal/mca/btl/ofi/btl_ofi_component.c
jsquyres: Ok.  Note that I didn't specifically call out each of the MCA params, but you might want to look at the levels of all your MCA params in the context of that wiki page.  The idea is that we want very, very few things in levels 1,2,3.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: We are dealing with multiple providers who has their own setting to satisfy. I don't see how this is different from when we hard-coded endpoint type or caps. 

I can have the MCA to override the MR setting. That won't be easy as it is a flag not an enum but I can try.

opal/mca/btl/ofi/configure.m4
thananon: This is my inexperience with m4. 
Fixed in upcoming commit.

opal/mca/btl/ofi/btl_ofi_providers.h
hppritcha: @thananon let's discuss this next week on our call.

opal/mca/btl/ofi/btl_ofi_component.c
thananon: I fixed all of the param and they are at least level 4.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: @hppritcha I would like to get this done as soon as possible.

Is there any suggestion for me that I can do now? I don't really like `!strcmp` but that's the only way I can see as the provider `prov_name` field is a string.

As for MCA override, I think the easy work around is having one to force FI_MR_BASIC and make it a level 8+. All the provider on 1.5 right now supports FI_MR_BASIC so it should be pretty safe.

opal/mca/btl/ofi/btl_ofi_providers.h
jsquyres: Shouldn't you validate (and set hints) based on what the BTL needs?

thananon: Yeah, right now those validate info functions are just empty. It was in early development where we have to force certain flags for certain providers. 

I still keep it because some provider might really need it. i.e,  provider returns FI_MR_UNSPEC and we want to force only FI_MR_VIRT_ADDR.

thananon: some provider also return FI_MR_UNSPEC which in API 1.5 assume FI_MR_SCALABLE but in some case we want FI_MR_BASIC (which provider totally support) so I have a place for the providers/dev to force it themselves.

thananon: I see what you are saying here. Let's not have this provider specific thing. 
will fix in upcoming commit.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: @hppritcha 

I just realized we dont need to hard code FI_MR_BASIC for gni anymore as now we requested FI_MR_UNSPEC and deal with that in `validate_info()`.  So this can go away.

Fixed in upcoming commit.

opal/mca/btl/ofi/README
thananon: I believe this is related to #5195 .  Will check again after we have a fix for that..

hjelmn:
@thananon Please squish this all into a single commit.

thananon:
Done.

opal/mca/btl/ofi/btl_ofi_module.c
```diff
@@ -0,0 +1,296 @@
...
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+
+    for (size_t i = 0 ; i < nprocs ; ++i) {
+        if (peers[i]) {
+            opal_list_remove_item (&ofi_btl->endpoints, &peers[i]->super);
+            OBJ_RELEASE(peers[i]);
+        }
+    }
+
+    return OPAL_SUCCESS;
```
jsquyres: Is it worth removing the av entry here, too?

opal/mca/btl/ofi/btl_ofi_module.c
```diff
@@ -0,0 +1,296 @@
...
+}
+
+int mca_btl_ofi_dereg_mem (void *reg_data, mca_rcache_base_registration_t *reg)
+{
+    mca_btl_ofi_reg_t *ur = (mca_btl_ofi_reg_t*)reg;
+
+    if (ur->ur_mr != NULL) {
+        if (0 != fi_close(&ur->ur_mr->fid)) {
+            opal_output(0, "%s: error unpinning memory mr=%p: %s",
```
jsquyres: Shouldn't this be a BTL_ERROR or opal_show_help()?

opal/mca/btl/ofi/btl_ofi_module.c
```diff
@@ -0,0 +1,296 @@
...
+/*
+ * Cleanup/release module resources.
+ */
+
+int mca_btl_ofi_finalize (mca_btl_base_module_t* btl)
+{
+    mca_btl_ofi_module_t *ofi_btl = (mca_btl_ofi_module_t *) btl;
+    mca_btl_ofi_endpoint_t *endpoint, *next;
+
+    fi_close(&ofi_btl->cq->fid);
```
jsquyres: These close and freeinfo statements might want to have conditionals around them to ensure that various pointers aren't NULL.  E.g.,

```c
if (NULL != ofi_btl->cq) {
    fi_close(&obi_btl->cq->fid);
}
```

just in case finalize is called early and somehow these pointers haven't been setup yet.  I.e., make it safe to call finalize at any time / at any point during the setup.

opal/mca/btl/ofi/btl_ofi_providers.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "btl_ofi.h"
+
+#define MCA_BTL_OFI_REQUIRED_CAPS       (FI_RMA | FI_ATOMIC)
+#define MCA_BTL_OFI_REQUESTED_MR_MODE   (FI_MR_UNSPEC)
+
+
+/* validate information returned from fi_getinfo().
+ * return OPAL_ERROR if we dont have what we need. */
+static int validate_info(struct fi_info *info)
```
jsquyres: Why is this in a header file?  It doesn't seem like this needs to be inlined.

opal/mca/btl/ofi/btl_ofi_providers.h
```diff
@@ -0,0 +1,42 @@
...
+{
+    int mr_mode;
+
+    /* we need exactly all the required bits */
+    if ((info->caps & MCA_BTL_OFI_REQUIRED_CAPS) != MCA_BTL_OFI_REQUIRED_CAPS)
+        return OPAL_ERROR;
```
jsquyres: Nit pick on the style: don't forget to `{}` *all* blocks, even 1-line blocks.  Sorry -- that's OMPI style.  😉

jsquyres:
Most of my comments are pretty minor, actually -- should probably be pretty easy to fix the issues.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,541 @@
...
+
+    return OPAL_SUCCESS;
+
+fail:
+    /* clean up */
+    fi_close(&av->fid);
+    fi_close(&cq->fid);
+    fi_close(&endpoint->fid);
+    fi_close(&domain->fid);
+    fi_close(&fabric->fid);
```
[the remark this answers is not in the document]
thananon: I was hoping that fi_close will handle NULL pointer. 
fixed in upcoming commit.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,541 @@
...
+    BTL_ERROR(("BTL OFI will now abort."));
+    exit(1);
+}
+
+/*
+ *  OFI component initialization:
+ *  (1) read interface list from kernel and compare against component parameters
+ *      then create a BTL instance for selected interfaces
+ *  (2) setup OFI listen socket for incoming connection attempts
+ *  (3) register BTL parameters with the MCA
```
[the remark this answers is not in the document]
thananon: remnant from UCT BTL. 
removed in upcoming commit.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,541 @@
...
+{
+    BTL_ERROR(("BTL OFI will now abort."));
+    exit(1);
+}
+
+/*
+ *  OFI component initialization:
+ *  (1) read interface list from kernel and compare against component parameters
+ *      then create a BTL instance for selected interfaces
+ *  (2) setup OFI listen socket for incoming connection attempts
```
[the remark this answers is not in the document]
thananon: removed in upcoming commit.

opal/mca/btl/ofi/btl_ofi_module.c
thananon: fixed.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: I tried to catch all of it. Seems like I missed this one. I got confused midway between libfabric and ompi standard.

opal/mca/btl/ofi/btl_ofi_providers.h
thananon: maybe we dont need this file anymore after I removed the provider hard-coded setting.  Let me move this onto component.c.

opal/mca/btl/ofi/btl_ofi_module.c
thananon: Why not?
fixed.

opal/mca/btl/ofi/btl_ofi_component.c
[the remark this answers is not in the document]
jsquyres: I don't think that's the issue -- you're dereferencing uninitialized pointers here.  So it would segv before it even called `fi_close()`.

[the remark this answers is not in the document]
thananon: gotcha.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,583 @@
...
+    char *linux_device_name;
+    char ep_name[FI_NAME_MAX];
+    struct fi_info *ofi_info;
+    struct fi_cq_attr cq_attr = {0};
+    struct fi_av_attr av_attr = {0};
+    struct fid_fabric *fabric;
+    struct fid_domain *domain;
+    struct fid_ep *endpoint;
+    struct fid_cq *cq;
+    struct fid_av *av;
```
jsquyres: I think you need to initialize all of these pointers to NULL to make sure your safety conditionals in the `fail` block work as expected.  😉

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -0,0 +1,583 @@
...
+    if (0 != rc) {
+        BTL_VERBOSE(("%s failed fi_enable with err=%s",
+                        linux_device_name,
+                        fi_strerror(-rc)
+                        ));
+        goto fail;
+    }
+
+    /* Everything succeeded, lets create a module for this device. */
+    module = (mca_btl_ofi_module_t*) malloc(sizeof(mca_btl_ofi_module_t));
```
jsquyres: I think you need to `calloc()` this (vs. `malloc()`) in order to ensure that the NULL protections against pointers in the module struct in the module finalize work as expected.

opal/mca/btl/ofi/README
```diff
@@ -0,0 +1,69 @@
...
+
+Component
+
+This BTL is requesting libfabric version 1.5 API and will not support older versions.
+
+The required capabilities of this BTL is FI_ATOMIC and FI_RMA with the endpoint type
+of FI_EP_RDM only. This BTL does NOT support libfabric provider that requires local
+memory registration (FI_MR_LOCAL).
+
+The provider specific settings can be added to btl_ofi_providers.h. This includes the
```
jsquyres: Bzzt. 😉

thananon:
This should earn that green tick or I will rename it to KYLO. It's unfair how other BTL got away with the restriction. (looking at you, vader).

opal/mca/btl/ofi/btl_ofi_component.c
jsquyres: You kids and your C99... (I'm not actually 100% sure if stack variables are guaranteed to be 0'ed out or not...?).

thananon: I know, we have it easy. 
No but really, does Open MPI build without C99?

jsquyres: No, OMPI's configure enforces building with a C99-compliant compiler.

hjelmn: stack variables are not zero by default

PR #5209, opened 2018-05-31
opal/asm: change ll/sc atomics to macros

hjelmn:
This commit fixes a hang that occurs with debug builds of Open MPI on
aarch64 and power/powerpc systems. When the ll/sc atomics are inline
functions the compiler emits load/store instructions for the function
arguments with -O0. These extra load/store arguments can cause the ll
reservation to be cancelled causing live-lock.

Note that we did attempt to fix this with always_inline but the extra
instructions are stil emitted by the compiler (gcc). There may be
another fix but this has been tested and is working well.

References #3697. Close when applied to v3.0.x and v3.1.x.

### DIFF
--- a/opal/class/opal_fifo.h
+++ b/opal/class/opal_fifo.h
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2007      Voltaire All rights reserved.
  * Copyright (c) 2010      IBM Corporation.  All rights reserved.
- * Copyright (c) 2014-2017 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2014-2018 Los Alamos National Security, LLC. All rights
  *                         reseved.
  * $COPYRIGHT$
  *
@@ -183,9 +183,10 @@ static inline opal_list_item_t *opal_fifo_pop_atomic (opal_fifo_t *fifo)
 static inline opal_list_item_t *opal_fifo_push_atomic (opal_fifo_t *fifo,
                                                        opal_list_item_t *item)
 {
+    const opal_list_item_t * const ghost = &fifo->opal_fifo_ghost;
     opal_list_item_t *tail_item;
 
-    item->opal_list_next = &fifo->opal_fifo_ghost;
+    item->opal_list_next = (opal_list_item_t *) ghost;
 
     opal_atomic_wmb ();
 
@@ -194,7 +195,7 @@ static inline opal_list_item_t *opal_fifo_push_atomic (opal_fifo_t *fifo,
 
     opal_atomic_wmb ();
 
-    if (&fifo->opal_fifo_ghost == tail_item) {
+    if (ghost == tail_item) {
         /* update the head */
         fifo->opal_fifo_head.data.item = item;
     } else {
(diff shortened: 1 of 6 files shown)


hjelmn:
FYI @shamisp, @kawashima-fj

hjelmn:
:bot:retest:ompi

opal/class/opal_fifo.h
```diff
@@ -194,7 +196,7 @@ static inline opal_list_item_t *opal_fifo_push_atomic (opal_fifo_t *fifo,
 
     opal_atomic_wmb ();
```
bosilca: Why we need this wmb ? there is nothing to flush, except the atomic swap right above.

bosilca: In fact I am not sure I understand how this code can be correct in the case where 2 threads are doing conflicting operations on a FIFO that contains only one element. Imagine a thread that pop while the second one push. The operation at line 202 (```tail_item->opal_list_next = item```) conflicts with the ```opal_atomic_sc_ptr``` in the pop, and can lead to a case where the head is NULL (it contains what the previous element assumed was his next) but the tail is not (it contains the result of the previous push).

opal/class/opal_fifo.h
```diff
@@ -258,11 +271,13 @@ static inline opal_list_item_t *opal_fifo_pop_atomic (opal_fifo_t *fifo)
     if (ghost == next) {
         void *tmp = item;
 
-        if (!opal_atomic_compare_exchange_strong_ptr (&fifo->opal_fifo_tail.data.item, &tmp, ghost)) {
+        if (!opal_atomic_compare_exchange_strong_ptr (&fifo->opal_fifo_tail.data.item, &tmp, (void *) ghost)) {
             while (ghost == item->opal_list_next) {
```
bosilca:
```
do {
    opal_atomic_rmb ();
} while (ghost != item->opal_list_next);
``` 
?

opal/include/opal/sys/arm64/atomic.h
```diff
@@ -162,28 +162,31 @@ static inline bool opal_atomic_compare_exchange_strong_rel_32 (volatile int32_t
...
+#define opal_atomic_ll_32(addr, ret)                                    \
+    do {                                                                \
+        volatile int32_t *_addr = (addr);                               \
+        int32_t _ret;                                                   \
+                                                                        \
+        __asm__ __volatile__ ("ldaxr    %w0, [%1]          \n"          \
+                              : "=&r" (_ret)                            \
+                              : "r" (_addr));                           \
+                                                                        \
+        ret = (typeof(ret)) _ret;                                       \
```
bosilca: Does this conversion work for any signed type ?

opal/class/opal_fifo.h
```diff
@@ -183,9 +183,11 @@ static inline opal_list_item_t *opal_fifo_pop_atomic (opal_fifo_t *fifo)
 static inline opal_list_item_t *opal_fifo_push_atomic (opal_fifo_t *fifo,
                                                        opal_list_item_t *item)
 {
+    const opal_list_item_t * const ghost = &fifo->opal_fifo_ghost;
     opal_list_item_t *tail_item;
 
-    item->opal_list_next = &fifo->opal_fifo_ghost;
+    assert (NULL == item->opal_list_next);
```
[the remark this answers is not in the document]
hjelmn: Some debugging that crept in. Will clean that out now.

opal/class/opal_fifo.h
hjelmn: This scenario is tested by test/class/opal_fifo.c. The exhaustive test causes the threads to empty the fifo. 

The code that keeps  things consistent is at line 276. If the pop detected that the fifo was empty but something was being pushed (next == ghost && tail != item) it waits until the next pointer has been updated then sets the head to the new item. This is the same strategy used by vader and nemesis.

opal/class/opal_fifo.h
hjelmn: Yeah, cleaner that way. Changing it now.

opal/class/opal_fifo.h
hjelmn: Yeah, looking at it there is no way it can get to one of the following lines without the read and write completing anyway. That means the wmb does nothing. It shouldn't hurt though.

shamisp: If this is not critical path, it is fine ...

opal/include/opal/sys/powerpc/atomic.h
```diff
@@ -165,31 +165,35 @@ static inline bool opal_atomic_compare_exchange_strong_32 (volatile int32_t *add
...
-                          "1:"
-                          : "=r" (ret), "=m" (*addr), "=r" (foo)
-                          : "r" (addr), "r" (newval)
-                          : "cc", "memory");
-    return ret;
-}
+/* NTH: the LL/SC support is done through macros due to issues with non-optimized builds. The reason
+ * is that even with an always_inline attribute the compiler may still emit instructions to store then
+ * load the arguments to/from the stack. This sequence may cause the ll reservation to be cancelled. */
+#define opal_atomic_ll_32(addr, ret)                                    \
```
shamisp: Makes sense. This is useful comment that should be present in some upper layer. It true for Power and Arm

shamisp:
👍

opal/include/opal/sys/powerpc/atomic.h
hjelmn: Yeah. I plan to add some additional documentation detailing what live-lock is with LL/SC.

PR #5210, opened 2018-06-01
fortran/mpif-h: fix MPI1 compatibility Makefile

ggouaillardet:
appends MPI1 compatible source files instead of redefining all the source files
fix a typo from open-mpi/ompi@89da9651bb2fe26900ac16aa074d0efb11073251

### DIFF
--- a/ompi/mpi/fortran/mpif-h/profile/Makefile.am
+++ b/ompi/mpi/fortran/mpif-h/profile/Makefile.am
@@ -15,7 +15,7 @@
 # Copyright (c) 2011-2013 Universite Bordeaux 1
 # Copyright (c) 2013-2014 Los Alamos National Security, LLC. All rights
 #                         reserved.
-# Copyright (c) 2015-2017 Research Organization for Information Science
+# Copyright (c) 2015-2018 Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


hjelmn:
Hah, shows how much i check fortran.

hjelmn:
Thanks for catching that. Its going to cause some MTT errors because it won't be in before the tarballs are generated but MTT will be a mess tonight anyway. Need to update more of the ompi-test repo to not use removed functions.

ggouaillardet:
we also need to revamp the detection of the `__error__` attribute.
on a stock CentOS 7 (gcc 4.8), I always get warnings such as
```
/home/gilles/local/ompi-master/include/mpi.h:423:9: warning: ‘__error__’ attribute ignored [-Wattributes]
         __mpi_interface_removed__("MPI_Comm_errhandler_fn was removed in MPI-3.0; use MPI_Comm_errhandler_function instead");
```

(it seems `__attribute__((__error__(msg)))` only causes an annoying warning instead of a failure.

with GNU compilers, we could try to use  `-Werror` in order to force configure `#define OPAL_HAVE_ATTRIBUTE_ERROR 0`

meanwhile, I will `opal_cv___attribute__error=0 configure ...`

PR #5213, opened 2018-06-02
Enable the PMIx ompi/rte component

rhc54:
Includes updates from PR #4677 for update to hwloc 2.0.1

Get the OMPI rte/pmix component working. This was tested using PRRTE as the RM, configuring OMPI using:

* autogen --no-orte

* with external libevent, external hwloc, and external PMIx master

* configuring PMIx master with the same libevent and hwloc

* execute the application using PRRTE's "prun" launcher, which has the same cmd line as ORTE's mpirun

Note that PMIx master appears to have a bug in the event notification system that caches job termination events. Thus, the first execution runs fine, but subsequent executions cause an "abort" when the OMPI default error handler is invoked upon notification of the prior job's termination. Will work that separately.


bgoglin:
Can you rebase on top of the master branch of bgoglin/ompi to get hwloc 2.0.1 for real, as well as my updates to OPAL for 2.0.x?

rhc54:
@bgoglin Is that going to conflict with the hwloc-related commits already here? If so, perhaps I should revert those and just take your branch?

bgoglin:
Yeah, just remove them, and take mine (Gilles' commit updating to hwloc 2.0.0 will be replaced with my direct update to 2.0.1, and Jeff's commit will come back from my branch). Hopefully your "Default to bind-to socket" won't be needed anymore.

rhc54:
@bgoglin Okay, done! And I was indeed able to leave out the change in default binding. Thanks!

Issue #5084, opened 2018-04-19
Call to MPI_Finalized hangs when called inside of MPI_Finalize


AndrewGaspar:
## Background information

My program creates some keyvals with delete functions. Inside one of the delete functions, a call to `MPI_Finalized` is made. If an attribute associated with the keyval is deleted during the call to `MPI_Finalize`, the call to `MPI_Finalized` hangs because it attempts to acquire (what I assume is) the same lock used to check if MPI is finalized. ~~It's not clear from the spec that `MPI_Finalized` is disallowed when called in the course of an `MPI_Finalize` call.~~ See EDIT.

I verified the recursive lock issue in a debugger.

This issue is low priority - if the delete function is being called, it's obvious that MPI is not yet finalized.

EDIT: It seems that the standard explicitly requires MPI_Finalized return false in this situation, at least for MPI_COMM_SELF. This issue reproduces even for MPI_COMM_SELF (see below code).

> ### 8.7.1 Allowing User Functions at Process Termination
> There are times in which it would be convenient to have actions happen when an MPI process finishes. For example, a routine may do initializations that are useful until the MPI job (or that part of the job that being terminated in the case of dynamically created processes) is finished. This can be accomplished in MPI by attaching an attribute to MPI_COMM_SELF with a callback function. When MPI_FINALIZE is called, it will first execute the equivalent of an MPI_COMM_FREE on MPI_COMM_SELF. This will cause the delete callback function to be executed on all keys associated with MPI_COMM_SELF, in the reverse order that they were set on MPI_COMM_SELF. If no key has been attached to MPI_COMM_SELF, then no callback is invoked. The “freeing” of MPI_COMM_SELF occurs before any other parts of MPI are a↵ected. **Thus, for example, calling MPI_FINALIZED will return false in any of these callback functions.** Once done with MPI_COMM_SELF, the order and rest of the actions taken by MPI_FINALIZE is not specified.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
```
> mpirun --version
mpirun (Open MPI) 3.0.0

Report bugs to http://www.open-mpi.org/community/help/
```

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

On Mac: [Spack](https://spack.io)
On CentOS7: system provided OpenMPI

### Please describe the system on which you are running

* Operating system/version: MacOS High Sierra, CentOS7
[...]

PR #5217, opened 2018-06-03
v2.x: Fix MPI_FINALIZED_HANG

jsquyres:
Per MPI-3.1:8.7.1 p361:11-13, it's valid for MPI_FINALIZED to be
invoked during an attribute destruction callback (e.g., during the
destruction of keyvals on MPI_COMM_SELF during the very beginning of
MPI_FINALIZE).  In such cases, MPI_FINALIZED must return "false".

Prior to this commit, we hung in FINALIZED if it were invoked during
a COMM_SELF attribute destruction callback in FINALIZE.  See
https://github.com/open-mpi/ompi/issues/5084.

This commit converts the MPI_INITIALIZED / MPI_FINALIZED
infrastructure to use a single enum (ompi_mpi_state, set atomically)
to represent the state of MPI:

- not initialized
- init started
- init completed
- finalize started
- finalize past COMM_SELF destruction
- finalize completed

The "finalize past COMM_SELF destruction" state is what allows us to
return "false" from MPI_FINALIZED before COMM_SELF has been fully
destroyed / all attribute callbacks have been invoked.

Since this state is checked at nearly every MPI API call (to see if
we're outside of the INIT/FINALIZE epoch), care was taken to use
atomics to *set* the ompi_mpi_state value in ompi_mpi_init() and
ompi_mpi_finalize(), but performance-critical code paths can simply
read the variable without needing to use a slow call to an
opal_atomic_*() function.

Thanks to @AndrewGaspar for reporting the issue.

(cherry picked from commit 35438ae9b521e3726b62efa76645e4af605ea3e9)

Refs #5084

### DIFF
--- a/oshmem/runtime/oshmem_shmem_finalize.c
+++ b/oshmem/runtime/oshmem_shmem_finalize.c
@@ -3,6 +3,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2018      Cisco Systems, Inc.  All rights reserved
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -70,8 +71,10 @@ int oshmem_shmem_finalize(void)
         /* Should be called first because ompi_mpi_finalize makes orte and opal finalization */
         ret = _shmem_finalize();
 
-        if ((OSHMEM_SUCCESS == ret) && ompi_mpi_initialized
-                && !ompi_mpi_finalized) {
+        int32_t state = ompi_mpi_state;
+        if ((OSHMEM_SUCCESS == ret) &&
+            (state >= OMPI_MPI_STATE_INIT_COMPLETED &&
+             state < OMPI_MPI_STATE_FINALIZE_PAST_COMM_SELF_DESTRUCT)) {
             PMPI_Comm_free(&oshmem_comm_world);
             ret = ompi_mpi_finalize();
         }
--- a/oshmem/runtime/oshmem_shmem_init.c
+++ b/oshmem/runtime/oshmem_shmem_init.c
@@ -148,13 +148,6 @@ int oshmem_shmem_init(int argc, char **argv, int requested, int *provided)
     if (!oshmem_shmem_initialized) {
         ret = ompi_mpi_init(argc, argv, requested, provided, true);
 
-        // It's posible that another thread is initializing MPI and
-        // has not completed yet.  Keep checking until it is
-        // completed.
-        while (ompi_mpi_state < OMPI_MPI_STATE_INIT_COMPLETED) {
-            usleep(1);
-        }
-
         if (OSHMEM_SUCCESS != ret) {
             return ret;
         }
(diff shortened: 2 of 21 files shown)


jsquyres:
@artpol84 @jladd-mlnx Can Mellanox please review at least the OSHMEM parts of this PR?  I had to port this from master, and it's a little different.  I'd appreciate if someone could actually test that I didn't break OSHMEM.  Thanks!

oshmem/runtime/oshmem_shmem_finalize.c
```diff
@@ -70,8 +71,10 @@ int oshmem_shmem_finalize(void)
         /* Should be called first because ompi_mpi_finalize makes orte and opal finalization */
         ret = _shmem_finalize();
 
-        if ((OSHMEM_SUCCESS == ret) && ompi_mpi_initialized
-                && !ompi_mpi_finalized) {
+        int32_t state = ompi_mpi_state;
```
artpol84: @jsquyres I see that atomics are used to change ompi_mpi_state, don't we need to use them when accessing it as well?
Or consideration is made that this will always be the same thread?

oshmem/runtime/oshmem_shmem_init.c
```diff
@@ -146,7 +146,7 @@ int oshmem_shmem_init(int argc, char **argv, int requested, int *provided)
     int ret = OSHMEM_SUCCESS;
 
     if (!oshmem_shmem_initialized) {
-        if (!ompi_mpi_initialized && !ompi_mpi_finalized) {
+        if (ompi_mpi_state < OMPI_MPI_STATE_INIT_COMPLETED) {
```
artpol84: Why not comparing to a not-initialized state here?
If initialization has started already it supposed to be started in some other thread. And doing ompi_init here will launch a parallel init which is not desired.
I'd expect to have a while look polling on a state (with proper memory barriers / atomics) to wait for the started init to complete.

oshmem/runtime/oshmem_shmem_finalize.c
jsquyres: See the comment starting at https://github.com/open-mpi/ompi/pull/5217/files#diff-b109613962ca7154bf7d060024f15fe2R196.

oshmem/runtime/oshmem_shmem_init.c
jsquyres: The new test is simply doing what the old test did (because `ompi_mpi_initialized` was not set until the end of `ompi_mpi_init()`).  I can change it to be `== OMPI_MPI_STATE_INIT_NOT_INITIALIZED`, since the new variable offers finer-grained control.

If you want it to check for `_STATE_INIT_STARTED` and have it spin for a while, that should probably be a separate PR.

artpol84:
@jsquyres so this PR will be discarded?

jsquyres:
No, the commits from #5234 will be added to this one (i.e., the commits on this PR are already on master).

jsquyres:
@artpol84 @bosilca Ready for review.

oshmem/runtime/oshmem_shmem_finalize.c
artpol84: :+1:

jsquyres:
bot:ibm:retest

jsquyres:
@hppritcha Good to go.

Issue #5223, opened 2018-06-05
ROMIO Lustre build fails with --enable-picky


benmenadue:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

master@8192eb9

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

* Operating system/version: CentOS 6.9 user-land
* Computer hardware: Fujitsu
* Network type: FDR

-----------------------------

## Details of the problem

Enabling picky-ness (`--enable-picky`) when configuring now makes the ROMIO builds fail in Lustre (`--with-io-romio-flags=--with-file-system=lustre+ufs`) with
```
../../../../../../../../ompi/mca/io/romio314/romio/adio/ad_lustre/ad_lustre_wrcoll.c: In function 'ADIOI_LUSTRE_Exch_and_write':
../../../../../../../../ompi/mca/io/romio314/romio/adio/ad_lustre/ad_lustre_wrcoll.c:423: error: implicit declaration of function 'MPI_Type_extent'
../../../../../../../../ompi/mca/io/romio314/romio/adio/ad_lustre/ad_lustre_wrcoll.c:504: error: implicit declaration of function 'MPI_Address'
```

Since these are both MPI1 functions, I think this is because of #5127.

jsquyres:
FYI @PeterGottesman

PR #5224, opened 2018-06-05
io/romio314: Replace deprecated MPI-1 functions

benmenadue:
Replace `MPI_Address` with `MPI_Get_address` and `MPI_Type_extent` with `MPI_Type_get_extent` in io/romio314. This addresses #5223 and allows ROMIO to build without needing `--enable-mpi1-compat`.

However, I could only test the changes to the Lustre ADIO - I don't have access to the other filesystems to test these.


jjhursey:
ok to test

hjelmn:
Hmm, guess i missed these ones. Looks good.

PR #5225, opened 2018-06-05
coll/tuned: Fix typo of an MCA value

yukiM-fj:
coll/tuned: Fix typo at coll_tuned_reduce_scatter_decision.c
"recursive_halfing" -> "recursive_halving"
Signed-off-by: Yuki ARAKAWA(MATSUMOTO) <yuki.matsumoto@jp.fujitsu.com>


kawashima-fj:
Hmm. This change introduces an incompatibility of the MCA parameter value.
Adding a synonym is a better solution (if MCA var enum supports it).

ok to test

hjelmn:
Closing since 3adf96fdb811ee2495affdc86a7be6c091de4e10 included this.

hjelmn:
@kawashima-fj Since the next release is v4.0.0 I *think* we are ok not supporting the typo name. I will check what others think.

PR #5227, opened 2018-06-05
PML/UCX: fixed hand on MPI_Finalize

hoopoepg:
fixes issue https://github.com/openucx/ucx/issues/2656

added flush for worker object to complete all pending operations


yosefe:
@hoopoepg pls port to v3.1.x

PR #5228, opened 2018-06-05
v3.0.x improved aarch64 support

hjelmn:
Adds patcher support for aarch64 and fixes debug builds on aarch64 and power.


shamisp:
👍

PR #5229, opened 2018-06-05
Preferentially use /dev/shm on Linux

hjelmn:
These commits add MCA variables to control where the osc/rdma and osc/sm components store shared memory segments. This variable is set to /dev/shm if it exists. This works around possible performance issues if the session directory is not a tmpfs.

Master will move this support into opal/shmem.


bosilca:
I somehow assumed that all shared memory backend should be located on the same support, and treat equally the vader and OSC shared memory.

hjelmn:
I agree. I plan to handle things differently in v4.0.0. There problem is if will change MCA variable names which can't happen within a release series. PR to master coming soon.

PR #5244, opened 2018-06-07
Fix openib support for RoCE HCA

nmorey:
We encountered at least 2 RoCE HCA that did not work out of the box with the openib BTL as they need specific parameters

These patches add support for them.
They should be added to the v2 and v3 maintenance branches aswell

### DIFF
--- a/opal/mca/btl/openib/mca-btl-openib-device-params.ini
+++ b/opal/mca/btl/openib/mca-btl-openib-device-params.ini
@@ -234,6 +234,11 @@ use_eager_rdma = 1
 mtu = 4096
 max_inline_data = 0
 
+[QLogic FastLinQ QL41000]
+vendor_id = 0x1077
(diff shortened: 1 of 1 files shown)


rhc54:
ok to test

jsquyres:
@nmorey Which release versions do you need this on?  2.1.x? 3.0.x? 3.1.x?  All of the above?

nmorey:
@jsquyres We really need it for v2.1.x as this is the branch we use in SLES.
But it seems simple enough to backport on all branches :)

jsquyres:
Ya, agreed -- was just curious where you needed it.

nmorey:
FYI, we pinged Broadcom and it seems they have other RoCE HCA that needs an entry in the ini file. They should send a PR soon :)

jsquyres:
@nmorey Exxxcellent.  Thanks!

PR #5253, opened 2018-06-08
V2.x: Fix openib support for RoCE HCA

PeterGottesman:
Copy @nmorey's  PR #5244 to V2.x.

> We encountered at least 2 RoCE HCA that did not work out of the box with the openib BTL as they need specific parameters


hppritcha:
bot:ompi:retest

PR #5263, opened 2018-06-12
ompio: fix abstraction

ggouaillardet:


ggouaillardet:
@edgargabriel this is still work in progress, I have to double check everything is correctly labelled, copyrights were not accidentally removed, and I might move a few more things down to `common/ompio`.

edgargabriel:
Gilles, thanks first of all for your help on this. Some high level comments:
  - I am ok with the fixes into the Makefile.am changes that adds the common_ompio stuff to the components as a dependency
 - I am fundamentally ok with moving the aggregators.[ch] files from io/ompio to common/ompio. 
 - I do not really like the idea of moving all the ompio mca paramaters from io/ompio to common/ompio and having to change their names. I was hoping that the interfaces that I introduced to access them using a function that is accessible as a function pointer from the ompio file structure should be sufficient for that. All components accessed the mca parameters only through that mechanism, the exception was common/ompio. 
 - There are a couple of smaller things, but before diving into the details we should probably decide on the high level solutions.

ggouaillardet:
Got it, I will see how to avoid moving the MCA parameters from tomorrow

ggouaillardet:
@edgargabriel I fixed the abstraction violations, and kept the function pointer in order to access MCA parmams of the `io/ompio` component.
do we need pointers in each `ompio_file_t` (the new name for `mca_io_ompio_file_t` since it was moved to `common/ompio`) ? of could we use global pointers in `common/ompio` ?
parameters are accessed via the `OMPIO_MCA_GET()` macro, but I just realized the macro is incorrect, and will fix that tomorrow.

edgargabriel:
Gilles, thank you, I think it looks fundamentally good, but since this is now  a huge change (touches literally every file in the ompio universe), I will try to run tomorrow some tests to make sure that we are not overlooking something before merging it into master.

ggouaillardet:
I fixed a few more things, and the PR is now ready for test.

edgargabriel:
@ggouaillardet looks good, thanks. All of our tests as well as the hdf5 testsuite pass correctly. The only item that wasn't working was a minor issue in the compilation of the pvfs2 fs component, I filed a pr against this branch, other wise we are good to go.

Thanks!

PR #5271, opened 2018-06-14
opal/asm: fix 32-bit fetch and op atomics

hjelmn:
These functions were erroneously returing the new, not the old value.


hjelmn:
@jjhursey Can you set up CI to test --disable-builtin-atomics. This wasn't caught because no one was testing this :(.

PR #5272, opened 2018-06-14
configure: check for __thread support.

thananon:
Add configure to check if compiler support `__thread` (older gcc).  This will make it easier for components to use TLS and implement fallback. We already have `OPAL_C_HAVE__THREAD_LOCAL` but that only detect `_Thread_local` (gcc 4.9+).

I'm still not sure where to put the check but here seems to work? (I don't think its c11. please suggest)

### DIFF
--- a/config/opal_setup_cc.m4
+++ b/config/opal_setup_cc.m4
@@ -46,7 +46,7 @@ AC_DEFUN([OPAL_PROG_CC_C11_HELPER],[
     opal_prog_cc_c11_helper_CFLAGS_save=$CFLAGS
     CFLAGS="$CFLAGS $1"
 
-    OPAL_CC_HELPER([if $CC $1 supports C11 thread local storage], [opal_prog_cc_c11_helper__Thread_local_available],
+    OPAL_CC_HELPER([if $CC $1 supports C11 _Thread_local], [opal_prog_cc_c11_helper__Thread_local_available],
                    [],[[static _Thread_local int  foo = 1;++foo;]], [OPAL_C_HAVE__THREAD_LOCAL],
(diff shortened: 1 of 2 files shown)


thananon:
I'm totally new to this autotool configury and I dont know where to put it. 
Where would you suggest it should be?

> Also, _Thread_local is in gcc 4.9.x as well (https://gcc.gnu.org/wiki/C11Status)

RHEL still ships with 4.8 :( and I can see the performance gain from supporting this.

hjelmn:
Just mentioning so you can update the PR message (looks like I could). Yeah, and RHELL, thats the way they roll. Always with the old software...

hjelmn:
You can move the check down to near where we check for C99.

thananon:
How about here?

hjelmn:
Looks good to me.

hjelmn:
If the semantics of __thread are the same as __Thread_local you might want to take one extra step and define an opal_thread_local keyword. Then define OPAL_HAVE_THREAD_LOCAL to 1 if this is defined.

hjelmn:
That way it will be easy to use whichever is available.

thananon:
I see, that is indeed simpler. Let me try that.

hjelmn:
Looks good.

ggouaillardet:
is `__thread` a C11 feature ?

thananon:
I don't think so.

hjelmn:
No. It's a compiler extension that predates C11. From what I can tell the semantics are there same.

hjelmn:
Note that the __thread identifier it is not of the form _[A-Z]* (reserved identifiers)

PR #5282, opened 2018-06-17
pml_ucx: add option to use opal memhooks instead of ucx internal hooks

yosefe:


yosefe:
bot:retest

yosefe:
bot:retest

yosefe:
@hoopoepg pls take a look
@jladd-mlnx @bureddy @vspetrov FYI

hjelmn:
I wonder if this should go in a more general place since it applies to any piece of code using UCX (not just the pml). Maybe  opal/common or ompi/common?

yosefe:
@hjelmn we are going to extract UCX common code to opal/common/ucx, and this will also be there. This is the reason it's done only in pml_ucx for now (for now we rely on pml being initialized before spml/osc)

jladd-mlnx:
👍

Issue #5279, opened 2018-06-16
SPC missing a function?


rhc54:
Looks like something isn't quite right on master:
```shell
In file included from pbarrier.c:28:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from pialltoall.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from palltoall.c:36:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from pallgatherv.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from piallgatherv.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from pbsend.c:34:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: ‘ompi_spc_get_count’ declared ‘static’ but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^~~~~~~~~~~~~~~~~~
In file included from pialltoallv.c:34:0:
[...]

Issue #5273, opened 2018-06-14
ompi/spc: compiler warnings


thananon:
**Compiler version:** gcc 4.8.
This is compiled without SPC.

```../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from pireduce.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from prput.c:34:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from pireduce_scatter.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from prget.c:33:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from pireduce_scatter_block.c:35:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from prsend.c:33:0:
../../../../ompi/runtime/ompi_spc.h:153:12: warning: 'ompi_spc_get_count' declared 'static' but never defined [-Wunused-function]
 static int ompi_spc_get_count(const struct mca_base_pvar_t *pvar, void *value, void *obj_handle);
            ^
In file included from pscatter.c:35:0:
[...]

PR #5283, opened 2018-06-18
spc: remove ompi_spc_get_count() prototype from ompi_spc.h

ggouaillardet:
This function is only used in ompi_spc.c and is hence declared as static.
Remove its prototype from the header file in order to silence compiler warnings who will typically consider ompi_spc_get_count() as a declared but not defined function.

Fixes open-mpi/ompi#5279
Fixes open-mpi/ompi#5273


ggouaillardet:
@rhc54 @thananon @davideberius this is quite a simple fix indeed. `ompi_spc_get_count()` is only used in one file and it should have never been inside the header.
The compiler warnings can be safely ignored for the time being.

thananon:
Thank you very much @ggouaillardet .

PR #5294, opened 2018-06-19
coll_hcoll: register progress callback directly without a proxy

yosefe:


ompi/mca/coll/hcoll/coll_hcoll_module.c
```diff
@@ -290,7 +280,8 @@ mca_coll_hcoll_comm_query(struct ompi_communicator_t *comm, int *priority)
            mxm bcol in libhcoll needs world_group fully functional during init
            world_group, i.e. ompi_comm_world, is not ready at hcoll component open
            call */
-        opal_progress_register(mca_coll_hcoll_progress);
+        printf("hcoll_progress_fn=%p\n", hcoll_progress_fn);
```
bosilca: Extraneous debug.

ompi/mca/coll/hcoll/coll_hcoll_module.c
```diff
@@ -290,7 +280,8 @@ mca_coll_hcoll_comm_query(struct ompi_communicator_t *comm, int *priority)
...
+        opal_progress_register(hcoll_progress_fn);
```
bosilca: The code will be more readable if the progress registration was the last step of a successful initialization.

bosilca: In fact the progress function should only be registered when coll_module_enable is called.

yosefe: @jladd-mlnx @vspetrov what do you think? is it possible to move or is progress needed for some phase during initialization?

jladd-mlnx: @vspetrov do we need it early for non-blocking context create?

vspetrov: No, i don't think we need hcoll internal progress to be registered into opal before we launch first collective via hcoll.

yosefe:
removed the printf(), but i prefer to not move hcoll progress to module_enable now - will also need to add accounting of how many times the module was opened/closed/enabled, since it's done per communicator and progress register/unregister is global.

jladd-mlnx:
👍

PR #5296, opened 2018-06-19
Decrease the size of the TCP BTL proc structure.

### LINKED ISSUE #5292: MPI_Init_thread crashes when called inside a thread on OSX

tried both
[openmpi-3.0.0.tar.bz2](https://download.open-mpi.org/release/open-mpi/v3.0/openmpi-3.0.0.tar.bz2)
and
[openmpi-3.1.0.tar.bz2](https://download.open-mpi.org/release/open-mpi/v3.1/openmpi-3.1.0.tar.bz2)

```bash
curl -vLOJ https://download.open-mpi.org/release/open-mpi/v3.1/openmpi-3.1.0.tar.bz2
tar -xf openmpi-3.1.0.tar.bz2
cd openmpi-3.1.0
./configure --prefix=$HOME/local/openmpi
make -j8 all
make install
```

```
$ uname -a
Darwin matrix.local 17.5.0 Darwin Kernel Version 17.5.0: Fri Apr 13 19:32:32 PDT 2018; root:xnu-4570.51.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro14,3 Darwin
```

-----------------------------

## Details of the problem

`MPI_Init_thread` crashes when called inside a thread on OSX.

Consider a hello world program `a.cpp`:

```c++
#include <thread>

#include <mpi.h>

static void BackgroundThreadLoop()
{
    int provided;
    printf("before: %s:%d\n", __func__, __LINE__);
    MPI_Init_thread(nullptr, nullptr, MPI_THREAD_MULTIPLE, &provided);
    printf("after: %s:%d\n", __func__, __LINE__);
}

int main()
{
    bool use_thread = true;
    if (use_thread) {
        // crashed
        auto th = std::thread(BackgroundThreadLoop);
        th.join();
    } else {
        // not crash
        BackgroundThreadLoop();
    }
    MPI_Finalize();
    return 0;
}

```

```shell
~/local/openmpi/bin/mpicxx a.cpp
./a.out                                                 # OK
~/local/openmpi/bin/mpiexec -np 2 ./a.out               # Crashed
~/local/openmpi/bin/mpiexec -np 2 -mca btl ^tcp ./a.out # OK
```

After debug, I found that the program crashed at

line 820 of ompi/runtime/ompi_mpi_init.c (in openmpi-3.1.0.tar.bz2)
```c++
    ret = MCA_PML_CALL(add_procs(procs, nprocs));
```

by adding the following debug info:

```c++
    printf("before: %s:%d\n", __func__, __LINE__);
    ret = MCA_PML_CALL(add_procs(procs, nprocs));
    printf("after: %s:%d\n", __func__, __LINE__);
```

```
$ ~/local/openmpi/bin/mpiexec -np 2 ./a.out               # Crashed

before: BackgroundThreadLoop:8
before: BackgroundThreadLoop:8
before: ompi_mpi_init:820
before: ompi_mpi_init:820
mpiexec noticed that process rank 1 with PID 0 on node matrix exited on signal 4 (Illegal instruction: 4).
```

bosilca:
As this structure is allocated on the stack, it should be small enopugh
to never overcome the size of the stack (even for threads). Fixes #5292.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_proc.h
+++ b/opal/mca/btl/tcp/btl_tcp_proc.h
@@ -60,7 +60,7 @@ typedef struct mca_btl_tcp_proc_t mca_btl_tcp_proc_t;
 OBJ_CLASS_DECLARATION(mca_btl_tcp_proc_t);
 
 /*	the highest possible interface kernel index we can handle */
-#define MAX_KERNEL_INTERFACE_INDEX 65536
+#define MAX_KERNEL_INTERFACE_INDEX 64
 
 /*	the maximum number of kernel interfaces we can handle */
 #define MAX_KERNEL_INTERFACES 8


jsquyres:
@bosilca I'm all for reducing the size of the stack array.

But do we know for a fact that kernel interfaces indexes with be <= 64?  Specifically: do we know that kernel interface indexes are densely packed, starting at 0?

E.g., if the index to the array is the kernel interface index and its values may be sparse, should we change that data structure from an array to a hash map?

bosilca:
I can't find such information, but on the few machines I looked at it was the case. If the kernel was handling the interfaces with a hash table the name would be different.

Can you take a look on how the usnic is exposed by the kernel ? You have multiple virtual interfaces, if each one get it's own index, it might be possible to overcome the 64 limit.

jsquyres:
usNIC won't play much of role here (because usNIC doesn't create new IP interfaces -- it's basically just an OS-bypass method to get to the same/existing IP interfaces).

But I asked an expert internally here in Cisco.  His detailed answer is below.

I think for our purposes, the answer is "in practice, they are usually dense, but they do not have to be."  For example, in my environment, it is quite common for us to load/unload kernel drivers (i.e., create/destroy IP interfaces).  This will lead to holes.

I propose #5326 -- it uses a hash map instead of an array.  Then there's no issue about a) how many interfaces there are, or b) what the values of their kindexes are.

----

Usually they are dense, but not necessarily.

Here are two simple examples that lead to not-dense case:

1) Moving devices between namespaces
You create a network namespace (this does not happen only in the context of containers), and you move an interface from an already existent network namespace (like the initial one that you get when you boot) to the new one.
The kernel will try to keep the original index of the device you are moving, and it will change the index only if that index is already in use in the new namespace.
This operation can create a hole in the index list of both namespaces: in the first namespace because you are removing a device, and in the new namespace because you are adding a new device that already has an ifindex assigned).

2) creation/deletion of devices
Each time a new network device is created/registered you assign a new ifindex, and that happens for example:
    - when you load a driver
    - when you create a virtual device (bridge, bonding, vlan, vxlan, etx)
You delete ifindexes and potentially create holes in the ifindex list when:
    - you rmmod a driver
    - you delete a virtual  device

You may endup adding or removing devices also when you play with the SRIOV configuration of a device (adding or removing VFs and their representations).

In summary:
- the kernel tries to assign consecutive (dense) ifindexes and the routine that does so is pretty simple (see below).
- however, there are system configurations and events that may lead to not-dense lists of ifindexes.

Note that once you have created a hole in the list of assigned `ifindexes`, you won't reuse those (now) free indexes until you have rolled-over a "int" (`net->ifindex` below). This is the kernel fn that is used to create/assign an `ifindex` when a new network device is registered ("struct net" represents a network namespace, and `__dev_get_by_index()` is used to check if a given `ifindex` is already in use by another device):

```c
static int dev_new_index(struct net *net)
{
	int ifindex = net->ifindex;

	for (;;) {
		if (++ifindex <= 0)
			ifindex = 1;
		if (!__dev_get_by_index(net, ifindex))
			return net->ifindex = ifindex;
	}
}
```

jsquyres:
This PR was made obsolete by #5326 (i.e., a different PR that fixed the same problem a different way).

PR #5300, opened 2018-06-19
oob/ud: remove as it has bitrotted

hjelmn:


jsquyres:
We discussed this on the 2018-06-19 webex and couldn't find anyone who cared about this component any more (LANL, Mellanox, ...etc.).  No one is actively maintaining it, and there have been some isolated cases where it seems to be / might be causing problems (no one looked deeply to find out for sure).

*If no one cares about oob/ud by 3 July 2018 (i.e., 2 weeks from today), we'll merge this PR and remove oob/ud.*

PR #5301, opened 2018-06-19
Fix the no-disconnect test

rhc54:
A race condition exists based on whether or not the userdata object attached to a hwloc_obj_t has been initialized. These objects are setup whenever we scan for resources under that location. You therefore must not set a variable to the pointer to the userdata object and then call a function that will initialize the data in it - you need to set the variable after the function call, and protect against a NULL pointer


rhc54:
@PeterGottesman @jsquyres

Issue #5302, opened 2018-06-19
PMIx mutex lock deadlock in logging routines


PeterGottesman:
On master head pmix is failing in the pmix_mutex_lock() function, seemingly related to the pmix logging service. 

configured with:
```
$ ./configure --prefix=/home/jsquyres/bogus --with-libfabric=/home/jsquyres/bogus --with-usnic --enable-mpirun-prefix-by-default --enable-debug --enable-mem-debug --enable-mem-profile --disable-mpi-fortran --enable-debug --enable-mem-debug --enable-picky

```

stdout:
```
$ mpirun --host host1,host2 hello_c
pmix_mutex_lock(): Resource deadlock avoided
pmix_mutex_lock(): Resource deadlock avoided
```

stack trace:
```
(gdb) bt
#0  0x00002aaaac0c51f7 in raise () from /lib64/libc.so.6
#1  0x00002aaaac0c6a28 in abort () from /lib64/libc.so.6
#2  0x00002aaaae7c2cc2 in pmix_mutex_lock (m=0x2aaab4023540)
    at /home/jsquyres/git/ompi/opal/mca/pmix/pmix3x/pmix/src/threads/mutex_unix.h:142
#3  0x00002aaaae7c2f2a in localcbfunc (status=0, cbdata=0x2aaab4023510) at base/plog_base_stubs.c:52
#4  0x00002aaaafeb656c in localcbfn (status=0, cbdata=0x2aaab401f6d0) at plog_default.c:93
#5  0x00002aaaae6f2089 in opal_opcbfunc (status=0, cbdata=0x2aaab4023a70) at pmix3x_server_north.c:159
#6  0x00002aaaaad3ac2e in pmix_server_log_fn (requestor=0x2aaab04bb520, info=0x2aaab4023b10, 
    directives=0x2aaab4023b88, cbfunc=0x2aaaae6f2033 <opal_opcbfunc>, cbdata=0x2aaab4023a70)
    at orted/pmix/pmix_server_gen.c:1072
#7  0x00002aaaae6f7cb2 in server_log (proct=0x2aaab04bb760, data=0x2aaab4023610, ndata=1, 
    directives=0x2aaab40232e0, ndirs=1, cbfunc=0x2aaaafeb652e <localcbfn>, cbdata=0x2aaab401f6d0)
    at pmix3x_server_north.c:1196
[...]

PR #5303, opened 2018-06-20
Prevent thread lock when show_help msgs are emitted

rhc54:


rhc54:
Fixes #5302

rhc54:
@ggouaillardet Would you mind merging this when the silly CI is done? It's getting a little late out here, but I don't want people impacted longer than necessary. Same for #5304

PR #5307, opened 2018-06-20
PML/UCX: fixed hang on MPI_Finalize

hoopoepg:
- added async UCX progress thread to allow
  pending requests to complete

fixes https://github.com/openucx/ucx/issues/2682

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -78,11 +78,6 @@ mca_pml_ucx_module_t ompi_pml_ucx = {
 #define PML_UCX_REQ_ALLOCA() \
     ((char *)alloca(ompi_pml_ucx.request_size) + ompi_pml_ucx.request_size);
 
-typedef struct mca_pml_progress_data {
-    ucp_worker_h worker;
-    int          complete;
-} mca_pml_progress_data_t;
-
 
 static int mca_pml_ucx_send_worker_address(void)
 {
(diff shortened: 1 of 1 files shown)


hoopoepg:
@yosefe I also tested this fix over v3.1 - it works

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -440,8 +459,27 @@ int mca_pml_ucx_del_procs(struct ompi_proc_t **procs, size_t nprocs)
...
     ucp_worker_flush(ompi_pml_ucx.ucp_worker);
 
+    /* Some peers may be in active communication phase and
+     * require progress from current process. Unfortunately
+     * PMIX could not provide progress call during Fence, and
+     * we will create own progress thread for UCX */
+    progress_data.worker   = ompi_pml_ucx.ucp_worker;
+    progress_data.complete = 0;
+
+    thread_err = pthread_create(&progress_thr, NULL, mca_pml_async_ucx_progress,
```
yosefe: we should aim for a solution in UCX, not OMPI. This problem can happen in other MPI's/PGAS libraries as well

rhc54: Or you could use the non-blocking PMIx _Fence_nb so PMIx can progress during the fence call.

hoopoepg: right, but it will require significant re-implementation of async progress in UCX, not sure if we have time for this now

hoopoepg: @rhc54 good idea, will try

hoopoepg: @yosefe done
cherry-pick it into v3.1?

yosefe:
BTW shall we remove worker flush?

hoopoepg:
let me check, not sure...

hoopoepg:
@yosefe yes, it works without flush too

PR #5308, opened 2018-06-20
coll/base: add MPI_Bcast based on a scatter followed by an allgather

mkurnosov:
Implements `MPI_Bcast` using a binomial tree scatter followed by an recursive doubling allgather.


jsquyres:
ok to test

PR #5320, opened 2018-06-22
pml/ucx: silence a warning

ggouaillardet:
declare 'fenced' volatile in order to silence CID 1437465


ggouaillardet:
@hoopoepg FYI

PR #5326, opened 2018-06-23
btl/tcp: use a hash map for kernel IP interface indexes

### LINKED ISSUE #5292: MPI_Init_thread crashes when called inside a thread on OSX

jsquyres:
The giant size of the TCP proc struct is causing a problem in some
environments (because it is allocated on the stack), and it was too
big, anyway.

Instead, use a hash map.  That way, it starts small and can grow if it
needs to.  It also makes no assumptions about the values of the kernel
interface indexes.

Fixes #5292.

This is an alternative solution to #5292 (@bosilca initially posted PR #5296; this is an alternative).

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_proc.c
+++ b/opal/mca/btl/tcp/btl_tcp_proc.c
@@ -833,7 +833,7 @@ void mca_btl_tcp_proc_accept(mca_btl_tcp_proc_t* btl_proc, struct sockaddr* addr
         mca_btl_base_endpoint_t* btl_endpoint = btl_proc->proc_endpoints[i];
         /* We are not here to make a decision about what is good socket
          * and what is not. We simply check that this socket fit the endpoint
-         * end we prepare for the real decision function mca_btl_tcp_endpoint_accept. */ 
+         * end we prepare for the real decision function mca_btl_tcp_endpoint_accept. */
         if( btl_endpoint->endpoint_addr->addr_family != addr->sa_family) {
             continue;
         }
@@ -888,7 +888,7 @@ void mca_btl_tcp_proc_accept(mca_btl_tcp_proc_t* btl_proc, struct sockaddr* addr
     }
     /* In this case the connection was inbound to an address exported, but was not in a CLOSED state.
      * mca_btl_tcp_endpoint_accept() has logic to deal with the race condition that has likely caused this
-     * scenario, so call it here.*/     
+     * scenario, so call it here.*/
     if (found_match) {
         (void)mca_btl_tcp_endpoint_accept(match_btl_endpoint, addr, sd);
         OPAL_THREAD_UNLOCK(&btl_proc->proc_lock);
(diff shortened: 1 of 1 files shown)


bosilca:
I confirm that by removing and reinserting the same kernel module multiple times we obtain a sparse kernel index interfaces array as each instance gets a different kernel interface index. But this is an extremely unnatural usage scenario. In this context using a hash table seems like an overkill, but it does solve a potential issue.

jsquyres:
I think the main point is that we don't have to assume how many IP interfaces there will be (i.e., don't have a fixed-size array).  The fact that the hash map also saves us from the admittedly-unlikely scenario of large kernel index values is a (small) bonus.

Issue #5256, opened 2018-06-08
osc/rdma: cannot handle more than 1 completion at a time.


thananon:
I discussed this with @hppritcha and @hjelmn and just creating this issue for tracking.  We mentioned this in #5205 .

The problem is when osc_rdma do aggrregate flush, it is doing a foreach over a list and flushing them one by one but some completion might trigger the callback to free the list member and lead to segfault.

If we fix this problem, we will see some performance bump from the BTL.

thananon:
I just want to report that for some reason, I dont see any problem when handling more than 1 completion at a time anymore.

hjelmn:
I got rid of the aggregation code some time ago. It did little but complicate the code and game single-threaded bandwidth benchmarks.

jsquyres:
@thananon Should this issue be closed?  It looks like @hjelmn fixed the problem...?

thananon:
Yes, I can no longer reproduce the behavior.

PR #5329, opened 2018-06-25
osc/rdma: fix data race on teardown

hjelmn:
The osc/rdma module did not wait for all pending atomics to complete
before tearing down. This could lead to weird issues as the target
location may no longer be registered or allocated.

This commit also fixes an offset calculation issue in
ompi_osc_get_data_blocking ().

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma.h
+++ b/ompi/mca/osc/rdma/osc_rdma.h
@@ -269,6 +269,9 @@ struct ompi_osc_rdma_module_t {
 
     /** number of time a get had to be retried */
     unsigned long get_retry_count;
+
+    /** outstanding atomic operations */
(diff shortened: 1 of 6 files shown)


hjelmn:
@thananon Fix for the atomics issue.

thananon:
I tried the patch and got segfault on Win_allocate (invalid window) just for sockets provider. I'm investigating. Please hold of merging.

thananon:
Ignore my previous message. It was my stupidity confusing between ofi provider name 'sockets' and the btl 'tcp' . This seems to fix #5256 . 

I was hoping that this will also fix our tcp tear down problem where I have to put sleep(1) in component_close but it doesn't. :(

thananon:
I also didnt notice that it got merged already XD.

Issue #5270, opened 2018-06-14
"__mpi_interface_removed__" implementation emits unintended warnings


jsquyres:
With gcc 5.2.0 on master builds, when compiling MPI applications, I see lots of warnings from `mpi.h` about the `__mpi_interface_removed__` implementation.  For example:

```
/home/jsquyres/bogus/include/mpi.h:423:9: warning: ‘__error__’ attribute ignored [-Wattributes]
         __mpi_interface_removed__("MPI_Comm_errhandler_fn was removed in MPI-3.0; u
         ^
/home/jsquyres/bogus/include/mpi.h:425:9: warning: ‘__error__’ attribute ignored [-Wattributes]
         __mpi_interface_removed__("MPI_File_errhandler_fn was removed in MPI-3.0; u
         ^
```

Per https://github.com/open-mpi/ompi/pull/5210#issuecomment-393737332, it seems like we don't have strong-enough configury to do `__mpi_interface_removed__` properly.

edgargabriel:
with gcc 4.8 the compilation fails now entirely when setting the --enable-mpi1-compatibility flag, it finishes however correctly without this flag:

```
make[2]: Entering directory '/home/gabriel/ompi/ompi/ompi/mpi/fortran/use-mpi-tkr'
  PPFC     mpi.lo
pmpi-f90-removed-interfaces.h:25.26:
    Included at mpi.F90:69:

interface PMPI_Attr_delete
                          1
Error: Unexpected INTERFACE statement in INTERFACE block at (1)
pmpi-f90-removed-interfaces.h:115.53:
    Included at mpi.F90:69:

subroutine PMPI_Type_extent(datatype, extent, ierror)
                                                     1
Error: Symbol 'pmpi_type_extent' at (1) already has an explicit interface
pmpi-f90-removed-interfaces.h:116.33:
    Included at mpi.F90:69:

  integer, intent(in) :: datatype
                                 1
Error: Unexpected data declaration statement in INTERFACE block at (1)
pmpi-f90-removed-interfaces.h:117.32:
    Included at mpi.F90:69:

  integer, intent(out) :: extent
                                1
Error: Unexpected data declaration statement in INTERFACE block at (1)
pmpi-f90-removed-interfaces.h:118.32:
    Included at mpi.F90:69:

  integer, intent(out) :: ierror
                                1
Error: Unexpected data declaration statement in INTERFACE block at (1)
pmpi-f90-removed-interfaces.h:119.3:
[...]

PR #5332, opened 2018-06-25
mpi.h: fix warning with gcc

hjelmn:


PeterGottesman:
Does this resolve #5270?

jsquyres:
@PeterGottesman @hjelmn Yes, it looks like it does.

PR #5334, opened 2018-06-25
btl/ofi: progress now happens after a threshold.

thananon:
This commit changed the way btl/ofi call progress. Before, we force
progression with every rdma/atomic call. This gives performance boost in
some case and slow down on others. Now we only force progression after
some number of rdma calls which result in better performance overall.

Also added new MCA parameter 'mca_btl_ofi_progress_threshold' to set
the threshold number. The new default is 64.

Also:
Added FI_DELIVERY_COMPLETE to tx_rtx flags to ensure that the completion
is generated after the message has been received on the remote side.

Signed-off-by: Thananon Patinyasakdikul <thananon.patinyasakdikul@intel.com>

### DIFF
--- a/opal/mca/btl/ofi/btl_ofi.h
+++ b/opal/mca/btl/ofi/btl_ofi.h
@@ -48,6 +48,8 @@
 BEGIN_C_DECLS
 #define MCA_BTL_OFI_MAX_MODULES         16
 #define MCA_BTL_OFI_MAX_CQ_READ_ENTRIES 128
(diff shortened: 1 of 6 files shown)


thananon:
progress_threshold is supposed to be a constant. We keep track of number of outstanding ops. If the number reach progress_threshold, we force progress.

thananon:
I updated and make the code a bit cleaner. Also match the default progress_threshold with num_cqe_read. We can change num_cqe_read from 1 to something else now as osc/rdma progress problem seems to be resolved.

PR #5337, opened 2018-06-26
ompi/requests: implement generalized request extensions

ggouaillardet:


ggouaillardet:
@hppritcha @jjhursey @hjelmn FYI

jjhursey:
@ggouaillardet Thanks!
Did you mean to commit `ompi/mca/io/romio321/romio/2.diff`?

@PHHargrove You mentioned a use case to me in the past if generalized requests were exposed. Does this PR provide you enough functionality, or do you need more?

jsquyres:
@ggouaillardet We discussed this PR on the webex today.  Should these be in `ompi/mpiext` (and probably default to being enabled, as opposed to other MPI extensions)?

PHHargrove:
@ggouaillardet I am not 100% certain, but suspect that if this work is sufficient for ROM-IO then it also meets the needs I had in mind.

bosilca:
Why was this merged without opportunity for other developers to review and comment ? In addition to the few issues mentionned above, there is no documentation and no explanation about the need for such extensions.

ggouaillardet:
@jjhursey `2.diff` was never meant to be included, and this PR fixes that (as stated in the first commit message)

@bosilca `ROMIO` from `MPICH 3.2.1` uses the Generalized request extensions in order to implement non blocking collective read/write. This full rationale for extending MPI generalized requests can be found at http://www.mcs.anl.gov/papers/P1417.pdf (note the paper contains several typos in the prototypes)

@jsquyres if I move this into `ompi/mpiext`, I would have to `#include` and other file in order to use these extensions, right ? In this case, that would require (yet an other small) patch to ROMIO and I initially wanted to avoid that. But I get your point, and I will be happy to move this into `ompi/mpiext` if requested.

@bosilca from my point of view, this PR is about finalizing ROMIO 3.2.1 refresh (and not so much about adding a new feature to Open MPI). I clearly should have done a better job at documenting/explaining here. I do not recall any guidelines about giving an opportunity for other developers to review and comment before merging into master. How much time should be given before merging such a non trivial PR ?

bosilca:
@ggouaillardet the paper is more than 10 years old, and the proposed extension didn't make it's way outside MPICH (not to the MPI Forum discussions nor anywhere other than ROMIO). It remains unclear what is the benefit from OMPI perspective to follow the MPICH prototyping there ?

The lack of written guideline is not an implicit approval for pushing new functionality without discussion and without review. Giving time for discussion would have allowed us to correct some obvious mistakes in this patch, and make sure the code is not only consistent but also beneficial to OMPI overall.

ggouaillardet:
@bosilca @jsquyres I made #5353 in order to move forward.

A major issue with the current `master` is `--enable-grequest-extensions` is not passed to `romio`, so extended requests are never used.

#5353 fixes that, and also remove public `PMIX_...` types/subroutines related to generalized request extensions. Via a few macros, ROMIO uses internal Open MPI generalized request extensions.

Let's move the discussion to #5353

PR #5339, opened 2018-06-26
SHMEM/ATOMIC: refactoring of module API

hoopoepg:
- removed atomic-basic-specific operand from module API
- added own calls for add and swap operations
- minor optimization for UCX atomics


hoopoepg:
@yosefe here is first step to new atomics. could you look?

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -444,7 +444,7 @@ int mca_pml_ucx_del_procs(struct ompi_proc_t **procs, size_t nprocs)
     mca_pml_ucx_waitall(dreqs, &num_reqs);
     free(dreqs);
 
-    opal_pmix.fence_nb(NULL, 0, mca_pml_fence_complete_cb, &fenced);
+    opal_pmix.fence_nb(NULL, 0, mca_pml_fence_complete_cb, (void*)&fenced);
```
yosefe: how is it related to atomics?

oshmem/mca/atomic/basic/atomic_basic_module.c
```diff
@@ -91,6 +92,86 @@ int mca_atomic_basic_finalize(void)
...
+        shmem_quiet();
+    }
+
+    atomic_basic_unlock(pe);
+
+    return rc;
+}
+
+static inline
+int mca_atomic_basic_fop(void *target,
```
yosefe: looks like no need for "mca_atomic_basic_op", could implement only with "mca_atomic_basic_fop" and pass dummy buffer for "prev"?

oshmem/mca/atomic/ucx/atomic_ucx_cswap.c
```diff
@@ -36,30 +36,21 @@ int mca_atomic_ucx_cswap_inner(void *target,
...
-        status = opal_common_ucx_wait_request(status_ptr, mca_spml_self->ucp_worker);
-        if (UCS_OK == status) {
-            assert(NULL != prev);
-            memcpy(prev, &val, nlong);
-            if (4 == nlong) {
-                *(uint32_t*)prev = val;
-            } else {
-                *(uint64_t*)prev = val;
-            }
+    val        = (4 == size) ? *(uint32_t*)value : *(uint64_t*)value;
```
yosefe: how about using sizeof(uint32_t) instead of 4?

oshmem/mca/atomic/atomic.h
```diff
@@ -86,17 +86,25 @@ struct mca_atomic_base_module_1_0_0_t {
     opal_object_t super;
 
     /* Collective function pointers */
+    int (*atomic_add)(void *target,
+                      const void *value,
```
yosefe: how about this optimization: instead of passing pointers (for "value" and "cond"), pass them as uint64_t number.
will avoid pointer dereference and some branches in the mca_atomic code

oshmem/mca/atomic/atomic.h
```diff
@@ -86,17 +86,25 @@ struct mca_atomic_base_module_1_0_0_t {
...
+                      size_t size,
+                      int pe);
     int (*atomic_fadd)(void *target,
                        void *prev,
                        const void *value,
-                       size_t nlong,
-                       int pe,
-                       struct oshmem_op_t *op);
+                       size_t size,
```
yosefe: how about making "prev" a "uint64_t*"? will avoid some branches

oshmem/mca/atomic/atomic.h
hoopoepg: it will not help because original shmem API operates by pointers, and we will have to convert ptr->value early using same if/else (size) construct

oshmem/mca/atomic/atomic.h
hoopoepg: not sure I got your idea...
for now UCX implementation operates by prev value directly, without any casting to datatype or branches

oshmem/mca/atomic/basic/atomic_basic_module.c
hoopoepg: it will require additional memcpy in case of single_op
ok, done

oshmem/mca/atomic/ucx/atomic_ucx_cswap.c
hoopoepg: fixed

oshmem/mca/atomic/atomic.h
yosefe: but original shmem api uses values: for example https://www.open-mpi.org/doc/v3.0/man3/shmem_long_fadd.3.php - `long value`

oshmem/mca/atomic/atomic.h
yosefe: for example in cswap:
```
 if (4 == size) {
            *(uint32_t*)prev = val;
        } else {
            *(uint64_t*)prev = val;
```

oshmem/mca/atomic/basic/atomic_basic_module.c
```diff
@@ -91,6 +92,86 @@ int mca_atomic_basic_finalize(void)
...
+{
+    int rc = OSHMEM_SUCCESS;
+    long long temp_value = 0;
+
+    atomic_basic_lock(pe);
+
+    rc = MCA_SPML_CALL(get(target, size, (void*)&temp_value, pe));
+
+    op->o_func.c_fn((void*) value,
+            (void*) &temp_value,
```
yosefe: indentation

oshmem/mca/atomic/basic/atomic_basic_module.c
```diff
@@ -91,6 +92,86 @@ int mca_atomic_basic_finalize(void)
...
+    int rc = OSHMEM_SUCCESS;
+    long long temp_value = 0;
+
+    atomic_basic_lock(pe);
+
+    rc = MCA_SPML_CALL(get(target, size, (void*)&temp_value, pe));
+
+    memcpy(prev, (void*) &temp_value, size);
+
+    op->o_func.c_fn((void*) value,
```
yosefe: indentation

oshmem/mca/atomic/ucx/atomic_ucx_module.c
```diff
@@ -32,6 +32,95 @@ int mca_atomic_ucx_finalize(void)
...
+        ATOMIC_ERROR("[#%d] Type size must be 4 or 8 bytes.", my_pe);
+        return OSHMEM_ERROR;
+    }
+
+    ucx_mkey = mca_spml_ucx_get_mkey(pe, target, (void *)&rva);
+    status_ptr = ucp_atomic_fetch_nb(mca_spml_self->ucp_peers[pe].ucp_conn,
+                                     op, val, prev, size,
+                                     rva, ucx_mkey->rkey,
+                                     opal_common_ucx_empty_complete_cb);
+    status = opal_common_ucx_wait_request(status_ptr, mca_spml_self->ucp_worker);
```
yosefe: looks like status is checked twice here: in opal_common_ucx_wait_request and in opal_common_ucx_wait_request
can we eliminate the extra branch? maybe opal_common_ucx_wait_request() should return opal_status_t?

oshmem/mca/atomic/atomic.h
hoopoepg: ok, done

oshmem/mca/atomic/atomic.h
yosefe: this is not "user data", it's a local variable created in the shmem wrapper, which can always be 64-bit
check https://www.open-mpi.org/doc/v3.0/man3/shmem_long_fadd.3.php - user does not pass pointer to prev, the prev value is returned from the API call

hoopoepg: got it
done

oshmem/mca/atomic/atomic.h
```diff
@@ -86,17 +86,26 @@ struct mca_atomic_base_module_1_0_0_t {
...
+                       void *prev,
+                       uint64_t value,
+                       size_t size,
+                       int pe);
     int (*atomic_cswap)(void *target,
-                        void *prev,
-                        const void *cond,
-                        const void *value,
-                        size_t nlong,
+                        uint64_t *prev, /* prev is used internally by wrapper, we may
```
yosefe: let's do the same for other atomics (fadd, swap)

oshmem/mca/atomic/ucx/atomic_ucx_module.c
```diff
@@ -32,6 +32,84 @@ int mca_atomic_ucx_finalize(void)
...
+                       size_t size,
+                       int pe,
+                       ucp_atomic_fetch_op_t op)
+{
+    ucs_status_t status;
+    ucs_status_ptr_t status_ptr;
+    spml_ucx_mkey_t *ucx_mkey;
+    uint64_t rva;
+
+    if ((8 != size) && (4 != size)) {
```
yosefe: can we do this as assertion (to avoid in release mode)?

opal/mca/common/ucx/common_ucx.h
```diff
@@ -56,6 +50,33 @@ ucs_status_t opal_common_ucx_wait_request(ucs_status_ptr_t request, ucp_worker_h
...
+int opal_common_ucx_wait_request_opal_status(ucs_status_ptr_t request, ucp_worker_h worker)
+{
+    /* check for request completed or failed */
+    if (OPAL_LIKELY(UCS_OK == request)) {
+        return OPAL_SUCCESS;
+    } else if (OPAL_UNLIKELY(UCS_PTR_IS_ERR(request))) {
+        return OPAL_ERROR;
+    }
+
+    return opal_common_ucx_wait_request_internal(request, worker) == UCS_OK ?
```
yosefe: this is not helpful, there is still branch in '?:' operator, need to revert or eliminate it by changing opal_common_ucx_wait_request_internal

hoopoepg: right, but it happened after few progress loops are completed and it could not affect to overal time

oshmem/mca/atomic/atomic.h
hoopoepg: why?
in other calls there is no postprocessing of prev value - it is updated directly from ucx_atomic calls

oshmem/mca/atomic/ucx/atomic_ucx_module.c
hoopoepg: done

oshmem/mca/atomic/atomic.h
yosefe: i see.. but in latest patch there is also no postprocessing for cswap.. or did i miss it? maybe it's not needed at all

opal/mca/common/ucx/common_ucx.h
yosefe: ok

oshmem/mca/atomic/atomic.h
hoopoepg: in force of result value to uint64_t postprocessing of result value is not required anymore - we know that there is no memory corruption :)

ibm-ompi:

 Gist: https://gist.github.com/ba30cc041fa571ca7558e80ab4020581

ibm-ompi:

 Gist: https://gist.github.com/e6070d7eca9064b75c7a27ce04da4dbc

ibm-ompi:

 Gist: https://gist.github.com/27b36ed96c47445e3fbddeb44fcb3423

ibm-ompi:

 Gist: https://gist.github.com/36913118e68260ac0c41a541dd3b6ae1

hoopoepg:
@yosefe could you look?

oshmem/mca/atomic/atomic.h
```diff
@@ -33,6 +33,8 @@
 
 BEGIN_C_DECLS
 
+#define ATOMIC_PTR_2_INT(ptr, size) ((size) == 8 ? *(uint64_t*)(ptr) : *(uint32_t*)(ptr))
```
yosefe: let's give it oshem prefix: OSHMEM_ATOMIC_PTR_2_INT

opal/mca/common/ucx/common_ucx.h
hoopoepg: @yosefe actually result of ***_flush is used for diagnostic only... leave opal result only?

yosefe:
@hoopoepg yes let's keep only opal result, and push diagnostic message inside the common code (instead of all other places which call common_wait like osc_ucx)

ompi/mca/osc/ucx/osc_ucx_active_target.c
```diff
@@ -260,29 +248,19 @@ int ompi_osc_ucx_post(struct ompi_group_t *group, int assert, struct ompi_win_t
...
             /* do fop first to get an post index */
-            status = opal_common_ucx_atomic_fetch(ep, UCP_ATOMIC_FETCH_OP_FADD, 1,
-                                                  &result, sizeof(result),
-                                                  remote_addr, rkey, mca_osc_ucx_component.ucp_worker);
-            if (status != UCS_OK) {
-                opal_output_verbose(1, ompi_osc_base_framework.framework_output,
-                                    "%s:%d: ucp_atomic_fadd64 failed: %d\n",
-                                    __FILE__, __LINE__, status);
-            }
+            opal_common_ucx_atomic_fetch(ep, UCP_ATOMIC_FETCH_OP_FADD, 1,
```
yosefe: seems broken to not check the error code.. but it wasn't introduced with this PR

hoopoepg: @yosefe right, but assumed flush can't return error code (or it failed permanently which cause complete isconnect)

PR #5351, opened 2018-06-28
btl/tcp: fix hash map usage

jsquyres:
Fix two facepalms:

1. The "uint32" in the hash map functions refer to the *key* size, not
   the *value* size.  The values are always 64 bits.
2. Pass the straight value to the "set" functions -- not the pointer
   to the value.

@bosilca I don't know how this worked at all -- much less survived and passed all my testing (!).  I'll need to cherry pick this to all the release branches, too (it's only been merged on v2.x so far).

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_proc.c
+++ b/opal/mca/btl/tcp/btl_tcp_proc.c
@@ -290,7 +290,7 @@ static mca_btl_tcp_interface_t** mca_btl_tcp_retrieve_local_interfaces(mca_btl_t
      */
     for( idx = opal_ifbegin(); idx >= 0; idx = opal_ifnext (idx) ) {
         int kindex;
-        uint32_t index;
+        uint64_t index;
         bool skip = false;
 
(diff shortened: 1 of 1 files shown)


opal/mca/btl/tcp/btl_tcp_proc.c
```diff
@@ -344,7 +344,7 @@ static mca_btl_tcp_interface_t** mca_btl_tcp_retrieve_local_interfaces(mca_btl_t
         /* create entry for this kernel index previously not seen */
         if (OPAL_SUCCESS != rc) {
             index = proc_data->num_local_interfaces++;
-            opal_hash_table_set_value_uint32(&proc_data->local_kindex_to_index, kindex, &index);
+            opal_hash_table_set_value_uint32(&proc_data->local_kindex_to_index, kindex, (void**) index);
```
bosilca: Why void** ? The correct way to cast this is to leave the index as uint32_t, and then first casting it to an uintptr_t and then to a void* (```(void*)((uintptr_t)index)```)

PR #5353, opened 2018-06-29
io/romio321: make grequest extensions internal

ggouaillardet:


jjhursey:
We have 1 known person asking that we expose the generalized request interface from OMPI (@PHHargrove) so we should make sure that there is a `ompi/mpiext/grequest` portion of this PR that exposes the interface.

I don't have strong feelings on whether the whole implementation should move to `mpiext` (which would require maybe more extensive changes to romio) or just some thin wrappers that call the internal interface.
For a first pass, maybe just add the thin wrappers in `mpiext` that call the internal interface.

By default we don't expose this functionality, but a user can build with the configure option (I think this is correct) `--enable-mpi-ext=grequest`

PR #5359, opened 2018-06-29
v2.x: tcp/btl: fix a cast

jsquyres:
The current cast is *functional*, but isn't really the way it should
be done.  This commit makes the cast the way it should be done.


jsquyres:
@hppritcha Fixing my overly-hasty-but-fixing-the-segv merge from last night.

PR #5368, opened 2018-07-02
btl/vader: add support for atomics and emulated rdma

hjelmn:
This commit adds support for atomic operations as well as rdma for
systems without rdma support. This support is implemented using an
internal send tag.


jsquyres:
@PeterGottesman This might fix our osc MTT woes.  Can you give it a whirl after this is merged?

PR #5374, opened 2018-07-03
opal/fifo: fix 128-bit atomic fifo on Power9

hjelmn:

### DIFF
--- a/opal/class/opal_fifo.h
+++ b/opal/class/opal_fifo.h
@@ -86,8 +86,11 @@ static inline opal_list_item_t *opal_fifo_push_atomic (opal_fifo_t *fifo,
                                                        opal_list_item_t *item)
 {
     opal_counted_pointer_t tail = {.value = fifo->opal_fifo_tail.value};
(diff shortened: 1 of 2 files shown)


hjelmn:
@jsquyres Turns out that Power 9 can support 128-bit compare-exchange. It has LL/SC instructions for reading/writing quad words.

hjelmn:
FYI @gpaulsen. Targeted for v3.0.x and v3.1.x.

gpaulsen:
Our CI is power8, but I think Josh disabled make-check.
I'm out rest of the week, but next week I can try it on our power8 and power9 by hand.

jsquyres:
@gpaulsen Any news?  😄

gpaulsen:
I'll give it a go.

gpaulsen:
Build and runs fine on power9 + RHEL 7.5 + gcc 4.8.5

'make check' reported all PASSES (including opal_fifo).

NOTE: mpool_memkind was skipped because configure was not able to build it (no memkind.h at configure time).  I assume this is not related to this opal_fifo change.

Thanks again @hjelmn

PR #5385, opened 2018-07-06
v3.0.x: Update to PMIx v2.1.2

rhc54:


rhc54:
Although the MTT shows clean, there were a couple of memory scrams - not sure why those tests reported success. Will investigate

rhc54:
Found and fixed. Note that the test was segfaulting but was reported as successful because it was supposed to fail. This looks like a bad bug in MTT to me - tests that are supposed to fail but segfault should _not_ be reported as "success".

PR #5388, opened 2018-07-07
coll/base/gather_intra_binomial: fix MPI_IN_PLACE processing

mkurnosov:
The call of `MPI_Gather` with sendbuf and sendtype parameters equal to MPI_IN_PLACE and NULL correspondingly, produces the segmentation fault in the root process.

The problem is that sendtype is used even when sendbuf value is MPI_IN_PLACE. But according to the standard (page 150, line 37), sendtype and sendcount parameters should be ignored in this case.

### DIFF
--- a/ompi/mca/coll/base/coll_base_gather.c
+++ b/ompi/mca/coll/base/coll_base_gather.c
@@ -65,9 +65,6 @@ ompi_coll_base_gather_intra_binomial(const void *sbuf, int scount,
     COLL_BASE_UPDATE_IN_ORDER_BMTREE( comm, base_module, root );
     bmtree = data->cached_in_order_bmtree;
 
-    ompi_datatype_type_extent(sdtype, &sextent);
-    ssize = opal_datatype_span(&sdtype->super, (int64_t)scount * size, &sgap);
-
     vrank = (rank - root + size) % size;
 
     if (rank == root) {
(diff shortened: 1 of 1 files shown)


mkurnosov:
A simple reproducer:

```
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>

{
    int rank, comm_size;
    int root = 0;

    MPI_Init(&argc, &argv);
    MPI_Comm_size(MPI_COMM_WORLD, &comm_size);

    int *rbuf, sbuf = rank;
    
    if (rank == root) {
        rbuf = malloc(comm_size * sizeof(*rbuf));
        rbuf[root] = root;
    }
    
    if (rank == root) {
        MPI_Gather(MPI_IN_PLACE, 0, NULL, rbuf, 1, MPI_INT, root, MPI_COMM_WORLD);
    } else {
        MPI_Gather(&sbuf, 1, MPI_INT, NULL, 0, MPI_DATATYPE_NULL, root, MPI_COMM_WORLD);
    }
    
    if (rank == root) {
        for (int i = 0; i < comm_size; i++) {
            if (rbuf[i] != i) {
                fprintf(stderr, "Error: [%d]=%d need %d\n", i, rbuf[i], i);
                break;
            }
        }
    }
    
    if (rank == root)
        free(rbuf);
    MPI_Finalize();
    return 0;
}
```

jsquyres:
Test added to open-mpi/ompi-tests IBM test suite.  Thanks!

Issue #5383, opened 2018-07-06
MPI_Ialltoallv with inputs larger than 2GB


clydestanfield:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

Tested on v2.1.0, v2.1.3, and v3.1.0. All three versions appear to have the issue.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Installed from the source/distribution tarball. Compiled with gcc 4.9.1.

### Please describe the system on which you are running

* Operating system/version: Redhat 7.4
* Computer hardware: AWS r3.8xlarge nodes
* Network type: AWS

-----------------------------

## Details of the problem

We are using MPI_Ialltoallv for an image processing algorithm. When doing this we pass in an MPI_Type_contiguous with an MPI_Datatype of MPI_C_FLOAT_COMPLEX which ends up being the size of multiple rows of the image (based on the number of nodes used for distribution). In addition sendcounts, sdispls, resvcounts, and rdispls all fit within a signed int. Usually this works without any issues, but when we lower our number of nodes we sometimes see failures.

What we found is that even though we can fit everything into signed ints, line 528 of nbc_internal.h ends up calling a malloc with an int that appears to be the size of the (num_distributed_rows * num_columns  * sizeof(std::complex<float>)) which in very large cases wraps back to negative.  As a result we end up seeing “Error in malloc()” (line 530 of nbc_internal.h) throughout our output.

We can get around this issue by ensuring the sum of our contiguous type never exceeds 2GB. However, this was unexpected to us as our understanding was that all long as we can fit all the parts into signed ints we should be able to transfer more than 2GB at a time. It appears the issue is that the malloc call in nbc_internal.h cannot handle inputs larger than 2GB.

jsquyres:
Looks like this was fixed on master, v3.0.x, and v3.1.x a long time ago.  Closing.

PR #5390, opened 2018-07-09
coll/libnbc: fix integer overflow

ggouaillardet:
Use internal pack/unpack subroutines that operate on MPI_Aint instead of int
and hence solve some integer overflows.

Thanks Clyde Stanfield for reporting this issue.

Refs open-mpi/ompi#5383


hjelmn:
Please PR to v3.0.x and v3.1.x.

Issue #5031, opened 2018-04-06
Change configury logic to prefer external hwloc/libevent/pmix


jsquyres:
Based on [Dallas March 2018 face-to-face meeting notes](https://github.com/open-mpi/ompi/wiki/Meeting-Minutes-2018-03#everything-else), we want to shift the focus in v4.0.x away from the embedded hwloc, libevent, and pmix to external versions of these libraries.

Specifically, change the configury logic:

1. If user preference for external was specified:
   * Look for external support.
   * If external support is found and is compatible (e.g., the version/API/ABI/whatever is good), use it.
   * Otherwise, abort `configure`
1. If user preference for internal was specified:
   * Use it.
1. If no user preference was specified, look for external support.
   1. If external support is found:
       * If the external version is compatible (e.g., the version/API/ABI/whatever is good):
           * If the external version is `>=` the internal version, use the external version.
           * If the external version is `<` the internal version, emit a warning and use the internal version.
       * If the external version is *not* compatible:
           * Emit a warning and use the internal version.
   1. If external support is *not* found
       * Use the internal version.

Also add relevant `ompi_info` output somewhere indicating whether we are using internal or external versions, and give the reason why.

jsquyres:
The bullets above were initially created from notes from the Dallas March 2018 meeting.  Discussion welcome, especially before I start implementing.  😄

ggouaillardet:
What is the rationale for using the internal version if external version `<` internal version __but__ the external version is compatible ?

At first glance, I would simply print a warning and use the external version.

jsquyres:
Per discussion on 2018-07-10 webex, for PMIx:
* If external PMIx < v1.2.5, it is not compatible
* External PMIx v1.2.5 is *compatible*, but is not *desirable* because there are PMIx v2.0.x features that the OMPI code base uses.

jsquyres:
Note that the current status of master moved a few steps towards this scheme, but it also did a few things that are explicitly *not* in this scheme.

For example, if the external component is configured successfully, it doesn't even try with the internal component -- which is contradictory to https://github.com/open-mpi/ompi/issues/5031#issuecomment-379304676: "The thought there was that if the user didn't specify a preference, then we should just give them the latest version, regardless of whether it is internal or external."

So there is definitely still work to be done.

ggouaillardet:
IIRC, when the end user does not explicitly request anything, the external component version is tested, and is __not__ selected if its version is lower than the embedded component.

[...]

PR #5395, opened 2018-07-09
Change defaults to prefer external libevent/hwloc

jsquyres:
This is an internal commit I made on the way towards #5031.  It is very much a WIP kind of thing; I honestly don't even remember what works / what doesn't work / what still needs to be done.

I'm pushing it to github in case someone else has time to advance this work.

### DIFF
--- a/opal/mca/event/external/configure.m4
+++ b/opal/mca/event/external/configure.m4
@@ -1,8 +1,8 @@
 # -*- shell-script -*-
 #
-# Copyright (c) 2009-2013 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2009-2018 Cisco Systems, Inc.  All rights reserved
 # Copyright (c) 2013      Los Alamos National Security, LLC.  All rights reserved.
-# Copyright (c) 2015-2017 Research Organization for Information Science
+# Copyright (c) 2015-2018 Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 #
 # Copyright (c) 2017      Intel, Inc. All rights reserved.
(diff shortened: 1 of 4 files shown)


ibm-ompi:

 Gist: https://gist.github.com/934424b4952d6c1d149e2c11c26b9df9

ibm-ompi:

 Gist: https://gist.github.com/952001875ca2f1598321dd7b41edcb0d

ibm-ompi:

 Gist: https://gist.github.com/fe5652436261f3f0374ac153fc7a82a4

ggouaillardet:
@jsquyres i fixed this PR so it does not abort if `event/external` cannot be built and is not explicitly requested

ggouaillardet:
I am having some issues pushing my changes to GitHub ...

the inline patch below should fix the remaining issue

```patch
diff --git a/opal/mca/event/external/configure.m4 b/opal/mca/event/external/configure.m4
index e2122b5b69..450b086c5b 100644
--- a/opal/mca/event/external/configure.m4
+++ b/opal/mca/event/external/configure.m4
@@ -73,7 +73,8 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
 
     # Did the user want us to check for libevent in a specific location?
     AC_MSG_CHECKING([for external libevent in])
-    AS_IF([test "$with_libevent" != "external" && \
+    AS_IF([test -n "$with_libevent" && \
+           test "$with_libevent" != "external" && \
            test "$with_libevent" != "internal" && \
            test "$with_libevent" != "yes" && \
            test "$with_libevent" != "no"],
```

ggouaillardet:
@jsquyres as far as I am concerned, these commits can be squashed and the PR can be merged.

jsquyres:
Yep -- am doing so now.  Sorry for the giant delay.  We'll get this in before the v4.0.x branch.

jsquyres:
Hmm.  Actually, this m4 doesn't implement all the logic described in #5031.

We argued over that logic long and hard and finally concluded on the outline that is given in #5031.

ggouaillardet:
:bot:mellanox:retest

jsquyres:
bot:mellanox:retest

opal/mca/hwloc/external/configure.m4
```diff
@@ -146,6 +143,34 @@ AC_DEFUN([MCA_opal_hwloc_external_CONFIG],[
...
+                          [[
+#if HWLOC_API_VERSION < 0x00020000
+#error "hwloc API version is less than 0x00020000"
+#endif
+                          ]])],
+                      [AC_MSG_RESULT([yes])],
+                      [AC_MSG_RESULT([no])
+                       AC_MSG_WARN([external hwloc version is less than internal version (2.0)])
+                       AC_MSG_WARN([using internal hwloc])
+                       opal_hwloc_external_support=no])])
```
jsquyres: I'm not clear on why all this code moved up to the previous block.  The code was previously split into two distinct parts:

1. Do we have external hwloc at all.
2. If we have it, does that external hwloc have all the things we need?

This is not a deal breaker -- it's just odd that you moved *some* of the "does external hwloc have the things we need?" (i.e., the version check) but left other things in the lower block (i.e., the XML check).

opal/mca/hwloc/external/configure.m4
```diff
@@ -146,6 +143,34 @@ AC_DEFUN([MCA_opal_hwloc_external_CONFIG],[
...
+
+           AS_IF([test "$opal_hwloc_external_support" = "yes" &&
+                  test -z "$with_hwloc"],
+                 [AC_MSG_CHECKING([if external hwloc version is 2.0 or greater])
+                  AC_COMPILE_IFELSE(
+                      [AC_LANG_PROGRAM([[#include <hwloc.h>]],
+                          [[
+#if HWLOC_API_VERSION < 0x00020000
+#error "hwloc API version is less than 0x00020000"
```
jsquyres: Is there a way to have this version by reference instead of by value?  I.e., I imagine that this will be one more place to forget to update if/when we update the embedded version of hwloc.

opal/mca/hwloc/external/configure.m4
```diff
@@ -172,21 +197,9 @@ AC_DEFUN([MCA_opal_hwloc_external_CONFIG],[
...
-               [AC_LANG_PROGRAM([[#include <hwloc.h>]],
-                   [[
-#if HWLOC_API_VERSION < 0x00010500
-#error "hwloc API version is less than 0x00010500"
-#endif
-                   ]])],
-               [AC_MSG_RESULT([yes])],
-               [AC_MSG_RESULT([no])
-                AC_MSG_ERROR([Cannot continue])])
-
```
jsquyres: Per #5031, if external support is not available, we need to emit a message at the bottom of this configure.m4:

1. If the user explicitly asked for external support, print why the external hwloc is not suitable and then abort.
2. If the user did not explicitly ask for external support, print why the external hwloc is not suitable and then allow processing to continue (which will end up selecting the internal hwloc).

opal/mca/event/external/configure.m4
```diff
@@ -56,72 +56,57 @@ AC_DEFUN([MCA_opal_event_external_POST_CONFIG],[
...
+                                opal_event_libdir="$with_libevent/lib"],
+                               [AC_MSG_RESULT([not found])
+                                AC_MSG_WARN([Library directories were not found:])
+                                AC_MSG_WARN([    $with_libevent/lib64])
+                                AC_MSG_WARN([    $with_libevent/lib])
+                                AC_MSG_WARN([Please use --with-libevent-libdir to identify it.])
+                                AC_MSG_ERROR([Cannot continue])])])])],
+          [AC_MSG_RESULT([(default search paths)])])
+    AS_IF([test ! -z "$with_libevent_libdir" && test "$with_libevent_libdir" != "yes"],
+          [opal_event_libdir="$with_libevent_libdir"])
```
jsquyres: Shouldn't all of the above be similar to hwloc?  I.e., shouldn't the `external/configure.m4` between hwloc and libevent be pretty similar in structure?  The exact tests will be different, of course (for various libraries, functions, header files, versions, ...etc.), but the *structure* should be similar -- especially when checking for lib vs. lib64, external-vs-non-external, ...etc.

opal/mca/event/external/configure.m4
```diff
@@ -136,24 +121,41 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
...
+
+           AS_IF([test "$opal_event_external_support" = "yes" &&
+                  test -z "$with_libevent"],
+                 [AC_MSG_CHECKING([if external libevent version is 2.0.21 or greater])
+                  AC_COMPILE_IFELSE(
+                      [AC_LANG_PROGRAM([[#include <event.h>]],
+                          [[
+#if _EVENT_NUMERIC_VERSION < 0x02001500
+#error "libevent API version is less than 0x02001500"
+#endif
```
jsquyres: This needs another version check to ensure it is >= the internal version of libevent (just like hwloc).

jsquyres:
bot:mellanox:retest

opal/mca/hwloc/external/configure.m4
ggouaillardet: I do not know ... strictly speaking, this is not even the library version (e.g. 2.0.1) but the API version (e.g. 2.0.0).

@bgoglin is there any way to test the __library__version instead of the API version at compile time ?
for example, `libevent` defines `_EVENT_NUMERIC_VERSION`.

opal/mca/hwloc/external/configure.m4
ggouaillardet: This is fixed in the latest commit

opal/mca/event/external/configure.m4
ggouaillardet: at first glance, my answer is yes.

But this is a revamp that is not in the scope of this PR.
Should `hwloc` or `libevent` be the reference ?

opal/mca/event/external/configure.m4
ggouaillardet: the purpose of this test is to ensure this.
(and unlike in `hwloc` we do not explicitly require a minimal external version for `libevent`)

opal/mca/hwloc/external/configure.m4
rhc54: @ggouaillardet Don't you also need to test that ```$with_hwloc``` wasn't given without any path (and so has a ```yes``` value)? I thought that ```test -z``` just means it wasn't given at all.

ggouaillardet: @rhc54 you are right, and I just pushed a commit that fixes that.

rhc54:
Looks good to me - @jsquyres is on vacation. I'd go ahead and merge it - the revamp is definitely outside scope and IMO not worth the trouble.

bgoglin:
@ggouaillardet (replying to a comment that might have been lost between patch updates) There's hwloc_get_api_version() which returns the HWLOC_API_VERSION built inside the lib.

ggouaillardet:
@bgoglin I want to test the hwloc version at configure time, so the `HWLOC_API_VERSION` macro is a better fit than the `hwloc_get_api_version()` subroutine (it would required to link and run a test program, and that would not work when cross-compiling).

My main concern is `HWLOC_API_VERSION` is an __API__ version and __not__ a __library__ version.

For example, Open MPI `master` embeds `hwloc v2.0.1`, but `HWLOC_API_VERSION=0x00020000`.
With this PR, that means we would prefer an external `hwloc 2.0.0` rather than the internal ` hwloc 2.0.1`
(same __API__ version, but some missing bug fixes), so I think we need an other (set of) macros, such as
`HWLOC_LIBRARY_VERSION` or `HWLOC_MAJOR_VERSION`, `HWLOC_MINOR_VERSION` and `HWLOC_RELEASE_VERSION`.

ashleypittman:
We've just hit some build issues which I think are probably the result of this pull request, our environment is that we want to test pmix so are using upstream pmix along with ompi and therefore by necessary are also using libevent, but I think we're using the libevent rather than a specific version or build.

Our configure like is this:

./configure --with-platform=optimized --enable-orterun-prefix-by-default --prefix=/testbin/ompi --with-pmix=/testbin/pmix --disable-mpi-fortran --enable-contrib-no-build=vt --with-libevent=external --with-hwloc=/testbin/hwloc

And the build is then failing shortly after with this:

13:58:23 --- MCA component pmix:ext2x (m4 configuration macro)
13:58:23 checking for MCA component pmix:ext2x compile mode... dso
13:58:23 checking if external component is version 2.x... yes
13:58:23 configure: WARNING: EXTERNAL PMIX SUPPORT REQUIRES USE OF EXTERNAL LIBEVENT
13:58:23 configure: WARNING: LIBRARY. THIS LIBRARY MUST POINT TO THE SAME ONE USED
13:58:23 configure: WARNING: TO BUILD PMIX OR ELSE UNPREDICTABLE BEHAVIOR MAY RESULT
13:58:23 configure: error: PLEASE CORRECT THE CONFIGURE COMMAND LINE AND REBUILD

Please let me know if I should file this as a ticket directly.

rhc54:
Fix is in the oven

PR #5398, opened 2018-07-10
coll/libnbc: fix integer overflow

ggouaillardet:
Use internal pack/unpack subroutines that operate on MPI_Aint instead of int
and hence solve some integer overflows.

Thanks Clyde Stanfield for reporting this issue.

Refs open-mpi/ompi#5383

(back-ported from commit open-mpi/ompi@76292951e530cfa8b859deeb5a403cb222be32c2)


jsquyres:
@hjelmn Can you please review?

PR #5400, opened 2018-07-10
Default to external PMIx installation

rhc54:


rhc54:
@jsquyres I didn't add output of a warning in the summary about dynamic ops not being supported if we use PMIx v1.2.x - not entirely sure of a good way to do it. Can you please take a look?

Issue #5376, opened 2018-07-04
MPI_Comm_spawn inherit job options from the original job


bosilca:
Thank you for taking the time to submit an issue!

## Background information

Processes started with MPI_Comm_spawn inherit [all] parameters from the original job. As an example, if the original job was spawned with "-npernode 1", all future dynamic processes carry the same constraint. I could not figured out a way using MPI info keys to clean the environment to remove the constraints.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

Original issue has been discovered in master, and can be replicated on all 3.x branches (and certainly in older versions but I haven't checked).

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Developer installation, aka. git clone followed configure and make, with --enable-debug.

### Please describe the system on which you are running

Issue can be replicated in multiple environments, with and without RM, and with multiple networks.

rhc54:
Yeah, this has always been a debatable point. Here is the relevant code (recall that ```npernode``` is translated to a ```ppr``` value):
```c
    if (NULL == jdata->map->ppr && NULL != orte_rmaps_base.ppr) {
        jdata->map->ppr = strdup(orte_rmaps_base.ppr);
    }
    if (NULL != jdata->map->ppr) {
        /* get the procs/object */
        ppx = strtoul(jdata->map->ppr, NULL, 10);
        if (NULL != strstr(jdata->map->ppr, "node")) {
            pernode = true;
        } else {
            pernode = false;
        }
    } else {
        if (orte_rmaps_base_pernode) {
            ppx = 1;
            pernode = true;
        } else if (0 < orte_rmaps_base_n_pernode) {
            ppx = orte_rmaps_base_n_pernode;
            pernode = true;
        } else if (0 < orte_rmaps_base_n_persocket) {
            ppx = orte_rmaps_base_n_persocket;
            persocket = true;
        }
    }
    if (0 == jdata->map->cpus_per_rank) {
        jdata->map->cpus_per_rank = orte_rmaps_base.cpus_per_rank;
    }
```
You can see that we apply the MCA params given at the start of the job unless you override them. However, it is a one-to-one process - i.e., you can change the value of a specific MCA param directive, but you can't turn it "off".

[...]

PR #5406, opened 2018-07-10
Control inheritance of launch directives by child jobs

rhc54:
Do not have child jobs inherit launch directives unless requested to do so. This affects the map-by, rank-by, bind-to, npernode, pernode, npersocket, persocket, and cpus-per-rank directives. Values provided in the spawn call always take precedence - if a particular value isn't specified, then the ORTE defaults will be used if inheritance is not requested, and the values specified by MCA param will be used if inheritance is set.

### DIFF
--- a/orte/mca/rmaps/base/base.h
+++ b/orte/mca/rmaps/base/base.h
@@ -12,7 +12,7 @@
  * Copyright (c) 2011      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2011-2013 Los Alamos National Security, LLC.
  *                         All rights reserved.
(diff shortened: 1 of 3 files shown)


rhc54:
Fixes #5376 

@bosilca Please see if this does what you wanted.

rhc54:
@hoopoepg Looks like something is still wrong with the Mellanox Jenkins environment:
```
13:33:52 --------------------------------------------------------------------------
13:33:52 No PCI locality information could be found on at least one node. Please, upgrade BIOS to expose NUMA info.
13:33:52 
13:33:52   Node: jenkins03
13:33:52 
13:33:52 Open MPI will map the application by default (BYSLOT).
13:33:52 --------------------------------------------------------------------------

```

PR #5407, opened 2018-07-10
osc/rdma: fix bug introduced in b90c838

hjelmn:
This commit fixes an bug that was introduced back in 2016 which
impacts request-based RMA in some cases.


hjelmn:
@bosilca FYI. This goes all the way back to v2.x.

PR #5418, opened 2018-07-12
v3.1: opal/fifo: fix 128-bit atomic fifo on Power9

hjelmn:

(cherry picked from commit 8b090103e22e31cf336ff3b821a5a4041ff97c68)


gpaulsen:
built and tested on Power9.  This fixes the issue.  Thanks!

PR #5419, opened 2018-07-12
v3.0.x: opal/fifo: fix 128-bit atomic fifo on Power9

hjelmn:

(cherry picked from commit 8b090103e22e31cf336ff3b821a5a4041ff97c68)

### DIFF
--- a/opal/class/opal_lifo.h
+++ b/opal/class/opal_lifo.h
@@ -74,7 +74,32 @@ static inline bool opal_update_counted_pointer (volatile opal_counted_pointer_t 
     return opal_atomic_cmpset_128 (&addr->value, old.value, new_p.value);
 }
 
+__opal_attribute_always_inline__
+static inline void opal_read_counted_pointer (volatile opal_counted_pointer_t *addr, opal_counted_pointer_t *value)
+{
+    /* most platforms do not read the value atomically so make sure we read the counted pointer in a specific order */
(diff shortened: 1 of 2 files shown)


opal/class/opal_lifo.h
```diff
@@ -142,10 +167,8 @@ static inline opal_list_item_t *opal_lifo_pop_atomic (opal_lifo_t* lifo)
     opal_list_item_t *item;
 
     do {
-
-        old_head.data.counter = lifo->opal_lifo_head.data.counter;
-        opal_atomic_rmb ();
-        old_head.data.item = item = (opal_list_item_t*)lifo->opal_lifo_head.data.item;
+        opal_read_counted_pointer (&lifo->opal_lifo_head, &old_head);
```
gpaulsen: Do you need an opal_atomic_rmb() here after the opal_read_counted_pointer(), or is there a read memory barrier even if that function gets inlined?

hjelmn: Should be ok since there is a rmb in opal_read_counted_pointer and old_head.data.item is a data dependency for the next line. Shouldn't be re-ordered by the compiler.

gpaulsen: thanks for taking a second look.
looks good to me.

gpaulsen:
I tested and verified this fixes the issue on Power9.

PR #5420, opened 2018-07-12
osc/rdma fixes for btl/vader

hjelmn:


jsquyres:
@PeterGottesman Nathan tells me that this should fix our one-sided woes with vader.  Can you give it a whirl when you get a chance?

PR #5422, opened 2018-07-12
v3.0.x update vader to favor /dev/shm for backing files

hjelmn:


hppritcha:
adding news since there is a new mca parameter

PR #5431, opened 2018-07-16
OSC/pt2pt: disable pt2pt when UCX PML is selected

hoopoepg:

### DIFF
--- a/ompi/mca/osc/pt2pt/osc_pt2pt_component.c
+++ b/ompi/mca/osc/pt2pt/osc_pt2pt_component.c
@@ -35,6 +35,7 @@
 #include "osc_pt2pt_data_move.h"
 
 #include "ompi/mca/osc/base/osc_base_obj_convert.h"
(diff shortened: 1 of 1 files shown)


hoopoepg:
@yosefe will it work?

jsquyres:
We strongly discourage string comparing against specific component names.

Is there a capability that can be checked, instead?

hoopoepg:
@jsquyres technically it is possible, but it will require osc->pml module dependency.
for now this hack allows eliminate crash of 1-sided tests, I'm going to investigate what is wrong with pt2pt over UCX, and try to fix it

hoopoepg:
@jsquyres right, components may be identified by name only

jsquyres:
If it's a bug in the UCX PML, that's likely what should be fixed -- not putting a hackaround over here in this component.  Make sense?

hjelmn:
no. there are times when ucx can not do rdma and osc/rdma can not be used. fix the bug. this pr will not be accepted

PR #5434, opened 2018-07-17
v3.0.x: osc/sm: avoid filename collision when multiple windows share same CID

PeterGottesman:
This commit fixes an issue identified by MTT where we can have two
different sets of processes on the same node creating a shared memory
window with communicators sharing the same CID. To avoid this issue
the temporary filename now includes the creating processes vpid.

References #5363

(cherry picked from commit 35a75a6bf5356b47939b5a336ed0dabac443e086)
Signed-off-by: Peter Gottesman <pgottesm@cisco.com>


PeterGottesman:
@bwbarrett Can we merge this and #5433? I'm trying to minimize the space MTT is losing to coredumps.

PR #5441, opened 2018-07-17
MCA/COMMON/UCX: shift opal memhooks into common UCX

hoopoepg:

### DIFF
--- a/opal/mca/common/ucx/common_ucx.c
+++ b/opal/mca/common/ucx/common_ucx.c
@@ -62,6 +62,8 @@ OPAL_DECLSPEC void opal_common_ucx_mca_register(void)
     opal_common_ucx.output = opal_output_open(NULL);
     opal_output_set_verbosity(opal_common_ucx.output, opal_common_ucx.verbose);
 
+    mca_base_framework_open(&opal_memory_base_framework, 0);
+
     /* Set memory hooks */
     if (opal_common_ucx.opal_mem_hooks &&
         (OPAL_MEMORY_FREE_SUPPORT | OPAL_MEMORY_MUNMAP_SUPPORT) ==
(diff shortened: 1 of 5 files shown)


jladd-mlnx:
👍

opal/mca/common/ucx/common_ucx.c
```diff
@@ -39,9 +52,27 @@ OPAL_DECLSPEC void opal_common_ucx_mca_register(void)
...
     opal_output_set_verbosity(opal_common_ucx.output, opal_common_ucx.verbose);
+
+    /* Set memory hooks */
+    if (opal_common_ucx.opal_mem_hooks &&
+        (OPAL_MEMORY_FREE_SUPPORT | OPAL_MEMORY_MUNMAP_SUPPORT) ==
+        ((OPAL_MEMORY_FREE_SUPPORT | OPAL_MEMORY_MUNMAP_SUPPORT) &
+         opal_mem_hooks_support_level()))
+    {
+        MCA_COMMON_UCX_VERBOSE(1, "%s", "using OPAL memory hooks as external events");
+        mca_base_framework_open(&opal_memory_base_framework, 0);
```
yosefe: need to call this before using opal_mem_hooks_support_level()

opal/mca/common/ucx/common_ucx.c
```diff
@@ -52,6 +83,7 @@ OPAL_DECLSPEC void opal_common_ucx_mca_deregister(void)
     if (opal_common_ucx.registered) {
         return;
     }
+    opal_mem_hooks_unregister_release(opal_common_ucx_mem_release_cb);
```
yosefe: is it going to remove memory hooks for the first component that is used? what if there are several of them?

opal/mca/common/ucx/common_ucx.c
hoopoepg: fixed

opal/mca/common/ucx/common_ucx.c
hoopoepg: here is reference counter at line 83:
```
opal_common_ucx.registered--;
if (opal_common_ucx.registered) {
    return;
}
```
uninitialization will be executed on last deregister only (assuming that numbers of register and deregister are same)

yosefe: BTW is it possible we can have race condition on the counter? in case MPI_Win_create is called from multiple threads for example?

hoopoepg: should not be issue because pml will be initialized first and all next initializations will not take effect (and for OSC initialization it will be executed in common OSC initialization stack)

Issue #5443, opened 2018-07-17
OMPIO: MPI_File_delete always calls POSIX unlink.


gbossu:
The issue is present on master, and also 3.0.x and 3.1.x (at least)

I am currently working on bringing the support for DDN’s IME burst buffer technology to OMPIO (probably a future PR). While doing so, I realized that the MPI_File_delete is implemented incorrectly in OMPIO: no matter the filesystem or file type, POSIX’s unlink is called. The delete function defined in each fs component actually never gets called.

Basically, when mca_common_ompio_file_delete is called (for MPI_File_open, the equivalent would be mca_common_ompio_file_open), no fs component is selected, POSIX unlink is called straight away. 

Instead, what we should do is just the same as for the file_open call: a selection algorithm should be ran to determine which fs component should handle the deletion.

However, we can’t use the already defined mca_fs_base_delete (which performs the fs selection algorithm), because it requires an ompio file handle, and we only have a file name when deleting a file. We could create a new selection function that takes only a file name, but it would be a bad practice as :
 * It would create highly redundant code
 * For the selection algorithm, we use the query functions defined in each fs component. And these functions also require a file handle

We also can’t create a complete ompio file handle, because it requires a piece of information we don’t have: the communicator. I couldn’t find a really good way to address this issue.

I will file a PR with my work, which is a quick fix for this issue. But it’s far from perfect and could even be considered a hack. Help would be much appreciated to write a better patch.

gpaulsen:
This was merged to master and v4.0.x only.
Closing...

PR #5444, opened 2018-07-17
io/ompio: Call component-specific file_delete function instead of POSIX unlink

gbossu:
This PR is a quick fix for the bug referenced in this issue: #5443
Again, this is a work in progress, and it shouldn’t be considered a definitive solution. This is more of a workaround, and any help is appreciated to improve my work.

This is what I have done:
I “moved” the code of mca_common_ompio_file_delete to fs_base_file_delete, which is where it should go because this code is common to both UFS and Lustre.

Then, in the mca_common_ompio_file_delete function, instead of calling POSIX unlink, I do the following:
* Create an “incomplete” ompio file handle: It will basically only contain the file name, and an MPI_COMM_NULL communicator.
* Run the fs selection algorithm (mca_fs_base_delete) with this file handle
* Use the delete function of the selected component

However, because the query functions of the fs components used to always work with a complete file handle, no checks were performed on the file handle attributes. For this reason, I had to add checks on the communicator validity for each component’s query function.

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_file_open.c
+++ b/ompi/mca/common/ompio/common_ompio_file_open.c
@@ -14,6 +14,7 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2017      IBM Corporation. All rights reserved.
+ * Copyright (c) 2018      DataDirect Networks. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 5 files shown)


gbossu:
Also, can someone add the needed tags ? I can't add them myself

ompi/mca/common/ompio/common_ompio_file_open.c
```diff
@@ -480,26 +481,76 @@ int mca_common_ompio_file_delete (const char *filename,
...
+    ret = mca_fs_base_file_select (fh, NULL);
+    /* We don't need the file handle anymore */
+    free(fh);
+    if (OMPI_SUCCESS != ret) {
+        opal_output(1, "mca_fs_base_file_select() failed\n");
+        return ret;
+    }
+
+    ret = fh->f_fs->fs_file_delete (filename, NULL);
+    if (ret != OMPI_SUCCESS) {
```
edgargabriel: Please move the free(fh) statement down, after the the line where you are de-referencing fh in line 509.
Otherwise looks good.

edgargabriel:
the general idea looks good, there are two minor changes that I would ask to implement before merging.

gbossu:
Thanks for the review, I will update my code based on your suggestions.
And mca_fs_base_get_fstype is actually called, because I define the rank as OMPIO_ROOT when I create the file handle. But it's definitely not clear and I will follow your advice :+1:

jsquyres:
ok to test

hppritcha:
we use labels on PRs for merges into specific branches.

gbossu:
@edgargabriel can you have a look again ? :)

edgargabriel:
@gbossu looks good, thanks. Question: is it sufficient to have this fix going forward to Open MPI v4.0.0 or would you like to backport it to v3.1 and/or v3.0 branch?

edgargabriel:
@sydidelot The branching for Open MPI v4.0 has either happened yesterday or will happen today/very soon. I think we would probably still be able to bring at least this fix into the release, if this is of any value for you. Not entirely sure however whether a major new feature for v4.0 would be accepted at this point in time, or whether that would have to wait for the next major release. The v4.0 release is scheduled for the early fall (Mid September) if I remember correctly. Maybe the release managers @hppritcha and @gpaulsen can comment on this.

Regarding the v3.x releases, there will probably be some more bug-fix releases for the 3.1 and 3.0 series.

PR #5450, opened 2018-07-18
coll-base-allgather: fix MPI_IN_PLACE processing

mkurnosov:
The call of `MPI_Allgather` with sendbuf and sendtype parameters equal to `MPI_IN_PLACE` and `NULL` correspondingly, produces the segmentation fault.

The problem is that sendtype is used even when sendbuf value is `MPI_IN_PLACE`. But according to the standard, sendtype and sendcount parameters should be ignored in this case.

### DIFF
--- a/ompi/mca/coll/base/coll_base_allgather.c
+++ b/ompi/mca/coll/base/coll_base_allgather.c
@@ -90,7 +90,7 @@ int ompi_coll_base_allgather_intra_bruck(const void *sbuf, int scount,
                                           mca_coll_base_module_t *module)
 {
     int line = -1, rank, size, sendto, recvfrom, distance, blockcount, err = 0;
-    ptrdiff_t slb, rlb, sext, rext;
+    ptrdiff_t rlb, rext;
     char *tmpsend = NULL, *tmprecv = NULL;
 
     size = ompi_comm_size(comm);
@@ -98,9 +98,6 @@ int ompi_coll_base_allgather_intra_bruck(const void *sbuf, int scount,
 
     OPAL_OUTPUT((ompi_coll_base_framework.framework_output,
                  "coll:base:allgather_intra_bruck rank %d", rank));
-
-    err = ompi_datatype_get_extent (sdtype, &slb, &sext);
-    if (MPI_SUCCESS != err) { line = __LINE__; goto err_hndl; }
 
     err = ompi_datatype_get_extent (rdtype, &rlb, &rext);
     if (MPI_SUCCESS != err) { line = __LINE__; goto err_hndl; }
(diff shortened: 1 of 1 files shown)


mkurnosov:
Reproducer

```
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>

{
    int rank, comm_size;
    MPI_Init(&argc, &argv);
    MPI_Comm_size(MPI_COMM_WORLD, &comm_size);
    int count = 5;

    int *rbuf = malloc(count * comm_size * sizeof(*rbuf));
    if (!rbuf) {
        fprintf(stderr, "Error: malloc failed\n");
        MPI_Abort(MPI_COMM_WORLD, 1);
    }
    for (int i = 0; i < count; i++)
        rbuf[rank * count + i] = rank + i + 1;

    MPI_Allgather(MPI_IN_PLACE, 0, NULL, rbuf, count, MPI_INT, MPI_COMM_WORLD);

    for (int i = 0; i < comm_size; i++) {
        for (int j = 0; j < count; j++) {
            if (rbuf[i * count + j] != i + j + 1) {
                fprintf(stderr, "*** Validation error: [%d]=%d, need %d\n", 
                        i * count + j, rbuf[i * count + j], i + j + 1);
            }
        }
    }
        
    free(rbuf);
    MPI_Finalize();
    return 0;
}

```

jsquyres:
bot:mellanox:retest

jladd-mlnx:
bot:mellanox:retest

hppritcha:
@mkurnosov please open a PR against 4.0.x with a cherry-pick of 540c2d1

mkurnosov:
@hppritcha done, I reopend the PR #5474

aravindksg:
In conjunction with the fix in this PR, would the tuned allgather algorithms also need to be fixed?

I ran into a segfault with the above reproducer code on master and the fault was in `ompi_coll_tuned_allgather_intra_dec_fixed()` where we have a call to `ompi_datatype_type_size(sdtype, &dsize)`. With `sdtype` being `NULL`, it produces a segfault. It seems we  need to check for `MPI_IN_PLACE` here too before calling `ompi_datatype_type_size()` ? 

Also, `MPI_IN_PLACE` check would be needed in `ompi_coll_tuned_allgatherv_intra_dec_fixed()` as the spec states sendbuf and sendtype needs to be ignored for Allgatherv.

aravindksg:
@hppritcha , @bosilca : Any thoughts on the above comment related to the segfault in the tuned algorithm as well? If this is a valid issue, I can create a PR for review.

bosilca:
A fix for tuned would be more than welcome.

PR #5452, opened 2018-07-18
OSC/UCX: fixed hang on OSC init

hoopoepg:
- there worked progress was missed on startup which caused hang
  on one of ranks

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_component.c
+++ b/ompi/mca/osc/ucx/osc_ucx_component.c
@@ -110,10 +110,7 @@ static int component_register(void) {
 }
 
 static int progress_callback(void) {
-    if ((mca_osc_ucx_component.ucp_worker != NULL) &&
-        mca_osc_ucx_component.num_modules) {
-        ucp_worker_progress(mca_osc_ucx_component.ucp_worker);
-    }
+    ucp_worker_progress(mca_osc_ucx_component.ucp_worker);
     return 0;
 }
 
@@ -140,7 +137,6 @@ static int component_finalize(void) {
     assert(mca_osc_ucx_component.num_incomplete_req_ops == 0);
     if (mca_osc_ucx_component.env_initialized == true) {
         OBJ_DESTRUCT(&mca_osc_ucx_component.requests);
-        opal_progress_unregister(progress_callback);
         ucp_cleanup(mca_osc_ucx_component.ucp_context);
         mca_osc_ucx_component.env_initialized = false;
     }
@@ -250,7 +246,7 @@ static int component_select(struct ompi_win_t *win, void **base, size_t size, in
     ucs_status_t status;
     int i, comm_size = ompi_comm_size(comm);
     int is_eps_ready;
-    bool progress_registered = false, eps_created = false, env_initialized = false;
+    bool eps_created = false, env_initialized = false;
     ucp_address_t *my_addr = NULL;
     size_t my_addr_len;
     char *recv_buf = NULL;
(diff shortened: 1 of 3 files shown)


ompi/mca/osc/ucx/osc_ucx_component.c
```diff
@@ -45,7 +45,12 @@ ompi_osc_ucx_component_t mca_osc_ucx_component = {
         .osc_query = component_query,
         .osc_select = component_select,
         .osc_finalize = component_finalize,
-    }
+    },
+    .ucp_context            = NULL,
+    .ucp_worker             = NULL,
+    .env_initialized        = false,
+    .num_incomplete_req_ops = 0,
+    .init_in_progress       = 1
```
yosefe: why isn't mca_osc_ucx_component.num_incomplete_req_ops>0 ?

hoopoepg: to avoid branch

hoopoepg: actually one counter is not enough - we can't know when initialization is complete, and assuming that init complete when at least one operation is complete

yosefe: i mean, why when there is lack of progress, mca_osc_ucx_component.num_incomplete_req_ops is 0?
maybe we should call worker_progress in osc/ucs once in X times regardless of num_incomplete_req_ops?

hoopoepg: right, it doesn't progress till at least one operation is scheduled, and on startup worker progress is not called at all. I played with number of forced calls of progress, but this solution is not robust (how many calls should be done?) and makes code more complex.

I'm not sure if this solution is correct: on many ranks it may hang too (I tested 4x32 ranks only), the most visible fix is remove num_incomplete_req_ops evaluation from progress callback

yosefe: can we progress only if, for example, there is at least one window?

hoopoepg: it seems there is no existing global counters of windows, I can add it into UCX, but it will take time to test. and as I can see in most of cases in app where RMA is used at least one windows exists from start to finish of application's life

yosefe: @hoopoepg can we add such counter to osc_ucx?
especially if we have RMA over UD/TCP, would need progress on RX side all the time

hoopoepg: @yosefe ok, done

hoopoepg:
@yosefe anything else to fix?

ompi/mca/osc/ucx/osc_ucx_component.c
```diff
@@ -105,24 +110,18 @@ static int component_register(void) {
 }
 
 static int progress_callback(void) {
-    if (mca_osc_ucx_component.ucp_worker != NULL &&
-        mca_osc_ucx_component.num_incomplete_req_ops > 0) {
+    if ((mca_osc_ucx_component.ucp_worker != NULL) &&
```
yosefe: maybe we could register/unregister this progress callback on-demand? to avoid the overhead of function call

hoopoepg: could be issue if we unregister callback during call from callback, not sure how ompi will fail

yosefe: could we destroy the module during call from callback?

yosefe: yeah, seems like _opal_progress_register() is safe to call from the callback, for example ob1 is doing that (in "add_request_to_send_pending" pml_ob1_sendreq.h)

ompi/mca/osc/ucx/osc_ucx_component.c
```diff
@@ -812,6 +819,14 @@ int ompi_osc_ucx_free(struct ompi_win_t *win) {
     ompi_comm_free(&module->comm);
 
     free(module);
+    mca_osc_ucx_component.num_modules--;
```
yosefe: code duplication with line 646 - maybe add a helper function?

hoopoepg: @yosefe done

Issue #5442, opened 2018-07-17
Incorrect intent on Fortran 08 interfaces to MPI_All* routines


PhilippOtte:
The intent of the receive buffer is declared incorrectly as intent(in) in all MPI_All* routines in the mpi-f08 module on the master branch of version 3.1. According to version 3.1 of the MPI standard (appendix A.3.3), the intent of the receive buffers should be intent(inout) (default). Note that the intent of the receive buffer in MPI_Recv is set according to the standard.

If the receive buffer is of intent(in), the compiler is free to optimize the routine in illegal ways.
For example, the compiler is allowed to interchange the following two calls:
```Fortran
call MPI_Alltoall(sndbuf,100,MPI_INT,recvbuf,100,MPI_INT,MPI_COMM_WORLD)
call MPI_Allreduce(sndbuf,recvbuf,1,MPI_INT,MPI_MAX,MPI_COMM_WORLD)
```
into:
```Fortran
call MPI_Allreduce(sndbuf,recvbuf,1,MPI_INT,MPI_MAX,MPI_COMM_WORLD)
call MPI_Alltoall(sndbuf,100,MPI_INT,recvbuf,100,MPI_INT,MPI_COMM_WORLD)
```

The problem could easily be fixed by setting the intent of recvbuf in all MPI_All* routines to intent(inout).

PhilippOtte:
Thank you Gilles for the explanation. I will pose a PR with corrected function definitions for the MPI_All* and MPI_Iall* interfaces and their ompi interfaces.

PhilippOtte:
I filed a PR. Please note that I was very careful while editing but was not able to compile the files as I did not manage to compile required parts of the code on my machine.

jsquyres:
@ggouaillardet I made a v3.1.x (#5781) and v4.0.x (#5780) PRs corresponding the master PR you made (#5453).  But I ran into conflicts trying to make a v3.0.x PR.

Could you make a backport PR for v3.0.x?

jsquyres:
This was merged against all the release branches a while ago.

PR #5453, opened 2018-07-18
Corrections to Fortran 2008 interfaces

PhilippOtte:
Following changes were carried out in ompi/mpi/fortran/use-mpi-f08:
- corrected the intent of the send buffer on most signatures in the Fortran 2008 interface
- added the ASYNCHRONOUS attribute to send and receive buffers and other array dummy arguments in non-blocking collective calls (also in mod/mpi-f08-interfaces.F90)
- corrected attributes in the C bindings defined in mpi-f-interfaces-bind.h

**Important**: I was not able to compile other parts of the code and could not check whether files compile but I think it would be wasted effort not to create a PR

Refs #5442

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/allgather_f08.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/allgather_f08.F90
@@ -13,7 +13,8 @@ subroutine MPI_Allgather_f08(sendbuf,sendcount,sendtype,recvbuf,recvcount,recvty
    use :: mpi_f08_types, only : MPI_Datatype, MPI_Comm
    use :: ompi_mpifh_bindings, only : ompi_allgather_f
    implicit none
-   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf, recvbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE :: recvbuf
    INTEGER, INTENT(IN) :: sendcount, recvcount
    TYPE(MPI_Datatype), INTENT(IN) :: sendtype
    TYPE(MPI_Datatype), INTENT(IN) :: recvtype
--- a/ompi/mpi/fortran/use-mpi-f08/allgatherv_f08.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/allgatherv_f08.F90
@@ -14,7 +14,8 @@ subroutine MPI_Allgatherv_f08(sendbuf,sendcount,sendtype,recvbuf,recvcounts,&
    use :: mpi_f08_types, only : MPI_Datatype, MPI_Comm
    use :: ompi_mpifh_bindings, only : ompi_allgatherv_f
    implicit none
-   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf, recvbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE :: recvbuf
    INTEGER, INTENT(IN) :: sendcount
    INTEGER, INTENT(IN) :: recvcounts(*), displs(*)
    TYPE(MPI_Datatype), INTENT(IN) :: sendtype
--- a/ompi/mpi/fortran/use-mpi-f08/allreduce_f08.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/allreduce_f08.F90
@@ -13,7 +13,8 @@ subroutine MPI_Allreduce_f08(sendbuf,recvbuf,count,datatype,op,comm,ierror)
    use :: mpi_f08_types, only : MPI_Datatype, MPI_Op, MPI_Comm
    use :: ompi_mpifh_bindings, only : ompi_allreduce_f
    implicit none
-   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf, recvbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE :: recvbuf
    INTEGER, INTENT(IN) :: count
    TYPE(MPI_Datatype), INTENT(IN) :: datatype
    TYPE(MPI_Op), INTENT(IN) :: op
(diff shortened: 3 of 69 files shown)


jsquyres:
ok to test

ibm-ompi:

 Gist: https://gist.github.com/1d987ba950d7a00a542a6a6e1fa75a58

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
```diff
@@ -810,7 +811,8 @@ subroutine ompi_iallgather_f(sendbuf,sendcount,sendtype,recvbuf, &
                             recvcount,recvtype,comm,request,ierror) &
    BIND(C, name="ompi_iallgather_f")
    implicit none
-   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf, recvbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN), ASYNCHRONOUS :: sendbuf, recvbuf
```
ggouaillardet: `sendbuf` and `recvbuf` cannot be defined twice

jsquyres: I saw, that too.  I'm also getting a compile error that I'm puzzling over:

```
ialltoallw_f08.F90:35:53:

    call ompi_ialltoallw_f(sendbuf,sendcounts,sdispls,sendtypes,&
                                                     1
Error: Type mismatch in argument ‘sendtypes’ at (1); passed TYPE(mpi_datatype) to INTEGER(4)
```

jsquyres: Oops -- I originally pasted in the wrong compile error.  This is the right one:

```
  PPFC     ialltoallw_f08.lo
ialltoallw_f08.F90:35:53:

    call ompi_ialltoallw_f(sendbuf,sendcounts,sdispls,sendtypes(1)%MPI_VAL,&
                                                     1
Error: Rank mismatch in argument ‘sendtypes’ at (1) (rank-1 and scalar)
```

jsquyres: Ah, I see the problem.  Directed comment coming up...

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
```diff
@@ -935,10 +946,11 @@ subroutine ompi_ialltoallw_f(sendbuf,sendcounts,sdispls,sendtypes, &
...
    implicit none
-   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN) :: sendbuf, recvbuf
-   INTEGER, INTENT(IN) :: sendcounts(*), sdispls(*), recvcounts(*), rdispls(*)
-   INTEGER, INTENT(IN) :: sendtypes
-   INTEGER, INTENT(IN) :: recvtypes
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, INTENT(IN), ASYNCHRONOUS :: sendbuf
+   OMPI_FORTRAN_IGNORE_TKR_TYPE, ASYNCHRONOUS :: recvbuf
+   INTEGER, INTENT(IN), ASYNCHRONOUS :: sendcounts(*), sdispls(*), recvcounts(*), rdispls(*)
+   INTEGER, INTENT(IN), ASYNCHRONOUS :: sendtypes(*)
+   INTEGER, INTENT(IN), ASYNCHRONOUS :: recvtypes(*)
```
jsquyres: You changed the types of `sendtypes` and `recvtypes` from scalars to arrays here.  They were specifically scalars (in both `ompi_alltoallw_f` and `ompi_ialltoallw_f`) -- see the lengthy comment here: https://github.com/open-mpi/ompi/blob/master/ompi/mpi/fortran/use-mpi-f08/ialltoallw_f08.F90#L24-L33

Yes, it's a hack, but there did not seem to be another performant way to get the values from one call to the other (i.e., without allocating a temporary array from the heap [because this is ialltoallw, and we'll return before sendtypes/recvtypes are finished being used], individually copying all the elements from sendtypes/recvtypes to the temporary array, and then passing the temporary array to the underlying call.  Then there will need to be some kind of hook when the request completes to know to free that temporary array, too).

Is there a better way?

PhilippOtte:
Is there a safe way of signing off the unsigned commits or should I create a new branch with the changes in it and create a new PR?

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
PhilippOtte: Fortran allows you to obtain the address of the first element in an array via the intrinsic c_loc which returns a type(c_ptr) which is equivalent to a void pointer in C. There would be no copying involved and you would prevent the compiler from doing any unwanted magic there.

ibm-ompi:

 Gist: https://gist.github.com/f6d76331dd80495c769a46dabc151ac3

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
jsquyres: Can you add that to this PR?  I'm wasn't enough of a Fortran expert to get that to work back when I originally wrote this stuff -- this hack was the best that I could come up with.  😳

jsquyres:
You can `git rebase -i origin/master` to edit the existing commits on this PR.  You'll then need to force-push back up to your branch (because you changed git history) -- that's just about the ***only*** socially-acceptable time to git-force-push.

image

ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h
PhilippOtte: According to Modern Fortran explained by Metcalf, Reid and Cohen the proposed way of using c_loc may lead to invalid pointers. So, your hack is probably the safer way of doing things.

jsquyres:
@ggouaillardet Can you re-review?  All the problems have been fixed.

jsquyres:
I just rebased -- we're now good.  Just need a new review from @ggouaillardet

gpaulsen:
Question, is this going to break binary compatibility with the older compiled apps?   Even if this is more correct, if it breaks binary compatibility, this might trigger a new major release version bump.

ggouaillardet:
the same changes have to be applied to `ompi/mpi/fortran/use-mpi-f08/profile` too. I will prepare a PR from now.

Should this PR make it for `v4.0.0` ?

ggouaillardet:
@jsquyres I fixed the signatures for the `PMPI` subroutines and fixed `MPI_Ireduce_scatter[_block]()` parameters too. As far as I am concerned, this is good for primetime.

gpaulsen:
First round of answers from Fortran experts here say:
```
As far as I know, the intent attribute does not change the calling convention and the signature of the procedure.  It should not cause binary incompatibilities.
```
But they forwarded the question, and I'd like to give it another day or so if this can wait.

ggouaillardet:
@gpaulsen is this something stated in the standard ? I might be a bit too paranoid/pedantic here, but unless stated in the standard, we should theoretically verify this technical choice was made on all supported compilers.

PR #5466, opened 2018-07-23
Leave opal_event_external_support exposed as global var

rhc54:


rhc54:
Refs #5395

rhc54:
looks like IBM CI is down

PR #5472, opened 2018-07-24
v4.0.x: Prefer external hwloc and libevent

ggouaillardet:
(cherry picked from #5395 and #5466)

### DIFF
--- a/opal/mca/event/configure.m4
+++ b/opal/mca/event/configure.m4
@@ -1,6 +1,6 @@
 dnl -*- shell-script -*-
 dnl
-dnl Copyright (c) 2010-2013 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2010-2018 Cisco Systems, Inc.  All rights reserved
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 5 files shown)


hppritcha:
bot:mellanox:retest

hppritcha:
bot:retest

jsquyres:
These commits nominally provide some of the requested functionality.  But:

1. Lots of warnings for external hwloc -- see #5524.
1. There are cases with the external hwloc component that the external hwloc is not used but no explanation is provided as to why (but in other cases, a good explanation is provided).
1. libevent and hwloc are quite different in their handling of internal vs. external.  This may end up being confusing to users, and is going to lead to maintenance problems down the road, IMHO.  I don't know how these two stack up against pmix's internal vs. external handling.
1. README and/or NEWS needs to be updated to document this behavior.
1. When the external components are selected, the configury of the internal components is confusing.  From the output, it looks like the internal configury succeeds, but then it says it can't compile the component.  Here's both hwloc and libevent internal configury outputs in this case:
```
checking whether hwloc configure succeeded... yes
checking for infiniband/verbs.h... (cached) yes
checking if MCA component hwloc:hwloc201 can compile... no
...
configure: /bin/sh './configure' succeeded for opal/mca/event/libevent2022/libevent
checking if MCA component event:libevent2022 can compile... no
```

At a minimum, I think #5524 needs to be solved before this PR is merged.

jsquyres:
Items 1 (warnings with external hwloc) and 5 (internal hwloc/libevent not configured -- and notice is given to explain why -- if external hwloc/libevent are selected) have been addressed) have been addressed with two new commits on this PR.

I think items 2 (more ?AC_MSG_NOTICEs? when external is not selected) and 4 (README+NEWS) should be addressed before this PR can be merged.

Given current resource constraints, item 3 (a wholesale revamp / unification of how internal-vs-external components are selected -- e.g., perhaps some common m4 between event/hwloc/pmix?) is outside the scope of this PR.

PR #5473, opened 2018-07-24
MCA/ATOMIC: atomic_init renamed to atomic_startup

hoopoepg:
- there is C11 naming conflict - atomic_init is C macro
  which cause building issue


hjelmn:
:+1:

hjelmn:
Wouldn't hurt. Don't know if we will bring the C11 stuff there yet.

PR #5474, opened 2018-07-24
coll/base/allgather: fix MPI_IN_PLACE processing

mkurnosov:
The call of MPI_Allgather with sendbuf and sendtype parameters equal to MPI_IN_PLACE and NULL correspondingly, produces the segmentation fault.

The problem is that sendtype is used even when sendbuf value is MPI_IN_PLACE. But according to the standard, sendtype and sendcount parameters should be ignored in this case.

Signed-off-by: Mikhail Kurnosov mkurnosov@gmail.com
(cherry picked from commit 540c2d1)


hoopoepg:
bot:mellanox:retest

mkurnosov:
@hoopoepg thanks

bwbarrett:
Is this is a cherry pick off master?  If so, can you add hash of the parent commit to the log message?

PR #5476, opened 2018-07-24
MTL OFI: Fix Deadlock in fi_cancel given completion during cancel

nrspruit:
- If a message for a recv that is being cancelled gets completed after
the call to fi_cancel, then the OFI mtl will enter a deadlock state
waiting for ofi_req->super.ompi_req->req_status._cancelled which will
never happen since the recv was successfully finished.

- To resolve this issue, the OFI mtl now checks ofi_req->req_started
to see if the request has been started within the loop waiting for the
event to be cancelled. If the request is being completed, then the loop
is broken and fi_cancel exits setting
ofi_req->super.ompi_req->req_status._cancelled = false;

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -1013,8 +1013,11 @@ ompi_mtl_ofi_cancel(struct mca_mtl_base_module_t *mtl,
                      */
                     while (!ofi_req->super.ompi_req->req_status._cancelled) {
                         opal_progress();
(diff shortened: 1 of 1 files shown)


rhc54:
ok to test

hppritcha:
@nrspruit please open a PR against 4.0.x with a cherry pick of this commit so we can get this fix into the 4.0.0 release.

aravindksg:
@rhc54 , @hppritcha Since Matias is out on vacation, could one of you please merge this PR.

PR #5477, opened 2018-07-24
MTL OFI: send/isend split into blocking/non-blocking paths

nrspruit:
-Updated blocking send to directly call functionality and
set completion events expected to 0 initally. This allows for optimization for
providers that support fi_tinject up to larger sizes. This also reduces
latency on running the OFI mtl with smaller sizes without requiring
calls to progress given fi_tinject is required to complete the messaging
before returning and will not create any events in the Completion Queue.

-Updated non-blocking send to directly call fi_tsend and avoid calling
fi_tinject as the functionality should not wait on completions. This
resolves a bug where applications calling MPI_Isend can overrun the
TX buffer with small (inject) messages causing a deadlock. In addition
this improves performance in message rates by preventing
waiting on any size message to complete in non-blocking send messages.

-Created common ompi_mtl_ofi_ssend_recv function to post the ssend recv
which is common between isend and send code paths.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -238,24 +238,71 @@ ompi_mtl_ofi_isend_callback(struct fi_cq_tagged_entry *wc,
 }
 
 __opal_attribute_always_inline__ static inline int
-ompi_mtl_ofi_send_start(struct mca_mtl_base_module_t *mtl,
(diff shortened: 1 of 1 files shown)


rhc54:
ok to test

jsquyres:
I think @hppritcha's review is good enough here.

aravindksg:
@rhc54 , @hppritcha Since Matias is out on vacation, could one of you please merge

PR #5480, opened 2018-07-25
PML/SPML/UCX: init global objects using C99 style

hoopoepg:
- to avoid value mix used C99 style of object initializations

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -49,33 +49,33 @@
 #define MODEX_KEY "pml-ucx"
 
 mca_pml_ucx_module_t ompi_pml_ucx = {
-    {
-        mca_pml_ucx_add_procs,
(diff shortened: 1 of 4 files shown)


hoopoepg:
@brminich could you review?

yosefe:
I think we should keep UCX components in line with the rest of OMPI components, which use the ANSI style initialization.

hoopoepg:
@yosefe there are lot of components where C99 init used (bml/r2, collectives, fs, mxm, etc), as I can see C99 is allowed in OMPI coding guidelines and it looks like much simpler to read

yosefe:
@hoopoepg ok. Should we initialize the components (eg mca_pml_ucx_component) this way too?

hoopoepg:
Yep, why not...

hoopoepg:
@yosefe done, pml/spml component init are updated too

yosefe:
@hoopoepg pls pick to v4.0.x

PR #5484, opened 2018-07-26
Fix the multiple pe/proc option

rhc54:
Things got a little out of whack and we weren't actually processing the map-by modifiers, plus an error crept into the display of the binding report. So clean those up.

Thanks to @tonyreina for the error report

(cherry picked from commit bcdb1f45aca3f6dfab2646bfdba99775f728ca3b)


hppritcha:
@hjelmn thanks for reviewing

hppritcha:
@hjelmn ping - please review

PR #5486, opened 2018-07-26
Cleanup pmix selection and map-by modifiers

### LINKED ISSUE #5482: Older mpirun syntax works but newer one doesn't

## Background information

I'm running a python script with TensorFlow and Horovod to train a model across 4 nodes that are on the same network. If I use the newer syntax of `mpirun --map-by ppr:pe` the job fails. If I use the older format of `mpirun --map-by socket -cpus-per-proc` the job succeeds.

3.0.0

I download and built 3.0.0 from source.

* Operating system/version: CentOS 7
* Computer hardware: Intel Xeon 6148 CPU
* Network type: LAN TCP/IP

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:

I download and build openMPI 3.0.0 from source and then use it with TensorFlow 1.8 and Horovod to train a model over 4 machines in a local network (all CPUs, no GPU).

If I run the command with the newer syntax `mpirun mpirun --map-by ppr:pe` the job fails with the error:

```--------------------------------------------------------------------------
 
  HNP daemon   : [[55014,0],0] on node skl-node1
  Remote daemon: [[55014,0],1] on node skl-node2
 
--------------------------------------------------------------------------```

However, if I run the same thing with the older (soon to be deprecated) syntax it works:  

`mpirun --map-by socket -cpus-per-proc`

I've run this on several different local networks with different machines. If I run the newer syntax but stick to running on a single node (i.e. -H localhost), then that works too. 

Is there some additional flag that the older syntax automatically adds that I need for the newer syntax?

Thanks.
-Tony

rhc54:
this  fixes #5482

### DIFF
--- a/opal/mca/pmix/ext2x/ext2x_component.c
+++ b/opal/mca/pmix/ext2x/ext2x_component.c
@@ -118,7 +118,7 @@ static int external_open(void)
     OBJ_CONSTRUCT(&mca_pmix_ext2x_component.dmdx, opal_list_t);
 
     version = PMIx_Get_version();
-    if ('2' >= version[0]) {
+    if ('2' > version[0]) {
         opal_show_help("help-pmix-base.txt",
                        "incorrect-pmix", true, version, "v2.x");
         return OPAL_ERROR;
--- a/opal/mca/pmix/ext2x/ext2x_component.c
+++ b/opal/mca/pmix/ext2x/ext2x_component.c
@@ -118,12 +118,12 @@ static int external_open(void)
     OBJ_CONSTRUCT(&mca_pmix_ext2x_component.dmdx, opal_list_t);
 
     version = PMIx_Get_version();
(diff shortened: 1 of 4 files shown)


hppritcha:
this fixes https://github.com/open-mpi/ompi/issues/5482

hppritcha:
@ggouaillardet please check when you have chance

hppritcha:
@ggouaillardet ping

ggouaillardet:
I will be back on 8/16.

Please reassign this PR if urgent.

PR #5493, opened 2018-07-27
v4.0.x: apply romio314 patch to romio321

markalle:
When romio314 was first pulled in an extra patch was applied to it, see commit
92f6c7c1e210c559471a05aaac9b19e0bd3d71bb. Most of that patch is already present
in vanilla romio321, but the fix for MPIO_DATATYPE_ISCOMMITTED() isn't.

If that macro doesn't set err_ then some paths end up with a variable being used
uninitialized. In particular you can trace through romio321/romio/mpi-io/read.c
to see what happens with error_code. It's an uninitialized stack variable that goes
through three MPIO_CHECK_* macros none of which set it. The macros consistently set
error_code to a failure if they see something wrong, but they don't consistently
set it to success when things are fine.

And then in the last macro MPIO_CHECK_DATATYPE it tries to look at the value
of error_code that was never set.

fixes 5489

### DIFF
--- a/ompi/mca/io/romio321/romio/mpi-io/mpioimpl.h
+++ b/ompi/mca/io/romio321/romio/mpi-io/mpioimpl.h
@@ -36,7 +36,12 @@
 #define ROMIO_THREAD_CS_ENTER()
 #define ROMIO_THREAD_CS_EXIT()
 #define ROMIO_THREAD_CS_YIELD()
-#define MPIO_DATATYPE_ISCOMMITTED(dtype_, err_) do {} while (0)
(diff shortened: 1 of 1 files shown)


hppritcha:
probably @markalle @gpaulsen we need indication of the sha of the commit into master, best if its in the commit message.  you can pick this up automatically by using

```
git cherry-pick -x sha_of_commit_on_master
```

jsquyres:
No, this does not look like a cherry-pick from master -- it looks like #5494 is the PR to master.

Please merge #5494 first, and then cherry pick that to the release branch (so we can more easily track what has migrated from master to the release branch(es)).

PR #5495, opened 2018-07-28
PML/SPML/UCX: init global objects using C99 style - v4.0

hoopoepg:
- to avoid value mix used C99 style of object initializations

(cherry picked from commit 2806504290ff2fdd10f254f878e92cae6e90c854)


hppritcha:
@gpaulsen this is not really a bug fix but since it improves code robustness  I think we should take it despite the enhancement label.

PR #5496, opened 2018-07-30
opal/atomic.h: prevent preprocessor macro redefinition

jsquyres:
This is not a cherry-pick from master because it is an Open MPI
v2.1.x-specific issue.

The embedded PMIx (v1.2.4) and Open MPI have duplicate definitions of
the "MB" preprocessor macro.  This was fixed in PMIx v1.2.5, but it's
not worth updating the embedded PMIx here in Open MPI because the
v2.1.x series is reaching its end of life.

Instead, just surround the OPAL "MB" definition with an #ifdef and
call it "good enough".

### DIFF
--- a/opal/include/opal/sys/sync_builtin/atomic.h
+++ b/opal/include/opal/sys/sync_builtin/atomic.h
@@ -45,7 +45,16 @@ static inline void opal_atomic_wmb(void)
     __sync_synchronize();
 }
 
+// This avoids a duplicate definition in PMIx <v1.2.5 (Open MPI v2.1.x
+// has PMIx 1.2.4).  This was fixed in PMIx v1.2.5, but it is not
+// worth updating the embedded PMIx in Open MPI v2.1.x because a) the
+// OMPI v2.1.x series is nearing its end of life, and b) it's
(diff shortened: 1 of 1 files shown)


opal/include/opal/sys/sync_builtin/atomic.h
```diff
@@ -45,7 +45,16 @@ static inline void opal_atomic_wmb(void)
     __sync_synchronize();
 }
 
+// This avoids a duplicate definition in PMIx <v1.2.5 (Open MPI v2.1.x
+// has PMIx 1.2.4).  This was fixed in PMIx v2.1.5, but it is not
```
rhc54: it was fixed in PMIx v1.2.5  <-- you typo'd the version

hppritcha:
@jsquyres please fix typo Ralph found then we'll merge

opal/include/opal/sys/sync_builtin/atomic.h
jsquyres: Thx -- fixed.

jsquyres:
@hppritcha approved above, so I'll merge.

PR #5498, opened 2018-07-31
pmix: added check for pmix fence status

### LINKED ISSUE #5086: Check PMIx Fence return status in MPI_Init
One of the customers recently encountered the problem in the Slurm/PMIx environment.
For some unrelated reasons PMIx_Fence was cancelled by timeout by RM. However because OMPI doesn't checks the return status of the operation all processes proceeded further:
* modex: https://github.com/open-mpi/ompi/blob/v3.0.x/ompi/runtime/ompi_mpi_init.c#L663
* barrier: https://github.com/open-mpi/ompi/blob/v3.0.x/ompi/runtime/ompi_mpi_init.c#L840
I think this is not a correct behavior and processes should take an error code path in this case.

karasevb:
Closes #5086

### DIFF
--- a/ompi/runtime/ompi_mpi_finalize.c
+++ b/ompi/runtime/ompi_mpi_finalize.c
@@ -257,7 +257,13 @@ int ompi_mpi_finalize(void)
              * communications/actions to complete.  See
              * https://github.com/open-mpi/ompi/issues/1576 for the
              * original bug report. */
-            opal_pmix.fence_nb(NULL, 0, fence_cbfunc, (void*)&active);
+            if (OMPI_SUCCESS != (ret = opal_pmix.fence_nb(NULL, 0, fence_cbfunc,
+                                                          (void*)&active))) {
+                OMPI_ERROR_LOG(ret);
+                /* Reset the active flag to false, to avoid waiting for
+                 * completion when the fence was failed. */
+                active = false;
+            }
             OMPI_LAZY_WAIT_FOR_COMPLETION(active);
         } else {
             /* However, we cannot guarantee that the provided PMIx has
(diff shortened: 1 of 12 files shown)


ibm-ompi:

 Gist: https://gist.github.com/c7391e8f58032139aeb5325d6e23f29c

ibm-ompi:

 Gist: https://gist.github.com/0fe463389af62da504c537cf69f68475

ibm-ompi:

 Gist: https://gist.github.com/1d1f14f27d9637dd6c832ac665a10a75

ompi/runtime/ompi_mpi_finalize.c
```diff
@@ -257,7 +257,10 @@ int ompi_mpi_finalize(void)
              * communications/actions to complete.  See
              * https://github.com/open-mpi/ompi/issues/1576 for the
              * original bug report. */
-            opal_pmix.fence_nb(NULL, 0, fence_cbfunc, (void*)&active);
+            if (OMPI_SUCCESS != (ret = opal_pmix.fence_nb(NULL, 0, fence_cbfunc,
+                                                          (void*)&active))) {
+                OMPI_ERROR_LOG(ret);
+            }
```
rhc54: If the fence operation returns an error, then the ```wait_for_completion``` is never going to happen. You need to skip that step.

hppritcha:
@karasevb we use the release stream tags on the underlying issue and for a PR into a non-master branch.

ompi/runtime/ompi_mpi_finalize.c
karasevb: Thanks. Fixed.

PR #5499, opened 2018-07-31
MTL OFI: Fix Deadlock in fi_cancel given completion during cancel

nrspruit:
- If a message for a recv that is being cancelled gets completed after
the call to fi_cancel, then the OFI mtl will enter a deadlock state
waiting for ofi_req->super.ompi_req->req_status._cancelled which will
never happen since the recv was successfully finished.

- To resolve this issue, the OFI mtl now checks ofi_req->req_started
to see if the request has been started within the loop waiting for the
event to be cancelled. If the request is being completed, then the loop
is broken and fi_cancel exits setting
ofi_req->super.ompi_req->req_status._cancelled = false;

(cherry picked from commit 767135c580f75d3dde9cb9c88601dd18afda949a)


rhc54:
ok to test

aravindksg:
@hppritcha : Could you please take a look. This is a cherry-pick of PR #5476

PR #5503, opened 2018-07-31
MTL OFI: Fix race condition due to global progress entries array

aravindksg:
Since progress entries array is globally allocated, it is susceptible
to race conditions when using multi-threaded applications. Allocating it
on the stack resolves any potential races as it is thread local by default.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -59,6 +59,7 @@ ompi_mtl_ofi_progress(void)
     int count = 0, i, events_read;
     struct fi_cq_err_entry error = { 0 };
     ompi_mtl_ofi_request_t *ofi_req = NULL;
+    struct fi_cq_tagged_entry wc[ompi_mtl_ofi.ofi_progress_event_count];
 
     /**
      * Read the work completions from the CQ.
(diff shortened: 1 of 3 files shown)


aravindksg:
Adding @jsquyres @hppritcha @bwbarrett @rhc54 for reviews.

rhc54:
Afraid I don't know enough to comment on this area.

jsquyres:
Me neither.

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -59,23 +59,25 @@ ompi_mtl_ofi_progress(void)
     int count = 0, i, events_read;
     struct fi_cq_err_entry error = { 0 };
     ompi_mtl_ofi_request_t *ofi_req = NULL;
+    struct fi_cq_tagged_entry wc[ompi_mtl_ofi.ofi_progress_event_count];
+
+    memset(&wc, 0, sizeof(struct fi_cq_tagged_entry) * ompi_mtl_ofi.ofi_progress_event_count);
```
hppritcha: isn't this kind of an expensive operation to be doing in a progress function?

aravindksg: Yeah, this affects latency particularly just after the inject_threshold when we start doing completions. But it's needed for THREAD_MULTIPLE functionality. The "memset()" is strictly speaking not required by OFI, so I think we can safely remove this (I tested after removing the memset and this seems to work fine). Thoughts?

aravindksg: Done. FWIW, I see ~5-6% improvement in latency for message sizes just after inject threshold after removing the memset.

aravindksg:
@hppritcha Just pinging back on this PR -  If no other concerns, please approve and merge?

aravindksg:
ping @hppritcha

hppritcha:
this looks good now.

PR #5505, opened 2018-08-01
orte/runtime: always set opal_using_threads for orted/mpirun

hjelmn:
Both orted and mpirun use threads to speed up local process spawing.
In order to avoid data corruption when calling the opal_output
interface we need to ensure that opal_using_threads is set to true.


hjelmn:
@rhc54 We will have to apply this to v4.0.x as well. I will check if it is needed for v3.0.x and v3.1.x.

PR #5506, opened 2018-08-01
v2.x: errhandler_predefined: remove erroneous assert()

jsquyres:
This assert() was added in 13ddcbfbaa2 (which was a back-ported cherry
pick from master: 35438ae9b521), but it is actually exactly the
opposite assert that should be there.  The assert() is actually
superflouous, because the "if" statement resulting in invoking this
function is exactly what the assert() should be.  Hence, the simplest
solution here is just to remove that assert().

This assert() is not present on master: it's only on the v2.x branch.
Hence, this is not a cherry-pick from master -- it's a direct PR to
the v2.x branch.

This issue found in Cisco MTT runs -- see https://mtt.open-mpi.org/index.php?do_redir=2657

### DIFF
--- a/ompi/errhandler/errhandler_predefined.c
+++ b/ompi/errhandler/errhandler_predefined.c
@@ -190,10 +190,6 @@ static void backend_fatal_aggregate(char *type,
     const char* usable_prefix = unknown_prefix;
     const char* usable_err_msg = unknown_error;
 
-    int32_t state = ompi_mpi_state;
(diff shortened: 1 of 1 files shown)


jsquyres:
@hppritcha As cited in the description, this is *not* a cherry pick because the issue is specific to the v2.x branch (i.e., the issue was introduced when master commit 35438ae was backported / cherry-picked to v2.x and committed in 13ddcbf).

hppritcha:
@jsquyres  please merge after CI

PR #5507, opened 2018-08-01
Adding -mca comm_method to print table of communication methods

markalle:
This is closely related to Platform-MPI's old -prot feature.

The long-format of the tables it prints could look like this:
```
>   Host 0 [myhost001] ranks 0 - 1
>   Host 1 [myhost002] ranks 2 - 3
>   Host 2 [myhost003] ranks 4
>   Host 3 [myhost004] ranks 5
>   Host 4 [myhost005] ranks 6
>   Host 5 [myhost006] ranks 7
>   Host 6 [myhost007] ranks 8
>   Host 7 [myhost008] ranks 9
>   Host 8 [myhost009] ranks 10
>
>    host | 0    1    2    3    4    5    6    7    8
>   ======|==============================================
>       0 : sm   tcp  tcp  tcp  tcp  tcp  tcp  tcp  tcp
>       1 : tcp  sm   tcp  tcp  tcp  tcp  tcp  tcp  tcp
>       2 : tcp  tcp  self tcp  tcp  tcp  tcp  tcp  tcp
>       3 : tcp  tcp  tcp  self tcp  tcp  tcp  tcp  tcp
>       4 : tcp  tcp  tcp  tcp  self tcp  tcp  tcp  tcp
>       5 : tcp  tcp  tcp  tcp  tcp  self tcp  tcp  tcp
>       6 : tcp  tcp  tcp  tcp  tcp  tcp  self tcp  tcp
>       7 : tcp  tcp  tcp  tcp  tcp  tcp  tcp  self tcp
>       8 : tcp  tcp  tcp  tcp  tcp  tcp  tcp  tcp  self
>
>   Connection summary:
>     on-host:  all connections are sm or self
>     off-host: all connections are tcp
```

In this example hosts 0 and 1 had multiple ranks so "sm" was more
meaningful than "self" to identify how the ranks on the host are
talking to each other. While host 2..8 were one rank per host so
"self" was more meaningful as their btl.

Above a certain number of hosts (12 by default) the above table gets too big
so we shrink to a more abbreviated looking table that has the same data:
```
>    host | 0 1 2 3 4       8
>   ======|====================
>       0 : A C C C C C C C C
>       1 : C A C C C C C C C
>       2 : C C B C C C C C C
>       3 : C C C B C C C C C
>       4 : C C C C B C C C C
>       5 : C C C C C B C C C
>       6 : C C C C C C B C C
>       7 : C C C C C C C B C
>       8 : C C C C C C C C B
>   key: A == sm
>   key: B == self
>   key: C == tcp
```
Then above 36 hosts we stop printing the 2d table entirely and just print the
summary:
```
>   Connection summary:
>     on-host:  all connections are sm or self
>     off-host: all connections are tcp
```
The options to control it are
```
    -mca comm_method 1   :   print the above table at the end of MPI_Init
    -mca comm_method 2   :   print the above table at the beginning of MPI_Finalize
```
The most important difference between these two is that when printing the table
during MPI_Init(), we send extra messages to make sure all hosts are connected to
each other. So the table ends up working against the idea of on-demand connections
(although it's only forcing the n^2 connections in the number of hosts, not the
total ranks).  If printing at MPI_Finalize() we don't create any connections that
aren't already connected, so the table is more likely to have "n/a" entries if
some hosts never connected to each other.

The other tunable is a simple environment variable MPI_COMM_METHOD_MAX that
defaults to 12 that controls at what host-count the unabbreviated / abbreviated
2d tables get printed:
```
    1 - n      : full size 2d table
    n+1 - 3n   : shortened 2d table
    3n+1 - inf : summary, no 2d table
```

The source of the information used in the table is the .mca_component_name

In the case of BTLs, the module always had a .btl_component linking back to the
component. This adds a similar field for .pml_component and .mtl_component to
those modules. Note, when setting the .pml_component field I noticed nobody was
setting .pml_flags, so I added a 0 for setting that field as well.

So with the new field linking back to the component, we can then access
the component name with code like this
```
    mca_pml.pml_component->pmlm_version.mca_component_name
```
See the three lookup_{pml,mtl,btl}_name() functions in hook_comm_method_fns.c,
and their use in comm_method() to parse the strings and produce an integer
to represent the connection type being used.

I think the weakest part is the strings_map[] list of recognized names in hook_comm_method_fns.c that's needed in order to have a decent mapping between the strings and integers. If someone adds a new btl/mtl/pml with a new name that's not in this list, it will map to 0 for unknown and print as "n/a".

### DIFF
--- a/ompi/mca/hook/comm_method/hook_comm_method_fns.c
+++ b/ompi/mca/hook/comm_method/hook_comm_method_fns.c
@@ -1,0 +1,882 @@
+/*
+ * Copyright (c) 2016-2018 IBM Corporation. All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+
+#include "hook_comm_method.h"
+
+#ifdef HAVE_DLFCN_H
+#include <dlfcn.h>
+#endif
+
+#include "ompi/communicator/communicator.h"
+#include "ompi/mca/pml/pml.h"
+#include "opal/mca/btl/btl.h"
+#include "ompi/mca/pml/base/base.h"
+#include "ompi/mca/bml/base/base.h"
+#include "ompi/mca/mtl/base/base.h"
+
+// In regular strncpy up to n bytes are copied, so if the 'to' buffer
+// was char string[16] and you called strncpy(string, , 16) you could
+// get 16 bytes of chars without a null.  My preferred API is to let
+// n be the size of the buffer, and to let n-1 chars be copied, and
+// to guarantee null termination.
+static void
+mystrncpy(char *to, const char *from, int n) {
+    strncpy(to, from, n-1);
+    to[n-1] = 0;
+}
+
+// For converting comm_method strings to comm_method id# and back.
+// This starts as our local set of strings, but gets Allreduced into
+// a global mapping so all the strings at all the ranks are represented.
+// If an MCA's name is more than 15 chars it gets truncated.
+#define COMM_METHOD_STRING_SIZE 16
+#define MAX_COMM_METHODS 50
+typedef struct {
+    int n;
+    char str[MAX_COMM_METHODS][COMM_METHOD_STRING_SIZE];
+} comm_method_string_conversion_t;
+
+static comm_method_string_conversion_t comm_method_string_conversion;
+
+#define MODE_IS_PML 1
+#define MODE_IS_MTL 2
+#define MODE_IS_BTL 3
+
+// ----------------------------------------------------------------------------
+
+// return the pml's module:component:name function pointer in fp
+static char*
+lookup_pml_name(void)
+{
+    return (char*) mca_pml_base_selected_component.pmlm_version.mca_component_name;
+}
+// return the mtl's module:component:name function pointer in fp
+static char*
+lookup_mtl_name(void)
+{
+    if (!ompi_mtl_base_selected_component) { return NULL; }
+    return (char*) ompi_mtl_base_selected_component->mtl_version.mca_component_name;
+}
+// Find the send btl's module:component:name for the incoming comm,rank
+static char*
+lookup_btl_name_for_send(ompi_communicator_t* comm, int rank) {
+    ompi_proc_t *dst_proc = ompi_group_peer_lookup_existing(comm->c_remote_group, rank);
+
+    mca_bml_base_endpoint_t* endpoint = mca_bml_base_get_endpoint(dst_proc);
+    if (endpoint &&
+        endpoint->btl_send.bml_btls &&
+        endpoint->btl_send.bml_btls[0].btl)
+    {
+        return (char*)
+            endpoint->btl_send.bml_btls[0].btl->btl_component->btl_version.mca_component_name;
+    }
+    return NULL;
+}
+
+// Use the above to lookup the mca_component_name for the rank's
+// sending BTL/MTL/PML.  The extra 3rd argument returns whether
+// MODE_IS_BTL / MTL / PML (can pass NULL if you don't want that info).
+//
+// And this one puts the result into malloced mem of size
+//     COMM_METHOD_STRING_SIZE
+// that the caller has to free.
+static char *
+comm_method_string(MPI_Comm comm, int rank, int *comm_mode) {
+    char *p;
+    char *string = malloc(COMM_METHOD_STRING_SIZE);
+
+    if (!string) { return NULL; }
+
+    p = lookup_pml_name();
+    if (p && 0==strncmp("ob1", p, 4)) {      // BTL
+        if (comm_mode) { *comm_mode = MODE_IS_BTL; }
+        mystrncpy(string, lookup_btl_name_for_send(comm, rank), COMM_METHOD_STRING_SIZE);
+    }
+    else if (p && 0==strncmp("cm", p, 3)) {  // MTL
+        if (comm_mode) { *comm_mode = MODE_IS_MTL; }
+        mystrncpy(string, lookup_mtl_name(), COMM_METHOD_STRING_SIZE);
+    } else {                        // PML
+        if (comm_mode) { *comm_mode = MODE_IS_PML; }
+        mystrncpy(string, p, COMM_METHOD_STRING_SIZE);
+    }
+    return string;
+}
+
+// ----------------------------------------------------------------------------
+
+// Managing the comm_method_string_conversion structure
+// and using it to convert strings to/from id numbers:
+
+// The data should be
+//     string 0 ==       "n/a" for unconnected / unknown
+//     string 1,2,... == "tcp" "self", etc, sorted
+// self is important enough we want to make sure it's always in the list
+static void
+init_string_to_conversion_struct(comm_method_string_conversion_t *data)
+{
+    data->n = 0;
+    strcpy(data->str[data->n], "n/a");
+    ++(data->n);
+    strcpy(data->str[data->n], "self");
+    ++(data->n);
+}
+
+static int
+lookup_string_in_conversion_struct(comm_method_string_conversion_t *data, char *string)
+{
+    int i;
+    for (i=0; i<data->n; ++i) {
+        if (0==strncmp(data->str[i], string, COMM_METHOD_STRING_SIZE)) {
+            return i;
+        }
+    }
+    return 0;
+}
+
+// For qsort of the str[] array in a comm_method_string_conversion_t
+static int mycompar(const void *a, const void *b) {
+    return strcmp(a, b);
+}
+
+static void
+add_string_to_conversion_struct(comm_method_string_conversion_t *data, char *string)
+{
+    int i;
+    if (0 == strcmp(string, "n/a")) { return; }
+
+    i = lookup_string_in_conversion_struct(data, string);
+    if (i == 0) { // didn't find string in list, so add it
+        if (data->n < MAX_COMM_METHODS) {
+            mystrncpy(data->str[data->n], string, COMM_METHOD_STRING_SIZE);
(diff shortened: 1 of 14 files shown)


ompi/mca/pml/pml.h
```diff
@@ -523,6 +524,8 @@ struct mca_pml_base_module_1_0_1_t {
     uint32_t                              pml_max_contextid;
     int                                   pml_max_tag;
     int                                   pml_flags;
+
+    mca_pml_base_component_t             *pml_component;
```
bosilca: Isn't this equivalent to using mca_pml_base_selected_component or ompi_mtl_base_selected_component (which are supposed to either be NULL if the corresponding type was not selected or contain the component selected)?

ompi/mca/hook/comm_method/hook_comm_method_fns.c
```diff
@@ -0,0 +1,800 @@
...
+            0, leader_comm, leader_comm->c_coll->coll_gather_module);
+    }
+    ompi_comm_free(&local_comm);
+    ompi_comm_free(&leader_comm);
+
+// Interception for testing purposes. Let rank-0 meddle with all its method[]
+// settings, this is only for testing, eg to make sure the printing comes out
+// right.
+    if (myleaderrank == 0) {
+        p = getenv("MPI_COMM_METHOD_FAKEFILE");
```
bosilca: undocumented.

ompi/mca/hook/comm_method/hook_comm_method_fns.c
```diff
@@ -0,0 +1,800 @@
...
+    // hpmp_nprocs = ompi_comm_size(MPI_COMM_WORLD);
+
+    if (hpmp_myrank == 0) {
+        max2Dprotptr = getenv("MPI_COMM_METHOD_MAX");
+        if (max2Dprotptr) {
+            max2Dprottable = atoi(max2Dprotptr);
+            max2D1Cprottable = 3 * max2Dprottable;
+        }
+
+        hostidprotptr = getenv("MPI_COMM_METHOD_BRIEF");
```
bosilca: undocumented.

ompi/mca/hook/comm_method/hook_comm_method_fns.c
jsquyres: Not only is this undocumented -- why wouldn't you use an MCA parameter for this?  (and the other place(s) where you use `getenv()`)

markalle:
I updated a few things, switched to -mca for the settings, even the "fakefile" one that's really just for internal debugging.

Thanks a bunch for the mca_pml_base_selected_component and ompi_mtl_base_selected_component variables, that helped simplify this checkin a lot.

So what about the other main comment about the highest priority item from the send list of BTLs. Is there anything else that would be more meaningful there?

I could probably redo it to understand the possibility of multiple BTLs between two hosts. I'd switch the internals to a mask, and for the printout use the second format for the table, using A,B,C,D etc and a key to identify things. I don't think the current code is very close to letting that work but it's not impossible.

I still lean toward thinking it's good enough without supporting printout of multiple BTLs between two hosts though. It still has a decent utility to let a person make sure they're really running what they think they're running

I'd still like to have a 1d table at some point with more per-host info like IP address, and that 1d table is where I'd be more prone to list multiple items for a host, eg list multiple IP addresses for a host or list it as having both an ibverbs BTL and a tcp BTL active there

ompi/mca/hook/comm_method/hook_comm_method_fns.c
```diff
@@ -0,0 +1,788 @@
...
+static char*
+lookup_mtl_name(void)
+{
+    if (!ompi_mtl_base_selected_component) { return NULL; }
+    return (char*) ompi_mtl_base_selected_component->mtl_version.mca_component_name;
+}
+// Find the send btl's module:component:name for the incoming comm,rank
+static char*
+lookup_btl_name_for_send(ompi_communicator_t* comm, int rank) {
+    ompi_proc_t *dst_proc = ompi_comm_peer_lookup (comm, rank);
```
bosilca: Using this function forces the allocation of the underlying proc structure, but not it's initialization (so the endpoint will be NULL). I suggest using `ompi_group_peer_lookup_existing(comm->c_remote_group, rank)` instead, which will prevent the allocation of the proc if the proc is still the sentinel.

ompi/mca/hook/comm_method/hook_comm_method_fns.c
```diff
@@ -0,0 +1,788 @@
...
+        endpoint->btl_send.bml_btls &&
+        endpoint->btl_send.bml_btls[0].btl)
+    {
+        return (char*)
+            endpoint->btl_send.bml_btls[0].btl->btl_component->btl_version.mca_component_name;
+    }
+    return NULL;
+}
+
+char *strings_map[] = {
```
bosilca: This approach is extremely static. Why not using a hash table for the names ?

markalle: I guess it could be a hash, but I'd still need a way to convert the hashes back to strings. So I'd either still have a static list of strings to walk or else I'd have all the ranks including their list of strings in the messages they send to rank 0 so it could have a complete list of strings.

The other part is right now the code can distinguish between "ofi" as an MTL and "ofi" as a BTL. If it was just a general string hash I don't think it would know that both existed. Maybe that's a minor enough feature I shouldn't care.

I'd hate for the table that currently says "tcp" to get unnecessarily bloated to "tcp(btl)" etc.

If we throw out the first long-format table
```
>    host | 0    1    2    3    4    5    6    7    8
>   ======|==============================================
>       0 : sm   tcp  tcp  tcp  tcp  tcp  tcp  tcp  tcp
>       1 : tcp  sm   tcp  tcp  tcp  tcp  tcp  tcp  tcp
>       2 : tcp  tcp  self tcp  tcp  tcp  tcp  tcp  tcp
>       3 : tcp  tcp  tcp  self tcp  tcp  tcp  tcp  tcp
>       4 : tcp  tcp  tcp  tcp  self tcp  tcp  tcp  tcp
>       5 : tcp  tcp  tcp  tcp  tcp  self tcp  tcp  tcp
>       6 : tcp  tcp  tcp  tcp  tcp  tcp  self tcp  tcp
>       7 : tcp  tcp  tcp  tcp  tcp  tcp  tcp  self tcp
>       8 : tcp  tcp  tcp  tcp  tcp  tcp  tcp  tcp  self
```
and only used the more abbreviated form
```
>    host | 0 1 2 3 4       8
>   ======|====================
>       0 : A C C C C C C C C
>       1 : C A C C C C C C C
>       2 : C C B C C C C C C
>       3 : C C C B C C C C C
>       4 : C C C C B C C C C
>       5 : C C C C C B C C C
>       6 : C C C C C C B C C
>       7 : C C C C C C C B C
>       8 : C C C C C C C C B
>   key: A == sm
>   key: B == self
>   key: C == tcp
```
then the extra characters wouldn't give me any heartburn. But I still find the long form more readable than the above so prefer it on the smaller runs

markalle:
Repushed just with the change to use ompi_group_peer_lookup_existing() in the BTL lookup, still using the static list in strings_map[] as the mapping that turns the known interconnect strings into numbers and back.

ibm-ompi:

 Gist: https://gist.github.com/2daa7e94eb91dd18632a74c859f0916b

jjhursey:
The IBM PGI failure seems unrelated. So can be disregarded. We updated the compiler from 17.10 to 18.10 and there must be a regression.

markalle:
retest
bot:retest

gpaulsen:
@markalle What's the fate of this PR?  Can you please rebase it, and try again for late April?

markalle:
Okay, I fixed the other big sticking point nobody liked: the static list of strings for the recognized interconnects.  Now they Allreduce their individual lists into a uniform list so the ranks can map strings to id#'s and all have the same result.

Related to this I decided that my original objection (that the "portals4" name for example exists both as a btl and an mtl, and is thus ambiguous) isn't a big deal.  The pml/mtl tables are always pretty boring with just a bunch of uniform entries anyway.  So I let the strings be used as is, and I added in the "Connection Summary" section a little note saying whether the thing being shown in the table was a btl/mtl/pml.

ompi/mca/hook/comm_method/hook_comm_method_fns.c
gpaulsen: Can you comment why some hosts talk to the same host over sm, and other hosts use self?
Can we infer that the hosts that talk to themselves over self only have 1 rank on them?

gpaulsen:
I added the Label:State Awaiting user information, as It'd be nice to have community buy-in on this approach before merging to master.

ompi/mca/hook/comm_method/hook_comm_method_fns.c
markalle: Yeah, it's identifying the local-host traffic.  If there's one rank on the host then it's just 'self'. If there are multiple ranks then both 'self' and 'vader' would be active between ranks on that host, so it shows 'vader' as that's the more useful piece of info

bosilca: I have to say the reading of such table is not trivial, as one might expect to read connectivity between MPI ranks when in fact this table shows connectivity (one of the possible BTL) between nodes (all ranks accumulated together). Moreover, it works best for regular and symmetric cases on homogeneous environments, the exact places where the best answer would have been "tcp across nodes, vader between processes on the same node, and self for communication in the same process".

AboorvaDevarajan:

gpaulsen:
@markalle Can you please rebase this, and we can try to get this into master?

markalle:
I think I've addressed all the concerns above, with maybe an exception being the fact that if multiple BTLs are used this only reports on btl[0].  I still think it's producing meaningful information though.

Issue #5514, opened 2018-08-03
Treematch warnings


rhc54:
In master using gcc-7.3:
```shell
topo_treematch_dist_graph_create.c: In function 'mca_topo_treematch_dist_graph_create':
topo_treematch_dist_graph_create.c:288:20: warning: 'depth' may be used uninitialized in this function [-Wmaybe-uninitialized]
             object = hwloc_get_obj_by_depth(opal_hwloc_topology, effective_depth, obj_rank);
             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
treematch/tm_tree.c: In function 'group_nodes':
treematch/tm_tree.c:1753:21: warning: variable 'list_size' set but not used [-Wunused-but-set-variable]
   unsigned long int list_size;
                     ^~~~~~~~~
treematch/tm_topology.c: In function 'topology_to_arch':
treematch/tm_topology.c:144:8: warning: argument 1 value '18446744073709551608' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
   arch = (double**)MALLOC(sizeof(double*)*nb_proc);
   ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /Users/rhc/openmpi/foobar/opal/mca/hwloc/hwloc201/hwloc/include/hwloc/helper.h:20:0,
                 from /Users/rhc/openmpi/foobar/opal/mca/hwloc/hwloc201/hwloc/include/hwloc.h:2242,
                 from treematch/tm_topology.c:1:
/usr/include/stdlib.h:160:7: note: in a call to allocation function 'malloc' declared here
 void *malloc(size_t __size) __result_use_check __alloc_size(1);
       ^~~~~~
treematch/tm_tree.c: In function 'partial_exhaustive_search':
treematch/tm_tree.c:979:7: warning: 'id' may be used uninitialized in this function [-Wmaybe-uninitialized]
       thread_derecurs_exhaustive_search(tab_group, id+1, n, arity, work->nb_groups, solution_size, val, best_val, selection, best_selection, indep_mat, lock, thread_id, tab_i, work->nb_groups);
[...]

PR #5517, opened 2018-08-03
Remove few warnings identified by @rhc in #5514.

bosilca:

Refs #5514 #5512

### DIFF
--- a/ompi/mca/topo/treematch/treematch/tm_topology.c
+++ b/ompi/mca/topo/treematch/treematch/tm_topology.c
@@ -141,7 +141,14 @@ double ** topology_to_arch(hwloc_topology_t topology)
   double **arch = NULL;
 
   nb_proc = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_PU);
+  if( nb_proc <= 0 ) {  /* if multiple levels with PUs */
+      return NULL;
+  }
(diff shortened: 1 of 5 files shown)


ompi/mca/topo/treematch/treematch/tm_topology.c
```diff
@@ -141,7 +141,12 @@ double ** topology_to_arch(hwloc_topology_t topology)
   double **arch = NULL;
 
   nb_proc = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_PU);
+  if( nb_proc <= 0 )  /* if multiple levels with PUs */
+      return NULL;
   arch = (double**)MALLOC(sizeof(double*)*nb_proc);
+  if( NULL == arch )
```
jsquyres: Minor: both of these blocks don't obey our code standards of always including blocks -- even 1-line blocks -- in `{}`.  😜

bosilca: Based on the code accepted lately I was under the impression that we are not following the code standard anymore...

jsquyres: That and 4-space tabs are about the only thing we ever enforced from the code standards (minimal that they were)...  Maybe putting `{` in the same line as `if`/`while`/etc., too.

PR #5525, opened 2018-08-07
opal/progress: protect against multiple threads in event base

hjelmn:
libevent does not support multiple threads calling the event loop on
the same event base. This causes external libevent's to print out
re-entrant warning messages. This commit fixes the issue by protecting
the call to the event loop with an atomic swap check.

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -10,7 +10,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2006-2016 Los Alamos National Security, LLC.  All rights
+ * Copyright (c) 2006-2018 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


hjelmn:
Example of the bug being triggered:

```
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
[warn] event_base_loop: reentrant invocation.  Only one event_base_loop can run on each event_base at once.
```

hjelmn:
Probably should be PR'd to v4.0.x, v3.0.x, and v3.1.x.

thananon:
Wait, so does this mean opal_progress can do 1 event base at a time? Is it still possible for multiple thread to go through btl_progress at the same time?

rhc54:
Not exactly. The problem is that libevent can only have one thread calling event_loop on a given event base at a time. Multiple threads can call event_loop so long as each is using its own event base. This is why we created the opal_progress_threads code in opal/runtime.

In the case of the opal_progress function itself, there is only one event_base involved - the opal global one. Since only one thread can call event_loop, we can only have one thread at a time traverse that particular section of the code.

Note that the "one thread" limit is _only_ imposed when the event library is to be advanced - it is _not_ imposed every time we call opal_progress. @hjelmn did a quick check and found no impact on performance, but it would probably be good to have others verify.

hppritcha:
Hmmmm.....should this be cherry picked into v4.0.x?

PR #5536, opened 2018-08-13
btl/vader: move memory barrier to where it belongs

hjelmn:
The write memory barrier was intended to precede setting a fast-box
header but instead follows it. This commit moves the memory barrier to
the intended location.

### DIFF
--- a/opal/mca/btl/vader/btl_vader_fbox.h
+++ b/opal/mca/btl/vader/btl_vader_fbox.h
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2011-2015 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2011-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


hjelmn:
@shamisp Here is the PR.

shamisp:
👍

jsquyres:
@hjelmn How bad is this bug?  Do we need to back port it to other branches / do immediate releases?

shamisp:
@jsquyres it is pretty bad. 
it causes random data corruption on arm and power. We actually observed data corruption with real application. 

The issue was introduced by this commit:
https://github.com/open-mpi/ompi/commit/a82f761a4a6670ff83d3c47601eeb26120d83be5#diff-821da35c17ae856f23e11bbedbc14976

hjelmn:
Really surprised this wasn't noticed earlier or during testing. @shamisp Can we get a reproducer we can add to MTT?

shamisp:
It shows up with Graph500. @nSircombe may have more details.

jsquyres:
Per 2018-08-13 discussion:
* Appears to be ARM/POWER specific
* Is in v2.1.x, v3.0.x, and v3.1.x (and recently-branched-but-not-yet-released v4.0.x)

nSircombe:
I don't have anything that reproduced the issues 100% of the time, this bug has manifested as intermittent (~1/3 of runs) failures (usually segfaults, occasionally harder to spot issues (failure of core solver to converge)) on the reference version of Graph500 and Tealeaf. So far I have focussed on mini-apps and benchmarks, although I have also experienced erratic behaviour with full-scale production applications, notable OpenFOAM.

However, 2.x and 3.0.x have not displayed the same issues,

jsquyres:
@hjelmn is out of range today, so I'm merging for him.

nSircombe:
Just to confirm, this patch has fixed the issues I'd seen with Graph500 and Tealeaf - 10s of test runs and all passed, no failures.

markalle:
Btw, did I hear this was affecting x86? If so did it appear with some new compiler version that was doing more compiler-level reordering? My understanding of the x86 CPU-level guarantees/non-guarantees is the only reordering it can do is "Loads may be reordered with older stores to different locations" which doesn't come up much.

So I wouldn't have expected x86 to be able to hit this unless the reordering happened at the compiler level.

jsquyres:
@markalle See the MTT link, above.

PR #5539, opened 2018-08-14
ompio: support for DDN's Infinite Memory Engine

gbossu:
This PR aims at bringing support for DDN’s Infinite Memory Engine (IME) in OMPIO.

To that end, both an fbtl and an fs component had to be created. These were named “ime”. Also, the close function of the ompio component was slightly modified, to allow IME to finalize at the same time OMPIO does. The patch was created using the master branch, but it can be adapted for 3.1.x with very little work.

We were also wondering: file locks are used in the POSIX fbtl component, for which reason are they actually necessary ? Shouldn't it be the developer’s responsibility to ensure i/o operations don’t overlap ? We kept the same logic as POSIX for file locks, but it would make the IME patch a lot cleaner if we could get rid of them.

Can we also know what is your testing process for OMPIO ? We couldn’t find the tests you use for MTT. If you want to compile and run some tests, we can provide a wrapper library to mock IME function calls using POSIX calls.

Signed-off-by: Gaëtan Bossu <gbossu@ddn.com>
Signed-off-by: Sylvain Didelot <sdidelot@ddn.com>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -55,6 +55,7 @@ Copyright (c) 2013-2017 Research Organization for Information Science (RIST).
                         All rights reserved.
 Copyright (c) 2017      Amazon.com, Inc. or its affiliates.  All Rights
                         reserved.
+Copyright (c) 2018      DataDirect Networks. All rights reserved.
 
 $COPYRIGHT$
 
--- a/config/ompi_check_ime.m4
+++ b/config/ompi_check_ime.m4
@@ -1,0 +1,62 @@
+dnl -*- shell-script -*-
+dnl
+dnl Copyright (c) 2018      DataDirect Networks. All rights reserved.
+dnl $COPYRIGHT$
+dnl
+dnl Additional copyrights may follow
+dnl
+dnl $HEADER$
+dnl
+
+# OMPI_CHECK_IME(prefix, [action-if-found], [action-if-not-found])
+# --------------------------------------------------------
+# check if IME support can be found.  sets prefix_{CPPFLAGS,
+# LDFLAGS, LIBS} as needed and runs action-if-found if there is
+# support, otherwise executes action-if-not-found
+AC_DEFUN([OMPI_CHECK_IME],[
+
+    check_ime_CPPFLAGS=
+    check_ime_LDFLAGS=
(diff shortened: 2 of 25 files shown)


hppritcha:
okay to test

edgargabriel:
Let me answer your questions. 
Regarding locking: explicit lock/unlock operations were introduced mostly for NFS file systems. Basically, the scenario that caused trouble were processes running on different nodes writing (small) data that ended up in the same block. Depending on the NFS mount options, the client side data caching could lead to the data of one of the nodes being overwritten by the other, despite of the fact that each process wrote to distinct locations that ended up however in the same fs block. I was not able to reproduce the problem on other file systems such as Lustre, but on NFS it was fairly common.

For your fbtl component, if you are sure that this scenario can not occur, feel free to remove the locking/unlocking, this is one of the benefits of having separate fbtl components for the various file systems.

Regarding testing: the nightly MTT tests do include some MPI I/O tests. However, for code development and verification, we have developed an in-house testsuite that I can give you access to if you want. In addition, we run regularly the parallel I/O tests from the hdf5 testsuite (and less regularly some other testsuites). 

If possible, it would be great to have access to the wrapper library that you mentioned just to be able to run some tests in case of a more generic change in ompio.

edgargabriel:
Let me know if you want to make any changes regarding the locking, otherwise the code looks good to me, I could merge it in.

gbossu:
Hello, thanks for the swift reply. I guess we will get rid of the file locks then, because it doesn't appear to be necessary in our case. Thank you for the explanation!

I will try to provide you the wrapper library tomorrow, because today is a holiday in France. And yes, it would be nice if we can access your in-house testsuite, this way we can ensure IME works as expected with OMPIO.

Thank you for the review. I think I will cone back to you again after we have made the necessary modifications regarding file locks.

gbossu:
So, I removed the file locks and updated the patch, as we judged the locks weren't necessary in our case.
However, I'm afraid the wrapper library might not become public until a week or two. We can notify you when it does.

edgargabriel:
@gbossu I sent you the testsuite by email

Issue #5529, opened 2018-08-09
incorrect logic in opal_config_asm.m4 for 128-wide atomic ops


flang-cavium:
Thank you for taking the time to submit an issue!

Patch is based on OpenMPI 3.1.1 - released tarball.

Compiled and installed from source.

Operating system: Ubuntu 16.04 AArch64 (ARM64).
Compilers:
- GCC 7.3 and 8.1
- clang/LLVM 6.0.1

Problem description:

The logic for discovering and enabling 128-bit wide atomic ops at ./configure time is incorrect. Essentially, 128-wide atomic ops are enabled regardless of whether or not the Target actually supports them. 

The file  opal_config_asm.m4 contains the following:

```
if test ! "$enable_cross_cmpset128" = "yes" ; then
### enable 128-wide atomic cmpset
fi
```

The logical test 

`if test ! "$enable_cross_cmpset128" = "yes" ; then`

is equivalent to

`if test "$enable_cross_cmpset128" = "no" ; then`

which means, cmpset128 will be enabled even though ./configure was told 

`--enable-cross-cmpset128=no`

Further down, there is another test for $enable_cross_cmpset128:

```
if test ! "$enable_cross_cmpset128" = "yes" ; then

### enable 128-wide atomic CAS

else

### enable 128-wide atomic CAS

fi
```

`if test ! "$enable_cross_cmpset128" = "yes" ; then`

is the equivalent of saying

`if test "$enable_cross_cmpset128" = "no" ; then`

Again, same logic problem, but this time around making certain 128-wide CAS really gets enable no matter what ./configure says.

Also: trying to attach a patch named opal_config_asm.m4.patch results in the following error:

**We don't support that file type. Try again with a GIF, JPEG, JPG, PNG, DOCX, GZ, LOG, PDF, PPTX, TXT, XLSX or ZIP.**

I seriously doubt anyone would ever submit a GIF, JPEG, JPG or PNG patch. I had to rename the patch file from opal_config_asm.m4.patch to opal_config_asm.m4.patch.txt.
[opal_config_asm.m4.patch.txt](https://github.com/open-mpi/ompi/files/2276111/opal_config_asm.m4.patch.txt)

bosilca:
[...]

PR #5445, opened 2018-07-17
Update opal to use C11 atomics if available

hjelmn:
This PR introduces a new set of types to be used for all opal atomics (opal_atomic_*) and adds support for C11 atomics. If C11 atomics are detected they are used exclusively.

This PR should be safe for v4.0.0 if we want to merge it today. I have no opinion either way.

PR #5546, opened 2018-08-15
Fix the detection of 128 bits atomics.

bosilca:
Thanks to @flang-cavium for identifying this issue and providing a proof-of-concept patch and a very accurate description of the issue in #5529.

It turns out we were missing the check for -latomic to get support for 128-bits atomics. Second we need to unset the local variables (_save) to avoid a conflict in OPAL_CHECK_SYNC_BUILTIN_CSWAP_INT128.

This patch might conflict with #5445, they are complementary and should be merged.

### DIFF
--- a/config/opal_config_asm.m4
+++ b/config/opal_config_asm.m4
@@ -2,14 +2,14 @@ dnl
 dnl Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
 dnl                         University Research and Technology
 dnl                         Corporation.  All rights reserved.
-dnl Copyright (c) 2004-2015 The University of Tennessee and The University
+dnl Copyright (c) 2004-2018 The University of Tennessee and The University
 dnl                         of Tennessee Research Foundation.  All rights
 dnl                         reserved.
 dnl Copyright (c) 2004-2006 High Performance Computing Center Stuttgart,
 dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
-dnl Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2008-2018 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
 dnl Copyright (c) 2015-2017 Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
@@ -24,65 +24,211 @@ dnl
 dnl $HEADER$
 dnl
 
+dnl This is a C test to see if 128-bit __atomic_compare_exchange_n()
+dnl actually works (e.g., it compiles and links successfully on
+dnl ARM64+clang, but returns incorrect answers as of August 2018).
+AC_DEFUN([OPAL_ATOMIC_COMPARE_EXCHANGE_N_TEST_SOURCE],[[
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+typedef union {
+    uint64_t fake@<:@2@:>@;
+    __int128 real;
+} ompi128;
+
+static void test1(void)
+{
+    // As of Aug 2018, we could not figure out a way to assign 128-bit
+    // constants -- the compilers would not accept it.  So use a fake
+    // union to assign 2 uin64_t's to make a single __int128.
+    ompi128 ptr      = { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
+    ompi128 expected = { .fake = { 0x11EEDDCCBBAA0099, 0x88776655443322FF }};
+    ompi128 desired  = { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};
+    bool r = __atomic_compare_exchange_n(&ptr.real, &expected.real,
+                                         desired.real, true,
+                                         __ATOMIC_RELAXED, __ATOMIC_RELAXED);
+    if ( !(r == false && ptr.real == expected.real)) {
+        exit(1);
+    }
+}
+
+static void test2(void)
+{
+    ompi128 ptr =      { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
+    ompi128 expected = ptr;
+    ompi128 desired =  { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};
+    bool r = __atomic_compare_exchange_n(&ptr.real, &expected.real,
+                                         desired.real, true,
+                                         __ATOMIC_RELAXED, __ATOMIC_RELAXED);
+    if (!(r == true && ptr.real == desired.real)) {
+        exit(2);
+    }
+}
+
+int main(int argc, char** argv)
+{
+    test1();
+    test2();
+    return 0;
+}
+]])
+
+dnl ------------------------------------------------------------------
+
+dnl This is a C test to see if 128-bit __sync_bool_compare_and_swap()
+dnl actually works (e.g., it compiles and links successfully on
+dnl ARM64+clang, but returns incorrect answers as of August 2018).
+AC_DEFUN([OPAL_SYNC_BOOL_COMPARE_AND_SWAP_TEST_SOURCE],[[
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+typedef union {
+    uint64_t fake@<:@2@:>@;
+    __int128 real;
+} ompi128;
+
+static void test1(void)
+{
+    // As of Aug 2018, we could not figure out a way to assign 128-bit
+    // constants -- the compilers would not accept it.  So use a fake
+    // union to assign 2 uin64_t's to make a single __int128.
+    ompi128 ptr    = { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
+    ompi128 oldval = { .fake = { 0x11EEDDCCBBAA0099, 0x88776655443322FF }};
+    ompi128 newval = { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};
+    bool r = __sync_bool_compare_and_swap(&ptr.real, oldval.real, newval.real);
+    if (!(r == false && ptr.real != newval.real)) {
+        exit(1);
+    }
+}
+
+static void test2(void)
+{
+    ompi128 ptr    = { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
+    ompi128 oldval = ptr;
+    ompi128 newval = { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};
+    bool r = __sync_bool_compare_and_swap(&ptr.real, oldval.real, newval.real);
+    if (!(r == true && ptr.real == newval.real)) {
+        exit(2);
+    }
+}
+
+int main(int argc, char** argv)
+{
+    test1();
+    test2();
+    return 0;
+}
+]])
+
+dnl ------------------------------------------------------------------
+
+dnl
+dnl Check to see if a specific function is linkable.
+dnl
+dnl Check with:
+dnl 1. No compiler/linker flags.
+dnl 2. CFLAGS += -mcx16
+dnl 3. LIBS += -latomic
+dnl 4. Finally, if it links ok with any of #1, #2, or #3, actually try
+dnl to run the test code (if we're not cross-compiling) and verify
+dnl that it actually gives us the correct result.
+dnl
+dnl Note that we unfortunately can't use AC SEARCH_LIBS because its
+dnl check incorrectly fails (because these functions are special compiler
+dnl intrinsics -- SEARCH_LIBS tries with "check FUNC()", which the
+dnl compiler complains doesn't match the internal prototype).  So we have
+dnl to use our own LINK_IFELSE tests.  Indeed, since these functions are
+dnl so special, we actually need a valid source code that calls the
+dnl functions with correct arguments, etc.  It's not enough, for example,
+dnl to do the usual "try to set a function pointer to the symbol" trick to
+dnl determine if these functions are available, because the compiler may
+dnl not implement these as actual symbols.  So just try to link a real
+dnl test code.
+dnl
+dnl $1: function name to print
+dnl $2: program to test
+dnl $3: action if any of 1, 2, or 3 succeeds
+dnl #4: action if all of 1, 2, and 3 fail
+dnl
+AC_DEFUN([OPAL_ASM_CHECK_ATOMIC_FUNC],[
+    OPAL_VAR_SCOPE_PUSH([opal_asm_check_func_happy opal_asm_check_func_CFLAGS_save opal_asm_check_func_LIBS_save])
+
+    opal_asm_check_func_CFLAGS_save=$CFLAGS
+    opal_asm_check_func_LIBS_save=$LIBS
+
+    dnl Check with no compiler/linker flags
+    AC_MSG_CHECKING([for $1])
+    AC_LINK_IFELSE([$2],
+        [opal_asm_check_func_happy=1
+         AC_MSG_RESULT([yes])],
+        [opal_asm_check_func_happy=0
+         AC_MSG_RESULT([no])])
+
+    dnl If that didn't work, try again with CFLAGS+=mcx16
+    AS_IF([test $opal_asm_check_func_happy -eq 0],
+        [AC_MSG_CHECKING([for $1 with -mcx16])
+         CFLAGS="$CFLAGS -mcx16"
+         AC_LINK_IFELSE([$2],
+             [opal_asm_check_func_happy=1
+              AC_MSG_RESULT([yes])],
+             [opal_asm_check_func_happy=0
+              CFLAGS=$opal_asm_check_func_CFLAGS_save
+              AC_MSG_RESULT([no])])
+         ])
+
+    dnl If that didn't work, try again with LIBS+=-latomic
+    AS_IF([test $opal_asm_check_func_happy -eq 0],
+        [AC_MSG_CHECKING([for $1 with -latomic])
+         LIBS="$LIBS -latomic"
+         AC_LINK_IFELSE([$2],
+             [opal_asm_check_func_happy=1
+              AC_MSG_RESULT([yes])],
+             [opal_asm_check_func_happy=0
+              LIBS=$opal_asm_check_func_LIBS_save
+              AC_MSG_RESULT([no])])
+         ])
+
+    dnl If we have it, try it and make sure it gives a correct result.
+    dnl As of Aug 2018, we know that it links but does *not* work on clang
+    dnl 6 on ARM64.
+    AS_IF([test $opal_asm_check_func_happy -eq 1],
+        [AC_MSG_CHECKING([if $1() gives correct results])
+         AC_RUN_IFELSE([$2],
+              [AC_MSG_RESULT([yes])],
+              [opal_asm_check_func_happy=0
+               AC_MSG_RESULT([no])],
+              [AC_MSG_RESULT([cannot test -- assume yes (cross compiling)])])
+         ])
+
+    dnl If we were unsuccessful, restore CFLAGS/LIBS
+    AS_IF([test $opal_asm_check_func_happy -eq 0],
+        [CFLAGS=$opal_asm_check_func_CFLAGS_save
+         LIBS=$opal_asm_check_func_LIBS_save])
+
+    dnl Run the user actions
+    AS_IF([test $opal_asm_check_func_happy -eq 1], [$3], [$4])
+
+    OPAL_VAR_SCOPE_POP
+])
+
+dnl ------------------------------------------------------------------
 
 AC_DEFUN([OPAL_CHECK_SYNC_BUILTIN_CSWAP_INT128], [
+  OPAL_VAR_SCOPE_PUSH([sync_bool_compare_and_swap_128_result])
 
-  OPAL_VAR_SCOPE_PUSH([sync_bool_compare_and_swap_128_result CFLAGS_save])
+  # Do we have __sync_bool_compare_and_swap?
+  # Use a special macro because we need to check with a few different
+  # CFLAGS/LIBS.
+  OPAL_ASM_CHECK_ATOMIC_FUNC([__sync_bool_compare_and_swap],
+      [AC_LANG_SOURCE(OPAL_SYNC_BOOL_COMPARE_AND_SWAP_TEST_SOURCE)],
+      [sync_bool_compare_and_swap_128_result=1],
+      [sync_bool_compare_and_swap_128_result=0])
 
-  AC_ARG_ENABLE([cross-cmpset128],[AC_HELP_STRING([--enable-cross-cmpset128],
-                [enable the use of the __sync builtin atomic compare-and-swap 128 when cross compiling])])
-
-  sync_bool_compare_and_swap_128_result=0
-
-  if test ! "$enable_cross_cmpset128" = "yes" ; then
-      AC_MSG_CHECKING([for processor support of __sync builtin atomic compare-and-swap on 128-bit values])
-
-      AC_RUN_IFELSE([AC_LANG_PROGRAM([], [__int128 x = 0; __sync_bool_compare_and_swap (&x, 0, 1);])],
-	  [AC_MSG_RESULT([yes])
-	      sync_bool_compare_and_swap_128_result=1],
-	  [AC_MSG_RESULT([no])],
-	  [AC_MSG_RESULT([no (cross compiling)])])
-
-      if test $sync_bool_compare_and_swap_128_result = 0 ; then
-	  CFLAGS_save=$CFLAGS
-	  CFLAGS="$CFLAGS -mcx16"
-
-	  AC_MSG_CHECKING([for __sync builtin atomic compare-and-swap on 128-bit values with -mcx16 flag])
-	  AC_RUN_IFELSE([AC_LANG_PROGRAM([], [__int128 x = 0; __sync_bool_compare_and_swap (&x, 0, 1);])],
-              [AC_MSG_RESULT([yes])
-		  sync_bool_compare_and_swap_128_result=1
-		  CFLAGS_save="$CFLAGS"],
-              [AC_MSG_RESULT([no])],
-	      [AC_MSG_RESULT([no (cross compiling)])])
-
-	  CFLAGS=$CFLAGS_save
-      fi
-  else
-      AC_MSG_CHECKING([for compiler support of __sync builtin atomic compare-and-swap on 128-bit values])
-
-      # Check if the compiler supports the __sync builtin
-      AC_TRY_LINK([], [__int128 x = 0; __sync_bool_compare_and_swap (&x, 0, 1);],
-	  [AC_MSG_RESULT([yes])
-	      sync_bool_compare_and_swap_128_result=1],
-	  [AC_MSG_RESULT([no])])
-
-      if test $sync_bool_compare_and_swap_128_result = 0 ; then
-	  CFLAGS_save=$CFLAGS
-	  CFLAGS="$CFLAGS -mcx16"
-
-	  AC_MSG_CHECKING([for __sync builtin atomic compare-and-swap on 128-bit values with -mcx16 flag])
-	  AC_TRY_LINK([], [__int128 x = 0; __sync_bool_compare_and_swap (&x, 0, 1);],
-              [AC_MSG_RESULT([yes])
-		  sync_bool_compare_and_swap_128_result=1
-		  CFLAGS_save="$CFLAGS"],
-              [AC_MSG_RESULT([no])])
-
-	  CFLAGS=$CFLAGS_save
-      fi
-  fi
-
-  AC_DEFINE_UNQUOTED([OPAL_HAVE_SYNC_BUILTIN_CSWAP_INT128], [$sync_bool_compare_and_swap_128_result],
-	[Whether the __sync builtin atomic compare and swap supports 128-bit values])
+  AC_DEFINE_UNQUOTED([OPAL_HAVE_SYNC_BUILTIN_CSWAP_INT128],
+        [$sync_bool_compare_and_swap_128_result],
+        [Whether the __sync builtin atomic compare and swap supports 128-bit values])
 
   OPAL_VAR_SCOPE_POP
 ])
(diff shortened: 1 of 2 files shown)


bosilca:
@jsquyres, @bwbarrett and @hppritcha once we converge on how we want to address this issue, we will need to push it to all stable releases. We can certainly address the cross-128 argument, but the purpose of this patch is to be minimally invasive and only address the detection of 128 atomic support.

hjelmn:
@bosilca The problem is often when the compare-exchange support for 128-bit atomics requires libatomic then the support is not lock-free (example gcc on x86-64). This is why I didn't add -latomic. Is there some case where libatomic is required and the resulting support is lock-free?

hjelmn:
To be fair I only tested that with icc and gcc. Both failed atomic_is_lock_free () for 128-bit atomics iff libatomic was required for atomic_compare_exchange_strong on 128-bit values.

bosilca:
@hjelmn I confirm, on the machines where I tested (x86_64 with different flavors of gcc and llvm) the atomics from -latomic are not lock-free. However, if we don't test for it we discard the check for 128-bits __sync and we would not mix the __atomic and __sync for x86_64. As a result all 128 bits atomics are disabled, which is not what we want.

We could modify the logic to always test for 128-bits sync if we fail to use the 128-bits __atomic. But, in the long term, compilers will fix their support for __atomic, and then with the current patch the detection will work seamlessly.

jsquyres:
@bosilca and I discussed this on the phone: see https://github.com/open-mpi/ompi/issues/5529#issuecomment-413291257.

The root cause of the problem, as identified by @flang-cavium in #5529, is that our `AC_TRY_RUN()` test is not strong enough on clang+ARM64.  Specifically: we're running this:

```c
__int128 x = 0;
__sync_bool_compare_and_swap (&x, 0, 1);
```

but we're not actually testing if that function *actually worked* -- we're just testing that that program *doesn't abort*.  Apparently, on clang+ARM64, this function doesn't cause the program to abort, and therefore `AC_TRY_RUN()` doesn't fail, so therefore we (incorrectly) conclude that this platform supports 128 bit atomics.

@bosilca is going to amend the tests to make sure that the function *actually worked*, and then hopefully we'll get a correct answer from this configure test on clang+ARM64.

Additionally, I'm going to add a commit or two on this PR about the `--enable-cross-*` CLI options -- see https://github.com/open-mpi/ompi/issues/5529#issuecomment-413291257 for more details.

flang-cavium:
Thank you for this.

OpenMPI is very important to us - Cavium. Our T99 (ARM64) chip is big in HPC/Super-Computing and OpenMPI is a very important part of that.

hjelmn:
@bosilca I though we already always checked for __sync. If that isn't working there is a problem.

https://github.com/open-mpi/ompi/blob/master/opal/include/opal/sys/gcc_builtin/atomic.h#L191

hjelmn:
And you are correct. A compiler may have a lock-free CAS-128 in libatomic so there is no problem with this PR.

hjelmn:
@jsquyres Crazy that it doesn't abort. That is an unexpected result. What does it do? Give a wrong answer or something?

jsquyres:
@hjelmn George hasn't added commits yet to make sure that the 128 bit test *actually works*.  So let's not merge yet!

hjelmn:
well. from the user comment it looked like they somewhat worked. that seems to be a compiler bug

jsquyres:
I think it might be worthwhile to use large values in the test that we run -- e.g., `~0` or something.  Perhaps the function actually only operate the lower 32 or 64 bits, leaving the upper bits alone...?  I think @flang-cavium would run whatever tests you ask him to run.

Worse would be if it *actually works* (e.g., via emulation), but not in an atomic fashion.  😲

bosilca:
I got access to am arm64. Just need to get all the compilers up-to-date.

flang-cavium:
> I think it might be worthwhile to use large values in the test that we run

Here's what happens when testing with values other than 0 and 1:

1. GCC 7.3 ARM64:

```
%> cat int128.c
#include <stdio.h>

int main()
{
  __int128 x = 123456789012345;
  __int128 y = 987654321098765;

  (void) fprintf(stderr, "before: x=%llu y=%llu\n", x, y);

  __int128 r = __atomic_compare_exchange_n (&x, &y,
                                            543210987654321,
                                            987650123498765,
                                            __ATOMIC_RELAXED,
                                            __ATOMIC_RELAXED);
  (void) fprintf(stderr, "after: x=%llu y=%llu r=%llu\n", x, y, r);

  return 0;
}
%> /usr/local/gcc73/bin/gcc -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -latomic -Wl,-rpath -Wl,/usr/local/gcc73/lib64 int128.c -o int128-gcc
%> echo $status
0
%> ./int128-gcc
before: x=123456789012345 y=0
after: x=123456789012345 y=0 r=123456789012345
%> echo $status
0
%>
```

2. clang 6.0.1 ARM64:

```
%> /opt/cavium/bin/clang -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -fslp-vectorize -fvectorize -Wl,-rpath -Wl,/opt/cavium/lib int128.c -o int128-clang
int128.c:8:53: warning: format specifies type 'unsigned long long' but the
      argument has type '__int128' [-Wformat]
  (void) fprintf(stderr, "before: x=%llu y=%llu\n", x, y);
                                    ~~~~            ^
int128.c:8:56: warning: format specifies type 'unsigned long long' but the
      argument has type '__int128' [-Wformat]
  (void) fprintf(stderr, "before: x=%llu y=%llu\n", x, y);
                                           ~~~~        ^
int128.c:15:59: warning: format specifies type 'unsigned long long' but the
      argument has type '__int128' [-Wformat]
  (void) fprintf(stderr, "after: x=%llu y=%llu r=%llu\n", x, y, r);
                                   ~~~~                   ^
int128.c:15:62: warning: format specifies type 'unsigned long long' but the
      argument has type '__int128' [-Wformat]
  (void) fprintf(stderr, "after: x=%llu y=%llu r=%llu\n", x, y, r);
                                          ~~~~               ^
int128.c:15:65: warning: format specifies type 'unsigned long long' but the
      argument has type '__int128' [-Wformat]
  (void) fprintf(stderr, "after: x=%llu y=%llu r=%llu\n", x, y, r);
                                                 ~~~~           ^
5 warnings generated.
%> echo $status
0
%> ./int128-clang
before: x=123456789012345 y=0
after: x=543210987654321 y=0 r=123456789012345
%> echo $status
0
%> uname -a
Linux VAL1-14 4.8.0-32-t99-perf #34 SMP Wed Jan 18 18:34:20 UTC 2017 aarch64 aarch64 aarch64 GNU/Linux
%>
```

Testing with 0 and 1 only is fraught with peril. Constant-folding to 0 and 1. Plus the 0 and 1 arguments are passed as integer constants to the __atomic* intrinsic call.

So the compiler will not actually use 128 bits to represent 0 or 1. It will use as little as possible.

flang-cavium:
> Worse would be if it actually works (e.g., via emulation), but not in an atomic fashion.

GCC's libatomic implements atomics as LL/SC with additional POSIX threads locking. Here's a snippet from the ARM64 disassembly:

```
0000000000001f00 <pthread_mutex_lock@plt>:
    1f00:       f0000090        adrp    x16, 14000 <memcpy@GLIBC_2.17>
    1f04:       f9401211        ldr     x17, [x16, #32]
    1f08:       91008210        add     x16, x16, #0x20
    1f0c:       d61f0220        br      x17

0000000000001f10 <pthread_mutex_unlock@plt>:
    1f10:       f0000090        adrp    x16, 14000 <memcpy@GLIBC_2.17>
    1f14:       f9401611        ldr     x17, [x16, #40]
    1f18:       9100a210        add     x16, x16, #0x28
    1f1c:       d61f0220        br      x17

[ ... ]

0000000000002b08 <libat_lock_1>:
    2b08:       927a1400        and     x0, x0, #0xfc0
    2b0c:       d0000081        adrp    x1, 14000 <memcpy@GLIBC_2.17>
    2b10:       91010021        add     x1, x1, #0x40
    2b14:       8b000020        add     x0, x1, x0
    2b18:       17fffcfa        b       1f00 <pthread_mutex_lock@plt>
    2b1c:       d503201f        nop

```
Also: LL/SC atomics can silently fail:

https://en.wikipedia.org/wiki/Load-link/store-conditional

hjelmn:
Yeah. They can fail which is why there is usually a loop around ll/sc.

bosilca:
But that is outside the scope of this discussion, as if __atomic_compare_exchange_n exists it's the implementors responsibility to make it follow the specifications. Otherwise, if they provide a function that has a broken implementation there is little we can do except preventing OMPI from being built with that specific compiler.

Btw in your example above where did you find the__atomic_compare_exchange_n prototype ? The 4th argument is supposed to be a bool and not a __int128 isn't it ?

flang-cavium:
> They can fail which is why there is usually a loop around ll/sc.

Here's a better idea: don't use LL/SC atomics on fully out-or-order CPU's. Which is why ARM came up with the lock-free atomics.

jsquyres:
@bosilca I think the issue is that `__atomic_compare_exchange_n()` exists and is buggy.  And potentially has been for a long time.  Should we submit a bug upstream?  Yes.  But we can also just strengthen our test a bit to make sure that `__atomic_compare_exchange_n()` isn't buggy.  We have done this with other functions, too.  It's just recognizing that the world isn't perfect.  ☹️

flang-cavium:
> But that is outside the scope of this discussion, as if __atomic_compare_exchange_n exists it's the implementors responsibility to make it follow the specifications.

1. It's not ouside scope, because we're discussing __int128 atomics on ARM64.
2. **__int128 is inherently non-portable.** It's not a Standard normative type, not in C and not in C++. It's a GNU extension. As such, "following the specification" doesn't really apply here, because there really is no normative specification for __int128. The most one can realistically say about the "specification" of __int128 is: it's Undefined Behavior. If it's Undefined Behavior, the compiler can do whatever it wants.

> Btw in your example above where did you find the__atomic_compare_exchange_n prototype?

copy-and-paste from OpenMPI's ./configure test for atomics.

bosilca:
You are wasting my time.

```$ ag __atomic_compare_exchange_n
opal_config_asm.m4
164:      AC_RUN_IFELSE([AC_LANG_PROGRAM([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);])],
175:          AC_RUN_IFELSE([AC_LANG_PROGRAM([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);])],
190:          AC_RUN_IFELSE([AC_LANG_PROGRAM([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);])],
218:      AC_TRY_LINK([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);],
228:          AC_TRY_LINK([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);],
242:          AC_TRY_LINK([], [__int128 x = 0, y = 0; __atomic_compare_exchange_n (&x, &y, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);],
267:__atomic_compare_exchange_n(&tmp, &old, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
269:__atomic_compare_exchange_n(&tmp64, &old64, 1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
```

flang-cavium:
> You are wasting my time.

That appears to be your default response/attitude when reality doesn't match your biases.

Not helpful.

bosilca:
In your code the arguments of __atomic_compare_exchange_n are incorrect. The return value should be bool and not __int128,  and you are only print 64 bits out of the 128. Moreover, running your code on an x86 show a similar output, despite the fact that we know for a fact that the atomic 128 bits are correct.

Now, the values printed as 64 bits make perfect sense as what you see as r is in fact the high of the y (because they are printed as 64 and not as 128 as push on the stack by the compiler).

flang-cavium:
GCC 7.3:

Test 1 passed! (r=0)
Test 2 passed! (r=1)

clang 6.0.1:

Test 1 failed! (r=0)
Test 2 failed! (r=0)

hjelmn:
Well. That is a compiler bug. Fun.

flang-cavium:
> That is a compiler bug.
It's not a compiler bug. 

As I have already explained:

**__int128 is non-portable.** It is not a Standard normative type.

__int128 is not mentioned anywhere in the C99 Standard, or in the C11 Standard. Or in the C++11 or C++14 Standard.

Undefined Behavior is not required to SEGV, SIGABRT, SIGILL, crash, panic the kernel or even print a diagnostic of any kind. The compiler is free to do whatever it wants with __int128 or Undefined Behavior. That includes making __int128 misbehave silently.

jsquyres:
It just means that we cannot use that function on that platform+compiler the way Open MPI currently uses it.  We'll update the `.m4` to use the stronger test (i.e., to see if that function behaves the way that Open MPI expects it to behave).

If someone from the ARM community would like to update our atomics to use something better on those platforms (e.g., the lockless methods you used), a pull request would be appreciated.  

Thanks!

hjelmn:
Since you seem to enjoy pointing out that int128 is a gcc extension then the C standard does not apply here but gcc's documentation does:

The ‘__atomic’ builtins can be used with any integral scalar or pointer type that is 1, 2, 4, or 8 bytes in length. 16-byte integral types are also allowed if ‘__int128’ (see __int128) is supported by the architecture.

Thus if clang claims to have __int128 then if if allows __atomic_compare_exchange on that type (defines it) then it *must* work and be atomic. So it absolutely is a compiler bug. Go file it with llvm.

flang-cavium:
1. clang is not GCC.
2. There are plenty of GCC extensions that clang doesn't support, and there are plenty of clang extensions that GCC doesn't support. Which is why it's never a good idea to rely on these types of compiler extensions.
3. Writing **portable software** implies staying away from non-Standard Language or Compiler extensions.
4. Is OpenMPI interested in being portable, or not?

bwbarrett:
Frankly, Open MPI is not portable.  It can not be portable without giving up performance.  It uses memory barriers, it depends on certain memory model implementations, it does a number of things required for high performance.

I do believe that this is a compiler bug that LLVM should fix.  It's claiming to support a GCC extension and doing a crappy job at it.  That's a compiler bug.  The moment it chose to provide __int128, it chose to support all expected variants of that usage.  Maybe that wasn't the right choice for them, maybe it was.  Either way, it puts developers who are trying to use those (rational, well defined) semantics in a tough spot.

jsquyres:
Folks: it doesn't matter who is right at this point.  Let's cut off the discussion here, because it has devolved into expressing opinions that aren't likely to change.

I mentioned above that I am working on updating the `.m4` to make Open MPI's configure script realize that it can't use these functions on this platform (I'll likely need to ask @flang-cavium to run another test for `__sync_bool_compare_and_swap()`, but it likely won't be until tomorrow at this point).  If the ARM community is interested in submitting pull requests to have Open MPI do something different on ARM64, that would be awesome (but fair warning: all this atomic code is pretty subtle, gnarly, and there be lots of dragons there).

hjelmn:
Jeff, before you do let me merge the c11 stuff. It has a bunch of configure changes

jsquyres:
@hjelmn No problem.

flang-cavium:
I'll be happy to test anything you'd like me to test for ARM64.

As long as you do not force ARM64 to use __int128 atomics. ;-)

gpaulsen:
I'm glad the ARM-wrestling is over.

jsquyres:
@flang-cavium Could you please try running this:

```c
#include <stdio.h>
#include <stdint.h>
#include <stdbool.h>
#include <stdlib.h>

typedef union {
    uint64_t fake[2];
    __int128 real;
} ompi128;

static void test3(void)
{
    ompi128 ptr    = { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
    ompi128 oldval = { .fake = { 0x11EEDDCCBBAA0099, 0x88776655443322FF }};
    ompi128 newval = { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};

    bool r = __sync_bool_compare_and_swap(&ptr.real,
                                          oldval.real,
                                          newval.real);

    if (r == false && ptr.real != newval.real) {
        printf("Test 3 passed!\n");
    } else {
        printf("Test 3 failed!\n");
    }
}

static void test4(void)
{
    ompi128 ptr    = { .fake = { 0xFFEEDDCCBBAA0099, 0x8877665544332211 }};
    ompi128 oldval = ptr;
    ompi128 newval = { .fake = { 0x1122DDCCBBAA0099, 0x887766554433EEFF }};

    bool r = __sync_bool_compare_and_swap(&ptr.real,
                                          oldval.real,
                                          newval.real);

    if (r == true && ptr.real == newval.real) {
        printf("Test 4 passed!\n");
    } else {
        printf("Test 4 failed!\n");
    }
}

int main(int argc, char** argv[])
{
    test3();
    test4();

    return 0;
}
```

bosilca:
To correct some of the misinformation on this discussion, according to the [ARM v8 ISA](http://www.keil.com/support/man/docs/armclang_asm/armclang_asm_pge1427897400799.htm) it is possible to implement 128-bits atomic compare-and-swap on ARM64. Something like the following might do the work:

```c
static inline bool
opal_atomic_compare_exchange_strong_rel_128( volatile __int128 *addr,
                                             __int128 *oldval,
                                             __int128 newval)
{
    __int128 prev;
    int tmp;
    bool ret;

    __asm__ __volatile__ ("1:  ldaxp   %0, %H0, [%2]   \n"
                          "    cmp     %0, %3          \n"
                          "    bne     2f              \n"
                          "    cmp     %H0, %H3        \n"
                          "    bne     2f              \n"
                          "    stlxp   %w1, %4, %H4, [%2]   \n"
                          "    cbnz    %w1, 1b         \n"
                          "2:                          \n"
                          : "=&r" (prev), "=&r" (tmp)
                          : "r" (addr), "r" (*oldval), "r" (newval)
                          : "cc", "memory");

    ret = (prev == *oldval);
    *oldval = prev;
    return ret;
}
```

flang-cavium:
1. GCC 7.2 X86_64:

```
/usr/local/gcc72/bin/gcc -D_GNU_SOURCE -O3 -std=c99 -mtune=corei7 -mfma -mavx -L/usr/local/gcc72/lib64 -latomic -Wl,-rpath -Wl,/usr/local/gcc72/lib64 ompi.c -o ompi-gcc
/tmp/cczyGOGE.o: In function `main':
ompi.c:(.text.startup+0x45): undefined reference to `__sync_bool_compare_and_swap_16'
ompi.c:(.text.startup+0x90): undefined reference to `__sync_bool_compare_and_swap_16'
```

2. GCC 7.3 ARM64:

```
/usr/local/gcc73/bin/gcc -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -L/usr/local/gcc73/lib64 -latomic -Wl,-rpath -Wl,/usr/local/gcc73/lib64 ompi.c -o ompi-gcc
/tmp/ccIhwsYm.o: In function `main':
ompi.c:(.text.startup+0x7c): undefined reference to `__sync_bool_compare_and_swap_16'
ompi.c:(.text.startup+0xe8): undefined reference to `__sync_bool_compare_and_swap_16'
```

3. clang 6.0.1 ARM64:

```
/opt/cavium/bin/clang -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -fslp-vectorize -fvectorize -Wl,-rpath -Wl,/opt/cavium/lib ompi.c -o ompi-clang
ompi.c:45:5: error: second parameter of 'main' (argument array) must be of type
      'char **'
int main(int argc, char** argv[])
    ^
1 error generated.
```

Let's try again with int main(int argc, char* argv[]):

```
/opt/cavium/bin/clang -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -fslp-vectorize -fvectorize -Wl,-rpath -Wl,/opt/cavium/lib ompi.c -o ompi-clang
%> ./ompi-clang
Test 3 failed!
Test 4 passed!
%>
```

jsquyres:
For the gcc tests, please compile with `-mcx16`

flang-cavium:
> For the gcc tests, please compile with -mcx16

Doesn't exist on ARM64:
```
/usr/local/gcc73/bin/gcc -D_GNU_SOURCE -O3 -std=c99 -mcpu=thunderx2t99 -march=armv8.1-a+lse -ffast-math -ffp-contract=fast -funroll-loops -finline-functions -mcx16 -L/usr/local/gcc73/lib64 -latomic -Wl,-rpath -Wl,/usr/local/gcc73/lib64 ompi.c -o ompi-gcc
gcc: error: unrecognized command line option ‘-mcx16’
```

hjelmn:
A key thing missing from this PR is the check __atomic_is_lock_free (). The test should fail if that returns false. See https://github.com/open-mpi/ompi/pull/5445/files#diff-2067567c58452d8c159cd3f59dbbb624R48

hjelmn:
@bosilca PowerPC has similar support and gcc uses it for compare-exchange 128. I will give that a try on our ARM systems and see what I get.

bosilca:
@flang-cavium `-latomic` is missing on the gcc compile command (`mcx16` is only for x86).

flang-cavium:
> -latomic is missing on the gcc compile command (mcx16 is only for x86).

I don't know why you keep making these kinds of innacurate comments, over, and over, and over, again. 

-latomic is present on compile line. Both for X86_64 and for ARM64.

ggouaillardet:
@flang-cavium would you mind posting a command line with `gcc` on `x86_64` that evidences `-latomic` is used ?

flang-cavium:
> would you mind posting a command line with gcc on x86_64 that evidences -latomic is used?

I already did. See the compile examples I posted above for GCC 7.2 on X86_64, GCC 7.3 on ARM64 and clang 6.0.1 on ARM64 (you have to scroll up a bit).

And also see my previous comment about __sync_bool_compare_and_swap on X86_64 with GCC. It's not a function call.

This is on X86_64:

```
%> readelf -sW /usr/local/gcc72/lib64/libatomic.so.1.2.0 | egrep compare_and_swap
%>
```

There is no symbol containing "compare_and_swap" defined in GCC's libatomic.so.

You can compile and link jsquyres' test case with GCC and -mcx16 and not pass -latomic on compile-line at all. The program will compile and link correctly:

```
%> /usr/local/gcc72/bin/gcc -D_GNU_SOURCE -O3 -std=c99 -mtune=corei7 -mfma -mavx -mcx16 -Wl,-rpath -Wl,/usr/local/gcc72/lib64 ompi.c -o ompi-gcc
%> echo $status
0
%> ./ompi-gcc
Test 3 passed!
Test 4 passed!
%> echo $status
0
%>
```

hjelmn:
Because it isn't compare_and_swap. Look at the C11 atomics for the API. The __atomic calls are the underlying implemention of both C11 and C++11 atomics.

flang-cavium:
On 08/17/2018 10:20 AM, Nathan Hjelm wrote:
> External Email
>
> Because it isn't compare_and_swap. Look at the C11 atomics for the API.
> The __atomic calls are the underlying implemention of both C11 and C++11
> atomics.

We are in C99. Yes? What does C11 have to do with any of this?

Please show me the Symbol Table Entry for __sync_bool_compare_and_swap 
in GCC's libatomic.so.

readelf -sW libatomic.so.1.X.X | egrep __sync_bool_compare_and_swap

That's what jsquyres' test program calls.

-- 
Stefan Teleman
Cavium
stefan.teleman@cavium.com

hjelmn:
@flang-cavium I am done with discussing this with you. You clearly know very little about atomic support, compilers, or language standards. I do not get paid to argue with trolls.

flang-cavium:
On 08/17/2018 10:44 AM, Nathan Hjelm wrote:
> External Email
>
> @flang-cavium <https://github.com/flang-cavium> I am done with
> discussing this with you.

Great idea.

-- 
Stefan Teleman
Cavium
stefan.teleman@cavium.com

hjelmn:
Sounds good. That is what we want for 4.0.x and earlier for sure. I will update the c11 pr and make sure we have everything covered there.

shamisp:
@bosilca please let me run this by arm compiler team.

bosilca:
@shamisp please do so. We are looking to get this in 4.0 so we are bound by the release time constraints.

config/opal_config_asm.m4
```diff
@@ -27,13 +27,13 @@ dnl
...
 dnl ARM64+clang, but returns incorrect answers as of August 2018).
-AC_DEFUN([OPAL_ATOMIC_COMPARE_EXCHANGE_N_TEST_SOURCE],[
+AC_DEFUN([OPAL_ATOMIC_COMPARE_EXCHANGE_N_TEST_SOURCE],[[
 #include <stdint.h>
 #include <stdbool.h>
 #include <stdlib.h>
 
 typedef union {
-    uint64_t fake@<:@2@:>@
+    uint64_t fake@<:@2@:>@;
```
jsquyres: Oops!  Good catch; thanks. 

Note that with the use of the quadrigraphs, the double quoting `[[]]` is not necessary.

bosilca: Without the double `[[]]` only the first few lines of the C code are in the configure test so all tests fails because the code is incomplete. I didn't try to understand why, I went for the simplest protection mechanism, the double `[[]]`.

jsquyres: Huh; weird.  I played with the quoting quite a bit -- for me, using Autoconf 2.69, the double quoting was not necessary.  Shrug.  It's not harmful; might as well leave it.

jsquyres:
Sorry I missed all the fun today -- it was a busy day.

@bosilca Thanks for the fixes.

The more I think about this sequence of tests, though, the more I'm convinced we didn't do it quite right.  I think both the tests for `__atomic_compare_exchange_n()` and `__sync_bool_compare_and_swap()` should be transformed into:

```
for X in __atomic_compare_exchange_n __sync_bool_compare_and_swap
    try with flags "" "-latomic" "mxc16":
        if function X is linkable:
            AC_RUN_IFELSE(...the correctness check for X..., happy=1, happy=0,
                  [happy=1 AC_MSG_WARN([you are cross compiling; we assume it works])])
```

Specifically: all we want to do is verify that the function is both prototyped and can be linked, and if so, if we're not cross compiling, actually try the function and make sure it works (because we know of a platform where it doesn't work the way that we want it to).

If we do it this way, the confusing use of the `--enable-cross-*` flags are unnecessary.

I will make an attempt at re-working the code to be this way.  It should be quite a bit simpler.

jsquyres:
@bosilca @hjelmn @bwbarrett @ggouaillardet @shamisp I just pushed a new commit that is a simpler approach (like I described above).  Please review.

The overall diff is a bit confusing (especially because I added a commit at the end that transforms tabs into spaces -- sorry).  You might want to look at the "SQUASHME: new, simpler approach" commit to see exactly what was done.

ibm-ompi:

 Gist: https://gist.github.com/813fcc3a1f56e9ac28a9f1ab404e3d60

ibm-ompi:

 Gist: https://gist.github.com/99abd2a407ef9d98d3d7e69bc76db20a

ibm-ompi:

 Gist: https://gist.github.com/9075a0081acf47bcc137ea87816685a5

jsquyres:
@gpaulsen @jjhursey Could use a little help figuring out why the IBM CI (XL Compiler) is failing CI here.  The Fortran linker is complaining that it can't find `-latomic`, but at this point in the build, the C linker has already succeeded (including `-latomic`) many times.

1. Any idea why the Fortran linker can't find `-latomic`, but the C linker can?
1. Can you send the output from configure + `config.log` + `config.status` from a master run on the same platform as CI (i.e., from master head -- not this PR) so that I can compare what this PR's configury is doing differently than master?

Thanks!

bosilca:
The missing header for printf is what's hurting. Instead of printf you can simply exit with different error codes.
```
configure:36508: gcc -o conftest -g -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing   conftest.c  -lz >&5
conftest.c:238:5: warning: ISO C does not support '__int128' types [-Wpedantic]
      __int128 real;
      ^~~~~~~~
 conftest.c: In function 'test1':
 conftest.c:254:9: error: implicit declaration of function 'printf' [-Werror=implicit-function-declaration]
          printf("Test 1 passed!\n");
          ^~~~~~
```

bosilca:
Too much fiddling, the code is from some perspective cleaner but not more readable. Btw, did we just lost the check for __atomic_always_lock_free (ito make sure the __atomic 128 CASis not implemented with a lock)?

ibm-ompi:

 Gist: https://gist.github.com/da52edc81c7083c52d11e61913fd5e2d

ibm-ompi:

 Gist: https://gist.github.com/b23f7a771b089b855df108530ec8a32b

ibm-ompi:

 Gist: https://gist.github.com/dd1c1cc00dabe7a129d6082aecb7da47

jsquyres:
@bosilca I'm confused.  On the ARM64 CI (Thunder), it's passing the "does the `__atomic` operation actually work" test -- but it shouldn't (right?).  

I added some printf's to verify that it is actually passing the test (since we can't easily get the `config.log`).  From https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.platforms/3429/Platform=ARMv8/console:

```
...
checking for fgrep... /bin/grep -F
checking for __atomic builtin atomics... yes
checking for __atomic_compare_exchange_n... no
checking for __atomic_compare_exchange_n with -mcx16... no
checking for __atomic_compare_exchange_n with -latomic... yes
checking if __atomic_compare_exchange_n works... Test 1 passed!
Test 2 passed!
yes
checking if __int128 atomic compare-and-swap is always lock-free... no
checking for __sync_bool_compare_and_swap... no
checking for __sync_bool_compare_and_swap with -mcx16... no
checking for __sync_bool_compare_and_swap with -latomic... no
checking if .proc/endp is needed... no
...
```

The "checking if __atomic_compare_exchange_n works..." line should *fail* on ARM64 (with both gcc and clang), right?

Ultimately, we get the "right" answer by concluding that neither `__atomic_*` nor `__sync_*` work the way we want, but it doesn't seem like we get the right answer about `__atomic_*` the right way...

shamisp:
Tagging @ nSircombe

jsquyres:
@bosilca Is armclang the same thing as clang?  To my un-educated eye (and a few very quick google searchers), that sounds like an ARM-specific version of clang (that requires a license).  

And if ARM has its own distribution of clang, then it's not the same thing as regular/upstream/community clang.  That may mean that community/open source clang does one thing, and ARM armclang does another (and therefore we have to support both)...?

nSircombe:
Armclang is optimised for AArch64 and the particular server-class implementations, and tweaked to help HPC workloads.
Armclang and clang should behave equivalently on this - taking the tests out of opal_config_asm.m4 from this PR (with the printf's still in) both tests 1 and 2 pass for clang 5.0.1 and Arm 18.4 on AArch64.

nSircombe:
Looking at the LLVM bugzilla example (https://bugs.llvm.org/show_bug.cgi?id=38621) with the Arm HPC Compiler 18.4, and 'vanilla' clang 5.0.1, on an AArch64 machine I'm seeing 4x passes.

So ___sync_bool_compare_and_swap, __sync_val_compare_and_swap appear to be implemented and working (not lockless) in clang5. 
ummm, I'm not sure what's going on with the clang6 results posted.

nSircombe:
Just tried the test from https://bugs.llvm.org/show_bug.cgi?id=38621 on an AArch64 machine with a newer version of Clang - v6.0.2 (http://llvm.org/git/clang.git 31f349d884b2007b6328de58832478a738759c73, http://llvm.org/git/llvm.git d88ca2927027d254931437c6f6c83f43724945af)

I get passes for all 4 tests.

nSircombe:
Setting -mcpu=thunderx2t99 appears to cause problems with LLVM 6. 
No -mcpu flag (or -mcpu=generic) results in 4 'passes'

clang5 and armclang both work with -mcpu=thunderx2t99, and armclang works with -mcpu=native (where the compiler will automatically pickup the TX2).

nSircombe:
I've been able to discuss this with some of the compiler developers at Arm.
There appears to be a bug in the Large System Extension (LSE) in LLVM 6 - switching to/from thunderx2 or turning on/off lse in the -march flag demonstrates this. 
There's some more info back on the LLVM bugzilla ticket (https://bugs.llvm.org/show_bug.cgi?id=38621).

The 128bit atomic calls are implemented (and working) in armclang and clang 5, but they are not lock-free, and can't be on AArch64 as it stand - as @flang-cavium noted during the discussion of these issues.

jsquyres:
@nSircombe Thanks for all the info!  I think that means that we're largely good-to-go (sorry, I don't think I previously grokked the difference between the AArch64 and ARM64 architectures).

@gpaulsen today confirmed on our weekly webex that IBM will look at the `-latomic` failure to see what's what there.  Once that's sorted out, we'll squash/merge.

nSircombe:
@jsquyres - I might have seeded some confusion with my choice of names! 
ARMv8-a is the architecture (there are a couple of revisions of this and some extensions (lse, sve...)), AArch64 is the 64-bit ISA. ARM64 is colloquially used to refer to AArch64 (usually)

So there's no lockless implementation available for anything built on a 64bit Arm platform.

shamisp:
@nSircombe just to clarify the clarification :) There are no lockless 128bit operation for aarch64.
Armv8-a defines LSE which can be used to implement lockless operation for 64bit.

nSircombe:
Yes, sorry. Missed out the 128bit caveat!

@jsquyres let me know if there’s anything you want me to check with armclang (or clang 5 or 6 for that matter).

________________________________

jsquyres:
@nSircombe @shamisp Can you guys test this PR and see that it behaves the way you expect it to on your platforms with the various compilers?

You can look in the `configure` stdout and look for the tests for `__sync_*` and `__atomic_*`.

ggouaillardet:
@kawashima-fj it would be worth testing this PR if/when the Fujitsu environment for ARM is ready.

nSircombe:
Only had chance to do a quick-pass, but here goes:
PR builds ok with the Arm compiler (clang 5), and aquick-check with a mini-app worked as expected.
`
$ cd build
$ CC=armclang CXX=armclang++ FC=armflang ../configure --prefix=${install_dir} --enable-mpirun-prefix-by-default 2>&1 | tee -a ompi-build.log
$ make -j all
$ make install
$ grep "__atomic_*" build/config.log 
configure:36329: checking for __atomic builtin atomics
configure:36448: checking for __atomic_compare_exchange_n
configure:36726: checking if __atomic_compare_exchange_n works
$ grep "__sync_*" build/config.log 
$ grep "__atomic_*" build/ompi-build.log 
checking for __atomic builtin atomics... yes
checking for __atomic_compare_exchange_n... yes
checking if __atomic_compare_exchange_n works... Test 1 passed!
checking for __atomic builtin atomics... yes
checking for processor support of __atomic builtin atomic compare-and-swap on 128-bit values... yes
checking for __atomic builtin atomics... yes
checking for __atomic_compare_exchange_n... yes
checking if __atomic_compare_exchange_n works... Test 1 passed!
$ grep "__sync_*" build/ompi-build.log 
`

nSircombe:
On closer inspection, I don't think this is working as expected. A quick grep through the stdout for 'atomic' gives...
With GCC 7.1 on AArch64:
```
checking if gcc  supports C11 atomic variables... yes
checking size of atomic_short... 2
checking size of atomic_int... 4
checking size of atomic_long... 8
checking size of atomic_llong... 8
checking for __atomic builtin atomics... yes
checking for __atomic_compare_exchange_n... no
checking for __atomic_compare_exchange_n with -mcx16... no
checking for __atomic_compare_exchange_n with -latomic... yes
checking if __atomic_compare_exchange_n works... Test 1 passed!
checking if __int128 atomic compare-and-swap is always lock-free... no
checking for __sync_bool_compare_and_swap with -latomic... no
checking for builtin atomics... BUILTIN_GCC
checking for struct ibv_exp_device_attr.exp_atomic_cap... no
checking if gcc  supports C11 atomic variables... yes
checking for __atomic builtin atomics... yes
checking for processor support of __atomic builtin atomic compare-and-swap on 128-bit values... no
checking for __atomic builtin atomic compare-and-swap on 128-bit values with -mcx16 flag... no
checking for builtin atomics... BUILTIN_GCC
checking for frameworks for oshmem... atomic, memheap, scoll, spml, sshmem
```

For Arm 18.4 (clang 5) on AArch64:

```
checking if armclang  supports C11 atomic variables... yes
checking size of atomic_short... 2
checking size of atomic_int... 4
checking size of atomic_long... 8
checking size of atomic_llong... 8
checking for __atomic builtin atomics... yes
checking for __atomic_compare_exchange_n... yes
checking if __atomic_compare_exchange_n works... Test 1 passed!
checking if __int128 atomic compare-and-swap is always lock-free... yes
checking for builtin atomics... BUILTIN_GCC
checking for struct ibv_exp_device_attr.exp_atomic_cap... no
checking if armclang  supports C11 atomic variables... yes
checking for __atomic builtin atomics... yes
checking for processor support of __atomic builtin atomic compare-and-swap on 128-bit values... yes
checking if __int128 atomic compare-and-swap is always lock-free... yes
checking for builtin atomics... BUILTIN_GCC
checking for frameworks for oshmem... atomic, memheap, scoll, spml, sshmem
```

'checking if __int128 atomic compare-and-swap is always lock-free... yes' - unless I'm mistaken, we know this isn't lock free?

nSircombe:
Hummm...
```
static void test5(void)
{

    bool r = __atomic_always_lock_free(16, 0);

    if (r == true) {
        printf("Test 5 passed!\n");
    } else {
        printf("Test 5 failed\n");
    }
}
```

Fails on GCC (as expected), passes with clang 5. I don't believe this this is correct.

nSircombe:
And looking in a bit more detail, the +lse issue is present (https://bugs.llvm.org/show_bug.cgi?id=38621#c1) in both clang 5 (and armclang 18.3) and clang 6. Building with -march=armv8.1a+lse and/or -mcpu=[something other than generic] at -O3, where Tests 1 & 3 (from https://bugs.llvm.org/show_bug.cgi?id=38621#c0) fail. I had missed the O3 flag off my clang5 test before.

bosilca:
What if they have it implemented with LL/SC ?

jsquyres:
@bosilca FYI: I squashed everything.  However, in my code review before pushing, I found one more use of `$enable_cross_*` that needs to be updated to the new philosophy.  I don't have time at this exact moment to fix it; it should be straightforward, though.

ibm-ompi:

 Gist: https://gist.github.com/15b6d10ac7daff3d196c8df27f47a795

bosilca:
The last cross-compile test is now gone.

ibm-ompi:

 Gist: https://gist.github.com/8179434c2fe6073cee6172c152a1f29c

jsquyres:
@bosilca Thanks!  I squashed your commit.

@gpaulsen Ping. We really need IBM's assistance to know why your CI is failing.

ibm-ompi:

 Gist: https://gist.github.com/6c3a6b79546f43d29415d8a8853d1452

jsquyres:
@jjhursey got this log from one of IBM's failed CI builds: [config.log.gz](https://github.com/open-mpi/ompi/files/2319405/config.log.gz). It's probably not enough info (i.e., doesn't answer why the C compiler can find `-latomic` and the Fortran compiler can't), but I figured I'd put it here on the issue for rerefence / safekeeping.

hjelmn:
Can we not change the configury so that the fortran bindings do not try to use -latomic. fortan has no notion of threading nor atomics and as long as ompi/orte/opal are linked against it we are fine.

jsquyres:
`$LIBS` and `$LDFLAGS` are shared between all the compilers (this is an Autotools assumption) -- the assumption is that the back-end linker is the same for all the front-end compilers.

jsquyres:
I'd like to compare a configure output from current master to a configure output to this PR.  I.e., why hasn't this been a problem with the XL Fortran compiler before?

bosilca:
The check for - latomic was added in this PR.

jsquyres:
Actually, come to think of it, if we need `-latomic` for C, then we'll also need it to compile C and Fortran MPI apps (i.e., in `mpicc` and `mpifort`).  So it doesn't make sense to separate out the C and Fortran `$LIBS` just for this issue.

Meaning: the Fortran linker needs to be able to find `-latomic`.

IBM: please advise.

ibm-ompi:

 Gist: https://gist.github.com/7a36c9443b32c20d4c2a135db2e915c8

jsquyres:
@jjhursey @gpaulsen Can you grab a configure output / `config.log` from the latest failed CI build?  I made a change in the configury, and I'd like to see the output.  Thanks.

hjelmn:
@jsquyres What was the problem?

jsquyres:
TL;DR: we weren't resetting CFLAGS/LIBS properly in one failure case.

More detail: it's still weird that IBM's XL CI environment has `-latomic` for the C compiler/linker but not for the Fortran compiler/linker.  But with closer examination of the config.log that Josh sent, we shouldn't have been using `-latomic` anyway -- because it wasn't always lock-free.  Once I fixed the "hey, we have -latomic but we don't want to use it" path to properly reset CFLAGS/LIBS, then IBM's XL CI environment isn't trying to use `-latomic`, and all is good.

FWIW, IBM may still want to report this to their XL dev team.  Even though this has turned out to not be an issue for Open MPI, it can still be an issue when you need XLF to link a combined C/Fortran application.

hjelmn:
I still don't think it is needed for fortran. If the C code was linked with -latomic that is sufficient. Fortran can link to it without issue.

bosilca:
@hjelmn we only have one `$LIBS`. If the 128 bits atomic implementation in `-latomic` become lockless, our current test will add `-latomic` to `$LIBS` and will therefore impact the linking of all Fortran code.

bosilca:
@jsquyres 👍 . I can't approve the PR, github has a legit policy that prevents authors from approving their own patches.

jsquyres:
george

jsquyres:
Before we merge, since there has been sooo many changes on this PR, I'd like to get a final round of confirmations from ARM people that this PR actually works for them.

@shamisp @nSircombe Can you confirm that this PR works properly?

shamisp:
Hi Jeff, I would like to run it by few more developers before we sign off.
Is it okay to wait until Monday COB ?

jsquyres:
@shamisp No problem.

nSircombe:
Some more digging with the HPC Compiler team clears things up a bit...

The equivalent of @bosilca suggestion (https://github.com/open-mpi/ompi/pull/5546#issuecomment-415454078) is implemented in clang on Armv8, on a pair of 64bit ints, in a loop. 

So the configure log all makes sense, and I think the configure logic is fine now.

There are some issues for us that we may need to look at (Armv8.1, .2... implement this slightly differently, which appears to impact the __atomics. And the lse extension is clearly causing problems).

But for this PR things look ok - if @shamisp is happy, then I'm happy too.

jsquyres:
@nSircombe Thanks!
@shamisp Are you happy?  😉

shamisp:
@jsquyres  @nSircombe  - I'm okay with existing check.

For the record @hjelmn - considering OpenMPI use case I want to implement ompi internal version of 128bit AMOs using LDXRP/STXRP.

PR #5549, opened 2018-08-15
v2.1.5: looks like this will be needed after all

jsquyres:
Doing a v1.2.5 release to fix a serious vader BTL bug.  Also pick up a
trivial UCX compile/link bug.

Note that no shared library version number changes are needed -- the only 2 changes were in btl/vader and pml/ucx.


bwbarrett:
Given --enable-static, I generally bump the revision for the overall library on a component change, but I don't have strong feelings.

Issue #5502, opened 2018-07-31
opal/util/info.c can still truncate info keys


jsquyres:
On a build on master HEAD with gcc 7.3.0, I see the following (also seen by @rhc54 in https://www.mail-archive.com/devel@lists.open-mpi.org/msg20671.html):

```
91  In file included from info.c:46:0:
92  info.c: In function ‘opal_info_dup_mode’:
93  ../../opal/util/info.h:112:31: warning: ‘%s’ directive output may be truncated writing up to 36 bytes into a region of size 27 [-Wformat-truncation=]
94   #define OPAL_INFO_SAVE_PREFIX "_OMPI_IN_"
95                                 ^
96  info.c:212:22: note: in expansion of macro ‘OPAL_INFO_SAVE_PREFIX’
97                        OPAL_INFO_SAVE_PREFIX "%s", iterator->ie_key);
98                        ^~~~~~~~~~~~~~~~~~~~~
99  info.c:212:45: note: format string is defined here
100                       OPAL_INFO_SAVE_PREFIX "%s", iterator->ie_key);
101                                              ^~
102 info.c:211:18: note: ‘snprintf’ output between 10 and 46 bytes into a destination of size 36
103                   snprintf(savedkey, OPAL_MAX_INFO_KEY,
104                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
105                       OPAL_INFO_SAVE_PREFIX "%s", iterator->ie_key);
106                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I.e., it looks like #5204 didn't actually fix the problem.

@markalle Can you please fix?

I did not check to which back release branches this will also need to be applied.

rhc54:
Also seeing these on my Mac using gcc-7.3:
```shell
In file included from /opt/local/lib/gcc7/gcc/x86_64-apple-darwin17/7.3.0/include-fixed/stdio.h:425:0,
                 from ../../opal/class/opal_list.h:71,
                 from ../../opal/util/info_subscriber.h:30,
                 from info_subscriber.c:45:
info_subscriber.c: In function 'opal_infosubscribe_change_info':
../../opal/util/info.h:112:31: warning: '%s' directive output may be truncated writing up to 36 bytes into a region of size 27 [-Wformat-truncation=]
 #define OPAL_INFO_SAVE_PREFIX "_OMPI_IN_"
                               ^
[...]

PR #5553, opened 2018-08-16
snprintf() length fix for info

markalle:
The important part of this fix is a couple places 5 was hard-coded that needed to be
strlen(OPAL_INFO_SAVE_PREFIX).

But also this contains a fix for a gcc 7.3.0 compiler warning about snprintf(). There
was an "if" statement making sure all the arguments had appropriate strlen(), but gcc
still complained about the following snprintf() because the size of the struct element
is iterator->ie_key[OPAL_MAX_INFO_KEY + 1].

Fixes https://github.com/open-mpi/ompi/issues/5502

### DIFF
--- a/opal/util/info.c
+++ b/opal/util/info.c
@@ -176,7 +176,7 @@ int opal_info_dup_mode (opal_info_t *info, opal_info_t **newinfo,
 {
     int err, flag;
     opal_info_entry_t *iterator;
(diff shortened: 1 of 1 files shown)


hppritcha:
Hmmm... should this be cherry-picked into v4.0.x and possibly other branches?

PR #5558, opened 2018-08-18
VERSION: Bump to v2.1.6a1.

jsquyres:
Let's hope we never need it.


jsquyres:
@hppritcha Please review and merge when CI has completed.

Issue #5560, opened 2018-08-20
common/ompio: ordering problem during file_open


edgargabriel:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
 
master, v4.0

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

compiled from git sources.

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

in theory any systems, although it was
-----------------------------

## Details of the problem

There is an ordering problem when opening a file w.r.t to setting file view vs. opening shared file pointer component. This can trigger a spurious error message from the lockedfile component,

PR #5562, opened 2018-08-20
common/ompio: fix an ordering problem during file_open

edgargabriel:
the sharedfp component has to be selected and opened before
we set the default file view during file_open. Otherwise
there is a sperious error message from the sharefp_file_seek
operation that is called during the file_set_view.

Fixes Issue #5560


raafatfeki:
@hppritcha Yes sure

PR #5564, opened 2018-08-20
Update to PMIx v2.1.3

rhc54:

### DIFF
--- a/opal/mca/pmix/pmix2x/pmix/NEWS
+++ b/opal/mca/pmix/pmix2x/pmix/NEWS
@@ -21,11 +21,12 @@ example, a bug might be fixed in the master, and then moved to the
 current release as well as the "stable" bug fix release branch.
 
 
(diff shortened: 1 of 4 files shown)


ggouaillardet:
I think we should address pmix/pmix#817 and then upgrade to `PMIx 2.1.4` (or backport the fix)

jsquyres:
Per 2018-08-21 webex, @rhc54 says that this is not a blocker for release.  It's a bug that occurs during PMIx notifications, and the only time OMPI uses those is when something else is already going wrong (i.e., error notifications).  It would be good to fix, but it's not a blocker.

jjhursey:
IT sounds like we are holding this one until we get a PMIx patch on top of 2.1.3, right?

rhc54:
No, we are holding until PMIx release 2.1.4 and then will update this PR

Issue #5584, opened 2018-08-23
v2.0.x: UCX links failure on 2 CI platforms


jsquyres:
Looks like we're getting the same UCX link failures on 2 different CI platforms (see the CI on #5582), seemingly indicating that this may be a legit issue:

```
In file included from pml_ucx.c:19:0:
pml_ucx_request.h:102:6: warning: ISO C99 doesn't support unnamed structs/unions [-Wpedantic]
     };
      ^
In file included from pml_ucx_request.c:13:0:
pml_ucx_request.h:102:6: warning: ISO C99 doesn't support unnamed structs/unions [-Wpedantic]
     };
      ^
pml_ucx.c: In function 'mca_pml_ucx_bsend':
pml_ucx.c:553:5: warning: format '%d' expects argument of type 'int', but argument 5 has type 'size_t' [-Wformat=]
     PML_UCX_VERBOSE(8, "bsend of packed buffer %p len %d", packed_data, packed_length);
     ^
  CCLD     mca_pml_ucx.la
libtool:   error: require no space between '-L' and '-lrt'
make[2]: *** [mca_pml_ucx.la] Error 1
make[2]: Leaving directory `/tmp/ompici/ompi_public_pr_release_gnu/ompi-src/ompi/mca/pml/ucx'
```

@hoopoepg Can you investigate?

FYI @derbeyn

jsquyres:
Fixed via #5587.

PR #5582, opened 2018-08-23
v2.0.x: osc/rdma: fix a segfault in MPI_Win_unlock due to an access to a NULL…

derbeyn:
fixes #5581.

Signed-off-by: Nadia Derbey <Nadia.Derbey@atos.net>

PR #5587, opened 2018-08-23
PML/UCX/BUILD: fix for build on non-default UCX location

hoopoepg:
- fixes https://github.com/open-mpi/ompi/issues/5584

### DIFF
--- a/config/ompi_check_ucx.m4
+++ b/config/ompi_check_ucx.m4
@@ -1,8 +1,11 @@
 # -*- shell-script -*-
 #
 # Copyright (C) 2015      Mellanox Technologies Ltd. ALL RIGHTS RESERVED.
-# Copyright (c) 2015-2016 Research Organization for Information Science
+# Copyright (c) 2015-2017 Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+#                         reserved.
+# Copyright (c) 2016 Cisco Systems, Inc.  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


hoopoepg:
@jsquyres fix should be in release of 2.1, but I'm not sure if this fix is enough, build error is really strange

jsquyres:
@hoopoepg Ok.  Please apply fixes to master first and then cherry pick to master whenever possible in the future.

(please also check if this needs to go to v3.0.x, v3.1.x, and v4.0.x -- thanks!)

hoopoepg:
actually all other branches have this fix, but there are more changes which should be applied to fix issue in v2.0 (that's why I added WIP label), let's wait for build complete & see does it helps or not :)

jsquyres:
@hoopoepg Ah, gotcha.  Thanks!

ibm-ompi:

 Gist: https://gist.github.com/069d254758bd4938578c3c11bccfe813

jsquyres:
Sad panda -- didn't fix the issue for IBM.  😦

hoopoepg:
yep, one more commit

config/ompi_check_ucx.m4
```diff
@@ -16,65 +19,72 @@
...
+	AS_IF([test "$ompi_ucx_version_ok" = "no"], [ompi_check_ucx_happy="no"])
+
+    fi
 
-    AS_IF([test "$ompi_ucx_version_ok" = "no"], [ompi_check_ucx_happy="no"])
 
     AS_IF([test "$ompi_check_ucx_happy" = "yes"],
-          [$2],
+          [$1_CPPFLAGS="[$]$1_CPPFLAGS $ompi_check_ucx_CPPFLAGS"
+	   $1_LDFLAGS="[$]$1_LDFLAGS $ompi_check_ucx_LDFLAGS"
```
jsquyres: Watch tabs vs. spaces here.

Open MPI is allergic to tabs.  😉

hoopoepg: will fix if it works

hoopoepg:
@jsquyres I squashed & beautify commits, it should work for now

config/ompi_check_ucx.m4
hoopoepg: @jsquyres btw, I copied this file from 2.x branch, it means that all release branches have tab/space mix :)

hoopoepg:
@jsquyres done, all tests are passed

jsquyres:
I did not notice that #5582 was on v2.0.x.  Let's wait to see what @derbeyn says on that issue as to whether we really need to fix that old branch or not.

PR #5590, opened 2018-08-23
MTL OFI: Ask for FI_THREAD_DOMAIN support as needed

aravindksg:
When an application is not using multiple threads to call into MPI, we can
safely ask for FI_THREAD_DOMAIN setting from the provider as it should
translate to the least amount of locking in provider.

Conversely, for applications using THREAD_MULTIPLE, explicitly ask for
FI_THREAD_SAFE to prevent race conditions.


rhc54:
git failure

bot:ompi:retest

PR #5591, opened 2018-08-23
osc/rdma: clean out stale aggregation code

hjelmn:
The aggregation code in osc/rdma is currently broken and will likely
not be reused. This commit cleans it out.


hppritcha:
@hjelmn  should this go back to v4.0.x?

hjelmn:
Probably worth putting into v4.0.x. Will open a PR

PR #5597, opened 2018-08-24
Allow run-as-root if 2 envars are set

### LINKED ISSUE #4451: Environment Var For --allow-run-as-root

openmpi 3.0.0

from source, via the [spack](https://github.com/spack/spack) package manager inside a Docker container.

* Operating system/version: Ubuntu 16.04
* Computer hardware: x86
* Network type: Ethernet

-----------------------------

## Details of the problem

I would like to control the `--allow-run-as-root` flag via an environment variable (or config file) as well.

Running as root in Docker containers is fine and partitioning by exporting an environment variable (e.g. autotools equivalent is `FORCE_UNSAFE_CONFIGURE=1`) would make it way easier to interact with mpirun/mpiexec for deeply integrated workflows (where directly editing the final mpi call might not always be possible).

rhc54:
Per suggestion by @bangerth, allow mpirun to execute as root if two
envars are set to specific values

Fixes #4451


simonbyrne:
Should this part be changed as well?:
https://github.com/open-mpi/ompi/blob/ae1f7e0c3bda617cb4f4f3a44b79ac56103bb8f1/orte/tools/orterun/orterun.c#L144-L160

PR #5600, opened 2018-08-25
Deal with special case during cleanup

rhc54:
In some scenarios, we can have a daemon sharing the node with mpirun. In
those cases, we need to avoid race conditions in cleanup


rhc54:
@hjelmn I _think_ this will fix the problem you've been having on trinitite - can you please confirm?

rhc54:
Woohoo!! Thanks!

rhc54:
Will bring this around to 4.x - do you need it on 3.x as well?

hppritcha:
@gpaulsen we probably want this in v4.0.x as well

Issue #5593, opened 2018-08-23
OPAL ERROR: Error in file %s/opal/mca/pmix/s1/pmix_s1.c at line 578


morrisonlevi:
## Background information

I am trying to get OpenMPI and Slurm working working with PMIx, but have been hitting issues in various places.

### What versions of software are you using?

  - OpenMPI 3.1.2
  - Slurm 17.11.9
  - PMIx 2.1.3

All three of these were built from source.

PMIx is installed to `/usr/local/pmix/2.1` and there is an `ld.so.conf.d` file that includes its libdir in the ld.so.cache.

Slurm is installed to `/usr/include/slurm`, `/usr/lib64/slurm`, etc.

OpenMPI is configured with:

    '--prefix=/apps/openmpi/3.1.2/gcc-7.3.0_cuda-9.2.88_pmix-2.1.3'
    '--enable-wrapper-rpath=no'
    '--enable-wrapper-runpath=no' '--with-slurm'
    '--with-pmix=/usr/local/pmix/2.1'
    '--with-pmi=/usr/local/pmix/2.1'
    '--with-pmi-libdir=/usr/local/pmix/2.1/lib'
    '--with-cuda=/apps/cuda/9.2.88'
    '--with-libevent=/usr'
    'CPPFLAGS=-I/usr/local/pmix/2.1/include'
    'LDFLAGS=-L/usr/local/pmix/2.1/lib'

Side note: I was unsure what to do for `--with-pmi` -- I have tried various other settings and this has been the version that has worked the best so far.

Side note 2: I got errors if I did not specific `-I` and `-L` in configure flags -- that shouldn't be necessary should it?

### Please describe the system on which you are running

* Operating system/version: RHEL 7
* Computer hardware: Probably N/A
* Network type: Probably N/A

-----------------------------

## Details of the problem

If I launch an MPI program with `srun --mpi=openmpi ...` I get this error:

> OPAL ERROR: Error in file %s/opal/mca/pmix/s1/pmix_s1.c at line 578

Which can be seen [here](https://github.com/open-mpi/ompi/blob/0a9c358ef8adc7225bdbe97c5fa7741ef0e26297/opal/mca/pmix/s1/pmix_s1.c#L578). Then, instead of the program using *n* ranks there are *n* independent runs of the program. In this example there are 4 tasks all on one node; the issue manifests in other configurations too:

```bash
$ srun -n 4 --mpi=openmpi $prog
[...]

PR #5607, opened 2018-08-27
sharedfp/sm and lockedfile: fix naming bug

edgargabriel:
If an application opens a file for reading from multiple processes
using MPI_COMM_SELF (or another communicator that has distinct
process groups but the same comm-id, as can happen as the result
of comm_split), the naming chosen for the lockedfile or the mmapped
file used by the sharedfp/sm component would collide. This patch
ensures that the filename is different by integrating the process id
of rank 0 for each sub-communicator.

This fixes one aspect of the problem reported in github issue #5593

### DIFF
--- a/ompi/mca/sharedfp/lockedfile/sharedfp_lockedfile_file_open.c
+++ b/ompi/mca/sharedfp/lockedfile/sharedfp_lockedfile_file_open.c
@@ -35,6 +35,7 @@
 #include <sys/stat.h>
 #endif
 #include <fcntl.h>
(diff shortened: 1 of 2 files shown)


hppritcha:
@edgargabriel could you check the "5593" above?  It doesn't seem to be the right number.

edgargabriel:
@hppritcha the issue number is correct. If you scroll down, there is an MPI I/O problem reported as well, and this is the fix that resulted from that report.

Issue #5611, opened 2018-08-27
file_delete does not trigger fs selection logic correctly under certain circumstances


edgargabriel:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
master, v4.0.x

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

checkout from git

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 
any
-----------------------------

## Details of the problem

MPI_File_delete segfaults if the first file I/O operation executed by an application is a File_delete operation, since the selection logic in io_base_delete does not trigger the mca_base_framework open of the fs framework and the corresponding mca_fs_base_find_available. 

This is only an issue on master and 4.0.x due to the recent changes introduced for the file_delete operation. The 'correct' implementation of file_delete does now trigger the fs selection logic and call the delete function of the selected fs-component, which is not possible if the fs framework has not yet been opened.

PR #5616, opened 2018-08-28
io/base: fixes to file_delete selection logic

edgargabriel:
file_delete triggers underneath the hood the full component selection
logic, since we do not have a file handle, just a file name.

As part of the selection logic, we have to however initiate the
framework-open of the fs component in case of ompio, since ompio
will call the delete function of the selected fs componentn, which
is based on the file system where the file is located.

This was not handled correctly so far. The problem however only
shows up if the first I/O operatin to be executed is a file_delete,
other wise the file_open will lead to the correct opening and initialization
of the fs framework. This commit ensures that we do the right thing
even if file_delete is the first file I/O operation in the application.

Fixes issue #5611


gpaulsen:
Thanks!

PR #5619, opened 2018-08-28
v4.0.x: atomic 128 fixes

jsquyres:
Bring over all 3 commits from #5546 (including the tabs->spaces commit -- it applied cleanly to the v4.0.x branch, whereas it did not apply cleanly to the v3.0.x or v3.1.x branches).


hppritcha:
@hjelmn please review, this caused enough issues on master, want to be extra careful that this is right.

PR #5621, opened 2018-08-28
MTL OFI: Fix race condition due to global progress entries array

aravindksg:
Since progress entries array is globally allocated, it is susceptible
to race conditions when using multi-threaded applications. Allocating it
on the stack resolves any potential races as it is thread local by default.

(cherry picked from commit ed2343034d09b33eb44a0a727bef97a108edc8aa)

Conflicts:
	ompi/mca/mtl/ofi/mtl_ofi_component.c
	ompi/mca/mtl/ofi/mtl_ofi_types.h

Conflicts were mainly due to some features in master and not in the release branch. So git was complaining about missing lines while cherry-picking.


jsquyres:
Per discussion in 2018-09-18 webex, good to go.  Merging...

PR #5630, opened 2018-08-29
osc/portal: use c99 subobject naming to initialize module

hjelmn:


hjelmn:
@tkordenbrock Can you take a look. The osc/portals module had the no-longer supported get and set info functions.

tkordenbrock:
@hjelmn Looks good to me.  Thanks for fixing this.

PR #5636, opened 2018-08-30
v4.0.x: make common/verbs-usnic actually check if it can compile

jsquyres:
Previously, we only checked to see if the user specified `--with-verbs-usnic`.  Now do the Right Thing and actually verify that it can compile, too.

### DIFF
--- a/README
+++ b/README
@@ -1020,15 +1020,33 @@ NETWORKING SUPPORT / OPTIONS
   configurations.
 
 --with-verbs-usnic
+  Note that this option is no longer necessary in recent Linux distro
+  versions.  If your Linux distro uses the "rdma-core" package (instead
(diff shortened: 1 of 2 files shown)


hppritcha:
@bturrubiates could you review this?

jsquyres:
Nope -- there's no version that I'm aware of in rdma-core where `ibv_device.ops` exists.

This is all we're looking for: this component in Open MPI is solely to address an old pre-`rdma-core` `libibverbs` bug.  If we're using an `rdma-core`-`libibverbs`, then this component is not needed.

jsquyres:
@hppritcha @gpaulsen Ben doesn't have write access to this repo, so Github didn't count his review.  Can one of you do a nominal review so that github allows this PR to be merged?

Thanks!

PR #5641, opened 2018-09-02
User is effective root on Haiku

extrowerk:
While it is possible to run the tools as different user, no software expects it, so let's just run it on Haiku.
Hope my logic is ok.

### DIFF
--- a/orte/orted/orted_submit.c
+++ b/orte/orted/orted_submit.c
@@ -323,7 +323,7 @@ int orte_submit_init(int argc, char *argv[],
      * us to proceed if the allow-run-as-root flag was given. Otherwise,
      * exit with a giant warning flag
      */
-    if (0 == geteuid() && !orte_cmd_options.run_as_root) {
+    if (0 == geteuid() && !orte_cmd_options.run_as_root && !defined(__HAIKU__)) {
         /* check for two envars that allow override of this protection */
         char *r1, *r2;
         if (NULL != (r1 = getenv("OMPI_ALLOW_RUN_AS_ROOT")) &&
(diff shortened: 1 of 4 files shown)


bgoglin:
Can't you change the default orte_cmd_options.run_as_root on Haiku instead of adding !defined HAIKU everywhere it's checked?

extrowerk:
@bgoglin : can you maybe explain with a bit more details? Thanks a lot.

bgoglin:
orte_cmd_options.run_as_root is here to allow running as root, which is exactly what you seem to want enabled by default on Haiku. So change it from false to true by default on Haiku.

extrowerk:
@bgoglin : that's clear, but where do i supposed to change that? Grepping the source tree doesn't gives me any hint.

rhc54:
If you look at the source code you are modifying, you will see that we allow you to define two envars that override the ```run-as-root``` check. Please just set those envars instead of modifying the code.

extrowerk:
@rhc54 Yep, that would solve the problem for me, but i'm trying to package openmpi for Haiku, and one cannot assume every user would do it. And if the buildbots building other ports wich requires openmpi, they would fail. So i see 2 options:
- changing the openmpi code to make it useful on Haiku
- exporting env vars in every recipe

For me, the first solution looks nicer.

rhc54:
I'm afraid that option is unacceptable. We have had similar discussions for quite some time and remain adamant that the code not default to allow run as root. We have had too many issues where users render the operating system useless by making a mistake. We only recently (as in a week ago) finally agreed to allow the two-envar system for bypassing it - not going to budge further than that, I'm afraid.

extrowerk:
@jsquyres : As i told, user is effective root on Haiku. One can create another user, but the system and the GUI still runs as root. Everything what you guys says makes sense, but not on Haiku.

Thanks for the nice explanation and the plenty warning however, you guys are rock!

PR #5642, opened 2018-09-03
Use less memory for the MCA variable names.

bosilca:
Small modifications to reduce the memory footprint of our MCA parameters.

The biggest memory hog in the MCA framework are the OPAL objects themselves, but this optimization reduces the memory needs by about 10% without any loss of functionality.

### DIFF
--- a/opal/mca/base/mca_base_var.c
+++ b/opal/mca/base/mca_base_var.c
@@ -1381,7 +1381,7 @@ static int register_variable (const char *project_name, const char *framework_na
         var->mbv_type        = type;
         var->mbv_flags       = flags;
         var->mbv_group_index = group_index;
-        var->mbv_info_lvl  = info_lvl;
+        var->mbv_info_lvl    = info_lvl;
         var->mbv_scope       = scope;
         var->mbv_synonym_for = synonym_for;
         var->mbv_bind        = bind;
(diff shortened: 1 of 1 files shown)


opal/mca/base/mca_base_var.c
```diff
@@ -1588,31 +1584,27 @@ int mca_base_var_register_synonym (int synonym_for, const char *project_name,
...
-    ret = asprintf (&source_env, "%sSOURCE_%s", mca_prefix, name);
-    if (0 > ret) {
-        return OPAL_ERROR;
-    }
-
-    ret = asprintf (&value_env, "%s%s", mca_prefix, name);
+    ret = snprintf(envvar, 128, "%s%s", mca_prefix, name);
     if (0 > ret) {
-        free (source_env);
+        opal_output(0, "Variable string too short to hold %s%s\n", mca_prefix, name);
```
jsquyres: This basically enforces an upper bound on MCA framework+param names.  If that's the case, we should probably enforce that when the MCA var is registered (with a helpful message to the developer -- because this is an Open MPI developer error, not a user error).

bosilca: Indeed it set an upper bound. The output will be seen by OMPI developers way before the code gets merged in master (or at least it is supposed to because of CI).

jsquyres: But shouldn't this check be at the registration layer?

bosilca: We could, but as we surround the MCA var with additional text to create the env name we would need to compute the correct upper-bound during registration.

jsquyres: Might be more deterministic to put it at the registration layer.  Here, the max length -- while huge -- is a function of the framework name, the component name, and the variable name.  Hence, some components could have longer var names than others.

Perhaps we should set max lengths on each of these and enforce them, and then the max length of the env variable will be obvious / easily calculable as the sum of those parts.

Sidenote: Don't we have max lengths for framework and component names already?

bosilca: Yes, we do for framework name (MCA_BASE_MAX_TYPE_NAME_LEN -> 31) and for component name (MCA_BASE_MAX_COMPONENT_NAME_LEN -> 63). We are missing upper-bounds for project name and variable name.

PR #5646, opened 2018-09-05
MCA base var followup

jsquyres:
Ensure that the project, framework, component, and variable names are lower than max lengths.  This is a follow-on to 992a8e8297f7f044ef3b30bfc235566d43d68cb5, per discussion on https://github.com/open-mpi/ompi/pull/5642.
    
NOTE: The current max lengths for project, framework, component, and variable name add up to more than 128 (per discussion on #5642):

https://github.com/open-mpi/ompi/blob/42b0e3bd6161cfaafcb66db008bdc96de84f52d6/opal/mca/mca.h#L247-L258

With MCA_BASE_VAR_MAX_VARIABLE_NAME_LEN set to 63 in this PR, that's (15+31+63+63)=172.

We *could* reduce the length of the TYPE_LEN and COMPONENT_NAME_LEN to, say, 15 and 31, respectively, bringing the total down to (15+15+31+63)=124.  But this:

1. Would be a major ABI break
1. Would require cherry picking to be in 4.0.0 (not 4.0.1 or later!)
1. Would likely require bumping the MCA version number from 2 to 3
1. Would require some extra code to safely handle opening DSOs from earlier versions (even if we just end up closing/discarding those components).

That seems like a bunch of extra work for a pretty minor win.  So I didn't do it in this PR.

### DIFF
--- a/opal/mca/base/mca_base_var.c
+++ b/opal/mca/base/mca_base_var.c
@@ -1284,6 +1284,24 @@ static int register_variable (const char *project_name, const char *framework_na
     /* Developer error. Storage can not be NULL and type must exist */
     assert (((flags & MCA_BASE_VAR_FLAG_SYNONYM) || NULL != storage) && type >= 0 && type < MCA_BASE_VAR_TYPE_MAX);
 
+    /* Developer error: check max length of strings */
+    if (NULL != project_name &&
+        strlen(project_name) > MCA_BASE_MAX_PROJECT_NAME_LEN) {
+        return OPAL_ERR_BAD_PARAM;
+    }
+    if (NULL != framework_name &&
+        strlen(framework_name) > MCA_BASE_MAX_TYPE_NAME_LEN) {
+        return OPAL_ERR_BAD_PARAM;
+    }
+    if (NULL != component_name &&
+        strlen(component_name) > MCA_BASE_MAX_COMPONENT_NAME_LEN) {
+        return OPAL_ERR_BAD_PARAM;
+    }
+    if (NULL != variable_name &&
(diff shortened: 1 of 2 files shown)


jsquyres:
@bosilca Fixed -- look at the commit to be squashed.

opal/mca/base/mca_base_var.c
```diff
@@ -1584,10 +1602,16 @@ int mca_base_var_register_synonym (int synonym_for, const char *project_name,
...
+        MCA_BASE_MAX_PROJECT_NAME_LEN +
+        MCA_BASE_MAX_TYPE_NAME_LEN +
+        MCA_BASE_MAX_COMPONENT_NAME_LEN +
+        MCA_BASE_MAX_VARIABLE_NAME_LEN +
+        4 + 1; // 4=max number of _'s, 1=trailing \0
+    char envvar[max_len];
     int ret;
 
-    ret = snprintf(envvar, 128, "%s%s", mca_prefix, name);
+    ret = snprintf(envvar, max_len, "%s%s", mca_prefix, name);
```
bosilca: you need to change the snprintf below as well. And there you have to account for the "SOURCE_" additional string.

jsquyres:
Isn't `alloca()` going to be morally equivalent to `char foo[X]` in this case?

bosilca:
Except you could do ```maxlen = 1 + strlen(mca_prefix) + strlen(SOURCE_PREFIX) + strlen(name);```

jsquyres:
Oh, ok.  Can do.

jsquyres:
New commit pushed.

PR #5647, opened 2018-09-05
patcher/base: improve instruction cache flush for aarch64

### LINKED ISSUE #5631: flush_and_invalidate_cache is not valid for ARM64 platform

## Background information
ARM64 based platforms need actions to invalidate instruction cache after code-modification

Actual for all OMPI versions 3.0+ (not tested on other versions)

* Operating system/version: 
* Computer hardware: 
ARM64
* Network type: 

-----------------------------

## Details of the problem
Implementing binary patching in UCX we found that ARM64 processor's instruction cache is actual for a long time which makes binary patching not effective some time after patching completed (may be executed original code instead of updated instructions). To force flushing instruction cache compiler intrinsic __clear_cache may be used, for reference UCX sources may be used:
https://github.com/openucx/ucx/blob/bac22184987518b990b006bf9f02bbff46745d28/src/ucs/arch/cpu.h#L79
and
https://github.com/openucx/ucx/blob/bac22184987518b990b006bf9f02bbff46745d28/src/ucs/arch/aarch64/cpu.h#L99

@hjelmn what do you think?

hjelmn:
This commit updates the patcher component to either use the
__clear_cache intrinsic or the correct assembly to flush the
instruction cache.

Fixes #5631

### DIFF
--- a/opal/mca/patcher/base/patcher_base_patch.c
+++ b/opal/mca/patcher/base/patcher_base_patch.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2016      Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2016-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
@@ -107,7 +107,11 @@ static void flush_and_invalidate_cache (unsigned long a)
 #elif OPAL_ASSEMBLY_ARCH == OPAL_IA64
     __asm__ volatile ("fc %0;; sync.i;; srlz.i;;" : : "r"(a) : "memory");
 #elif OPAL_ASSEMBLY_ARCH == OPAL_ARM64
-    __asm__ volatile ("dsb sy");
+    __asm__ volatile ("dc cvau, %0\n\t"
+                      "dsb ish\n\t"
+                      "ic ivau, %0\n\t"
+                      "dsb ish\n\t"
+                      "isb":: "r" (a));
 #endif
 }
 
@@ -138,9 +142,26 @@ static inline void apply_patch (unsigned char *patch_data, uintptr_t address, si
 {
     ModifyMemoryProtection (address, data_size, PROT_EXEC|PROT_READ|PROT_WRITE);
     memcpy ((void *) address, patch_data, data_size);
-    for (size_t i = 0 ; i < data_size ; i += 16) {
+#if HAVE___CLEAR_CACHE
+    /* do not allow global declaration of compiler intrinsic */
+    void __clear_cache(void* beg, void* end);
+
+    __clear_cache ((void *) address, (void *) (address + data_size));
+#else
+    size_t offset_jump = 16;
+
+#if OPAL_ASSEMBLY_ARCH == OPAL_ARM64
+    offset_jump = 32;
+#endif
+
+    /* align the address */
+    address &= ~(offset_jump - 1);
+
+    for (size_t i = 0 ; i < data_size ; i += offset_jump) {
         flush_and_invalidate_cache (address + i);
     }
+
+#endif
 
     ModifyMemoryProtection (address, data_size, PROT_EXEC|PROT_READ);
 }
--- a/configure.ac
+++ b/configure.ac
@@ -858,7 +858,7 @@ OPAL_SEARCH_LIBS_CORE([ceil], [m])
 # -lrt might be needed for clock_gettime
 OPAL_SEARCH_LIBS_CORE([clock_gettime], [rt])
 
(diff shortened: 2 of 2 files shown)


hjelmn:
@shamisp @hoopoepg This should fix the aarch64 cache flush.

jsquyres:
@jjones-cavium Can you test this PR, too?  Thanks!

shamisp:
Adding @nSircombe

opal/mca/patcher/base/patcher_base_patch.c
```diff
@@ -138,10 +141,24 @@ static inline void apply_patch (unsigned char *patch_data, uintptr_t address, si
 {
     ModifyMemoryProtection (address, data_size, PROT_EXEC|PROT_READ|PROT_WRITE);
     memcpy ((void *) address, patch_data, data_size);
-    for (size_t i = 0 ; i < data_size ; i += 16) {
+#if HAVE___CLEAR_CACHE
+        /* do not allow global declaration of compiler intrinsic */
```
jsquyres: Minor indenting issue here.

hoopoepg: it seems tab/space mix :)

hjelmn: C&P from UCX. Will fix.

opal/mca/patcher/base/patcher_base_patch.c
```diff
@@ -107,7 +107,10 @@ static void flush_and_invalidate_cache (unsigned long a)
 #elif OPAL_ASSEMBLY_ARCH == OPAL_IA64
     __asm__ volatile ("fc %0;; sync.i;; srlz.i;;" : : "r"(a) : "memory");
 #elif OPAL_ASSEMBLY_ARCH == OPAL_ARM64
-    __asm__ volatile ("dsb sy");
+    __asm__ volatile ("dc cvau, %0\n\t"
+                      "dsb ish\n\t"
+                      "ic ivau, %0\n\t"
+                      "dsb ish\n\t" :: "r" (a));
```
shamisp: @hjelmn few comments. 
```dsb ish``` -> ```dsb ish; isb```
it is not that obvious but ```isb``` (instruction barrier) is required here because DSB does not impose order on all instruction but memory access only. We don't want some thing like jump executed out of order so we have to use ISB in addition to DSB.

File: opal/mca/patcher/base/patcher_base_patch.c [hunk 3]
```diff
@@ -138,10 +141,24 @@ static inline void apply_patch (unsigned char *patch_data, uintptr_t address, si
...
+#if HAVE___CLEAR_CACHE
+    /* do not allow global declaration of compiler intrinsic */
+    void __clear_cache(void* beg, void* end);
+
+    __clear_cache ((void *) address, (void *) (address + data_size));
+#else
+    size_t offset_jump = 16;
+
+#if OPAL_ASSEMBLY_ARCH == OPAL_ARM64
+    offset_jump = 4;
```
shamisp: how is the offsets are calculated ? (4, 16)

opal/mca/patcher/base/patcher_base_patch.c
hjelmn: Ok, will add isb.

opal/mca/patcher/base/patcher_base_patch.c
shamisp: Double checking ...

shamisp: Arm (Section D3.4) advices to use cache line increments obtained from ```ctr_el0``` (DminLine, IminLine fields),similar to: https://github.com/openucx/ucx/blob/bac22184987518b990b006bf9f02bbff46745d28/src/ucs/arch/aarch64/cpu.h#L144 

it is log2 based value. I think the minimum value for Armv8 is 32 bytes (16 words) but your setup most likely to have 64 bytes. I guess, 32 will work but you just hitting the same d/i-cacheline twice.

hjelmn: Ok, I see. I will just use 32-bytes then. Not like this routine is being run too often :).

hjelmn: Ok, pushed.

opal/mca/patcher/base/patcher_base_patch.c
```diff
@@ -138,10 +142,27 @@ static inline void apply_patch (unsigned char *patch_data, uintptr_t address, si
...
+    size_t offset_jump = 16;
+
+#if OPAL_ASSEMBLY_ARCH == OPAL_ARM64
+    offset_jump = 32;
+#endif
+
+    /* align the address */
+    address &= ~(offset_jump - 1);
+
+    for (size_t i = 0 ; i < data_size ; i += ofset_jump) {
```
nSircombe: Typo? 
ofset_jump should be offset_jump

hjelmn: Yup. Typo. Thanks for catching that.

hjelmn:
@hppritcha This will be headed to v4.0.x once merged.

nSircombe:
I'm building on AArch64 with the Arm compiler, and trying out  a few test codes just as a sanity check - I'll let you know how I get on.

nSircombe:
I've not found any issues building with GCC 7 and Arm 18.4 (and LLVM 7) all working as expected so far.

PR #5650, opened 2018-09-06
v4.0.x: MTL PSM2: Remove shadow variables from v4.0.x

matcabral:
As agreed on #4574, where removed in past release branches
to avoid perfomance impacts in the default values for
some paramters.

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -84,129 +84,6 @@ mca_mtl_psm2_component_t mca_mtl_psm2_component = {
     }
 };
 
-struct ompi_mtl_psm2_shadow_variable {
-    int variable_type;
-    void *storage;
-    mca_base_var_storage_t default_value;
-    const char *env_name;
-    mca_base_var_info_lvl_t info_level;
-    const char *mca_name;
-    const char *description;
-    mca_base_var_flag_t flags;
-};
-
-struct ompi_mtl_psm2_shadow_variable ompi_mtl_psm2_shadow_variables[] = {
-    {MCA_BASE_VAR_TYPE_STRING, &ompi_mtl_psm2.psm2_devices, {.stringval = "self,shm,hfi"}, "PSM2_DEVICES", OPAL_INFO_LVL_3,
-     "devices",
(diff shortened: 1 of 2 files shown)


jsquyres:
@matcabral Is this a cherry pick from master?  If it's not, can you explain why?  Thanks!

hjelmn:
I thought we had agreed on leaving the variables but not setting the environment variable unless the user sets the parameter. I haven't looked at this in awhile though.

matcabral:
@jsquyres  this is NOT a cherry-pick from master. 
@hjelmn  Summarizing the discussion in #4574, we agreed to leave this on master until fully supported by PMIx. Therefore, I have been removing it from release branches. With that said, I see that https://github.com/open-mpi/ompi/blob/master/opal/mca/pmix/pmix3x/pmix/src/mca/pnet/opa/pnet_opa_component.c#L79  is apparently already is supporting this functionality. However, I have not been able to test it and confirm it is working (@rhc54 is out this week) .Therefore, I’m filing this PR as a way to avoid the shadow variables in 4.0.0 RC1. So, the question now would be if it would be better to wait and confirm the PMIx functionality, skipping 4.0.0 RC1. However, would definitely like this in the final release. 

Thanks,

rhc54:
It requires that you set the ```orte_enable_instant_on``` MCA param to ```1``` or some non-zero value. I added that requirement solely for initial testing purposes - I would recommend removing that param for a production release as it isn't desirable/necessary. We already protect against the feature not being supported by the active opal/pmix component and so it isn't really needed.  I'll remove it in master and @matcabral can include the commit here.

There are also a couple of other issues to consider here:

- we still have no way of telling the user what PMIx components built and are available. This becomes a significant issue when features like this are being supported as users have no easy way to tell if they configured things correctly.
- we don't expose the PMIx configure options to the OMPI configure level, which makes it harder for users to know how to point to non-standard locations.

Perhaps these should be discussed on the next telecon?

gpaulsen:
lets leave this commit out of v4.0.0rc1 which we're building today, and discuss this on tomorrow's web-ex.

matcabral:
As agreed on 09/11/18 conf call, we will remove this variables from v4.0.x  in this PR and postpone the removal from master until the PMIx solution is complete (some configure time details pending)

PR #5651, opened 2018-09-06
btl/openib: don't complain about no NICs

jsquyres:
Since openib is on its long, slow way out the door, don't let it
complain about not being able to find any NICs at run time.


jsquyres:
Does anyone care if I change this behavior on master and v4.0.0? 

Currently, if the openib BTL is available in the installation but isn't used at `mpirun` time (i.e., it finds 0 IBV devices), it emits a warning message.

This PR changes that behavior to *not* emit the warning.

PR #5652, opened 2018-09-06
mtl/ofi: ask for Libfabric v1.5 API first

jsquyres:
First ask for Libfabric API v1.5.  If that fails, then ask for v1.0,
but with the domain hint FI_MR_SCALABLE (this hint is automatic
starting with API v1.5).

-----

I sent an email to @hppritcha earlier today about this, but I needed to shake up my git tree, so I figured I'd just go ahead and make a PR about this and get feedback this way.

I notice that the OFI MTL assumes FI_MR_SCALABLE.  According to https://ofiwg.github.io/libfabric/v1.6.1/man/fi_mr.3.html:

> (For library versions 1.4 and earlier, this is indicated by setting mr_mode to FI_MR_SCALABLE, with the fi_info mode bit FI_LOCAL_MR set to 0). For versions 1.5 and later, scalable is implied by the lack of any mr_mode bits being set.

The OFI MTL asks for FI_VERSION(1,0) to fi_getinfo(), but then the OFI MTL doesn't specify FI_MR_SCALABLE in the domain hints.

Is this a bug?

Shouldn't it set FI_VERSION(1,5) and if that doesn't work, then try again with FI_VERSION(1,0) and set FI_MR_SCALABLE in the domain hints?

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -512,8 +512,11 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
      * FI_VERSION provides binary backward and forward compatibility support
      * Specify the version of OFI is coded to, the provider will select struct
      * layouts that are compatible with this version.
+     *
+     * We start by asking for API version 1.5 -- that has
+     * FI_MR_SCALABLE implied by default.
      */
-    fi_version = FI_VERSION(1, 0);
+    fi_version = FI_VERSION(1, 5);
 
     /**
      * fi_getinfo:  returns information about fabric  services for reaching a
@@ -527,9 +530,25 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                      hints,         /* In: Hints to filter providers            */
                      &providers);   /* Out: List of matching providers          */
     if (FI_ENODATA == -ret) {
+        // If we got nothing back, try again with FI_VERSION(1, 0) and
+        // explicitly set the FI_MR_SCALABLE flag.
+        fi_version = FI_VERSION(1, 0);
+        hints->domain_attr->mr_mode = FI_MR_SCALABLE;
+
+        ret = fi_getinfo(fi_version, /* OFI version requested                */
+                         NULL,       /* Optional name or fabric to resolve   */
+                         NULL,       /* Optional service name or port to req */
+                         0ULL,       /* Optional flag                        */
+                         hints,      /* In: Hints to filter providers        */
+                         &providers);/* Out: List of matching providers      */
+
         // It is not an error if no information is returned.
-        goto error;
-    } else if (0 != ret) {
+        if (FI_ENODATA == -ret) {
+            goto error;
+        }
+    }
+
+    if (0 != ret) {
         opal_show_help("help-mtl-ofi.txt", "OFI call fail", true,
                        "fi_getinfo",
                        ompi_process_info.nodename, __FILE__, __LINE__,


matcabral:
Hi @jsquyres, the OFI MTl only uses the fi_tagged(3) APIs. Therefore, there is no need for fi_mr(3) stuff. So far, we have not found any reason to be more restrictive than v1.0.  On the other hand, the OFI BTL does look at this, in fact asks for V1.5: 
https://github.com/open-mpi/ompi/blob/master/opal/mca/btl/ofi/btl_ofi_component.c#L217

jsquyres:
@matcabral I'm confused: Using fi_tagged(3) doesn't inherently mean you need to register or not.  Don't you have to use `FI_MR_SCALABLE` (either implicitly or explicitly)?  Or is it since you didn't specify `FI_MR_LOCAL` in the mode bits that you only get providers back that do not require registration?

matcabral:
All of the above :). The pre and post 1.5 thing makes this quite confusing. So, without setting anything in the MR realm we are taking the defaults implicitly. As you say, pre 1.5, by not setting `FI_MR_LOCAL` all providers requiring registration will be discarded, and after 1.5 I would expect it will implicitly default to `FI_MR_SCALABLE`.  So, I think the change that would probably add some clarity would be bringing the pre 1.5 behavior to the explicit plane:
`
libfabric_api = fi_version();
  if (libfabric_api < FI_VERSION(1, 5)) {
 hints->domain_attr->mr_mode = FI_MR_SCALABLE;
}

makes sense?

hppritcha:
This is going to break using GNI provider without some additional smarts.

jsquyres:
So is `FI_MR_SCALABLE` a red herring here?  I.e., is the real desired feature the lack of `FI_MR_LOCAL` in the hints so that you only get providers who do not require registration?  If that's the case, then this PR is moot / not necessary (and the whole pre-1.5 / post-1.5 issue is also moot).

More specifically: does the OFI MTL require OFI providers to not require memory registration?

-----

@hppritcha also told me in an email that additional updates would be required to update the OFI MTL to adhere to the libfabric v1.5 API, so that probably pretty much kills this PR.  But let me ask an additional question before closing this PR anyway...

How do the PSM libfabric providers handle the usual horror story of:

- user calls mmap() to get some new memory at virtual address 0x1000
- user calls MPI_SEND with buffer=0x1000
- OFI MTL sends the buffer
- Because MR_SCALABLE and/or lack of MR_LOCAL, OFI provider registers the memory behind the scenes
- MPI_SEND completes
- user calls munmap(0x1000)
- user calls mmap() again at virtual address 0x1000
- user calls MPI_SEND with buffer=0x1000
- ...what happens next?

@hppritcha tells me that the GNI provider has a hook to Cray's kernel MMU notifier system (based on [Roland's original ummunotify patch](https://marc.info/?l=linux-mm-commits&m=127352503417787&w=2) from several years ago).

What does PSM / PSM2 do?

j-xiong:
Before 1.5, there is no FI_MR_LOCAL, but a mode bit FI_LOCAL_MR. Applications that asks for API 1.0 and want memory registration support should add FI_LOCAL_MR to the mode bits (not mr mode) regardless of the mr mode setting (FI_MR_BASIC or FI_MR_SCALABLE).

PSM/PSM2 doesn't require local memory registration so the "horror" story will just work as usual.

jsquyres:
Ok, I guess I was looking at the most recent libfabric man pages (e.g., that's where it says SCALABLE does not require memory registration).

How does PSM/PSM2 get away without requiring memory registration (even implicit memory registration / under the covers)?  Does it always copy?

-----

Regardless, it sounds like the OFI MTL:

* Requires the use of OFI API 1.0
* Requires providers to not ask for FI_LOCAL_MR (i.e., not utilize explicit memory registration)

j-xiong:
For PSM/PSM2 there are implicit registration inside the driver. But nothing is exposed at the API level.

jsquyres:
Closing this PR as moot.

PR #5653, opened 2018-09-07
Allow to override build user and host

bmwiedemann:
Allow to override build user and host, using the standard $USER and $HOSTNAME environment variables to make reproducible builds possible.
See https://reproducible-builds.org/ for why this is good.

This helps improve issue #3759


jsquyres:
ok to test

jsquyres:
Thanks for bringing this to our attention again; I admit it kinda fell off our radar.  😦 

Are you looking for both this PR *and* #3779 to be merged?  (it looks like that one kinda fell off the radar, too...)

Also, per the discussion in #3759, are you interested in an output line in `ompi_info`, something like "Reproducible build: yes|no"?  (`ompi_info` is the tool users use to see information about the Open MPI installation)

bmwiedemann:
I'd say, this change perfectly complements #3779

We could use `$SOURCE_DATE_EPOCH` as an indicator for "Reproducible build: yes"
but I'd prefer if we could keep the code simple and use this patch as is
and to compensate, in our .spec file we do something like
```bash
export HOSTNAME=reproducible-SUSE-rpm-build
```
which should show up in ompi_info as well

jsquyres:
I'm good with this PR.  Mind if I add a commit to it with a short README blurb about setting`$USER` and/or `$HOSTNAME` and citing Reproducible Builds?  (what's the best URL to cite for RB's, anyway -- https://reproducible-builds.org/ ?)

bmwiedemann:
oops, missed the other comment. Yes, please add blurb with that URL.

jsquyres:
@bmwiedemann I rebased to the current HEAD (since this commit was super old) and added a 3rd commit with a README blurb.  Look ok?

@ggouaillardet You might want to expand on this README blurb in #3779...?

bmwiedemann:
LGTM

bmwiedemann:
Can you merge this PR, please?

jsquyres:
Thanks for your patience -- merged!

Issue #5447, opened 2018-07-17
v4.0.0: Check for ABI compatibility with v3.1.x


jsquyres:
Per https://www.mail-archive.com/ompi-packagers@lists.open-mpi.org/msg00015.html and webex discussion on 2018-07-17, @hppritcha and @gpaulsen will be conducting an ABI compatibility check between the v4.0.x branch (which will be created tomorrow) and the v3.1.x branch.

If possible, we would like to not change libmpi/liboshmem's `.so` major version for all the reasons cited in https://www.mail-archive.com/ompi-packagers@lists.open-mpi.org/msg00015.html.  We're pretty sure that this means increasing **both** `c` **and** `a` by 10.

Howard and Geoff will need to check the branches manually, and probably do a bunch of testing, such as:
1. Compile and install Open MPI v3.1.x into `/path`
1. Compile a bunch of tests against that Open MPI installation
1. `rm -rf /path`
1. Compile and install Open MPI v4.0.x (with appropriate `c:r:a` values and possibly `--enable-mpi1-compat`) into `/path`
1. Run the tests (without recompiling/relinking) and make sure they all work properly

jsquyres:
FYI: @amckinstry

gpaulsen:
I have completed this testing.  It is important to acknowledge that testing alone can never fully guarantee ABI compatibility.  This tested was done in combination with discussion in the weekly web-ex discussions with our core developers and deemed the gating factor if we can keep our .so versions compatible so as to NOT require users' to rebuild/link their mpi applications.

I built and ran the open-mpi _ibm_ test suite (generally considered good coverage of the MPI API interface) with top of v3.1.x branch.  I then deleted that OpenMPI v3.1.x install, and built Open MPI v4.0.x two time (I repeated this process both with and without the new configure flag: `-enable-mpi1-compatibility`).  I then updated the variables:
```
OPAL_PREFIX
OPAL_LIBDIR
PATH
LD_LIBRARY_PATH
```
I was then able to RUN the binaries under each subdirectory of the ibm test suite.
A few tests failed both test runs due to not enough ranks (I used --np 4, along with vader btl).  I also killed the int_overflow test in collectives subdir as that test is a resource HOG!.  These failures ensured that I corectly saw what a failure looked like.  I then checked return codes for all of the other tests, and they all correctly returned 0 return codes.

*PASSED*.

[...]

PR #5657, opened 2018-09-07
Updating VERSION for shared libs for v4.0.0

gpaulsen:
This was done after discussions with core developers about any
  potential ABI breakage for any of the libs the user directly
  links against.  Also compaitiblity tests were done using the
  ibm test suite and building with v3.1.x and running with v4.0.x
  see: https://github.com/open-mpi/ompi/issues/5447


gpaulsen:
Not that this was NOT cherry-picked from master, as it's a release branch only PR.

PR #5665, opened 2018-09-10
v4.0.x: patcher/base: improve instruction cache flush for aarch64

### LINKED ISSUE #5631: flush_and_invalidate_cache is not valid for ARM64 platform

hjelmn:
This commit updates the patcher component to either use the
__clear_cache intrinsic or the correct assembly to flush the
instruction cache.

Fixes #5631

(cherry picked from commit 1cdbceb0951c30a04b730b78f31d07543d8c3d2a)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -858,7 +858,7 @@ OPAL_SEARCH_LIBS_CORE([ceil], [m])
 # -lrt might be needed for clock_gettime
 OPAL_SEARCH_LIBS_CORE([clock_gettime], [rt])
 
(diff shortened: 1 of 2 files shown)


hppritcha:
@shamisp please review if you want this in the 4.0.0 release - which will happen in the next week.

shamisp:
👍

PR #5666, opened 2018-09-11
UCX: added missing UCX libs to UCX detection

hoopoepg:


gpaulsen:
Is this desired to go into v4.0.x or earlier?

hoopoepg:
issue is actual to all release branches
will push fix to branches 3.0+

hppritcha:
@gpaulsen we need this fix.  without it UCX is never found now on any of the systems I've tried.  Not sure how this escaped MTT testing.

PR #5669, opened 2018-09-11
misc: compiler warning fixes

jsquyres:
A variety of small compiler warning fixes.  The 2 PMIx fixes are
already committed upstream.

### DIFF
--- a/ompi/debuggers/ompi_msgq_dll.c
+++ b/ompi/debuggers/ompi_msgq_dll.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2007-2016 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2007-2018 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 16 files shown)


jsquyres:
@bosilca Could you verify the "ompi/monitoring" commit on this PR?  It's mainly code shuffling, but I changed some #define's so that #includes work properly.

jsquyres:
CI correctly points out the ompi/monitoring commit does not correctly handle the no-Fortran build case.  Need to fix that...

jsquyres:
@bosilca The more I look at the monitoring commit, the more it looks like `monitoring_prof_mpi_init_f2c()` and `monitoring_prof_mpi_finalize_f2c()` don't do anything different than the real Open MPI bindings do.

What is their purpose -- should we just remove them?

bosilca:
@jsquyres please rollback your modifications on these 2 files. They define an interposition layer (using the standard MPI approach) that can be LD_PRELOAD by applications to gain automatic monitoring capabilities.

jsquyres:
@bosilca I'm not done with this PR yet, but check out the common/monitoring commit now -- I didn't separate into 2 files, but I did fix the #includes.

jsquyres:
@bosilca This PR is now good-to-go.

Issue #5672, opened 2018-09-11
Double free in oob/ud


jsquyres:
A user reports in https://www.mail-archive.com/users@lists.open-mpi.org/msg32650.html that they're getting a double free in `orted`.  Looking at the stack trace, it's in oob/ud:

```
libibverbs: GRH is mandatory For RoCE address handle
*** Error in `/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/bin/orted': double free or corruption (out): 0x00002aaab4001680 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81499)[0x2aaaabdfd499]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/openmpi/mca_oob_ud.so(mca_oob_ud_peer_from_uri+0xb9)[0x2aaaac9b1f39]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/openmpi/mca_oob_ud.so(+0x5fd8)[0x2aaaac9affd8]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/openmpi/mca_oob_ud.so(+0x4b2f)[0x2aaaac9aeb2f]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/libopen-rte.so.40(orte_oob_base_send_nb+0x1131)[0x2aaaaab32c21]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/libopen-pal.so.40(opal_libevent2022_event_base_loop+0xd89)[0x2aaaaac00459]
/apps/brussel/CO7/ivybridge-ib/software/OpenMPI/3.1.1-GCC-7.3.0-2.30/lib/libopen-pal.so.40(+0x369ae)[0x2aaaaabbf9ae]
/lib64/libpthread.so.0(+0x7e25)[0x2aaaabb67e25]
/lib64/libc.so.6(clone+0x6d)[0x2aaaabe7abad]
```

[...]

PR #5673, opened 2018-09-11
v2.1.x: oob/ud: remove double free

jsquyres:
And just for the hecuvit, add in a clarifing comment and a unique
return code that more precisely defines the error that occurred.

Note that this is not a cherry pick from master because oob/ud has already been removed from master.

Refs #5672


jsquyres:
@hppritcha Good to go.

Issue #5643, opened 2018-09-04
Question about correct file format in external32


adammoody:
## Background information

My understanding is that external32 should write integer values in big-endian format, but it seems to be writing in little-endian format instead.  I will attach a test case that writes a series of MPI_UINT64_T values to two files: one in "native" data representation and one in "external32".  Running hexdump on the resulting file shows that both appear to store those values in little-endian representation.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

Open MPI 3.0.1

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Installed from v3.0.1 distribution tarball.

### Please describe the system on which you are running

* Operating system/version:  RHEL 7.5
* Computer hardware: Intel(R) Xeon(R) CPU E5-2695 v4
* Network type: Intel Omni-path

-----------------------------

## Details of the problem

I am trying to debug some data format issues with a different MPI library, which got me testing the difference between “native” and “external32” data representations.  I wrote a program that writes 4 consecutive MPI_UINT64_T values to a file in both formats.  According to the standard external32 should be using big-endian in the file, but it looks to me like it’s using little-endian instead.  This is with Open MPI 3.0.1 on an machine with Intel processors.  The sequence of MPI_UINT64_T values the test case writes is {1, 2, 3, 4}.

```shell
>>: hexdump -C openmpi/testfile_ext32
00000000  01 00 00 00 00 00 00 00  02 00 00 00 00 00 00 00  |................|
00000010  03 00 00 00 00 00 00 00  04 00 00 00 00 00 00 00  |................|
 ```

This looks to be little-endian format.

**Comment by adammoody** [comment]
Reproducer builds with mpicc, run as a single process (will fail with more than one).

```
#include <stdlib.h>
#include <stdio.h>
#include <inttypes.h>
#include <string.h>

#include "mpi.h"

static void create_stattype(int chars, MPI_Datatype* dt_stat)
{
    int size = chars + sizeof(uint64_t);

    MPI_Datatype dt_filepath;
    MPI_Type_contiguous(chars, MPI_CHAR, &dt_filepath);

    MPI_Datatype types[2];
    types[0] = dt_filepath;  /* file name */
    types[1] = MPI_UINT64_T; /* file type */

    int blocklens[2] = {1, 1};
    MPI_Aint displs[2];
    displs[0] = 0;
    displs[1] = chars;

[...]

PR #5680, opened 2018-09-12
coll/libnbc: fix NBC_Unpack()

ggouaillardet:
always initialize 'size'

### DIFF
--- a/ompi/mca/coll/libnbc/nbc_internal.h
+++ b/ompi/mca/coll/libnbc/nbc_internal.h
@@ -516,6 +516,11 @@ static inline int NBC_Unpack(void *src, int srccount, MPI_Datatype srctype, void
   int res;
   ptrdiff_t ext, lb;
 
+  res = ompi_datatype_pack_external_size("external32", srccount, srctype, &size);
+  if (OMPI_SUCCESS != res) {
+    NBC_Error ("MPI Error in ompi_datatype_pack_external_size() (%i)", res);
+    return res;
+  }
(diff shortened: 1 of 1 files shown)


jsquyres:
bot:ompi:retest

jsquyres:
@ggouaillardet As you pointed out in https://github.com/open-mpi/ompi/issues/5643#issuecomment-420557817, it might be worth putting in the commit message that this is a real bug, but it's in an algorithm that was implemented but not used (i.e., it's not a NEWS-worthy comment, nor anything for users to worry about -- and not even worth PR'ing to any of the current release branches).

jsquyres:
OS X CI keeps failing: `ERROR: Error fetching remote repo 'origin'`

jsquyres:
Confirmed: OS X CI "piper" is having networking problems at the moment.  Ok to merge this PR since it was the only failure.

bwbarrett:
@jsquyres what?  How do you know this doesn't break OS X?  seems much better idea to retest.

jsquyres:
Ok, sorry -- do we have an ETA when OS X CI will be back?

rhc54:
@bwbarrett I understand your concern, but you can't ask everything to hit a dead stop because the OSX tester is hung/down with no prognosis for recovery. Some of us are quite capable of building on OSX and can see that nothing was broken - no sense in holding everything up in such cases.

ggouaillardet:
@jsquyres per your comment, I updated the commit message.

hppritcha:
bot:ompi:retest

PR #5689, opened 2018-09-12
Silence warning of unused function

rhc54:
Requires protection for HAVE___CLEAR_CACHE


rhc54:
Stuck in OSX CI - but this was built and tested in OSX, so proceeding.

PR #5700, opened 2018-09-14
Handle error cases in TCP BTL

abouteiller:
When an error is returned by the socket operations, trigger the
appropriate error path in the PML to give an opportunity for
rerouting/error handling.


bwbarrett:
bot:ompi:retest

OMPI testers ran out of disk space...

PR #5701, opened 2018-09-14
Adding error handling in OpenIB BTL

abouteiller:
This improves the handling of some error cases in the Open IB BTL.
1. Mark the IB endpoint as failed when invoking an error (this resolves an UDCM connexion deadlock in error cases).
2. Return send-credits spent on fragments pending on a failed endpoint (resolves a resource exhaustion deadlock in error cases).
3. Do not trigger an error for simple out-of-credit events.

### DIFF
--- a/opal/mca/btl/openib/btl_openib_component.c
+++ b/opal/mca/btl/openib/btl_openib_component.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2007 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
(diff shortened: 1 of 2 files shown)


bwbarrett:
bot:ompi:retest

Looks like we ran out of disk space for some odd reason.

hppritcha:
@abouteiller you know we only support RoCE and iwarp now for the openib BTL and that its going to removed most likely in the next major release if we find that UCX and libfabric both support RocE and iwarp well enough?

abouteiller:
Yes, I am pushing out already completed work here. I'll make a pass on UCX error behavior next.

PR #5710, opened 2018-09-14
v3.0.x.: io/romio314: fix minmax datatypes

jsquyres:
romio assumes that all predefined datatypes are contiguous. Because of
the (terribly named) composed datatypes MPI_SHORT_INT, MPI_DOUBLE_INT,
MPI_LONG_INT, etc this is an incorrect assumption. The simplest way to
fix this is to override the MPI_Type_get_envelope and
MPI_Type_get_contents calls with calls that will work on these
datatypes. Note that not all calls to these MPI functions are
replaced, only the ones used when flattening a non-contiguous
datatype.

References #5009

(cherry picked from commit 4d876ec6fedd2f5d299259eff53ba37b12c5d7e0)


bwbarrett:
bot:ompi:retest

Host ran out of disk space

PR #5717, opened 2018-09-16
shmem/lock: progress communications while waiting for shmem_lock

yosefe:


gpaulsen:
In the future, for PRs like this, it would be nice to have a link to an issue with more details such as:

- How the issue was discovered
- Under what circumstances users might hit the bug.
- Does this result in a hang or just poor performance?
- What components need to be loaded to hit the issue etc.

The reason this is important for PRs like this, is because additional calls to opal_progress can have wide reaching performance issues, and if any performance regression is traced back to this PR, we'd need to understand if this a bit better.

Thanks.

PR #5718, opened 2018-09-17
coll/libnbc: add recursive doubling algorithm for MPI_Iexscan

mkurnosov:
Implements recursive doubling algorithm for MPI_Iexscan. The algorithm preserves order of operations so it can be used both by commutative and non-commutative operations.

The MCA parameter `coll_libnbc_iexscan_algorithm` was added for dynamic algorithm selection. 


mkurnosov:
I think that more discussion on dynamic algorithm selection in the libnbc is needed. For example: use approach from the coll/tuned (dynamic decisions) in the libnbc or add non-blocking collectives into the coll/tuned.

PR #5741, opened 2018-09-20
v4.0.x: fortran/use-mpi-f08: clean [p]ompi_FOO_f bindings

ggouaillardet:


hppritcha:
@gpaulsen I think we should pull this in.

gpaulsen:
I noticed that this file was deleted in this PR, is that intentional?
`ompi/mpi/fortran/use-mpi-f08/pmpi-f-interfaces-bind.h`

ggouaillardet:
@gpaulsen this is intentional. The rationale is we do not need the bindings for the internal `pompi_*` subroutines

hppritcha:
@jsquyres please review when you have some cycles

Issue #3972, opened 2017-07-27
CUDA GPUDirect blocks on message size greater than GPUDirect RDMA limit with openib btl


Felandric:
## Background information
When using CUDA GPUDirect to Send and Recv directly from GPU buffers on message sizes greater than the RDMA limit, rather than expected behavior of staging buffers through host memory, Open MPI simply hangs. 

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
v3.0.x

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
Git clone.
Commands used:
```
./autogen.pl

./configure --prefix=/labhome/ompi/install --with-knem=/path/to/knem --with-mxm=/path/to/mxm --with-slurm --with-pmi --with-platform=contrib/platform/mellanox/optimized --with-hcoll=/path/to/hcoll --with-ucx=/path/to/ucx --with-cuda

make all install
```

### Please describe the system on which you are running

* Operating system/version: Red Hat Enterprise Linux Server 7.2 (Maipo) (3.10.0-327.el7.x86_64)
CUDA distribution: 8.0
CUDA GPUDirect Driver: Mellanox OFED GPUDirect RDMA
* Computer hardware: 2 nodes, 48 cores/node, 2 GPUs per node
CPUs: Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz
GPUs: NVIDIA Tesla P100-PCIE-16GB

* Network type: 
Infiniband
-----------------------------

## Details of the problem
GPUDirect is intended to improve latency and bandwidth by allowing RDMA transfers from GPU to GPU and bypassing the CPU. However, there is a message size limit on RDMA transfers. Above this limit, GPUDirect is expected to stage buffers through host memory with cudaMemcpy. 
The limit can be changed with the parameter "-mca btl_openib_cuda_rdma_limit x", where x is the message size in bytes.
Unfortunately, rather than behave as expected, the program simply hangs on MPI_Recv.  
According to [this](https://www.research-collection.ethz.ch/bitstream/handle/20.500.11850/155349/eth-48403-01.pdf), this may occur due to OpenMPI using blocking copies. The option "-mca mpi_common_cuda_cumemcpy_async 1" is meant to instruct OpenMPI to use non-blocking asynchronous copies when staging buffers through host.  
[...]

PR #5748, opened 2018-09-20
openib: Disable CUDA async by default

bwbarrett:
Disable async receive for CUDA under OpenIB.  While a performance
optimization, it also causes incorrect results for transfers
larger than the GPUDirect RDMA limit.  This change has been validated
and approved by Akshay.

References #3972

(cherry picked from commit 9344afd485d23c41a18ab57d6df8550826c24b9e)


Akshay-Venkatesh:
@bwbarret would you mind changing the commit message to 'disabling cuda async recv'? Cuda async is broader.

bwbarrett:
@Akshay-Venkatesh kind of ugly to do that for just the branch commits :/.

Akshay-Venkatesh:
Should be ok then.

PR #5526, opened 2018-08-08
Fix script abstraction break: mv make_manpage.pl to config

jsquyres:
Having the "make_manpage.pl" script in the ompi/ tree broke
"./autogen.pl --no-ompi" (specifically: "make distcheck" of --no-ompi
builds would break).

Signed-off-by: Jeff Squyres <jsquyres@cisco.com>

PR #5753, opened 2018-09-21
Fix script abstraction break: mv make_manpage.pl to config

gpaulsen:
v4.0.x version of PR #5526 

Having the "make_manpage.pl" script in the ompi/ tree broke
"./autogen.pl --no-ompi" (specifically: "make distcheck" of --no-ompi
builds would break).


jsquyres:
@gpaulsen This commit does not have the usual cherry pick line in the commit message.  Is that a mistake?

gpaulsen:
Yes, that's a mistake.  how come the CI didn't catch that?  Don't we have CI for that?

gpaulsen:
Ok, repushed with correct cherry-picked from message in PR comment.

PR #5759, opened 2018-09-22
Fix/monitoring

bosilca:
The monitoring PML was unloaded too early, and all its PVARs were marked as invalid, totally defeating the purpose of the PML. Reshape the highjacking order to make it simpler and more robust. There is however a caveat, the monitoring module cannot be unloaded from memory (or we will have to return on a memory region that was just dlclosed).

### DIFF
--- a/ompi/mca/pml/base/pml_base_select.c
+++ b/ompi/mca/pml/base/pml_base_select.c
@@ -193,6 +193,14 @@ int mca_pml_base_select(bool enable_progress_threads,
         modex_reqd = true;
     }
 
+    /* Save the winner */
+
+    mca_pml_base_selected_component = *best_component;
+    mca_pml = *best_module;
+    opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
+                         "select: component %s selected",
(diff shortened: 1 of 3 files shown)


jsquyres:
@bosilca Don't we have a refcount mechanism for loaded MCA components?

bosilca:
Yes we do and we use it here to prevent early unloading of the component, but this mechanism does not help with the underlying issue. In general the framework base release the last reference to the component once the component `close` has been called.  In this particular case we remove the component from the base list by intercepting the interface of the selecting component, the only entity that can remove the last refcount is the component itself.

ibm-ompi:

 Gist: https://gist.github.com/411839f6e4e2e2a925480b8f28e3a836

ibm-ompi:

 Gist: https://gist.github.com/644e981107dcd2224441813047914be0

ibm-ompi:

 Gist: https://gist.github.com/340924647b3f16e37aa1497cfec94300

jjhursey:
bot:ibm:retest (CI script broke, should be fixed now)

PR #5771, opened 2018-09-25
v4.0.x: README: Add note about --with-foo and RPATH

jsquyres:
Specifically mention our intended behavior about /usr and /usr/lib
(and why we don't add /usr/lib[64] and /usr/local/lib[64] to RPATH).

(cherry picked from commit 9367440e3210cf2bfae33d3c72411ab8b9fa6622)

### DIFF
--- a/README
+++ b/README
@@ -769,17 +769,21 @@ Additionally, if a search directory is specified in the form
 --with-<foo>=<dir>, Open MPI will:
 
 1. Search for <foo>'s header files in <dir>/include.
-2. Search for <foo>'s library files in <dir>/lib, and if they are not
-   found there, search again in <dir>/lib64.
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -765,6 +765,22 @@ Open MPI is unable to find relevant support for <foo>, configure will
 assume that it was unable to provide a feature that was specifically
 requested and will abort so that a human can resolve out the issue.
 
+Additionally, if a search directory is specified in the form
+--with-<foo>=<dir>, Open MPI will:
```
rhc54: Given the user that reported this was using ```--with-foo-libdir```, you probably should include that here as well

jsquyres: Will add #5773 when CI completes.

PR #5772, opened 2018-09-25
coll/libnbc: add knomial tree algorithm for MPI_Ibcast

mkurnosov:
Implements k-nomial tree algorithm for `MPI_Ibcast`.


mkurnosov:
bot:mellanox:retest

mkurnosov:
bot:retest

ggouaillardet:
@mkurnosov thanks for this PR !

@bosilca I noted it defines some new MCA parameters in order to select the algo used by `coll/libnbc`.
From a user point of view, this is more or less similar to how `coll/tuned` can be used.

should we take this opportunity to dispatch `coll/libnbc` into `coll/base` and `coll/tuned` ?
That would make things even simpler for the user (same MCA prefix, ability to use a file for algo selection instead of simply forcing an algo regardless communicator and/or data size) ?

If not, should we at least fully mimick `coll/tuned` (use a file, split static vs dynamic algo selection, ...) ?

This is something I am willing to work on if this is the way we should move forward.

bosilca:
We are in the process of replacing tuned with better collective support, but the selection logic from tuned could be migrated to allow users to select their algorithm. I'll be happy to help to extract it as an independent entity.

mkurnosov:
@ggouaillardet  @bosilca I plan to add other collective algorithms to the libnbc. I can use current approach with MCA parameter `coll_libnbc_ibcast_*_algorithm` as a temporary solution. What do you think?

bosilca:
At long as you keep a consistent approach with tuned we're all good.

PR #5775, opened 2018-09-25
pmix: check the old topo key to keep compatibility with old RMs

### LINKED ISSUE #5177: PMIx: backward compatibility with previous RMs
I noticed this in v3.1.x: In the hwloc topo get function a new PMIx variable is used:
https://github.com/open-mpi/ompi/blob/5d920bcc1a268c22fd1a5770ee29d294f6af7cd9/opal/mca/hwloc/base/hwloc_base_util.c#L347

This change went unnoticed by us and Slurm plugin wasn't updated. So Slurm plugin provides the old key named `PMIX_LOCAL_TOPO` instead of `PMIX_HWLOC_XML_V1` and I thought that it would be correct to test this key as well.

karasevb:

Closes #5177

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -350,6 +350,11 @@ int opal_hwloc_base_get_topology(void)
         OPAL_MODEX_RECV_VALUE_IMMEDIATE(rc, OPAL_PMIX_HWLOC_XML_V1,
                                         &wildcard_rank, &val, OPAL_STRING);
 #endif
+        if (rc != OPAL_SUCCESS) {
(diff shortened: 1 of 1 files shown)


rhc54:
I think this is okay - however, I'm not sure what will happen if OMPI is built against a different version of HWLOC than the RM. Have you checked what happens if Slurm is built against HWLOC v1 and OMPI is using v2 (and vice versa)?

I'm not sure if HWLOC will gracefully return an error, crash, or just construct garbage. @bgoglin ?

karasevb:
@rhc54 
I checked both cases which you mentioned (Slurm+HWLOC v1 vs OMPI + HWLOC v2, and vice versa), this did not lead to failures.

PR #5777, opened 2018-09-25
Expose PMIx tools

rhc54:
Provide users with a mechanism for seeing how the embedded PMIx was
built and what PMIx MCA params are supported. Fix a strange issue with
how the PMIX_CONFIGURE_CLI value gets passed to pmix_info.


ibm-ompi:

 Gist: https://gist.github.com/a8fa7507dbdda7770bd98ef53096bc94

ibm-ompi:

rhc54:
I'm not aware of us coming up with any suggestions on a solution to the disable-dlopen problem here. Unless someone has one, I propose to (sadly) abandon the attempt. The only option I could think of would be to just not expose the tools if disable-dlopen is given, but I don't know if that is acceptable.

rhc54:
Abandoning as "don't know how to do it"

PR #5779, opened 2018-09-26
Squash a bunch of harmless compiler warnings.

jsquyres:

### DIFF
--- a/orte/mca/rmaps/base/rmaps_base_map_job.c
+++ b/orte/mca/rmaps/base/rmaps_base_map_job.c
@@ -556,17 +556,17 @@ void orte_rmaps_base_display_map(orte_job_t *jdata)
                 if (NULL == (proc = (orte_proc_t*)opal_pointer_array_get_item(node->procs, j))) {
                     continue;
                 }
-                memset(tmp1, 0, 1024);
+                memset(tmp1, 0, sizeof(tmp1));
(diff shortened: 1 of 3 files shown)


orte/mca/rmaps/base/rmaps_base_map_job.c
```diff
@@ -556,17 +556,17 @@ void orte_rmaps_base_display_map(orte_job_t *jdata)
                 if (NULL == (proc = (orte_proc_t*)opal_pointer_array_get_item(node->procs, j))) {
                     continue;
                 }
-                memset(tmp1, 0, 1024);
+                memset(tmp1, 0, sizeof(tmp1));
                 if (orte_get_attribute(&proc->attributes, ORTE_PROC_HWLOC_BOUND, (void**)&bd, OPAL_PTR)) {
                     if (NULL == bd) {
-                        (void)strncpy(tmp1, "UNBOUND", strlen("UNBOUND"));
+                        (void)strncpy(tmp1, "UNBOUND", sizeof(tmp1));
```
bwbarrett: I know this PR is closed, but you're asking me to review the PRs into release branches.  This patch isn't really any better.  strncpy() is not guaranteed to NULL terminate the string, so this can still cause memory corruption down stream.  Better would be to explicitly set `tmp1[sizeof(tmp1) - 1] = '\0';` after the strncpy() to make sure the string is NULL terminated.

PR #5780, opened 2018-09-26
v4.0.x: Fortran 08 bindings fixes

jsquyres:
Refs #5442 (this is the v4.0.x version of #5443).

**NOTE:** This PR does *not* change the `mpi_f08` ABI as far as I can tell.  I tried with and without this PR using gcc 8.2 and intel 2018, and the symbols were always the same.  This is not definitive proof, but I think `ASYNCHRONOUS` and `INTENT` do not change symbols.

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/allgather_f08.F90
+++ b/ompi/mpi/fortran/use-mpi-f08/allgather_f08.F90
@@ -13,7 +13,8 @@ subroutine MPI_Allgather_f08(sendbuf,sendcount,sendtype,recvbuf,recvcount,recvty
    use :: mpi_f08_types, only : MPI_Datatype, MPI_Comm
    use :: ompi_mpifh_bindings, only : ompi_allgather_f
    implicit none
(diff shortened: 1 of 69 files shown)


jsquyres:
Per https://github.com/open-mpi/ompi/pull/5453#issuecomment-425088103, let's wait a day or so.

gpaulsen:
@ggouaillardet ?

jsquyres:
@gpaulsen We were waiting for you to hear back from your Fortran experts...?  (see the comment that I cited, where @gpaulsen said he was still checking with IBM Fortran experts)

kawashima-fj:
@jsquyres May I add a commit to this PR? I want to make a PR of #5799 for v4.0.x but must wait until this PR is merged because my commit is based on changes in this PR. If this PR is merged before your answer, I'll make a separate PR.

gpaulsen:
@jsquyres , @kawashima-fj, I was able to reach another IBM XL Fortran Engineer, who confirmed that INTENT does not change the calling convention (which is good since this was merged to master already).

kawashima-fj:
@gpaulsen , @jsquyres Thanks. I created a separate PR #5843.

PR #5781, opened 2018-09-26
v3.1.x: Fortran 08 bindings fixes

jsquyres:
Refs #5442 (this is the v3.1.x version of #5443).

NOTE: This PR does not change the mpi_f08 ABI as far as I can tell. I tried with and without this PR using gcc 8.2 and intel 2018, and the symbols were always the same. This is not definitive proof, but I think ASYNCHRONOUS and INTENT do not change symbols.


jsquyres:
Per https://github.com/open-mpi/ompi/pull/5453#issuecomment-425088103, let's wait a day or so.

jsquyres:
@ggouaillardet Might as well review this one, too (it's the 3.1 version of #5780)

PR #5790, opened 2018-09-27
shmem/lock: progress communications while waiting for shmem_lock

yosefe:
When the underlying SPML requires active progress on target side (e.g RMA emulation over active messages), need to call opal_progress() to avoid a deadlock.
For example, when UCX is using TCP transport.

### DIFF
--- a/oshmem/shmem/c/shmem_lock.c
+++ b/oshmem/shmem/c/shmem_lock.c
@@ -708,6 +708,7 @@ static int shmem_lock_wait_for_ticket(void *lock,
 
     do {
         shmem_int_get(&remote_turn, lock_turn, 1, server_pe);
+        opal_progress();
     } while (remote_turn != ticket);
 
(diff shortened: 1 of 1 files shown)


gpaulsen:
As I said on the linked master PR:

> In the future, for PRs like this, it would be nice to have a link to an issue with more details such as:
> 
> - How the issue was discovered
> - Under what circumstances users might hit the bug.
> - Does this result in a hang or just poor performance?
> - What components need to be loaded to hit the issue etc.
> 
> The reason this is important for PRs like this, is because additional calls to opal_progress can have wide reaching performance issues, and if any performance regression is traced back to this PR, we'd need to understand if this a bit better.
> 
> Thanks.
>

gpaulsen:
@hppritcha I'm okay PRing this now, since we don't yet have a good automated way to tell if a particular PR causes a performance regression.

yosefe:
@gpaulsen sorry, missed it, i've updated the PR description here and on master.

PR #5794, opened 2018-09-27
mtl ofi: Change from opt-in to opt-out provider selection

bwbarrett:
Change default provider selection logic for the OFI MTL.  The
old logic was whitelist-only, so any new HPC NIC provider would
have to ask users to do extra work or wait for an OMPI release
to be whitelisted.  The reason for the logic was to avoid
selecting a "generic" provider like sockets or shm that would
frequently have worse performance than the optimized BTL options
Open MPI supports.

With the change, we blacklist the (small, relatively static) list
of providers that duplicate internal capabilities.  Users can use
one of thse blacklisted providers in two ways: first, they can
explicitly request the provider in the include list (which will
override the default exclude list) and second, the can set a new
empty exclude list.

Since most HPC networks require special libraries and therefore
an explicit build of libfabric, it is highly unlikely that this
change will cause users to use libfabric when they didn't want to
do so.  It does, however, solve the whitelisting problem.

(cherry picked from commit c5eaa38491c7197f7dbc74c299ade18e09bf5f64)

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -124,7 +124,7 @@ ompi_mtl_ofi_component_register(void)
                                     MCA_BASE_VAR_SCOPE_READONLY,
                                     &param_priority);
 
-    prov_include = "psm,psm2,gni";
+    prov_include = NULL;
     mca_base_component_var_register(&mca_mtl_ofi_component.super.mtl_version,
                                     "provider_include",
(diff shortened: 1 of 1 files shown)


gpaulsen:
@hppritcha any thoughts?  I'm ready to pull this in.

gpaulsen:
Oops, I merged in the NEWS for this before merging in this PR.  I'd like @hppritcha's okay before we merge this, but he's traveling this week.  I'll give him another 6-8 hours and then merge this if I haven't heard from him.

PR #5802, opened 2018-09-28
mpi.h: remove MPI_UB/MPI_LB when not enabling MPI-1 compat

jsquyres:
This is one of the 2 commits from master PR #5800 (the other commit would change ABI).  This one does *not* change ABI: if MPI_UB is used in a user application (and `--enable-mpi1-compatibility` was not specified), it will currently fail to compile with a confusing warning about an undefined `ompi_mpi_ub` symbol in a preprocessor macro.  With this change, it will still fail to compile, but with a much clearer about `MPI_UB` not existing.

----

When --enable-mpi1-compatibility was specified, the ompi_mpi_ub/lb
symbols were #if'ed out of mpi.h.  But the #defines for MPI_UB/LB
still remained.  This commit also #if's out the MPI_UB/LB macros when
--enable-mpi1-compatibility is specified.
    
(cherry picked from commit 7223334d4dc1225d49cd2c63714870c3a04ad953)


gpaulsen:
Jeff said:

>  This commit also #if's out the MPI_UB/LB macros when --enable-mpi1-compatibility is specified.

I think you meant to say:

>  This commit also #if's out the MPI_UB/LB macros _UNLESS_ --enable-mpi1-compatibility is specified.

Correct?

PR #5803, opened 2018-09-28
opal: Remove outdated MacOS workaround

bwbarrett:
Remove the pack/unpack pragma around net/if.h on MacOS, which was added to fix a bug in MacOS X 10.4.x on 64-bit platforms. The bug was fixed in Mac OS X 10.5.0 and, sometime in the last 11 years, compilers started emitting warnings about the fact that the Apple header stomped over the pragma pack settings from the workaround.  We already don't support versions of MacOS earlier than 10.5, so there's no point in keeping the workaround.
    
### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -53,7 +53,7 @@ Copyright (c) 2014-2015 Hewlett-Packard Development Company, LP.  All
                         rights reserved.
 Copyright (c) 2013-2017 Research Organization for Information Science (RIST).
                         All rights reserved.
-Copyright (c) 2017      Amazon.com, Inc. or its affiliates.  All Rights
+Copyright (c) 2017-2018 Amazon.com, Inc. or its affiliates.  All Rights
                         reserved.
 Copyright (c) 2018      DataDirect Networks. All rights reserved.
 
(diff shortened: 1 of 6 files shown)


bwbarrett:
@rhc54 Note that the same warning is being sent from PMIx's pif.h and the same fix will remove the warning.  Of course, you'd also drop support for 10.4.x and earlier if you take this patch.  A bit of configure magic could probably detect 10.4.x and keep the patch, but that seems like a lot of work for an operating system that has zero users.

rhc54:
Got it - will upstream. Thanks!

README
```diff
@@ -143,7 +145,7 @@ General notes
 Platform Notes
 --------------
 
-- N/A
+- Mac OS X versions prior to 10.5.0 are not supported on 64-bit systems.
```
jsquyres: Per this chunk of configure.ac, I think you don't need this README note:

https://github.com/open-mpi/ompi/blob/8b228c1314dc81c799e14b9988b7b0c41d125f8a/configure.ac#L328-L347

rhc54:
Hmmm...I also see this in the ipv4 components. Did you miss them in the OPAL ones?

bwbarrett:
> Hmmm...I also see this in the ipv4 components. Did you miss them in the OPAL ones?

@rhc54, I don't see the pack/unpack workaround in any of the IPv4 files that weren't fixed in this patch.  I see some includes of net/if.h in those files, but they didn't have the #pragma in question around them.  If you see one, happy to fix.

README
bwbarrett: Oops, missed that.  I removed the README change and updated the commit message to match this new reality.

rhc54:
odd - this code was still in my v4 components. I guess somebody removed them in OPAL with a different commit? anyway, they do seem to be gone from OPAL now.

Issue #5801, opened 2018-09-28
Add missing C MPI datatypes to mpi_f08


jsquyres:
Per https://github.com/open-mpi/ompi/pull/5799#pullrequestreview-159873191, we are missing the C MPI datatypes in `mpi_f08` (e.g., MPI_INT, MPI_CHAR, MPI_OFFSET, MPI_COUNT, ...etc.).

This prevents us from building against the MPICH test suite.  Here's how I tried to build the MPICH test suite:

```sh
#!/bin/sh

# First, build and install Open MPI and ensure it is in your path.

git clone git@github.com:pmodels/mpich.git
cd mpich
git submodules init
git submodules update
./autogen.sh
./configure --prefix=$HOME/bogus CC=gcc CXX=g++ FC=gfortran
cd test

# Edit mpi/include/mpitestconf.h to reset anything necessary
# E.g., on my system, mpich decided that INTEGER16 was available
# but Open MPI did not build support for it.  So I had to comment
# out the definition of HAVE_MPI_INTEGER16

make
```

jsquyres:
Removed v3.0.x and v3.1.x targets.

gpaulsen:
https://github.com/open-mpi/ompi/pull/5811 Merged into v4.0.x.  Closing

PR #5811, opened 2018-10-01
fortran/use-mpi-f08: add MPI C types

ggouaillardet:
Refers open-mpi/ompi#5801


ibm-ompi:

 Gist: https://gist.github.com/f17fe24388064c34fcdb730d3d0cd27e

jsquyres:
Got a compile error w.r.t. `OMPI_MPI_C_FLOAT`.

Also, how about also adding these to the `mpi` module and `mpif.h`?

I.e., these constants should be available in all of the Fortran interfaces.

ggouaillardet:
I fixed `use mpi_f08` and will double check `mpif.h` and `use mpi` tomorrow

ggouaillardet:
@jsquyres the `allctypes` tests work for all Fortran bindings, so I think we are good here.

gpaulsen:
Added target v4.0.x as that's also on open-mpi/ompi#5801

ggouaillardet:
@jsquyres there are `allctypes` tests for all Fortran flavors and they all pass, so from my point of view, it means we are good.

jsquyres:
Ok.

PR #5812, opened 2018-10-01
fortran: add CHARACTER and LOGICAL support to MPI_Sizeof()

ggouaillardet:

### DIFF
--- a/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-sizeof.pl
@@ -1,8 +1,8 @@
 #!/usr/bin/env perl
 #
 # Copyright (c) 2014-2015 Cisco Systems, Inc.  All rights reserved.
-# Copyright (c) 2015      Research Organization for Information Science
-#                         and Technology (RIST). All rights reserved.
+# Copyright (c) 2015-2018 Research Organization for Information Science
+#                         and Technology (RIST).  All rights reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@jsquyres can you please double check this PR ?
It is required in order to successfully build the mpich test suite.
I cannot recall if there is a good reason not to have `MPI_Sizeof()` support these types

jsquyres:
bot:ompi:retest

jsquyres:
@ggouaillardet CI now passes.  Can you extend the `ompi-tests/ibm/random/mpi_sizeof_*` tests to check CHARACTER and LOGICAL and make sure they are working properly?  If so, then let's merge this PR and cherry pick it to 3.0.x and beyond (don't bother with 2.1.x).

ggouaillardet:
@jsquyres are you sure you want me to do that __now__ ?

If I take a naive approach, the `ibm` testsuite will not even build with the release branches, which means no test will ever be performed.
A more convoluted approach is to do the test at `configure` time, and if it success, do it again at runtime, or _hardcode_ a failure. That is definitely something I can do, unless you suggest we wait until the fix lands in the release branches before updating the test.

jsquyres:
Meh; yeah, this is effectively a compile-time test (although we want to make sure that the value returned by the subroutine is also correct).  Probably the safest thing to do would be to add a configure test, as you suggest -- just try to compile it and see if it works, and set an AC_DEFINE with the result.

PR #5827, opened 2018-10-02
Remove stale ORTE code

rhc54:
Functionality moved to PMIx

See https://www.mail-archive.com/devel@lists.open-mpi.org/msg20747.html


artpol84:
Thanks!

PR #5840, opened 2018-10-04
mpiext/pcollreq: Correct f08 routine signatures

kawashima-fj:
Changes of nonblocking collectives in e98d794e8b and f750c6932c are applied to persistent collectives. And the module file is updated.

Need PR to v4.0.x branch.

@ggouaillardet FYI


kawashima-fj:
bot:mellanox:retest

PR #5848, opened 2018-10-05
Allow disabling vader fastbox support

bwbarrett:


bwbarrett:
This doesn't disable on any particular platform, just gives us the ability to disable in general at configure time.

opal/mca/btl/vader/btl_vader_fbox.h
```diff
@@ -9,7 +9,7 @@
  * $HEADER$
  */
 
-#if !defined(MCA_BTL_VADER_FBOX_H)
+#if !defined(MCA_BTL_VADER_FBOX_H) && OPAL_BTL_VADER_FBOX_SUPPORT
```
jsquyres: A minor nit: `OPAL_BTL_VADER_FBOX_SUPPORT` won't be defined until `opal_config.h` was previously included.  If `btl_vader_fbox.h` is the first file included, then that macro won't be defined.

I realize that our coding guidelines say "always include `opal_config.h` first!", but sometimes people don't/forget to do that.

Do you want to put an `#include "opal_config.h"` here at the top of `btl_vader_fbox.h` to cover all the bases / be defensive?

bwbarrett: You'll get a compiler error if you include fbox.h before opal_config.h, because we error on undefined preprocessor define behavior.  So I don't think it's worth being over defensive here.

hjelmn:
Just want to point out that this is affecting more than i386 so I need to fix the bug. Since nothing has really changed in the fast boxes for a long time its probably a bug in the "fixes". I wouldn't waste time on this PR.

bwbarrett:
> Just want to point out that this is affecting more than i386 so I need to fix the bug. Since nothing has really changed in the fast boxes for a long time its probably a bug in the "fixes". I wouldn't waste time on this PR.

Why?  Would it not be helpful to have amckinstry run without fboxes to make sure the issue actually is in the fbox code?  Would it not be helpful to have an out so we can ship 4.0, if it comes to that?  Plus, unless I'm missing something, the hard work's already done, so unless there's a reason it is 1) wrong or 2) has some moral objection I'm not missing, it seems like committing and moving on is the right option.

hjelmn:
Nope. This is the fast path for messages. Shipping without it working on x86_64 is a blocker.

hjelmn:
FWIW I have recreated the hang on x86_64 in a docker container. Digging into it now.

bwbarrett:
> Nope. This is the fast path for messages. Shipping without it working on x86_64 is a blocker.

Ok, so we go from "i don't have time to work on i386 broken" to "we can't ship without it".  You can't have it both ways.  So pick one, and be prepared to stick with it.

As an RM, I'd ship with slightly lower shared memory performance in a heartbeat, if it meant getting the correct answer.  100% of my customers are willing to make that tradeoff.

hjelmn:
The fixes we put in broke x86_64. Either we back out the "fixes" and do this PR or I have to figure out what is going on. Once x86_64 broke I have to find time to work on this.

hjelmn:
Think I identified the problem on x86_64. Its a really subtle bug in the fixes. testing now. So far in 100 iterations it hasn't hung/faulted.

hppritcha:
4.0 release managers are getting really nervous about all these last minute vader fixes. 
We will want this patch or something similar to allow for an out so as to not have to release a 4.0.1 within days of the 4.0.0.

hjelmn:
@hppritcha I disagree. The code that was in the releases worked on 64-bit platforms but had issues on 32-bit (discovered recently). In an attempt to make it work for 32-bit we broke 64-bit. The recent changes fix 64-bit while still allowing it to work with 32-bit/

gpaulsen:
@hjelmn, When @hppritcha and I talked earlier today, we were discussing that it would be nice to have either an mca parameter or configure parameter to enable|disable the fastbox feature of vader (defaulting to enabled).
Could you take a look at this PR, and see if you think it's suitable for disabling fastbox if a user so chose?

hjelmn:
FWIW its possible to disable fast boxes without this. Just need to set an MCA parameter.

gpaulsen:
Oh, that might be sufficient.  What is that parameter?  I can't seem to find it.

hjelmn:
btl_vader_fbox_max. I completely forgot about it. It was meant to control the memory footprint of vader.

bwbarrett:
Re-closing this PR.  While Nathan and I might disagree about how critical fbox support is to Open MPI as a whole, he's right that btl_vader_fbox_max is a better control than this PR for A/B testing with/without fbox support and as a workaround to users.

Issue #5842, opened 2018-10-04
New vader SEGV possibly PR 5829


gpaulsen:
@amckinstry commented on https://github.com/open-mpi/ompi/pull/5829

Original issue was: https://github.com/open-mpi/ompi/issues/5638

Since that was merged to all release branches, this is a *blocker* on all release branches.

```
Unfortunately we now see a related crash on other codes (lammps):

#0 0x0000000000000000 in ()
#1 0x00007f7a700f8a5f in mca_btl_vader_poll_handle_frag (hdr=0x7f7a6a1bf049, endpoint=endpoint@entry=0x55e02df87bd0) at btl_vader_component.c:603
#2 0x00007f7a700f8f83 in mca_btl_vader_check_fboxes () at btl_vader_fbox.h:225

#3 0x00007f7a700f8f83 in mca_btl_vader_component_progress () at btl_vader_component.c:702
#4 0x00007f7a828fb1cc in opal_progress () at runtime/opal_progress.c:228
#5 0x00007f7a8d816ebd in ompi_request_wait_completion (req=0x55e02df8d900) at ../ompi/request/request.h:413
#6 0x00007f7a8d816ebd in ompi_request_default_wait (req_ptr=0x7fffa70c3708, status=0x7fffa70c3710) at request/req_wait.c:42
#7 0x00007f7a8d86ec41 in ompi_coll_base_sendrecv_actual (sendbuf=sendbuf@entry=0x55e02df97de0, scount=scount@entry=1, sdatatype=sdatatype@entry=0x55e02cace1e0 <ompi_mpi_int>, dest=dest@entry=0, stag=stag@entry=-12, recvbuf=recvbuf@entry=0x7fffa70c3934, rcount=1, rdatatype=0x55e02cace1e0 <ompi_mpi_int>, source=0, rtag=-12, comm=0x55e02cacf700 <ompi_mpi_comm_world>, status=0x0) at base/coll_base_util.c:59

it looks like hdr->tag is invaliid, hence segdfault
```

gpaulsen:
@amckinstry, can you please confirm that you have commit dfa8d3a81ac64e32d2bfb13a9afb20b83d747e03 in your master build that you saw this with?  There were a few other vader related fixes that went in around #5829

gpaulsen:
@amckinstry - @hjelmn asked what platforms this affects.  If it only affects i386, we may not have a developer to fix this.

jsquyres:
[...]

PR #5852, opened 2018-10-05
btl/vader: fix race condition in writing header

hjelmn:

### DIFF
--- a/opal/mca/btl/vader/btl_vader_fbox.h
+++ b/opal/mca/btl/vader/btl_vader_fbox.h
@@ -29,6 +29,10 @@ typedef union mca_btl_vader_fbox_hdr_t {
         /** sequence number */
         uint16_t  seq;
     } data;
+    struct {
+        uint32_t value0;
+        uint32_t value1;
+    } data_i32;
     uint64_t ival;
 } mca_btl_vader_fbox_hdr_t;
 
@@ -50,19 +54,20 @@ void mca_btl_vader_poll_handle_frag (mca_btl_vader_hdr_t *hdr, mca_btl_base_endp
 static inline void mca_btl_vader_fbox_set_header (mca_btl_vader_fbox_hdr_t *hdr, uint16_t tag,
                                                   uint16_t seq, uint32_t size)
 {
-    mca_btl_vader_fbox_hdr_t tmp = {.data = {.tag = 0, .seq = seq, .size = size}};
-    hdr->ival = tmp.ival;
+    mca_btl_vader_fbox_hdr_t tmp = {.data = {.tag = tag, .seq = seq, .size = size}};
+    /* clear out existing tag/seq */
+    hdr->data_i32.value1 = 0;
     opal_atomic_wmb ();
-    hdr->data.tag = tag;
+    hdr->data_i32.value0 = size;
+    opal_atomic_wmb ();
+    hdr->data_i32.value1 = tmp.data_i32.value1;
 }
 
 static inline mca_btl_vader_fbox_hdr_t mca_btl_vader_fbox_read_header (mca_btl_vader_fbox_hdr_t *hdr)
 {
-    mca_btl_vader_fbox_hdr_t tmp;
-    uint16_t tag = hdr->data.tag;
+    mca_btl_vader_fbox_hdr_t tmp = {.data_i32 = {.value1 = hdr->data_i32.value1}};;
     opal_atomic_rmb ();
-    tmp.ival = hdr->ival;
-    tmp.data.tag = tag;
+    tmp.data_i32.value0 = hdr->data_i32.value0;
     return tmp;
 }
 


hjelmn:
@amckinstry This appears to fix #5842. The only thing I can think of is when the fast-box rolls empties there is a chance for a race on the header the way we tried to fix i386. This PR should fix the race. Let me know.

hjelmn:
FWIW, I have run the crack LAMMPS problem ~ 15k times and have not seen the issue with this change. I was able to reproduce it in a ubuntu docker container.

hjelmn:
@bwbarrett Please review. The race is really subtle and only happens when end == start. I don't fully understand the race but its gone as far as I can tell.

bwbarrett:
I can't figure out why this fixes anything.  You were able to reproduce this in a VM?  What was your setup to do so?

To make the problem go away, do you need both write memory barriers?  I'm having trouble figuring out why a wmb after the first word but before size is necessary.

hjelmn:
I can  try without the first memory barrier. Was able to reproduce the hang in qemu docker image. I will look at the assembly output and see if I can figure out why this fixes the issue. As I said, it only happens when the start catches up to the end.

Only thing I can think of is if somehow the sequence number gets written before the tag is set to 0 then there would be a chance for the message to be processed before the tag is written.

hjelmn:
Yeah, works fine without the first memory barrier. Was being too cautious.

hjelmn:
Ran ~ 1k times without the first memory barrier. Looks good.

bwbarrett:
Ok, I think I understand why this fixes the problem.  I think this fix works because compilers are unlikely to prefer two 16 bit stores over a single 32 bit store, but it would be nice to figure out how to zero out the fbox header on the previous write, when the compiler is unlikely to see it, in the case of wrap-around in the fbox.  Nathan and I need to iterate a little on a better long term fix (perhaps in San Jose makes the most sense).

There are two issues with the code.  First, when the send wraps to the start of the fbox buffer, the wrap code doesn't zero the next fbox header (like it does for the non-wrap case).  This would be ok, except for the second problem.  The code before this patch is:
```
mca_btl_vader_fbox_hdr_t tmp = {.data = {.tag = 0, .seq = seq, .size = size}};
hdr->ival = tmp.ival;
opal_atomic_wmb ();
hdr->data.tag = tag;
```
If we optimize out the temprary assignments, which is perfectly legal for the compiler to do, we end up with:
```
hdr->data.tag = 0;
hdr->data.seq = seq;
hdr->data.size = size;
opal_atomic_wmb ();
hdr->data.tag = tag;
```
This code still looks ok at a high level, but I think is the bug.  Because there's no indication of side effects to the write of `tag`, the compiler is able to eliminate the first write, which leaves the following code:
```
hdr->data.seq = seq;
hdr->data.size = size;
opal_atomic_wmb ();
hdr->data.tag = tag;
```
Normally, this wouldn't be a problem, because the cleanup code at the end of fbox_sendi() sets the next header to 0.  But, as I mentioned earlier, on a wrap around to the start of the buffer, the header isn't 0ed out, so tag may have a valid value.  And since the `seq` field *was* written with the new value, the receive side thinks there's a valid buffer.  So the bug wasn't every time that `start == end`, but on the `start == end` that happens on a wrap.

With the new code, we have:
```
mca_btl_vader_fbox_hdr_t tmp = {.data = {.tag = tag, .seq = seq, .size = size}};
hdr->data_i32.value1 = 0;
hdr->data_i32.value0 = size;
opal_atomic_wmb ();
hdr->data_i32.value1 = tmp.data_i32.value1;
```
which expands to:
```
hdr->data_i32.value1 = 0;
hdr->data_i32.value0 = size;
opal_atomic_wmb ();
hdr->data_i32.value1 = (tag | seq << 32);
```
if we then optimize out the duplicate assignment, we get:
```
hdr->data_i32.value0 = size;
opal_atomic_wmb ();
hdr->data_i32.value1 = (tag | seq << 32);
```
As long as the compiler emits `hdr->data_i32.value1 = (tag | seq << 32);` instead of `hdr->data.tag = tag; hdr->data.seq = seq`, this code will work.  If it tries to save the shift and or and emits the two stores, then the code is still potentially wrong.  Looking at my assembly, it looks like it's doing the single store.  I think if we want to make sure we're 100% safe, we should change the code to:
```
hdr->data.size = size;
hdr->data.tag = tag;
opal_atomic_wmb();
hdr->data.seq = seq;
```
The current receive code validates the sequence number and tag both have the expected value (next sequence number, and a non-zero tag), so forcing the tag to be valid and using the sequence number for validation the tag is valid is a reasonable approach.  This is one of the problems `READ_ONCE()/WRITE_ONCE()` are intended to deal with in the Linux kernel, but their implementation depends on particular behaviors of GCC, which we can't count on in the general case.

jsquyres:
Should we put all of @bwbarrett's above comment in a code comment somewhere so that this knowledge isn't lost?

hjelmn:
I can put in a comment. I think @bwbarrett is correct in his analysis.

bwbarrett:
I wasn't going to worry about comments just yet, because I think there's some opportunities to clean it up just a little bit and simplify the code/comments.  Might be easier to describe what our ordering dependencies are and then point to the Linux kernel docs on these issues.   Search for ACCESS_ONCE and you'll find them; they're really quite well written (unlike most kernel docs).

hjelmn:
@bwbarrett Ok, lets do this one right now and deal with the cleanup later.

hjelmn:
We have confirmation from the user that this fixes the issue on all affected platforms.

bwbarrett:
@hjelmn yeah, totally agree.  Let's get this version in and iterate a bit on both the code and comments :).

PR #5856, opened 2018-10-07
Clean up asprintf warnings

bwbarrett:
Ralph beat me to the orte cleanup, but finish the job of calling
opal_asprintf instead of asprintf.  Open MPI's developers have
long relied on behavior of asprintf (that the ptr is set to NULL
on error) which Linux does not guarantee.


bwbarrett:
@rhc54, nope, my fault.  I pushed the branch to my fork and forgot to post the PR.  Sigh :).  Fixing wasn't bad at all.  And now, we're almost warning free on MacOS.

PR #5866, opened 2018-10-09
v2.x: util/hostfile: fix a double free error

ggouaillardet:
As reported at https://stackoverflow.com/questions/52707242/mpirun-segmentation-fault-whenever-i-use-a-hostfile
mpirun crashes when the hostfile contains a "user@host" line.
The root cause is username was not strdup'ed and free'd twice by opal_argv_free() and free()

(back-ported from commit open-mpi/ompi@5803385d447957ef5999fb93052c0712a0b47a81)


jsquyres:
@hppritcha Trivial error+fix for not-uncommon case; worthy of v2.x.  Good to go.

PR #5874, opened 2018-10-09
v4.0.0: Fix configury for internal PMIx

rhc54:
Cherry-pick several fixes for when users configure ```--with-hwloc=internal``` or ```--with-hwloc=external```

Fixes #5871

Issue #5871, opened 2018-10-09
4.0rc4 won't compile with external PMIx 2.1.2


angainor:
I'm having hard time getting v4 to work on our SLURM cluster. I configure it with external PMIx / HWLOC / libevent libraries:

```
./configure --prefix=[...] --enable-mpi-cxx --with-pmi --with-pmix=external --with-hwloc=external --with-libevent=external --with-mxm=${HPCX_MXM_DIR} --with-hcoll=${HPCX_HCOLL_DIR} --with-ucx=${HPCX_UCX_DIR} --with-knem=${KNEM_DIR} --with-platform=contrib/platform/mellanox/optimized
```
configure works ok
```
Miscellaneous
-----------------------
CUDA support: no
HWLOC support: external
Libevent support: external
PMIx support: 2x
```
but compilation fails
```
make[2]: Entering directory `/tmp/marcink/openmpi-4.0.0rc4/opal/mca/pmix/ext2x'
  CC       mca_pmix_ext2x_la-ext2x_local.lo
  CC       mca_pmix_ext2x_la-ext2x.lo
  CC       mca_pmix_ext2x_la-ext2x_client.lo
  CC       mca_pmix_ext2x_la-ext2x_component.lo
  CC       mca_pmix_ext2x_la-ext2x_server_north.lo
  CC       mca_pmix_ext2x_la-ext2x_server_south.lo
ext2x_local.c:21:10: fatal error: ext2x.h: No such file or directory
 #include "ext2x.h"
          ^~~~~~~~~
compilation terminated.
make[2]: *** [mca_pmix_ext2x_la-ext2x_local.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
ext2x_component.c:26:10: fatal error: ext2x.h: No such file or directory
 #include "ext2x.h"
          ^~~~~~~~~
compilation terminated.
make[2]: *** [mca_pmix_ext2x_la-ext2x_component.lo] Error 1
ext2x.c:42:10: fatal error: ext2x.h: No such file or directory
 #include "ext2x.h"
          ^~~~~~~~~
compilation terminated.
ext2x_client.c:37:10: fatal error: ext2x.h: No such file or directory
 #include "ext2x.h"
          ^~~~~~~~~
compilation terminated.
make[2]: *** [mca_pmix_ext2x_la-ext2x.lo] Error 1
ext2x_server_south.c:43:10: fatal error: ext2x.h: No such file or directory
 #include "ext2x.h"
[...]

PR #5883, opened 2018-10-10
pmix/ext: misc fixes

ggouaillardet:


ggouaillardet:
@rhc54 @jsquyres I fixed several issues related to #5871 
These errors are also in master, so these two commits could/should be cherry-picked instead of the last (and one-shot) commit from #5874

rhc54:
> @rhc54 @jsquyres I fixed several issues related to #5871
> These errors are also in master, so these two commits could/should be cherry-picked instead of the last (and one-shot) commit from #5874

Horse-manure. Leave #5874 alone as it was tested as a group. 😄  Besides, the errors in v4.0 were only in a single area (ext2x), so most of this PR wouldn't apply.

PR #5886, opened 2018-10-10
osc_ucx: fix hang/timeout in component finalize

yosefe:
Fixes a hang found in MTT testing of UCX RMA emulation


yosefe:
@hoopoepg can you pls review?

yosefe:
it's actually required to have such barrier
using collective call, rather than PMI-x, so it would be faster

PR #5890, opened 2018-10-10
osc_ucx: add worker flush before osc module free

yosefe:
@jsquyres @hoopoepg this addresses https://github.com/open-mpi/ompi/pull/5886#issuecomment-428657968


jsquyres:
👍 Thanks.

PR #5892, opened 2018-10-11
btl tcp: Work around shutdown race bug

bwbarrett:
Work around a shutdown race in the TCP BTL when one process closes
the endpoint socket and the FIN arrives at the other process before
the second process shuts down.  In this case, the socket will
close, readv() will return 0, and the TCP BTL will push the error
to the PML, which will abort the job (from MPI_FINALIZE, which is
just rude).

There are two commits that appear to cause this issue.  First, in
ed8141e7c0 (Dec 2014), we started setting the endpoint state to
FAILED on a readv() returning 0.  Second, in 6acebc40a1 we started
pushing the error to the PML for any FAILED state endpoints that
enter endpoint_close().

The best workaround I could think of without rewriting lots of
code is to only error to the PML if there were frags in flight.
There is a long comment in the commit as to why I think this is
a reasonable approach for now, as well as future approaches.  The
future approach, I think, is to add some auto-retry to the endpoint
so that a single TCP disconnect doesn't abort a job.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_endpoint.c
+++ b/opal/mca/btl/tcp/btl_tcp_endpoint.c
@@ -542,19 +542,44 @@ void mca_btl_tcp_endpoint_close(mca_btl_base_endpoint_t* btl_endpoint)
      * re-route or re-schedule the fragments.
      */
     if( MCA_BTL_TCP_FAILED == btl_endpoint->endpoint_state ) {
+        bool frag_cleaned = false;
         mca_btl_tcp_frag_t* frag = btl_endpoint->endpoint_send_frag;
         if( NULL == frag )
             frag = (mca_btl_tcp_frag_t*)opal_list_remove_first(&btl_endpoint->endpoint_frags);
         while(NULL != frag) {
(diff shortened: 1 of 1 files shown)


bwbarrett:
I was able to replicate Jeff's stream of "Socket closed" error message/test failures on master on a 5 node cluster with 3 IP interfaces / node, running 18 tasks per node (so 90 total), approximately 30% of the runs.  With this patch, I could no longer replicate the failure.

I'm open to better ideas, but this seemed simple and clean enough.  Also curious if George remembers why he did the FAILED change; the commit message wasn't enlightening.

bwbarrett:
@abouteiller You added the original bug when you started bubbling up errors to the PML.  I don't have time to figure out how to better handle this situation in your error handling code (that doesn't work in the PML most customers use).  We're going to push this patch, and you're welcome to iterate on a fix that works for everyone.

bosilca:
Excuse me ?

bwbarrett:
@bosilca if you'd prefer, I can roll back the patch that broke master.  This seems less bad.

bwbarrett:
Closed in favor of https://github.com/open-mpi/ompi/pull/5916.

PR #5895, opened 2018-10-11
BTL/OB1: added own copy of iface-attr for every ep

hoopoepg:
- fixed issue in incorrect put limits when MEMIC is used

fix for https://github.com/openucx/ucx/issues/2937


hoopoepg:
as I can see you create own iface per ep, all such ifaces may have different limits for short's

hjelmn:
ah, ok. to save space though we should only save the relevant values (max short, max bcopy, etc). saving the whole attribute structure is a little much (since it is big).

hoopoepg:
ok, do you want I update this PR or you create another one?

hjelmn:
might be better to create another one as there will be fewer changes. thanks for looking into this. i still find this behavior to uct a little odd. probably should be documented at some point.

PR #5896, opened 2018-10-11
coll/libnbc: add recursive doubling algorithm for MPI_Iallgather

mkurnosov:
Implements recursive doubling algorithm for MPI_Iallgather. The algorithm can be used only for power-of-two number of processes.


hjelmn:
At some point we should add a C enum. This is fine as is though.

PR #5901, opened 2018-10-11
btl/uct: use the correct tl interface attributes

hjelmn:
It is apparently possible for different instances of the same UCT
transport to have different limits (max short put for example). To
account for this we need to store the attributes per TL context not
per TL. This commit fixes the issue.


hjelmn:
@hoopoepg This PR does essentially the same thing as #5895 but puts the attributes on the TL context.

hoopoepg:
did you have chance to test this fix? does it resolve issue?

hjelmn:
Yes. Seems to. The first thread gets to be a little faster due to the higher short limit but that isn't a big deal.

PR #5879, opened 2018-10-09
OSC/UCX: fixed zero-size window processing

hoopoepg:
- added processing of zero-size MPI window

fixes https://github.com/open-mpi/ompi/issues/5873

Signed-off-by: Sergey Oblomov <sergeyo@mellanox.com>

PR #5911, opened 2018-10-12
OSC/UCX: fixed zero size window - v3.1

hoopoepg:
added processing of zero-size MPI window

backport from https://github.com/open-mpi/ompi/pull/5879, but with additional fix for logging infrastructure


yosefe:
@bwbarrett ok to merge?

Issue #5849, opened 2018-10-05
"Socket closed" messages from TCP BTL


jsquyres:
Cisco MTT is getting a lot of "Socket closed" messages from the TCP BTL.

For example: https://mtt.open-mpi.org/index.php?do_redir=2762

That message appears to come from here:

https://github.com/open-mpi/ompi/blob/5f1c940c8b252763b9714f5a3b813b3bf2847479/opal/mca/btl/tcp/btl_tcp_endpoint.c#L542-L558

I don't know if this is an actual error (that the user should be informed about) or a "it's ok, we'll just ignore it and move on to the next interface in the list" kind of issue (that could probably only be reported with a high enough verbosity).  Regardless, "Socket closed" is probably not enough of a detailed message to convey meaningful information to the end user.  😄 

I don't know if this is related to https://github.com/open-mpi/ompi/issues/3035 or https://github.com/open-mpi/ompi/issues/5818, but it's worth cross-referencing them here.

bwbarrett:
https://github.com/open-mpi/ompi/pull/5892 is a workaround for this problem.  This is unrelated to  #3035 or #5818 and only happens on master.

Jeff actually got the code slightly wrong.  The issue is on the next line down, where the BTL calls back to the PML with the disconnect.  OB1 just aborts the job on an error callback, but in this case, the callback is fired because readv() returned 0 because the other side was in MPI_FINALIZE and shut down the socket.

bwbarrett:
Based on MTT results, da1189d77142df412592fb815cb59f688028e528 (which reverted 6acebc4) fixed the test failures.  Since there are open issues for all the other issues we're seeing, I'm going to close this ticket.

PR #5916, opened 2018-10-12
Revert "Handle error cases in TCP BTL"

bwbarrett:
This reverts commit 6acebc40a194c92ab38a28553c2c8b04eb391820.

This patch is causing numerous "Socket closed" messages which are
causing most of the failures on Cisco's MTT run.  See
https://github.com/open-mpi/ompi/issues/5849 for more information.


bwbarrett:
Discussion at the developer forum was that we'd commit this to unblock nightly testing.  @bosilca and @abouteiller will work on a more correct fix to OB1 and TCP to add this feature back in the tree.

Issue #5915, opened 2018-10-12
NUMA issues with shared memory


devreal:
## Background information

I've been tracking down a performance issue in an application I ported to a PGAS programming model that is implemented on top of MPI-3 RMA. To avoid NUMA issues, I ran one multi-threaded process per socket on a dual-socket Haswell system, which led to one process on a node being seemingly slower than the other. I was able to reproduce a similar performance difference with a STREAM benchmark. 

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

3.1.2 build from release tarball using GCC 8

### Please describe the system on which you are running

Reproduced on 2 systems: Cray XC40 and an Infiniband Cluster, both dual-socket 2x12 core Haswell.

-----------------------------

## Details of the problem

I am attaching the stream benchmark that is based on the official STREAM release and extended to use memory allocated either through `malloc` or through `MPI_Win_allocate`. All tests were done using 10 repetitions on 1GB of memory per process.

### Cray XC40

On the Cray system, running with 2 processes, 12 threads each, I see the following performance with Open MPI when running two processes per node:

```
$ PMI_NO_FORK=1 OMP_NUM_THREADS=12 aprun -n 2 -N 2 -d 12 -cc numa_node ./stream_mpi_rma 
******************************
* STREAM using  2x12 threads *
******************************
******************************
* STREAM using malloc()      *
******************************
[0] Function    Best Rate MB/s  Avg time     Min time     Max time
[0] Copy:          39145.8     0.054980     0.054859     0.055323
[0] Scale:         39068.1     0.055055     0.054968     0.055153
[0] Add:           44531.7     0.072443     0.072336     0.072537
[0] Triad:         44591.2     0.072339     0.072239     0.072419
[1] Function    Best Rate MB/s  Avg time     Min time     Max time
[1] Copy:          39144.1     0.054931     0.054861     0.055305
[1] Scale:         39062.1     0.055025     0.054976     0.055075
[1] Add:           44544.4     0.072359     0.072315     0.072416
[1] Triad:         44599.0     0.072303     0.072226     0.072381
***********************************
* STREAM using MPI_Win_allocate() *
***********************************
[0] Function    Best Rate MB/s  Avg time     Min time     Max time
[0] Copy:          22285.0     0.096559     0.096365     0.097148
[0] Scale:         22229.7     0.096724     0.096604     0.096872
[0] Add:           25377.1     0.127100     0.126934     0.127537
[...]

PR #5921, opened 2018-10-16
 RDMA OSC: initialize segment memory before registering the segment 

devreal:
This PR causes memory in a shared memory segment to be initialized by the ranks participating in its creation before rank 0 registers the memory with the network device. The goal is to have the pages for each rank's portion properly allocated in the corresponding NUMA domain. 

This PR addresses #5915

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -620,6 +620,7 @@ static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, s
 
         if (size && MPI_WIN_FLAVOR_ALLOCATE == module->flavor) {
             *base = (void *)((intptr_t) module->segment_base + my_base_offset);
+            memset (*base, 0, size);
         }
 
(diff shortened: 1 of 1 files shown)


hjelmn:
Looks fine to me. Please squash this down to a single commit.

hjelmn:
Oh, and when squashed run ```git commit -s --amend```  to add the signed-off by line.

hjelmn:
Odd, I wonder why my fifo warning fix is in there. Can you rebase on master?

devreal:
Sorry, that was my mistake while doing the squash. It should be clean and signed now.

hjelmn:
Thanks for tracking this down. I assumed the app first touch would be enough to ensure the locality of the memory but guess that is not always true.

jsquyres:
@hjelmn This didn't run thought IBM or AWS CI because this came from a non-core contributor.  One of needs to type "ok to test" to give IBM and AWS CI the ok to run.

hjelmn:
Ah, ok. Didn't notice that.

PR #5922, opened 2018-10-16
opal/free_list: fix race condition

### LINKED ISSUE #2921: master: MPI_Irecv seems not to be thread safe?
This is the message I get from multithreaded injection rate benchmark. The benchmark post multiple Irecv on the receiver side (multiple threads at a time). 

This happens on both [Artem's benchmark](https://github.com/artpol84/poc) and my own benchmark and only happens on Open MPI and has been happening for a while. (I tested with Intel MPI and everything is ok)

```
[mpi002:27049] *** An error occurred in MPI_Irecv
[mpi002:27049] *** reported by process [56688641,1]
[mpi002:27049] *** on communicator MPI COMMUNICATOR 6 DUP FROM 0
[mpi002:27049] *** MPI_ERR_INTERN: internal error
[mpi002:27049] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[mpi002:27049] ***    and potentially your MPI job)
```

It happens with different BTLs and I think @jsquyres can confirm this.

It's not happening all the time. This will happens more often if multiple threads are bounded to the same core. This suggests something happens between context switching?

hjelmn:
There was a race condition in opal_free_list_get. Code throughout the
Open MPI codebase was assuming that a NULL return from this function
was due to an out-of-memory condition. In some cases this can lead to
a fatal condition (MPI_Irecv and MPI_Isend in pml/ob1 for
example). Before this commit opal_free_list_get_mt looked like this:

```c
static inline opal_free_list_item_t *opal_free_list_get_mt (opal_free_list_t *flist)
{
    opal_free_list_item_t *item =
        (opal_free_list_item_t*) opal_lifo_pop_atomic (&flist->super);

    if (OPAL_UNLIKELY(NULL == item)) {
        opal_mutex_lock (&flist->fl_lock);
        opal_free_list_grow_st (flist, flist->fl_num_per_alloc);
        opal_mutex_unlock (&flist->fl_lock);
        item = (opal_free_list_item_t *) opal_lifo_pop_atomic (&flist->super);
    }

    return item;
}
```

The problem is in a multithreaded environment is *is* possible for the
free list to be grown successfully but the thread calling
opal_free_list_get_mt to be left without an item. The happens if
between the calls to opal_lifo_push_atomic in opal_free_list_grow_st
and the call to opal_lifo_pop_atomic other threads pop all the items
added to the free list.

This commit fixes the issue by ensuring the thread that successfully
grew the free list **always** gets a free list item.

Fixes #2921


hjelmn:
@thananon This is the race that caused IRecv to fail.

thananon:
#2921 for reference.

Thank you Nathan, I will test it as soon as I can. I'm at face to face meeting right now.

Issue #5935, opened 2018-10-17
testmpio bug fixes


edgargabriel:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

master, v4.0

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

OpenSuSE Leap 42.3
-----------------------------

## Details of the problem

Based on a discussion on the users mailing list, we looked at the testmpio test suite for parallel I/O, which revealed a couple of bugs. Specifically:

- return the correct error codes if trying to write a file that was opened in RDONLY mode, and reading a file that was opened using WRONLY
- check that the fileview data type is a multiple of the etype 
- a bug in the position calculation in File_seek if SEEK_SET is being used
- a bug in the File_sync due to missing Barrier before the sync operation
- detect and report an error if an invalid data representation is being used

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

edgargabriel:
This issue has been fixed on master and v4.0, not sure whether we should backport this also to the 3.1 release series?

PR #5936, opened 2018-10-17
Pr/testmpio v4.0.x

edgargabriel:
these commits bring over the fixes for issue #5935 to the v4.0 branch.

### DIFF
--- a/ompi/mca/io/ompio/io_ompio_file_open.c
+++ b/ompi/mca/io/ompio/io_ompio_file_open.c
@@ -400,8 +400,9 @@ int mca_io_ompio_file_seek (ompi_file_t *fh,
         }
         break;
     case MPI_SEEK_CUR:
-        offset += data->ompio_fh.f_position_in_file_view;
-        offset += data->ompio_fh.f_disp;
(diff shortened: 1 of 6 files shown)


ompi/mca/io/ompio/io_ompio_file_open.c
```diff
@@ -372,6 +372,13 @@ int mca_io_ompio_file_sync (ompi_file_t *fh)
         OPAL_THREAD_UNLOCK(&fh->f_lock);
         return MPI_ERR_ACCESS;
     }        
+    // Make sure all processes reach this point before syncing the file.
+    ret = data->ompio_fh.f_comm->c_coll->coll_barrier (data->ompio_fh.f_comm,
+                                                       data->ompio_fh.f_comm->c_coll->coll_barrier_module);
```
gpaulsen: @edgargabriel How important is this barrier in mca_io_ompio_file_sync ?

edgargabriel:
@gpaulsen there are scenarios that require this to have properly, It is however not a regression, this bug was in ompio since the very beginning. I would put also the bug in the file_seek calculation (also part of this pr) at least en par with file_sync problem.

PR #5937, opened 2018-10-17
remove some dead crs components

hppritcha:


hppritcha:
@jsquyres need review for rc5

jsquyres:
@bwbarrett Did we just accidentally remove something that you might care about?  (these components were all `.ompi_ignore`d, and Howard and I were just cleaning house -- let us know if we need to revert this...?)

Issue #5818, opened 2018-10-01
TCP BTL: problems when there are multiple IP addresses on a single interface


jsquyres:
This issue is split off from #3035 and #5817.

On one of my development servers that has multiple IP addresses on a single IP interface, I get 100% failure with:

* Running our "ring_c" example MPI app on a single server
* TCP BTL and no shared memory
* After apply #6817

Here's the details:

```
$ mpirun -np 2 --mca btl tcp,self ring_c
Process 0 sending 10 to 1, tag 201 (2 processes in ring)
--------------------------------------------------------------------------
Open MPI detected an inbound MPI TCP connection request from a peer
that appears to be part of this MPI job (i.e., it identified itself as
part of this Open MPI job), but it is from an IP address that is
unexpected.  This is highly unusual.

The inbound connection has been dropped, and the peer should simply
try again with a different IP interface (i.e., the job should
hopefully be able to continue).

  Local host:          savbu-usnic-a
  Local PID:           4674
  Peer hostname:       savbu-usnic-a ([[17212,1],0])
  Source IP of socket: 10.193.184.48
  Known IPs of peer:   
        10.0.8.254
        10.193.184.49
        10.10.0.254
        10.2.0.254
        10.3.0.252
        10.50.0.254
--------------------------------------------------------------------------
[hang]
```

Here's the `ip addr` from this machine:

```
$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 16436 qdisc noqueue state UNKNOWN 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    link/ether 60:73:5c:68:f3:68 brd ff:ff:ff:ff:ff:ff
    inet 10.0.8.254/16 brd 10.0.255.255 scope global eth0
    inet6 fe80::6273:5cff:fe68:f368/64 scope link 
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN qlen 1000
    link/ether 60:73:5c:68:f3:69 brd ff:ff:ff:ff:ff:ff
[...]

PR #5938, opened 2018-10-17
Always use same IP address for module and modex

bwbarrett:
Two patches that end up with simplifying the modex send so that we're not racy in a multi-IP per NIC design.  We were already storing a source address in the local BTL, for use to bind() and force the source address to be known to us.  With this patch, we guarantee that the address we use for bind() is the same we publish in the modex.

This should help with #5818.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_component.c
+++ b/opal/mca/btl/tcp/btl_tcp_component.c
@@ -513,17 +513,12 @@ static int mca_btl_tcp_create(int if_kindex, const char* if_name)
         btl->tcp_send_handler = 0;
 #endif
 
-       struct sockaddr_storage addr;
-       opal_ifkindextoaddr(if_kindex, (struct sockaddr*) &addr,
-                                          sizeof (struct sockaddr_storage));
-#if OPAL_ENABLE_IPV6
-        if (addr.ss_family == AF_INET6) {
-            btl->tcp_ifaddr_6 =  addr;
-        }
-#endif
-        if (addr.ss_family == AF_INET) {
-           btl->tcp_ifaddr = addr;
-        }
+        /* save the address associated with this kindex in the module
+           to publish in the modex and use as the source of all packets
+           from this module */
+        opal_ifkindextoaddr(if_kindex, (struct sockaddr*)&btl->tcp_ifaddr,
+                            sizeof (struct sockaddr_storage));
+
         /* allow user to specify interface bandwidth */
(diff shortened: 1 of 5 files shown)


opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -1121,97 +1117,64 @@ static int mca_btl_tcp_component_create_listen(uint16_t af_family)
...
-                     addrs[current_addr].addr_inuse   = 0;
-                     addrs[current_addr].addr_ifkindex =
-                         opal_ifindextokindex (index);
-                     current_addr++;
-                     opal_output_verbose(30, opal_btl_base_framework.framework_output,
-                                         "btl: tcp: component_exchange: "
-                                         "%s IPv6 %s", ifn,
-                                         opal_net_get_hostname((struct sockaddr*)&my_ss));
-                 } else
+         if (AF_INET6 == addr->sa_family) {
```
bosilca: We lost the check for mca_btl_tcp_component.tcp_disable_family (there is no need to pack an address that will never be used).

bwbarrett: Gah, you're right.  Will fix.

ibm-ompi:

 Gist: https://gist.github.com/be6a1effefe5880ecad68620abda07a9

ibm-ompi:

 Gist: https://gist.github.com/49f4e6aeed925d5420add9331e8d9b30

ibm-ompi:

 Gist: https://gist.github.com/a433e13e3416ae2c7b680f36a3ad1b16

jjhursey:
bot:ibm:retest

opal/mca/btl/tcp/btl_tcp_component.c
```diff
@@ -490,11 +490,59 @@ static int mca_btl_tcp_component_close(void)
...
-static int mca_btl_tcp_create(int if_kindex, const char* if_name)
+static int mca_btl_tcp_create(const int if_kindex, const char* if_name)
 {
     struct mca_btl_tcp_module_t* btl;
     char param[256];
     int i;
+    struct sockaddr_storage addr;
+    bool found = false;
+
+    /*
```
bosilca: I am not sure we are not overdoing it. We have the unique name of the iface we want to use, why are we trying to play around parsing all interfaces in order to find the correct kindex instead of just using the if_name via opal_ifnametokindex ?

PR #5963, opened 2018-10-23
platform/mellanox: disable btl-uct by default — v4

amaslenn:
(cherry picked from commit 074e9cc92c79e6b9f8c492b2613eac3f27e1c522)

### DIFF
--- a/contrib/platform/mellanox/optimized
+++ b/contrib/platform/mellanox/optimized
@@ -1,4 +1,4 @@
-enable_mca_no_build=coll-ml
+enable_mca_no_build=coll-ml,btl-uct
 enable_debug_symbols=yes
 enable_orterun_prefix_by_default=yes
 with_verbs=no


amaslenn:
@yosefe

yosefe:
@gpaulsen can we merge this? it's important for us because of btl_uct compilation issues

hjelmn:
@yosefe Again, #5966 is the correct approach. Instead of disabling (and not testing with CI) code it should be fixed.

yosefe:
@hjelmn in general I agree, but currently we need to remove btl/uct from platform/mellanox/optimized, due to issues of supporting it. Since it's Mellanox-only configuration, I hope it wouldn't be a problem.

jsquyres:
Ya, this is an organization-specific platform file.  We generally don't care what individual organizations do in their own platform files.

hjelmn:
As long as it doesn't affect CI I don't really care :). My understanding is this platform file is used by Mellanox CI, is that incorrect?

yosefe:
@hjelmn yes, it's used by Mellanox CI

hjelmn:
Then that I find as a problem. With #5966 the build issue is fixed and for v4.0.x the component is disabled by default so I see no reason to not build it. That way your CI can at least run a build test on btl/uct changes. I don't know of any other CI system which has OpenUCX right now.

hjelmn:
You can also force the component not to load by adding btl = ^uct to your MCA parameters file.

yosefe:
@hjelmn looks like you are not going to merge the fix for a week at least, and in meantime our CI is broken

jsquyres:
Oh, I didn't realize that was the Mellanox CI config file (i.e., the one that is used by open-mpi/ompi github CI, right?).  If so, then I would tend to agree with @hjelmn -- we should just fix the problem.

yosefe:
@jsquyres @hjelmn I didn't mean the CI which tests OMPI PRs, but another one we have internally

hjelmn:
@yosefe Then I withdraw my objection. My primary concern is CI for PRs :).

yosefe:
@hjelmn actually, the PR CI also uses platform/mellanox/optimized, but it is not broken now because it compiles with UCX v1.4 and not UCX master. But i guess we can update the PR CI to run at least build test without a platform config.

PR #5968, opened 2018-10-24
Correctly propagate the oversubscribe flag to the spawnees

abouteiller:
In order for the 'oversubscribe' flag to be propagated to spawnees in an MPI_COMM_SPAWN and friends, the 'SUBSCRIBE_GIVEN' flag must also be set. 

The discussion in #5376 states that by default  OVERSUBSCRIBE would not be propagated to spawns. It however appears that the existing code is not conditional on the inherit mca parameter, hence, by making the feature work, this patch will also make it inherited 'by default'. This is unlike what was discussed in the issue. I believe that, unlike the slot assignments etc, it is useful to inherit `-oversubscribe` by default (same goes for `-nolocal` IMO), but can be debated. 

Signed-off-by: Aurélien Bouteiller <bouteill@icl.utk.edu>


CrashLaker:
Hi,
How do i set this?

Issue #5969, opened 2018-10-24
osc/rdma v3.1.3rc2: PR #5924 seems to introduce segfaults on simple tests


bsergentm:
### What version of Open MPI are you using?
v3.1.3rc2, commit 6187b7a2097f81e3e4a7d628213cb5dfa4d96b42
Open MPI repo revision from  `ompi_info --all` : v3.1.3rc1-2-gfa3d929

### Describe how Open MPI was installed

```shell
./configure '--prefix=/homes/sergentm/ompi-installdir' '--with-devel-headers' '--enable-binaries' '--enable-mpi-cxx' '--enable-mpirun-prefix-by-default' '--with-pmi=/homes/sergentm/export/slurm/usr' '--with-lustre=/opt/lustre_210_for_deps/usr' '--enable-wrapper-rpath=no' '--enable-wrapper-runpath=no' '--with-knem=/homes/sergentm/export/knem_dir/opt/knem-1.1.3.90mlnx1' '--with-cma' '--with-portals4=no' '--with-verbs=/usr' '--with-verbs-libdir=/usr/lib64' '--disable-openib-dynamic-sl' '--with-mxm=/opt/mellanox/mxm' '--with-ucx=no' 'CFLAGS=-g -pipe -m64' 'CXXFLAGS=-g -pipe -m64' 'FCFLAGS=-g -pipe -m64'
```

### Please describe the system on which you are running

* Operating system/version: 
Linux 3.10.0-693.21.1.el7.x86_64 #1 SMP Fri Feb 23 18:54:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Computer hardware: 
Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
* Network type: 
InfiniBand 100 Gb/sec (4X EDR)

More details on file "env.txt" attached here : [v3.1.x_osc-rdma_segment-register.tar.gz](https://github.com/open-mpi/ompi/files/2509788/v3.1.x_osc-rdma_segment-register.tar.gz)

-----------------------------

## Details of the problem

The commit 4fbe078359feb8606952cb991091e32dc4787140 : "RDMA OSC: initialize segment memory before registering the segment" of PR https://github.com/open-mpi/ompi/pull/5924 seems to introduce a regression on MPI-RMA applications which uses the Passive Target mode (lock/unlock).

I attached to this issue a simple reproducer of the issue. I launched it on the described platform both with srun and mpirun and have the same issue.

```shell
[...]

PR #5971, opened 2018-10-24
Fix regression in RDMA shared memory registration

devreal:
Fix regression introduced in #5921 NUMA by re-adding a node-local barrier to wait for memory registration to complete. Addresses #5969 

Tested on a Cray XC40 using the provided reproducer.


jsquyres:
ok to test

PR #5972, opened 2018-10-24
MTL/OFI: Add OFI Scalable Endpoint support

aravindksg:
OFI MTL supports OFI Scalable Endpoints feature as a means to improve
multi-threaded application throughput and message rate. Currently, the feature
is designed to utilize multiple TX/RX contexts exposed by the OFI provider in
conjunction with a multi-communicator MPI application model. For more
information, refer to README under mtl/ofi.

While there is some scaling from using multiple threads with this patch, in conjunction with PR #5241 (which allows multithreaded opal_progress()), there is significant improvement in performance.

### DIFF
--- a/ompi/mca/mtl/ofi/README
+++ b/ompi/mca/mtl/ofi/README
@@ -1,5 +1,5 @@
-OFI MTL
-
+OFI MTL:
+--------
 The OFI MTL supports Libfabric (a.k.a. Open Fabrics Interfaces OFI,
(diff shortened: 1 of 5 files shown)


hppritcha:
@matcabral honestly I'm not going to have time to test this anytime soon with GNI provider.  Go ahead and merge although you may want to wait till we've fixed the AWS jenkins.

matcabral:
bot:retest

PR #5975, opened 2018-10-25
Avoid a double lock interlock when calling pmix_finalize

abouteiller:
Due to taking multiple locks simultaneously, pmix-finalize may occasionally deadlock if there are PMIx event notifications pending. 

This patch delegates the deletion of the events outside of the locked section so that only one lock is held at any one time.


abouteiller:
@rhc54 I propagated the same change to the external PMIx interfaces ext2x and ext3x, as well as to the server_south which had the same erroneous code.

rhc54:
Excellent - thanks!

PR #5986, opened 2018-10-29
Add OPAL_VPID to unpacking

jjhursey:
* Needed to properly read PMIx job data like the following
   - `OPAL_PMIX_LOCALLDR`
   - `OPAL_PMIX_RANK`
   - `OPAL_PMIX_GLOBAL_RANK`
   - `OPAL_PMIX_APPLDR`
   - `OPAL_PMIX_APP_RANK`


jjhursey:
Yeah that's right. I was trying to access the `OPAL_PMIX_LOCALLDR` key, and found that there was no way to get to it without adding the unpack. Unpacking as an integer gives you the job portion, not the vpid.

PR #6002, opened 2018-10-31
Fix misc issues evidenced whey building on msys2 arch

ggouaillardet:


ggouaillardet:

the third commit open-mip/ompi@079fd7ee8745d05254bef507e4724491521a96dc fixes parallel build (`configure --disable-mca-dso && make -j` on www.msys2.org arch, not sure the arch is involved though). Serial `make` always works, but I added missing dependency (create the symlink after the linked file is built) in order to make `make -j` a happy panda. I revamped the logic (e.g. replace  `if` tests at build time with a `if` at `configure` time), and I would like to you review this since I might have missed something.

ibm-ompi:

 Gist: https://gist.github.com/f083c52b6188d8cfcff4183f5d490f03

jsquyres:
A few ancillary notes:

1. I note that `$(OMPI_V_LN_S)` is in `$topdir/Makefile.ompi-rules`.  We should probably just be including that file in these `.am` files rather than re-defining those macros (I'm guessing this just happened organically over time -- i.e., they were probably first defined in some `common/foo/Makefile.am` and then got copied to other `common/*/Makefile.am`s).
2. These `Makefile.am`s (and possibly/probably others?) have fragmented comments describing the `.la` creation scheme (e.g., some say "... case 2, described above" -- but it's not described above).
    * Is `opal/mca/common/sm/Makefile.am` the origin of all of these comments?  If so, it might be good to just refer to that `sm` `Makefile.am` in all the others rather than having fragmented / incoherent comments.

Per this PR:

1. Did you have to move the `all-local` and `clean-local` targets up inside the AM `if` because of the new dependency on `all-local`?  Otherwise, I'm not sure why you moved them up inside the `if`.  I ask because I have a super-dim recollection that there may be problems with conditionally defining targets...?

ggouaillardet:
@jsquyres I made some of the requested changes.

It seems I do not have to move the `all-local` and `clean-local` targets after all.
That being said, unless there is a good reason not to do so, I'd rather do it anyway for aesthetic reasons
(why generate a test that will be performed at `make` time when we already know the outcome at `configure` time)

ggouaillardet:
note the last commit is not meant to be committed.

ompi/mca/common/monitoring/Makefile.am
```diff
@@ -50,15 +54,7 @@ libmca_common_monitoring_la_LDFLAGS =  \
...
-# These two rules will sym link the "noinst" libtool library filename
-# to the installable libtool library filename in the case where we are
-# compiling this component statically (case 2), described above).
-V=0
-OMPI_V_LN_SCOMP = $(ompi__v_LN_SCOMP_$V)
-ompi__v_LN_SCOMP_ = $(ompi__v_LN_SCOMP_$AM_DEFAULT_VERBOSITY)
-ompi__v_LN_SCOMP_0 = @echo "  LN_S    " `basename $(component_install)`;
-
-all-local:
+all-local: $(component_noinst)
```
jsquyres: Since we're now including `Makefile.ompi-rules`, I think you need to also change `OMPI_V_LN_SCOMP` to `OMPI_V_LN_S` everywhere.

ompi/mca/common/monitoring/Makefile.am
```diff
@@ -13,11 +13,15 @@
 # $HEADER$
 #
 
+include $(top_srcdir)/Makefile.ompi-rules
```
jsquyres: The commit message on this commit no longer matches what the commit actually does.

I don't know / care if you want to split this into multiple commits:
* Add the dependencies
* Fix the comments
* Use Makefile.ompi-rules

jsquyres:
@ggouaillardet I added the DNM label so that you don't accidentally merge with that last commit ("note the last commit is not meant to be committed").

AboorvaDevarajan:

ibm-ompi:

 Gist: https://gist.github.com/8226fa70da089f68c8f204ce17635812

lanl-ompi:

jsquyres:
@ggouaillardet Do you plan to finish this PR?

ibm-ompi:

 Gist: https://gist.github.com/6d27c9e1e19019490cb6acaffd8162f4

hppritcha:
Closing.  Please reopen if needed.

PR #6015, opened 2018-11-01
MTL/OFI: Check threshold number of peers allowed per rank

aravindksg:
When the provider does not support FI_REMOTE_CQ_DATA, the OFI tag does not have
sizeof(int) bits for the rank. Therefore, unexpected behavior will occur when
this limit is crossed.

Check the max allowed number of ranks during add_procs() and return if there is
danger of exceeding this threshold.


matcabral:
This also added a sanity check for add_procs() during lazy get_ep.

PR #6023, opened 2018-11-03
VERSION: setting version for final 4.0.0 build on 11/5

gpaulsen:


bwbarrett:
We don’t do this by hand. Just tag with the VERSION as is and use the build utility to build the release version of the tarballs, similar to the rcs

gpaulsen:
I removed the date change and repushed.  @bwbarrett, @hppritcha please re-review.

gpaulsen:
If I understand correctly, in the future we should not modify VERSION at all and just use correctly named tags and the jenkins tarball builder.  That would be much easier (and more correct as a tag tags exactly one commit).

Since we've been using the greek field in VERSION for v4.0.0rc's I'd still like to remove the rc5 from the greek field in this PR for our final v4.0.0 build.  Sound reasonable @bwbarrett?

bwbarrett:
@gpaulsen, no, we never commit the VERSION file with an empty greek field.  Look at the value in any of the releases on the 2.x or 3.x series.  We do keep it up to date across the RCs, as you've been doing to date.

gpaulsen:
okay, closing this, thanks.

PR #6027, opened 2018-11-03
v4.0.0: text updates

jsquyres:
This is the v4.0.x version of #6026.

Also need to pull the contents of #6028 into this PR.

PR #6028, opened 2018-11-03
README: More updates for v4.0.0

jsquyres:
Move the UCX and MXM text up to flow better with the rest of the text / context.

@yosefe Turns out that I had more changes to make w.r.t. the UCX / MXM text, anyway.  😄

### DIFF
--- a/README
+++ b/README
@@ -641,11 +641,11 @@ OpenSHMEM Collectives
 Network Support
 ---------------
 
-- There are four main MPI network models available: "ob1", "cm",
-  "yalla", and "ucx".  "ob1" uses BTL ("Byte Transfer Layer")
+- There are several main MPI network models available: "ob1", "cm",
+  "ucx", and "yalla".  "ob1" uses BTL ("Byte Transfer Layer")
   components for each supported network.  "cm" uses MTL ("Matching
-  Transport Layer") components for each supported network.  "yalla"
-  uses the Mellanox MXM transport.  "ucx" uses the OpenUCX transport.
+  Transport Layer") components for each supported network.  "ucx" uses
+  the OpenUCX transport.
 
   - "ob1" supports a variety of networks that can be used in
     combination with each other:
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -679,31 +679,64 @@ Network Support
...
+  While the openib BTL depended on libibverbs, the UCX PML depends on
+  the UCX library.
+
+  Once installed, Open MPI can be built with UCX support by adding
+  --with-ucx to the Open MPI configure command. Once Open MPI is
+  configured to use UCX, the runtime will automatically select the UCX
+  PML if one of the supported networks is detected (e.g., InfiniBand).
+  It's possible to force using UCX in the mpirun or oshrun command
+  lines by specifying any or all of the following mca parameters:
+  "-mca pml ucx" for MPI point-to-point operations, "-mca spml ucx"
```
yosefe: guess need to use "--mca" here as well (instead of -mca)?

README
```diff
@@ -641,11 +641,11 @@ OpenSHMEM Collectives
 Network Support
 ---------------
 
-- There are four main MPI network models available: "ob1", "cm",
-  "yalla", and "ucx".  "ob1" uses BTL ("Byte Transfer Layer")
+- There are several main MPI network models available: "ob1", "cm",
+  "ucx", and "yalla".  "ob1" uses BTL ("Byte Transfer Layer")
```
yosefe: pls remove mentions of mxm/yalla

README
```diff
@@ -679,31 +679,64 @@ Network Support
...
+
+    UCX currently supports:
+
+    - OpenFabrics Verbs (including InfiniBand and RoCE)
+    - Cray's uGNI
+    - TCP
+    - Shared memory
+    - NVIDIA CUDA drivers
+
+  - "yalla" uses the Mellanox MXM transport library.  MXM is the
```
yosefe: pls remove this section

README
jsquyres: Good catch; will fix.

README
jsquyres: Wait; you just told me in an email that Mellanox wants to keep the MXM support in v4.0.0.

So I'm confused: do you want MXM support in 4.0.0 or not?

yosefe: can we keep it in the code base for now, but not document it in README, or document it as deprecated?

jsquyres: Ah, I see.

Yes, I think we should mark it as deprecated.  If we leave it there undocumented, someone *will* go poking around and use it "just because it's there".  If we at least have it documented as deprecated, that will actually deter most people from using it.  I'll update the README and the web site PR (https://github.com/open-mpi/ompi-www/pull/117).

jsquyres: @yosefe Is ikrit deprecated, too?

yosefe: yes

jsquyres:
New README text pushed (including a discussion about default of PML/etc. selection) -- everyone please have a read.

README
```diff
@@ -668,42 +668,93 @@ Network Support
...
+
+  The "ikrit" OpenSHMEM network model is also available, but is
+  deprecated; it uses the deprecated Mellanox Message Accelerator
+  (MXM) library.
+
+- In prior versions of Open MPI, InfiniBand and RoCE support was
+  provided through the openib BTL and ob1 PML plugins.  Starting with
+  Open MPI 4.0.0, InfiniBand support through the openib+ob1 plugins is
+  both deprecated and superseded by the ucx PML component.
+
```
hjelmn: Not entirely true. ob1 use is not deprecated, just btl/openib. There are cases where pml/ob1 + uct,vader,self will beat pml/ucx. Same is true for osc/rdma + btl/uct vs osc/ucx.

bosilca: Good catch. However this is for 4.0 and I was under the impression that uct BTL + OB1 is not yet ready for primetime (outside master) ?

jsquyres: After some discussion on the webex, I'll remove the `+ob1` part of the clause.

jsquyres: Fixed in #6027; will cherry pick all the #6027 commits over here to this v4.0.x branch PR when it is merged to master.

PR #6032, opened 2018-11-05
README: updating LSF version supported to 9.1.1 or later

gpaulsen:
(cherry picked from commit 010059589877a4e7985a1c4daa86fd74ea840ab0)


rhc54:
Since you changed the README, can we assume there is a corresponding change to configure.m4?? Otherwise, please do so - we shouldn't build against earlier versions of LSF if we are simultaneously saying we don't support them.

Issue #6014, opened 2018-11-01
Compiler bug? / potential hang up issue with vader btl


t-niemi:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

3.0.3, 3.1.3, 4.0.0rc5

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Downloaded source tarball, compiled with gcc 5.4.0 using ./configure --prefix=... & make all install

### Please describe the system on which you are running

* Operating system/version: ubuntu 16.04, 18.04
* Computer hardware: x86_64
* Network type: shared memory

-----------------------------

## Details of the problem

Following the recent updates to vader btl, I tested 3.0.3, 3.1.3 and 4.0.0rc5 to check that the segfault problems are resolved. 3.0.3 seemed to work fine, but for some reason 3.1.3 and 4.0.0rc5 seemed to hang almost immediately in all test cases. I was able to reduce the problem to the following minimal test code:

```c
#include <mpi.h>

int main(int argc, char** argv) {
  int rounds = atoi(argv[1]);
  int i;
  MPI_Init(NULL, NULL);
  
  for (i = 0; i < rounds; i++) {
    MPI_Barrier(MPI_COMM_WORLD);
  }
  
  MPI_Finalize();
}
```

This code reliably hangs when rounds > 100 and np 2 or more.

As far as I understand, 3.0.3, 3.1.3 have pretty much identical code regarding the vader-part. However, for some reason, 3.0.3 works while 3.1.3 does not. I tried to compile a debug version of 3.1.3 with --enable-debug, but that one worked, so the problem could be related to some compiler optimization. 

On another machine I have gcc 4.9.2 and it showed the same behaviour. On a third machine I have gcc 7.3.0, but it seemed to work ok.

As a conclusion, it seems that if older gcc (5.4.0 or 4.9.2) is used in release mode, the vader part can hang up. With newer gcc (eg. 7.3.0) things seem to work.

t-niemi:
I also tried to play around with btl_vader_fbox.h and making the following change

```c
static inline mca_btl_vader_fbox_hdr_t mca_btl_vader_fbox_read_header (mca_btl_vader_fbox_hdr_t *hdr)
{
    /*mca_btl_vader_fbox_hdr_t tmp = {.data_i32 = {.value1 = hdr->data_i32.value1}};;
[...]

PR #6033, opened 2018-11-05
opal/asm: work around possible gcc compiler bug

hjelmn:
It seems in some cases (gcc older than v6.0.0) the __atomic_thread_fence is a
no-op with __ATOMIC_ACQUIRE. This appears to be the case with X86_64 so go
ahead and use __ATOMIC_SEQ_CST for the x86_64 read memory barrier. This should
not cause any performance issues as it is equivalent to the memory barrier
in the hand-written atomics.

References #6014


hjelmn:
:bot:ompi:retest

bwbarrett:
bot:ompi:retest

PR #6057, opened 2018-11-07
v3.1.x: pmix2x: add PMIX_EXPORT decoration to mca_psensor_file_component

### LINKED ISSUE #6056: undefined symbol: mca_psensor_file_component

3.1.3

release tarball, `./configure && make && make install`

* Operating system/version: CentOS 6.10
* Computer hardware: various
* Network type: IB

-----------------------------

## Details of the problem

Launching an program produces these warnings:
```
[raijin7:07223] pmix_mca_base_component_repository_open: "mca_psensor_file" does not appear to be a valid psensor MCA dynamic component (ignored): /apps/openmpi/3.1.3/lib/pmix/mca_psensor_file.so: undefined symbol: mca_psensor_file_component. ret -1
```
There's one from each MPI process and 1 for the `mpirun`. The same happens regardless of whether we use the internal pmix distribution of an external build of PMIx 2.1.4.

I think the problem is a missing `PMIX_EXPORT` on https://github.com/open-mpi/ompi/blob/17040631627b760efe3da95b4fc75eaf1d1aca4c/opal/mca/pmix/pmix2x/pmix/src/mca/psensor/file/psensor_file.h#L32 I'm just testing this at the moment; if it fixes it, can I PR into the 3.1.x branch? (I can't PR it to master since these files are branch-specific.)

benmenadue:
`mca_psensor_file_component` needs to be decorated with `PMIX_EXPORT` so that it can be found by `dlsym` when opening the component.

Refs #6056.

### DIFF
--- a/opal/mca/pmix/pmix2x/pmix/src/mca/psensor/file/psensor_file.h
+++ b/opal/mca/pmix/pmix2x/pmix/src/mca/psensor/file/psensor_file.h
@@ -29,7 +29,7 @@ typedef struct {
     pmix_list_t trackers;
 } pmix_psensor_file_component_t;
 
-extern pmix_psensor_file_component_t mca_psensor_file_component;
+PMIX_EXPORT extern pmix_psensor_file_component_t mca_psensor_file_component;
 extern pmix_psensor_base_module_t pmix_psensor_file_module;
 
 


benmenadue:
@jsquyres Actually, this doesn't need to go to master, since that's tracking upstream PMIx. This is a PR to the v3.1 branch, which uses PMIx 2.1.4. What's interesting is why my build of 3.0.3 doesn't have the same issue, even though it also uses PMIx 2.1.4...

@rhc54 It's missing from the v2.1 branch in pmix, so I just PR the change in there (pmix/pmix#931).

jsquyres:
ok to test

PR #6059, opened 2018-11-07
v3.0.x: pmix2x: add PMIX_EXPORT decoration to mca_psensor_file_component

benmenadue:
`mca_psensor_file_component` needs to be decorated with `PMIX_EXPORT` so that it can be found by `dlsym` when opening the component.

Refs #6056 (but for v3.0.x).


jsquyres:
ok to test

PR #6061, opened 2018-11-08
orte-rmaps-base: update out-of-slots show_help message

jsquyres:
Update the show_help message for when there are not enough slots to
run an application.

Also, remove a bunch of copies of this message in various show_help
text files that aren't used/referred to anywhere in the code.

### DIFF
--- a/orte/mca/rmaps/base/help-orte-rmaps-base.txt
+++ b/orte/mca/rmaps/base/help-orte-rmaps-base.txt
@@ -10,7 +10,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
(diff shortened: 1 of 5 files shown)


orte/mca/rmaps/base/help-orte-rmaps-base.txt
```diff
@@ -23,12 +23,35 @@
...
+  2. The --host:N command line parameter (N defaults to 1 if not
+     provided)
+  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
+  4. If neither a hostfile, the --hosts command line parameter, nor an
+     RM is present, Open MPI defaults to the number of processor
+     cores
+
+In all the above cases, if you want Open MPI to default to the number
+of hardware threads instead of the number of processor cores, use the
+--use-hwthread-cpu option.
```
rhc54: should be "--use-hwthread-cpus"

PR #6067, opened 2018-11-08
v4.0.x: README: Make LSF text more accurate

jsquyres:
Also remove a now-outdated LSF reference.

(cherry picked from commit 419852ab433e578a790e8882711a20f2f570f0a2)

### DIFF
--- a/README
+++ b/README
@@ -115,7 +115,7 @@ General notes
 - The run-time systems that are currently supported are:
   - rsh / ssh
   - PBS Pro, Torque
-  - Platform LSF (v9.1.1 and later)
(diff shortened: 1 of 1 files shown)


gpaulsen:
I've been discussing this with Ralph, and apparently there are all sorts of thorns around what to do for this.

This change looks good to me, but we should discuss the right thing to do here for our README.

We don't KNOW that older LSF versions won't work, but IBM doesn't want to support older than 5 year old versions.
The NOTE might be useful if someone who DOES wish to compile with OLD LSF, if it happens to work.  The NOTE is there in older v3.x so that should be a good enough place for the linking info.

PR #6073, opened 2018-11-12
OSC: set UCX module used by default

hoopoepg:
- OSC/UCX module set priority to 200 to be used by default

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_component.c
+++ b/ompi/mca/osc/ucx/osc_ucx_component.c
@@ -106,7 +106,7 @@ static int component_open(void) {
 
 static int component_register(void) {
     char *description_str;
-    mca_osc_ucx_component.priority = 0;
+    mca_osc_ucx_component.priority = 200;
     opal_asprintf(&description_str, "Priority of the osc/ucx component (default: %d)",
              mca_osc_ucx_component.priority);
     (void) mca_base_component_var_register(&mca_osc_ucx_component.super.osc_version, "priority", description_str,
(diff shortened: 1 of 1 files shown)


yosefe:
@jladd-mlnx @xinzhao3 pls bless

yosefe:
We reduced the priority of OSC/UCX because of missing functionality (emulation layer) in UCX.
Now that is available in UCX master (for v1.5 release) so we want to raise the priority

rhc54:
Given that it won't be the correct priority for those with earlier versions of UCX, I'd suggest having a configure test for UCX version and then set the priority accordingly.

hoopoepg:
@yosefe what do you think? add version check?

yosefe:
let's add runtime version check

hoopoepg:
bot:retest

hoopoepg:
bot:retest

hoopoepg:
@yosefe ok to merge?

yosefe:
@hoopoepg let's port to v4.0.x too

hjelmn:
Umm, ok. You should have asked for my review before increasing the priority above 101. osc/rdma is still faster in most cases and is only enabled if btl/uct is explicitly enabled.

yosefe:
@hjelmn @jsquyres would 60 be a good priority for ucx?

hjelmn:
Yes, that would be better. Just need it higher than osc/pt2pt. If/when btl/uct becomes available by default we need to revisit and figure out how to handle the relative priorities of the osc/rdma and osc/ucx. That will likely be in the Open MPI v5.0.0 timeframe. I still need to make sure when btl/uct is enabled but not used it doesn't slow any other ucx components down or use up any resources.

PR #6080, opened 2018-11-14
configure: fix lustre detection

ggouaillardet:


ggouaillardet:
@edgargabriel could you please review this ?

also, do you have any recollection why we `OPAL_LOG_COMMAND()` instead of a simpler `AC_COMPILE_IFELSE()` ?

`OPAL_CHECK_PACKAGE()` now automatically checks `<dir>/lib64` and `<dir>/lib` so I think there is also room to simplify that part.

edgargabriel:
@ggouaillardet sorry for the delay: the OPAL_LOG_COMMAND was not introduced by me, if I recall correctly @jsquyres added a fix sometime in spring, maybe it stems from his commit.

roblatham00:
I'll only point out that ROMIO does this differently (not to say correctly/incorrectly).  Autotools guidelines say header files and libraries should be passed in as part of CFLAGS, LIBS, and LDFLAGS, and the --with-whatever argument is for... I don't know something else.  I find the distinction annoying, personally, but have tried to stick to "the autotools way".

roblatham00:
from personal experience,  `AC_COMPILE_IFELSE` is not only simpler but has solved or preempted more than one portability issue for MPICH/ROMIO.

jsquyres:
@roblatham00 `OPAL_CHECK_PACKAGE` ultimately invokes things like `AC_CHECK_HEADERS` and `AC_SEARCH_LIBS` -- we don't try to roll that stuff ourselves.  `OPAL_CHECK_PACKAGE` is just a wrapper around checking for multiple things in a package: header files, libraries, ...etc.  We've been using it for years.  Example:

```
          [OPAL_CHECK_PACKAGE([$2],
                              [netlink/netlink.h],
                              [nl],
                              [nl_connect],
                              [-lm],
                              [$1],
                              [],
                              [opal_libnlv1_happy=1],
                              [opal_libnlv1_happy=0])],
```

Which looks for `netlink/netlink.h`, and looks for the `nl_connect` symbol in `-lnl`, and also adds `-lm` to `$LIBS` before checking for that.  I forget what `$6` and `$7` are for offhand, but they also influence what/how things are searched for.  `$8` is what to do if everything is found; `$9` is what to do if anything is not found.

And as for the difference between `AC_COMPILE_IFELSE` and `OPAL_LOG_COMMAND`, we added the `OPAL_LOG_COMMAND` macro several years ago as a way to make sure that the command that is run (and its output) are emitted to `config.log`.  In certain cases, this is extraordinarily helpful (e.g., because `AC_COMPILE_IFELSE` output is not emitted to `config.log`).  Hence, `OPAL_LOG_COMMAND` is not a replacement for `AC_COMPILE_IFELSE` -- rather, it's just a different tool that can be used (and isn't limited to just compiling things -- you can run *any* command).

ggouaillardet:
:bot:ompi:retest

AboorvaDevarajan:

awlauria:
bot:ompi:retest

Issue #6075, opened 2018-11-13
v4.0.0 and libfabric-devel-1.5.3-1.el7.x86_64 on Intel OPA1 systems don't mix


hppritcha:
Turns out that the current OFI MTL - which in the v4.0.0 has higher priority than the PSM2 MTL - gets confused if built/run against the libfabric distributed in  libfabric-devel-1.5.3-1.el7.x86_64, resulting in flame out of things like IMB.  

```
[gr0540.localdomain:180658] mca:base:select: Auto-selecting mtl components
[gr0540.localdomain:180658] mca:base:select:(  mtl) Querying component [ofi]
[gr0540.localdomain:180658] mca:base:select:(  mtl) Query of component [ofi] set priority to 25
[gr0540.localdomain:180658] mca:base:select:(  mtl) Querying component [psm2]
[gr0540.localdomain:180658] mca:base:select:(  mtl) Query of component [psm2] set priority to 10
[gr0540.localdomain:180658] mca:base:select:(  mtl) Selected component [ofi]
[gr0540.localdomain:180658] mca: base: close: component psm2 closed
[gr0540.localdomain:180658] mca: base: close: unloading component psm2
[gr0540.localdomain:180658] select: initializing mtl component ofi
[gr0540.localdomain:180659] mca:base:select: Auto-selecting mtl components
[gr0540.localdomain:180659] mca:base:select:(  mtl) Querying component [ofi]
[gr0540.localdomain:180659] mca:base:select:(  mtl) Query of component [ofi] set priority to 25
[gr0540.localdomain:180659] mca:base:select:(  mtl) Querying component [psm2]
[gr0540.localdomain:180659] mca:base:select:(  mtl) Query of component [psm2] set priority to 10
[gr0540.localdomain:180659] mca:base:select:(  mtl) Selected component [ofi]
[gr0540.localdomain:180659] mca: base: close: component psm2 closed
[gr0540.localdomain:180659] mca: base: close: unloading component psm2
[gr0540.localdomain:180659] select: initializing mtl component ofi
[gr0540.localdomain:180658] mtl_ofi_component.c:315: mtl:ofi:provider_include = "(null)"
[gr0540.localdomain:180658] mtl_ofi_component.c:318: mtl:ofi:provider_exclude = "shm,sockets,tcp,udp,rstream"
[gr0540.localdomain:180658] mtl_ofi_component.c:347: mtl:ofi:prov: psm2
[...]

PR #6082, opened 2018-11-14
MTL/OFI: Lower priority when all procs are local

matcabral:
So far Vader is faster than OFI MTL for doing shared memory.
Therefore, let it run by default when all procs are local.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -116,8 +116,21 @@ ompi_mtl_ofi_component_register(void)
     int ret;
     mca_base_var_enum_t *new_enum = NULL;
     char *desc;
+    int num_total_procs = (int)ompi_process_info.num_procs;
+    /* num_local_peers does not include us in
+     * its calculation, so adjust for that */
+    int num_local_procs = (int)(1 + ompi_process_info.num_local_peers);
 
-    param_priority = 25;   /* for now give a lower priority than the psm mtl */
+    /* Lower priority when all porcesses are loal (vader is faster for shm).
+     * However, if running only one process assume it is ompi_info or this
+     * is most likely going to spawn.
+     */
+    if ((num_local_procs == num_total_procs) && (1 < num_total_procs)) {
+        param_priority = 10;
+    } else {
+        param_priority = 25;   /* for now give a lower priority than the psm mtl */
+    }
+
     mca_base_component_var_register(&mca_mtl_ofi_component.super.mtl_version,
                                     "priority", "Priority of the OFI MTL component",
                                     MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,


matcabral:
solves #6075

bwbarrett:
I'm no sure how a patch to punt to OB1 is the right fix for an issue with the PSM provider in libfabric, but I didn't read closely.  The big problem with this patch is that it effectively kills the dynamic process use case with the MTL.  OB1 will be selected in the case all procs are local, then any connect/spawn must use OB1.

matcabral:
Hi @bwbarrett , In #6075 there are two different issues. This is not a solution for problem 1), which is a bug in and older version of the PSM2 provider reporting CQ_DATA support when some APIs didn't support it. Solved in v1.6. I don't think it should be fixed inside OMPI.
Problem 2), the root comes form https://github.com/open-mpi/ompi/blob/master/ompi/mca/mtl/psm2/mtl_psm2_component.c#L246 , which was originally wrote to avoid "reserving" HW resources in the HFI card when all process are local, but then got extended to lower priority in favor of OB1. However, the side effect is that now OFI MTL is selected (in builds that have both). So, OFI needs to mimic that behavior to avoid being selected(*). An alternative solution "this" patch is to remove the priority lowering logic in the PSM2 MTL (keeping the device thing).  Dynamic process limitation, yes agreed. Little mitigation by not lowering the P if only one proc is running. 


(*) back again in the field of having the MTL with OFI provider driven priority...

bwbarrett:
I guess my problem with this patch is that it assumes a behavior that is right for psm, but not right for other providers.  Perhaps the right thing to do is record the provider selected and only do this logic if the provider is psm/psm2?

hppritcha:
This isn't perfect but its okay for medium term..  At this time, I'm 99% sure we don't have an OFI libfabric provider that can beat vader's latency, at least on x86_64.   I understand this is not a fix for issue #6075.

bwbarrett:
Why did this patch get merged?  It’s not what we want for dynamic process cases. For those of us without a BTL (Amazon EFA), this breaks a use case for no tangible benefit.

matcabral:
This was agreed on the conf call on Tuesday with no objections, based on the fact that no OFI provider is faster than Vader for intranode communication.

bwbarrett:
I was on this Tuesday’s call and this was not discussed. I had objections above and am an OFI user. Please revert this commit. It should not have been pushed.

matcabral:
Sure, I can revert the commit to avoid the conflict. Just for the record, I talked about this PR on the conf call and there were no objections.

matcabral:
reverted in #6142

matcabral:
Hi @bwbarrett, the mca option to control lowering the MTL priority based on the provider is not so straight forward: the priority for the component must be defined before it is queried (_ompi_mtl_ofi_component_query_). However, the MTL will not know about providers until _fi_getinfo_ is called during _init_ (much after the component priorities were defined) and the component was already selected). I see this falls under the topic of "Dynamic MTL priority" based on the providers found/selected. I ran a quick POC moving the provider selection logic to  _ompi_mtl_ofi_component_register_ and works. However, I prefer to spend more time designing that change as a feature rather than rushing it to fix a bug.  Thanks,

PR #6108, opened 2018-11-25
pml_ucx: initialize req_mpi_object.comm for error handler

### LINKED ISSUE #6078: segfault in ompi_errhandler_request_invoke
One of our users triggered a segfault while calling `MPI_Waitall`. Currently running with OpenMPI 3.1.2, this is the stack saved in a coredump:

```
(gdb) bt
#0  0x00007faac3ca7547 in ompi_errhandler_request_invoke () from /usit/abel/u1/marcink/software/openmpi.gnu/3.1.2/lib/libmpi.so.40
#1  0x00007faac3ce5921 in PMPI_Waitall () from /usit/abel/u1/marcink/software/openmpi.gnu/3.1.2/lib/libmpi.so.40
#2  0x000000000042f06a in puMigrate ()
#3  0x0000000000449574 in mccMode ()
#4  0x0000000000402ccd in main () at src/main.c:45
```

I've tried to debug a little, and the segfault happens because `comm` is `NULL` when calling the following line:

```
    switch (type) {                                                                                                                                                                                         
    case OMPI_REQUEST_PML:                                                                                                                                                                                  
        return ompi_errhandler_invoke(mpi_object.comm->error_handler,                                                                                                                                       
                                      mpi_object.comm,                                                                                                                                                      
                                      mpi_object.comm->errhandler_type,                                                                                                                                     
                                      ec, message);                                                                                                                                                         

```
So `mpi_object.comm->` de-references a `NULL` pointer. `message` simply says `MPI_Waitall`. The segfault happens repeatedly, at the same time on quite a few (~40) ranks.

There could be some problem with the user's code, which triggers the error. I'm not entirely sure of course if user's application doesn't cause any memory corruption, but in either case I guess OpenMPI should not segfault in the error handler. 

Anyone has ideas as to why this can happen?

yosefe:
Fixes #6078

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -477,11 +477,12 @@ int mca_pml_ucx_irecv_init(void *buf, size_t count, ompi_datatype_t *datatype,
     PML_UCX_TRACE_RECV("irecv_init request *%p=%p", buf, count, datatype, src,
                        tag, comm, (void*)request, (void*)req);
 
-    req->ompi.req_state = OMPI_REQUEST_INACTIVE;
-    req->flags          = 0;
-    req->buffer         = buf;
-    req->count          = count;
-    req->datatype.datatype = mca_pml_ucx_get_datatype(datatype);
+    req->ompi.req_state           = OMPI_REQUEST_INACTIVE;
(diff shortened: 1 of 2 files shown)


yosefe:
@hoopoepg can you pls take a look?

PR #6118, opened 2018-11-26
Re-add removed deprecate-only MPI-2.0 symbols

bertwesarg:
See #6114


hjelmn:
Yeah, a little surprised I didn't catch this before I merged the branch to remove the actually removed symbols. At one point I wanted to go a step further and make the deprecated symbols go away too but that was rejected :).

PR #6122, opened 2018-11-27
mpi/c: add back (some more) deprecated subroutines

ggouaillardet:
- MPI_NULL_DELETE_FN
 - MPI_NULL_COPY_FN
 - MPI_DUP_FN


hjelmn:
Good catch. Thanks for that. Users shouldn't be using any of these but the forum is never going to do the right thing and remove them.

Issue #6130, opened 2018-11-28
MPI spawn and "not enough slots" issue


iassiour:
Hi

I am getting "not enough slots" issues with MPI_Comm_Spawn in latest ompi 4.0.0. 
The issue can be reproduced if a slave is spawned on the host that the master runs.

I have set the oversubscribe option. (rmaps_base_oversubscribe = 1 in openmpi-mca-params.conf)

but this seems to be ignored for MPI_comm_spawn. 

Is there a way to specify oversubscribe for mpi spawn or disable the slots mechanism altogether.

Regards

rhc54:
```ompi_info``` is your friend:
```shell
$ ompi_info --level 9 --param rmaps base
.
.
.
         MCA rmaps base: parameter "rmaps_base_inherit" (current value:
                          "false", data source: default, level: 9 dev/all,
                          type: bool)
                          Whether child jobs shall inherit launch directives
                          Valid values: 0: f|false|disabled|no|n, 1:
                          t|true|enabled|yes|y
```
So you need to add:
```
rmaps_base_inherit = 1
```
to your param file

iassiour:
Hi,

I think that does not work. The following program would still complain about slots in the second call to MPI_comm_spaen even if I have added

rmaps_base_oversubscribe = 1
rmaps_base_inherit = 1

in the openmpi-mca-params.conf file

`  mpistat = MPI_Info_create(&minfo);
  mpistat = MPI_Info_set(minfo, "add-host", "host26");
  mpistat = MPI_Comm_spawn(slavejobtospawn, MPI_ARGV_NULL,
           1, minfo, 0, MPI_COMM_WORLD, &wcomm_, MPI_ERRCODES_IGNORE);

  int value = 500;

  mpistat = MPI_Send(&value, 1, MPI_INT, 0, 0, wcomm_);
  MPI_Status MPIerr;
  mpistat = MPI_Recv(&value, 1, MPI_INT, 0, 0, wcomm_, &MPIerr);

  mpistat = MPI_Info_create(&minfo2);
  mpistat = MPI_Info_set(minfo2, "add-host", "h26");
   mpistat = MPI_Comm_spawn(slavejobtospawn, MPI_ARGV_NULL,
           1, minfo2, 0, MPI_COMM_WORLD, &wcomm2_, MPI_ERRCODES_IGNORE);`

Thanks

iassiour:
Hi,

I think that the slots issue is not limited to mpi spawn

[...]

PR #6139, opened 2018-11-29
v4.0.x: Fix typo for rmaps_base_oversubscribe

rhc54:
Causes the MCA param to be ignored, while the cmd line option still
works.

Thanks to @iassiour for the report!


rhc54:
Fixes #6130

rhc54:
Well that's odd - the email from that issue were all tagged with @gn164. I simply copied it from there. I can easily change it.

jsquyres:
Looks like "gn164" is the name set on the Github account, but @iassiour is the github username.

gpaulsen:
thank you

PR #6140, opened 2018-11-30
Fix use of bitwise operation in CPP condition

bertwesarg:
Signed-off-by: Bert Wesarg <bert.wesarg@tu-dresden.de>

### DIFF
--- a/ompi/mpi/cxx/mpicxx.h
+++ b/ompi/mpi/cxx/mpicxx.h
@@ -44,7 +44,7 @@
 
 #include <stdarg.h>
 
-#if !defined(OMPI_IGNORE_CXX_SEEK) & OMPI_WANT_MPI_CXX_SEEK
+#if !defined(OMPI_IGNORE_CXX_SEEK) && OMPI_WANT_MPI_CXX_SEEK
 // We need to include the header files that define SEEK_* or use them
 // in ways that require them to be #defines so that if the user
 // includes them later, the double inclusion logic in the headers will


hjelmn:
@jsquyres Do we have enough changes on master to require the next release be v5.0.0? If so we should remove the C++ bindings on master and this should go just to the releases.

bertwesarg:
@hjelmn @jsquyres I'm fine, if this goes only into branches, where it apply, but it is still not merged.

jsquyres:
WOW.  Somehow a) I totally missed this PR (thanks for the poke, @bertwesarg), and b) this bug has been there for *years*.

...actually, I just did the `git blame` and it's only been there since b9315edb854 (Sep 2017).  It might be worth checking which release branches this is in (v3.0.x, v3.1.x, v4.0.x).

I think we might as well fix this, even if the C++ bindings are going to go away sometime in the not-distant future (possibly Open MPI v5.0.0).  If nothing else, this gives us something to cherry-pick from for the release branches.

And yes, to answer @hjelmn's question, we do have enough on master to say that it breaks ABI / backwards compatibility, so it'll be v5.0.0 (not v4.1.x.).  See https://github.com/open-mpi/ompi/issues?utf8=%E2%9C%93&q=label%3A%22%F0%9F%98%B3+Backward+compat+break%22+ (although I thought there were more issues than that... OIC, I think there are more *pending* potential backwards compat fixes; see https://github.com/open-mpi/ompi/wiki/5.0.x-FeatureList).

I'll go ahead and merge this long-forgotten PR.  @bertwesarg could you make PRs for the relevant release branches?  Thanks!

bertwesarg:
> I'll go ahead and merge this long-forgotten PR. @bertwesarg could you make PRs for the relevant release branches? Thanks!

I did, but no one merges them.

jsquyres:
I just tagged them properly; we somehow totally missed them before, and since they weren't tagged, we continued to miss them.  😦 

They missed v3.0.4 / v3.1.4 (which might even get released today), but it looks like we'll still have a v3.0.5 / v3.1.5 someday (due to a `--host` / `--hostfile` ordering issue), so we can still get this fix in the last releases on the v3.0.x and v3.1.x series.

bertwesarg:
> I just tagged them properly; we somehow totally missed them before, and since they weren't tagged, we continued to miss them. frowning

So who should have tagged them then? I can't and all other outside contributors too. Maybe its possible to automatically set the _Target_ labels based on the target branch? If not, maybe search by the  `base:` [1] filter instead of these manual labels. And reinterpret the _Target_ labels as "This merge should also go into that one" (a "Target: master" label should not be needed anymore then).

[1] https://help.github.com/en/articles/searching-issues-and-pull-requests#search-by-branch-name

PR #6144, opened 2018-12-01
ucx: add owner.txt for components

### LINKED ISSUE #6119: UCX components missing owner.txt

## Background information

We require all components to have an owner.txt file giving the current owner and development status of the component.

All versions

## Details of the problem

The owner of pml/ucx and osc/ucx need to add owner.txt files to master. I do not know if this is required for release branches (probably not as we don't typically delete/heavily modify components from releases).

yosefe:
Fixes #6119


yosefe:
@jsquyres @hjelmn can you pls take a look?

PR #6030, opened 2018-11-05
mpiext: keep include path for extension short

ggouaillardet:
in order to cope with the 72 characters per line limit, move
openmpi/ompi/mpiext/FOO/mpif-h/mpiext_FOO_mpifh.h to
openmpi/ompi/mpiext/FOO_mpifh.h

Refs. open-mpi/ompi#6019

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>

PR #6145, opened 2018-12-01
mpiext: updates for header file locations

jsquyres:
Continuing the discussion from #6030:

For reference -- prior to #6030, MPI extension header files were installed like this:

```
<prefix>/include/openmpi/ompi/mpiext/pcollreq/mpif-h/mpiext_pcollreq_mpifh.h
<prefix>/include/openmpi/ompi/mpiext/pcollreq/c/mpiext_pcollreq_c.h
<prefix>/include/openmpi/ompi/mpiext/pcollreq/c/pmpiext_pcollreq_c.h
<prefix>/include/openmpi/ompi/mpiext/cuda/c/mpiext_cuda_c.h
<prefix>/include/openmpi/ompi/mpiext/affinity/c/mpiext_affinity_c.h
```

As of #6030, MPI extension header files were installed like this (specifically done to reduce the length of the Fortran header file make the `include ...` Fortran statement in the installed `mpif-ext.h` be less than 72 characters):

```
<prefix>/include/openmpi/ompi/mpiext/affinity_c.h
<prefix>/include/openmpi/ompi/mpiext/pcollreq_c.h
<prefix>/include/openmpi/ompi/mpiext/pcollreq_mpifh.h
<prefix>/include/openmpi/ompi/mpiext/ppcollreq_c.h
<prefix>/include/openmpi/ompi/mpiext/cuda_c.h
```

I think we got here because the original assumption was that the path in the source tree had to be exactly the same as the path in the install tree.  I do not think that that is a correct assumption (or: it may have been true back when the MPI extension system was written; it isn't true any more).

More generally: rather than think about this about "what can we make better (shorter)?", let's look at this through the lens of "what do we really want?"

I think that having a different *filename* in the source tree and installation tree would be... difficult.  But I think we can manage a different path.  How's this for a suggestion for "what do we want?":

```
<prefix>/include/openmpi/mpiext/affinity_c.h
<prefix>/include/openmpi/mpiext/pcollreq_c.h
<prefix>/include/openmpi/mpiext/pcollreq_mpifh.h
<prefix>/include/openmpi/mpiext/ppcollreq_c.h
<prefix>/include/openmpi/mpiext/cuda_c.h
```

Calculating that out:
* We have 72 chars to play with
* The `include` line in `<includedir>/mpif-ext.h` file takes this many characters:
    * spaces in front: 8
    * `include '`: 9
    * `openmpi/mpiexec/NAME_mpifh.h`: without NAME, it's 25
    * trailing `'`: 1
    * 8 + 9 + 25 + 1: 43
* That leaves us with (72-43)=29 characters for NAME.  If any extension name is over 29 characters, we have probably hit the apocalypse. 🙉 🙈 🙊 

This PR moves installing the MPI extension header files in a different path in the installation tree compared to their path in the source tree.  Since these header files do not include any other files, this is very do-able.

One more thing, though: I notice that we renamed the C and mpifh headers to ditch the `mpiext_` prefix.  But we left the `mpiext_` prefix on:

* All C source code
* All Fortran source code
* The `use-mpi` and `use-mpi-f08` header files

This is a bit of an odd discrepancy.

@ggouaillardet I think we *need* the prefix on the C/Fortran source code files so that the linker is guaranteed to have unique `.o` filenames for them in `libmpi` and friends (i.e., follow the prefix rule).  Were the `use-mpi` and `use-mpi-f08` header filenames left unaltered in #6030 for a reason?


ggouaillardet:
@jsquyres I did not initially rename the `use-mpi` and `use-mpi-f08` headers because they are not installed (e.g. they are only used internally). For the sake of consistency (and to fix this PR too) it is a good idea to rename them as well. I will shortly push a new commit you can squash with the first one.

ggouaillardet:
@jsquyres on second thought, we could keep the name `mpiext_FOO_LANG` with the new install directory. that would make the extension name limited to 22 characters (instead of 29) which should be far enough.

jsquyres:
Your new PR (#6158) is a good replacement for this one.  Closing without merging.

PR #6158, opened 2018-12-06
mpiext: updates for header file locations

ggouaillardet:


ggouaillardet:
This is a replacement of #6145. It basically reverts #6030, keeps the `mpiext_NAME_LANG.h` name for the header files, and install them in `<prefix>/include/openmpi/mpiext`. The comments and commit messages are yours.

There was some error in calculating a 29 characters max name for a MPI extension with mpif.h binding.
The new magic number is now `26` which should be far enough.

PR #6163, opened 2018-12-07
Refactoring of osc/ucx component for MT

artpol84:
This is the first take on osc/ucx optimization. Targeted optimizations will follow.


artpol84:
@xinzhao3 FYI

artpol84:
@xinzhao3 f73377c3649b was cherry-picked

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -0,0 +1,1345 @@
...
+        rc = OPAL_ERROR;
+        goto err_get_addr;
+    }
+
+    rc = _wpool_list_put(wpool, &wpool->idle_workers, winfo);
+    if (rc) {
+        goto err_wpool_add;
+    }
+
+    pthread_key_create(&wpool->tls_key, _tlocal_cleanup);
```
hjelmn: Why not use opal_thread_local instead of all this pthread stuff? If you need pthread tls then please use the opal tsd interface. Using pthreads directly is discouraged.

artpol84: Does it provide an option to incorporate TLS into a structure field as we do here?

artpol84: And more importantly, does it provides an option to install a cleanup callback.
I checked - it doesn't. Maybe it can be a good thing to implement this in OPAL level by using pthread_key if available and providing a custom implementation otherwise.
Though I'm afraid it could be tricky.

artpol84:
@xinzhao3 1867f83 was cherry-picked.

opal/mca/common/ucx/common_ucx_wpool.c
hjelmn: No, you can't use C11 TLS in a structure. You can, however, use opal/tsd to register a cleanup function for anything you put in C11 TLS. That way you still get fast loading of TLS but still get the teardown functionality. Also note that pthreads does not call the teardown function for the main thread so anything you allocate for it is leaked if you do not call the teardown for that context separately.

hjelmn:
Out of curiosity, why does the code for the worker pool live in opal? Are you intending to use the same code for oshmem at some point?

opal/mca/common/ucx/common_ucx_wpool.c
hjelmn: BTW, you still should not be calling pthreads directly. See opal/threads/tsd.h.

artpol84: > Also note that pthreads does not call the teardown function for the main thread so anything you allocate for it is leaked if you do not call the teardown for that context separately.

We cleanup all unreleased resources at the worker pool finalize (as we keep track of all TLS pointers there) [ [link](https://github.com/open-mpi/ompi/pull/6163/files#diff-9227c793e9006a6483ac7a7479b6bbe7R249) ]
We also considered the case where threads are still active when MPI_Finalize is called and we still have to cleanup.

> No, you can't use C11 TLS in a structure. You can, however, use opal/tsd to register a cleanup function for anything you put in C11 TLS. That way you still get fast loading of TLS but still get the teardown functionality of pthreads. 

So we will have to maintain some sort of the table and map the structures tere by the index. Is that correct?
This will probably be beneficial if something else instead of pthreads is used to implement SW treads.

>  You can, however, use opal/tsd to register a cleanup function for anything you put in C11 TLS.

In the wpool cleanup we currently call pthread_key_deregister and this ensures that if a therad exiting in parallel with MPI_Finalize call there will be no race condition between thread cleanup callback and `opal_common_ucx_wpool_finalize`. Does `opal/tsd` provides this kind of a guarantee?

artpol84:
Yes, definitely. We will update wpool accordingly to maintain a minimal level of locking required by OSHMEM spec and will use them there.
This will be a next step planned in early 2019.

opal/mca/common/ucx/common_ucx_wpool.c
artpol84: > Why not use opal_thread_local instead of all this pthread stuff? If you need pthread tls then please use the opal tsd interface. Using pthreads directly is discouraged.

In general I agree with this comment. Though using opal/tsd looks like a little more work: we have only one variable that can't be placed in the structure, so we will have to maintain a table to do the same. I don't see this hurting the performance (as we can get fast access by the index), but this will take some time to implement. I think it actually makes sense to implement this in opal/tsd as another convenience tool.

Implementation wise I'd prefer to merge this PR as is and address this comment in a separate PR later. Before that we would keep the code in the master and not placing it in the release branches.

artpol84:
@hjelmn I checked opal/tsd and it seems to be a 1-to-1 wrapper over pthread_key functionality.
So I would guess that one can use `opal_tsd_key_t` in the same way we use `pthread_key_t` now: i.e. place it as a field in the structure. 
It seems to me that we can just replace pthread_key's with `opal_tsd_key_t`'s to address your concern. Please confirm.

artpol84:
@jladd-mlnx I'm finishing up with @yosefe and will advise once it is done.
I'd guess it is better to wait till the next week. We are going to go over it face-to-face.

artpol84:
@xinzhao3 please double-check me with e6ec302

artpol84:
@bwbarrett @jsquyres The build faile for cray_cle for seemingly an unrelated reason.
Can you confirm? I'm about to merge this one.

jsquyres:
bot:ompi:retest

PR #6181, opened 2018-12-12
hwloc/base: fix hwloc_topology_set_xmlbuffer() usage

ggouaillardet:
pass the length of the xml buffer plus one in order to
include the NULL termination character.

The buffer is then sscanf'ed and a NULL terminated string
is expected here.

This fixes an out-of-bound error reported by valgrind.

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_dt.c
+++ b/opal/mca/hwloc/base/hwloc_base_dt.c
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2011-2012 Cisco Systems, Inc.  All rights reserved.
- * Copyright (c) 2017      Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2017-2018 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


bgoglin:
Your patch above is good for preventing against rogue buffer attacks, so I'll apply it anyway. Replacing sscanf would be easy, but I'd have to verify all the other nolibxml parsing routines as well, your patch is much easier.

It seems the exported buffer always contains the ending 0, but the reported size doesn't include it. So OMPI might miss the ending 0 if it transmits the reported size instead of size+1. I guess I'll increase the reported size by 1 and better document things.

Now I need to understand why I can't reproduce the issue. I tried called set_xmlbuffer() on a copy of the buffer that doesn't include the ending 0, valgrind didn't complain.

ggouaillardet:
FWIW I run on an up to date CentOS 7 box.

You should be able to reproduce the issue with `sscanf()`, just make sure you did not malloc space for the ending 0.
I do not know if this behavior depends on the length of the buffer, and/or the internals of the glibc.

jsquyres:
This fix was included in #6185 -- this PR is now moot.  Closing without merging.

PR #6185, opened 2018-12-12
hwloc: remove public hwloc macros from opal_config.h

ggouaillardet:


ggouaillardet:
@bgoglin FYI

The issue can be evidenced when building Open MPI with an external hwloc >= 2.0.1

Since the hwloc m4 are slurped into Open MPI, `HWLOC_VERSION` (among other) is now in `opal_config.h` regardless an external or the embedded hwloc is used.
This macro is also (indirectly) defined in the `hwloc.h` of the external hwloc (if >= 2.0.1) and this will issue tons of warnings.

The first commit is a possible fix (a bit ugly IMHO though).
An other option is __not__ to have these macros ending up in `opal_config.h` but I do not know how to achieve that, nor if it is even feasable.
Yet an other option would be to integrate `hwloc` as a truly independent package with its own `configure` (just like PMIx for example)

The second commit is a simple cleanup (the line addition was committed 10 years ago, and I cannot find any rationale for it)

jsquyres:
@ggouaillardet Per the comments on #6314, let's continue over here.  I just pushed some changes here:

1. Move the #undef's to `opal/mca/hwloc/external/external.h`
1. Split the `+1` fixes into a separate commit.

What do you think?

ibm-ompi:

ibm-ompi:

jjhursey:
IBM Ci had a network event - retrying bot:ibm:retest

bgoglin:
The +1 comes from #6181, which was never applied because we mostly discussed improvements to hwloc instead of that specific fix. But I think it's sane.

PR #6190, opened 2018-12-12
mtl/ofi: Fix MPI_ANY_SOURCE for i(m)probe in v3.1.x

mohanasudhan:
Fix invalid handling of MPI_ANY_SOURCE for i(m)probe
by using the any_source address handle rather than
the 0 address handle. Intel test cases
MPI_Probe_source_c and MPI_Probe_tag_c now passes.

Commit e667455 in master (which has the fix) was a
feature change and we are just backporting the bug fix
to 3.1.x and 3.0.x branches.

Signed-off-by: Mohan Gandhi <mohgan@amazon.com>


mohanasudhan:
Adding Brian to review list as the checklist says "At least 1 approving review is required by reviewers with write access"

Issue #6182, opened 2018-12-12
Scalable Endpoints broken when number of local processes greater than number of contexts


bwbarrett:
109d0569ffdc29f40518d02ad7a4d5bca3adc3d1 introducted scalable endpoints support for the OFI MTL.  It also causes a crash whenever the number of local processes is greater than the number of available contexts.  For example, using the sockets provider from 1.6.2:

```
% mpirun -npernode 17 -hostfile ~/h-c5n -mca mtl ofi -mca pml cm -mca mtl_ofi_provider_include sockets ./examples/ring_c
--------------------------------------------------------------------------
Reached limit (0) for number of OFI contexts that can be opened with the
provider. Creating new communicators beyond this limit is possible but
they will re-use existing contexts in round-robin fashion.
Using new communicators beyond the limit will impact performance.

  Local host: ip-172-31-30-126
  Location: mtl_ofi.h:1250
--------------------------------------------------------------------------
[ip-172-31-30-126:47221] *** Process received signal ***
[ip-172-31-30-126:47221] Signal: Segmentation fault (11)
[ip-172-31-30-126:47221] Signal code: Address not mapped (1)
[ip-172-31-30-126:47221] Failing at address: 0x38
```

The number of contexts is being computed in ompi_mtl_ofi_init_sep() in a way that doesn't seem to account for the case where number of local processes is greater than number of contexts, but the socket provider defaults to 16 contexts and were running on hosts with 18 cores, so we set `ompi_mtl_ofi.max_ctx_cnt` to 0 and blow up.

aravindksg:
Yep, I see the issue. It seems to me that for this case, we can simply fall back to using regular endpoints.  Shall work on a patch to fix this.

aravindksg:
@bwbarrett : This issue can now be closed. Could you please close it (I don't have the permissions)?

bwbarrett:
Verify this appears fixed.  Closing.

PR #6192, opened 2018-12-13
Fix for SEP when num local procs is greater than available contexts

aravindksg:
For cases when the number of local processes is greater than the number of
available contexts, the SEP initialization phase would calculate the number of
contexts to provision for each rank to be 0 and would eventually crash.

Fix the issue here by using regular endpoints in the event the number of local
processes is more than available contexts. This fixes issue #6182.


aravindksg:
@bwbarrett Tested this with sockets provider by using more local ranks than available contexts and it works fine

PR #6199, opened 2018-12-17
opal/sync: Fix assert during multi-threaded progress invocation

aravindksg:
PR #5241 provided an MCA variable to allow multi-threaded opal_progress.
However, it allowed to update the linked list even when multiple threads was
allowed to call opal_progress. This caused a scenario when a more recent thread
could complete it's progress and fail the assert(sync ==
wait_sync_list).

Allowing to update the linked list only for the case when the number of threads
exceeds the threshold fixes the problem.

### DIFF
--- a/opal/threads/wait_sync.c
+++ b/opal/threads/wait_sync.c
@@ -47,17 +47,19 @@ int ompi_sync_wait_mt(ompi_wait_sync_t *sync)
     }
 
     /* Insert sync on the list of pending synchronization constructs */
-    OPAL_THREAD_LOCK(&wait_sync_lock);
-    if( NULL == wait_sync_list ) {
-        sync->next = sync->prev = sync;
-        wait_sync_list = sync;
-    } else {
-        sync->prev = wait_sync_list->prev;
-        sync->prev->next = sync;
-        sync->next = wait_sync_list;
-        wait_sync_list->prev = sync;
+    if (num_thread_in_progress >= opal_max_thread_in_progress) {
+        OPAL_THREAD_LOCK(&wait_sync_lock);
+        if( NULL == wait_sync_list ) {
+            sync->next = sync->prev = sync;
+            wait_sync_list = sync;
+        } else {
+            sync->prev = wait_sync_list->prev;
+            sync->prev->next = sync;
+            sync->next = wait_sync_list;
+            wait_sync_list->prev = sync;
+        }
+        OPAL_THREAD_UNLOCK(&wait_sync_lock);
     }
-    OPAL_THREAD_UNLOCK(&wait_sync_lock);
 
     /**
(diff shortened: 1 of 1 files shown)


aravindksg:
@thananon , @bosilca for reviews

rhc54:
bot:ompi:retest

thananon:
I'm not sure we should remove the link list path as it might be useful to the next milestone of progress function (Tasks, smarter thread assignment). I suggest we should just remove the assert. @bosilca ?

aravindksg:
While removing the assert works, it degrades performance as we still need to take locks to update the list (multi-threaded BW was affected by as much as 2x in my testing with mtl/ofi and SEP). With the approach in this patch, we still don't remove the linked list right? We just update it only as needed (when the threshold is exceeded).

bosilca:
Here is my understanding of how the code will behave with this patch. Only the extra threads (aka over the ```opal_max_thread_in_progress```) will be added to the ```wait_sync_list``` list of "pending" threads. As a result the number of locks on the critical path will be decreased on all scenarios where the number of active threads (aka. threads doing progress) is below the ```opal_max_thread_in_progress```, and the management will instead be done with a single atomic update (the ```num_thread_in_progress```). All other cases will behave as before. If my understanding is correct this is a nice optimization for a common case. :+1:

aravindksg:
> Here is my understanding of how the code will behave with this patch. Only the extra threads (aka over the `opal_max_thread_in_progress`) will be added to the `wait_sync_list` list of "pending" threads. As a result the number of locks on the critical path will be decreased on all scenarios where the number of active threads (aka. threads doing progress) is below the `opal_max_thread_in_progress`, and the management will instead be done with a single atomic update (the `num_thread_in_progress`). All other cases will behave as before. If my understanding is correct this is a nice optimization for a common case. 👍

@bosilca , Yes, that's exactly right

opal/threads/wait_sync.c
```diff
@@ -89,20 +91,23 @@ int ompi_sync_wait_mt(ompi_wait_sync_t *sync)
     }
     OPAL_THREAD_ADD_FETCH32(&num_thread_in_progress, -1);
 
-    assert(sync == wait_sync_list);
+    if (NULL != wait_sync_list) {
+        assert(sync == wait_sync_list);
```
thananon: My code crashed with this assert. I think we have to rethink this PR. Here is the problem..
1. This assert is problematic because maybe you wake up more than 1 thread (more than 1 sync completed)
2. The check of `opal_max_thread_in_progress` to decide if we are going to update the list is open to race condition. (read after write on active thread count, etc)

thananon:
@abouteiller and I encountered the same problem.
@bosilca @aravindksg I'm going to revert this patch for now until we can find a better approach.

PR #6209, opened 2018-12-20
PML/UCX: suppressed multi-thread warning

hoopoepg:
- in case if UCX doesn't support for multi-thread mode
  then PML UCX fallback into single thread mode and
  reduce priority of component

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -454,6 +454,7 @@ int mca_pml_ucx_enable(bool enable)
     PML_UCX_FREELIST_INIT(&ompi_pml_ucx.convs,
                           mca_pml_ucx_convertor_t,
                           128, -1, 128);
+    ompi_pml_ucx.enabled = enable;
     return OMPI_SUCCESS;
 }
 
@@ -1019,5 +1020,7 @@ int mca_pml_ucx_dump(struct ompi_communicator_t* comm, int verbose)
 
 static void mca_pml_ucx_mpi_init_bottom(int argc, char **argv, int requested, int *provided)
 {
-    *provided = MPI_THREAD_SINGLE;
+    if (ompi_pml_ucx.enabled) {
+        *provided = MPI_THREAD_SINGLE;
+    }
 }
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -21,6 +21,7 @@
 #include "ompi/mca/pml/base/pml_base_bsend.h"
 #include "opal/mca/common/ucx/common_ucx.h"
 #include "pml_ucx_request.h"
+#include "ompi/mca/hook/base/base.h"
 
 #include <inttypes.h>
 
@@ -77,6 +78,13 @@ mca_pml_ucx_module_t ompi_pml_ucx = {
     },
     .ucp_context           = NULL,
     .ucp_worker            = NULL
+};
+
+static void mca_pml_ucx_mpi_init_bottom(int argc, char **argv, int requested, int *provided);
+
+static ompi_hook_base_component_t hook_ucx_extra_component = {
+    /* Component functions */
+    .hookm_mpi_init_bottom = mca_pml_ucx_mpi_init_bottom
 };
 
 #define PML_UCX_REQ_ALLOCA() \
@@ -188,7 +196,7 @@ int mca_pml_ucx_close(void)
     return OMPI_SUCCESS;
 }
 
-int mca_pml_ucx_init(void)
+int mca_pml_ucx_init(int enable_mpi_threads)
 {
     ucp_worker_params_t params;
     ucp_worker_attr_t attr;
@@ -199,8 +207,7 @@ int mca_pml_ucx_init(void)
 
     /* TODO check MPI thread mode */
     params.field_mask  = UCP_WORKER_PARAM_FIELD_THREAD_MODE;
-    params.thread_mode = UCS_THREAD_MODE_SINGLE;
-    if (ompi_mpi_thread_multiple) {
+    if (enable_mpi_threads) {
         params.thread_mode = UCS_THREAD_MODE_MULTI;
     } else {
         params.thread_mode = UCS_THREAD_MODE_SINGLE;
@@ -222,12 +229,14 @@ int mca_pml_ucx_init(void)
         goto err_destroy_worker;
     }
 
-    if (ompi_mpi_thread_multiple && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
-        /* UCX does not support multithreading, disqualify current PML for now */
-        /* TODO: we should let OMPI to fallback to THREAD_SINGLE mode */
-        PML_UCX_ERROR("UCP worker does not support MPI_THREAD_MULTIPLE");
-        rc = OMPI_ERR_NOT_SUPPORTED;
-        goto err_destroy_worker;
+    if (enable_mpi_threads && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
+        PML_UCX_VERBOSE(1, "UCP worker does not support MPI_THREAD_MULTIPLE, "
+                           "force MPI_THREAD_SINGLE");
+        /* reduce priority of PML UCX to minimal to allow other components
+         * to operate in multi-thread mode, add hook to modify provided
+         * thread model */
+        ompi_pml_ucx.priority = 1;
+        ompi_hook_base_register_callbacks(&hook_ucx_extra_component);
     }
 
     rc = mca_pml_ucx_send_worker_address();
@@ -268,6 +277,7 @@ int mca_pml_ucx_cleanup(void)
 
     PML_UCX_VERBOSE(1, "mca_pml_ucx_cleanup");
 
+    ompi_hook_base_deregister_callbacks(&hook_ucx_extra_component);
     opal_progress_unregister(mca_pml_ucx_progress);
 
     if (ompi_pml_ucx.datatype_attr_keyval != MPI_KEYVAL_INVALID) {
@@ -1005,4 +1015,9 @@ int mca_pml_ucx_start(size_t count, ompi_request_t** requests)
 int mca_pml_ucx_dump(struct ompi_communicator_t* comm, int verbose)
 {
     return OMPI_SUCCESS;
+}
+
+static void mca_pml_ucx_mpi_init_bottom(int argc, char **argv, int requested, int *provided)
+{
+    *provided = MPI_THREAD_SINGLE;
 }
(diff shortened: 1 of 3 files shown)


ibm-ompi:

 Gist: https://gist.github.com/f2e0e8f85ca60b3605273ec39f7c403b

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -225,7 +225,7 @@ int mca_pml_ucx_init(void)
     if (ompi_mpi_thread_multiple && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
         /* UCX does not support multithreading, disqualify current PML for now */
         /* TODO: we should let OMPI to fallback to THREAD_SINGLE mode */
-        PML_UCX_ERROR("UCP worker does not support MPI_THREAD_MULTIPLE");
+        PML_UCX_VERBOSE(1, "UCP worker does not support MPI_THREAD_MULTIPLE");
```
yosefe: so if UCX does not support multi-threading, another PML would be used?
Maybe we should keep using PML_UCX, but report a lower thread-support level to the SHMEM application?

yosefe: i don't think this is the behavior we want

hoopoepg: ok, fixed: it will fallback into single thread mode

manjugv:
@hoopoepg If I understand this right, one can use either PML/UCX or SPML/UCX and it will fallback to single thread mode, if multithreaded mode is not supported, correct ?

hoopoepg:
@manjugv right, MPI_Init_thread allows to fallback into single thread mode even for cases when multiple thread mode requested

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -222,12 +231,11 @@ int mca_pml_ucx_init(void)
         goto err_destroy_worker;
     }
 
-    if (ompi_mpi_thread_multiple && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
-        /* UCX does not support multithreading, disqualify current PML for now */
-        /* TODO: we should let OMPI to fallback to THREAD_SINGLE mode */
-        PML_UCX_ERROR("UCP worker does not support MPI_THREAD_MULTIPLE");
-        rc = OMPI_ERR_NOT_SUPPORTED;
-        goto err_destroy_worker;
+    if ((ompi_mpi_thread_requested == MPI_THREAD_MULTIPLE) &&
```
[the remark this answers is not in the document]
hoopoepg: in case if single thread is requested - then there is no reason to have UCX MT enabled (less build limitations and runtime optimizations)

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -193,14 +193,23 @@ int mca_pml_ucx_init(void)
...
+        rc = OMPI_ERROR;
+        goto err;
+    }
+
     /* TODO check MPI thread mode */
-    params.field_mask  = UCP_WORKER_PARAM_FIELD_THREAD_MODE;
-    params.thread_mode = UCS_THREAD_MODE_SINGLE;
-    if (ompi_mpi_thread_multiple) {
+    params.field_mask = UCP_WORKER_PARAM_FIELD_THREAD_MODE;
+    if (ompi_mpi_thread_multiple && (ctx_attr.thread_mode == UCS_THREAD_MODE_MULTI)) {
```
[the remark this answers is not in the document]
hoopoepg: didn't get... in case if multi-thread is requested - then check for MT is supported by UCX, and make decision about worker attributes. Actually MPI allows to fallback into single thread mode even in case if Multi-Thread is requested (actual mode is returned)

ompi/mca/pml/ucx/pml_ucx.c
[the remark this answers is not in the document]
yosefe: how does setting `ompi_mpi_thread_provided = MPI_THREAD_MULTIPLE` affect runtime optimizations?

ompi/mca/pml/ucx/pml_ucx.c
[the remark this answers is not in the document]
yosefe: `ctx_attr.thread_mode == UCS_THREAD_MODE_MULTI` does not indicate if MT is supported by UCX

ompi/mca/pml/ucx/pml_ucx.c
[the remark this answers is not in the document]
hoopoepg: single thread build may avoid some synchronization calls

ompi/mca/pml/ucx/pml_ucx.c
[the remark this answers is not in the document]
hoopoepg: hmmmmm... not good. will fix

ompi/mca/pml/ucx/pml_ucx.c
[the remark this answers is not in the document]
yosefe: but `ompi_mpi_thread_provided` does not affect UCX behavior

hjelmn:
I don't think this is the right way forward. The logic is going to have to get more complex. If thread multiple is requested and there are btls that can be used it should always fall back to ob1 unless pml/ucx was explicitly requested.

yosefe:
maybe if no MT-supporting pml was found on the first round, need to do another round and this time do not require MT support?

yosefe:
or reduce the priority?

jsquyres:
I agree with @hjelmn -- just because UCX doesn't support THREAD_MULTIPLE doesn't mean that you should reduce the thread supported level to SINGLE.  Yes, it's *allowed* by the spec, but you're doing a disservice to the user by automatically downgrading the thread level, especially given that  it's non-deterministic as to whether the UCX or OB1 PML will be checked first.

Perhaps we need multiple priorities -- one for MULTIPLE and one for NOT_MULTIPLE.  That way, if a user asks for MULTIPLE, selection can take the highest priority module on the MULTIPLE list.  And if there's nothing in MULTIPLE, then take the highest priority for NOT_MULTIPLE.

hoopoepg:
ok, what is priority rule for single thread PML?
what about "MULTI_THREAD" + 100?

yosefe:
@jsquyres @hjelmn We plan the following solution: each pml component would report *both* supported thread-level and priority. Then, ompi would select the highest-prio pml amongst the pml's with the best thread level. Then, the global variable of provided thread level would be set to the thread-level of that component.

hoopoepg:
@yosefe implemented bit simpler algo to select component: PML UCX itself set priority based on requested thread model and used init callback to update provided thread model. it allowed to minimize changes and don't touch other components

ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -1006,3 +1016,8 @@ int mca_pml_ucx_dump(struct ompi_communicator_t* comm, int verbose)
 {
     return OMPI_SUCCESS;
 }
+
+static void mca_pml_ucx_mpi_init_bottom(int argc, char **argv, int requested, int *provided)
+{
+    *provided = MPI_THREAD_SINGLE;
```
yosefe: do we assume that if UCX was not unloaded, it means it was selected as PML?

hoopoepg: right, OMPI unloads all non-selected PML's

hoopoepg: there is cleanup loop:
https://github.com/open-mpi/ompi/blob/ae2bda9aa045c4dca7f1cbd12f2a20a084336b60/ompi/mca/pml/base/pml_base_select.c#L206

yosefe: i think the fact that unused components are unloaded is more of an optimization, and we better not rely on that (is this still the case in static compilation?)
need a better way to know if ucx was selected
Maybe we can set thread support level when pml_ucx module is initialized?

hoopoepg: hmm, there is no other way to know is component selected or not (except component is closed), list of opened components is local in ```select``` function and we may be sure that unselected components are cleaned

yosefe: can we check if a ucx module was created?

hoopoepg: @yosefe done

yosefe:
@jsquyres @hjelmn can you pls take a look?

hoopoepg:
@jsquyres @hjelmn ping
are you ok with these changes?

hjelmn:
It is a holiday. Will take a look later in the week.

jsquyres:
I defer to @hjelmn's review.

hoopoepg:
@hjelmn could you look?
thank you

yosefe:
@hjelmn @jsquyres ping

jsquyres:
Do we know for a fact that patching up the thread level at the end of `ompi_mpi_init()` is sufficient / safe?  I.e., what about the rest of the OMPI infrastructure that initialized itself to be thread safe?  It seems kinda weird that the UCX PML will be operating in a thread single mode, but the rest of OMPI will be operating in a thread multi mode.

@bosilca Thoughts?

hoopoepg:
actually there is no way to update supported thread level except callback

jsquyres:
@hoopoepg Yes, I know -- I think that was actually part of the initial design.  I don't think it was designed for the components to *alter* the thread level -- they could only *react* to the thread level.

This is why I think there may be a bunch of unintended consequences of *changing* the thread level after all of `ompi_mpi_init()` has already executed.

hoopoepg:
I see

So, what is the best solution? Disable component of it can't fit required thread level?

jsquyres:
Ya, that's what has traditionally been done.

I don't think we've ever discussed a solution that changes the thread level *after* the entire OMPI infrastructure has been setup (i.e., oriented to one thread level).

I'm not sure, but it sounds kinda hard to change the entire thread level infrastructure after the fact.  More investigation would need to occur to determine

* what is safe and what is not.  E.g., are there any consequences to the infrastructure being setup for MULTIPLE but then the global variable changes to SINGLE?
* what may just result in [slightly?] lower performance.  E.g., if the infrastructure uses locks because it is setup for MULTIPLE, but the process actually ends up running in SINGLE (and ostensibly does not need the locks).

It might be a better approach to update the selection mechanism.  E.g., go through all the components at MULTIPLE.  If none are found, reduce the thread level (safely, ...somehow?) and try again.

gpaulsen:
What's the status of this PR?  It hasn't been updated in almost 3 months.

I'd think a better approach would be to have the component disable itself if it cannot support the requested thread level.

hoopoepg:
@gpaulsen we will disable pml ucx in case if multithreading is requested and not supported

hoopoepg:
closed via https://github.com/open-mpi/ompi/pull/6680

PR #6210, opened 2018-12-20
Use mpi_f08 module in mpi_f08_ext module

kawashima-fj:
I cannot add a new datatype in the `mpi_f08_ext` module without this change in PR #6205 .

@jsquyres Could you review? I'm not familiar with Fortran and Autoconf. Please see commit messages for details.

I'll merge this PR after receiving review and doing more build tests.

### DIFF
--- a/ompi/mpi/fortran/mpiext-use-mpi/Makefile.am
+++ b/ompi/mpi/fortran/mpiext-use-mpi/Makefile.am
@@ -21,6 +21,7 @@ if OMPI_BUILD_FORTRAN_USEMPI_OR_USEMPIF08_EXT
 
 AM_FCFLAGS = -I$(top_builddir)/ompi/include -I$(top_srcdir)/ompi/include \
              $(OMPI_FC_MODULE_FLAG)$(top_builddir)/ompi/mpi/fortran/base \
+             $(OMPI_FC_MODULE_FLAG)$(top_builddir)/ompi/mpi/fortran/use-mpi-ignore-tkr \
              -I$(top_srcdir) $(FCFLAGS_f90)
 
 flibs =
--- a/ompi/mpiext/example/configure.m4
+++ b/ompi/mpiext/example/configure.m4
@@ -32,3 +32,11 @@ AC_DEFUN([OMPI_MPIEXT_example_CONFIG],[
 
 # only need to set this if the component needs init/finalize hooks
 AC_DEFUN([OMPI_MPIEXT_example_NEED_INIT], [1])
+
+# By default, mpiext_example_mpifh.h is included in the source file
+# of the mpi_ext module. To disable it, define this macro as 0.
+#AC_DEFUN([OMPI_MPIEXT_example_INCLUDE_MPIFH_IN_USEMPI], [0])
+
+# By default, mpiext_example_mpifh.h is included in the source file
+# of the mpi_f08_ext module. To disable it, define this macro as 0.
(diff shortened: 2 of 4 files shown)


kawashima-fj:
@jsquyres (or someone) Could you check whether this is a sane fix? I'm not familiar with Fortran and Autoconf. I tested some build patterns and all worked fine.

kawashima-fj:
@jsquyres Ping.

ompi/mpiext/example/configure.m4
```diff
@@ -32,3 +32,11 @@ AC_DEFUN([OMPI_MPIEXT_example_CONFIG],[
 
 # only need to set this if the component needs init/finalize hooks
 AC_DEFUN([OMPI_MPIEXT_example_NEED_INIT], [1])
+
+# By default, mpiext_example_mpifh.h is included in the source file
+# of mpi_ext module. To disable it, define this macro.
+#AC_DEFUN([OMPI_MPIEXT_example_NO_MPIFH_IN_USEMP], [1])
```
jsquyres: Can we reverse the names/logic?  Positive logic is always easier to read than negative logic.

E.g., `OMPI_MPIEXT_example_INCLUDE_MPIFH_IN_USEMPI` (vs. vs `NO` version).

Also, can we *always* define these values (i.e., to 0 or 1), vs. define them to 1 or undefine them?  This allows using `#if OMPI_MPIEXT_example_INCLUDE_MPFIH_IN_USEMPI` in source code, and you'll get a preprocessor error if you have made a typo in that name (which you won't get if you `#ifdef A_MACRO_NAME_WITH_A_TYPO_THAT_DOES_NOT_EXIST`).

ompi/mpi/fortran/mpiext-use-mpi/Makefile.am
```diff
@@ -21,6 +21,7 @@ if OMPI_BUILD_FORTRAN_USEMPI_OR_USEMPIF08_EXT
 
 AM_FCFLAGS = -I$(top_builddir)/ompi/include -I$(top_srcdir)/ompi/include \
              $(OMPI_FC_MODULE_FLAG)$(top_builddir)/ompi/mpi/fortran/base \
+             $(OMPI_FC_MODULE_FLAG)$(top_builddir)/ompi/mpi/fortran/use-mpi-ignore-tkr \
```
jsquyres: Is this always guaranteed to be `use-mpi-ignore-tkr`?  I.e., is it possible for OMPI to be building the `use-mpi-tkr` in the same build as Open MPI extensions with Fortran `use mpi` bindings?

jsquyres:
Sorry for the huge delay.

ompi/mpiext/example/configure.m4
kawashima-fj: @jsquyres I reversed the name/logic. Please review again.

This macro is an Autoconf macro which is used only in `ompi_ext.m4`, not a C preprocessor macro. We never see the preprocessor error.

ompi/mpi/fortran/mpiext-use-mpi/Makefile.am
kawashima-fj: Yes, it is guaranteed. Fortran bindings of MPI Extensions are not enabled for `use-mpi-tkr`. You wrote the code and comment :)
https://github.com/open-mpi/ompi/blob/master/config/ompi_ext.m4#L182

PR #6212, opened 2018-12-20
treematch: fix global common symbol

jsquyres:


bosilca:
@jsquyres you should also make the variable static.

jsquyres:
Haha!  I should make `global_b` be static?  Too funny!  😉

bosilca:
;) @GuillaumeMercier might be able to fix it upstream.

PR #6217, opened 2018-12-22
odls_base_default_fns.c: remove errant free

jsquyres:
Fixes CID 1441826.


jsquyres:
v4.0.x version of this PR filed (#6218); confirmed that this does not affect any other release branches.

jsquyres:
Fixed!

PR #6220, opened 2018-12-23
MPI_Type_set_name: fix string length at target

jsquyres:
opal_string_copy() takes care of all the string computations.
Specifically: when we converted to opal_string_copy(), we accidentally
left the *source* length as the argument, not the *target* length,
which resulted in one less character being copied than intended (as
was showing up in MTT C++ testing results).


jsquyres:
This only affects master; it does not affect any release branches.

PR #6234, opened 2019-01-01
oshmem/scoll: fix shmem_collect32/64 for zero-size length

yosefe:
Fixes https://github.com/openucx/ucx/issues/3063#issuecomment-445710345


yosefe:
@hoopoepg pls take a look

PR #6239, opened 2019-01-04
orte: shutdown

hppritcha:
be more careful about closing frameworks as part of
orte_finalize.  Owing to recent restructuring in opal to handle
finalize in a more general fashion, the missing framework
closes were causing meltdowns as the mca vars subsystem
was cleaning itself up.

This problem was recently reported by Siegmar:

https://www.mail-archive.com/users@lists.open-mpi.org//msg32946.html

### DIFF
--- a/orte/mca/ess/base/ess_base_std_orted.c
+++ b/orte/mca/ess/base/ess_base_std_orted.c
@@ -12,7 +12,7 @@
  * Copyright (c) 2009      Institut National de Recherche en Informatique
  *                         et Automatique. All rights reserved.
  * Copyright (c) 2011      Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


hppritcha:
@rhc54 i'll open a PR on prrte if you'd like.

hppritcha:
@siegmargross you may want to try this branch and see if it fixes the problem you reported.

rhc54:
> @rhc54 i'll open a PR on prrte if you'd like.

Hmmm...we haven't picked up the new OPAL finalize logic, so a PR won't work over there. Guess this is part of the "what do we do about OPAL" issue we haven't really resolved yet.

siegmargross:
Thank you very much for the solution. It works again.

Issue #6201, opened 2018-12-17
MPI_Win_create failure under 4.0.0 when UCX enabled


AdamSimpson:
### What version of Open MPI are you using?

v4.0.0 release

UCX is at v1.4.0

Configured with `./configure --with-ucx`

### Please describe the system on which you are running

* Operating system/version:  Ubuntu/16.04, Ubuntu/18.04
* Computer hardware: x86 CPU
* Network type: single host 
* no IB, no xpmem, no knem
-----------------------------

## Details of the problem
Calls to `MPI_Win_create` fail with the following:

```
$ mpirun -np 2 ./a.out
[HOST] *** An error occurred in MPI_Win_create
[HOST] *** reported by process [858718209,1]
[HOST] *** on communicator MPI_COMM_WORLD
[HOST] *** MPI_ERR_WIN: invalid window
[HOST] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[HOST] ***    and potentially your MPI job)
[HOST] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[HOST] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

I can get the test to run using the following variations:
```
$ mpirun -np 2 --mca pml ^ucx ./a.out
$ mpirun -np 2 --mca pml ob1 --mca btl self,vader ./a.out
$ mpirun -np 2 --mca btl ^vader ./a.out
$ OMPI_MCA_btl_vader_single_copy_mechanism=none mpirun -np 2 ./a.out
$ mpirun -np 2 --mca osc ^rdma ./a.out
```
Compiling OMPI without UCX also seems to work as well.

The tests above were all conducted with `test.c`, provided below.

------------------

test.c:
```c
/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil ; -*- */
/*
 *
 *  (C) 2003 by Argonne National Laboratory.
 *      See COPYRIGHT in top-level directory.
 */
#include <mpi.h>
#include <stdio.h>

#define ELEM_SIZE 8

int main( int argc, char *argv[] )
{
    int     rank;
    int     errors = 0, all_errors = 0;
    int    *flavor, *model, flag;
    void   *buf;
    MPI_Win window;

    MPI_Init(&argc, &argv);
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);

    /** Create using MPI_Win_create() **/

    if (rank > 0)
[...]

PR #6249, opened 2019-01-07
v4.0.x: btl/vader: don't try to set reachabilty in add_procs if not requested

hjelmn:
This commit fixes a bug where add_procs can incorrectly return an
error when going through the dynamic add_procs path. This doesn't
happen normally, only when pml/ob1 is not in use.

References #6201

(cherry picked from commit 30b8336cb40e586e5d926b2b52cd78bf3751e5d3)


hjelmn:
@jsquyres Fairly straightforward. In the dynamic add_procs case reachabilty is always NULL so we don't want to return an error in this case. This bug flew under the radar until I added atomic support to vader.

hjelmn:
@jsquyres Addressed your comment with the latest commit.

Issue #6247, opened 2019-01-07
v4.x REGRESSION: Updating to PMIx v3.1.0 has an issue


rhc54:
Courtesy of @amckinstry (filed originally on PMIx repo as https://github.com/pmix/pmix/issues/1032):

This is testing within Debian.

3.1.0rc1 works fine; 3.1.0rc2 fails on 32-bit archs.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918157

This is with openmpi 3.1.3.  This will not compile as it stands with rc2 (rc1 was fine), so there was a patch needed:
https://salsa.debian.org/hpc-team/openmpi/blob/debian/master/debian/patches/pmix-modex.patch

Which would be instantly suspect, except the combination works with 64-bit archs (arm64, amd, etc).

The problem is easily reproduced with a simple MPI code on i386:

```
/*
  sudo apt-get install mpi-default-bin mpi-default-dev
  export OMPI_MCA_plm_rsh_agent=/bin/false
  export OMPI_MCA_rmaps_base_oversubscribe=1
  mpicc -o mpi-test mpi-test.c && mpirun -np 2 ./mpi-test
*/
#include <mpi.h>
int main(int argc, char** argv)
{
  MPI_Init(&argc, &argv);
  MPI_Finalize();
  return 0;
}
```

giving:
```
export OMPI_MCA_orte_base_help_aggregate=0
export OMPI_MCA_btl_base_verbose=100
alastair@debian32:~$ mpirun -n 2 -mca btl self ./a.out
[debian32:04634] mca: base: components_register: registering framework btl components
[debian32:04634] mca: base: components_register: found loaded component self
[debian32:04634] mca: base: components_register: component self register function successful
[debian32:04634] mca: base: components_open: opening btl components
[debian32:04634] mca: base: components_open: found loaded component self
[debian32:04634] mca: base: components_open: component self open function successful
[debian32:04634] select: initializing btl component self
[debian32:04634] select: init of component self returned success
[debian32:04635] mca: base: components_register: registering framework btl components
[debian32:04635] mca: base: components_register: found loaded component self
[debian32:04635] mca: base: components_register: component self register function successful
[debian32:04635] mca: base: components_open: opening btl components
[debian32:04635] mca: base: components_open: found loaded component self
[debian32:04635] mca: base: components_open: component self open function successful
[debian32:04635] select: initializing btl component self
[debian32:04635] select: init of component self returned success
[...]

PR #6251, opened 2019-01-08
pmix/ext3x: fix support for external PMIx v3.1

ggouaillardet:
The PMIX_MODEX and PMIX_INFO_ARRAY macros were removed from the PMIx 3.1 standard.
Open MPI does not really need them (they are only used to be reported as not supported),
so smply #ifdef protect them to support an external PMIx v3.1

Refs. open-mpi/ompi#6247

### DIFF
--- a/opal/mca/pmix/ext3x/ext3x.c
+++ b/opal/mca/pmix/ext3x/ext3x.c
@@ -1,7 +1,7 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@rhc54 that should be good enough to support an external pmix v3.1 in `master` and `v4.0.x` branches.
is it good for you too ?

I do not mind backporting it to (ompi) `v.3.1.x` branch too.

rhc54:
Definitely! Thanks!

@bwbarrett This is a pretty small change that would make life easier for packagers - I'd recommend taking it into the v3.1 series.

ggouaillardet:
@bwbarrett the PR for `v3.1.x` is #6253 
I agree with Ralph recommending taking it to the v3.1 series.

PR #6268, opened 2019-01-10
Revert "v4.0.x: Re-add removed deprecate-only MPI-2.0 symbols"

gpaulsen:
Reverts open-mpi/ompi#6120

### DIFF
--- a/ompi/include/mpi.h.in
+++ b/ompi/include/mpi.h.in
@@ -279,7 +279,7 @@
 #                    define __mpi_interface_deprecated__(msg) __attribute__((__deprecated__))
 #                endif
 #            endif
-#            if (OMPI_ENABLE_MPI1_COMPAT && !OMPI_BUILDING)
+#            if OMPI_ENABLE_MPI1_COMPAT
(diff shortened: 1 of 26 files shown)


gpaulsen:
https://github.com/open-mpi/ompi/pull/6120 was not quite ready to go into v4.0.x

In my testing I found that while it was correctly warning of removed symbols when configured without `--enable-mpi1-compatability`, it SHOULD have been a fatal build error, rather than just a warning.

After we revert this, I'll push the newer code which fixes this.

This PR has a bit of an issue in that I hit the revert button in github to create this PR, but that does not "sign-off" the Commit, which our Signed-off-by-checker then yells about.

I'll bring up on today's RM meeting.

jsquyres:
@gpaulsen and I talked about this on the phone today.  Let's just close this PR without merging; @gpaulsen is going to open a new PR in the immediate future with the corrections to #6120 and the additional fix that is required.

gpaulsen:
New PR is: https://github.com/open-mpi/ompi/pull/6281

PR #6282, opened 2019-01-15
opal: add basic memory accounting for free lists and objects

hjelmn:
This commit adds support for exposing free list and object memory
allocation through the MPI_T pvar interface. Memory tracking of free
list memory is enabled by default for all builds. OPAL object memory
tracking is only enabled for debug builds.

In addition this commit updates the pvar interface to pass the pvar
handle to the read function instead of the pvar. This gives the read
function access to the count returned when the handle was bound.

### DIFF
--- a/opal/class/opal_free_list.c
+++ b/opal/class/opal_free_list.c
@@ -15,6 +15,8 @@
  * Copyright (c) 2011      NVIDIA Corporation.  All rights reserved.
  * Copyright (c) 2012-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019      Triad National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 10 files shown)


hjelmn:
@bosilca @jsquyres Any comments? This should have little-no overhead for optimized builds and will provide a nice capability for debug builds.

hjelmn:
@bosilca That isn't a bad idea. We can re-purpose the mpool framework to allow the base allocators to be used before the framework is opened. Then we can make all internal allocations use the default allocator and do all the accounting on mpools (which would also get free lists). I don't think it would take too much work.

hjelmn:
One question is how much info we want to track. OpenUCX tracks each individual allocation but we probably only really care about total size (for most cases).

awlauria:
@bosilca @hjelmn if you want this in for 5.0 please target getting this into master by end of April. I'll add it to the 5.0 milestone for tracking purposes for now.

lanl-ompi:

jsquyres:
@hjelmn Looks like this PR got forgotten about.  Do you still want to work on it?

hjelmn:
Maybe. Let me figure out where I was on it.

gpaulsen:
@hjelmn ?

hjelmn:
Will look tomorrow.  Finding time where I can.

hppritcha:
closing this ancient PR.  reopen if its worth reviving.

Issue #6285, opened 2019-01-17
MPI I/O Crashes and Corruption in HDF5


ax3l:
### What version of Open MPI are you using?

3.1.3

Update: affects 2.X, 3.X and 4.0.0

### Describe how Open MPI was installed

From source via spack or as modules build from source on various different HPC systems.

### Please describe the system on which you are running

* Operating system/version: Debian 9.6 and derivates such as Ubuntu
* Computer hardware: Laptops and HPC
* Network type: local and remote (Ethernet and IB)

## Details of the problem

We are currently reporting two parallel HDF5 issues that either crash on write or corrupt data. The issues only occur with OpenMPI, not with MPICH 3.3 which we used for comparison.

We (@ax3l, @psychocoderHPC) are not the upstream authors of HDF5, but wanted to inform and connect you, so you are aware of these issues as they might be rooted in the OpenMPI I/O layer somewhere.

Link to HDF5 parallel I/O write issues with reproducers:
- [x] crash with some special-size struct types: https://forum.hdfgroup.org/t/cannot-write-more-than-512-mb-in-1d/5118
 -> #6286 #6287 (a specific OpenMPI I/O issue)
- [x] data corruption: https://forum.hdfgroup.org/t/hdf5bug-h5fd-mpio-collective-chunking/5279
 -> #6295 #6326 #6172 (a general OpenMPI Datatype issue)

edgargabriel:
Looking into the issues.

edgargabriel:
The first issue (cannot write more than 512MB) is an easy fix. First, the test passes with all collective I/O components except for 'individual', which is however the selected one for 1-process scenarios. 

The culprit is a floating point division of two very close, large values.

no _of_cyles = ceil(536870916/636870912)

should be 2, but is 1 if you use floats (which is what the individual component used). If you use double precision floating point values, the result is as expected 2 (which is what all other collective I/O components use). A fix is coming soon.

If you want to run the code before the next release is available, simply exclude the individual fcoll component for this scenario, e.g. 

mpirun --mca fcoll ^individual -np 1 ./....

ax3l:
Thank you for so quickly looking into this!
So the first issue is located at:
- https://github.com/open-mpi/ompi/blob/v4.0.0/ompi/mca/common/ompio/common_ompio_file_write.c#L112
- https://github.com/open-mpi/ompi/blob/v4.0.0/ompi/mca/common/ompio/common_ompio_file_read.c#L134

[...]

PR #6286, opened 2019-01-17
common/ompio: fix a floating point division problem

edgargabriel:
This commit fixes  a problem reported on the mailing list with
individual writes larger than 512 MB.

The culprit is a floating point division of two large, close values.
Changing the datatypes from float to double (which is what is being
used in the fcoll components) fixes the problem.

See issue #6285 and

 https://forum.hdfgroup.org/t/cannot-write-more-than-512-mb-in-1d/5118

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_file_read.c
+++ b/ompi/mca/common/ompio/common_ompio_file_read.c
@@ -9,7 +9,7 @@
  *                          University of Stuttgart.  All rights reserved.
  *  Copyright (c) 2004-2005 The Regents of the University of California.
  *                          All rights reserved.
- *  Copyright (c) 2008-2018 University of Houston. All rights reserved.
+ *  Copyright (c) 2008-2019 University of Houston. All rights reserved.
  *  Copyright (c) 2018      Research Organization for Information Science
  *                          and Technology (RIST). All rights reserved.
  *  $COPYRIGHT$
@@ -33,8 +33,8 @@
 
 #include "common_ompio.h"
 #include "common_ompio_request.h"
-#include "math.h"
 #include <unistd.h>
+#include <math.h>
 
 #if OPAL_CUDA_SUPPORT
 #include "common_ompio_cuda.h"
@@ -132,8 +132,8 @@ int mca_common_ompio_file_read (ompio_file_t *fh,
     else {
 	bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
-
+    cycles = ceil((double)max_data/bytes_per_cycle);
+    
 #if 0
 	printf ("Bytes per Cycle: %d   Cycles: %d max_data:%d \n",bytes_per_cycle, cycles, max_data);
 #endif
--- a/ompi/mca/common/ompio/common_ompio_file_write.c
+++ b/ompi/mca/common/ompio/common_ompio_file_write.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2008-2018 University of Houston. All rights reserved.
+ * Copyright (c) 2008-2019 University of Houston. All rights reserved.
  * Copyright (c) 2015-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
@@ -31,8 +31,8 @@
 
 #include "common_ompio.h"
 #include "common_ompio_request.h"
-#include "math.h"
 #include <unistd.h>
+#include <math.h>
 
 #if OPAL_CUDA_SUPPORT
 #include "common_ompio_cuda.h"
@@ -116,7 +116,7 @@ int mca_common_ompio_file_write (ompio_file_t *fh,
     else {
 	bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = ceil((double)max_data/bytes_per_cycle);
 
 #if 0
     printf ("Bytes per Cycle: %d   Cycles: %d\n", bytes_per_cycle, cycles);
@@ -409,7 +409,7 @@ int mca_common_ompio_file_iwrite_at_all (ompio_file_t *fp,
 /**************************************************************/
 
 int mca_common_ompio_build_io_array ( ompio_file_t *fh, int index, int cycles,
-                                      size_t bytes_per_cycle, int max_data, uint32_t iov_count,
+                                      size_t bytes_per_cycle, int  max_data, uint32_t iov_count,
                                       struct iovec *decoded_iov, int *ii, int *jj, size_t *tbw, 
                                       size_t *spc)
 {


ompi/mca/common/ompio/common_ompio_file_read.c
```diff
@@ -132,7 +132,7 @@ int mca_common_ompio_file_read (ompio_file_t *fh,
     else {
 	bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = ceil((double)max_data/bytes_per_cycle);
```
ax3l: please don't change it from
```C
ceil((float)max_data/bytes_per_cycle;
// to
ceil((double)max_data/bytes_per_cycle;
```

which only defers the issue.
Instead, just cast both `max_data` and `bytes_per_cycle` to `uint64_t` to be platform agnostic and convert it to an ["integer-`ceil`"](https://stackoverflow.com/questions/2745074/fast-ceiling-of-an-integer-division-in-c-c) via:
```C
(max_data + bytes_per_cycle - 1) / bytes_per_cycle
```

jsquyres:
@edgargabriel Can you put a "Thanks to ..." note in your commit message?  It helps us when creating NEWS items for the release.

Thanks!

edgargabriel:
done both requests, thanks!

ompi/mca/common/ompio/common_ompio_file_write.c
```diff
@@ -116,7 +115,7 @@ int mca_common_ompio_file_write (ompio_file_t *fh,
     else {
 	bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = (max_data + bytes_per_cycle - 1) / bytes_per_cycle;
```
ax3l: From reviewing the types, `cycles` here should, as in `read`, be at least `size_t` instead of `int`.

Saver would be to make all byte counters in both methods fixed to `uint64_t`...

Anyway, `int`'s range is on 64bit architectures/data models smaller than the range of `size_t` and at least read/write should implement the same overflow ranges :)

ompi/mca/common/ompio/common_ompio_file_read.c
```diff
@@ -132,7 +131,7 @@ int mca_common_ompio_file_read (ompio_file_t *fh,
     else {
 	bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = (max_data + bytes_per_cycle - 1) / bytes_per_cycle;
```
psychocoderHPC: As @ax3l wrote [here](https://github.com/open-mpi/ompi/pull/6286#discussion_r248764855) I would suggest to use `uint64_t` at this place else this equation will reduce the possible range of `max_data` on a 32Bit system to `[0;max_data - bytes_per_cycle)`. All values larger for `max_data` will produce an overflow. 
With the current PR it would not be possible to write ~4GiB data on a 32Bit system.

psychocoderHPC: A possible other solution if the type should stay `size_t` is: `cycles = max_data / bytes_per_cycle + max_data % bytes_per_cycle ? 1u : 0u ;`

edgargabriel: Well, honestly I think going back to the ceil solution might the `easiest way out, and simple avoid the type cast to a float. The problem is that changing the 'cycles' variable to uint64_t forces you to change that also for the 'index; variable (otherwise you get a warning in the for loop), and the interface of the build_io_array function which is used all over the code base. Not sure I want to make this sweeping change at this point.

edgargabriel: Another solution could be to calculate the no of cycles into a temporary a uint64_t, and then assign that value using an int typcast to 'cycles'. This would have the advantage that we do not have to change the interfaces, and till get rid of the ceil function. An int as the datatype for the no_of_cycles is perfectly fine, you can read/write 512 MB per cycle, so (if my math is correct) that would still allow for 1024 PB of data overall being written, that should be good enough for the next decade.

ax3l:
Rather the latter, if you ask me, but I can check tomorrow.
Be aware that there is no int-ceil in C, which would mean even without the cast it is an implicit cast to a float type...

What's the exact warning you see now?

ompi/mca/common/ompio/common_ompio_file_read.c
ax3l: Rather the latter, if you ask me, but I can check tomorrow. Be aware that there is no int-ceil in C, which would mean even without the cast it is an implicit cast to a float type...

What's the exact warning you see now?

edgargabriel:
The failure in th
e pull request build checker is unrelated to this commit:
java.nio.file.FileSystemException: /global/homes/h/hpp/jenkins_cori/workspace/open-mpi.build.platforms/Platform/cray_cle: Disk quota exceeded

ompi/mca/common/ompio/common_ompio_file_read.c
ax3l: @edgargabriel have you seen this comment? :) We just mentioned that `max_data` on 32-bit systems will limit writes to 4 Gbyte as it is only `size_t`. That's our concern.

Regarding the `cycles` var: it should probably also be at least of `size_t`, better also a fixes large integer type as you just added. An `int` is most of the time 2 to 4 bytes. With an upper bound of 32767 cycles on some 32-bit systems and 512 MByte per cycle this is as well "only" 16 PByte. (4 Byte `int`s are indeed opening this to 1 EByte). Not as critical here, but `max_data` is a concern and should be of larger type.

edgargabriel:
You are right. I spent now more time on this (what was supposed to be ) one line fix than I can justify for my work. I am going back to my original solution, which was technically perfectly fine, and introduces the least amount of change from one release of the 3.1.x series to the next one (and dito for 4.0.x), and therefore the least surprise. Applications and tests that worked before, will also continue to work, and the bug originally reported in this issue is fixed. Thanks.

ax3l:
Let us please just do a minimal fix as in #6287, the double-cast is not a clean solution, independent of the other aspect discussed regarding ranges.

psychocoderHPC:
> Let us please just do a minimal fix as in #6287, the double-cast is not a clean solution, independent of the other aspect discussed regarding ranges.
> 
> These two lines are all that need to be changed, no better or worse with regards to ranges, but fixing the reported crash from the rounding issue in a future-proof manner:
> 
> ```diff
> diff --git a/ompi/mca/common/ompio/common_ompio_file_read.c b/ompi/mca/common/ompio/common_ompio_file_read.c
> index 3203e2a697..49f0c89532 100644
> --- a/ompi/mca/common/ompio/common_ompio_file_read.c
> +++ b/ompi/mca/common/ompio/common_ompio_file_read.c
> @@ -132,7 +132,7 @@ int mca_common_ompio_file_read (ompio_file_t *fh,
>      else {
>         bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
>      }
> -    cycles = ceil((float)max_data/bytes_per_cycle);
> +    cycles = (max_data + bytes_per_cycle - 1) / bytes_per_cycle;
>  
>  #if 0
>         printf ("Bytes per Cycle: %d   Cycles: %d max_data:%d \n",bytes_per_cycle, cycles, max_data);
> diff --git a/ompi/mca/common/ompio/common_ompio_file_write.c b/ompi/mca/common/ompio/common_ompio_file_write.c
> index e53a1d080b..6ea8bae064 100644
> --- a/ompi/mca/common/ompio/common_ompio_file_write.c
> +++ b/ompi/mca/common/ompio/common_ompio_file_write.c
> @@ -116,7 +116,7 @@ int mca_common_ompio_file_write (ompio_file_t *fh,
>      else {
>         bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
>      }
> -    cycles = ceil((float)max_data/bytes_per_cycle);
> +    cycles = (max_data + bytes_per_cycle - 1) / bytes_per_cycle;
>  
>  #if 0
>      printf ("Bytes per Cycle: %d   Cycles: %d\n", bytes_per_cycle, cycles);
> ```
> 
> Everything else one can address and refactor later.

`cycles = max_data / bytes_per_cycle + (max_data % bytes_per_cycle ? 1u : 0u);` (without overflow risk) must be used insead of `cycles = (max_data + bytes_per_cycle - 1) / bytes_per_cycle;` to guarantee that you can write 4GiB on 32bit systems.

ax3l:
@psychocoderHPC true, otherwise we cap the upper range.
So the diff we propose is:
```diff
diff --git a/ompi/mca/common/ompio/common_ompio_file_read.c b/ompi/mca/common/ompio/common_ompio_file_read.c
index 3203e2a697..fb863a3821 100644
--- a/ompi/mca/common/ompio/common_ompio_file_read.c
+++ b/ompi/mca/common/ompio/common_ompio_file_read.c
@@ -132,7 +132,7 @@ int mca_common_ompio_file_read (ompio_file_t *fh,
     else {
        bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = max_data / bytes_per_cycle + (max_data % bytes_per_cycle ? 1u : 0u);
 
 #if 0
        printf ("Bytes per Cycle: %d   Cycles: %d max_data:%d \n",bytes_per_cycle, cycles, max_data);
diff --git a/ompi/mca/common/ompio/common_ompio_file_write.c b/ompi/mca/common/ompio/common_ompio_file_write.c
index e53a1d080b..d60c68828f 100644
--- a/ompi/mca/common/ompio/common_ompio_file_write.c
+++ b/ompi/mca/common/ompio/common_ompio_file_write.c
@@ -116,7 +116,7 @@ int mca_common_ompio_file_write (ompio_file_t *fh,
     else {
        bytes_per_cycle = OMPIO_MCA_GET(fh, cycle_buffer_size);
     }
-    cycles = ceil((float)max_data/bytes_per_cycle);
+    cycles = max_data / bytes_per_cycle + (max_data % bytes_per_cycle ? 1u : 0u);
 
 #if 0
     printf ("Bytes per Cycle: %d   Cycles: %d\n", bytes_per_cycle, cycles);
```

edgargabriel:
There is no advantage to your solution compared to mine. So I do not see the point of this. This is the end of the discussion from my point of view.

jsquyres:
bot:ompi:retest

psychocoderHPC:
@edgargabriel The point is that `double` floating point numbers has 53 significant digits. This means we can represent all byte numbers up to 8PiB without precision loss. After 8PiB we get the same problem again. So will need run into the same issue in the next 5 to 10 years.

edgargabriel:
I doubt that in my lifetime one 'internal iteration' will be increased from 512MB to 8PiB. This is not the overall amount of data that we are talking about.

jsquyres:
Let me ask what might be a naieve question (I'm not an IO expert and I'm not very familiar with this code).  It looks like this code is looping over reading `bytes_per_cycle` bytes each iteration.

Is it feasible to read 8PiB in a single iteration?  Or will we be using some other algorithm for that?

ax3l:
All fine with us, we are not trying to nag.
We just searched for a solution that avoids a cast, since this is pure byte-counter math, and is type-agnostic in case the underlying types gets refactored some day.

edgargabriel:
bot:ompi:retest

edgargabriel:
bot:ompi:retest

ax3l:
One of the CI servers has run out of disk quota.

hppritcha:
bot:lanl:retest

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

PR #6287, opened 2019-01-18
common/ompio: possible rounding issue

psychocoderHPC:
Similar to #6286 rounding number of bytes into a single precision floating point value to round up the result of a division is a potential risk due to rounding errors.

- remove floating point operations for `round up`
- removes floating point conversion for round down (native behavior of integer division)

Signed-off-by: René Widera <r.widera@hzdr.de>

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_aggregators.c
+++ b/ompi/mca/common/ompio/common_ompio_aggregators.c
@@ -896,11 +896,14 @@ int mca_common_ompio_split_initial_groups(ompio_file_t *fh,
     int size_smallest_group = 0;
     int num_groups = 0;
     int ret = OMPI_SUCCESS;
+    OMPI_MPI_COUNT_TYPE bytes_per_agg_group = 0;
 
     OMPI_MPI_OFFSET_TYPE max_cci = 0;
     OMPI_MPI_OFFSET_TYPE min_cci = 0;
 
-    size_new_group = ceil ((float)OMPIO_MCA_GET(fh, bytes_per_agg) * fh->f_init_procs_per_group/ bytes_per_group);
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
To avoid style changes I keeped the indention for this PR equal to the lines I changed even if there are tabs in the first column. This will not follow the contribution guide lines but I tried to avoid to change other code around the touched lines.

ompi/mca/common/ompio/common_ompio_aggregators.c
```diff
@@ -896,11 +896,14 @@ int mca_common_ompio_split_initial_groups(ompio_file_t *fh,
     int size_smallest_group = 0;
     int num_groups = 0;
     int ret = OMPI_SUCCESS;
+    OMPI_MPI_COUNT_TYPE bytes_per_agg_group = 0;
```
psychocoderHPC: Is `OMPI_MPI_COUNT_TYPE` the right type for a byte count here or should it be `MPI_Aint`?

jsquyres:
ok to test

edgargabriel:
I am honestly still unclear why the ceil and floor functions are considered inadequate, i.e. not convinced this is truly necessary.

psychocoderHPC:
@edgargabriel Since `io_ompio_bytes_per_agg` (default on my system: 33554432) is an integer which can be chosen by the user form the environment it can result into the same rounding issue than #6286 and if the value is larger than 24bit it is not anymore possible to represent the value without rounding as a float value.

edgargabriel:
after thinking a bit more about it, the problem pinpointed by this fix is a legitimate concern. Whether we use a double precision value and the ceil function for resolving it, or your approach doesn't really matter in my opinion, so we can this solution as well.

Issue #6291, opened 2019-01-21
Memory Leak in MPI_Get_elements_x


ax3l:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

`master` as of 352b667323d02c0eb7c2df6325554645ffe56920

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

From source via `spack`.

### Please describe the system on which you are running

* Operating system/version: Debian 9
* Computer hardware: ThinkPad
* Network type: Local

-----------------------------

## Details of the problem

The function `opal_datatype_compute_ptypes` in https://github.com/open-mpi/ompi/blob/352b667323d02c0eb7c2df6325554645ffe56920/ompi/datatype/ompi_datatype_get_elements.c#L52 allocates datatypes in https://github.com/open-mpi/ompi/blob/98afc838aa53da88cba339f6dcbab256806a5745/opal/datatype/opal_datatype_get_count.c#L161 that don't seem to be freed.

Seen while running a small MPI-I/O example with HDF5 `develop`: https://forum.hdfgroup.org/t/hdf5bug-h5fd-mpio-collective-chunking/5279

ggouaillardet:
@ax3l thanks for the report !

Meanwhile, feel free to manually download and apply the patch at https://github.com/open-mpi/ompi/pull/6292.patch

ggouaillardet:
The fix has landed into both `master` and `v4.0.x` branch, so I am now closing this issue.

ax3l:
Thanks a lot!

PR #6292, opened 2019-01-22
opal/datatype: plug a memory leak in opal_datatype_t destructor

ggouaillardet:
correctly free ptypes if the datatype is not pre-defined.

Thanks Axel Huebl for reporting this.

Refs. open-mpi/ompi#6291

### DIFF
--- a/opal/datatype/opal_datatype_create.c
+++ b/opal/datatype/opal_datatype_create.c
@@ -11,6 +11,8 @@
  * Copyright (c) 2004-2006 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:

I quickly checked the code and at first glance, `id` from `opal_datatype_t` is not used (e.g. left untouched at `0`) if the datatype is not a predefined one.

jsquyres:
bot:ompi:retest

hppritcha:
bot:ompi:retest

bosilca:
All OMPI datatypes are supposed to be created with the id set to -1 (as showed [here](https://github.com/open-mpi/ompi/blob/master/ompi/datatype/ompi_datatype_create.c#L32)). In any case using the predefined check is a safe alternative.

PR #6294, opened 2019-01-23
mtl/ofi: Print descriptive error message on modex failure

bwbarrett:
With MTLs, there's no "other transport" when the remote side
does not have an active NIC, so we should print a useful error
message when the modex failed (indicating lack of a NIC on
the remote side).

### DIFF
--- a/ompi/mca/mtl/ofi/help-mtl-ofi.txt
+++ b/ompi/mca/mtl/ofi/help-mtl-ofi.txt
@@ -65,3 +65,13 @@ are more threads than the available contexts.
 
   Local host: %s
   Location: %s:%d
+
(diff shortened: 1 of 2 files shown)


bwbarrett:
@rhc54 I asked for your review because I can't remember the existence guarantees on the remote hostname.

bwbarrett:
bot:ompi:retest

It looks like the cray had an internal failure.

bwbarrett:
bot:ompi:retest

cori still having problems.

bwbarrett:
bot:ompi:retest

The Cray configs are having a bad day...  Removed one of the others that had a full disk quota.

PR #6302, opened 2019-01-24
mtl/ofi: Provide av count hint during initialization

bwbarrett:
Provide the av_attr.count hint (number of addresses that will be
inserted into the address vector through the life of the process)
at initialization of the address vector.  It's ok to be a bit
wrong, but some endpoints (RxR) can benefit by not going through
the slow growth realloc churn.


bwbarrett:
bot:ompi:retest

Looks like we have sanity with the Cray testers again.

hppritcha:
bot:ompi:retest

Issue #6303, opened 2019-01-25
Possible Race condition in close_open_file_descriptors()


uberlinuxguy:
Thank you for taking the time to submit an issue!

## Background information
There is a possible close/read race condition in the `close_open_file_descriptors()` function of all odls modules.  The trigger is unknown, and the condition is not easily reproducible.  It is possible that the underlying trigger is a bug in libc or the underlying open/read/close system calls in the kernel, but there is a safe work around.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
Looking at the code, this effects all active branches, and possibly stale branches, of openmpi.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

OpenMPI was installed from source when this bug was detected.

### Please describe the system on which you are running

* Operating system/version: CentOS 7
* Computer hardware: x86_64 arch
* Network type: OpenIB

-----------------------------

## Details of the problem
`close_open_file_descriptors()` seems to go through the open fs in `/proc/self/fd/` and closes them all.  However, one of the last fd's it closes is the fd on the `DIR` structure returned by `opendir()`.  In most instances, this works fine.  Under certain, currently unknown, circumstances (possibly kernel or libc related) a segmentation fault happens in the `readdir()` function on the DIR that is closed by while loop.

The proposal is to skip the fd of the open DIR structure.  I actually have a working patch for the `odls_default_module` and plan to port the patch to the other odls modules.  I do not, however, have a way to test the patch with the Cray Alps launcher.  I also do not have a good way to trigger this possible race condition, but the proposed patch will avoid it.

I am planning to issue pull requests for all of the active branches as well as master per the guidelines.

uberlinuxguy:
All pull requests were merged for this issue.  I can now be closed.

PR #6307, opened 2019-01-28
Adding changes for issue #6303 for branch v4.0.x.

uberlinuxguy:
Signed-off-by: Jason Williams <uberlinuxguy@gmail.com>

### DIFF
--- a/orte/mca/odls/alps/odls_alps_module.c
+++ b/orte/mca/odls/alps/odls_alps_module.c
@@ -298,6 +298,15 @@ static int close_open_file_descriptors(int write_fd,
         return ORTE_ERR_FILE_OPEN_FAILURE;
     }
     struct dirent *files;
+
+    /* grab the fd of the opendir above so we don't close in the 
+     * middle of the scan. */
+    int dir_scan_fd = dirfd(dir);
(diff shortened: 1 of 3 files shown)


ggouaillardet:
@uberlinuxguy Thanks for the fix and the PR !

Since open-mpi/ompi@98d81a5f7a619d5a19615297a6fe8a18d8e3781c has been merged into master, the way we do things in Open MPI is `cherry-pick -x 98d81a5f7a619d5a19615297a6fe8a18d8e3781c` into the release branches, so the commit message contains the following line
```
(cherry picked from commit 98d81a5f7a619d5a19615297a6fe8a18d8e3781c)
```
you might have to manually fix a few things because the cherry-pick does not work out of the box,
in this case I replace `cherry picked` with `back-ported`. In some cases, there is no cherry-pick possible because `master` does things differently than the release branches, and we explicitly state in the commit message this is a one-off commit for the release branches.

Could you please fix the PRs so they include the reference to the commit in the `master` branch ?
Then you will have to `git push --force ...` in order to overwrite the (commits of the) existing PRs.

ggouaillardet:
This one looks good to me, thanks !

Issue #6275, opened 2019-01-14
RMA accumulate is busted in 4.0


jeffhammond:
ARMCI-MPI is experiencing incorrect results with shared-memory accumulate with Open-MPI 4.0.  Version 3.x is fine.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

4.0.0

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Homebrew or built from source.

### Please describe the system on which you are running

* Operating system/version:  Mac or Linux
* Computer hardware: Intel CPUs.
* Network type: Shared-memory communication.

-----------------------------

## Details of the problem

https://travis-ci.org/pmodels/armci-mpi/jobs/477111650 is but one example.

### Local testing

MacOS Mojave

Open-MPI from Homebrew:
```sh
jrhammon-mac02:prk-repo jrhammon$ mpicc -show
clang -I/usr/local/Cellar/open-mpi/4.0.0/include -L/usr/local/opt/libevent/lib -L/usr/local/Cellar/open-mpi/4.0.0/lib -lmpi
jrhammon-mac02:prk-repo jrhammon$ brew info open-mpi
open-mpi: stable 4.0.0 (bottled), HEAD
High performance message passing library
https://www.open-mpi.org/
Conflicts with:
  mpich (because both install MPI compiler wrappers)
/usr/local/Cellar/open-mpi/4.0.0 (753 files, 10.7MB) *
  Poured from bottle on 2019-01-14 at 11:27:52
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/open-mpi.rb
==> Dependencies
Required: gcc ✔, libevent ✔
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 8,380 (30 days), 32,457 (90 days), 95,919 (365 days)
install_on_request: 2,965 (30 days), 11,847 (90 days), 37,813 (365 days)
build_error: 0 (30 days)
```

Not using datatypes works fine:
```sh
ARMCI_STRIDED_METHOD=IOV ARMCI_IOV_METHOD=BATCHED ARMCI_USE_WIN_ALLOCATE=1 mpirun -oversubscribe -n 4 ./tests/contrib/armci-test
[...]

PR #6312, opened 2019-01-29
ompi/op: fix support of non predefined datatypes with predefined oper…

ggouaillardet:
…ators

ACCUMULATE, unlike REDUCE, can use with derived
datatypes with predefinied operations, with some
restrictions outlined in MPI-3:11.3.4.  The derived
datatype must be composed entierly from one predefined
datatype (so you can do all the construction you want,
but at the bottom, you can only use one datatype, say,
MPI_INT).

Refs. open-mpi/ompi#6275


hjelmn:
please PR to v3.0.x, v3.1.x, and v4.0.x

Issue #2519, opened 2016-12-05
Patch for linking libfabric


amckinstry:
Linking libfabric breaks on Debian/Ubuntu systems (at least) without the following patch:

```
Author: Gianfranco Costamagna <locutusofborg@debian.org>

--- openmpi-2.0.1.orig/ompi/mca/mtl/ofi/Makefile.am
+++ openmpi-2.0.1/ompi/mca/mtl/ofi/Makefile.am
@@ -43,7 +43,7 @@ mca_mtl_ofi_la_SOURCES = $(mtl_ofi_sourc
mca_mtl_ofi_la_LDFLAGS = \
$(ompi_mtl_ofi_LDFLAGS) \
-module -avoid-version
-mca_mtl_ofi_la_LIBADD = $(ompi_mtl_ofi_LIBS) \
+mca_mtl_ofi_la_LIBADD = $(ompi_mtl_ofi_LIBS) $(opal_common_libfabric_LIBS) \
$(OPAL_TOP_BUILDDIR)/opal/mca/common/libfabric/lib@OPAL_LIB_PREFIX@mca_common_libfabric.la

noinst_LTLIBRARIES = $(component_noinst)
```

ggouaillardet:
@amckinstry thanks for the patch

@rhc54 @jsquyres that is an intersting one ...

we link ```libmca_common_libfabric.la``` with ```-lfabric``` and naively hope ```libfabric.so``` is a dependency.
as shown by ```ldd```, that is true on Centos 7, but not on debian (i tested ubuntu 14.04.3-LTS). if i understand correctly, the reason is ```libmca_common_libfabric.la``` does not need ```libfabric.so``` at all (!)

indeed, ```opal/mca/common/libfabric/common_libfabric.s``` only contains
```c
int mca_common_libfabric_register_mca_variables(void)
{
    return OPAL_SUCCESS;
}
```

a possible fix is the suggested patch.
an other one is to really make ```libmca_common_libfabric.la``` depend on ```libfabric.so```
for example

```c
int mca_common_libfabric_register_mca_variables(void)
{
    if (fi_version() >= FI_VERSION(1,0)) {
        return OPAL_SUCCESS;
    } else {
        return OPAL_ERROR;
   }
}
```

and an other one is to simple remove ```opal/mca/common/libfabric``` (it does not do much today)

any thoughts ?

rhc54:
There was some logic behind that component, but I honestly don't recall. I'd just use your patch for now.

[...]

PR #6313, opened 2019-01-29
configury: fix ofi components dependencies

ggouaillardet:
use $(opal_common_ofi_*) variables since these are the only
one defined (by opal/mca/common/ofi/configure.m4)

Refs. open-mpi/ompi#2519

Thanks Alastair McKinstry for the report and initial fix.
Thanks Rashika Kheria for the reminder.

### DIFF
--- a/ompi/mca/mtl/ofi/Makefile.am
+++ b/ompi/mca/mtl/ofi/Makefile.am
@@ -5,6 +5,8 @@
 # Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
 #                         reserved.
 # Copyright (c) 2017      IBM Corporation.  All rights reserved.
+# Copyright (c) 2019      Research Organization for Information Science
+#                         and Technology (RIST).  All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


rhc54:
Please delete the rml/ofi component - it never worked very well and there is no interest in pursuing it further.

bwbarrett:
Oi, there's a trail of sadness in this one.  Calling other component's configure macro works, but is super hinky from a "components should be independent" standpoint.  At least it would be an error during autogen.

I'm pretty sure common:ofi was originally meant to hold all the MCA parameters that are currently set by the various OFI components.  It looks like most of those parameters ended up being unique to the components (ie, most of the mtl:ofi parameters only make sense when dealing with the point-to-point interface), so the common parameter pool remains empty.

This is the minimal patch to remove the suck.  Long term, we need to rethink our sharing strategy; this one doesn't match what other interfaces are doing.

jsquyres:
I agree with Ralph -- if rml/ofi is the problem and it is unmaintained / does not work, then perhaps we should just remove that component.

Additionally, as Brian mentioned, it looks like common/ofi was meant to hold common MCA vars.  But it's empty.  Meaning: this never happened (and if we remove rml/ofi, then we're down to a single OFI component, and perhaps common/ofi should go away as well).

I.e., perhaps the real fix should be removing rml/ofi and common/ofi, and then fixing the macros in `mtl/ofi/Makefile.am` (which also means moving the `OPAL_CHECK_OFI` from `opal/mca/common/ofi/configure.m4` to `ompi/mca/mtl/ofi/configure.m4`)

jsquyres:
We talked about this on the webex today:

1. `opal/mca/common/ofi` has no C code that anyone is using
1. But `opal/mca/common/ofi` does call `OPAL_CHECK_OFI`, the output of which is used by multiple other components (contrary to what I said in a comment, above)
1. @matcabral volunteered to look deeper into this

rhc54:
My contribution is https://github.com/open-mpi/ompi/pull/6338

jsquyres:
@ggouaillardet Can we close this one without merging in favor of #6363?

PR #6314, opened 2019-01-29
hwloc/external.h: fix a clash with external HWLOC_VERSION

jsquyres:
Top-level configure will *always* configure the embedded hwloc
component, even if we already know that we'll be using an external
hwloc (because of complicated reasons).  A side-effect of this is that
the embedded hwloc will AC_DEFINE HWLOC_VERSION in opal_config.h.  If
the external hwloc defines a different value of HWLOC_VERSION, we'll
get zillions of warnings about the two HWLOC_VERSION values not
matching.

So in hwloc/external.h, #undef HWLOC_VERSION after including
<opal_config.h>, but before including the external <hwloc.h>.

### DIFF
--- a/opal/mca/hwloc/external/external.h
+++ b/opal/mca/hwloc/external/external.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2017 Cisco Systems, Inc.  All rights reserved
+ * Copyright (c) 2011-2019 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2016      Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@jsquyres I tried to solve the same issue, but ended with a different fix in #6185 (note this PR should only contain one commit). If you believe this is a better fix (you might have to `#undef` some more macros though), I am fine with it and will close #6185.

jsquyres:
Given the time zone difference between us, I just updated #6185 -- let's continue over there.

PR #6315, opened 2019-01-29
oob/tcp: correctly fallback to the loopback interface

ggouaillardet:
vir* interfaces are ignored but they prevented from falling
back to the loopback interface if needed.

Thanks Patrick Begou for reporting this

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_component.c
+++ b/orte/mca/oob/tcp/oob_tcp_component.c
@@ -16,8 +16,8 @@
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2018 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014      NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2015-2017 Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2015-2019 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
@@ -484,6 +484,10 @@ static int component_available(void)
     char name[32];
     struct sockaddr_storage my_ss;
     int kindex;
+    char **ipv4loops = NULL;
+#if OPAL_ENABLE_IPV6
+    char **ipv6loops = NULL;
+#endif
 
     opal_output_verbose(5, orte_oob_base_framework.framework_output,
                         "oob:tcp: component_available called");
@@ -566,16 +570,6 @@ static int component_available(void)
                     continue;
                 }
             }
-        } else {
-            /* if no specific interfaces were provided, we ignore the loopback
-             * interface unless nothing else is available
-             */
-            if (1 < opal_ifcount() && opal_ifisloopback(i)) {
-                opal_output_verbose(20, orte_oob_base_framework.framework_output,
-                                    "%s oob:tcp:init rejecting loopback interface %s",
-                                    ORTE_NAME_PRINT(ORTE_PROC_MY_NAME), name);
-                continue;
-            }
         }
 
         /* Refs ticket #3019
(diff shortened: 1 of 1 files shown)


rhc54:
I have no real objection to this PR, but I don't understand how your comment has anything to do with it. We still ignore vir interfaces after this change. FWICT, the only way this change impacts anything is if opal_ifcount is returning a value greater than 1 when only the loopback device is present - which seems incorrect. Or are you saying that someone has virtual interfaces active, but no real ones, and so opal_ifcount is coming back with an incorrect number?

ggouaillardet:
if a machine only has `lo` and `virbr0`, `opal_ifcount()` is `2` so `lo` is skipped by `oob/tcp`, and since we always skip `virbr0`, we end up with no available interface and `mpirun` fails.

this PR fixes that.

Can you please suggest a better wording ?

ggouaillardet:
all I can say for now is `oob/tcp` does explicitly discard virtual interfaces, though `btl/tcp` uses them.
I cannot remember the rationale for `oob/tcp` (`git blame` points to https://github.com/open-mpi/ompi/commit/a2919174 but the commit message does not mention this part).

@jsquyres @bosilca any thoughts ?

ggouaillardet:
Now I remember ... we disable the virtual interfaces because it is not uncommon they all share the same IP on all the nodes of a cluster. In some cases, we need to use it so we should add yet an other MCA param.

rhc54:
IIRC, I skipped the virtual interfaces because we would pick them up, but they don't support loopback. Thus, mpirun couldn't use them to send to itself. You raise another good point as well - the problem is that we might need to talk to a daemon in a local VM but also need to talk to daemons on other host machines that include the same virtual interface.

I guess the question is: how is the btl/tcp component sorting this out? Can we use their logic here? No point in recreating the wheel, and I'm a little concerned that we might not allow a daemon-to-daemon connection when we would allow proc-to-proc communications.

jsquyres:
We talked about this on the webex today.

1. The consensus seemed to be that if we try to "fix" this (even if we just change the default values for if include/exclude for BTL and OOB), we'll fix it for this guy and likely break it for someone else.
1. For v4.0.x, the solution then feels like we should just document the MCA params that fix it for this user.
1. For v5.0.x, we should finish the "reachable" framework -- that's a pretty definitive way to determine whether a peer is reachable via a given interface or not.  Then we don't have to rely on heuristics that may or may not work for everyone.

jsquyres:
@bosilca We agreed on the call today to *not* merge this PR.  See https://github.com/open-mpi/ompi/pull/6315#issuecomment-458612812.  Do you feel differently?

bosilca:
No I'm fine either way. It's just that we have here a simple fix for most 'trivial' cases but we prefer to delay without a clear deadline a potential solution .

ggouaillardet:
@jsquyres My understanding of your previous comment is this PR might break something, and unless I am missing something, I think it does not.

Here is the issue the user is facing and this PR fixes
 - 3 interfaces on his laptop : `lo`, `virbr0`, and `wifi`
 - `oob/tcp` always discards `virbr0` (this is hardcoded), and since `opal_ifcount() == 3`, `lo` is discarded too
 - `oob/tcp` (only) uses the `wifi` interface, and he can `mpirun hostname` on his laptop.

Now the users takes the train and the `wifi` interface goes down.
 - `oob/tcp` still discards `virbr0` since this is hardcoded
 - `opal_ifcount() == 2`, so `lo` is discarded too
 - `oob/tcp` ends up with no available interface and `mpirun` cannot start.

If the user did not have `virbr0`, then `oob/tcp` would have used `lo` (since `opal_ifcount() == 1`) and `mpirun hostname` would have always worked regardless the status of his wifi connection.

My analysis is `oob/tcp` cannot use `opal_ifcount()` to determine whether `lo` should be used or discarded, and this PR fixes that __and only that__. Also, I believe it does not break anything else.

- `opal_ifcount()` is not updated (long story short, `btl/tcp` is behavior is not altered with to `vir*` interfaces, and in this specific case it does not even matter since `btl/vader` will be used).
- `virbr0` is still never used by `oob/tcp` (which means `mpirun` cannot start a MPI app on a VM/container only accessible from the host via `virbr0`). This is currently unsupported (`vir*` interfaces are never used by `oob/tcp` and this is hardcoded), and this PR does not address this.

I already provided a workaround to the user
(either shutdown `virbr0` or `mpirun --mca oob_tcp_if_include lo` when the wifi is down),
but this is not very user-friendly.

I definitely agree we have to improve things in a near future, and unless we have a plan/roadmap/deadline for that __or__ this PR does break something, I'd rather merge it and backport it to the release branches in order to fix the issue reported by the user (no more, no less).

rhc54:
Guess I'm confused. As Jeff documented, we _do_ have a plan/roadmap to go forward - it consists of documenting the very simple MCA parameter solution to the problem. Remember, we are talking about a near-zero population of people experiencing the problem. The MCA solution for v4.0.x is just fine and low risk. The problem with this PR is that we won't know that we broke somebody else until they start complaining about it - and then it's too late. We've been there - this problem of how to handle loopback and virtual interfaces has continued to haunt us over the years. Every time we fix it, we break it.

Fixing it for v5 will take more time, but we have agreement on doing it and a rough timetable. It is the long-term correct solution.

ggouaillardet:
I think I got it now, the short term plan is to do nothing in existing release branches, but instead document

> use `--mca oob_tcp_if_include lo` if the only available interfaces are `lo` and `vir*`.

rhc54:
We should go ahead and close this as we are not going to use it. Someone can do the recommended documentation if so inclined.

PR #6319, opened 2019-01-29
pml/ob1: fix deadlock with communicator flag ALLOW_OVERTAKE.

thananon:
We missed an assert to check if ALLOW_OVERTAKE is set or not before
validating the sequence number and this will cause deadlock.

Signed-off-by: Thananon Patinyasakdikul <tpatinya@utk.edu>


ibm-ompi:

 Gist: https://gist.github.com/468f43825d4a35f93a680d48905f7fec

PR #6326, opened 2019-01-30
Provide a better fix for #6285.

bosilca:
The issue was a little complicated due to the internal stack used in the
convertor. The main issue was that in the case where we run out of iov
space to save the raw description of the data while hanbdling a
repetition (loop), instead of saving the current position and bailing out
directly we reading of the next predefined type element. It worked in
most cases, except the one identified by the HDF5 test. However, the
biggest issue here was the drop in performance for all ensuing calls to
the convertor pack/unpack, as instead of handling contiguous loops as a
whole (and minimizing the number of memory copies) we copied data
description by data description.

### DIFF
--- a/opal/datatype/opal_convertor_raw.c
+++ b/opal/datatype/opal_convertor_raw.c
@@ -102,7 +102,7 @@ opal_convertor_raw( opal_convertor_t* pConvertor,
                     /* now here we have a basic datatype */
                     OPAL_DATATYPE_SAFEGUARD_POINTER( source_base, blength, pConvertor->pBaseBuf,
                                                 pConvertor->pDesc, pConvertor->count );
-                    DO_DEBUG( opal_output( 0, "raw 1. iov[%d] = {base %p, length %lu}\n",
+                    DO_DEBUG( opal_output( 0, "raw 1. iov[%d] = {base %p, length %" PRIsize_t "}\n",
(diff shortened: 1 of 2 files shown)


ggouaillardet:
minor typo in the commit message, it should refer #6285

hjelmn:
```
06:32:13 FAIL: ddt_raw2
```

Looks like something is not right either with the test or the PR.

edgargabriel:
@bosilca something is not quite right, the result is still not correct (although the offsets are a different now). Will look into it later today to see whether I can provide more details.

bosilca:
ok let me check again, on my machine all datatype tests pass.

ggouaillardet:
@bosilca I pushed an other commit to this PR. `ddt_raw2` fails without it, and passes with.
If it works for you, please squash both commits and merge into master, I will PR for the release branches after.

bosilca:
Oh the newbie mistake! I did had the feeling that my pull-rebase didn't quite worked but all my tests passed so I went ahead and pushed, and indeed I inherited both my fix and @ggouaillardet. Thanks @ggouaillardet for the better fix of the better fix.

edgargabriel:
I can confirm that the hdf5 data corruption reported by @ax3l and @psychocoderHPC is fixed with this commit.

PR #6337, opened 2019-01-30
btl/vader: fix fragment sizes used by free lists

hjelmn:
This commit fixes a bug introduced in
f62d26ddbc8cda4d985cceee531a2ec32406d1f6. That commit changed how
vader allocates fragment memory from the shared memory
segment. Unfortunately, the values used for the fragment sizes did not
include space for the fragment header. This can cause an overrun of
data from one fragment to the header of the next fragment.

### DIFF
--- a/opal/mca/btl/vader/btl_vader_module.c
+++ b/opal/mca/btl/vader/btl_vader_module.c
@@ -17,7 +17,7 @@
  * Copyright (c) 2014-2015 Intel, Inc. All rights reserved.
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


jsquyres:
We, the overworked disks in Cisco's MTT cluster, confirm this fix.

For the past few weeks, Cisco MTT has been dumping terabytes of corefiles.  Our human finally got to look at it today, and saw that all the corefile stacktraces were in vader.  A quick test shows that https://github.com/open-mpi/ompi/commit/f62d26ddbc8cda4d985cceee531a2ec32406d1f6 did, indeed, show the problem (i.e., dump core, particularly in IBM reduce_scatter tests), but HEAD after this PR does not.

Thanks for the fix.

Issue #6316, opened 2019-01-29
warning: use of old-style cast to ‘void*’ [-Wold-style-cast]


Shadow-fax:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
openmpi-4.0.0

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
Compiled from a source/distribution tarball.

### Please describe the system on which you are running

* Operating system/version: Ubuntu 18.10
* Computer hardware: x86_64 arch
* Network type: 

-----------------------------

## Details of the problem
This macro in mpi.h produces so many warnings about old-style-cast in any code using OpenMPI, e.g. OpenFOAM opensource package.

    #if !OMPI_BUILDING
    #define OMPI_PREDEFINED_GLOBAL(type, global) ((type) ((void *) &(global)))
    #else
    #define OMPI_PREDEFINED_GLOBAL(type, global) ((type) &(global))
    #endif

the warnings:

    In file included from db/IOstreams/Pstreams/OPwrite.C:29:
    db/IOstreams/Pstreams/OPwrite.C: In static member function ‘static bool Foam::OPstream::write(Foam::Pstream::commsTypes, int, const char*, std::streamsize, int, Foam::label)’:
    /home/shadowfax/foam/foam-extend-4.1/ThirdParty/packages/openmpi-4.0.0/platforms/linux64GccDPOpt/include/mpi.h:329:72: warning: use of old-style cast to ‘void*’ [-Wold-style-cast]
    #define OMPI_PREDEFINED_GLOBAL(type, global) ((type) ((void *) &(global)))
                                                                        ^
    /home/shadowfax/foam/foam-extend-4.1/ThirdParty/packages/openmpi-4.0.0/platforms/linux64GccDPOpt/include/mpi.h:1049:18: note: in expansion of macro ‘OMPI_PREDEFINED_GLOBAL’
    #define MPI_BYTE OMPI_PREDEFINED_GLOBAL(MPI_Datatype, ompi_mpi_byte)
                  ^~~~~~~~~~~~~~~~~~~~~~
    db/IOstreams/Pstreams/OPwrite.C:101:13: note: in expansion of macro ‘MPI_BYTE’
             MPI_BYTE,

jsquyres:
@Shadow-fax What compiler and version are you using?

[...]

PR #6340, opened 2019-01-31
v4.0.x: mpi.h.in: use C++ static_cast<> where appropriate

jsquyres:
When compiling mpi.h with a modern C++ compiler and a high degree of
pickyness (e.g., -Wold-style-cast), casting using (void*) in the
OMPI_PREDEFINED_GLOBAL and MPI_STATUS*_IGNORE macros will emit
warnings.  So if we're compiling with a C++ compiler, use C++'s
static_cast<> instead of (void*).

Thanks to @shadow-fax for identifying the issue.

(cherry picked from commit 30afdcead915c9c5a62305b93460e2ba8cc6f801)

Refs #6316

### DIFF
--- a/ompi/include/mpi.h.in
+++ b/ompi/include/mpi.h.in
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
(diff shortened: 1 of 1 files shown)


hppritcha:
@bwbarrett please review when you have a chance

PR #6306, opened 2019-01-28
regex: fixed host ordering for different prefixes

karasevb:
This PR fixes a couple bugs:
1) **Wrong host ordering**:
```bash
user@jjssjb ~/ompi $ mpirun --np 3 -H jjss0000001,jjss000000A,jjss0000002 --mca regx fwd --mca routed radix --mca routed_radix 2 --map-by node hostname
jjss0000001
jjss000000A
jjss000000A <-- orted run twice this node
```
Regex test:
```bash
user@jjssjb ~/ompi/system $ export OMPI_MCA_regx=fwd
user@jjssjb ~/ompi/system $ ./regex jjss,jjss0000001,jjss000000A,jjss0000002
REGEX: jjss,jjss[7:1-2],jjss[6:0]A
ERROR: jjss,jjss0000001,jjss0000002,jjss000000A 
                                    ^^^^^^^^^^^ wrong hosts ordering
```

2) **Wrong regex for regx/reverse component**:
```
user@jjssjb ~/ompi $ mpirun --np 3 -H jjss0000001,jjss000000A,jjss0000002 --mca regx reverse --mca routed radix --mca routed_radix 2 -mca regx_base_verbose 5 --map-by node hostname
...
[jjss000000A:03220] [[24634,0],2] node[1].name jjss0000001 daemon 1
[jjss000000A:03220] [[24634,0],2] node[2].name jjss0000002 daemon 2
[jjss000000A:03220] [[24634,0],2] node[3].name jjss0000000 daemon 3 <--  this node is not existed
[jjss0000001:02888] [[24634,0],1] node[0].name jjssj0 daemon 0
[jjss0000001:02888] [[24634,0],1] node[1].name jjss0000001 daemon 1
[jjss0000001:02888] [[24634,0],1] node[2].name jjss0000002 daemon 2
[jjss0000001:02888] [[24634,0],1] node[3].name jjss0000000 daemon 3 <--  this node is not existed
...
ssh: Could not resolve hostname jjss0000000: Name or service not known      
                                ^^^^^^^^^^^ attempt connect to not existed host
...
```
```bash
user@jjssjb ~/ompi/system $ export OMPI_MCA_regx=reverse
user@jjssjb ~/ompi/system $ ./regex jjss,jjss0000001,jjss000000A,jjss0000002
REGEX: jjss[0:0],jjss[7:1-2],jjss000000A[0:0]
ERROR: jjs0,jjss0000001,jjss0000002,jjss0000000
[...]

PR #6213, opened 2018-12-20
Clean up ompi_mpi_init and ompi_mpi_finalize.

hjelmn:
This commit updates ompi_mpi_init and ompi_mpi_finalize to use the
opal finalize domain functionality. All subsystem initialization
functions have been updated to append their finalize function to the
current opal_finalize_domain_t. The ompi finalize domain is left open
when ompi_mpi_init returns to allow additional cleanup functions to be
registered between MPI_Init/MPI_Finalize.
    
One note on this commit: attribute initialization has been broken down
to seperate the initialization of the subsystem and creating the
predefined attributes. The subsystem is now initialized before the
communicator system and the predefined attributes are created after
the MPI window subsystem. This change fixed a teardown ordering
issue.

PR #6350, opened 2019-02-01
Move from the use of regex to compression

rhc54:
We've been fighting the battle of trying to create a regex generator and
parser that can handle arbitrary hostname schemes - without long-term
success. The worst of it is that there is no way of checking to see if
the computed regex is correct short of parsing it and doing a
character-by-character comparison with the original string. Ugh...there
has to be a better solution.

One option is to investigate using 3rd-party regex libraries as
those are coming from communities whose sole focus is resolving that
problem. However, someone would need to spend the time to investigate
it, and we'd have to find a license-friendly implementation.

Another option is to quit beating our heads against the wall and just
compress the information. It won't be as much of a reduction, but we
also won't keep hitting scenarios where things break. In this case, it
seems that "perfection" is definitely the enemy of "good enough".

This PR implements the compression option while retaining the
possibility of people adding regex-generating components. The
compression code used in ORTE is consolidated into the opal/compress
framework. That framework currently held bzip and gzip components for
use in compressing checkpoint files - since we no longer support C/R, I
have .opal_ignore'd those components.

However, I have left the original framework APIs alone in case someone
ever decides to redo C/R. The APIs of interest here are added to the
framework - specifically, the "compress_block" and "decompress_block"
functions. I then moved the ORTE zlib compression code into a new
component in this framework.

Unfortunately, the framework currently is a single-select one - i.e.,
only one active component at a time. Since I .opal_ignore'd the other
two and made the priority of zlib high, this isn't a problem. However,
if someone wants to re-enable bzip/gzip or add another component, they
might need to transition opal/compress to a multi-select framework.

Included changes:

* Consolidate the compression code into the opal/compress framework

* Move the ORTE zlib compression code into a new opal/compress/zlib
  component

* Ignore the bzip and gzip components in opal/compress framework

* Add a "compress_base_limit" MCA param to set the threshold above which
  we compress data - defaults to 4096 bytes

* Delete stale brucks and rcd components from orte/grpcomm framework

* Delete the orte/regx framework

* Update the launch system to use opal/compress instead of string regex


rhc54:
@ggouaillardet Can you give this a try on a multi-node system? I don't think any of our CI does that, and I'd like to ensure we verify at least minimal behavior on more than one setup.

rhc54:
Closes #6306

rhc54:
@ggouaillardet There is still a bug in multi-node operations, but I'm having trouble investigating it. My Virtualbox installs are Ubuntu, which speaks a foreign language to me when it comes to core files. Anyway, if/when you get a chance, please see if you can spot the problem - I know I still need to do something with the topology...that may be the remaining issue. I'll try to do something about it over the weekend.

ibm-ompi:

ibm-ompi:

rhc54:
I have this working now, but it crashes during finalize with something new to me:
```shell
[rhc001:05345] *** Process received signal ***
[rhc001:05345] Signal: Segmentation fault (11)
[rhc001:05345] Signal code: Address not mapped (1)
[rhc001:05345] Failing at address: 0x7fcd3cda6268
[rhc001:05345] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7fcd453d5890]
[rhc001:05345] [ 1] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(+0x7aa0b)[0x7fcd44ce6a0b]
[rhc001:05345] [ 2] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(+0x7602b)[0x7fcd44ce202b]
[rhc001:05345] [ 3] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(+0x787f1)[0x7fcd44ce47f1]
[rhc001:05345] [ 4] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(opal_finalize_cleanup_domain+0x3e)[0x7fcd44ca897c]
[rhc001:05345] [ 5] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(opal_finalize_util+0x56)[0x7fcd44ca8adc]
[rhc001:05345] [ 6] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-pal.so.0(opal_finalize+0xd3)[0x7fcd44ca8c53]
[rhc001:05345] [ 7] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-rte.so.0(orte_finalize+0x1ba)[0x7fcd455fae0a]
[rhc001:05345] [ 8] /media/sf_rhc/ubuntu/openmpi/build/master/lib/libopen-rte.so.0(orte_daemon+0x40f1)[0x7fcd456317b2]
[rhc001:05345] [ 9] /media/sf_rhc/ubuntu/openmpi/build/master/bin/orted(+0x9a1)[0x55a5679ce9a1]
[rhc001:05345] [10] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7fcd44ff3b97]
[rhc001:05345] [11] /media/sf_rhc/ubuntu/openmpi/build/master/bin/orted(+0x82a)[0x55a5679ce82a]
[rhc001:05345] *** End of error message ***
bash: line 1:  5345 Segmentation fault      (core dumped) orted -mca ess "env
 -mca ess_base_jobid "2250702848" -mca ess_base_vpid 1
-mca ess_base_num_procs "2" -mca plm "rsh" --tree-spawn
-mca orte_parent_uri "2250702848.0;tcp://10.0.2.4,10.0.3.15:54805"
-mca pmix "^s1,s2,cray,isolated"
```

I'm afraid I don't know anything about this new init/finalize system and couldn't find anything explaining it on the wiki. I tried to emulate the other frameworks, but I guess something in the compress framework doesn't quite match the expected behavior?

@ggouaillardet @hjelmn Can someone familiar with this new init/finalize scheme tell me what is wrong?

hjelmn:
@rhc54 There is a bug in the current implementation that is fixed on the ompi cleanup branch. That branch is failing because one of the osc components does something it shouldn't. I will try to get the fix into master sooner rather than later. For now, if you don't use ucx, you can run with #6213.

rhc54:
Afraid that didn't help - still getting same error. It must be something wrong with the way I added the opal_compress framework, but I don't see the problem.

ibm-ompi:

 Gist: https://gist.github.com/bfdd22fd331a5ba64aa441602e3c85ff

ibm-ompi:

 Gist: https://gist.github.com/853a5b65203d5b111d8ef2fda80a94f5

hjelmn:
Usually a crash means something is out of order on the finalize. Can take a look tomorrow.

rhc54:
@ggouaillardet Can you please take a look at this? It's crashing during finalize of a compute node daemon (i.e., it's the orted that crashes, not mpirun). I don't currently have access to a cluster where I can catch a core dump to see why.

ggouaillardet:
@rhc54 I pushed an other commit with two fixes
 - do not call `opal_finalize_pop_domain ()`
 - do close the `rmaps` framework (not sure why that was removed in the first place)
you should be good now

ibm-ompi:

 Gist: https://gist.github.com/3d94583f45ad4f00f4f68c10c26d811c

ibm-ompi:

 Gist: https://gist.github.com/9de5d4b196d79f3962805be17e04f1f3

rhc54:
Current state: non-PMIx jobs launch and run, but we get an error log output from the PMIx gds/hash component which is detecting that the number of nodes listed for a job isn't matching the number of ppn entries. This will prevent PMIx-based jobs from operating. The data is coming from the orte/orted/pmix/pmix_server_register_fns.c call into PMIx_server_register_nspace.

This is only occurring on the orted - mpirun is fine. The root-cause is a mismatch in the data being passed to the orted - there is clearly a bug in orte/util/nidmap.c. The way this code now works is:

- for ssh-based tree launch, we send the nidmap (i.e., the list of node names and daemon vpids) to each layer of the tree. The daemon parses the info and then launches its children.

- for non-ssh based launch, we send the nidmap upon completion of daemon callback using an xcast

- for both types of launch, we send the number of slots, the "slots are given" flag, and the topology for each daemon upon completion of daemon callback. We check to see if all the values are the same for each of those areas - if so, we just send a flag indicating it. Thus, we only send the data if there is an actual difference to report (which is the unusual case).

- when we launch a job, we send the node list and the ppn for each node.

The rest of the procedure remains unchanged. My guess is that there remains a bug in that last stage. However, it requires a multi-node system, which means working on VMs for me - and my VMs are incredibly slow (compiling OMPI takes nearly an hour). Thus, debugging this will take a little while.

ibm-ompi:

ibm-ompi:

ggouaillardet:
Sometimes `mpirun -n 1 ./ring_c` crashes (no `orted` is involved) and some other times it works just fine ...

```
$ ~/debuginfo/ompi-reg/bin/mpirun -n 1 ./ring_c
*** Error in `/home/usersup/gilles/debuginfo/ompi-reg/bin/mpirun': double free or corruption (out): 0x00007ffc2fe9df80 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81499)[0x2b1be0a01499]
/home/usersup/gilles/debuginfo/ompi-reg/lib/libopen-rte.so.0(orte_util_pass_node_info+0x808)[0x2b1bdeff394b]
/home/usersup/gilles/debuginfo/ompi-reg/lib/libopen-rte.so.0(orte_plm_base_daemons_reported+0xb27)[0x2b1bdf03eb34]
/lib64/libevent-2.0.so.5(event_base_loop+0x774)[0x2b1be0329a14]
/home/usersup/gilles/debuginfo/ompi-reg/bin/mpirun[0x40169a]
/home/usersup/gilles/debuginfo/ompi-reg/bin/mpirun[0x40105d]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x2b1be09a2445]
/home/usersup/gilles/debuginfo/ompi-reg/bin/mpirun[0x400f79]
======= Memory map: ========
```

ggouaillardet:
I just pushed a new commit to address the double free issue.

ggouaillardet:
if this is not enough, please tell me your `mpirun` command line

rhc54:
Nope - still see the problem:
```shell
$ mpirun -n 2 -H rhc001,rhc002 hostname
[rhc002:06916] PMIX ERROR: BAD-PARAM in file gds_hash.c at line 222
[rhc002:06916] PMIX ERROR: BAD-PARAM in file gds_hash.c at line 452
rhc002
rhc001
$
```
As per my prior note, the number of nodes isn't matching the number of ppn entries when registering the nspace on the orted. I think you'll need to check the input/output of the nidmap construct/parsers as there seems to be some mismatch.

ggouaillardet:
@rhc54 I just pushed a new commit.

Where do you invoke `mpirun` ? It works for me with two nodes (for example if `mpirun` is invoked on `rhc001`) but I get OOB no route if three nodes are used (for example if `mpirun` is invoked on `rhc003`)

ggouaillardet:
@rhc54 I am not sure of what is going wrong ...

with 3 hosts
- `mpirun --mca routed radix,debruijn ...` fails (daemon 2 has no route to parent = daemon 1)
- `mpirun --mca routed radix ...` works (daemon 2 only talks to parent = daemon 0)
- `mpirun --mca routed debruijn` works too (daemon 2 is able to talk to parent = daemon 1)

rhc54:
That seems odd. It sounds like the parent is being incorrectly identified. What should happen is:

- mpirun sets up the routed component so it knows its direct children. It then launches those daemons, passing its oob contact info on the cmd line as the --parent-uri

- each daemon calls back to mpirun (since that is their parent) and receives back the nidmap (see the node_regex_report function at the end of orte/orted/orted_main.c). The daemon updates its nidmap and then updates its routing plan so it now knows its children in the tree. The daemon then launches its children, again putting its oob contact info on that cmd line.

The key here is that the nidmap must correctly identify which daemon vpids are on which nodes so that the routing plan is correctly computed. I would set plm_base_verbose=5 to see the orted cmd line from each daemon. Typically, I'll set the routed radix value to 1 so the daemons link together in a linear chain - makes it easier to see the tree spawn in action.

ggouaillardet:
It looks like the changes evidence a bug that has always been there ...

with `master`, if you `mpirun -H rhc001,rhc002,rhc003 -n 2 --mca routed radix,debruijn hostname` from `rhc001`, here is what happens on `rhc003` (daemon's vpid is 2)
 - `orte_ess_base_orted_setup()` (indirectly) calls `orte_routed_base_update_routing_plan()`
 - `update_routing_plan()` from `routed/debruijn` sets `ORTE_PROC_MY_PARENT->vpid = 1` (at first glance, that is fishy ...)
- `orte_rml_base_parse_uris()` parse the URI from the command line, __and__ resets `ORTE_PROC_MY_PARENT->vpid = 0` (at first glance, this is good again)
- `report_orted()` sends an `ORTE_RML_TAG_ORTED_CALLBACK` message to the parent (aka vpid __0__, that is reachable since we have its URI on the command line)


now, with this PR

- `orte_rml_base_parse_uris()` parse the URI from the command line and sets `ORTE_PROC_MY_PARENT->vpid = 0`
-  `node_regex_report()` calls `orte_routed_base_update_routing_plan()`
 - `update_routing_plan()` from `routed/debruijn` sets `ORTE_PROC_MY_PARENT->vpid = 1` (at first glance, that is fishy ...)
- `report_orted()` sends an `ORTE_RML_TAG_ORTED_CALLBACK` message to the parent (aka vpid __1__, that is __not__ reachable since we only have the URI for vpid 0 on the command line)


Bottom line, I think there is something fishy with `routed/debruijn`, and that is not coming from this PR.
I do not know the right fix, and for the time being, I suggest we simply blacklist this component.

hjelmn:
@ggouaillardet Not surprised. The component has not be maintained it some time. Shouldn't be too difficult to fix but I won't have time to look until next week. If you have any questions let me know.

ggouaillardet:
Thanks @hjelmn. I noted `owner.txt` states this component is unmaintained. Is simply removing this component an option ?

rhc54:
Let's just make it not be the default for now - let radix be the default again. If/when we get a chance to fix it, we can bring it back. Note that @bwbarrett had noted that we were failing to launch multi-layer trees but hadn't had time to track down the reason.

We need to check, however, that radix is actually working. Remember that the radix defaults to 64, so you need to set the radix to 1 to get a multi-layer launch with small numbers of nodes. From your prior comments, I'm gather that works - just pointing this out to ensure we don't forget and mislead ourselves.

ggouaillardet:
`routed/radix` priority is `70` and `routed/debruijn` priority is `30`. The thing is __both__ are _used_ by default, so `routed/debruijn` has to be blacklisted.

rhc54:
Not sure I understand - how can both be active at the same time? The orted's all have to agree on the active routed component, so while all might be considered, only one can be active - yes?

ggouaillardet:
I hardly understand the routing part, but here is what I did observe when running `orted` under the debugger

```c
void orte_routed_base_update_routing_plan(char *module)
{
    orte_routed_base_active_t *active;

    OPAL_LIST_FOREACH(active, &orte_routed_base.actives, orte_routed_base_active_t) {
        if (NULL == module ||
            0 == strcmp(module, active->component->base_version.mca_component_name)) {
            if (NULL != active->module->update_routing_plan) {
                active->module->update_routing_plan();
            }
        }
    }

    return;
}
```

`module` was `NULL`, so the `update_routing_plan()` functions from __both__ `radix` and `debruijn` were invoked.

rhc54:
Ouch! Okay, this is leftover cruft that is my fault. I'll fix it this morning.

rhc54:
I think I have this about cleaned up - should push something soon. One thing to note, however: although we updated the routing plan of all active components, only the highest priority component was actually being used. The others were being ignored. So I believe this is most likely a red herring in terms of diagnosing the problem.

rhc54:
Okay, I have this up-and-running now! @ggouaillardet Can you please give it a smoke test just to be sure?

karasevb:
@rhc54 
Will it be included in v4.0.0 release?

rhc54:
@karasevb v4.0.0 has already been released. This would be included in v5.0.0 scheduled for this summer. This change is too large to go back to v4.0.1, but you could PR your change to the v4.0.x series as it resolves a currently known issue.

karasevb:
@rhc54 Thanks. Will be done.

karasevb:
Refs. https://github.com/open-mpi/ompi/pull/6375

PR #6353, opened 2019-02-03
Fix the PVAR allocation usage.

bosilca:
According to the MPI standard the obj_handle is a pointer to an MPI
object, and therefore cannot be MPI_COMM_WORLD. The MPI standard example
14.6 highlight this usage.


jsquyres:
@bosilca Did you forget to merge this one?

PR #6355, opened 2019-02-04
fix problem in pmix for osx

hppritcha:
at least for me on my OS-x box (running Yosemite) PMIx server was failing to compile without this patch.


rhc54:
That is really odd - it builds fine on my Mac, but I'm running Mojave at this point. We obviously don't want an OPAL reference down in the PMIx code, so I'll add something equivalent to the PMIx side.

PR #6357, opened 2019-02-05
v3.1.x: Add -g compilation flag for all files that are present in the stack when attaching with MPIR

James-A-Clark:
This PR adds the `-g` compilation flag for all files that are present in the stack from the `MPI_Init()` call. This is so that when a debugger attaches using MPIR, it can step out of this stack back into main. This cannot be done with certain aggressive optimisations and missing debug information.

This issue appeared when OpenMPI 3.1.2 was built with GCC 7.3.0 on Power8 with the following configuration line:
` ./configure --enable-mpirun-prefix-by-default`

The stack that we get when attaching to the user process is this:
```
#0  pthread_cond_wait@@GLIBC_2.17 () from /lib64/libpthread.so.0
#1  pmix2x_client_init (...) at pmix2x_client.c:151
#2  rte_init () at ess_pmi_module.c:130
#3  orte_init (...) at runtime/orte_init.c:271
#4  ompi_mpi_init (...) at runtime/ompi_mpi_init.c:513
#5  PMPI_Init (...) at pinit.c:66
#6  main (...) at hello.c:22
```

And without this patch, GDB can't unwind because of missing unwind information and an optimisation that puts a branch in the function preamble of the `orte_init()` function. Looking like this:
```
#0  pthread_cond_wait@@GLIBC_2.17 () from /lib64/libpthread.so.0
#1  pmix2x_client_init () from /software/mpi/openmpi-3.1.2_gnu-7.3.0/lib/openmpi/mca_pmix_pmix2x.so
#2  rte_init () from /software/mpi/openmpi-3.1.2_gnu-7.3.0/lib/openmpi/mca_ess_pmi.so
#3  orte_init () from /software/mpi/openmpi-3.1.2_gnu-7.3.0/lib/libopen-rte.so.40
Backtrace stopped: frame did not save the PC
```

The user impact is that after attaching using MPIR, they can't get back to their call to `MPI_Init()`. This can be fixed minimally by compiling with `-fasynchronous-unwind-tables`, but the existing pattern of using `-g` for MPIR files (which implies unwind tables) was copied from the file `orte/orted/Makefile.am` and applied to other files in this stack.

This issue is occurring on 3.1.x and up, but I'm not sure about the branching model for PRs. Would I also have to submit a PR for future version branches? Or do changes to 3.1.x get automatically pulled in?

Thanks

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -18,6 +18,8 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (C) 2018      Mellanox Technologies, Ltd. 
  *                         All rights reserved.
+ * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -1920,14 +1922,14 @@ int opal_hwloc_base_cset2str(char *str, int len,
         for (core_index = 0; core_index < num_cores; ++core_index) {
             if (map[socket_index][core_index] > 0) {
                 if (!first) {
-                    strncat(str, ", ", len - strlen(str));
+                    strncat(str, ", ", len - strlen(str) - 1);
                 }
                 first = false;
 
                 snprintf(tmp, stmp, "socket %d[core %d[hwt %s]]",
                          socket_index, core_index,
                          bitmap2rangestr(map[socket_index][core_index]));
-                strncat(str, tmp, len - strlen(str));
+                strncat(str, tmp, len - strlen(str) - 1);
             }
         }
     }
@@ -1983,7 +1985,7 @@ int opal_hwloc_base_cset2mapstr(char *str, int len,
     for (socket = hwloc_get_obj_by_type(topo, HWLOC_OBJ_SOCKET, 0);
          NULL != socket;
          socket = socket->next_cousin) {
-        strncat(str, "[", len - strlen(str));
+        strncat(str, "[", len - strlen(str) - 1);
 
         /* Iterate over all existing cores in this socket */
         core_index = 0;
@@ -1995,7 +1997,7 @@ int opal_hwloc_base_cset2mapstr(char *str, int len,
                                                         socket->cpuset,
                                                         HWLOC_OBJ_CORE, ++core_index)) {
             if (core_index > 0) {
-                strncat(str, "/", len - strlen(str));
+                strncat(str, "/", len - strlen(str) - 1);
             }
 
             /* Iterate over all existing PUs in this core */
@@ -2010,13 +2012,13 @@ int opal_hwloc_base_cset2mapstr(char *str, int len,
 
                 /* Is this PU in the cpuset? */
                 if (hwloc_bitmap_isset(cpuset, pu->os_index)) {
-                    strncat(str, "B", len - strlen(str));
+                    strncat(str, "B", len - strlen(str) - 1);
                 } else {
-                    strncat(str, ".", len - strlen(str));
+                    strncat(str, ".", len - strlen(str) - 1);
                 }
             }
         }
-        strncat(str, "]", len - strlen(str));
+        strncat(str, "]", len - strlen(str) - 1);
     }
 
     return OPAL_SUCCESS;


jsquyres:
I'm still a little uncertain for the need for this patch.  Don't debuggers offer other / different ways to get out of `MPI_INIT`, even if the symbols aren't known?

E.g., set a breakpoint out in main and "continue" until execution gets there.

I ask for a few reasons:

1. Having users step through code in the deep, deep magic of `MPI_INIT` is... a little frightening.
1. The exact stack of code that is executed in `MPI_INIT` is highly system-dependent.  Different components can/will be selected on different machines and different environments.  Adding `-g` to this particular stack where you ran into the issue may change a) over time, and b) on different machines.

jsquyres:
ok to test

James-A-Clark:
@rhc54 the -g flag on its own shouldn't impact runtime. The binary size will increase slightly for these compilation units, but those affected sections of the ELF will only be loaded by debuggers, not during normal execution. 

Optimisation flags also seem to be appended to the compile line after the debugger flags. So even if `$(DEBUGGER_CFLAGS)` ever included an `-O0` it would be overwritten by a following `-O` option. The full compile line for  `orted_submit.c` looks like this, with an `-O3` following the `-g`:
```
/opt/gnu/7.3.0/bin/gcc -DHAVE_CONFIG_H -I. -I../opal/include -I../ompi/include -I../oshmem/include -I../opal/mca/hwloc/hwloc1117/hwloc/include/private/autogen -I../opal/mca/hwloc/hwloc1117/hwloc/include/hwloc/autogen -I../ompi/mpiext/cuda/c -I.. -I../orte/include -I/home/jclark/ompi/opal/mca/event/libevent2022/libevent -I/home/jclark/ompi/opal/mca/event/libevent2022/libevent/include -I/home/jclark/ompi/opal/mca/hwloc/hwloc1117/hwloc/include -DNDEBUG -frecord-gcc-switches -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -fno-strict-aliasing -pthread -g -O3 -DNDEBUG -Wall -Wundef -Wno-long-long -Wsign-compare -Wmissing-prototypes -Wstrict-prototypes -Wcomment -pedantic -Werror-implicit-function-declaration -finline-functions -fno-strict-aliasing -pthread -MT orted/liborted_mpir_la-orted_submit.lo -MD -MP -MF orted/.deps/liborted_mpir_la-orted_submit.Tpo -c orted/orted_submit.c  -fPIC -DPIC -o orted/.libs/liborted_mpir_la-orted_submit.o
```

@jsquyres This isn't about symbols exactly, to fix this bug we only really need async unwind information which is also used for exceptions. I chose to use the `-g` from `$(DEBUGGER_CFLAGS)` instead of `-fasynchronous-unwind-tables` because I wasn't sure of an easy way of making the latter portable and it was already done for other MPIR files in the build system.

The reason that we can't set a breakpoint in main and run to it is because, as you can see from the stack, GDB doesn't know where in main we started from. The user would normally have to parse which line in main this stack starts from, and then set a breakpoint on the line after. But with this bug, GDB is unable to show this info and stops unwinding at `orte_init()`.

> Having users step through code in the deep, deep magic of MPI_INIT is... a little frightening.
The exact commands that you would issue to GDB when using MPIR to attach would be:
```
frame 0
finish
```
Which would tell GDB to step-out up the stack until frame 0 was reached and bring you back to user code. The user wouldn't have to do any stepping in MPI internals per say.

> The exact stack of code that is executed in MPI_INIT is highly system-dependent. Different components can/will be selected on different machines and different environments. Adding -g to this particular stack where you ran into the issue may change a) over time, and b) on different machines.
That's a good point. Initially I only added the flag to the one frame that GDB was having a problem unwinding out of. But we thought that we should be more thorough and add it to everything we saw in the stack in case a compiler change in the future optimises the link register saving out of other functions.

We're only worried about the stack that leads up to the MPIR holding point, and not everything under MPI_INIT. Do you think that stack would still be platform/environment dependent?

James-A-Clark:
Tagging @shamisp

James-A-Clark:
@rhc54 Is there a way that we can test this patch to make sure that it doesn't impact performance?

The `-g` flag was only added to a handful of files with a few functions in them each, which mostly seem related to startup and not runtime features. Hopefully this minimises the impact.

rhc54:
I don't have a way to do it myself, but I also agree with you that I can't see how this would impact application performance. I only raised the question in case someone out there in the MPI team sees something we don't 😄

jsquyres:
What about a different approach: test in `configure` to see if `-fasynchronous-unwind-tables` works.

I assume we wouldn't want to add that flag *everywhere*, but we could sprinkle it around in the Right Places (potentially in, or as a supplement to `$(DEBUGGER_CFLAGS)`...?  I'd have to cache back in exactly what/how `DEBUGGER_CFLAGS` is used...).

Specifically: `configure` can check if that flag works.  If it is, add it to something that gets used in relevant `Makefile.am`s.  If not, the value that gets used in relevant `Makefile.am`s is empty (or just doesn't contain that flag).

James-A-Clark:
I've updated the PR to add a configure check for the compiler version and to add only the async unwind flag. I think there are still a few open questions:

* Where do we want to put the flags? I've assumed everything in the orte/mca/ess folder and added it to the toplevel makefile. But some of the subfolders overwrite CFLAGS, so I had to add it again manually to pmi/Makefile. Do we need to do this to any of the other subfolders?
* For the IBM compiler, although we're not seeing the issue currently, there also doesn't seem to be an equivalent flag. So I've added `-g`. The alternative would be no flag.

rhc54:
@jsquyres Can we move this forward?

config/orte_setup_debugger_flags.m4
```diff
@@ -53,4 +53,26 @@ AC_DEFUN([ORTE_SETUP_DEBUGGER_FLAGS],[
...
+        MPIR_UNWIND_CFLAGS="-g"
+    elif test "x$opal_cv_c_compiler_vendor" = "xgnu"; then
+        MPIR_UNWIND_CFLAGS="-fasynchronous-unwind-tables"
+    elif test "x$opal_cv_c_compiler_vendor" = "xintel"; then
+        MPIR_UNWIND_CFLAGS="-fasynchronous-unwind-tables"
+    elif test "x$opal_cv_c_compiler_vendor" = "xportland group"; then
+        MPIR_UNWIND_CFLAGS="-Meh_frame -Mframe"
+    else
+        MPIR_UNWIND_CFLAGS=
+    fi
```
jsquyres: It might be better to actually try the flags in question, rather than assume that they work for a given compiler vendor.  E.g., perhaps something (warning: typed directly into github -- might not be 100% correct):

```
dnl Check to see if specific CFLAGS work
dnl $1: compiler flags to check
dnl $2: Action if the flags work
dnl $3: Action if the flags do not work
AC_DEFUN([_ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS],[
    OPAL_VAR_SCOPE_PUSH([ORTE_SETUP_DEBUGGER_FLAGS_CFLAGS_save])
    ORTE_SETUP_DEBUGGER_FLAGS_CFLAGS_save=$CFLAGS
    AC_MSG_CHECKING([if $1 compiler flag works])
    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([],[int i = 3;])],
        [$2; AC_MSG_RESULT([yes])],
        [$3; AC_MSG_RESULT([no])])
    CFLAGS=$ORTE_SETUP_DEBUGGER_FLAGS_CFLAGS_save
    OPAL_VAR_SCOPE_POP
])
```

Then where you have this current `if` block checking compiler vendors, instead do something like:

```
    _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-fasynchronous-unwind-tables],
        [MPIR_UNWIND_CFLAGS=-fasynchronous-unwind-tables],
        [_ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-Meh_frame -Mframe],
            [MPIR_UNWIND_CFLAGS="-Meh_frame -Mframe"],
            [MPIR_UNWIND_CFLAGS=-g])
        ])
```

Waving my hands here a little -- try something along these lines (it doesn't have to be exactly this).  My point is the assuming values based on checking vendors and/or versions is inherently imprecise.  It's better to actually just test for the behavior and see if it works.

orte/mca/ess/Makefile.am
```diff
@@ -17,7 +17,9 @@
 # $HEADER$
 #
 
-AM_CPPFLAGS = $(LTDLINCL)
+# Add unwind flags because all of the files in this tree are
+# involved in startup.
+AM_CPPFLAGS = $(LTDLINCL) $(MPIR_UNWIND_CFLAGS)
```
jsquyres: CPPFLAGS are different than CFLAGS.  I think you want to set `AM_CFLAGS`.

Also, setting the value here in this `Makefile.am` will propagate it to `orte/mca/ess/base`, but not to the individual components.  You might need to put it in each of the ess components' `Makefile.am`s.  ☹️ 

I see you edited `pmi/Makefile.am` -- do you need to also do the others for completeness?

orte/mca/ess/pmi/Makefile.am
```diff
@@ -11,7 +11,7 @@
 # $HEADER$
 #
 
-AM_CPPFLAGS = $(ess_pmi_CPPFLAGS)
+AM_CPPFLAGS = $(ess_pmi_CPPFLAGS) $(MPIR_UNWIND_CFLAGS)
```
jsquyres: Similar comment as above -- I think you want to use `AM_CFLAGS`.

orte/mca/ess/Makefile.am
rhc54: I don't think this need be done elsewhere - only the ess/pmi component is used by application procs. The other components are solely used by daemons and thus are not accessible by the debugger.

jsquyres: Ok, cool.  Might be worth a comment in the Makefile.am so that we don't have this conversation again in 6 months.

James-A-Clark:
Thanks for the feedback. I've made the changes to detect which flags work, rather than doing it in compiler name. And also done the other CPPFLAGS -> CFLAGS fixes.

I didn't manage to do the flag detection with nested calls to _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS, that doesn't seem to be an option. But I did manage to do it with an extra "foundFlags" variable.

config/orte_setup_debugger_flags.m4
```diff
@@ -53,4 +68,31 @@ AC_DEFUN([ORTE_SETUP_DEBUGGER_FLAGS],[
...
+    # Test for PGI compiler flags
+    if test $foundFlags = no; then
+        _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-Meh_frame -Mframe],
+            [MPIR_UNWIND_CFLAGS="-Meh_frame -Mframe" foundFlags=yes])
+    fi
+
+    # Finally fall back to -g if neither of the above were available
+    if test $foundFlags = no; then
+        _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-g],
+            [MPIR_UNWIND_CFLAGS=-g foundFlags=yes])
```
jsquyres: Minor nit: `foundFlags=yes` is unnecessary here.

config/orte_setup_debugger_flags.m4
```diff
@@ -24,6 +24,21 @@ dnl
 dnl $HEADER$
 dnl
 
+dnl Check to see if specific CFLAGS work
+dnl $1: compiler flags to check
+dnl $2: Action if the flags work
```
jsquyres: Curiosity question: did you not like the 3-var variant that was proposed?  (i.e., `$3` = action if flags do not work).  I ask because this is a common mechanism (some test, and then "do this if it works" / "do this if it doesn't work").  I'm not strictly tied to this, but we do use that convention all throughout our m4.

Such an approach lets you have a compact form down below:

```
_ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-fasynchronous-unwind-tables],
        [MPIR_UNWIND_CFLAGS=-fasynchronous-unwind-tables],
        [_ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-Meh_frame -Mframe],
            [MPIR_UNWIND_CFLAGS="-Meh_frame -Mframe"],
            [MPIR_UNWIND_CFLAGS=-g])
        ])
```

config/orte_setup_debugger_flags.m4
```diff
@@ -53,4 +68,31 @@ AC_DEFUN([ORTE_SETUP_DEBUGGER_FLAGS],[
...
+    if test $foundFlags = no; then
+        _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-Meh_frame -Mframe],
+            [MPIR_UNWIND_CFLAGS="-Meh_frame -Mframe" foundFlags=yes])
+    fi
+
+    # Finally fall back to -g if neither of the above were available
+    if test $foundFlags = no; then
+        _ORTE_SETUP_DEBUGGER_FLAGS_TRY_CFLAGS([-g],
+            [MPIR_UNWIND_CFLAGS=-g foundFlags=yes])
+    fi
```
jsquyres: Minor nit: could you OPAL_VAR_SCOPE_PUSH / OPAL_VAR_SCOPE_POP for `foundFlags`?

Those macros do a few things:

1. Make sure `foundFlags` is not already set at the beginning (i.e., make sure we're not causing a side effect now or later)
1. Unset `foundFlags` at the end

config/orte_setup_debugger_flags.m4
James-A-Clark: I tried this one a couple of different ways, and always ended up with a syntax error in the generated configure script. I assumed that nesting functions wasn't supported. 

This also seems to suggest the same: http://www.slac.stanford.edu/comp/unix/package/rtems/doc/html/autoconf/autoconf.info.Multiple_Cases.html

config/orte_setup_debugger_flags.m4
James-A-Clark: Fixed

config/orte_setup_debugger_flags.m4
James-A-Clark: Fixed

config/orte_setup_debugger_flags.m4
jsquyres: m4 is a fickle beast.  It's like a bully on the playground: if you give it all your lunch money, it might not beat you up.  😄 

m4 nesting is fully supported.  You just have to be careful about side effects -- remembering that m4 macros are *expanded*, not *called*.  My original suggestion typed here in Github wasn't quite correct.  I pushed a commit to this PR that seems to work.  What do you think?

I also rebased this PR on top of the current v3.1.x branch to get rid of the orte_rml_ofi warnings during `autogen.pl` (we fixed those in a different PR).

James-A-Clark: Thanks for the help, this seems much better. I've rebased it onto master and made a new PR: https://github.com/open-mpi/ompi/pull/6527

rhc54:
ok to test

jsquyres:
Ok, #6527 merged (the master version of this PR).  Could you refresh this PR as a cherry-pick from that PR?  And then file corresponding PRs for v3.0.x and v4.0.x?

Thank you!

James-A-Clark:
Thanks Jeff, I made the new PRs but ended up making a new one for 3.1.x because it seemed cleaner.

PRs: 
https://github.com/open-mpi/ompi/pull/6554
https://github.com/open-mpi/ompi/pull/6553
https://github.com/open-mpi/ompi/pull/6551

PR #6361, opened 2019-02-06
mtl/ofi: Fix segfault when not using Thread-Grouping feature

aravindksg:
For the non thread-grouping paths, only the first (0th) OFI context
should be used for communication. Otherwise, this would access a non-existent
array item and causes segfaults.

### DIFF
--- a/ompi/mca/mtl/ofi/README
+++ b/ompi/mca/mtl/ofi/README
@@ -72,7 +72,7 @@ by reducing the bits available for the communicator ID field in the OFI tag.
 
 SCALABLE ENDPOINTS:
 -------------------
-OFI MTL supports OFI Scalable Endpoints feature as a means to improve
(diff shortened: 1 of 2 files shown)


bwbarrett:
bot:ompi:retest  the "not in tree" issue strikes again.

aravindksg:
@bwbarrett : Have addressed your concern in the new patch. Also, just to note here that I am not adding a specialized function at this point for SEP or Thread-grouping feature. I tested the impact of the branch and it seems negligible. In the future, if there is a further divergence in code path based on SEP enabled or not, a specialized function can be added then.

bwbarrett:
@aravindksg are you ok with me merging this commit?

aravindksg:
> @aravindksg are you ok with me merging this commit?

Sure, thanks @bwbarrett

PR #6375, opened 2019-02-11
v4.0.x/regex: fixed host ordering for different prefixes

karasevb:
Corresponds to https://github.com/open-mpi/ompi/pull/6306

### DIFF
--- a/orte/mca/regx/fwd/regx_fwd.c
+++ b/orte/mca/regx/fwd/regx_fwd.c
@@ -154,38 +154,25 @@ static int nidmap_create(opal_pointer_array_t *pool, char **regex)
         }
         /* is this node name already on our list? */
         found = false;
-        for (item = opal_list_get_first(&nodenms);
-             !found && item != opal_list_get_end(&nodenms);
-             item = opal_list_get_next(item)) {
-            ndreg = (orte_regex_node_t*)item;
-            if (0 < strlen(prefix) && NULL == ndreg->prefix) {
-                continue;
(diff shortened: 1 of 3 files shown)


gpaulsen:
@hppritcha, I think we should merge into v4.0.1

rhc54:
@jjhursey It occurs to me that we now are getting different answers from PMIx_resolve_node and friends then we get from OMPI. IIRC, you had users that were trying to utilize the PMIx functions. Do we need this change (and your "reverse" component) backported to PMIx?

jjhursey:
As long as `PMIx_resolve_node` preserves ordering (which it does now in PMIx) then we are ok. From the `master` [testing](https://github.com/open-mpi/ompi/pull/6306#issuecomment-458800769) on this PR @karasevb verified that this patch fixes that case for OMPI too when using the forward regex component.

The reverse component was helpful in making a more compact representation for the customer than the forward component. So we are ok with the forward component as long as it works correctly, which this PR seems to address for OMPI. For PMIx, it would be nice to bring over the reverse functionality to improve compression for those hostnames, but if we are moving to a more general technique in PMIx then we can just wait for that.

rhc54:
Interesting - I wonder what the difference is between the PMIx and OMPI regex components? I thought they were just copies of each other.

Yeah, we'll be moving to compression in PMIx as well. I also have a skeleton for a new regex parser that doesn't distinguish between alpha and numeric patterns, but that will have to come later.

PR #6378, opened 2019-02-12
Fix inherited_linker_flags processing when building with NAG compiler

skosukhin:
If we build OpenMPI with a combination of GCC and NAG and need to link to an external Libtool library that has `-pthread` in the `inherited_linker_flags` variable, we might get the following problem. For example, we link to an external hwloc, and `libhwloc.la` contains the mentioned `-pthread` entry.  Apparently, this particular problem was addressed with a [patch for libtool](https://github.com/open-mpi/ompi/blob/c5709663b6acc5b8c08295b06e6c4eda3d58da41/config/ltmain_nag_pthread.diff). The patch changes ` -pthread`, which NAG compiler does not understand to ` -Wl,-pthread`, which NAG does understand. Now, `libmpi_mpifh.la` is linked with the C compiler (gcc in this case), which (conditionally) requires a convenience library `libmpi_mpifh_sizeof.la`, which is linked with the Fortran (nagfor) compiler. As a result, `libmpi_mpifh_sizeof.la` inherits the `-pthread` flag from `libhwloc.la` but stores it in the modified form as `-Wl,-pthread`. In turn, `libmpi_mpifh.la` inherits `-Wl,-pthread`, but it is linked with gcc, which interprets this as a flag to be passed to ld, and for ld, of course, `-pthread` is an invalid flag. So, we get an error.

There are two solutions (with variations) that I found for this problem. The first one, which this PR offers, is fast and simple: update the patch, so Libtool would revert `-Wl,-pthread` back to `-pthread`. The drawback of this solution is that `libmpi_usempi.la`, which is linked with the Fortran compiler, would have the original `-pthread` entry but this would also introduce the consistency between both of the Fortran interface libraries. Given that most of the users would use the MPI wrappers for linking to the MPI libraries, the contents of the `*.la` files that correspond to them are not so important (at least they are correct since the original `-pthread` inherited from hwloc remains unchanged).

The second possible solution questions the necessity of the convenience library `libmpi_mpifh_sizeof.la`. The [comment](https://github.com/open-mpi/ompi/blob/c5709663b6acc5b8c08295b06e6c4eda3d58da41/ompi/mpi/fortran/mpif-h/Makefile.am#L87-L106) says that the reason for having this library is to enforce usage of the C compiler for linking. But, as far as I know, there is a more [straightforward way](https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html#How-the-Linker-is-Chosen) to do this. [Here is the implementation of that approach](https://github.com/skosukhin/ompi/commit/a35cfc3e4721d05717f1c634516e1f5e61710ce3). The only drawback of this solution that I found is that the silent rule reports `GEN libmpi_mpifh.la` instead of `CCLD libmpi_mpifh.la` but the actual actions are the same. Another way of doing almost the same but keep the message intact is to declare `libmpi_mpifh_sizeof.la` as an `EXTRA` library and link `libmpi_mpifh.la` directly to `sizeof_f.lo`. [Here is the implementation of this](https://github.com/skosukhin/ompi/commit/ca733f9f2250e40aef28b03d6c13ee5ea72155cc).

If you are interested at all, please, let me know which of the solutions you prefer and I will update the branch of this PR with the solution of choice.

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -1038,10 +1038,28 @@ sub patch_autotools_output {
 	  whole_archive_flag_spec${tag}=
 	  tmp_sharedflag='--shared' ;;
 	nagfor*)			# NAGFOR 5.3
-	  tmp_sharedflag='-Wl,-shared';;
+	  tmp_sharedflag='-Wl,-shared' ;;
 	xl";
 
-        push(@verbose_out, $indent_str . "Patching configure for NAG compiler ($tag)\n");
+        push(@verbose_out, $indent_str . "Patching configure for NAG compiler #1 ($tag)\n");
+        $c =~ s/$search_string/$replace_string/;
+
+        # Newer versions of Libtool have the previous patch already. Therefore,
+        # we add the support for convenience libraries separetly
+        my $search_string = "whole_archive_flag_spec${tag}=" . '\n\s+' .
+            "tmp_sharedflag='--shared' ;;" . '\n\s+' .
+            'nagfor\052.*# NAGFOR 5.3\n\s+' .
+            "tmp_sharedflag='-Wl,-shared' ;;" . '\n\s+' .
+            'xl';
+        my $replace_string = "whole_archive_flag_spec${tag}=
+	  tmp_sharedflag='--shared' ;;
+        nagfor*)                        # NAGFOR 5.3
+          whole_archive_flag_spec${tag}='\$wl--whole-archive`for conv in \$convenience\\\"\\\"; do test  -n \\\"\$conv\\\" && new_convenience=\\\"\$new_convenience,\$conv\\\"; done; func_echo_all \\\"\$new_convenience\\\"` \$wl--no-whole-archive'
+          compiler_needs_object=yes
+          tmp_sharedflag='-Wl,-shared' ;;
+	xl";
+
+        push(@verbose_out, $indent_str . "Patching configure for NAG compiler #2 ($tag)\n");
         $c =~ s/$search_string/$replace_string/;
     }
 
--- a/config/ltmain_nag_pthread.diff
+++ b/config/ltmain_nag_pthread.diff
@@ -1,18 +1,40 @@
 --- config/ltmain.sh
 +++ config/ltmain.sh
-@@ -6417,8 +6417,14 @@
- 	func_source "$lib"
-
+@@ -7862,6 +7862,13 @@ func_mode_link ()
  	# Convert "-framework foo" to "foo.ltframework"
-+        # and "-pthread" to "-Wl,-pthread" if NAG compiler
  	if test -n "$inherited_linker_flags"; then
--	  tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g'`
-+          case "$CC" in
-+            *nagfor*)
-+	      tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g' | $SED 's/-pthread/-Wl,-pthread/g'`;;
-+            *)
-+	      tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g'`;;
-+          esac
+ 	  tmp_inherited_linker_flags=`$ECHO "$inherited_linker_flags" | $SED 's/-framework \([^ $]*\)/\1.ltframework/g'`
++
++	  # Additionally convert " -pthread" to " -Wl,-pthread" for nagfor
++	  func_cc_basename $CC
++	  case $func_cc_basename_result in
++	    nagfor*) tmp_inherited_linker_flags=`$ECHO "$tmp_inherited_linker_flags" | $SED 's/ -pthread/ -Wl,-pthread/g'` ;;
++	  esac
++
  	  for tmp_inherited_linker_flag in $tmp_inherited_linker_flags; do
  	    case " $new_inherited_linker_flags " in
  	      *" $tmp_inherited_linker_flag "*) ;;
+@@ -8881,7 +8888,8 @@ func_mode_link ()
+ 	  xlcverstring="$wl-compatibility_version $wl$minor_current $wl-current_version $wl$minor_current.$revision"
+ 	  verstring="-compatibility_version $minor_current -current_version $minor_current.$revision"
+           # On Darwin other compilers
+-          case $CC in
++          func_cc_basename $CC
++          case $func_cc_basename_result in
+               nagfor*)
+                   verstring="$wl-compatibility_version $wl$minor_current $wl-current_version $wl$minor_current.$revision"
+                   ;;
+@@ -9493,6 +9501,13 @@ EOF
+ 	  ;;
+       esac
+ 
++      # Time to revert the changes made for nagfor.
++      func_cc_basename $CC
++      case $func_cc_basename_result in
++        nagfor*)
++          new_inherited_linker_flags=`$ECHO " $new_inherited_linker_flags" | $SED 's% -Wl,-pthread% -pthread%g'` ;;
++      esac
++
+       # move library search paths that coincide with paths to not yet
+       # installed libraries to the beginning of the library search list
+       new_libs=


jsquyres:
@skosukhin Thanks for the patch!  Can you sign off your commit, as per https://github.com/open-mpi/ompi/blob/master/.github/CONTRIBUTING.md#contributors-declaration?

jsquyres:
@skosukhin Your approaches do look simpler / better.  Let me think about this a bit.

@ThemosTsikas This involves your compiler, so: FYI.  😄

jsquyres:
FWIW: @cagoelz just confirmed to me that Absoft did, in fact, fix the issue (way back in 2014) that I mentioned in the [comment](https://github.com/open-mpi/ompi/blob/c5709663b6acc5b8c08295b06e6c4eda3d58da41/ompi/mpi/fortran/mpif-h/Makefile.am#L87-L106) that @skosukhin cited.

skosukhin:
I have added one more patch for NAG fixing injection of libtool convenience libraries into shared ones. This might be useful if the linking of the Fortran library is done with Fortran compiler in the future.

awlauria:
@skosukhin is this ready for review?

skosukhin:
@awlauria yes, I think so

awlauria:
bot:ompi:retest

gpaulsen:
It looks fine to me, but I have no way to test with the NAG compiler.

jsquyres:
I have no way to testing this, either.

@ThemosTsikas Does anyone at NAG want to test this PR?  It's a fairly involved patch for Open MPI specifically for the NAG compiler.

ThemosTsikas:
Hello, thanks for the heads up. 

I tried it and it seems to be working ok. I get a division by zero in sone of the PnetCDF ptests, though.

jsquyres:
@ThemosTsikas Ok, thanks for the confirmation.

Is the div/zero issue a new issue, or related to this one?

ThemosTsikas:
I can't tell yet. I might find some time to investigate. It only seems to happen for 4 ranks and up.

jsquyres:
@ThemosTsikas Any news on this?

ThemosTsikas:
Last time I checked, libtool has no active maintainer. I don't see the point in supporting abandonware. 

Openmpi 4.1.0 seems to build ok with the NAG Fortran compiler.

jsquyres:
@ThemosTsikas For better or for worse, lots of things build with GNU Libtool, including Open MPI.  We're unlikely to change away from it any time soon.

I'm glad v4.1.x builds ok with NAG.  Did you check the use case that the original PR author cited in the description, perchance?

ThemosTsikas:
Checked the Parallel-NetCDF nagfor build with --enable-shared. 

```
make[2]: Entering directory '/2TB-disk/themos/GIT-projects/PnetCDF/src/libs'
  GEN      libpnetcdf.la

*** Warning: Linking the shared library libpnetcdf.la against the
*** static library /usr/local/lib/NAG_Fortran/libf70rts.a is not portable!
/usr/bin/ld: unrecognized option '-pthread'
/usr/bin/ld: use the --help option for usage information
make[2]: *** [Makefile:573: libpnetcdf.la] Error 1

```
Looks like libtool(1) feeding ld(1) bad options, no evidence that nagfor is involved.

jsquyres:
Is that what this PR addresses?

gpaulsen:
@skosukhin ?

skosukhin:
The reported problem manifests itself at the build time of `OpenMPI`, not at the build time of a package that uses it. A trigger for the problem is `inherited_linker_flags=' -pthread'` in the `*.la` file of a library that `OpenMPI` depends on. The following script reproduces the problem:
```bash
#!/bin/bash

set -e

# Build directory:
build_dir="$(pwd)/build"

# Install directory:
install_dir="$(pwd)/install"

# Set up the building environment:
export CC=gcc
export CXX=g++
export FC=nagfor
export F77=nagfor
export PKG_CONFIG_PATH="${install_dir}/lib/pkgconfig:${PKG_CONFIG_PATH}"
export PATH="${install_dir}/bin:${PATH}"
export LD_LIBRARY_PATH="${install_dir}/lib:${LD_LIBRARY_PATH}"

# Create the build directory:
mkdir -p "${build_dir}" && cd "${build_dir}"

# Install xz:
( wget https://tukaani.org/xz/xz-5.2.5.tar.gz
  tar xvf xz-5.2.5.tar.gz && cd xz-5.2.5
  ./configure --prefix="${install_dir}" && make -j8 install )

# Install libxml2:
( wget ftp://xmlsoft.org/libxml2/libxml2-2.9.10.tar.gz
  tar xvf libxml2-2.9.10.tar.gz && cd libxml2-2.9.10
  ./configure --prefix="${install_dir}" --without-python
  make -j8 install )

# Install hwloc:
( wget https://download.open-mpi.org/release/hwloc/v2.4/hwloc-2.4.1.tar.gz
  tar xvf hwloc-2.4.1.tar.gz && cd hwloc-2.4.1
  ./configure --prefix="${install_dir}"
  make -j8 install )

# Install OpenMPI:
( wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.0.tar.gz
  tar xvf openmpi-4.1.0.tar.gz && cd openmpi-4.1.0
  ./configure --prefix="${install_dir}" --with-hwloc="${install_dir}"
  make -j8 install )
```
The result is:
```console
  CCLD     libmpi_mpifh.la
/usr/bin/ld: unrecognized option '-pthread'
/usr/bin/ld: use the --help option for usage information
Makefile:2858: recipe for target 'libmpi_mpifh.la' failed
```

The problem disappeared with this PR:
```bash
# ...
# Same as above
# ...
# Install OpenMPI:
( wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.0.tar.gz
  wget https://github.com/open-mpi/ompi/commit/ad16407c6a79c457ec3bc4251234af4281faa893.patch
  wget https://github.com/open-mpi/ompi/commit/3b624dbc1b4d6efb58b9ce7e54dc4acf19043e88.patch
  tar xvf openmpi-4.1.0.tar.gz && cd openmpi-4.1.0
  patch -p1 < ../ad16407c6a79c457ec3bc4251234af4281faa893.patch
  patch -p1 < ../3b624dbc1b4d6efb58b9ce7e54dc4acf19043e88.patch
  ./autogen.pl --force
  ./configure --prefix="${install_dir}" --with-hwloc="${install_dir}"
  make -j8 install )
```

I have also checked `PnetCDF` and it worked:
```bash
# ...
# Same as above
# ...
# Install PnetCDF:
( wget https://parallel-netcdf.github.io/Release/pnetcdf-1.12.2.tar.gz
  tar xvf pnetcdf-1.12.2.tar.gz && cd pnetcdf-1.12.2
  ./configure --prefix="${install_dir}" --with-mpi="${install_dir}" --enable-fortran --enable-shared MPICC=mpicc MPICXX=mpicxx MPIF90=mpif90 MPIF77=mpif77
  make -j8
  make -j8 check
  make -j8 ptest
  make -j8 install )
```

Also:
```console
$ gcc --version
gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ nagfor -V
NAG Fortran Compiler Release 6.2(Chiyoda) Build 6223
Product NPL6A62NA for x86-64 Linux
Copyright 1990-2018 The Numerical Algorithms Group Ltd., Oxford, U.K.
```

ThemosTsikas:
Can confirm the above script works with later versions on x86-64

```
$ gcc --version
gcc (GCC) 8.3.1 20191121 (Red Hat 8.3.1-5)
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ nagfor -V
NAG Fortran Compiler Release 7.0(Yurakucho) Build 7043
Product NPL6A70NA for x86-64 Linux
Copyright 1990-2020 The Numerical Algorithms Group Ltd., Oxford, U.K.

```

awlauria:
Based on the above comments, I think this is good to merge + backport to v5.0.x?

awlauria:
@jsquyres @gpaulsen @janjust

janjust:
@awlauria looks good to me

PR #6380, opened 2019-02-13
v4.0.x: oob/tcp: add cross version compatibility support

hppritcha:
Since we intend to provide cross version compatibility
between versions with the same major and minor, use
MAJOR.MINOR.0 instead of orte_version_string
(e.g. MAJOR.MINOR.RELEASEGREEK).

Open MPI 4.0.0 has already been released, so in order to make
it compatible with future 4.0.x releases, we have to use 4.0.0
as the version string, that is why we use MAJOR.MINOR.0 instead
of MAJOR.MINOR

Relates to #6157 

### DIFF
--- a/orte/mca/oob/tcp/oob_tcp_connection.c
+++ b/orte/mca/oob/tcp/oob_tcp_connection.c
@@ -14,8 +14,8 @@
  * Copyright (c) 2009-2018 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


hppritcha:
Do we want to make this more general and just check Major, so 4.1 and 4.0 can hopefully talk to each other @ggouaillardet 's "half way" approach for using containers with Open MPI?

rhc54:
IIRC, the conclusion of the telecon was to indeed only check the major number, and we will separate that from the ORTE_MAJOR_VERSION - i.e., change the code so we only pass/check "4". We will update that number if/when there is a change that impacts the oob wire protocol.

jsquyres:
This PR -- and the master PR #6157 -- does not reflect that decision.

ibm-ompi:

PR #6382, opened 2019-02-13
mtl/ofi: add a .gitignore

jsquyres:
Ignore generated files.

### DIFF
--- a/ompi/mca/mtl/ofi/.gitignore
+++ b/ompi/mca/mtl/ofi/.gitignore
@@ -1,0 +1,5 @@
+mtl_ofi_improbe_opt.c
+mtl_ofi_iprobe_opt.c
+mtl_ofi_irecv_opt.c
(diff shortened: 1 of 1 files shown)


jsquyres:
Open question to Open MPI dev community: does anyone know/remember why we have a giant, monolithic `.gitignore` at the top of the repository (vs. having smaller, per-directory `.gitignore` files)?

This PR adds a small `.gitignore` into `ompi/mca/mtl/ofi` that ignores a small number of generated files in that directory.

Do we have any objection to splitting the large, top-level `.gitignore` into the directories where the files are to be ignored?

jsquyres:
Fixed.  Not sure how I managed to add both the top-level `.gitignore` and the MTL-level `.gitignore` into the same commit...

jsquyres:
@bwbarrett Ping.

PR #6384, opened 2019-02-14
PML/UCX: Use net worker address for remote peers

brminich:
For remote node peers pack smaller worker address, which contains
network device addresses only. This would reduce amount of OOB traffic
during startup.


brminich:
@artpol84, @karasevb, @yosefe

jsquyres:
ok to test

yosefe:
@brminich pls pick to v4.0.x

brminich:
@yosefe - #6395

Issue #6383, opened 2019-02-14
build fails with both --enable-mpi-java --with-devel-headers


g-raffy:
Hi there,

The latest version of openmpi 2.x (2.1.6) fails to build when options `--enable-java-mpi` and `--with-devel-headers` are both active.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

openmpi v2.1.6 (latest 2.x)

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
from the source/distribution tarball [https://download.open-mpi.org/release/open-mpi/v2.1/openmpi-2.1.6.tar.gz](https://download.open-mpi.org/release/open-mpi/v2.1/openmpi-2.1.6.tar.gz)
### Please describe the system on which you are running

* Operating system/version: debian 9 (stretch)
* Computer hardware: apple xserve 2,1
* Network type: 1Gb ethernet

-----------------------------

### symptoms

Here are the last lines displayed by the failed build :
```
make[3]: Entering directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi/mpi/java/java'
  JAVAC    MPI.class
JAVADOC    doc
make  all-am
make[4]: Entering directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi/mpi/java/java'
    JAR    mpi.jar
make[4]: *** No rule to make target 'mpi_CartParms.h', needed by 'all-am'.  Stop.
make[4]: Leaving directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi/mpi/java/java'
Makefile:1751: recipe for target 'all' failed
make[3]: *** [all] Error 2
make[3]: Leaving directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi/mpi/java/java'
Makefile:1659: recipe for target 'all-recursive' failed
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi/mpi/java'
Makefile:3348: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/root/bug2587/builds/without-debian/build1/openmpi-2.1.6/ompi'
Makefile:1808: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
[...]

PR #6386, opened 2019-02-15
v3.1.x: java: Fix Java devel header build error

kawashima-fj:
Fix a compilation error under the following conditions. This bug was a regression introduced in v3.1.0.

- `--enable-mpi-java` is enabled (default: disabled)
- `--with-devel-headers` is enabled (default: disabled)
- OpenJDK 8 or later is used

(cherry picked from commit 132ea1a6b06fa233c09a672f6572bcef98a54be1)

@ggouaillardet Please review. This is your commit.

Ref. #6383


jsquyres:
This was accepted into v3.0.x already.

PR #6399, opened 2019-02-19
rml/ofi: remove

hppritcha:
per discussion at the 2/19/19 devel-core meeting,
remove rml/ofi from 4.0.x


hppritcha:
the rml/Makefile.in is generated as part of autogen.pl  if you do a git grep rather than grep, you will not see it.

PR #6416, opened 2019-02-21
request/greq: fix fortran generalized request

AboorvaDevarajan:
Copy the return error from the fortran query function into the C status

Signed-off-by: Aboorva Devarajan <aburvadevarajan@gmail.com>

### DIFF
--- a/ompi/request/grequest.c
+++ b/ompi/request/grequest.c
@@ -204,21 +204,25 @@ int ompi_grequest_invoke_query(ompi_request_t *request,
 
     /* MPI-2:8.2 does not say what to do with the return value from
        the query function (i.e., the int return value from the C
-       function or the ierr argument from the Fortran function).
-       Making the command decision here to ignore it.  If the handler
-       wants to pass an error back, it should set it in the MPI_ERROR
-       field in the status (which is always kept, regardless if the
-       top-level function was invoked with MPI_STATUS[ES]_IGNORE or
-       not).  */
+       function or the ierr argument from the Fortran function).*/
     if (NULL != g->greq_query.c_query) {
         if (g->greq_funcs_are_c) {
+            /* In the case of C, Making the command decision here to ignore the return value.  
+               If the handler wants to pass an error back, it should set it in 
+               the MPI_ERROR field in the status (which is always kept, regardless 
+               if the top-level function was invoked with MPI_STATUS[ES]_IGNORE or
+               not).  */
             rc = g->greq_query.c_query(g->greq_state, status);
         } else {
+            /* In the case of Fortran, status doesn't have a MPI_ERROR field so the 
+               return error from the Fortran query function is copied into the C status
+               to avoid passing garbage value */
             MPI_Fint ierr;
             MPI_Fint fstatus[sizeof(MPI_Status) / sizeof(int)];
             g->greq_query.f_query((MPI_Aint*)g->greq_state, fstatus, &ierr);
             MPI_Status_f2c(fstatus, status);
             rc = OMPI_FINT_2_INT(ierr);
+            status->MPI_ERROR = rc;
         }
     }
 


AboorvaDevarajan:
This is related to the issue #4597 and PR #6416 
The old PR has become stale since I deleted the fork, so opening a new one.
I have added the comments as mentioned in the PR #6416, if any changes are required in man pages please let me know. Thanks.

jsquyres:
I should note that MPI-3.1 has exactly the same text, but it's on page 479 line 10 (not 477, as it is in MPI-3.0).

jjhursey:
ok to test

bosilca:
The entire 12.2 section is a mess with little coherence. However, that section describe how OMPI should report errors to the user, and not how we should deal with error internally. To clarify why this matters: internally we always use the status associated with the request and when necessary we copy back into the user provided status. Thus, internally we should always copy the callback error into the MPI_ERROR field of the status, and let the upper level manage the MPI expected behavior.

Here is what I propose instead: the callback can do whatever it wants/needs with the status (which btw should be a valid object as indicated MPI 3.1 page 477 line 26), but if an error is returned by the callback function we update the MPI_ERROR field in the status.
```diff
diff --git a/ompi/request/grequest.c b/ompi/request/grequest.c
index 10d8885d82..c895b4232b 100644
--- a/ompi/request/grequest.c
+++ b/ompi/request/grequest.c
@@ -202,14 +202,13 @@ int ompi_grequest_invoke_query(ompi_request_t *request,
     int rc = OMPI_SUCCESS;
     ompi_grequest_t *g = (ompi_grequest_t*) request;

-    /* MPI-2:8.2 does not say what to do with the return value from
-       the query function (i.e., the int return value from the C
-       function or the ierr argument from the Fortran function).
-       Making the command decision here to ignore it.  If the handler
-       wants to pass an error back, it should set it in the MPI_ERROR
-       field in the status (which is always kept, regardless if the
-       top-level function was invoked with MPI_STATUS[ES]_IGNORE or
-       not).  */
+    /* MPI-3 mandates that the return value from the query function
+     * (i.e., the int return value from the C function or the ierr
+     * argument from the Fortran function) must be returned to the
+     * user. Thus, if the return of the query function is not MPI_SUCCESS
+     * we will update the MPI_ERROR field. Otherwise, the MPI_ERROR
+     * field is untouched (or left to the discretion of the query function).
+     */
     if (NULL != g->greq_query.c_query) {
         if (g->greq_funcs_are_c) {
             rc = g->greq_query.c_query(g->greq_state, status);
@@ -221,7 +220,9 @@ int ompi_grequest_invoke_query(ompi_request_t *request,
             rc = OMPI_FINT_2_INT(ierr);
         }
     }
-
+    if( MPI_SUCCESS != rc ) {
+        status->MPI_ERROR = rc;
+    }
     return rc;
 }
```

gpaulsen:
@AboorvaDevarajan What do you think about @bosilca's proposal?

AboorvaDevarajan:
I agree with it, but seems like for the ```MPI_SUCCESS``` case for FORTRAN we are still propagating the ```status.MPI_ERROR``` which is not initialized. (will be a garbage value) if it is not set in the query_fn.

Without that the below test case will still fail if we don't mandate the users to set the ```MPI_ERROR``` to ```MPI_SUCCESS``` in the ```query_fn```,

https://gist.github.com/AboorvaDevarajan/38340bc0fee37f50fb7583683ca243c9

But as Jeff mentioned MPI Standard doesn't allow the users to set the ```status.MPI_ERROR``` in the ```query_fn```.

AboorvaDevarajan:
I just checked how MPICH is handling this, seems like after the ```query_fn``` is invoked the function errors out if ```ierr != MPI_SUCCESS```, later copies all the attributes of the internal status back to the application status object except for the ```MPI_ERROR``` and uses it for further checking.

https://github.com/pmodels/mpich/blob/master/src/mpi/request/mpir_request.c#L339
https://github.com/pmodels/mpich/blob/master/src/mpi/request/mpir_request.c#L127

jsquyres:
@AboorvaDevarajan I think that the gist test case you cite should get the error code back in `ierr`, not the `MPI_ERROR` field in the status, because MPI_TEST/MPI_WAIT are used -- not their array values.  If MPICH is looking at the `MPI_ERROR` field to determine if the `query_fn` succeeded or failed, that seems erroneous (per the MPI-3.1 text).  Right?

If so, then it should probably be reported to the MPICH people.

AboorvaDevarajan:
@jsquyres  the test doesn't check the MPI_ERROR, thought the issue with @bosilca's suggestion will be that for  ```MPI_SUCCESS``` case ompi/request_wait will error out [here](https://github.com/open-mpi/ompi/blob/master/ompi/request/req_wait.c#L73) since ```req->req_status.MPI_ERROR``` will be  a garbage value,
https://github.com/open-mpi/ompi/blob/master/ompi/request/req_wait.c#L73

gpaulsen:
@jsquyres @bosilca, What do we want to do about this?

bosilca:
I think we're done here, this patch is already in master. For the sake of future reading, the comment indicates what the standard mandates and how we handle it.

jsquyres:
Yes, I think this PR can be closed as a dup (it may not have started as a dup, but it ended up as a dup).

markalle:
Can I reopen this just for further discussion purposes?  Not saying I want this PR to go in.

My understanding is the only time we care about `status[i].MPI_ERROR` is for the functions that are returning MPI_ERR_IN_STATUS, which are MPI_Wait/test_all/some() and those functions always call both query_fn() and also free_fn(), and it's the free_fn()'s return value that the standard says goes into the `status[i].MPI_ERROR`.

I think that's all consistent with how the standard tells users don't set `.MPI_ERROR` in their query_fn() because that's managed by MPI based on the callback return values.

So I think the status change currently at the bottom of ompi_grequest_invoke_query needs to be at the bottom of ompi_grequest_destruct() where it would be based on the free_fn() return value, and I guess the status is available there via `greq->greq_base.req_status`.  And the rc from ompi_grequest_invoke_query() would just get used in request_get_status.c which is the only pace where query_fn() is called without it being paired with free_fn() right?

ibm-ompi:

 Gist: https://gist.github.com/2b88aaf80478bd7131d4ba2f2b751002

markalle:
Well I've had to change my mind here about the Fortran side too.

It's like Aboorva said, ompi_request_default_wait() is still using
a value that originated in an uninialized stack variable over in
ompi_grequest_invoke_query()'s fstatus.  And I can make a simple
test fail if I deliberately put garbage into the starting fstatus
value.

I think I agree with the design's intent, where
req->req_status.MPI_ERROR is initialized to success and it's
meant to be unmodified in the case of callback success,
and set to the callback's return value if the callback fails
(and then higher levels decide whether req->req_status
goes back into the user's status).  But the MPI_Status_f2c
on fstatus is violating that design.

I think I should re-close this and make a new PR for the
fortran part.

PR #6417, opened 2019-02-21
Cart/Graph create would not run the next_cid  algorithm

abouteiller:
Cart/Graph create would not run the next_cid algorithm and create disjoint communicator with inconsistent cid.

The problem arises from calling `comm_nextcid` before the `new_comm->c_local_grp->my_rank` is set to something else than `MPI_UNDEFINED`. This causes the nextcid algorithm to not "participate" and creates a `newcomm` with an inconsistent cid.


bosilca:
This highlight a critical bug in a part of the code that somehow was never correctly tested. Therefore, we need to have a test that create unbalanced communicators to test such cases.

PR #6428, opened 2019-02-23
Pr/v3.0.x OMPIO Crash on Large Data

ax3l:
Same as #6344, but applied to v3.0.x.

Contains #6286 #6287.

cc @edgargabriel @jsquyres

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_file_read.c
+++ b/ompi/mca/common/ompio/common_ompio_file_read.c
@@ -9,7 +9,9 @@
  *                          University of Stuttgart.  All rights reserved.
  *  Copyright (c) 2004-2005 The Regents of the University of California.
  *                          All rights reserved.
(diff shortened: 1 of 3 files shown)


jsquyres:
ok to test

edgargabriel:
I would like to emphasize that having 

https://github.com/open-mpi/ompi/pull/6326 

pr'ed to 3.0 and 3.1 is as important ( maybe even more important ) as this issue. That is however a datatype issue, and I am not sure whether it applies cleanly from master to the 3.x branches, or whether it requires some additional work.

jsquyres:
@edgargabriel Thanks for bringing that up.  I tried to cherry pick 5a82c4fd0769ed007de774ef9115bde73b75de32 from #6326, but ran into conflicts on v3.0.x and v3.1.x.  I'll file a separate issue and ping George.

PR #6432, opened 2019-02-25
hwloc/external: configure: delay feature tests until after version tests

### LINKED ISSUE #6430: 4.0.0 configure fails when --enable-visibility is used and a pre-v2 hwloc is found
Hi,

4.0.0 fails to configure when `--enable-visibility` is passed as a configure option if a pre-v2 hwloc library is found:
```
checking if gcc -std=gnu99 -std=gnu99 supports -fvisibility=hidden... no
configure: error: Symbol visibility support requested but compiler does not seem to support it.  Aborting
```
This is because of a two different values for macro `HAVE_DECL_HWLOC_OBJ_OSDEV_COPROC` ending up in `confdefs.h`:
```
configure:385403: checking if gcc -std=gnu99 -std=gnu99 supports -fvisibility=hidden
configure:385419: gcc -std=gnu99 -std=gnu99 -o conftest -DNDEBUG -g -march=native -fvisibility=hidden -Werror -I/short/z00/bjm900/tmp/tmp.pwJnqtgC0F/openmpi-4.0.0/gcc-opt/opal/mca/event/libevent2022/libevent/include -I/short/z00/bjm900/tmp/tmp.pwJnqtgC0F/openmpi
-4.0.0/openmpi-4.0.0/opal/mca/event/libevent2022/libevent -I/short/z00/bjm900/tmp/tmp.pwJnqtgC0F/openmpi-4.0.0/openmpi-4.0.0/opal/mca/event/libevent2022/libevent/include -I/short/z00/bjm900/tmp/tmp.pwJnqtgC0F/openmpi-4.0.0/gcc-opt/opal/mca/hwloc/hwloc201/hwloc/i
nclude -I/short/z00/bjm900/tmp/tmp.pwJnqtgC0F/openmpi-4.0.0/openmpi-4.0.0/opal/mca/hwloc/hwloc201/hwloc/include     conftest.c -lrt -lm -lutil  -lz  >&5
conftest.c:654:1: error: "HAVE_DECL_HWLOC_OBJ_OSDEV_COPROC" redefined
conftest.c:553:1: error: this is the location of the previous definition
configure:385419: $? = 1
configure: failed program was:
...
| #define HAVE_DECL_HWLOC_OBJ_OSDEV_COPROC 0
...
| #define HAVE_DECL_HWLOC_OBJ_OSDEV_COPROC 1
```
Previous tests generate a warning for the duplication, but the visibility test includes `-Werror` which makes it error instead.

The first copy of the definition is coming from the test of the external version of hwloc (in our case v1.5, from the CentOS 6 repos), while the second is hard-enabled by the internal `hwloc201` component:
```
openmpi-4.0.0/opal/mca/hwloc/hwloc201/configure.m4:          [AC_DEFINE([HAVE_DECL_HWLOC_OBJ_OSDEV_COPROC], [1])
openmpi-4.0.0/opal/mca/hwloc/external/configure.m4:                  AC_CHECK_DECLS([HWLOC_OBJ_OSDEV_COPROC], [], [], [#include <hwloc.h>])
```

Should the macros set by the tests of the external library should be removed when the internal library is used?

Thanks,
Ben

benmenadue:
Hold off running hwloc/external feature tests until after we decide if we're using the internal or external component.

Refs #6430.

### DIFF
--- a/opal/mca/hwloc/external/configure.m4
+++ b/opal/mca/hwloc/external/configure.m4
@@ -133,8 +133,6 @@ AC_DEFUN([MCA_opal_hwloc_external_CONFIG],[
                  [CPPFLAGS="$CPPFLAGS $opal_hwloc_external_CPPFLAGS"
                   LDFLAGS="$LDFLAGS $opal_hwloc_external_LDFLAGS"
                   LIBS="$LIBS $opal_hwloc_external_LIBS"
-                  AC_CHECK_DECLS([HWLOC_OBJ_OSDEV_COPROC], [], [], [#include <hwloc.h>])
-                  AC_CHECK_FUNCS([hwloc_topology_dup])
 
                   AC_MSG_CHECKING([if external hwloc version is 1.5 or greater])
                   AC_COMPILE_IFELSE(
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

PR #6439, opened 2019-02-27
Fix cross-mpirun connect/accept operations

rhc54:
Ensure we publish all the info required to be returned to the other
mpirun when executing this operation. We need to know the daemon (and
its URI) that is hosting each of the other procs so we can do a direct
modex operation and retrieve their connection info.


rhc54:
Fixes #3458

Issue #6433, opened 2019-02-25
mpi_yield_when_idle setting from etc/openmpi-mca-params.conf is ignored in 4.0.0


iassiour:
Hi ,

I have set   "yield when idle" in etc/openmpi-mca-params.conf

tail openmpi-mca-params.conf

plm_rsh_agent = rsh
rmaps_base_oversubscribe = 1
hwloc_base_binding_policy = none
mpi_yield_when_idle = 1

However the setting seems to be ignored. 

I get 100 % us 0% sys for processes waiting on MPI calls.

But if I set the environment variable:

setenv OMPI_MCA_mpi_yield_when_idle 1

or 
mpirun --mca mpi_yield_when_idle true

I get the more expected 15% us 85 % sys.

Could someone please check that the setting is read from the configuration file as expected.

jsquyres:
You may not have gotten an email about it, but @ggouaillardet proposed a PR about this -- see #6440.

iassiour:
Greetings jsquyres ,

Thanks, Could you please let me know what future minor version of ompi will have this fix.

gpaulsen:
@hppritcha should just cherry-pick the original https://github.com/open-mpi/ompi/pull/6440 fix to v4.0.x for inclusion into v4.0.3 ?

hppritcha:
closed via #7168 and #6440

planetA:
@iassiour @ggouaillardet 

Sorry for reviving an old issue, but as far as I know yield does nothing on modern Linux kernel. Or rather, immediately returns to the process. At least with modern CFS.

Considering that, why OpenMPI would need to use yield?

The man page for sched_yield says:

> Use  of sched_yield() with nondeterministic scheduling policies such as SCHED_OTHER is unspecified and very likely means your application design is broken.

devreal:
@planetA 

> Sorry for reviving an old issue, but as far as I know yield does nothing on modern Linux kernel. Or rather, immediately returns to the process. At least with modern CFS.

Can you point out documentation for this behavior? There is plenty of noise out there, the only useful thing I found was https://www.kernel.org/doc/html/latest/scheduler/sched-design-CFS.html which states:

> yield_task(…)
> 
> This function is basically just a dequeue followed by an enqueue [...]

I would expect that any runnable task with a similar priority gets scheduled for execution before the task calling `sched_yield` gets a chance to run again. I'm not deep into CFS specifics though...

> Considering that, why OpenMPI would need to use yield?

[...]

PR #6440, opened 2019-02-27
schizo/ompi: correctly handle the yield_when_idle option

ggouaillardet:
only sets the OMPI_MCA_mpi_yield_when_idle environment variable
if this directive was not given (for example via a config file).

Refs. open-mpi/ompi#6433

### DIFF
--- a/orte/mca/schizo/ompi/schizo_ompi.c
+++ b/orte/mca/schizo/ompi/schizo_ompi.c
@@ -15,8 +15,8 @@
  * Copyright (c) 2011-2017 Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2017      UT-Battelle, LLC. All rights reserved.
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
- * Copyright (c) 2015      Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2015-2019 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * Copyright (c) 2018      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
@@ -908,9 +908,9 @@ static int setup_fork(orte_job_t *jdata,
 
     /* setup yield schedule - do not override any user-supplied directive! */
     if (oversubscribed) {
-        opal_setenv("OMPI_MCA_mpi_yield_when_idle", "1", false, &app->env);
+        opal_setenv("OMPI_MCA_mpi_oversubscribe", "1", true, &app->env);
     } else {
-        opal_setenv("OMPI_MCA_mpi_yield_when_idle", "0", false, &app->env);
+        opal_setenv("OMPI_MCA_mpi_oversubscribe", "0", true, &app->env);
     }
 
     /* set the app_context number into the environment */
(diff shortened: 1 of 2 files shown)


ggouaillardet:

ggouaillardet:
@rhc54 you are right, I will update the PR tomorrow

orte/mca/schizo/ompi/schizo_ompi.c
```diff
@@ -906,11 +908,34 @@ static int setup_fork(orte_job_t *jdata,
...
+    ompi_mpi_yield_when_idle = false;
+    (void) mca_base_var_register("ompi", "mpi", NULL, "yield_when_idle",
+                                 "Yield the processor when waiting for MPI communication (for MPI processes, will default to 1 when oversubscribing nodes)",
+                                 MCA_BASE_VAR_TYPE_BOOL, NULL, 0, MCA_BASE_VAR_FLAG_DEF_UNSET,
+                                 OPAL_INFO_LVL_9,
+                                 MCA_BASE_VAR_SCOPE_READONLY,
+                                 &ompi_mpi_yield_when_idle);
+
+    bool set_yield_when_idle = true;
+    int var_index = mca_base_var_find("ompi", "mpi", NULL, "yield_when_idle");
```
jsquyres: In addition to what @rhc54 said:

1. My $0.02: make `ompi_mpi_yield_when_idle` level 5 or 6, not 9 (per https://github.com/open-mpi/ompi/wiki/MCAParamLevels).
1. IIRC, `mca_base_var_register()` returns the newly-created index -- I don't think you need to explicitly `(void)`-ignore its return and then `mca_base_var_find()`.

ggouaillardet:
I updated the PR based on the received feedback.

orte/mca/schizo/ompi/schizo_ompi.c
```diff
@@ -908,9 +908,9 @@ static int setup_fork(orte_job_t *jdata,
 
     /* setup yield schedule - do not override any user-supplied directive! */
     if (oversubscribed) {
-        opal_setenv("OMPI_MCA_mpi_yield_when_idle", "1", false, &app->env);
+        opal_setenv("OMPI_MCA_mpi_oversubscribe", "1", true, &app->env);
```
jsquyres: I know the previous code was directly setting an MPI env variable, but this feels wrong -- it's an abstraction violation for ORTE to set anything in the OMPI layer.

ggouaillardet: I understand how that can feel wrong. That being said, doesn't an `ompi` component in `ORTE` feel wrong in the same way ? Since `orte` is scheduled to be replaced with `prrte`,  I did not want to use a MCA parameter related to `orte`. Would setting `OMPI_MCA_opal_oversubscribe` instead feel more right ?

rhc54: Let's not get too hung up on component names, guys - I could have called it anything. We even set "MPI_foo" envars in this component and freely note that it is an abstraction violation because the community long-ago requested that we do so.

Perhaps instead of defining an MCA parameter we could just pass something via PMIx in the job info to indicate oversubscription. Have to define a new attribute for it, but that's trivial to do, and it would translate easier to PRRTE. You'd need to see where you would want to access it and deal with setting the yield variable without overriding any user setting, but that should be doable.

jsquyres: Actually, I should have looked closer and realized that this is a component exclusively for `ompi`.  I think that changes the equation here.  I retract my original comment.

PR #6446, opened 2019-03-01
v4.0.x: Fix cross-mpirun connect/accept operations

rhc54:
Ensure we publish all the info required to be returned to the other
mpirun when executing this operation. We need to know the daemon (and
its URI) that is hosting each of the other procs so we can do a direct
modex operation and retrieve their connection info.

(cherry picked from commit 60961ceb41cde6fbdc84694748696e90ddcfb345)


rhc54:
Fixes #3458

hppritcha:
@gpaulsen we should take this as part of a rc2

hppritcha:
@ggouaillardet please review when you have a chance

PR #6448, opened 2019-03-04
in-place conversion macro writes into INPUT argument

markalle:
In fint_2_int.h there are some conversion macros for logicals. It has
one path for OMPI_SIZEOF_FORTRAN_LOGICAL != SIZEOF_INT where a new array
would be allocated and the conversions then might expand to
```    c_array[i] = (array[i] == 0 ? 0 : 1)```
and another path for OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT where it
does things "in place", so the same conversion there would just be
```    array[i] = (array[i] == 0 ? 0 : 1)```

The problem is some of the logical arrays being converted are INPUT
arguments. And it's possible for some compilers to even put the argument
in read-only memory so the above "in place" conversion SEGV's.  A
testcase I have used
```    call MPI_CART_SUB(oldcomm, (/.true.,.false./), newcomm, ierr)```
and gfortran put the second arg in read-only mem.

In cart_sub_f.c you can trace the ompi_fortran_logical_t *remain_dims arg.
remain_dims[] is for input only, but the file uses
```
    OMPI_LOGICAL_ARRAY_NAME_DECL(remain_dims);
    OMPI_ARRAY_LOGICAL_2_INT(remain_dims, ndims);
    PMPI_Cart_sub(..., OMPI_LOGICAL_ARRAY_NAME_CONVERT(remain_dims), ...);
    OMPI_ARRAY_INT_2_LOGICAL(remain_dims, ndims);
```
to convert it to c-ints make a C call then restore it to Fortran logicals
before returning.

It's not always wrong to convert purely in-place, eg cart_get_f.c has
a periods[] that's exclusively for OUTPUT and it would be fine with the
macros as they were. But I still say the macros are invalid because they
don't distinguish whether they're being used on INPUT or OUTPUT args and
thus they can't be used in a way that's legal for both cases.

It might be possible to fix the macros by adding more of them so that
cart_create_f.c and cart_get_f.c would use different macros that give
more context. But my fix here is just to turn off the first block and
make all paths run as if OMPI_SIZEOF_FORTRAN_LOGICAL != SIZEOF_INT.

The main macros that get enlarged by this change are
```
    define OMPI_ARRAY_LOGICAL_2_INT_ALLOC : mallocs now
    define OMPI_ARRAY_LOGICAL_2_INT : also mallocs now
```
But these are only used in 4 places, three of which are the purpose of
this checkin, to avoid the former in-place expansion of an INPUT arg:
```
    cart_create_f.c
    cart_map_f.c
    cart_sub_f.c
```
and one of which is an OUPUT arg that was fine and that gets
unnecessarily expanded into a separate array by this checkin.
```
    cart_get_f.c
```

So I think an unnecessary malloc in cart_get_f.c is the only downside
to this change, where the logicals array argument could have been used
and converted in place.

### DIFF
--- a/ompi/mpi/fortran/base/fint_2_int.h
+++ b/ompi/mpi/fortran/base/fint_2_int.h
@@ -11,8 +11,8 @@
  *                         All rights reserved.
  * Copyright (c) 2011-2012 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2014-2019 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -161,9 +161,24 @@
 
 /*
  * Define MACROS to take account of different size of logical from int
+ *
+ * There used to be an in-place option for the below conversions of
+ * logical arrays.  So if mpi_cart_create(..., periods, ...) took an
+ * input array of Fortran logicals, it would walk the array converting
+ * the elements to C-logical values, then at the end it would restore
+ * the values back to Fortran logicals.
+ *
+ * The problem with that is periods is an INPUT argument and some
+ * Fortran compilers even put it in read-only memory because of that.
+ * So writing to it wasn't generally okay, even though we were restoring it
+ * before returning.
+ *
+ * The in-place option is hence only valid if no conversion is ever needed
+ * (e.g. Fortran logical and C int have the same size *and** Fortran logical
+ * .TRUE. value is 1 in C.
  */
 
-#if OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT
+#if (OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT) && (OMPI_FORTRAN_VALUE_TRUE == 1)
 #  define OMPI_LOGICAL_NAME_DECL(in)               /* Not needed for int==logical */
 #  define OMPI_LOGICAL_NAME_CONVERT(in)        in  /* Not needed for int==logical */
 #  define OMPI_LOGICAL_SINGLE_NAME_CONVERT(in) in /* Not needed for int==logical */
@@ -172,37 +187,15 @@
 #  define OMPI_ARRAY_LOGICAL_2_INT_ALLOC(in,n)     /* Not needed for int==logical */
 #  define OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in)     /* Not needed for int==logical */
 
-#  if OMPI_FORTRAN_VALUE_TRUE == 1
-#    define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    0
-#    define OMPI_LOGICAL_2_INT(a) a
-#    define OMPI_INT_2_LOGICAL(a) a
-#    define OMPI_ARRAY_LOGICAL_2_INT(in, n)
-#    define OMPI_ARRAY_INT_2_LOGICAL(in, n)
-#    define OMPI_SINGLE_INT_2_LOGICAL(a)            /* Single-OUT variable -- Not needed for int==logical, true=1 */
-#  else
-#    define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    1
-#    define OMPI_LOGICAL_2_INT(a) ((a)==0? 0 : 1)
-#    define OMPI_INT_2_LOGICAL(a) ((a)==0? 0 : OMPI_FORTRAN_VALUE_TRUE)
-#    define OMPI_SINGLE_INT_2_LOGICAL(a) *a=OMPI_INT_2_LOGICAL(OMPI_LOGICAL_NAME_CONVERT(*a))
-#    define OMPI_ARRAY_LOGICAL_2_INT(in, n) do { \
-       int converted_n = (int)(n); \
-       OMPI_ARRAY_LOGICAL_2_INT_ALLOC(in, converted_n + 1); \
-       while (--converted_n >= 0) { \
-         OMPI_LOGICAL_ARRAY_NAME_CONVERT(in)[converted_n]=OMPI_LOGICAL_2_INT(in[converted_n]); \
-       } \
-     } while (0)
-#    define OMPI_ARRAY_INT_2_LOGICAL(in, n) do { \
-       int converted_n = (int)(n); \
-       while (--converted_n >= 0) { \
-         in[converted_n]=OMPI_INT_2_LOGICAL(OMPI_LOGICAL_ARRAY_NAME_CONVERT(in)[converted_n]); \
-       } \
-       OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in); \
-     }  while (0)
-
-#  endif
+#  define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    0
+#  define OMPI_LOGICAL_2_INT(a) a
+#  define OMPI_INT_2_LOGICAL(a) a
+#  define OMPI_ARRAY_LOGICAL_2_INT(in, n)
+#  define OMPI_ARRAY_INT_2_LOGICAL(in, n)
+#  define OMPI_SINGLE_INT_2_LOGICAL(a)            /* Single-OUT variable -- Not needed for int==logical, true=1 */
 #else
 /*
- * For anything other than Fortran-logical == C-int, we have to convert
+ * For anything other than Fortran-logical == C-int or some .TRUE. is not 1 in C, we have to convert
  */
 #  define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    1
 #  define OMPI_LOGICAL_NAME_DECL(in)           int c_##in
@@ -238,7 +231,7 @@
        } \
        OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in); \
      }  while (0)
-#endif /* OMPI_SIZEOF_FORTRAN_LOGICAL */
+#endif /* OMPI_SIZEOF_FORTRAN_LOGICAL && OMPI_FORTRAN_VALUE_TRUE */
 
 
 #endif /* OMPI_FORTRAN_BASE_FINT_2_INT_H */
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ompi/mpi/fortran/mpif-h/cart_map_f.c [hunk 1]
```diff
@@ -89,6 +90,10 @@ void ompi_cart_map_f(MPI_Fint *comm, MPI_Fint *ndims, MPI_Fint *dims,
     if (NULL != ierr) *ierr = OMPI_INT_2_FINT(c_ierr);
 
     OMPI_ARRAY_FINT_2_INT_CLEANUP(dims);
+#if OMPI_ALLOW_INPLACE_PATH_FOR_LOGICALS
```
ggouaillardet: The related `C` binding is `MPI_Cart_create(..., const int periods[], ...)` so it looks like the real issue is we cannot convert `periods` back into Fortran, and the right fix here is to replace `OMPI_ARRAY_INT_2_LOGICAL(periods, size)` with `OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(periods)`

The same comment applies to other parts of this PR.

make sense ?

jsquyres:
I think 2 out of the 4 cases are communicator constructors (which almost certainly involve communication), and the other 2 are not common-use functions (i.e., they'd be using during setup of the app).  So I think an extra malloc+free is ok / worth the cost, especially if it makes the code a little simpler.

I'd be in favor of just:

1. Putting in a comment "We used to try to do this stuff in place, but we really can't because ..." (and Gilles is right that the new C `const` markings make it even more problematic).
2. Stripping out the old code / leaving only the 1 code path that does the extra malloc+free

The old code will still be in git if we want to get it back.  That's what git history is for.  😄

More specifically: I'd rather *not* effectively have the following hard-coded in our code base:

```c
#if 0
    do_it_a_way_that_we_know_doesnt_work()
#else
    do_it_the_right_way();
#endif
```

markalle:
Thanks for the review, I agree and took out the #ifdef and just left the new code

ompi/mpi/fortran/base/fint_2_int.h
```diff
@@ -163,7 +164,29 @@
  * Define MACROS to take account of different size of logical from int
  */
 
-#if OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT
```
ggouaillardet: @markalle I believe you do not need this any more.
I added some tests in open-mpi/ompi-tests#104, built both Open MPI and the ibm test suite with `-fdefault-integer-8` and had no issue with the stock/current `ompi/mpi/fortran/base/fint_2_int.h`

markalle: Oh gosh, this is embarrassing. I agree. My local tree is changed like you described but apparently in the thing I pushed I failed to "git add" that file. Let me re-push

markalle:
Okay, re-pushed. In my previous push I failed to update fint_2_int.h in the same vein as the .c files. Eg to remove the in-place path entirely and just leave the allocate/copy/free path

ompi/mpi/fortran/base/fint_2_int.h
```diff
@@ -163,46 +164,19 @@
...
-#  define OMPI_LOGICAL_NAME_DECL(in)               /* Not needed for int==logical */
-#  define OMPI_LOGICAL_NAME_CONVERT(in)        in  /* Not needed for int==logical */
-#  define OMPI_LOGICAL_SINGLE_NAME_CONVERT(in) in /* Not needed for int==logical */
-#  define OMPI_LOGICAL_ARRAY_NAME_DECL(in)         /* Not needed for int==logical */
-#  define OMPI_LOGICAL_ARRAY_NAME_CONVERT(in)  in  /* Not needed for int==logical */
-#  define OMPI_ARRAY_LOGICAL_2_INT_ALLOC(in,n)     /* Not needed for int==logical */
-#  define OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in)     /* Not needed for int==logical */
+// There used to be an in-place option for the below conversions of
+// logical arrays.  So if mpi_cart_create(..., periods, ...) took an
+// input array of jortran logicals, it would walk the array converting
```
jsquyres: jortran -> Fortran

ggouaillardet:
@markalle I am sorry I am still missing why you need to change `fint_2_int.h`
if `OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT` then we can use the data in place. Using a copy is not wrong but it is clearly suboptimal. If this is an input parameter, then the Fortran binding should not write it, and you did fix that in the three Fortran bindings.
Can you produce a counter example ? (e.g. a simple reproducer that crashes if `fint_2_int.h` is __not__ modified)

ompi/mpi/fortran/base/fint_2_int.h
markalle: hahaha, will repush in a sec

markalle:
Sure:
```
call MPI_CART_SUB(oldcomm, (/.true.,.false./), newcomm, ierr)
```
The second arg is for input only. The cart_sub_f.c contains

ggouaillardet:
@markalle I finally understand what you are trying to solve in `ompi/mpi/fortran/base/fint_2_int.h` !

And I believe we want to keep the in-place option when possible.
I cannot push to your branch (you might have unchecked the box in the github ui), so here is an inline patch you can squash into your branch

```patch
From 66c77107200836100a354326fec84077af7e327e Mon Sep 17 00:00:00 2001
From: Gilles Gouaillardet <gilles@rist.or.jp>
Date: Fri, 29 Mar 2019 09:51:15 +0900
Subject: [PATCH] fortran/base: keep the in-place option if possible

---
 ompi/mpi/fortran/base/fint_2_int.h | 50 +++++++++++++++++++++---------
 1 file changed, 35 insertions(+), 15 deletions(-)

diff --git a/ompi/mpi/fortran/base/fint_2_int.h b/ompi/mpi/fortran/base/fint_2_int.h
index d707719c82..44ce128956 100644
--- a/ompi/mpi/fortran/base/fint_2_int.h
+++ b/ompi/mpi/fortran/base/fint_2_int.h
@@ -11,9 +11,8 @@
  * Copyright (c) 2011-2012 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
- * Copyright (c) 2014      Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2019 IBM Corporation. All rights reserved.
+ * Copyright (c) 2014-2019 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * $COPYRIGHT$
  *
@@ -162,21 +161,41 @@
 
 /*
  * Define MACROS to take account of different size of logical from int
+ *
+ * There used to be an in-place option for the below conversions of
+ * logical arrays.  So if mpi_cart_create(..., periods, ...) took an
+ * input array of Fortran logicals, it would walk the array converting
+ * the elements to C-logical values, then at the end it would restore
+ * the values back to Fortran logicals.
+ *
+ * The problem with that is periods is an INPUT argument and some
+ * Fortran compilers even put it in read-only memory because of that.
+ * So writing to it wasn't generally okay, even though we were restoring it
+ * before returning.
+ *
+ * The in-place option is hence only valid if no conversion is ever needed
+ * (e.g. Fortran logical and C int have the same size *and** Fortran logical
+ * .TRUE. value is 1 in C.
  */
 
-// There used to be an in-place option for the below conversions of
-// logical arrays.  So if mpi_cart_create(..., periods, ...) took an
-// input array of Fortran logicals, it would walk the array converting
-// the elements to C-logical values, then at the end it would restore
-// the values back to Fortran logicals.
-//
-// The problem with that is periods is an INPUT argument and some
-// Fortran compilers even put it in read-only memory because of that.
-// So writing to it wasn't okay, even though we were restoring it
-// before returning.
-
+#if (OMPI_SIZEOF_FORTRAN_LOGICAL == SIZEOF_INT) && (OMPI_FORTRAN_VALUE_TRUE == 1)
+#  define OMPI_LOGICAL_NAME_DECL(in)               /* Not needed for int==logical */
+#  define OMPI_LOGICAL_NAME_CONVERT(in)        in  /* Not needed for int==logical */
+#  define OMPI_LOGICAL_SINGLE_NAME_CONVERT(in) in /* Not needed for int==logical */
+#  define OMPI_LOGICAL_ARRAY_NAME_DECL(in)         /* Not needed for int==logical */
+#  define OMPI_LOGICAL_ARRAY_NAME_CONVERT(in)  in  /* Not needed for int==logical */
+#  define OMPI_ARRAY_LOGICAL_2_INT_ALLOC(in,n)     /* Not needed for int==logical */
+#  define OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in)     /* Not needed for int==logical */
+
+#  define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    0
+#  define OMPI_LOGICAL_2_INT(a) a
+#  define OMPI_INT_2_LOGICAL(a) a
+#  define OMPI_ARRAY_LOGICAL_2_INT(in, n)
+#  define OMPI_ARRAY_INT_2_LOGICAL(in, n)
+#  define OMPI_SINGLE_INT_2_LOGICAL(a)            /* Single-OUT variable -- Not needed for int==logical, true=1 */
+#else
 /*
- * Allocate and convert Fortran-logicals into C-int
+ * For anything other than Fortran-logical == C-int or some .TRUE. is not 1 in C, we have to convert
  */
 #  define OMPI_FORTRAN_MUST_CONVERT_LOGICAL_2_INT    1
 #  define OMPI_LOGICAL_NAME_DECL(in)           int c_##in
@@ -212,6 +231,7 @@
        } \
        OMPI_ARRAY_LOGICAL_2_INT_CLEANUP(in); \
      }  while (0)
+#endif /* OMPI_SIZEOF_FORTRAN_LOGICAL && OMPI_FORTRAN_VALUE_TRUE */
 
 
 #endif /* OMPI_FORTRAN_BASE_FINT_2_INT_H */
-- 
2.20.1 (Apple Git-117)

```

markalle:
I like that patch. I'll ask geoffrey later how to fix my github setup

markalle:
squashed and repushed

markalle:
bot:retest

ggouaillardet:
@markalle https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork explains how to allow maintainers to push into your PR's.
(fwiw, the box is checked on all my PRs by default)

I think it does not work because you did not (github) fork your `ompi` repository from `open-mpi/ompi` but `dsolt/ompi` instead.

markalle:
I agree about the 1sided not belonging there, will fix

gpaulsen:
@markalle Please PR this to v4.0.x branch.
Does it also belong in either v3.0.x or v3.1.x?

PR #6453, opened 2019-03-05
opal/asm: Fix a compiler warning on POWER arch

nysal:
OPAL_XLC_INLINE_ASSEMBLY was removed in commit ebce88b7ad.
Removing dead code, which also fixes a compiler warning.


rhc54:
@jjhursey Just mnaking a note that we'll need this backported to PMIx and PRRTE.

PR #6450, opened 2019-03-05
v4.0.x: btl/usnic: fix usnic_btl_run_tests CPPFLAGS

ggouaillardet:
do define the OMPI_LIBMPI_NAME macro via the CPPFLAGS.
The issue occurs when Open MPI is configured with
--enable-opal-btl-usnic-unit-tests

Thanks George Marselis for reporting this issue

Refs. open-mpi/ompi#6441

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>

(cherry picked from commit open-mpi/ompi@b4097626ab256d2ace27a99b7fd174cd2709533e)

PR #6459, opened 2019-03-05
btl/usnic: amend Makefile.am fix from b4097626ab

jsquyres:
Use $(AM_CPPFLAGS) in $(usnic_btl_run_tests_CPPFLAGS) so that we don't
have to replicate hard-coded values.

Refs #6441 #6450


ggouaillardet:
@jsquyres, though we do not require `$(opal_ofi_CPPFLAGS)` nor `-DBTL_IN_OPAL=1` (and the latter looks counter-intuitive) I have no objection with your changes.

jsquyres:
Yeah, I noticed the `-DBTL_IN_OPAL` while I was doing this and took a mental note to PR removing that.  It's left over from when we transitioned the BTL from the OMPI layer to the OPAL layer; I had some code in there that let me maintain the same BTL code in both OMPI and OPAL (for PR'ing back to release branches), which was quite convenient at the time.

PR #6183, opened 2018-12-12
oshmem: fix macro usage in pshmem.h

ggouaillardet:
pshmem.h now includes shmem.h (since open-mpi/ompi@f46130cd20914a3c60b19c3b2ed7c4b380319ee9) and some macros were removed at that time.

Use the OSHMEM_HAVE_C11 macro (defined in shmem.h) instead of the
previous OSHMEMP_HAVE_C11 macrso previously defined in pshmem.h

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>

PR #6132, opened 2018-11-28
OSHMEM: Let `pshmem.h` include `shmem.h` to be stand-alone again

bertwesarg:
See #6093

Signed-off-by: Bert Wesarg <bert.wesarg@tu-dresden.de>

PR #6465, opened 2019-03-05
V4.0.x self contained pshmem

bertwesarg:
Pick #6132 and #6183 for v4.0.1.

Original bug was introduced with 4.0.0, thus should be fixed in v4.0.1.


bertwesarg:
@ggouaillardet @hppritcha please consider for v4.0.1. Thanks.

Issue #6470, opened 2019-03-07
v4.0.x: ompi_info field for MPI-1 compat symbols in mpi.h?


jsquyres:
While reading https://github.com/open-mpi/ompi/pull/6469, I realized that there's no `ompi_info` field indicating whether `mpi.h` includes the MPI-1 compat symbols or not.

We should probably add this.  Should be a trivial thing to add.

It might be specific just to v4.0.x, but *maybe* we might want to add it to master (just hard-coded to "no") for continuity into v5.0.x...?

gpaulsen:
Good idea.  I'll give this a go today.

hppritcha:
@gpaulsen see #6471

hppritcha:
closing as both master and v4.0.x have been updated

PR #6471, opened 2019-03-07
ompi_info: report whether MPI1 compat is enabled

hppritcha:
Its so easy to misspell compatability (sic) that we need
to have ompi_info help us out.

Related to #6470


jsquyres:
I like.  👍 Let me add one more quick commit to this PR -- a line for the bottom of configure.

jsquyres:
@hppritcha This should replace #6471

hppritcha:
hmm...okay I'll open a PR on 4.0.x

gpaulsen:
If we're going to have an info to tell users if this functionality exists or doesn't exist, wouldn't we want this on master as well?
Otherwise master and v3.1 would look the same, even though they are very different.

jsquyres:
Fair enough.  Should this just be hard-wired to output "no" from `ompi_info` on master, then?  I.e., `OMPI_ENABLE_MPI1_COMPAT` doesn't exist here on master.

PR #6476, opened 2019-03-11
configury: Remove --enable-mpi1-compatibility

jsquyres:
This configure CLI option should probably have been removed as part of
a6d6be2853488cfb20128f97b381b3c94a921cd7.  Regardless, the #defines it
sets no longer exist in the code base at all, so they should be
removed.  Also remove the text about it in README.


jsquyres:
@hppritcha This should replace #6471

PR #6482, opened 2019-03-13
Fix integer overflows with indexed datatype creation.

awlauria:
The types of count, disp, and extent passed into
ompi_datatype_add() should be size_t, ptrdiff_t and ptrdiff_t,
respectively. This prevents integer overflows and errors in
computing the size of large indexed datatypes.

### DIFF
--- a/ompi/datatype/ompi_datatype_create_indexed.c
+++ b/ompi/datatype/ompi_datatype_create_indexed.c
@@ -15,6 +15,7 @@
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

jjhursey:
I'm not sure why the IBM CI didn't fire on this PR. Let's try to force it.
bot:ibm:retest

jsquyres:
I assume this is a somewhat esoteric bug fix (i.e., not likely to happen in many real world cases).  

* Should we cherry pick it to v4.0.x?
* Should we cherry pick it to v3.1.x / v3.0.x?

jjhursey:
It's a bit of an edge case, but a solid fix. We will want this in at least the v4.0.x series, and should be considered for the v3.x series too - IMHO.

awlauria:
For some more context on how it was found,  it can be reproduced by calling `MPI_Type_create_hindexed(2238, 477312832, 477321792 MPI_BYTE, &type)`.
followed by querying the size `MPI_Type_size_x(type, &size)` would show a negative value returned in `size`.

So it is probably not a common case.

jsquyres:
That works for me.  I 👍'ed @jjhursey's post; I wouldn't mind seeing this PR'ed to v3.0.x/v3.1.x/v4.0.x.

awlauria:
Sure.

4.0.x: https://github.com/open-mpi/ompi/pull/6485
3.1.x: https://github.com/open-mpi/ompi/pull/6483
3.0.x: https://github.com/open-mpi/ompi/pull/6484

PR #6477, opened 2019-03-11
opal_hwloc_base_cset2str() off-by-1 in its strncat()

markalle:
I think the strncat() calls here need to be of the form
```    strncat(str, new_str_to_add, len - strlen(new_str_to_addstr) - 1);```
since in the OMPI calls len is being used as total number of bytes
in str.  Eg if our string was char str[64] len would be 64.

strncat(dest,src,n) on the other hand is documented as writing up to
n chars from the incoming string plus 1 for the null, for n+1 total
bytes it can write.

Signed-off-by: Mark Allen <markalle@us.ibm.com>

PR #6488, opened 2019-03-14
v3.1.x: opal_hwloc_base_cset2str() off-by-1 in its strncat()

markalle:
I think the strncat() calls here need to be of the form
    strncat(str, new_str_to_add, len - strlen(new_str_to_addstr) - 1);
since in the OMPI calls len is being used as total number of bytes
in str.

strncat(dest,src,n) on the other hand is documented as writing up to
n chars from the incoming string plus 1 for the null, for n+1 total
bytes it can write.

(cherry picked from commit 30d60994d258f5f0b7c432efd284d1b6b8333faf)

Conflicts:
	opal/mca/hwloc/base/hwloc_base_util.c


markalle:
cherry pick from https://github.com/open-mpi/ompi/pull/6477

PR #6499, opened 2019-03-19
NEWS: add a few news items for 4.0.1rc2

hppritcha:
a little late, but a couple of bullets for the
4.0.1rc2 NEWS.

[skip ci]


hppritcha:
@gpaulsen we forgot to update the NEWS for the 4.0.1rc2.  I'd suggest we merge this in, then if we don't do any other fixes that lead to a rc3, tag an rc3 and build the tarball, then release that as 4.0.1.

PR #6503, opened 2019-03-19
Do not force 'hash' gds on direct modex

jjhursey:
* Forcing the 'hash' gds component should not be necessary any more.

Port of PR #6498 (component names changed so a cherry-pick would not work)

### DIFF
--- a/opal/mca/pmix/pmix3x/pmix3x_client.c
+++ b/opal/mca/pmix/pmix3x/pmix3x_client.c
@@ -11,6 +11,7 @@
  * Copyright (c) 2017-2018 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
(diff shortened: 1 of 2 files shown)


hppritcha:
@jjhursey  does not getting this fix into 4.0.1 cause some loss of functionality/correctness?

jjhursey:
@hppritcha It's mostly performance related for those that choose to skip the modex/fence inside MPI_Init, but may do put/get (or even a later modex/fence) operations in pmix later. I've noticed some scaling restrictions limiting the gds to hash in this mode too when used with a later modex/fence, but I don't have exact numbers there.

SMPI is impacted in one of our run configurations - so we made this change internally so I would like to see it make it into the v4.0.x series - but it doesn't necessarily need to be v4.0.1.

jjhursey:
bot:ompi:retest

hppritcha:
@jjhursey we’ve already taken another PR post rc2 so may as well Take this.

PR #6504, opened 2019-03-19
Do not force 'hash' gds on direct modex in pmix4x

jjhursey:
* Forcing the 'hash' gds component should not be necessary anymore.

### DIFF
--- a/opal/mca/pmix/pmix4x/pmix4x_client.c
+++ b/opal/mca/pmix/pmix4x/pmix4x_client.c
@@ -11,6 +11,7 @@
  * Copyright (c) 2017-2018 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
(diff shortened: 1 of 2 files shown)


jjhursey:
I missed this change for master from PR #6498

jjhursey:
The `ext2x` component was not impacted (it didn't have this code change). When we moved from the `ext2x` to the `ext3x` component is when we first noticed the regression.

I've tried to do a bit of digging to figure out why this was needed at the time it was introduced, and couldn't quite trace it back. We did some internal testing and it looked ok for our use case. I'm hoping that others will test their use cases in this mode to see if anything come up. If it does then they can always force the `PMIX_MCA_gds=hash` environment variable, so there is a workaround if it does become an issue. I don't think there will be, but you know just in case.

jjhursey:
bot:ompi:retest

PR #6510, opened 2019-03-21
shmem/fortran: Fix invalid datatype size in call to atomic cswap

sam6258:
Sending in a size of 8 for a 4 byte cswap causes problems in a couple of the atomic components. In particular, with `atomic/ucx` component, this causes ucx completion errors and wrong data. This test can be used as a reproducer for the issue: https://github.com/openshmem-org/tests-cray/blob/master/smaf/shmem_finc_inc_cswap.F90

As for the compiler optimization that is mentioned in the git blame of this file, I am not sure of this fix's effect on performance. But it is needed for correctness. 

It looks like only the v4.0.x and master branches are affected by this bug. I'll PR to v4.0.x once we approve this fix. Thanks.


gpaulsen:
@hoopoepg can you please review?

PR #6520, opened 2019-03-25
Describing Issue 6114 with v4.0.0 in README.

gpaulsen:

### DIFF
--- a/README
+++ b/README
@@ -12,7 +12,7 @@ Copyright (c) 2006-2018 Cisco Systems, Inc.  All rights reserved.
 Copyright (c) 2006-2011 Mellanox Technologies. All rights reserved.
 Copyright (c) 2006-2012 Oracle and/or its affiliates.  All rights reserved.
 Copyright (c) 2007      Myricom, Inc.  All rights reserved.
-Copyright (c) 2008-2018 IBM Corporation.  All rights reserved.
+Copyright (c) 2008-2019 IBM Corporation.  All rights reserved.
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -523,15 +523,21 @@ MPI Functionality and Features
...
+      contains examples of how to update legacy MPI applications using
+      these deleted symbols to use the "new" symbols.
 
   All that being said, if you are unable to immediately update your
   application to stop using these legacy MPI-1 symbols, you can
   re-enable them in mpi.h by configuring Open MPI with the
   --enable-mpi1-compatibility flag.
 
+  NOTE: Open MPI v4.0.0 had an error where these declarations were not
+        included in the library if configured without --enable-mpi1-compatibility
```
jsquyres: Is this right?

I don't remember the exact behavior of v4.0.0.  There are two things:

1. The declarations in `mpi.h` (and friends)
1. The symbols in `libmpi`

Which did `--enable-mpi1-compatibility` enable/disable in 4.0.0?  Your text says it was only `#1`.  Is that correct?

gpaulsen: In searching around https://github.com/open-mpi/ompi/issues/6114 would be a better Issue to reference from the README, describing what the issue with v4.0.0 was.
without `--enable-mpi1-compatibility` the header produced a deprecation (very confusing) warning message, and the symbols were un-intentionally removed from the library (i.e. link error).

gpaulsen:
I just updated commit refering to Issue 6114, which better describes the issue on v4.0.0 that was fixed on v4.0.1 (also now described in the README)

ibm-ompi:

PR #6527, opened 2019-03-27
Add compilation flag to allow unwinding through files that are present in the stack when attaching with MPIR

James-A-Clark:
This is a rebase onto master of this pull request: https://github.com/open-mpi/ompi/pull/6357


jsquyres:
ok to test

PR #6531, opened 2019-03-27
shmat/shmdt additions for patcher

markalle:
This is mostly based off recent UCX additions to their patcher:
    https://github.com/openucx/ucx/pull/2703

They added triggers for
* mmap when (flags & MAP_FIXED) && (addr != NULL)
* shmat when (shmflg & SHM_REMAP) && (shmaddr != NULL)

Beyond that I noticed they already had a trigger for
* madvise when (advice == MADV_FREE)
that we didn't so I added that.

And the other main thing is we didn't really have shmat/shmdt
active for some systems because we only had a path for
syscall(SYS_shmdt, ) but we needed to also have a path for
syscall(SYS_ipc, IPCOP_shmdt, ) and same for shmat.

### DIFF
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -15,7 +15,7 @@
  *                         reserved.
  * Copyright (c) 2016-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2019 IBM Corporation.  All rights reserved.
  *
  * $COPYRIGHT$
  *
@@ -47,6 +47,9 @@
 #endif
 #if defined(HAVE_LINUX_MMAN_H)
 #include <linux/mman.h>
+#endif
+#if defined(HAVE_SYS_IPC_H)
+#include <sys/ipc.h>
 #endif
 
 #include "memory_patcher.h"
@@ -104,15 +107,7 @@ opal_memory_patcher_component_t mca_memory_patcher_component = {
  * data. If this can be resolved the two levels can be joined.
  */
 
-/*
- * The following block of code is #if 0'ed out because we do not need
- * to intercept mmap() any more (mmap() only deals with memory
- * protection; it does not invalidate any rcache entries for a given
- * region).  But if we do someday, this is the code that we'll need.
- * It's a little non-trivial, so we might as well keep it (and #if 0
- * it out).
- */
-#if 0
+#if defined (SYS_mmap)
 
 #if defined(HAVE___MMAP) && !HAVE_DECL___MMAP
 /* prototype for Apple's internal mmap function */
@@ -121,12 +116,11 @@ void *__mmap (void *start, size_t length, int prot, int flags, int fd, off_t off
 
 static void *(*original_mmap)(void *, size_t, int, int, int, off_t);
 
-static void *intercept_mmap(void *start, size_t length, int prot, int flags, int fd, off_t offset)
+static void *_intercept_mmap(void *start, size_t length, int prot, int flags, int fd, off_t offset)
 {
-    OPAL_PATCHER_BEGIN;
     void *result = 0;
 
-    if (prot == PROT_NONE) {
+    if ((flags & MAP_FIXED) && (start != NULL)) {
         opal_mem_hooks_release_hook (start, length, true);
     }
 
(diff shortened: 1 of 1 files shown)


opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -404,6 +407,68 @@ static size_t memory_patcher_get_shm_seg_size (const void *shmaddr)
...
+
+    if ((shmflg & SHM_REMAP) && (shmaddr != NULL)) {
+// I don't really know what REMAP combined with SHM_RND does, so I'll just
+// guess it remaps all the way down to the lower attach_addr, and all the
+// way up to the original shmaddr+size
+        uintptr_t attach_addr = (uintptr_t)shmaddr;
+        
+        if (shmflg & SHM_RND) {
+            attach_addr -= attach_addr % SHMLBA;
+            size += attach_addr % SHMLBA;
```
yosefe: isn't  `attach_addr % SHMLBA` equal to 0 now, after subtracting the remainder?

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -404,6 +407,68 @@ static size_t memory_patcher_get_shm_seg_size (const void *shmaddr)
...
+        opal_mem_hooks_release_hook ((void*)attach_addr, size, false);
+    }
+
+    if (!original_shmat) {
+#if defined(SYS_shmat)
+        result = memory_patcher_syscall(SYS_shmat, shmid, shmaddr, shmflg);
+#else // IPCOP_shmat
+        unsigned long ret;
+        ret = memory_patcher_syscall(SYS_ipc, IPCOP_shmat,
+            shmid, shmflg, &shmaddr, shmaddr);
```
yosefe: indent on column

opal/mca/memory/patcher/memory_patcher_component.c
markalle: yeah, that's definitely not what I meant to do

opal/mca/memory/patcher/memory_patcher_component.c
markalle: reindented

markalle:
Thanks for the review comments, repushed

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -121,12 +116,11 @@ void *__mmap (void *start, size_t length, int prot, int flags, int fd, off_t off
...
 static void *(*original_mmap)(void *, size_t, int, int, int, off_t);
 
-static void *intercept_mmap(void *start, size_t length, int prot, int flags, int fd, off_t offset)
+static void *_intercept_mmap(void *start, size_t length, int prot, int flags, int fd, off_t offset)
 {
-    OPAL_PATCHER_BEGIN;
     void *result = 0;
 
-    if (prot == PROT_NONE) {
+    if ((flags & MAP_FIXED) && (start != NULL)) {
```
yosefe: BTW before this change, mmap(PROT_NONE) was calling release hook, and after this change - it doesn't. Is this the desired behavior?

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -404,6 +407,68 @@ static size_t memory_patcher_get_shm_seg_size (const void *shmaddr)
...
+
+    if ((shmflg & SHM_REMAP) && (shmaddr != NULL)) {
+// I don't really know what REMAP combined with SHM_RND does, so I'll just
+// guess it remaps all the way down to the lower attach_addr, and all the
+// way up to the original shmaddr+size
+        uintptr_t attach_addr = (uintptr_t)shmaddr;
+
+        if (shmflg & SHM_RND) {
+            attach_addr -= ((uintptr_t)shmaddr) % SHMLBA;
+            size += ((uintptr_t)shmaddr) % SHMLBA;
```
yosefe: IMHO it should be something like this (i will probably have to fix UCX as well):
```
            attach_addr -= ((uintptr_t)shmaddr) % SHMLBA;
            size        += ((uintptr_t)shmaddr) % SHMLBA;
            size        += (SHMLBA - (size % SHMLBA)) % SHMLBA;  /* Round size up */
```

yosefe: actually, thinking again, we can't really change the size. we got the size from shmctl() so it's the actual segment size and should be aligned to SHMLBA So we probably should not change 'size' at all, just 'attach_addr' - since the segment attach point just moves by %SHMLBA, but the segment size cannot really change.

opal/mca/memory/patcher/memory_patcher_component.c
markalle: It should be okay because previously that entire section was #if 0 so mmap() wasn't being intercepted at all. The PROT_NONE was old code and was #if 0 after a decision a long time ago that mmap PROT_NONE didn't need to be trigger a cache deregistration. So when I re-added mmap() for this commit I just used the UCX code to decide what should be intercepted.

opal/mca/memory/patcher/memory_patcher_component.c
markalle: I'm leaning toward agreeing. I agree the new mapping will have the original 'size' as its size, and it'll only occupy the range (attach_addr ..+original_size).

And when I read the manpage for REMAP I'm sure they intended to write that it replaces any mapping that existed in the address range where the new mapping goes. But they wrote more explicitly that it replaces any mapping in (addr ..+size) eg where the new mapping would go w/o RND in play.

So being logical I'd say with RND any mapping that's in (attach_addr ..+original_size) gets replaced.

I'm still maybe 5% concerned that a pedantic reading of the manpage would say that they specified the exact range that gets replaced, and the above skips a little bit on the right hand side of that range.

Is there any downside other than an unnecessary cache invalidation to my current code that uses the lower of the low ends and the higher of the high ends of the two interpretations of the manpage? My code would be telling you a certain virtual address had been invalidated when it really hadn't.

yosefe: ok, let's be pedantic

opal/mca/memory/patcher/memory_patcher_component.c
yosefe: ok, i guess if flags do not have MAP+FIXED, it will not overwrite an existing mapping anyway

PR #6532, opened 2019-03-28

### LINKED ISSUE #1089: need ability to not use --enable-new-dtags option in linker scripts.

hppritcha:
causes issues.

The new configury option is

--enable-wrappers-runpath

To disable use of --enable-new-dtags in the wrappers, add

--disable-wrappers-runpath

to the Open MPI configury line.

Fixes #1089

(cherry picked from commit ebb30c15f2a3808a51c94bf7e0f382ba096ade2f)

### DIFF
--- a/config/opal_setup_wrappers.m4
+++ b/config/opal_setup_wrappers.m4
@@ -130,6 +130,16 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
                   [enable rpath/runpath support in the wrapper compilers (default=yes)])])
     AS_IF([test "$enable_wrapper_rpath" != "no"], [enable_wrapper_rpath=yes])
     AC_MSG_RESULT([$enable_wrapper_rpath])
+
+    AC_MSG_CHECKING([if want wrapper compiler runpath support])
+    AC_ARG_ENABLE([wrapper-runpath],
+                  [AS_HELP_STRING([--enable--wrapper-runpath],
+                  [enable runpath in the wrapper compilers if linker supports it (default: enabled,  unless wrapper-rpath is disabled).])])
+    AS_IF([test "$enable_wrapper_runpath" != "no"], [enable_wrapper_runpath=yes])
+    AC_MSG_RESULT([$enable_wrapper_runpath])
+
+    AS_IF([test "$enable_wrapper_rpath" = "no" && test "$enable_wrapper_runpath" = "yes"],
+          [AC_MSG_ERROR([--enable-wrapper-runpath cannot be selected with --disable-wrapper-rpath])])
 ])
 
 # Check to see whether the linker supports DT_RPATH.  We'll need to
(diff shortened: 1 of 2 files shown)


config/opal_setup_wrappers.m4
```diff
@@ -220,18 +230,19 @@ EOF
...
-                    runpath_args="-Wl,--enable-new-dtags"
-                    AC_MSG_RESULT([yes (-Wl,--enable-new-dtags)])],
-                   [AC_MSG_RESULT([no])])
-    AC_LANG_POP([C])
+    AS_IF([test x"$enable_wrapper_runpath" = x"yes"],
+           [AC_LANG_PUSH([C])
+            AC_MSG_CHECKING([if linker supports RUNPATH])
+            AC_LINK_IFELSE([AC_LANG_PROGRAM([], [return 7;])],
+                           [WRAPPER_RPATH_SUPPORT=runpath
+                            runpath_args="-Wl,--enable-new-dtags"
```
jsquyres: This doesn't seem right.  Can you please double check this in the context of the rest of the configure script?

Specifically: `$runpath_args` is used later down in the code, but you've now only assigned it if `--disable-wrapper-runpath` was not specified.

Are the wrappers the *only* place that runpath is used?

hppritcha: hmmm looks like we may need some other commits from master, I'm surprised the cherry-pick worked without a problem.

hppritcha: actually runpath_args  is initialized empty at line 234.  the variable is used at line 315.  The admin wanting the patch on 3.0.x noticed an issue with the fortran wrappers not obeying the --enable/disable-runpath, but I'll fix that in a separate PR.

hppritcha: I was wrong.  the patch looks correct. see comments on the file.

jsquyres: Right, I see it's initialized in 234.  But it's *used* down below.  My question is: are there any other places we use runpath besides the wrappers?  Because if so, this patch is disabling runpath *everywhere*, not just for the wrappers.

If we're only using runpath for the wrappers, then this patch is likely fine.

hppritcha: It doesn't look like it.  buildrpm.sh does add --enable-new-dtags to LDFLAGS passed into configure for the spec file.  Looks like this change has been in master for almost 2 years now.

hppritcha:
looks like mellanox is having internet problems.
bot:mellanox:retest

jsquyres:
@hppritcha Should this be a PR for master, and then cherry-picked over to the relevant release branches?

jsquyres:
@hppritcha Please also PR this to v3.1.x.

hppritcha:
its already in v3.1.x  it got in before v3.1.x was branched from master

jsquyres:
@hppritcha Ah, perfect!

Issue #6522, opened 2019-03-26
Repeated calls to MPI_Cart_create segfaults or hangs when first argument is a cart comm


wrs20:
Fails in OpenMPI 4.0.0 and 3.1.3, passes in OpenMPI 2.1.1 (and MPICH).
OpenMPI installed from 4.0.0 source tar. Ubuntu 18.04.2.

Source:

```c
// cart_break.c
#include <mpi.h>
int main(int argc, char * argv[]){
   
    MPI_Init(&argc, &argv);
    MPI_Comm parent0, parent1, child0, child1;

    int ndims2[3] = {2, 1, 1};
    int ndims1[3] = {1, 1, 1};
    int periods[3] = {1, 1, 1};
    
    // parent 0 (works)
    if (MPI_Cart_create(MPI_COMM_WORLD, 3, ndims2, periods, 0, &parent0) != MPI_SUCCESS) { return -1; }

    // child0 from parent0 (works)
    if(MPI_Cart_create(parent0, 3, ndims1, periods, 0, &child0) != MPI_SUCCESS) { return -1; }


    // parent 1 (works)
    if(MPI_Cart_create(MPI_COMM_WORLD, 3, ndims2, periods, 0, &parent1) != MPI_SUCCESS){ return -1; }

    // child1 from parent1 (hangs in mpi4py, segfaults in c) passes if parent1 is replaced with either parent0 or MPI_COMM_WORLD
    if(MPI_Cart_create(parent1, 3, ndims1, periods, 0, &child1) != MPI_SUCCESS) {return -1;}
    

    // cleanup
    if (child0 != MPI_COMM_NULL){ MPI_Comm_free(&child0); }
    if (child1 != MPI_COMM_NULL){ MPI_Comm_free(&child1); }
    if (parent0 != MPI_COMM_NULL){ MPI_Comm_free(&parent0); }
    if (parent1 != MPI_COMM_NULL){ MPI_Comm_free(&parent1); }


    MPI_Finalize();
    return 0;
}

```

To reproduce:
```shell
shell$ mpicc cart_break.c
shell$ mpirun -n 2 ./a.out
```

Traceback:
```
[pypc-dm-07:10294] *** Process received signal ***
[pypc-dm-07:10294] Signal: Segmentation fault (11)
[pypc-dm-07:10294] Signal code: Address not mapped (1)
[pypc-dm-07:10294] Failing at address: 0x8
[...]

PR #6538, opened 2019-03-28
Prevent a segfault when accessing a rank outside a communicator.

bosilca:
This is not fixing any issue, it is simply preventing a sefault if the
communicator creation has not happened as expected. Thus, this code path
should never really be hit in a correct MPI application with a valid
communicator creation support.

This PR provides a better conclusion to #6522.

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_comm.h
+++ b/ompi/mca/pml/ob1/pml_ob1_comm.h
@@ -82,6 +82,16 @@ static inline mca_pml_ob1_comm_proc_t *mca_pml_ob1_peer_lookup (struct ompi_comm
 {
     mca_pml_ob1_comm_t *pml_comm = (mca_pml_ob1_comm_t *)comm->c_pml_comm;
 
+    /**
+     * We have very few ways to validate the correct, and collective, creation of
+     * the communicator, and ensure all processes have the same cid. The least we
+     * can do is to check that we are not using a rank that is outside the scope
+     * of the communicator.
+     */
+    if( OPAL_UNLIKELY(rank >= (int)pml_comm->num_procs) ) {
+        ompi_rte_abort(-1, "PML OB1 received a message from a rank outside the"
+                       " valid range of the communicator. Please submit a bug request!");
+    }
     if (OPAL_UNLIKELY(NULL == pml_comm->procs[rank])) {
         OPAL_THREAD_LOCK(&pml_comm->proc_lock);
         if (NULL == pml_comm->procs[rank]) {


ompi/mca/pml/ob1/pml_ob1_comm.h
```diff
@@ -82,6 +82,16 @@ static inline mca_pml_ob1_comm_proc_t *mca_pml_ob1_peer_lookup (struct ompi_comm
...
 
+    /**
+     * We have very few ways to validate the correct, and collective, creation of
+     * the communicator, and ensure all processes have the same cid. The least we
+     * can do is to check that we are not using a rank that is outside the scope
+     * of the communicator.
+     */
+    if( OPAL_UNLIKELY(rank >= (int)pml_comm->num_procs) ) {
+        ompi_rte_abort(-1, "PML OB1 received a message from a rank outside the"
+                       " valid range of the communicator. Please submit a bug request!");
```
jsquyres: Shouldn't this invoke the error handler on the communicator -- not just always abort?

bosilca: As the comment states, if we ever hit this case the communicator creation did something really bad and the communicator is broken globally, so there is only one realistic path forward, abort the job.

jsquyres: This is a funny statement coming from the ULFM guy.  😉 

Shouldn't we let the application try to save its own state (e.g., if they selected ERRORS_RETURN)?  I.e., yes, the state of OMPI is borked -- but it still may be desirable to do something outside the scope of MPI.

bosilca: My position remains coherent, ULFM is about controlled behaviors in well understood scenarios (mainly process fault) while this particular PR is about coping with a broken state of the MPI implementation itself.

It makes no sense to allow the error handler on the communicator in question to be called, simply because we know (due to the triggered condition) that, at least, this communicator (and potentially all communicators created after it) are completely broken (mismatched cid on the participants). If we escalate the issue to MPI_COMM_WORLD, and trigger the error handler there, it might come at an unexpected time for the application leading to even more badness.

To summarize my position, when we reach this condition there is a possibility that we have already mismatched messages in this broken communicator, which would result in the state of the MPI application being inconsistent and thus unsafe for saving.

jsquyres: Fair enough points.  Should we have an error code about that?  E.g., MPI_ERR_INTERNAL_STATE_IS_BOKED?  That would let the application decide whether it wants to save or just abort.

bosilca: It would clearly be a better approach. However, we currently profusely call ompi_rte_abort in all critical places where we do not want to clean the return path. Fixing this is a desirable long term improvement, until then this patch provides a bandaid (that hopefully will never be triggered).

jsquyres: Ok, fair enough.  Merge away.  😄

PR #6550, opened 2019-03-31
v4.0.x: Cleanup race condition in finalize that leads to incomplete vader cleanup

### LINKED ISSUE #6547: Shared memory not getting cleaned up
Using Open MPI v4.0.0 compiled from source on RHEL7.4

-----------------------------

## Details of the problem
I've already let @rhc54 know about this, but figured I'd add this here for record keeping/tracking.

On our CI testing of libfabric + Open MPI, I've noticed a lot of "vader_segments" in /dev/shm that are accumulating from all our testing. I think this is related to issue #5798 and was fixed with commit c076be52afc19b1a8c1884ff7e66b04122c7ab23 but I don't think this patch was pulled into 4.0.x. If this was fixed in a different version, which version and sorry about the spam!

rhc54:
@aingerson Please verify this fixes the problem

Fixes #6547


aingerson:
I think that did the trick. Thanks, Ralph!

hppritcha:
@ggouaillardet could you please review?

hppritcha:
Intel engineers have verified that this patch fixes the issue.

rhc54:
Just an FYI: the second commit removes stale code that cannot be reached any more as the functionality moved to PMIx. I only included it because the desired commit was based on it - i.e., I would have had to make some significant mods to the desired commit in order to get it to build/work.

hppritcha:
@gpaulsen let's merge this.

PR #6555, opened 2019-04-01
pml/ob1: fixed exit from get_frag_fail when falling back on btl_put

EmmanuelBRELLE:
In the case the btl_get fails Ob1 tries to fallback on btl_put first but
the return code was ignored. So the code fell back on both btl_put and
btl_send.
I suggest to exit the function 'mca_pml_ob1_recv_request_get_frag_failed' with an OMPI_SUCCESS in case of successful fallback on btl_put other case will instead fall back on btl_send


hjelmn:
ok to test

hppritcha:
We probably want this pushed back to release branched too.

jsquyres:
@bosilca @hjelmn Can you guys comment on this PR?  Looks like a no-brainer.

hjelmn:
Looks ok to me.

PR #6566, opened 2019-04-03
Component: Changes

davideberius:
Renamed the communicator to a more appropriate name to avoid naming collisions.


bosilca:
ok to test

PR #6567, opened 2019-04-04
opal/sys: Native timer cleanup

kawashima-fj:
Just code cleanup. See commit comment for meanings.


bosilca:
I do not see what is the benefit of ```OPAL_HAVE_SYS_TIMER_GET_FREQ```. The new code does not bring any benefit, it is exactly the same code except with one additional layer of indirection. Everything else looks like a good catch.

kawashima-fj:
@bosilca The benefit is removing `OPAL_ASSEMBLY_ARCH == OPAL_ARM64` (architecture name) from `opal/mca/timer/linux/timer_linux_component.c`. It was a kind of abstraction violation.

PR #6574, opened 2019-04-08
Fixes the O(N^2) loop in the mca_scoll_mpi_comm_query

vspetrov:
The new proc group is created from the "world_group" based on the
      ranks mapping which can be directly taken from proc_name->vpid.


jladd-mlnx:
@vspetrov please cherry pick into 4.0.x

Issue #5020, opened 2018-04-05
mpiexec --cpu-set  ORTE error 


GuillaumeMercier:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
Git repo version: v2.x-dev-5428-g17a5416

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
from a git clone (v2.x-dev-5428-g17a5416)

### Please describe the system on which you are running

* Operating system/version:  3.10.0-327.36.3.el7.x86_64
* Computer hardware:  Xeon E5 V3
* Network type: OPA

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

When trying to launch a programm   with the `--cpu-set` option, I've got this:

```shell
[miriel016:115300] [[3216,0],0] ORTE_ERROR_LOG: Data unpack would read past end of buffer in file ../../../../ompi-sources/orte/mca/odls/base/odls_base_default_fns.c at line 565
```

My command line is:
```shell
mpiexec --cpu-set 0,1,2,12,13,14 --bind-to core  -np 4 -host localhost sample_prog   4
```

ggouaillardet:
note `Git repo version` is misleading here (a known git thing ...), and the issue occurs on the (latest) `master` branch.

can you please give the inline patch a try ?

```patch
diff --git a/orte/mca/odls/base/odls_base_default_fns.c b/orte/mca/odls/base/odls_base_default_fns.c
index ff9426b..be20292 100644
--- a/orte/mca/odls/base/odls_base_default_fns.c
+++ b/orte/mca/odls/base/odls_base_default_fns.c
@@ -15,7 +15,7 @@
  *                         All rights reserved.
  * Copyright (c) 2011-2017 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2013-2018 Intel, Inc.  All rights reserved.
- * Copyright (c) 2014-2017 Research Organization for Information Science
+ * Copyright (c) 2014-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2017      Mellanox Technologies Ltd. All rights reserved.
[...]

PR #6584, opened 2019-04-12
binding: -cpu-set as a constraint rather than as a binding

markalle:
There are two commits here, but one is pretty small. The bulk of this PR is -cpu-set related:

The first category of issue I'm addressing is that recent code changes seem to only consider -cpu-set as a binding option. Eg a command like this
```
  % mpirun -np 2 --report-bindings --use-hwthread-cpus \
      --bind-to cpulist:ordered --map-by hwthread --cpu-set 6,7 hostname
```
which just round robins over the --cpu-set list.

```
Example output which seems fine to me:
> MCW rank 0: [..../..B./..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
> MCW rank 1: [..../...B/..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
```

It should also be possible though to pass a --cpu-set to most other map/bind options and have it be a constraint on that binding. Eg
```
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by hwthread --cpu-set 6,7 hostname
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by ppr:2:node,pe=2 --cpu-set 6,7,12,13 hostname
```

The first command above errors that
> Conflicting directives for mapping policy are causing the policy
> to be redefined:
>   New policy:   RANK_FILE
>   Prior policy:  BYHWTHREAD

The error check in orte_rmaps_rank_file_open() is likely too aggressive. The intent seems to be that any option like "--map-by whatever" will check to see if a rankfile is in use, and report that mapping via rmaps and using an explicit rankfile is a conflict.

But the check has been expanded to not just check
    NULL != orte_rankfile
but also errors out if
    (NULL != opal_hwloc_base_cpu_list &&
    !OPAL_BIND_ORDERED_REQUESTED(opal_hwloc_binding_policy))
which seems to be only recognizing -cpu-set as a binding option and ignoring -cpu-set as a constraint on other binding policies.

For now I've changed the
    NULL != opal_hwloc_base_cpu_list
to
    OPAL_BIND_TO_CPUSET == OPAL_GET_BINDING_POLICY(opal_hwloc_binding_policy)
so it hopefully only errors out if -cpu-set is being used as a binding policy.  Whether I did that right or not it's enough to get to the next stage of testing the example commands I have above.

Another place similar logic is used is hwloc_base_frame.c where it has
```
    /* did the user provide a slot list? */
    if (NULL != opal_hwloc_base_cpu_list) {
        OPAL_SET_BINDING_POLICY(opal_hwloc_binding_policy, OPAL_BIND_TO_CPUSET);
    }
```
where it used to (long ago) only do that if
    !OPAL_BINDING_POLICY_IS_SET(opal_hwloc_binding_policy)
I think the new code is making it impossible to use --cpu-set as anything other than a binding policy.

That brings us past the error detection and into the real functionality, some of which has been stripped out, probably in moving to hwloc-2:
```
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by hwthread --cpu-set 6,7 hostname
> MCW rank 0: [B.../..../..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
> MCW rank 1: [.B../..../..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
```

The rank_by() function in rmaps_base_ranking.c makes an array out of objects returned from
    opal_hwloc_base_get_obj_by_type(,,,i,)
which uses df_search().  That function changed quite a bit from hwloc-1 to 2 but it used to include a check for
    available = opal_hwloc_base_get_available_cpus(topo, start)
which is where the bitmask from --cpu-set goes.  And it used to skip objs that had hwloc_bitmap_iszero(available).

So I restored that behavior in ds_search() by adding a "constrained_cpuset" to replace start->cpuset that it was otherwise processing.  With that change in place the first command works:
```
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by hwthread --cpu-set 6,7 hostname
> MCW rank 0: [..../..B./..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
> MCW rank 1: [..../...B/..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
```

The other command uses a different path though that still ignored the available mask:
```
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by ppr:2:node:pe=2 --cpu-set 6,7,12,13 hostname
> MCW rank 0: [BB../..../..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
> MCW rank 1: [..BB/..../..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
```
In bind_generic() the code used to call opal_hwloc_base_find_min_bound_target_under_obj() which used opal_hwloc_base_get_ncpus(), and that's where it would intersect objects with the available cpuset and skip over ones that were't available. To match the old behavior I added a few lines in bind_generic() to skip over objects that don't intersect the available mask. After that we get
```
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by ppr:2:node:pe=2 --cpu-set 6,7,12,13 hostname
> MCW rank 0: [..../..BB/..../..../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
> MCW rank 1: [..../..../..../BB../..../..../..../..../..../..../..../....][..../..../..../..../..../..../..../..../..../..../..../....]
```

I think the above changes are improvements, but I don't feel like they're comprehensive.  I only traced through enough code to fix the two specific bugs I was dealing with.

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_frame.c
+++ b/opal/mca/hwloc/base/hwloc_base_frame.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
  * Copyright (c) 2016-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -217,7 +218,14 @@ static int opal_hwloc_base_open(mca_base_open_flag_t flags)
          * we do bind to the given cpus if provided, otherwise this would be
          * ignored if someone didn't also specify a binding policy
          */
-        OPAL_SET_BINDING_POLICY(opal_hwloc_binding_policy, OPAL_BIND_TO_CPUSET);
+// Restoring pre ef86707fbe3392c8ed15f79cc4892f0313b409af behavior.
+// Formerly -cpu-set #,#,# along with -use_hwthread-cpus resulted
+// in the binding policy staying OPAL_BIND_TO_HWTHREAD
+// I think that should be right because I thought -cpu-set was a contraint you put
+// on another binding policy, not a binding policy in itself.
+        if (!OPAL_BINDING_POLICY_IS_SET(opal_hwloc_binding_policy)) {
+            OPAL_SET_BINDING_POLICY(opal_hwloc_binding_policy, OPAL_BIND_TO_CPUSET);
+        }
     }
 
     /* if we are binding to hwthreads, then we must use hwthreads as cpus */
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -765,15 +765,34 @@ static hwloc_obj_t df_search(hwloc_topology_t topo,
         return found;
     }
     if (OPAL_HWLOC_AVAILABLE == rtype) {
+// The previous (3.x) code included a check for
+// available = opal_hwloc_base_get_available_cpus(topo, start)
+// and skipped objs that had hwloc_bitmap_iszero(available)
+        hwloc_obj_t root;
+        opal_hwloc_topo_data_t *rdata;
+        root = hwloc_get_root_obj(topo);
+        rdata = (opal_hwloc_topo_data_t*)root->userdata;
+        hwloc_cpuset_t constrained_cpuset;
+
+        constrained_cpuset = hwloc_bitmap_alloc();
+        if (rdata && rdata->available) {
+            hwloc_bitmap_and(constrained_cpuset, start->cpuset, rdata->available);
+        } else {
+            hwloc_bitmap_copy(constrained_cpuset, start->cpuset);
+        }
+
         unsigned idx = 0;
         if (num_objs)
-            *num_objs = hwloc_get_nbobjs_inside_cpuset_by_depth(topo, start->cpuset, search_depth);
+            *num_objs = hwloc_get_nbobjs_inside_cpuset_by_depth(topo, constrained_cpuset, search_depth);
         obj = NULL;
-        while ((obj = hwloc_get_next_obj_inside_cpuset_by_depth(topo, start->cpuset, search_depth, obj)) != NULL) {
-            if (idx == nobj)
+        while ((obj = hwloc_get_next_obj_inside_cpuset_by_depth(topo, constrained_cpuset, search_depth, obj)) != NULL) {
+            if (idx == nobj) {
+                hwloc_bitmap_free(constrained_cpuset);
                 return obj;
+            }
             idx++;
         }
+        hwloc_bitmap_free(constrained_cpuset);
         return NULL;
     }
     return NULL;
--- a/orte/mca/rmaps/base/rmaps_base_binding.c
+++ b/orte/mca/rmaps/base/rmaps_base_binding.c
@@ -16,6 +16,7 @@
  * Copyright (c) 2015-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2018      Inria.  All rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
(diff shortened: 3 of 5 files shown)


rhc54:
I don't see any `cpulist:ordered` binding policy in OMPI - is that some Spectrum thingy?

I'm leery of these changes unless we can demonstrate they have been tested against a wide range of mapping/binding combinations as you are touching code used by all of them - but your comments imply you have only looked at rankfile (which is the least-used mapper we have)

markalle:
The description here:
https://www.open-mpi.org/doc/v4.0/man1/mpirun.1.php
talks about --bind-to cpu-list:ordered and has some syntactic synonyms for it. It's not a spectrum addition.

But I was only using that command to show the context in which -cpu-set currently seems to be functional.

In the 3.x branch -cpu-set could be used in other contexts like these:
```
 % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by hwthread --cpu-set 6,7 hostname
  % mpirun -np 2 --report-bindings \
      --bind-to hwthread --map-by ppr:2:node,pe=2 --cpu-set 6,7,12,13 hostname
```
Those are the example commands re-enabled in this PR.

rhc54:
Dunno where that came from - I don't recognize that binding policy, but maybe someone added it?

Anyway, I still want to see some broader testing than just rankfile - as I said, you're touching code that many map/bind combinations traverse, even when --cpu-set isn't specified. We have lots of examples where someone "fixed" it for one use-case and "broke" it for others 😢

markalle:
I agree about wanting more testing. But the commands I'm re-enabling here aren't using rankfile. They're simple mapping options like --map-by hwthread or --map-by ppr:2:node,pe=2 but operating within a constrained -cpu-set:
```
% mpirun -np 2 --report-bindings \
    --bind-to hwthread --map-by hwthread --cpu-set 6,7 hostname
% mpirun -np 2 --report-bindings \
    --bind-to hwthread --map-by ppr:2:node,pe=2 --cpu-set 6,7,12,13 hostname
```

The only way I see rankfile coming into this is that the first error-check is (I think wrongly) conflating any use of -cpu-set as being equivalent to using a rankfile. I don't know all the ways -cpu-set can be used, but I think my examples above shouldn't be considered on par with the explicit bindings that come from a rankfile.

The part that I guess you're saying is most dangerous is the change that makes this statement in hwloc_base_frame.c
```
        OPAL_SET_BINDING_POLICY(opal_hwloc_binding_policy, OPAL_BIND_TO_CPUSET);
```
less active. I would like to know the original rationale for the above. But the code that's there now makes it as if the user typed --bind-to cpu-list even if all they did was add --cpu-set to another map/bind option like my examples.

But the rest of the changes are relatively simple, using hwloc_root->userdata->rdata if it exists, and is only replicating functionality that existed in 3.x prior to changing from hwloc-1 to hwloc-2. The part that isn't simple is the fact that it's likely not comprehensive -- I traced through the code for my two examples comparing old and new behavior. I suspect there are plenty more places essentially the same functionality was removed in moving from hwloc-1 to hwloc-2.

markalle:
For me this PR also appears to fix
https://github.com/open-mpi/ompi/issues/5020

gpaulsen:
Thanks.

awlauria:
@markalle @gpaulsen this never went back to the v4 series. Is this intentional?

awlauria:
I brought brack bdd92a7a643f40e76dfe7a332f4ae98b54702f9c to v4/v4.1.

However, the second commit, bf3980d70c1783c808edeba6e3732508aa384979, seems to have been fixed by another PR that did go back - I think this one specifically: https://github.com/open-mpi/ompi/pull/7643

PR #6429, opened 2019-02-25
mpiext/shortfloat: Work around empty archives

kawashima-fj:
These dummy functions are required for the following reason.

- The `libmpiext_shortfloat_{c,mpifh,usempif08}.la` files must be built because the `OMPI_EXT_MAKE_LISTS` macro in the `config/ompi_ext.m4` file adds the files to the lists of the `OMPI_MPIEXT_{C,MPIFH,USEMPIF08}_LIBS` output variables and the following files use the output variable.
    * `ompi/Makefile.am`
    * `ompi/mpi/fortran/mpif-h/Makefile.am`
    * `ompi/mpi/fortran/use-mpi-f08/Makefile.am`
- The ar command of OS X refuses to create an archive file which does not contain any object files.

The `usempi` binding is not affected because `OMPI_MPIEXT_USEMPIF_LIBS` is not used anywhere by nature. Generally it only includes `mpifh`.

See https://github.com/open-mpi/ompi/pull/6205#issuecomment-466363071

@ggouaillardet I cannot access a macOS machine. Could you test this PR?

PR #6589, opened 2019-04-15
shortfloat MacOS / Fortran fixes

jsquyres:
Short build system fixes for the "shortfloat" MPI extension.


kawashima-fj:
@jsquyres Your first commit is equivalent to #6429 (though probably you forgot `git add ompi/mpiext/shortfloat/c/mpiext_shortfloat_c_dummy.c`), which is superseded by #6431. Your second commit is unnecessary if #6431 is merged. Could you see which PR is preferred?

jsquyres:
@kawashima-fj You are exactly right.  a) shame on me for forgetting to `git add` the additional files, and b) for forgetting about #6429 and then #6431.

Let's followup on #6431.

PR #6596, opened 2019-04-16
opal_summary.m4: make message consistent with others

jsquyres:
Fix one message in opal_summary.m4 be consistent with other messages
in the same area of opal_summary.m4.


jsquyres:
False Cray failure.

bot:lanl:retest

PR #6597, opened 2019-04-16
IPUT: delay flush till remote memory is updated

hoopoepg:
- on some systems iput operation may delay delivering data
  from HCA to memory on remote side. to workaround this issue
  add non-blocking get operation to ensure that remote memory
  is updated

### DIFF
--- a/oshmem/shmem/c/shmem_iput.c
+++ b/oshmem/shmem/c/shmem_iput.c
@@ -23,6 +23,25 @@
  * been copied out of the source array on the local PE but not necessarily before the data has
  * been delivered to the remote data object.
  */
+
+/* Some transports (IB, etc) may guarantee delivery to remote HCA only, not remote memory,
+ * which may cause incorrect data. To workaround this add 'read' request of last put
+ * chunk which guarantee flush completion after remote memory is updated. */
+#if OSHMEM_SPEC_COMPAT == 1
+#  define SHMEM_REMOTE_DELIVERY(ctx, type, target, tst, nelemes, pe)       \
+        do {                                                               \
+            static type _tmp;                                              \
+            rc = MCA_SPML_CALL(get_nb(ctx,                                 \
+                                      (void*)(target + nelemes - 1 * tst), \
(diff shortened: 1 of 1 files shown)


ibm-ompi:

 Gist: https://gist.github.com/f7bb53544f987c4815cf6287b6f026e3

ibm-ompi:

 Gist: https://gist.github.com/674e0925c47b647dfe4952f340f96a31

yosefe:
@hoopoepg can this happen only with iput, or also with a sequence of non-strided put-s?

hoopoepg:
@yosefe technically it may happen on all put operations, but iput split message by small chunks which cause PCI overload

oshmem/shmem/c/shmem_iput.c
```diff
@@ -23,6 +23,25 @@
...
+#if OSHMEM_SPEC_COMPAT == 1
+#  define SHMEM_REMOTE_DELIVERY(ctx, type, target, tst, nelemes, pe)       \
+        do {                                                               \
+            static type _tmp;                                              \
+            rc = MCA_SPML_CALL(get_nb(ctx,                                 \
+                                      (void*)(target + nelemes - 1 * tst), \
+                                      element_size,                        \
+                                      (void*)(&_tmp),                      \
+                                      pe, NULL));                          \
+        }while(0)
```
dmitrygladkov: minor: wrong indentation

sam6258:
I tested the fix on the original issue and it has gone away. Looks good to me.

manjugv:
@hoopoepg Unless I'm missing something (from the code perspective), I don't think there is a need for this here. The semantics **does not** guarantee the completion on the remote memory, when it returns. It just guarantees that the source buffers are reusable. 

For PUTs, you need to ensure this on subsequent invocation of shmem_quiet / barrier.

hoopoepg:
@manjugv right, this fix just adds non-blocking get to end of put operation which guarantee delivery on flush

PR #6598, opened 2019-04-16
Remove all the linkages back to libpmix in pmix components

rhc54:
This link-back seems to be breaking OMPI for some reason. I'm not sure we need it in PMIx anyway, but we'll investigate over there.


rhc54:
@jsquyres Please merge when CI completes

rhc54:
The Cray platform is running incredibly slow and may not complete. Source of the problem isn't clear

jsquyres:
I confirm that this seems to work for me on my Mac.

PR #6602, opened 2019-04-17
common/ompio: refactor the build_io_array function

edgargabriel:
abstract out the io_array structure to be used in common_ompio_build_io_array function.
This is preparation for a future component that would like to use the same function,
but not modify the io_array stored on the file handle itself.


edgargabriel:
bot:ompi:retest

PR #6604, opened 2019-04-18
SPML/UCX: Add put_with_long_inc_all routine to shmemx.h

brminich:
The new routine transfers the data asynchronously from the source PE to all PEs in the OpenSHMEM job. The routine returns immediately. The source and target buffers are reusable only after the completion of the routine. After the data is transferred to the target buffers, the counter object is updated atomically. The counter object can be read either using atomic operations such as shmem_atomic_fetch or can use point-to-point synchronization routines such as shmem_wait_until and shmem_test.


hoopoepg:
ok to test

brminich:
to be reopened after corresponding master PR

Issue #6565, opened 2019-04-03
Memory leak with persistent MPI sends and the ob1 "get" protocol


s-kuberski:
## Background information
A memory leak appears when using persistent communication with the vader BTL and large message sizes.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
4.0.0 on local computer, 2.1.2 and 3.1.3 on clusters

### Describe how Open MPI was installed 
4.0.0: from tarball

### Please describe the system on which you are running
* Operating system/version: Ubuntu 18.04 / CentOS 7.5 / Scientific Linux release 7.5
* Computer hardware: Laptop / Intel Skylake / Intel Nehalem
* Network type: vader RDMA

-----------------------------

## Details of the problem
Running a simulation program with Open MPI, memory leaks appeared causing the application to crash. The behaviour can be reproduced with the attached code block.

When the vader BTL is used, the used memory increases linearly over time. This bug is directly connected to the message size:
With `btl_vader_eager_limit 4096` and a message size of 4041 bytes, the bug appears. If the eager limit is raised or the message size is decreased, no problem occurs. 

Only `btl_vader_single_copy_mechanism` set  to `cma` could be tested, no problem is seen if the value is set to `none`. 

If buffered communication is used and the buffer is detached and attached manually, the problem does not appear. 

Only the shared-memory communication with vader is affected. If the processes are located on different nodes or `-mca btl ^vader` is set, everything is fine.

```c
#include <stdlib.h>
#include <stdio.h>
#include "mpi.h"

int main(int argc, char **argv) {
  int rank, size, rplus, rminus, cnt, max = 4000000, i;
  char *sbuf, *rbuf;
  static MPI_Request req;
  int lbuf = 4041;

  /* Initialize MPI and assert even size */
  MPI_Init(&argc,&argv);
  MPI_Comm_size(MPI_COMM_WORLD,&size);
  MPI_Comm_rank(MPI_COMM_WORLD,&rank);

  if ( size & 1 ) {
    if ( rank == 0 ) fprintf(stderr,"ERROR: Invalid number of MPI tasks: %d\n", size);
    MPI_Finalize();
    return -1;
  }

  /* Optional arguments: max and lbuf*/
  if ( argc > 1 ) {
    if(atoi(argv[1])>0) max = atoi(argv[1]);
    if ( rank == 0 ) printf("max=%d\n", max);
  }
  if ( argc > 2 ) {
[...]

PR #6621, opened 2019-04-27
Fix the leak of fragments for persistent sends (issue #6565)

bosilca:
Fix for #6565.

The rdma_frag attached to the send request was not correctly released
upon request completion, leaking until MPI_Finalize. A quick solution
would have been to add RDMA_FRAG_RETURN at different locations on the
send request completion, but it would have unnecessarily made the
sendreq completion path more complex. Instead, I added the length to
the RDMA fragment so that it can be completed during the remote ack.

Be more explicit on the comment.

The rdma_frag can only be freed once when the peer forced a protocol
change (from RDMA GET to send/recv). Otherwise the fragment will be
returned once all data pertaining to it has been trasnferred.

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_rdmafrag.h
+++ b/ompi/mca/pml/ob1/pml_ob1_rdmafrag.h
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2013 The University of Tennessee and The University
+ * Copyright (c) 2004-2019 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
@@ -46,7 +46,8 @@ struct mca_pml_ob1_rdma_frag_t {
     mca_bml_base_btl_t *rdma_bml;
     mca_pml_ob1_hdr_t rdma_hdr;
     mca_pml_ob1_rdma_state_t rdma_state;
-    size_t rdma_length;
+    size_t rdma_length;  /* how much the fragment will transfer */
+    opal_atomic_size_t rdma_bytes_remaining;  /* how much is left to be transferred */
     void *rdma_req;
     uint32_t retries;
     mca_pml_ob1_rdma_frag_callback_t cbfunc;
@@ -71,7 +72,6 @@ OBJ_CLASS_DECLARATION(mca_pml_ob1_rdma_frag_t);
 
 #define MCA_PML_OB1_RDMA_FRAG_RETURN(frag)                              \
     do {                                                                \
-        /* return fragment */                                           \
         if (frag->local_handle) {                                       \
             mca_bml_base_deregister_mem (frag->rdma_bml, frag->local_handle); \
             frag->local_handle = NULL;                                  \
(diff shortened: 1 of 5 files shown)


ompi/mca/pml/ob1/pml_ob1_sendreq.c
```diff
@@ -262,12 +258,20 @@ mca_pml_ob1_rget_completion (mca_pml_ob1_rdma_frag_t *frag, int64_t rdma_length)
...
 
     /* count bytes of user data actually delivered and check for request completion */
     if (OPAL_LIKELY(0 < rdma_length)) {
-        OPAL_THREAD_ADD_FETCH_SIZE_T(&sendreq->req_bytes_delivered, (size_t) rdma_length);
+        frag_remaining = OPAL_THREAD_SUB_FETCH_SIZE_T(&frag->rdma_bytes_remaining, (size_t)rdma_length);
         SPC_USER_OR_MPI(sendreq->req_send.req_base.req_ompi.req_status.MPI_TAG, (ompi_spc_value_t)rdma_length,
                         OMPI_SPC_BYTES_SENT_USER, OMPI_SPC_BYTES_SENT_MPI);
+
+        if( 0 == frag_remaining ) {  /* this frag is now completed. Update the request and be done */
+            OPAL_THREAD_ADD_FETCH_SIZE_T(&sendreq->req_bytes_delivered, frag->rdma_length);
```
kawashima-fj: Just a question.

This protocol may segment a message to fragments and this function (`mca_pml_ob1_rget_completion`) is called per fragment. Before your change, `req_bytes_delivered` was decreased per fragment. After your change, `req_bytes_delivered` is decreased (to zero) when all fragments are transfered. Is it appropriate?

ompi/mca/pml/ob1/pml_ob1_rdmafrag.h
```diff
@@ -46,7 +46,8 @@ struct mca_pml_ob1_rdma_frag_t {
     mca_bml_base_btl_t *rdma_bml;
     mca_pml_ob1_hdr_t rdma_hdr;
     mca_pml_ob1_rdma_state_t rdma_state;
-    size_t rdma_length;
+    size_t rdma_length;  /* how much the fragment will transfer */
+    opal_atomic_size_t rdma_bytes_remaining;  /* how much has been currently transferred */
```
kawashima-fj: The comment is confusing. This variable holds remaining bytes.

ompi/mca/pml/ob1/pml_ob1_sendreq.c
bosilca: Indeed, before each RGET completion was updating the request itself, and never manage the life cycle of the fragment (it was release with the request, so we had a strong limit of one fragment per request). Now the RGET completion works at the fragment level, and decreases it's payload until the fragment is completed and only then it touches the corresponding request. The issue here is that we need to know when the remote got all the data from a fragment in order to be able to clean it. With this change, the fragment cleanup can be done in a single place, and the protocol can now support fragmentation of the RGETs.

ompi/mca/pml/ob1/pml_ob1_rdmafrag.h
bosilca: fixed.

jsquyres:
FWIW, I verified that this PR actually does fix the memory leak.

@kawashima-fj If you're happy, let's merge this PR (and cherry pick it to the release branches).

kawashima-fj:
@bosilca Thanks for your explanation.

I also verified the leak fix yesterday. I cannot access a terminal until the next week (because it is ten-day holiday in Japan now!). I'll cherry-pick next Tuesday or so. If you cannot wait, please cherry-pick instead of me.

Issue #6628, opened 2019-04-30
buildrpm.sh no longer respects the value of rpmtopdir


mwheinz:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

3.1.3.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Building from a git clone.

### Please describe the system on which you are running

* Operating system/version: Any RPM based distribution 
* Computer hardware: Any
* Network type: Any

-----------------------------

## Details of the problem

buildrpm.sh no longer respects the value of environment variable rpmtopdir

In OMPI 2.1.2, buildrpm.sh could work with a value of rpmtopdir that was
set in the environment. In newer versions this is no longer true,
causing such values to be ignored. This creates an issue with our updating our internal build environment to use newer versions of OMPI.

The proposed patch will be to add a new argument, "-R" to buildrpm.sh that allows the caller to pass in the name of the directory being used to build the RPMs.

```

mwheinz:
Okay - Jeff's approved the pull for Master; I've been side-tracked into other issues with our internal build system/CI, but I'll try to get the patches for the other branches pushed tomorrow morning.

PR #6629, opened 2019-04-30
buildrpm.sh no longer respects the value of rpmtopdir

mwheinz:
Refs #6628 

RPMs.

### DIFF
--- a/contrib/dist/linux/buildrpm.sh
+++ b/contrib/dist/linux/buildrpm.sh
@@ -58,6 +58,9 @@
 #    file from the tarball specified on the command line. By default,
 #    the script will look for the specfile in the current directory.
 #
+# -R directory
+#    Specifies the top level RPM build direcotry.
+#
 # -h
 #    Prints script usage information.
 #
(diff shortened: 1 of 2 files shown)


contrib/dist/linux/buildrpm.sh
```diff
@@ -267,24 +276,30 @@ fi
...
+    if test -r $file; then
+        rpmtopdir=${rpmtopdir:-"`grep %_topdir $file | awk '{ print $2 }'`"}
+    fi
 fi
+
+# If needed, initialize the $rpmtopdir directory. If no $rpmtopdir was
+# specified, try various system-level defaults.
 if test "$rpmtopdir" != ""; then
-	rpmbuild_options="$rpmbuild_options --define '_topdir $rpmtopdir'"
+    rpmbuild_options="$rpmbuild_options --define '_topdir $rpmtopdir'"
```
jsquyres: Minor quibble: it would be nice to split this commit into two commits:
1. One that makes the desired `-R` code change
1. Another that makes the whitespace fixes (and no other code/logic changes).

mwheinz: I can do that.

jsquyres: Done!

PR #6635, opened 2019-05-03
buildrpm.sh no longer respects the value of rpmtopdir

mwheinz:
Ref #6638

buildrpm.sh no longer respects the value of rpmtopdir
    
RPMs.


mwheinz:
This is on the wrong branch. Ignore.

PR #6637, opened 2019-05-03
v3.0.x buildrpm.sh no longer respects the value of rpmtopdir

mwheinz:
Refs #6628 

RPMs.

### DIFF
--- a/contrib/dist/linux/buildrpm.sh
+++ b/contrib/dist/linux/buildrpm.sh
@@ -58,6 +58,9 @@
 #    file from the tarball specified on the command line. By default,
 #    the script will look for the specfile in the current directory.
 #
+# -R directory
+#    Specifies the top level RPM build direcotry.
+#
 # -h
 #    Prints script usage information.
 #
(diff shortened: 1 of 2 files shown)


contrib/dist/linux/buildrpm.sh
```diff
@@ -267,11 +276,16 @@ fi
 # Find where the top RPM-building directory is
 #
 
-rpmtopdir=
-file=~/.rpmmacros
-if test -r $file; then
-    rpmtopdir=${rpmtopdir:-"`grep %_topdir $file | awk '{ print $2 }'`"}
+# if the user did not specify an $rpmtopdir, check for an .rpmmacros file.
+if test "$rpmtopdir" == ""; then
```
jsquyres: Actually, I think the use of `==` is wrong here.  Sorry we missed it on the master PR.  😦 

I think it works with Linux `test`, but it fails (IIRC) with BSD flavors of `test`.

It might be a little better to `if test -z "$rpmtopdir"; then` ...

jsquyres: Actually, I think this is moot -- this script is for building RPMs.  Which is specific to Linux -- not BSD.  So the use of `==` is probably ok here.

I therefore retract my prior comment.  😄

mwheinz: *phew*

PR #6643, opened 2019-05-08
btl/uct: remove this from 4.0.x release branch

### LINKED ISSUE #6640: Installation Failed! (UCX compile error)

4.0.1

distribution tarball.

* Operating system/version: Ubuntu 18.04
* Computer hardware: Tesla K40c (2 slots)
* Network type: 

-----------------------------

## Details of the problem

I am having problem with installation of openmpi v4.0.1 with cuda support. Previously I installed mpi v3.1.4 and it installs perfectly but v4.0.1 fails to install giving out following error:

```shell
btl_uct_module.c: In function ‘mca_btl_uct_reg_mem’:
btl_uct_module.c:214:22: error: ‘UCT_MD_MEM_ACCESS_REMOTE_GET’ undeclared (first use in this function); did you mean ‘MCA_RCACHE_ACCESS_REMOTE_READ’?
         uct_flags |= UCT_MD_MEM_ACCESS_REMOTE_GET;
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
                      MCA_RCACHE_ACCESS_REMOTE_READ
btl_uct_module.c:214:22: note: each undeclared identifier is reported only once for each function it appears in
btl_uct_module.c:217:22: error: ‘UCT_MD_MEM_ACCESS_REMOTE_PUT’ undeclared (first use in this function); did you mean ‘UCT_MD_MEM_ACCESS_REMOTE_GET’?
         uct_flags |= UCT_MD_MEM_ACCESS_REMOTE_PUT;
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
                      UCT_MD_MEM_ACCESS_REMOTE_GET
btl_uct_module.c:220:22: error: ‘UCT_MD_MEM_ACCESS_REMOTE_ATOMIC’ undeclared (first use in this function); did you mean ‘UCT_MD_MEM_ACCESS_REMOTE_PUT’?
         uct_flags |= UCT_MD_MEM_ACCESS_REMOTE_ATOMIC;
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                      UCT_MD_MEM_ACCESS_REMOTE_PUT
btl_uct_module.c:225:21: error: ‘UCT_MD_MEM_ACCESS_ALL’ undeclared (first use in this function); did you mean ‘UCT_MD_MEM_ACCESS_REMOTE_PUT’?
         uct_flags = UCT_MD_MEM_ACCESS_ALL;

```

Any help please?

hppritcha:
The UCX UCT layer is not really intended to be used outside
UCX itself.  The OpenUCX community does not guarantee API
compatibility of the UCT layer between major releases.

we will leave the UCT BTL in master so it can be used for
experiments, etc. but will remove it from releases, at
least the 4.0.x release.

Fixes #6640

Signed-off-by: Howard Pritchard <hpritchard@st-master.usrc>


jladd-mlnx:
👍

jsquyres:
@hppritcha I strongly suggest you do not remove this component -- it will break our backwards compatibility guarantees.

Instead, I suggest two things:

1. Make this component's `configure.m4` smarter to determine whether it can actually build or not (i.e., we should not get a compile error from btl/uct that blocks building the rest of OMPI).
1. Make this component not be the default at run time.  I understand that no one wants to support it, but we *did* ship it in v4.0, so we can't just delete it in mid-4.0.x stream.  But we *can* make it not run by default, I think.

Thoughts?

bosilca:
@thananon

gpaulsen:
perhaps we can leave everything as-is, but add some configury to v4.0.x to disable building it by default.

thananon:
I have no objection in removing it from 4.0.x but I just want to drop this here as a reference.

- Without btl/uct and btl/ofi, we kinda lost osc/rdma support too. I'm not sure if each PML have their one-sided API implemented.
- btl/uct is performing best in MT, even for two sided, by miles. If this get removed, when someone look for multithreaded support, we will have to point them to master instead.

jsquyres:
We can't remove it from v4.0.x, and we should keep building it by default (if possible), because that's what we did in v4.0.0.  Changing that in mid-stream of the v4.0.x release train is probably not a good idea (IMNSHO).

I'm personally in favor of the 2 recommendations I made above (https://github.com/open-mpi/ompi/pull/6643#issuecomment-490664889).  Other thoughts on that?

rhc54:
I agree that it cannot be removed from v4.0.x, especially if it _does_ function with at least some UCX versions and provides a performance benefit to boot. Seems to me that extending the configure logic to ensure it builds only with compatible UCX versions makes sense (along with something in the FAQ about supported versions), and it should remain as the default when it does build since that is the current priority. Question then becomes: who is going to do the configure update and the FAQ entry?

PR #6644, opened 2019-05-09
adding OMP_PLACES%d and basic CUDA_VISIBLE_DEVICES for LSF

markalle:
LSF can set OMP_PLACES%d for each rank, with the expectation that
MPI will set OMP_PLACES to the corresponding value (where %d is
a 1-based  host-local index of the rank). LSF also can provide a list
of CUDA_VISIBLE_DEVICES that it might want split between the ranks.

This checkin looks for an OMP_PLACES%d for each rank, and also
parses the CUDA_VISIBLE_DEVICES to be handed out to the ranks in a
round robin manner.

We support a comma separated list of somewhat arbitrary strings
like GPU-1234-abcd in CUDA_VISIZBLE_DEVICES. An invalid GPU entry
in the list is used to signify that all entries past that point are
to be ignored.  We can't really recognize all invalid GPU strings,
but we will recognize negative integers as invalid.

We also use
    --mca ncuda_per_rank <number>
as the control for how many GPUs to hand out to each rank, with
the option
    --mca ncuda_per_rank all
which assignes the full list of CUDA_VISIBLE_DEVICES to each rank.

Also LSF will support multiple MPS on one host, so we translate
CUDA_MPS_PIPE_DIRECTORY1/2/3... to a CUDA_MPS_PIPE_DIRECTORY
environment variable per specified rank.

### DIFF
--- a/orte/mca/schizo/ompi/schizo_ompi.c
+++ b/orte/mca/schizo/ompi/schizo_ompi.c
@@ -17,7 +17,7 @@
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
  * Copyright (c) 2015-2019 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
- * Copyright (c) 2018      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2018-2019 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -51,6 +51,8 @@
 #include "orte/runtime/orte_globals.h"
 
 #include "orte/mca/schizo/base/base.h"
+#include "orte/mca/schizo/schizo.h"
+#include "orte/mca/schizo/ompi/schizo_ompi.h"
 
 static int define_cli(opal_cmd_line_t *cli);
 static int parse_cli(int argc, int start, char **argv);
@@ -1182,6 +1184,9 @@ static int setup_fork(orte_job_t *jdata,
 }
 
 
+static int parse_rank_specific_launcher_settings(int mylrank, int mynlranks,
+    orte_app_context_t *app, char ***env);
+
 static int setup_child(orte_job_t *jdata,
                        orte_proc_t *child,
                        orte_app_context_t *app,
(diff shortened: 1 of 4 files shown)


jjhursey:
Since we need to resolve this envar before the fork of the process (so the OpenMP runtime will pick it up) we needed to hook into the `setup_child` call. We considered forking `schizo/ompi` into `schizo/ompi_lsf` (with proper checks for LSF environments), but that didn't seem maintainable. There didn't seem to be another way to stack on top of the `schizo/ompi` with an LSF specific modification. Is there a better place to put this code?

What we could do is:
 * Add a check before the new `parse_rank_specific_launcher_settings` call that only triggers it in an LSF environment.
 * Add an MCA parameter that disables the feature altogether if the user desires.

If I remember correctly, LSF only injects these envars if the user asks for it to as `bsub` time. So they are already opting-in to the feature.

rhc54:
That's great for LSF, but may not reflect what everybody else does 😄 

The schizo components are scanned in priority order, so if you need something lsf-specific to execute after the schizo/ompi component, then just add an lsf component and give it a lower priority. Ensure it only builds when LSF is present so you don't mess everyone else up.

What seems odd here is that you are implying these envars are assigned at time of allocation, which means you are expecting to only run one job during an allocation? I would have expected them to be assigned at the job level as that would make more sense, but I don't know much about LSF. If they are assigned per-job, then the problem will be that mpirun won't see them as it is started at the cmd line, but the daemons will. Still okay so long as no procs are started by mpirun itself, but that might be something you'd need to force if/when someone wants this option and it would probably take some mods to the plm/lsf component to get the cmd line right for starting the daemons.

Again, I'll only caution you that a similar "feature" was added to master once before and generated major backlash due to unexpected behavior for existing apps. This was pre-Spectrum days, so maybe feelings have changed, but at that time IBM's OMP team was one of the complainers. IIRC, that implementation passed the MPI bindings down to OMP so they would know what cores had been assigned to the proc, so maybe this is different enough that your OMP guys won't get annoyed again.

However, you will wind up with the rather odd situation where the MPI proc is bound to one set of resources, and the OMP threads think they can/should use some other place. This sounds to me like trouble, especially if they choose different NUMA locations.

jjhursey:
Thanks for the note about the schizo framework. We could leverage that to have a specific component just for this feature that would run after the ompi component for the `setup_child` function.

I see your concern for the general case. I agree that we could get into situations where the user would not want these to be resolved - so an MCA parameter would be useful here. Particularly if the user is trying to use mpirun binding options in addition to the LSF binding options.

It was my understanding that these envars are only set if the user requests the affinity via LSF. So there is an LSF affinity file that determines the process mapping and as long as the user doesn't try to fight that LSF binding that they requested then the envars will be correct. But if they want to tinker inside that allocation then having an MCA option to disable this feature/component would be useful.

rhc54:
You should probably look at ras/lsf as it already reads the LSF affinity file and adjusts accordingly. Not sure how that will interact with these changes. You also might want to set the default binding policy to "no-bind" when these are set to avoid unexpected conflicts.

Finally, it would probably be worth looking at this "set at allocation" vs "set at execution" issue. I still don't understand how LSF can set these values at allocation as it doesn't have any idea how many procs you are going to run via mpirun nor any way to constrain that number. Remember, the way we launch in OMPI under mpirun, LSF _never_ sees the actual MPI processes - it only sees a single process running on each node (that just happens to be our orted). So as far as LSF is concerned, there is a job that executes one proc/node. Only mpirun knows that there are multiple procs running underneath that daemon - LSF has no idea they exist.

I'm guessing that the person requesting this feature doesn't understand that situation, or is expecting that the user will request an allocation that exactly matches the number of procs they will subsequently start via mpirun. If so, then it would be wise to check that assumption is true or else you will get some quite unexpected behaviors.

jjhursey:
There is a scenario where this is functional and necessary that is driven by at least one client. We will talk with the LSF team about the concerns.

AboorvaDevarajan:

gpaulsen:
@jjhursey I think you've talked to LSF about this or something similar.  Is this PR still desirable?  Any changes needed for it?

jjhursey:
@gpaulsen We do still want this PR, but I have a couple of notes to polish it off. It does have the assumption that the user told LSF how it wants to map/order/bind and mpirun should launch according to that spec. I think that is fine in the basic usage case that LSF is targeting here.

Related PR #6869 which allows overriding of the LSF affinity file if the user adjusts the map/bind/order policy. We may want to extend the ras/lsf to disable this feature if the user specifies their own directive on the command line. Maybe by simply adding `--mca schizo ^lsf` to the launching environment to disable this feature. Maybe we can do something more subtle than that, but that would work for now.

We will want to make the MCA parameter `ncuda_per_rank` part of the schizo component that way it's all compartmentalized to that one component. It's only used by this component so should be declared with it instead of as an ORTE global option.

@markalle Is this something you can work on to finish up this ticket? Let me know if you need a hand.

ibm-ompi:

 Gist: https://gist.github.com/e05a2adbaa4555978a591b1ba35ca50b

rhc54:
Just saw the CI run on this PR - can we drop this thing? If you want to retarget it to PRRTE, it will need some major changes as I'm still uncomfortable with you setting ```OMP_PLACES``` in the ```OMPI``` component where it applies to everyone. This looks highly LSF-specific (and likely tightly tied to a particular compiler too, which is what broke the earlier attempt) and should be implemented as such.

jjhursey:
Yes, we should close this PR since ORTE is gone from `master`. We should try to port it to PRRTE with some of the notes from this thread. @markalle and I can work on that.

PR #6645, opened 2019-05-09
btl/ugni: fix 32-bit compare-and-swap atomics

hjelmn:
This commit fixes an error in the 32-bit compare-and-swap atomic support
for Aries networks. The code was incorrectly using the non-fetching
version of cswap which was causing the routing to return
OPAL_ERR_BAD_ARG.


hjelmn:
@devreal This is one of two issues preventing network atomics for 32-bit on Aries. Will submit a PR for the other as well.

hjelmn:
@hppritcha FYI. Performance bug should go into all releases.

hjelmn:
:bot:retest:ompi

PR #6646, opened 2019-05-09
osc/rdma: fix CAS 32-bit network atomic compatibility check

hjelmn:
When checking for btl compatibility with 32-bit CAS osc/rdma was
checking the incorrect flag field.


hjelmn:
@devreal This is the other half of the 32-bit CAS performance fix.

Issue #6575, opened 2019-04-08
Possibly, incorrect MPI_Type_indexed behavior with zero blockLenght


vspetrov:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
master

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
compiled from source, git clone open-mpi/ompi/master

### Please describe the system on which you are running

* Operating system/version: RHEL 7.4 
* Computer hardware: x86
* Network type: n/a

-----------------------------

## Details of the problem
Consider the test case:
```
#include <mpi.h>
#include <stdio.h>

int main (int argc, char **argv) {
    int rank, size;
    MPI_Datatype dtype;
    int dt_count = 1;
    int pBlockLength[1] = {0};
    int pDisps[1] = {0};
    MPI_Aint lb, extent;

    MPI_Init(&argc, &argv);

    MPI_Type_indexed(dt_count, pBlockLength, pDisps, MPI_DOUBLE_COMPLEX, &dtype);
    MPI_Type_commit(&dtype);
    MPI_Type_get_extent(dtype, &lb, &extent);

    printf("lb: %llu, extent %llu\n",
           (long long unsigned)lb, (long long unsigned)extent);
    MPI_Finalize();
}
```
Compile and run on a single node:
```
mpicc mpi_create_indexed.c -g -o mci
mpirun -np 1 ./mci
lb: 9223372036854775807, extent 1
```

In this test user creates an Indexed dtype with count = 1 and pBlockLength[0] = 0;
The resulting dtype has huge LB  and extent = 1. I would expect this dtype to have LB=UB=TRUE_UB=TRUE_LB=EXTENT=0. 

1. Do i understand correctly that the test is MPI spec compliant ? From the spec i read that the entries of pBlockLength have to be non-negative. So all looks legit. 
2. Do you agree that the resulting dtype must have zero extent and lb,ub?
3. MPICH handles this case this way: [mpich_type_indexed](https://github.com/pmodels/mpich/blob/66218a1a6ab2f58e6fd1c5eef3968aa6ac32d76d/src/mpi/datatype/type_indexed.c#L158) So, the check if at least one entry of pBlockLength is non-zero and only then create the dtype otherwise mpich would return an empty type.

[...]

PR #6647, opened 2019-05-09
Fix/length 0

bosilca:
Change the datatype creation function to return a duplicate of MPI_DATATYPE_NULL if the count or the blockLength for the datatype is 0. Fix the format in the datatype_dump function to prevent any compile warnings.

Provide a fix for #6575.

### DIFF
--- a/ompi/datatype/ompi_datatype_create_contiguous.c
+++ b/ompi/datatype/ompi_datatype_create_contiguous.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2013 The University of Tennessee and The University
+ * Copyright (c) 2004-2019 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2006 High Performance Computing Center Stuttgart,
@@ -29,13 +29,12 @@ int32_t ompi_datatype_create_contiguous( int count, const ompi_datatype_t* oldTy
 {
     ompi_datatype_t* pdt;
 
-    if( 0 == count ) {
-        pdt = ompi_datatype_create( 0 );
-        ompi_datatype_add( pdt, &ompi_mpi_datatype_null.dt, 0, 0, 0 );
-    } else {
-        pdt = ompi_datatype_create( oldType->super.desc.used + 2 );
-        opal_datatype_add( &(pdt->super), &(oldType->super), count, 0, (oldType->super.ub - oldType->super.lb) );
+    if( (0 == count) || (0 == oldType->super.size) ) {
+        return ompi_datatype_duplicate( &ompi_mpi_datatype_null.dt, newType);
     }
+
+    pdt = ompi_datatype_create( oldType->super.desc.used + 2 );
+    opal_datatype_add( &(pdt->super), &(oldType->super), count, 0, (oldType->super.ub - oldType->super.lb) );
     *newType = pdt;
     return OMPI_SUCCESS;
 }
--- a/ompi/datatype/ompi_datatype_create_darray.c
+++ b/ompi/datatype/ompi_datatype_create_darray.c
@@ -192,9 +192,7 @@ int32_t ompi_datatype_create_darray(int size,
     if (ndims < 1) {
         /* Don't just return MPI_DATATYPE_NULL as that can't be
            MPI_TYPE_FREE()ed, and that seems bad */
-        *newtype = ompi_datatype_create(0);
-        ompi_datatype_add(*newtype, &ompi_mpi_datatype_null.dt, 0, 0, 0);
-        return MPI_SUCCESS;
+        return ompi_datatype_duplicate( &ompi_mpi_datatype_null.dt, newtype);
     }
 
     rc = ompi_datatype_type_extent(oldtype, &orig_extent);
--- a/ompi/datatype/ompi_datatype_create_indexed.c
+++ b/ompi/datatype/ompi_datatype_create_indexed.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2013 The University of Tennessee and The University
+ * Copyright (c) 2004-2019 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2006 High Performance Computing Center Stuttgart,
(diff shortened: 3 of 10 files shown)


markalle:
I like it except I'm a little uncomfortable with the degenerate cases.  I think all of these should indeed produce size-0 types, but some of them should have extra lb/ub settings carrying over too.

The standard puts the lb/ub entries in the typemap, and describes the type creations in terms of what the new typemap looks like relative to the old one.  So I think the standard takes us to some strange looking results.  Eg take MPI_Type_contiguous().  If the count is non-0 and the incoming datatype is size-0 but still has lb/ub fields, then the new datatype would still contain special lb/ub values.  On the other hand if the count==0 I think I'd agree the result is a dup of null.dt because the typemap of the new type contains no entries and logically wouldn't carry over any lb/ub info.

I have to admit I'm having a hard time coming up with examples where the extent meta-data on a size-0 datatype really matters though.

jsquyres:
I'm a little confused.  Shouldn't this

```c
MPI_Datatype datatype;
MPI_Type_contiguous(0, MPI_INT, &datatype);
```

give me a valid datatype in `datatype`?  If I understand this PR properly, I'll instead get a clone of `MPI_DATATYPE_NULL` back, which is an invalid datatype.

Indeed, consider these 4 cases:

```c
// 1. This is valid
MPI_Send(0, MPI_INT, ...);

// 2. This is invalid / *can* trigger a run-time error
// (I don't think it *must* generate a run-time error because
// the count is 0, but it's valid for an MPI implementation to error)
MPI_Send(0, MPI_DATATYPE_NULL, ...);

// 3. This devolves to the same as case 2
MPI_Datatype datatype;
MPI_Type_contiguous(0, MPI_INT, &datatype);
MPI_Send(0, datatype, ...);

// 4. What behavior is expected here?
MPI_Send(4, datatype, ...);
```

For case 4, if you just sent 4 datatypes of 0 length, that should be fine -- you send an empty message.  But if you send 4 `MPI_DATATYPE_NULL` datatypes, that will definitely error.

...right?

bosilca:
In answer to @markalle question. Part of the beauty of losing the MPI_LB and MPI_UB hard markers in MPI 3.0 is that there is nothing to propagate when dealing with 0 bytes data as the soft markers are also subject to count. So I think we are good on that side.

To @jsquyres you are right we should not return directly the copy of MPI_DATATYPE_NULL, we need to make it valid first. Which explain the old code where we were add the datatype_null to an empty datatype. Good catch!

bosilca:
@jsquyres I take that back, sorry. The code is correct as is, we do not allow explicit use of MPI_DATATYPE_NULL for communication, but a duplicate is safe.

Here is  a quick validator
```c
#include <mpi.h>

int main( int argc, char* argv[] )
{
  int me;
  MPI_Datatype type;

  MPI_Init(NULL, NULL);

  MPI_Comm_rank(MPI_COMM_WORLD, &me);

  MPI_Type_contiguous(0, MPI_INT, &type);
  MPI_Type_commit(&type);

  if( 0 == me ) {
    MPI_Send(&me, 0, type, 1, 10, MPI_COMM_WORLD);
  } else if( 1 == me ) {
    MPI_Recv(&me, 0, type, 0, 10, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
  } /* else do nothing */

  MPI_Type_free(&type);
  MPI_Finalize();
}
```

bosilca:
This PR is ready to go.

jsquyres:
@bosilca What happens if you change the count of 0 in the send/receive calls to 4?

bosilca:
It still works. MPI_DATATYPE_NULL is in fact an zero-fied datatype (opal_datatype_empty) predefined and committed. We explicitly prevent MPI_DATATYPE_NULL from being used in communications and duplication, via a pointer comparison at the MPI level. As the duplicate here is done using our internal datatype functions, we obtain a totally valid, zero-fied datatype. 👍

jsquyres:
Roger!

ibm-ompi:

 Gist: https://gist.github.com/abf798fcf1ca238442688d0afba49c27

ibm-ompi:

 Gist: https://gist.github.com/f6cbea340431abcf1c291efa59d33a0d

bosilca:
You're right the changes are a little pedantic, but we better be on the safe side. I don't think the MPI standard mandates what to return as lb and ub for an empty datatype, but sane values (0 and 0) are definitively user-friendly.

markalle:
But the standard only removed the explicit user-level MPI_LB and MPI_LB markers didn't it?

They still have MPI_Type_create_resized() that sets the lb and ub markers. And that metadata is visible via MPI_Type_get_extent().  Unless I'm missing something the typemap at the heart of the MPI standard datatypes still contains lb/ub just as much as the standard did pre 3.0.

So for example if I had
```
    // dt1 = basic size-0 type
    MPI_Type_contiguous(0, MPI_INT, &dt1);
    MPI_Type_commit(&dt1);
    // dt2 = size-0 with LB -8 and extent 20
    MPI_Type_create_resized(dt1, -8, 20, &dt2);
    MPI_Type_commit(&dt2);
    // dt3 = type contiguous 10x of dt2
    MPI_Type_contiguous(10, dt2, &dt3);
    MPI_Type_commit(&dt3);
```
I think the typemaps above would amount to
```
dt1 = { }
dt2 = { (LB, -8), (UB, 12) }
dt3 = { (LB, -8), (UB, 12),
             (LB, -8 + 20), (UB, 12 + 20),
             (LB, -8 + 40), (UB, 12 + 40),
             (LB, -8 + 60), (UB, 12 + 60),
             (LB, -8 + 80), (UB, 12 + 80),
             (LB, -8 + 80), (UB, 12 + 100),
             (LB, -8 + 80), (UB, 12 + 120),
             (LB, -8 + 80), (UB, 12 + 140),
             (LB, -8 + 80), (UB, 12 + 160),
             (LB, -8 + 80), (UB, 12 + 180) }
that reduces to
dt3 = { (LB, -8), (UB, 192) }
``` 
so if we queried the last datatype with MPI_Type_get_extent() I'd expect to see LB -8 and extent 200.

Having said all that, I'm still not sure I care. The extent of a size-0 type has no use (beyond the fact that it can be explicitly queried).

markalle:
(hehe, I failed to update some of the numbers in the above, but you get the idea... that's what I get for using a contiguous count of 10)

bosilca:
MPI_LB and MPI_UB were hard markers, and has specific rules on how they propagate. The new resized capability is considered as soft markers and they propagate more naturally without special rules/exceptions. So, with the old MPI_LB and MPI_UB your example would have been correct and valid, because even if the datatype contained no data per sei, it had the markers that needed to be explicitly propagated. The new scheme (resize) prevents you from viewing the operations as you did above, simply because markers do not exists anymore, and therefore an empty datatype is an empty datatype whatever boundaries you try to set to it.

Obviously this is my own interpretation of the MPI datatype chapter, and might be wrong.

PR #6649, opened 2019-05-10
 OSC rdma: make sure accumulating in shared memory is safe 

### LINKED ISSUE #6536: Regression of osc_rdma_acc_single_intrinsic
There seems to be a regression in Open MPI's implementation of atomic operations if the MCA parameter `osc_rdma_acc_single_intrinsic` is set to `true`: the attached test case results in what appears to be random results whereas without the MCA parameter the results meet the expectations. 

Example runs: 

```
$ mpirun -n 2 -N 1 ./mpi_fetch_op_local_remote
result:1000
```

```
$ mpirun -n 2 -N 1 -mca osc_rdma_acc_single_intrinsic true ./mpi_fetch_op_local_remote
result:1015
mpi_fetch_op_local_remote: mpi_fetch_op_local_remote.c:98: main: Assertion `sum == 1000*(comm_size-1)' failed.
```
[mpi_fetch_op_local_remote.tar.gz](https://github.com/open-mpi/ompi/files/3017913/mpi_fetch_op_local_remote.tar.gz)

Built with:
```
$ mpicc mpi_fetch_op_local_remote.c -o mpi_fetch_op_local_remote
```

I just tested this with the 4.0.0 release. Setting the parameter with 3.1.2 works as expected. 
This problem was observed on both a Cray XC40 and an IB-based cluster. Interestingly, the issue is only present if the local rank performs atomic updates (subtracting and readding a value), otherwise everything is fine.

I first reported this on the user ML but lost track of it: https://www.mail-archive.com/users@lists.open-mpi.org/msg32834.html

devreal:
Even if the peer is reachable through shared memory there may still be remote peers updating directly through the network, which leads to conflicting updates, i.e., if `osc_rdma_acc_single_intrinsic` is set. This PR adds more thorough checks to make sure it's safe to accumulate in shared memory.

Fixes #6536 

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma.h
+++ b/ompi/mca/osc/rdma/osc_rdma.h
@@ -147,6 +147,9 @@ struct ompi_osc_rdma_module_t {
 
     bool acc_use_amo;
 
+    /** whether the group is located on a single node */
+    bool single_node;
+
     /** flavor of this window */
     int flavor;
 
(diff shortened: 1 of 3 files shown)


jsquyres:
ok to test

hppritcha:
@devreal could you PR this over to 4.0.x?

PR #6650, opened 2019-05-10
slurm-19.05.0-0rc1 srun needs dash

jdhayes:
The underscore (--cpu_bind) is no longer recognized as a valid srun option.
I changed this to be a dash (--cpu-bind), which is now supported.

### DIFF
--- a/orte/mca/plm/slurm/plm_slurm_module.c
+++ b/orte/mca/plm/slurm/plm_slurm_module.c
@@ -278,7 +278,7 @@ static void launch_daemons(int fd, short args, void *cbdata)
      * constraint, but will ensure it doesn't get
      * bound to only one processor
      */
-    opal_argv_append(&argc, &argv, "--cpu_bind=none");
+    opal_argv_append(&argc, &argv, "--cpu-bind=none");
 
 #if SLURM_CRAY_ENV
     /*
--- a/orte/mca/plm/slurm/plm_slurm_module.c
+++ b/orte/mca/plm/slurm/plm_slurm_module.c
@@ -31,6 +31,9 @@
 
 #include <string.h>
 #include <sys/types.h>
+#include <sys/wait.h>
+#include <errno.h>
+#include <sysexits.h>
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
@@ -165,6 +168,44 @@ static int plm_slurm_launch_job(orte_job_t *jdata)
     return ORTE_SUCCESS;
 }
 
+/* Slurm 19 has removed support for --cpu_bind, and now only
+*  uses --cpu-bind. Use this function to test for
+*  supported srun parameters.
+*/
+int srun_check_parms(char *const argv[])
+{
+    int status;
+    int srun_pid;
+
+    srun_pid = fork();
+    if (0 == srun_pid) {
+        FILE* file = fopen("/dev/null", "w");
+        int fd = fileno(file);
+
+        dup2(fd, 1);
+        dup2(fd, 2);
+
+        execvp(argv[0], argv);
+
+        close(fd);
+        exit(errno);
+    }
+    else{
+        /* Wait for child process to end */
+        wait(&status);
+
+        /* Once child process, srun, has ended check exit code */
+        if (WIFEXITED(status)) {
+            if (255 == WEXITSTATUS(status)) {
+                return(0);
+            }
+        }
+    }
+    /* All other exit codes mean parms are good */
+    return(1);
+}
+
+
 static void launch_daemons(int fd, short args, void *cbdata)
 {
     orte_app_context_t *app;
@@ -278,7 +319,18 @@ static void launch_daemons(int fd, short args, void *cbdata)
      * constraint, but will ensure it doesn't get
      * bound to only one processor
      */
-    opal_argv_append(&argc, &argv, "--cpu-bind=none");
+    int srun_good_parms;
+    /* Use dummy job to check srun parms */
+    char *const srun_argv[] = {"srun","--cpu_bind=none","--test-only", NULL};
+    /* Check srun params, returns 0 if bad */
+    srun_good_parms = srun_check_parms(srun_argv);
+
+    if (1 == srun_good_parms) {
+        opal_argv_append(&argc, &argv, "--cpu_bind=none");
+    }
+    else {
+        opal_argv_append(&argc, &argv, "--cpu-bind=none");
+    }
 
 #if SLURM_CRAY_ENV
     /*


rhc54:
We are likely to need some kind of version checking, then, as earlier versions of Slurm will be broken with this change.

jdhayes:
Can the 'srun' command just use both (with dash and with underscore) as a
kind of alias?
This seems reasonable to me, keeping backward compatibility with OMPI
versions.

rhc54:
Don't know - what does srun do when it doesn't recognize a command line option? Is Slurm accepting both variations?

jdhayes:
Currently srun crashes when using the underscore, so calling mpirun
directly does not work under Slurm.
   https://bugs.schedmd.com/show_bug.cgi?id=6993

jsquyres:
FWIW, this sounds like SLURM broke us: they changed their acceptable CLI options.  😦

I think this is why @rhc54 was saying we'll need version checking somehow: Open MPI supports many versions of SLURM; it's going to need to decide whether to pass `--cpu_bind` or `--cpu-bind`. 

This kinda sucks, because we'll have to check the version at run time (this can't be decided when Open MPI is configured or built, because you could configure Open MPI with one version of SLURM and then run it with another).  Super sad panda.  ☹️

jdhayes:
A cheap and easy way?
   srun --version
   slurm 19.05.0-0rc1

jdhayes:
Or...
Oh, how about running the srun command first with the "--test-only" flag to
test the srun parameters?

srun --test-only --ntasks-per-node=1 --kill-on-bad-exit
--cpu_bind=none --nodes=1 --nodelist=c39 --ntasks=1 orted -mca ess
"slurm" -mca ess_base_jobid "2507341824" -mca ess_base_vpid "1" -mca
ess_base_num_procs "2" -mca orte_node_regex "c[2:38-39]@0(2)" -mca
orte_hnp_uri "2507341824.0;tcp://10.102.11.38,10.104.11.38:49214"
--mca plm_base_verbose "100"

If the exit code is a success, then it works (all flags are good).

jsquyres:
How far back is `--test-only` supported?  I.e., what version of SLURM started accepting it?

jdhayes:
Slurm 16.05.4 has it.
How far back do you need it?
Should I check further back?

jdhayes:
Found it in a NEWS file under the 16.05.4 source code:
   * Changes in Slurm 14.03.0pre5
 -- Add --test-only option to sbatch command to validate the script and
options.

jdhayes:
Oh, sorry that is for sbatch.
I guess srun has supported it for longer:
    * Changes in Slurm 2.5.7
    -- srun - Don't check for executable if --test-only flag is used.

jdhayes:
The first fix that I can find for srun --test-only was for SLURM 2.1.10.
Perhaps this feature has been there since the start?

jsquyres:
Amusingly enough, I have SLURM 14.03 on my cluster (gasp!).  So I'd be good.

@rhc54 Any idea how far back we need to support this?  I'm kinda guessing that 14.03 isn't far enough -- i.e., `--test-only` isn't going to be a good enough test.  Instead, it'll likely need to be:

* run `srun --version` and looking for the magic version where `cpu_bind` changed to `cpu-bind`, or
* run with `cpu_bind`, and if that fails, try running again with `cpu-bind`

jdhayes:
Seems to have happened here:

https://github.com/SchedMD/slurm/commit/a57fafb263506a439697c1dbd270a031f6f17d16

jdhayes:
I guess that was originally mentioned in RELEASE NOTES FOR SLURM VERSION
17.11.
However it was supporting both underscore and dash at that time. Then it
stops supported underscore later...
I'll see if I can find it the exact commit/version.

jdhayes:
I have been trying to track this file through time:
   https://github.com/SchedMD/slurm/blob/slurm-18.08/src/srun/libsrun/opt.c

At some point after 19.05.0 0pre1 "--cpu_bind" was removed.
There are many commits and merges, it is hard to tell just by manually
clicking though Github.

rhc54:
> The first fix that I can find for srun --test-only was for SLURM 2.1.10. Perhaps this feature has been there since the start?

Going back that far is more than enough - I don't think we support anywhere near that far back anyway. I guess you could try it with `cpu_bind` and if that fails try `cpu-bind`. If both fail, then abort. Whichever succeeds, re-issue it without the `--test-only`

jdhayes:
Before
https://github.com/SchedMD/slurm/blob/7e7fd1bc649691516a357f012aa1415308d6f54e/src/srun/libsrun/opt.c

Then after this commit, some how "Merge branch 'slurm-18.08' into
slurm-19.05" removed it?
https://github.com/SchedMD/slurm/blob/62dc419e8be24535da33a548afa5f86cdb743568/src/srun/libsrun/opt.c

rhc54:
AFAICT the `--test-only` option remains - they simply moved the code that defined it elsewhere, but the code that executes it is still present.

jdhayes:
Oh, yes. I am sure that is true. The --test-only option works in the latest
version of srun.

I am trying to find the exact commit where the "--cpu_bind" is no longer
supported.
It looks like it occurs between those 2 commits.
Although I am unsure exactly where.

rhc54:
Does it really matter? If we are going to test anyway, I don't see what difference it makes when they created the breakage.

jdhayes:
I agree.
Although, Jeff had asked, so I am trying my best to help :)

rhc54:
I see - he was actually just wondering what release version it changed in, not the precise hash, in case we had to actually check the Slurm version. Sorry for the confusion.

jsquyres:
Yes, @rhc54 is right -- sorry, I wasn't so much looking for the exact commit as the release version.  Sorry!

But he's also right: if you're going to update this PR to test / run both versions, then we don't much care about the version any more.

For future reference, you might have better luck trying to chase down git history through the git command line.  The github UI is pretty great, but the git command line -- especially for history spelunking -- is quite a bit more powerful.

jdhayes:
Oh, I see...overachiever, oops :)
I like command line. Maybe I'll try that next time.

OK, then you already have enough info regarding Slurm end of things.

Here is a POC that I drafted, would something like this work right before
the srun is executed:

main()
{
    char *const argv[] = {"srun","--cpu_bind=none","--test-only", NULL};

    if (fork() == 0) {
    int fd = open("/dev/null");

    dup2(fd, 1);   // make stdout go to file
    dup2(fd, 2);   // make stderr go to file - you may choose to not do this
                   // or perhaps send stderr to another file

    close(fd);     // fd no longer needed - the dup'ed handles are
sufficient        execvp(argv[0], argv);
        exit(errno);
    }
    else{
      //parent stuff
      int status;
      wait(&status);       /*you made a exit call in child you
                           need to wait on exit status of child*/
      if(WIFEXITED(status))
         printf("child exited with = %d\n",WEXITSTATUS(status));
                              //you should see the errno here
    }
}

Errno is 255 when argument --cpu_bind does not exist.
I can use this to do, something like if 255 use dash otherwise use
underscore?

I will finish up the patch by Monday and do some testing under Slurm 16 and
19.

ggouaillardet:
could it be safer/faster to add a MCA parameter ?

For example `mpirun --mca plm_slurm_srun_cli <value> ...`
First, `value` could either be `legacy` or `current`.
Then we could add a third value `auto` that detects which syntax should be used via the `--test-only` option ?

IIRC, the `srun` arguments can also be passed via the environment, so `export SLURM_CPU_BIND=none; srun ...` could be an other possibility (note I did not test that yet)

jdhayes:
Oh, that would be a faster way to explicitly define through MCA parameter.
Perhaps no parameter would assume auto detect through --test-only?

Here is a cleaner version of the --test-only srun param check:
    https://cluster.hpcc.ucr.edu/~jhayes/slurm/19.05.0/srun_check_parms.c

jdhayes:
Hello,
Yesterday, I integrated my srun_check_parms.c
<https://cluster.hpcc.ucr.edu/~jhayes/slurm/19.05.0/srun_check_parms.c>
function
into the openmpi source code.
The new pull request is: Validate slurm params function. #6654
That is about all I can do for now.

PR #6652, opened 2019-05-12
OSHMEM: Add support for shmemx_malloc_with_hint() - v4.0.x

yosefe:
- added multiple segments processing
- added shmemx_malloc_with_hint call + set of hints

(picked from master 94b5e91)

### DIFF
--- a/oshmem/include/pshmemx.h
+++ b/oshmem/include/pshmemx.h
@@ -16,6 +16,11 @@
 extern "C" {
 #endif
 
+/*
+ * Symmetric heap routines
+ */
+OSHMEM_DECLSPEC  void* pshmemx_malloc_with_hint(size_t size, long hint);
+
 
 /*
  * Legacy API
--- a/oshmem/include/shmemx.h
+++ b/oshmem/include/shmemx.h
@@ -18,10 +18,28 @@
 extern "C" {
 #endif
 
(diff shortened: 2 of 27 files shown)


hppritcha:
@yosefe please correct sign off problem with commit 
2b2a413

yosefe:
@hppritcha thanks, fixed

hppritcha:
this was discussed on the devel call today.  the consensus is that we normally do not take this type of addition into a release branch.

gpaulsen:
Howard and I discussed today, and agreed to take this as an experimental extension to OSHMEM, but wanted to relay the following comments:

Typically we don't accept new functionality into a release branch, but our thoughts are as following:
1) This is an extension to OSHMEM, which reduces risk to existing OSHMEM users.
2) This is isolated only to the OSHMEM component, and doesn't impact any Open MPI, reduces risk to majority of users.
3) This passes OSHMEM CI regression, that also reduces risk.

Given it's importance, we understand and will take it.  We don't want this to start a precedence of accepting new functionality on release branches.  We hope to keep the release branches as stable and minimal as possible, which means keeping them to bugfixes only.

Finally, we noticed there isn't much OSHMEM in MTT, it would be nice to have some specific OSHMEM tests in MTT.

PR #6654, opened 2019-05-13
Validate slurm params function.

jdhayes:
I successfully tested this patch under Slurm 16.05.4 and 19.05.0 rc1.

### DIFF
--- a/orte/mca/plm/slurm/plm_slurm_module.c
+++ b/orte/mca/plm/slurm/plm_slurm_module.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2006-2018 Cisco Systems, Inc.  All rights reserved
+ * Copyright (c) 2006-2019 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2007-2015 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2014-2018 Intel, Inc.  All rights reserved.
@@ -272,14 +272,6 @@ static void launch_daemons(int fd, short args, void *cbdata)
         opal_argv_append(&argc, &argv, "--kill-on-bad-exit");
     }
 
-    /* ensure the orteds are not bound to a single processor,
-     * just in case the TaskAffinity option is set by default.
-     * This will *not* release the orteds from any cpu-set
-     * constraint, but will ensure it doesn't get
-     * bound to only one processor
-     */
-    opal_argv_append(&argc, &argv, "--cpu_bind=none");
-
 #if SLURM_CRAY_ENV
     /*
      * If in a SLURM/Cray env. make sure that Cray PMI is not pulled in,
@@ -419,6 +411,23 @@ static void launch_daemons(int fd, short args, void *cbdata)
 
     /* setup environment */
     env = opal_argv_copy(orte_launch_environ);
+
+    /* ensure the orteds are not bound to a single processor,
+     * just in case the TaskAffinity option is set by default.
+     * This will *not* release the orteds from any cpu-set
+     * constraint, but will ensure it doesn't get
+     * bound to only one processor
+     *
+     * NOTE: We used to pass --cpu_bind=none on the command line.  But
+     * SLURM 19 changed this to --cpu-bind.  There is no easy way to
+     * test at run time which of these two parameters is used (see
+     * https://github.com/open-mpi/ompi/pull/6654).  There was
+     * discussion of using --test-only to see which one works, but
+     * --test-only is only effective if you're not already inside a
+     * SLURM allocation.  Instead, set the env var SLURM_CPU_BIND to
+     * "none", which should do the same thing as --cpu*bind=none.
+     */
+    opal_setenv("SLURM_CPU_BIND", "none", true, &env);
 
     if (0 < opal_output_get_verbosity(orte_plm_base_framework.framework_output)) {
         param = opal_argv_join(argv, ' ');


rhc54:
ok to test

@jdhayes We do need you to signoff on the commit - can you please do so?

jdhayes:
OK, appended signoff.

orte/mca/plm/slurm/plm_slurm_module.c
```diff
@@ -165,6 +168,44 @@ static int plm_slurm_launch_job(orte_job_t *jdata)
...
+
+        /* Once child process, srun, has ended check exit code */
+        if (WIFEXITED(status)) {
+            if (255 == WEXITSTATUS(status)) {
+                return(0);
+            }
+        }
+    }
+    /* All other exit codes mean parms are good */
+    return(1);
```
jsquyres: I think the logic isn't quite right here -- e.g., what if `srun` fails due to a signal?

It might be easier to check for the one condition that is good, and all others are failures.

Isn't the only good case where `srun` exits with exit status 0?

orte/mca/plm/slurm/plm_slurm_module.c
```diff
@@ -278,7 +278,7 @@ static void launch_daemons(int fd, short args, void *cbdata)
      * constraint, but will ensure it doesn't get
      * bound to only one processor
      */
-    opal_argv_append(&argc, &argv, "--cpu_bind=none");
+    opal_argv_append(&argc, &argv, "--cpu-bind=none");
```
jsquyres: This commit doesn't have a signoff line.  You can probably delete this commit, anyway, since the next commit makes this commit moot.

jsquyres:
I just did some checking on my ancient SLURM:

```
$ srun --version
slurm 14.03.0
```

Here's what I mean by looking for the one good case:

```sh
# Bad case
$ srun --cpu-bind=none --test-only
srun: unrecognized option '--cpu-bind=none'
Try "srun --help" for more information
$ echo $status
1

# Good case
$ srun --cpu_bind=none --test-only
srun: cluster configuration lacks support for cpu binding
srun: Job 2664107 to start at 2019-05-14T14:08:20 using 16 processors on ivy13
$ echo $status
0
```

jdhayes:
Yeah, I thought about that first.
However, it looks like there is more than 1 good case:

srun --version
slurm 19.05.0-0rc1

srun --cpu-bind=none --test-only; echo $?
allocation failure: Requested node configuration is not available
1

srun -p batch --cpu-bind=none --test-only; echo $?
srun: Job 285 to start at 2019-05-14T14:24:30 using 4 processors on nodes
c[38-39] in partition batch
0

srun --cpu_bind=none --test-only; echo $?
srun: unrecognized option '--cpu_bind=none'
srun: unrecognized option '--cpu_bind=none'
Try "srun --help" for more information
255

Also, it seemed faster to let it fail, since srun would check the params
before it tried to "mock" schedule the job.
So, I opted to check for the exact failure that I knew I could handle, 255
with the bad params.

jsquyres:
My point is that 255 does not happen in my old version of SLURM -- it's only 0 and 1.

How about just trying to execute the full command line (with no `--test-only`), and if it works, great.  If it doesn't, try again with the other CPU bind option.  If it works, great.  If not, error.

jdhayes:
Yes, you are right.
Even 16.05.4 seems to give a exit code of 1, similar to your 14.03.0.

srun --version
slurm 16.05.4

srun --cpu-bind=none --test-only; echo $?
srun: unrecognized option '--cpu-bind=none'
Try "srun --help" for more information
1

Looks like the error codes changed at some point.

jdhayes:
Although the patch would still work since the case where the exit code does
equal 255, (ie. Slurm 19) it does detect the issue and uses the correct
cpu-bind flag.
In other cases, it would default to the already working cpu_bind flag.

orte/mca/plm/slurm/plm_slurm_module.c
```diff
@@ -165,6 +168,44 @@ static int plm_slurm_launch_job(orte_job_t *jdata)
...
+        dup2(fd, 2);
+
+        execvp(argv[0], argv);
+
+        close(fd);
+        exit(errno);
+    }
+    else{
+        /* Wait for child process to end */
+        wait(&status);
```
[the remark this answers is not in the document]
rhc54: Why would you do this? Just do:
```C
rc = system("srun --test-only --cpu_bind=none");
```
There is no reason to do something more complex here, is there?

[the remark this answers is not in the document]
jsquyres: @rhc54 That won't redirect the stdout/stderr.

Regardless, I don't think this `--test-only` approach is a good one -- return values / behavior have changed over releases of SLURM.  See https://github.com/open-mpi/ompi/pull/6654#issuecomment-492416970

jdhayes:
Perhaps your right.
It did not have to be that complex, but I was trying to use the style of
coding that already existed.

jdhayes:
OK, that's fine.
Perhaps setting the SLURM_CPU_BIND environment variable is more
simplistic/homogeneous across Slurm versions.

jsquyres:
That's an interesting idea.  Does `SLURM_CPU_BIND` exist back as far as say my ancient version of SLURM (14.03)?  Is the meaning the same in all versions?  And if so, can we pass the same values in it that we were previously passing in `--cpu_bind`?

orte/mca/plm/slurm/plm_slurm_module.c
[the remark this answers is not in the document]
rhc54: You can certainly add to that cmd to redirect things - I wasn't trying to write out a detailed prescription 😄 
Will join the discussion over on the main thread.

jdhayes:
I will not take credit for it, it was ggouaillardet who first mentioned it.
You can check the man page for srun, it should be there.

rhc54:
I honestly don't understand this discussion thread. Who cares what the error code is? If it doesn't return zero, then it failed and that is all we need to know. We don't have to worry about all these cmd line variations - we are the ones controlling it! Just do the simple "srun --test-only --cpu_bind=none" and be done with it.

I think you guys are going down the perfectionist rathole here.

jdhayes:
There is a case where I get an error code of 1, even though there is no
issue with the params.
I thought it best to check and handle only the case that I know of, and
ignore all other exit codes.
That is why I only change the params if I detect an error code of 255,
which seems to only happen in Slurm 19.
The patch works under Slurm 19 and 16, and I am guessing 14 as well.
If anyone else want to take a crack at this patch, or implement the
environment variable instead, feel free...I am just running out of time to
do it :)

rhc54:
The allocation failure - that is what you are referring to? Why do we care? It failed - nothing we do is going to change that, which means we cannot launch anyway. In this case, both your check tries will fail, and so mpirun might as well error out.

What am I missing here?

jsquyres:
It looks like much of this is moot: `--test-only` does not work when you are inside an existing allocation (e.g., if you're inside an `salloc`, `sbatch`, or `srun`).

jsquyres:
I pushed an amended commit to this PR that simply sets the env variable.  According to the man page, it's a synonym for `--cpu*bind`.

@jdhayes @rhc54 Please review.

jdhayes:
Looks OK to me.
Much simpler! :)

orte/mca/plm/slurm/plm_slurm_module.c
```diff
@@ -277,8 +277,17 @@ static void launch_daemons(int fd, short args, void *cbdata)
...
+     * NOTE: We used to pass --cpu_bind=none on the command line.  But
+     * SLURM 19 changed this to --cpu-bind.  There is no easy way to
+     * test at run time which of these two parameters are used (see
+     * https://github.com/open-mpi/ompi/pull/6654) -- there was
+     * discussion of using --test-only to see which one works, but
+     * --test-only is only effective if you're not already inside a
+     * SLURM allocation.  Instead, set the env var SLURM_CPU_BIND to
+     * "none", which should do the same thing as --cpu*bind.
+    */
+    opal_setenv("SLURM_CPU_BIND", "none", true, &orte_launch_environ);
```
rhc54: You really should move this down to line 422 (pre-PR) and add that envar to the srun "env" environment, not the orte_launch_environ, as that is what we pass to our client procs. I don't _think_ it will hurt anything for the clients, but no reason to take a chance.

jsquyres: Will do!

jsquyres:
bot:mellanox:retest

PR #6658, opened 2019-05-14
usnic: fix Coverity false positives

jsquyres:
Add some Coverity inline notation to tell Coverity that these
functions never return.


jsquyres:
False Mellanox jenkins error.

bot:mellanox:retest

PR #6662, opened 2019-05-15
OSHMEM/MMAP/SYSV: Return ERR_NOT_IMPLEMENTED if segment hint != 0

yosefe:


yosefe:
@hoopoepg pls take a look

PR #6668, opened 2019-05-15
btl/uct: check for support before disabling UCX memory hooks

### LINKED ISSUE #6666: BTL UCT causing problems with UCX memory hooks
From https://www.mail-archive.com/devel@lists.open-mpi.org/msg20895.html and https://www.mail-archive.com/devel@lists.open-mpi.org/msg20904.html:

> The issue below is caused by btl_uct – it mistakenly calls ucm_set_external_event() without checking opal_mem_hooks_support_level().
> This leads UCX to believe that memory hooks would be provided by OMPI, but in fact they are not, so pinned physical pages become out-of-sync with process virtual address.
>
> btl_uct wrong call: https://github.com/open-mpi/ompi/blob/master/opal/mca/btl/uct/btl_uct_component.c#L132
> Correct way: https://github.com/open-mpi/ompi/blob/master/opal/mca/common/ucx/common_ucx.c#L104

FYI @yosefe 

v4.0.x RMs: this feels like a blocker for v4.0.x.  Thoughts?

hjelmn:

Fixes #6666

### DIFF
--- a/opal/mca/btl/uct/btl_uct_component.c
+++ b/opal/mca/btl/uct/btl_uct_component.c
@@ -17,6 +17,7 @@
  * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights reserved.
  * Copyright (c) 2018      Triad National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019      Google, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


jsquyres:
Many thanks, @hjelmn!

jsquyres:
Mellanox Jenkins git clone fail.

bot:mellanox:retest

jsquyres:
bot:mellanox:retest

PR #6669, opened 2019-05-15
Fixed a bug that SPC/MPI_T interface is not able to initialize properly.

yqin:
This patch helps to fix a bug that SPC/MPI_T interface is not able to initialize properly.


yqin:
Ah, need to rebase it.

PR #6670, opened 2019-05-16
btl/uct: add support for UCX 1.6.x

hjelmn:
This commit updates the uct btl to support the v1.6.x release of
UCX. This release breaks API.

### DIFF
--- a/opal/mca/btl/uct/btl_uct_endpoint.c
+++ b/opal/mca/btl/uct/btl_uct_endpoint.c
@@ -4,6 +4,7 @@
  *                         reserved.
  * Copyright (c) 2018      Triad National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019      Google, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


hjelmn:
@jsquyres Tested and working with UCX 1.6.x branch.

opal/mca/btl/uct/btl_uct_endpoint.c
```diff
@@ -101,6 +102,28 @@ static void mca_btl_uct_process_modex (mca_btl_uct_module_t *uct_btl, unsigned c
     }
 }
 
+static inline ucs_status_t mca_btl_uct_ep_create_connected_compat (uct_iface_h iface, uct_device_addr_t *device_addr,
+                                                                   uct_iface_addr_t *iface_addr, uct_ep_h *uct_ep)
+{
+#if UCT_API >= UCT_VERSION(1, 6)
```
hoopoepg: what about check if uct_ep_create_connected is available using autotools/configure? this is standard way to check if some API available.
for example look at constructions AC_CHECK_DECLS() at config/ompi_check_ucx.m4

jsquyres: We talked about this and decided for a generic version test -- which, you're right, is unusual / not the preferred way for `configure`-based tests.

The reason is that the UCT API is not guaranteed to have stability over time.  As I understand it, UCT is an internal API that may change at any time without any warning.  So technically, if we went the "usual" way of `configure` tests, we'd have to test for every single UCT API symbol (including parameters) that is used in this BTL.  

That's why we compromised on the version check, instead: this BTL is known to work with UCX versions X through Y.

jsquyres:
I would still like to see a `configure` test for the version UCX version number.  Specifically, because the UCT API does not have any API stability guarantees, I really do not want to be in a position again where we are today: a new version of UCX causes a default build of Open MPI to fail.

How about this for a compromise:

* BTL UCT `configure.m4` will allow itself to be built by default if it finds UCX version X through Y (I don't know offhand what the lower bound/X value is, but Y should be 1.6).
* If BTL UCT `configure.m4` finds UCX version Z, where Z>Y, then it will not build itself unless `--enable-uct-btl-ignore-ucx-version` (or some CLI option like that) is specified.

This gives us multiple important properties:

1. If a future version of UCX changes the UCT API, it won't cause a default build of Open MPI to break.
1. If a future version of UCX *doesn't* change the UCT API, there's still a path for users to use the UCT BTL.
1. CI and/or MTT testing can force the attempted compilation of the UCT BTL so that we can be notified of compile failures if a future version of UCX changes the UCT API.

How's that sound?

hjelmn:
@jsquyres UCT is not really internal-only. Any layer of UCX may change with a new minor version. This can be the case with UCP as well. That would break pml/ucx.

hjelmn:
The real problem is that UCX doesn't really version their API at all. The versions available are tarball version. I prefer a version check over a configure check because I know when the API changed and I the UCT API is supposedly stabilizing.

gpaulsen:
@jsquyres @hoopoepg Are you comfortable with merging this to master (and I'll then PR it to v4.0.x)?

gpaulsen:
fixes #6640

PR #6672, opened 2019-05-16
v4.0.x: plm_slurm_module: adjust for new SLURM CLI options

jsquyres:
SLURM 19 discontinued the use of --cpu_bind (and changed it to
--cpu-bind).  There's no easy way to test at run time which one is
accepted, so set the environment variable SLURM_CPU_BIND to "none",
which should do the same thing as the srun CLI parameter.

Signed-off-by: Jordan Hayes <jhayes@ucr.edu>
(cherry picked from commit 7dad74032e30259506da7fa582dd8c4351e6e0a1)

FYI @jdhayes

Release managers: I marked this as "NEWS" because this commit fixes the SLURM plm (for `mpirun`-based launching, not native launch/`srun`-based launching) with regard to SLURM 19, because SLURM 19 deleted a CLI option that previous versions of Open MPI were using.  This PR works around the issue by using a SLURM env var instead of the problematic CLI option.  This env var works with really old versions of SLURM (we checked back to SLURM 14) up to SLURM 19.

### DIFF
--- a/orte/mca/plm/slurm/plm_slurm_module.c
+++ b/orte/mca/plm/slurm/plm_slurm_module.c
@@ -9,7 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
- * Copyright (c) 2006-2018 Cisco Systems, Inc.  All rights reserved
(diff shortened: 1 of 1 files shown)


gpaulsen:
Okay to merge.  Just leaving open a while in case someone wants to look at https://github.com/open-mpi/ompi/issues/6684.  Can probably merge anway next week.

hppritcha:
bot:lanl:retest

PR #6675, opened 2019-05-16
COMMON/UCX: init memhooks infra on external hooks only

hoopoepg:
- initialize memory hooks infrastructure only in case
  if external memory hooks are requested

### DIFF
--- a/opal/mca/common/ucx/common_ucx.c
+++ b/opal/mca/common/ucx/common_ucx.c
@@ -16,7 +16,6 @@
 #include "opal/memoryhooks/memory.h"
 
 #include <ucm/api/ucm.h>
-#include <ucs/sys/math.h>
 
 /***********************************************************************/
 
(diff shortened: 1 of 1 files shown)


opal/mca/common/ucx/common_ucx.c
```diff
@@ -91,23 +92,22 @@ OPAL_DECLSPEC void opal_common_ucx_mca_register(void)
...
+    if (opal_common_ucx.opal_mem_hooks) {
+        ret = mca_base_framework_open(&opal_memory_base_framework, 0);
+        if (OPAL_SUCCESS != ret) {
+            /* failed to initialize memory framework - just exit */
+            MCA_COMMON_UCX_VERBOSE(1, "failed to initialize memory base framework: %d, "
+                                      "memory hooks will not be used", ret);
+            return;
+        }
+
+        if (ucs_test_all_flags(opal_mem_hooks_support_level(),
```
yosefe: i think better avoid using ucs stuff in here, to avoid any compatibility issues in the future

hoopoepg: ok, removed

PR #6676, opened 2019-05-17
oshmem: remove useless reference to orte header

ggouaillardet:


ggouaillardet:
@yosefe FYI most `ompi` and `oshmem` code do not need `orte` any more (not to mention it might be soon somehow replaced with `prrte`)

PR #6680, opened 2019-05-17
PML/UCX: disable PML UCX if MT is requested but not supported

hoopoepg:
- in case if multithreading requested but not supported
  disable PML UCX


hoopoepg:
@yosefe port it into v4.0?

yosefe:
@hoopoepg yes

PR #6696, opened 2019-05-20
btl/uct: check for support before disabling UCX memory hooks

gpaulsen:
v4.0.x version of PR #6668

(cherry picked from commit 3e1dd362411f1da5564d3402f65e9b3b74f50759)


gpaulsen:
I added the WIP-DNM because:
In today's call it sounded like there may be some additional commits going into master around ucx configury.  We can either add those commits here, or just PR this to v4.0.x and then PR the additional ones as they come in (my vote).

AboorvaDevarajan:

gpaulsen:
@hppritcha and I agree to pull this in now, and then pull in additional configury later.

PR #6699, opened 2019-05-22
java: Fix compilation error in allToAllw using Java arrays

t-kurita:
Java bindings in Open MPI support Java arrays and direct buffers
as buffers. All non-blocking methods must use direct buffers and
only blocking methods can choose between Java arrays and
direct buffers.
Though Comm.allToAllw() is a blocking method, Java applications
using Java arrays as buffers get compilation errors.
This fix enables using Java arrays in Comm.allToAllw().

Signed-off-by: Kurita, Takehiro <fj6370fp@aa.jp.fujitsu.com>

### DIFF
--- a/ompi/mpi/java/c/mpiJava.h
+++ b/ompi/mpi/java/c/mpiJava.h
@@ -11,6 +11,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019       FUJITSU LIMITED.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -94,6 +95,15 @@ void ompi_java_getReadPtrv(
         jobject buf, jboolean db, int off, int *counts, int *displs,
         int size, int rank, MPI_Datatype type, int baseType);
 
+/* Gets a buffer pointer for reading, but it
+ * 'size' is the number of processes.
+ * if rank == -1 it copies all data from Java.
+ * if rank != -1 it only copies from Java the rank data. */
+void ompi_java_getReadPtrw(
+        void **ptr, ompi_java_buffer_t **item, JNIEnv *env,
+        jobject buf, jboolean db, int *offs, int *counts, int *displs,
+        int size, int rank, MPI_Datatype *types, int *baseTypes);
+
 /* Releases a buffer used for reading. */
 void ompi_java_releaseReadPtr(
         void *ptr, ompi_java_buffer_t *item, jobject buf, jboolean db);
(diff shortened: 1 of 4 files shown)


kawashima-fj:
ok to test

@t-kurita You changed types of arguments of `allToAllw` public method (`Buffer` -> `Object`). Probably this change causes an ABI break. In other words, if a user compile a Java program using an old `mpi.jar` and update the `mpi.jar`, the user `.class` file cannot be run using the new `mpi.jar` unless the user recompile the Java program. For the master branch, this change may be accepted. However, for release branches, it cannot be accepted.

We have two options.

- Accept this ABI break in the master branch. Don't merge this change to release branches. 
  This change (accepting Java arrays in `allToAllw`) is a new feature in a sense. Therefore we don't necessarily need merge it to relase branches.
- Keep the old method by overloading.

I think the former is suffient. @jsquyres do you have a comment?

kawashima-fj:
To OMPI core developers (especially @jsquyres):
Is it OK to break ABI of Java bindings in the master branch?

This PR fixes a bug that a program which uses a Java array in the `sendBuf` or `recvBuf` argument of the `Comm.allToAllw` method cannot be compiled. This PR changes the type of these arguments to a more generic one (`Object`). Though source code compatibility is retained, binary compatibility is not retained (see the comment above). Adding another method (method overloading) can retain binary compatibility. However, we must keep the old method forever to retain compatibility. So I don't think it's a good idea for this case.

I don't think this PR should be cherry-picked to release branches because this bug is a minor issue.

jsquyres:
@kawashima-fj I agree with your analysis: ABI break is ok on master, but not on release branches.

I'm not clear on two points, though:

1. Do we know for 100% sure that this is an ABI break?  I.e., does Java use RTTI to select the right method at run time, or is the method chosen at compile/link time?  I do not know enough about how Java works under the covers to know the answer to this question.  It might be worth testing this specific case (i.e., compile a test code against the original `.jar`, then run against the new `.jar`).
1. I'm not clear on the exact scope of this error: is it possible for user apps to compile against it at all?  I.e., it is a compile error, or is it a run-time error?  If user apps genuinely can't compile against this old API, then I think we're ok breaking ABI (because it was never possible to compile against that API, anyways).

kawashima-fj:
@jsquyres 

**To your question 1**:

The method is chosen at compile time. The signature of the method (including argument types) is encoded into the caller's `.class` file at compile time. If a method of the same signature does not exist (i.e. disappeared) in the callee's `.class` or `.jar` at run-time, `NoSuchMethodError` occurs.

I checked it by:

1. compile a `Foo` class with a `foo(Buffer)` method
2. compile a `Bar` class which calls `Foo.foo(Buffer)`
3. run the `Bar` -> success
4. recompile the `Foo` class with changing `Buffer` to `Object`
5. run the `Bar` -> `Exception in thread "main" java.lang.NoSuchMethodError: Foo.foo(Ljava/nio/Buffer;)Ljava/lang/Object; at Bar.main(Bar.java:7)`

**To your question 2**:

Before this change:

call `allToAllw(DirectBuffer)`: OK (should be OK)
call `allToAllw(int[])`: compile error (should be OK; other collectives accept it)

After this change (with recompiling OMPI but *without* recompiling user app):

call `allToAllw(DirectBuffer)`: run-time error (**ABI break!**)
call `allToAllw(int[])`: OK (bug fixed)

After this change (with recompiling OMPI and *with* recompiling user app):

call `allToAllw(DirectBuffer)`: OK
call `allToAllw(int[])`: OK (bug fixed)

Is it your expected answer?

@t-kurita Thanks for your contribution. If my answer is incorrect, please let us know.

Issue #6703, opened 2019-05-23
division by zero problem for empty fileviews


edgargabriel:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

v4.0.x, v3.1.x, v.3.0.x

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

When using an empty fileview, a division by zero bug can occur in ompio. Not entirely sure why the problem did not show up previously, but some recent changes trigger that bug in one of our tests. A fix has been already applied to master, and PRs fixing the same issue on the v4.0.x, v3.1.x, and maybe v3.0.x will come shortly.

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ mpirun -np 2 ./hello_world
```

hppritcha:
closing as fix has been patched back to all release branches.

PR #6706, opened 2019-05-23
common/ompio: fix division by zero problem with empty fview

edgargabriel:
When using an empty fileview, a division by zero bug can occur in ompio. Not entirely sure why the problem did not show up previously, but some recent changes trigger that bug in one of our tests.

This pr is part of a fix applied in commit f6b3a0a

Fixes Issue #6703


edgargabriel:
the failure has nothing to do with this pr

bot:retest

edgargabriel:
bot:retest

Issue #6691, opened 2019-05-20
mpirun run hostname cmd unexpected failed


zrss:
Thank you for taking the time to submit an issue!

## Background information

run openmpi in a kubernetes cluster

```
mpirun -np 256 --bind-to socket -mca pml ob1 -mca btl ^openib --mca oob_base_verbose 100 hostname 2>&1 | tee server_out.txt
```

hostfile

```
jobca1e368c-worker-0-2yi-21-tcmhw slots=8
jobca1e368c-worker-1-kf5-562-7q2l4 slots=8
jobca1e368c-worker-10-ytx-394-tjndh slots=8
jobca1e368c-worker-11-4lo-879-tvd2g slots=8
jobca1e368c-worker-12-am0-360-5bf7n slots=8
jobca1e368c-worker-13-14r-69-djjn6 slots=8
jobca1e368c-worker-14-hij-487-rwll5 slots=8
jobca1e368c-worker-15-bl2-479-ldw6c slots=8
jobca1e368c-worker-16-g5l-33-2qcbg slots=8
jobca1e368c-worker-17-m4z-333-g9l5r slots=8
jobca1e368c-worker-18-q0q-773-n7plw slots=8
jobca1e368c-worker-19-ti7-822-n5zf7 slots=8
jobca1e368c-worker-2-rc1-485-rst7w slots=8
jobca1e368c-worker-20-n9s-551-pn5vt slots=8
jobca1e368c-worker-21-013-239-v57fm slots=8
jobca1e368c-worker-22-prf-680-jbwt9 slots=8
jobca1e368c-worker-23-gv6-714-h8xlp slots=8
jobca1e368c-worker-24-4mb-421-47xbg slots=8
jobca1e368c-worker-25-pre-992-kbnkt slots=8
jobca1e368c-worker-26-fvk-685-2tlht slots=8
jobca1e368c-worker-27-5v7-407-dmt5w slots=8
jobca1e368c-worker-28-dpb-517-tkkkq slots=8
jobca1e368c-worker-29-mvh-486-2r7lz slots=8
jobca1e368c-worker-3-ivv-417-fblnz slots=8
jobca1e368c-worker-30-25h-242-klwh6 slots=8
jobca1e368c-worker-31-17k-516-mbr9b slots=8
jobca1e368c-worker-4-9uv-292-8htt5 slots=8
jobca1e368c-worker-5-lxw-492-w7927 slots=8
[...]

PR #6713, opened 2019-05-30
orte/orted: work around the unexpected change of parent vpid in node_regex_report

zrss:
Fixes: #6691 

keep the vpid of parent in node_regex_report

Signed-off-by: zhesihuang <huangzhesi@gmail.com>

### DIFF
--- a/orte/orted/orted_main.c
+++ b/orte/orted/orted_main.c
@@ -1161,6 +1161,7 @@ static void node_regex_report(int status, orte_process_name_t* sender,
                               orte_rml_tag_t tag, void *cbdata) {
     int rc;
     bool * active = (bool *)cbdata;
+    opal_vpid_t parent_vpid;
 
     /* extract the node info if needed, and update the routing tree */
     if (ORTE_SUCCESS != (rc = orte_util_decode_nidmap(buffer))) {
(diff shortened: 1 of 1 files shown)


zrss:
@Ralph Castain , hi Castain, could u have a look ? i am a fresh man in openmpi repo :smile:, maybe someone knows these code well, and have a better way to solve it

and this problem [6691](https://github.com/open-mpi/ompi/issues/6691#issuecomment-497245597) i met in openmpi 4.0.0 and 4.0.1, but master branch i didn't test it yet

zrss:
also ping @jsquyres to have a look

rhc54:
This isn't going to work as the routed component may now be pointed to the wrong place. The first question is: how are you winding up using the debrujin routed component? The default is the radix component - debrujin is way down the priority list. Do you have some MCA parameter set that is directing you to use it?

rhc54:
I see what's going on - we cycle across all active routed components and debrujin does mess with the parent value. I don't know how to fix that component (it may not be possible to redefine the algo so it works around the provided parent), so the best "fix" really is to simply disable/remove it. You can disable it by just putting `routed = ^debrujin` in the default MCA parameter file - you can disable it by just removing the routed/debrujin directory completely (would require you to re-run autogen.pl prior to reconfiguring) or dropping an empty `.opal_ignore` file in that directory.

zrss:
thx for the reply

> we cycle across all active routed components and debrujin does mess with the parent value

yes, it will change the origin parent when we update the route plan (`debrujin`, `binomial`'s parent is diff then `radix`, generally saying `radix` leads to 0 vpid parent, as the default radix is 64, and i only have 32 hosts)  ...

currently, i can specific the routed mod (`radix`, `binomial`, `debruijn`, `direct`) to work around it, i.e.

```
mpirun -np 256 --bind-to socket --mca pml ob1 --mca oob tcp --mca plm rsh --mca btl ^openib --mca routed radix hostname
```

or disable the tree spawn

```
mpirun -np 256 --bind-to socket --mca pml ob1 --mca oob tcp --mca plm rsh --mca btl ^openib --mca plm_rsh_no_tree_spawn true hostname
```

can we assign this question to a colleague who is familiar with this part of the code to fix it ? we just leave a hint here

as `horovod` (deep learning trainning framework) becomes quite popular today, and it is bootstrapped by `mpirun`, there maybe some persons will hit this problem

PR #6716, opened 2019-05-30
Fix tree spawn at scale

rhc54:
The debrujin component is using an algorithm that doesn't respect the
previously assigned parent ID. This causes the other components to have
their routing trees broken whenever debrujin updates routes. This
happens whenever more than 256 nodes are involved, thus breaking tree
spawn for sizes >= 256

Thanks to @zrss for the report and diagnosis!


bwbarrett:
We talked about this change (and related PRs on the v3.1.x and v4.0.x branches) on the call today and decided that removing the component is the right path forward.  While we try very hard to avoid removing components in the middle of a release series, in this case we are going to do so.  The component never worked properly in this release series, because of the stateful component selection and the bug in component selection logic for this component, and because of the cost of a proper fix, this is the right choice.

Issue #6712, opened 2019-05-29
 int32_t versions of shmem_wait_until() shmem_test() expect 64-bit values


sam6258:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

I have reproduced this issue in OMPI master and v4.0.x.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Built from source.

### Please describe the system on which you are running

* Operating system/version: RHEL 7.6
* Computer hardware: POWER9 ppc64le
* Network type: N/A (Issue reproducible on a single node)

-----------------------------

## Details of the problem

Unit tests (`cxx_test_shmem_wait_until.cpp` and `cxx_test_shmem_test.cpp`) from this bucket are failing: https://github.com/Sandia-OpenSHMEM/SOS/tree/master/test/unit

I have a proposed fix coming and will link the PR to this issue.

```> oshrun -n 2 cxx_test_shmem_wait_until
PE 1 received incorrect value with TEST_SHMEM_WAIT_UNTIL(int32_t)
PE 1 received incorrect value with TEST_SHMEM_WAIT_UNTIL(uint32_t)
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
oshrun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[54404,1],1]
  Exit code:    1
--------------------------------------------------------------------------
```

artpol84:
Thanks, @sam6258.
Looking forward to the fix.

PR #6724, opened 2019-05-30
shmem/c: Fix shmem type for calls to shmem_test and shmem_wait_until …

sam6258:
…with [u]int32_t and [u]int64_t

Signed-off-by: Scott Miller <scott.miller1@ibm.com>

This PR is intended to fix https://github.com/open-mpi/ompi/issues/6712. The invalid size of shmem type is causing comparisons in `shmem_test` and `shmem_wait_until` to fail. The sizes for `uint32_t` and `int32_t` types do not match the size of `SHMEM_LLONG`, so we need to use the respective `SHMEM_INT32_T` type.

This also fixes up the case for `SHMEM_INT64_T` even though it is the same size as `SHMEM_LLONG` on the architecture I used to test.

### DIFF
--- a/oshmem/shmem/c/shmem_wait.c
+++ b/oshmem/shmem/c/shmem_wait.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2013      Mellanox Technologies, Inc.
  *                         All rights reserved.
+ * Copyright (c) 2019      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


artpol84:
Looks OK to me,
@hoopoepg please double check.

sam6258:
CI failure looks unrelated to PR. Anyone that can comment on that?

sam6258:
bot:retest

PR #6735, opened 2019-06-04
Fixed orted:  atoi() cause corruption (out)

perrynzhou:
### fixed on v4.0.x
- please review code

### DIFF
--- a/orte/mca/regx/base/regx_base_default_fns.c
+++ b/orte/mca/regx/base/regx_base_default_fns.c
@@ -1056,7 +1056,7 @@ static int regex_parse_node_range(char *base, char *range, int num_digits, char 
     for (found = false, i = 0; i < len; ++i) {
         if (isdigit((int) range[i])) {
             if (!found) {
-                start = atoi(range + i);
+		start = strtol(range + i, NULL, 10);
                 found = true;
                 break;
             }
--- a/orte/mca/regx/base/regx_base_default_fns.c
+++ b/orte/mca/regx/base/regx_base_default_fns.c
@@ -1056,7 +1056,7 @@ static int regex_parse_node_range(char *base, char *range, int num_digits, char 
     for (found = false, i = 0; i < len; ++i) {
         if (isdigit((int) range[i])) {
             if (!found) {
-		start = strtol(range + i, NULL, 10);
+                start = strtol(range + i, NULL, 10);
                 found = true;
                 break;
             }


orte/mca/regx/base/regx_base_default_fns.c
```diff
@@ -1056,7 +1056,7 @@ static int regex_parse_node_range(char *base, char *range, int num_digits, char
     for (found = false, i = 0; i < len; ++i) {
         if (isdigit((int) range[i])) {
             if (!found) {
-                start = atoi(range + i);
+                start = strtol(range + i, NULL, 10);
```
rhc54: Isn't this code block simply wrong?? range is a pointer to a character string, so why isn't this:
```
start = strtol(range[i], NULL, 10)
```
???
If that were corrected, then the original line would be just fine and behave as expected, I believe.

perrynzhou: @rhc54  that will case convert interger overflow.case memory problem

rhc54: I'm afraid I don't understand your answer. Are you saying that you believe `strtol(range + i,...` is correct?  Could you provide an example showing how that statement can yield the correct answer?

Or are you saying that `atoi(range[i])` will still fail? Could you provide a concrete example showing how that happens? Are you saying that "range" contains some characters that `atoi` fails to realize are not digits and attempts to convert?

perrynzhou: @rhc54  gdb info
```
Thread 1 "orted" hit Breakpoint 9, regex_parse_node_range (base=0x710660 "pery-", range=0x710669 "4254712551", num_digits=14, 
    suffix=0x714340 "-worker-1", names=0x7fffffff9a48) at base/regx_base_default_fns.c:1037
1037    {
(gdb) 
1044        if (NULL == base || NULL == range) {
(gdb) 
1048        len = strlen(range);
(gdb) 
1049        base_len = strlen(base);
(gdb) 
1052        start = end = 0;
(gdb) 
1056        for (found = false, i = 0; i < len; ++i) {
(gdb) 
1057            if (isdigit((int) range[i])) {
(gdb) list
1052        start = end = 0;
1053
1054        /* Look for the beginning of the first number */
1055
1056        for (found = false, i = 0; i < len; ++i) {
1057            if (isdigit((int) range[i])) {
1058                if (!found) {
1059                    start = atoi(range + i);  // integer overflow case  memory problem
1060                    found = true;
1061                    break;
(gdb) p range
$121 = 0x710669 "4254712551" 
(gdb) p i
$122 = 0
```

rhc54: I see - so this is yet another corner-case caused by a strange hostname combined with a problem in `atoi` (not protecting itself against an input larger than int). The logic in this code block still bothers me (e.g., the check for `!found` does nothing as we break when `found` gets set to true anyway), but that is for another day.

Just to be clear: this patch has nothing to do with a "double free" in orted - it's corruption due to a bug in `atoi`. Yes? Sorry for all the questions - I'm just trying to understand the proposed solution (which is simple enough) in the context of the reported problem (double free).

perrynzhou: @rhc54  yes,what need I to do next?can push pr to v4.0.x?

ggouaillardet: @rhc54 `range+i` is correct (read `range[i]` is not). The issue is indeed `atoi()` not protecting itself against an input larger than `int`. Strictly speaking, I guess we could build a hostname that has `strtol()` overflows too, but from a pragmatic point of view, this patch looks good enough to me.

I think the overflow could cause different hosts ending up having the same name, and could then indirectly triggers a double free().

perrynzhou: > @rhc54 `range+i` is correct (read `range[i]` is not). The issue is indeed `atoi()` not protecting itself against an input larger than `int`. Strictly speaking, I guess we could build a hostname that has `strtol()` overflows too, but from a pragmatic point of view, this patch looks good enough to me.
> 
> I think the overflow could cause different hosts ending up having the same name, and could then indirectly triggers a double free().

@ggouaillardet  you can control string range,judge  length of range string and control range string.because many scene used openmpi was in distributed computing or distributed traning in kubernetes.

ggouaillardet: Good point, I will take a crack at it.

perrynzhou: > Good point, I will take a crack at it.

@ggouaillardet  what should i do ? I send PR,can you review it ? go on workflow of merge  to v4.0.x

Issue #6740, opened 2019-06-05
OSC rdma: deadlock if memory registration fails


devreal:
We ran into a situation in which Open MPI got stuck in `MPI_Win_allocate` while allocating a series of windows on a ConnectX-5 cluster, like in the following example:

```C
for (i = 0; i < n; ++i) {
  MPI_Win_allocate(1<<30, 1, MPI_INFO_NULL, MPI_COMM_WORLD, &bases[i], &wins[i]);
}
```

On a 64GB with two ranks, we expect all allocations to succeed up to `n=30` (60GB in total).

Running with two ranks, the processes hang with the stack trace of process 0 looking like:

```
#0  mca_btl_vader_component_progress ()
    at btl_vader_component.c:701
#1  0x00002aaaab3ff3ba in opal_progress ()
    at opal_progress.c:228
#2  0x00002aaaaab301b7 in ompi_request_wait_completion (req=0x7be688)
    at request.h:413
#3  0x00002aaaaab301f5 in ompi_request_default_wait (req_ptr=0x7fffffffb890, 
    status=0x7fffffffb870)
    at req_wait.c:42
#4  0x00002aaaaabd6f60 in ompi_coll_base_sendrecv_zero (dest=1, stag=-16, 
    source=1, rtag=-16, comm=0x9b7cd0)
    at coll_base_barrier.c:64
#5  0x00002aaaaabd7620 in ompi_coll_base_barrier_intra_two_procs (
    comm=0x9b7cd0, module=0x9b9410)
    at coll_base_barrier.c:300
#6  0x00002aaaaf7fb11f in ompi_coll_tuned_barrier_intra_dec_fixed (
    comm=0x9b7cd0, module=0x9b9410)
    at coll_tuned_decision_fixed.c:196
#7  0x00002aaaaf8718b4 in ompi_osc_rdma_free (win=0x9b5230)
    at osc_rdma_module.c:61
#8  0x00002aaaaf88272a in ompi_osc_rdma_component_select (win=0x9b5230, 
    base=0x7fffffffbac0, size=1073741824, disp_unit=1, 
    comm=0x404140 <ompi_mpi_comm_world>, info=0x404340 <ompi_mpi_info_null>, 
    flavor=2, model=0x7fffffffbacc)
    at osc_rdma_component.c:1255
#9  0x00002aaaaabf067a in ompi_osc_base_select (win=0x9b5230, 
    base=0x7fffffffbac0, size=1073741824, disp_unit=1, 
[...]

PR #6746, opened 2019-06-07
OSC rdma win allocate: propagate errors to avoid deadlocks

devreal:
This PR changes the shared memory allocation in the rdma osc to synchronize error codes among the ranks in the shared memory communicator to prevent rank 0 from aborting without notifying the other ranks. The changes add two additional allreduce calls on the shared memory communicator and one on the window communicator and replaces two barrier calls with the error code synchronization.

There are still some cases in which ranks abort without notification, most notably if memory allocation fails. I'm not sure what the policy is in Open MPI for this type of error. Catching those would require some additional collective communication, both on the shared memory communicator and the communicator of the window.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -523,6 +523,19 @@ struct _local_data {
     size_t size;
 };
 
+static int synchronize_errorcode(int errorcode, ompi_communicator_t *comm)
+{
+    int ret;
+    int err = errorcode;
+    /* This assumes that error codes are negative integers */
+    ret = comm->c_coll->coll_allreduce (MPI_IN_PLACE, &err, 1, MPI_INT, MPI_MIN,
+                                        comm, comm->c_coll->coll_allreduce_module);
+    if (OPAL_UNLIKELY (OMPI_SUCCESS != ret)) {
+        err = ret;
+    }
+    return err;
+}
+
 static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, size_t size)
 {
     ompi_communicator_t *shared_comm;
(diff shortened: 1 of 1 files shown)


gpaulsen:
ok to test.

hjelmn:
Well, in my defense I never saw a partial failure :). Looking at this now...

hjelmn:
@devreal Generally we offer no real guarantees when we run out of memory. We do *try* to handles OOM condition gracefully.

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -593,28 +606,35 @@ static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, s
...
         }
 
         module->segment_base = opal_shmem_segment_attach (&module->seg_ds);
         if (NULL == module->segment_base) {
             OSC_RDMA_VERBOSE(MCA_BASE_VERBOSE_ERROR, "failed to attach to the shared memory segment");
             ret = OPAL_ERROR;
+        }
+
+        ret = synchronize_errorcode(ret, shared_comm);
+        if (OPAL_UNLIKELY(OMPI_SUCCESS != ret)) {
```
hjelmn: I am curious. Have you run into a scenario where some of the local ranks fail to attach?

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -1235,6 +1248,9 @@ static int ompi_osc_rdma_component_select (struct ompi_win_t *win, void **base,
 
     /* fill in our part */
     ret = allocate_state_shared (module, base, size);
+
+    /* notify all others if something went wrong */
+    ret = synchronize_errorcode(ret, module->comm);
```
hjelmn: This I see as the most likely scenario. Run into problems on one/some nodes but not all nodes.

ompi/mca/osc/rdma/osc_rdma_component.c
devreal: Indeed we have seen this happen when trying to allocate windows beyond 50GB accumulated memory on the node the registration fails (on a ConnectX-3 system). I should have linked the ticket: #6740 

Interestingly, the problem only occurs with the 3.x releases but not on the 4.x releases or master. Still the issue of handling errors exists in all branches.

Issue #6758, opened 2019-06-13
FPE due to int rounding in cost_calc of common_ompio_aggregators.c


haraldkl:
### Please describe the system on which you are running
* Operating system/version: Arch Linux
* Computer hardware: Intel x86
* Network type: Single node

-----------------------------

## Details of the problem
When compiling our application with floating point error checking activated, the resulting executable linked against OpenMPI fails with an floating point exception when calling MPI_File_set_view.
Debugging this revealed that this is due to the cost_calc function in ompi/mca/common/ompio/common_ompio_aggregators.c.
There a real division is assigned to an intermediate variable c, which is then again assigned to the real variable n_as. If the division happens to be smaller than 1, this can become 0. As n_as is subsequently used in another division, we end up with a division by 0 in this case.

e79ecd4 fixes the issue for me, by avoiding the intermediate assignment to an integer variable. Please let me know whether that would be suitable for a pull-request, or if it needs some other kind of fixing.

edgargabriel:
@haraldkl please feel free to file a pr with this patch, it looks good to me. If possible, having a reproducer for the bug would also be helpful, since I could add it to our regular testsuite to avoid that we break it later accidentally.

jsquyres:
@haraldkl @edgargabriel Can you add "target" labels to this issue so that we know what release branches are affected?  Thanks.

edgargabriel:
@jsquyres sure, will do. This part of the code is only in master and 4.0.x

haraldkl:
Here is a (probably not minimal Fortran) program to reproduce the problem:
[crash_mpiio.txt](https://github.com/open-mpi/ompi/files/3287098/crash_mpiio.txt)

To reproduce: compile with floating point exceptions active and run with 5 processes (or some other count that does not divide 16 and is larger than 4), i.e. with gfortran:
`$ mpif90 -ffpe-trap=invalid,zero,overflow crash_mpiio.f90`
`$ mpiexec -np 5 ./a.out`

hppritcha:
#6774 fixes this in 4.0.x

PR #6759, opened 2019-06-13
Suggestion to fix division by zero in file view.

haraldkl:
This is the fix mentioned for #6758.

In common_ompi_aggregators calc_cost routine:
do not cast the real division to an int intermediately.
This patch removes the obsolete int variable c and assigns
the result of the P_a/P_x division directly to n_as.

With the intermediate int c variable, n_as gets 0 if P_a < P_x,
resulting in a division by 0 when computing n_s.


jsquyres:
@haraldkl Can you please sign off your commit, in agreement with the [Contributor's Declaration](https://github.com/open-mpi/ompi/wiki/Administrative-rules#contributors-declaration)?

jsquyres:
ok to test

haraldkl:
@jsquyres I pushed the amended commit message now with my sign-off. Thanks.

jsquyres:
False cray CI failure

bot:lanl:retest

hppritcha:
another case of ws not being cleaned up
bot:ompi:retest

hppritcha:
bot:ompi:retest

hppritcha:
@edgargabriel can you take a look at this and see if we need to PR back to other branches?

edgargabriel:
@hppritcha this part of the code exists in the 4.0.x branch, so we will have to file a pr for that. It is not present however in the 3.1.x or 3.0.x series

PR #6769, opened 2019-06-21
OSC rdma win allocate: propagate errors to avoid deadlocks (v3.1.x)

devreal:
This PR changes the shared memory allocation in the rdma osc to synchronize error codes among the ranks in the shared memory communicator to prevent rank 0 from aborting without notifying the other ranks. The changes add two additional allreduce calls on the shared memory communicator and one on the window communicator and replaces two barrier calls with the error code synchronization.

Cherry-pick of #6746 to v3.1.x

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -550,6 +550,19 @@ struct _local_data {
     size_t size;
 };
 
+static int synchronize_errorcode(int errorcode, ompi_communicator_t *comm)
+{
+    int ret;
+    int err = errorcode;
+    /* This assumes that error codes are negative integers */
+    ret = comm->c_coll->coll_allreduce (MPI_IN_PLACE, &err, 1, MPI_INT, MPI_MIN,
+                                        comm, comm->c_coll->coll_allreduce_module);
+    if (OPAL_UNLIKELY (OMPI_SUCCESS != ret)) {
+        err = ret;
+    }
+    return err;
+}
+
 static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, size_t size)
 {
     ompi_communicator_t *shared_comm;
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

jsquyres:
Looks like a legit CI fail:

```
15:45:06 osc_rdma_component.c: In function ‘allocate_state_shared’:
15:45:06 osc_rdma_component.c:631:13: error: implicit declaration of function ‘opal_asprintf’ [-Werror=implicit-function-declaration]
15:45:06              ret = opal_asprintf (&data_file, "%s" OPAL_PATH_SEP "osc_rdma.%s.%x.%d",
15:45:06              ^
```

ibm-ompi:

 Gist: https://gist.github.com/e06b509cbeda012ece3e0d1ccaa17054

ibm-ompi:

 Gist: https://gist.github.com/3f231118d78251cf6f311832e9ac3615

bosilca:
For the general case aborting is a good approach. All the collectives on the 2 communicators can be replaced by a single collective upon global completion of the allocation function with a MAX_LOC operation.

devreal:
@jsquyres Dang! That was the second time I got bitten by the missing backport of `opal_asprintf` to 3.1.x when cherry-picking. Should be more careful next time.

@bosilca

> All the collectives on the 2 communicators can be replaced by a single collective upon global completion of the allocation function with a MAX_LOC operation.

That is essentially the state so far before this PR, which leads to the deadlock situation I'm trying to fix. In some cases, the process detecting the error could continue to contribute to the collectives all other processes execute and pretend nothing happened before signalling the error at the very end (for example if the registration fails). In some other cases that is not possible, i.e., if the shared memory allocation fails. Looking at the code again, we can probably get rid of one or two synchronization steps by slightly reordering the code and continuing execution despite the error. I will give it a try and report back.

jsquyres:
> @jsquyres Dang! That was the second time I got bitten by the missing backport of `opal_asprintf` to 3.1.x when cherry-picking. Should be more careful next time.

Not to worry; this is exactly why we have CI.  😄

hppritcha:
bot:ompi:retest

jsquyres:
bot:mellanox:retest

devreal:
@bosilca going through the code I found that we can piggy-back the error signalling onto the broadcast (shmem comm) and the allgather (global comm) and reorder to get down to one place where the error code needs to be synchronized explicitely. Compared to the original code, this shaves off one shmem barrier and one global allreduce. On the downside, it adds some complexity to the control flow and makes the error handling a little less obvious.

I'm happy to post a PR for this, preferably as a follow-up to this PR (which has already been merged into master). The diff on master can be found here: https://github.com/open-mpi/ompi/compare/master...devreal:osc_winalloc_err?diff=split&w=1

bosilca:
@devreal I like your proposed approach. With the correct amount of comments it will be a good addition to the code. I'll be happy to review.

hppritcha:
bot:mellanox:retest

hppritcha:
give mellanox another try
bot:mellanox:retest

hppritcha:
bot:mellanox:retest

PR #6782, opened 2019-06-28
mpi/c: fix param checks in [I]Neighbor_alltoall{v,w}

ggouaillardet:
do not check some input parameters when an {in,out}degree is zero

Thanks Junchao Zhang for analyzing and reporting this issue.


hjelmn:
Should probably add a unit test for this case to ensure we don't introduce a regression for this case.

PR #6783, opened 2019-06-28
Prevent EPIPE on OSX.

abouteiller:
Some BSD flavors generate EPIPE when we write to a disconnected peer. We need the prevent this signal to be able to trap socket shutdown and cleanly release the endpoint.


abouteiller:
bot:retest

abouteiller:
@ggouaillardet would you mind making a review?

PR #6784, opened 2019-06-28
Address a race condition in libevent select.

abouteiller:
In some rare instances (race), when an fd is closed, the libevent will enter an infinite loop.
If we do not remove a bad fd from the select list we keep getting the same error from select, and we stop doing any progress on the communication side. Thus, we forcefully disable all bad fd as soon
as select fails, and we are back in track, progress ensure and everything seems to work as expected (no leftover events in the event base).

### DIFF
--- a/opal/mca/event/libevent2022/libevent/select.c
+++ b/opal/mca/event/libevent2022/libevent/select.c
@@ -42,6 +42,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <errno.h>
+#include <fcntl.h>
 
 #include "event-internal.h"
 #include "evsignal-internal.h"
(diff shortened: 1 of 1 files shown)


rhc54:
Would you please push this upstream? It affects a lot of us who use libevent but don't embed it.

jsquyres:
Any reason this should not be merged?

awlauria:
This looks like a good change. Not merged with upstream libevent though - but no complaints over it.

Adding it to the 5.0 milestone for tracking.

awlauria:
5.0 is slated to branch on April 30th. Since this is a bug fix, it doesn't _need_ to get in by then, but the sooner the better. Thanks!

awlauria:
I ping'd upstream libevent on it - if they sign off I think we should pull it in.

awlauria:
Based on the comments in libevent/libevent#911 this should be good to go - they just have a minor nit on syntax and want a unit test before they approve.

awlauria:
@gpaulsen do you agree this is good to go?

awlauria:
I forgot we don't patch other components anymore - my bad. Revert here:

https://github.com/open-mpi/ompi/pull/7940

PR #6790, opened 2019-07-03
fortran/mpif-h: correctly handle array_of_errcodes in ompi_comm_spawn…

ggouaillardet:
…[_multiple]_f

Since array_of_errcodes is only allocated when MPI_ERRCODES_IGNORE is not used,
it should not be cleaned when MPI_ERRCODES_IGNORE is used.

Correctly allocate array_of_errcodes with the right size (e.g. maxprocs).

Thanks Gyevi-Nagy Laszlo for reporting this issue.

### DIFF
--- a/ompi/mpi/fortran/mpif-h/comm_spawn_f.c
+++ b/ompi/mpi/fortran/mpif-h/comm_spawn_f.c
@@ -10,8 +10,8 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2010-2018 Cisco Systems, Inc.  All rights reserved
- * Copyright (c) 2015-2017 Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2015-2019 Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -76,7 +76,7 @@ void ompi_comm_spawn_f(char *command, char *argv, MPI_Fint *maxprocs,
 {
     MPI_Comm c_comm, c_new_comm;
     MPI_Info c_info;
-    int size, c_ierr;
+    int c_ierr;
     int *c_errs;
     char **c_argv;
     char *c_command;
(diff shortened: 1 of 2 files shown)


jsquyres:
False failure at cray (git pull fail).

bot:lanl:retest

ompi/mpi/fortran/mpif-h/comm_spawn_f.c
```diff
@@ -119,9 +118,7 @@ void ompi_comm_spawn_f(char *command, char *argv, MPI_Fint *maxprocs,
         opal_argv_free(c_argv);
     }
     if (!OMPI_IS_FORTRAN_ERRCODES_IGNORE(array_of_errcodes)) {
-	OMPI_ARRAY_INT_2_FINT(array_of_errcodes, size);
-    } else {
-	OMPI_ARRAY_FINT_2_INT_CLEANUP(array_of_errcodes);
```
jsquyres: I think you still need the CLEANUP here, right?  I.e., for the non-IGNORE case, we have to have a matching CLEANUP for the ALLOC on new line 94, right?

ompi/mpi/fortran/mpif-h/comm_spawn_multiple_f.c
```diff
@@ -138,9 +140,7 @@ void ompi_comm_spawn_multiple_f(MPI_Fint *count, char *array_commands,
     }
 
     if (!OMPI_IS_FORTRAN_ERRCODES_IGNORE(array_errcds)) {
-	OMPI_ARRAY_INT_2_FINT(array_errcds, size);
-    } else {
-	OMPI_ARRAY_FINT_2_INT_CLEANUP(array_errcds);
```
jsquyres: Similar to above: do we still need the CLEANUP here?

ompi/mpi/fortran/mpif-h/comm_spawn_f.c
ggouaillardet: `OMPI_ARRAY_INT_2_FINT()` includes the cleanup of the array(will double check tomorrow)

jsquyres: Ah yes, you are correct.  That's confusing.  ☹️  But so be it!

hppritcha:
bot:lanl:retest

PR #6797, opened 2019-07-05
Fix if_linux_ipv6 verbose output of interface addresses

orivej:
Previously verbose output of `if_linux_ipv6_open` looked like this:

    found interface ab c: 0ab: a b: abc: 0 0: a 0:abcd: 0 0 scope 0

This changes the output to:

    found interface eth0 inet6 ab0c:ab:a0b:abc:0:a00:abcd:0 scope 0

### DIFF
--- a/opal/mca/if/linux_ipv6/if_linux_ipv6.c
+++ b/opal/mca/if/linux_ipv6/if_linux_ipv6.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2010-2019 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
  * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights
  *                         reserved.
@@ -48,9 +48,12 @@
 #include "opal/constants.h"
 #include "opal/util/if.h"
 #include "opal/util/output.h"
+#include "opal/util/show_help.h"
 #include "opal/util/string_copy.h"
 #include "opal/mca/if/if.h"
 #include "opal/mca/if/base/base.h"
+
+#define LOG_PREFIX "mca: if: linux_ipv6: "
 
 static int if_linux_ipv6_open(void);
 
@@ -77,7 +80,36 @@ opal_if_base_component_t mca_if_linux_ipv6_component = {
     },
 };
 
-/* configure using getifaddrs(3) */
+#if OPAL_ENABLE_IPV6
+static bool hex2int(char hex, int *dst)
+{
+    if ('0' <= hex && hex <= '9') {
+        *dst = hex - '0';
+    } else if ('A' <= hex && hex <= 'F') {
+        *dst = hex - 'A' + 10;
+    } else if ('a' <= hex && hex <= 'f') {
+        *dst = hex - 'a' + 10;
+    } else {
+        return false;
+    }
+    return true;
+
+}
+
+static bool hexdecode(const char *src, uint8_t *dst, size_t dstsize)
+{
+    int hi, lo;
+    for (size_t i = 0; i < dstsize; i++) {
+        if (hex2int(src[i * 2], &hi) && hex2int(src[i * 2 + 1], &lo)) {
+            dst[i] = 16 * hi + lo;
+        } else {
+            return false;
+        }
+    }
+    return true;
+}
+#endif
+
 static int if_linux_ipv6_open(void)
 {
 #if OPAL_ENABLE_IPV6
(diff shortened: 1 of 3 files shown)


bwbarrett:
ok to test

opal/mca/if/linux_ipv6/if_linux_ipv6.c
```diff
@@ -77,7 +77,21 @@ opal_if_base_component_t mca_if_linux_ipv6_component = {
     },
 };
 
-/* configure using getifaddrs(3) */
+static int hex2int(char hex)
+{
+    if ('0' <= hex && hex <= '9') return hex - '0';
+    if ('A' <= hex && hex <= 'F') return hex - 'A' + 10;
+    if ('a' <= hex && hex <= 'f') return hex - 'a' + 10;
+    abort();
```
jsquyres: This is a bit anti-social -- if an unexpected misconfiguration occurs, this could end up dropping core on tens / hundreds / thousands of times across a cluster.  Could you make this a bit more friendly of a failure mode?  Perhaps a call to `opal_show_help()` followed by a call to `exit(1)`, or somesuch.  I realize this is an unlikely error, but I don't think we want to take the chance of dropping TB worth of corefiles on a networked filesystem, for example.

orivej: I though that since this file is Linux specific and Linux is very stable in this regard, we can expect the `/proc/net/if_inet6` format to never change, and this function to always succeed.

Please see the second commit that makes it robust.

jsquyres: You're probably not wrong, but the last time I used `abort()` in a situation that should never have reached that code, a few months later, something else happened in the source to make that condition actually occur -- which resulted in `abort()` being called in an embedded controller.  This filled up the embedded filesystem on dozens on machines, bricking them.  QA was *NOT* happy with me.  😦 

So I've always been super shy of using `abort()` ever since then.  😄

jsquyres:
@orivej I just did the following:

1. Squashed your 2 commits down to 1
1. Added two more small commits (which we should squash before merging):
    1. Use `opal_show_help()` instead of `opal_output()`.  `opal_show_help()` de-duplicates error messages, so that you don't see the same error message from every single MPI process.
    1. Fix 2 minor compiler warnings.

I left these 2 additional commits so that you can easily see the changes.

If you're good with them, I can re-squash all down into a single commit and we can merge when CI completes.

opal/mca/if/linux_ipv6/if_linux_ipv6.c
```diff
@@ -126,8 +127,11 @@ static int if_linux_ipv6_open(void)
             opal_if_t *intf;
 
             if (!hexdecode(addrhex, a6.s6_addr, sizeof a6.s6_addr)) {
-                opal_output(0, LOG_PREFIX "skipped interface %s: failed to hexdecode %s\n",
-                            ifname, addrhex);
+                char hostname[OPAL_MAXHOSTNAMELEN + 1];
```
orivej: `OPAL_MAXHOSTNAMELEN` already accounts for `+ 1`: https://github.com/open-mpi/ompi/blob/v4.0.1/opal/include/opal_config_bottom.h#L295-L302

orivej: Please use:
```
char hostname[OPAL_MAXHOSTNAMELEN];
gethostname(hostname, sizeof hostname);
```
(as in the rest of ompi codebase), because it lets `gethostname` write the final `\0` when the hostname reaches `MAXHOSTNAMELEN` (which is `OPAL_MAXHOSTNAMELEN - 1`).

jsquyres: Good catch; thanks.

I just checked the man page for `gethostname(2)` on RHEL 6 and RHEL 7: it looks like `gethostname()` does *not* guarantee to copy over the null terminator if `len` is too small. 😲

As you cite, however, this is likely a larger issue for the rest of the code base, and likely mitigated by having `OPAL_MAXHOSTNAMELEN` be (the_max+1) (even though glibc `gethostname(2)` does not state that there is any relationship between `MAXHOSTNAMELEN` or `HOST_NAME_MAX` and the maximum string length returned by `gethostname(3)`  ...but glibc `gethostname(3p)` *does* state that `HOST_NAME_MAX` is the max length.  🤷‍♂  FWIW, MacOS 10.14 `gethostname(3)` states that the max hostname length will be no less than `_POSIX_HOST_NAME_MAX`, which is effectively a synonym for `MAXHOSTNAMELEN`).

Probably the simplest "99%" fix to do across the code base is therefore:

```c
char hostname[OPAL_MAXHOSTNAMELEN] = {0};
```

If we assume that `OPAL_MAXHOSTNAMELEN` is, indeed, 1 greater than the max, then the `={0}` will be enough to guarantee that the result is null terminated.

Regardless, this seems like a larger scope (but still probably trivial) issue than this PR.  I put the `={0}` compromise in this PR; I'll open a "help wanted" github issue to apply this kind of trivial fix elsewhere in the code base.

orivej:
```
char hostname[OPAL_MAXHOSTNAMELEN] = {0};
```
The first line has no effect. If `len(gethostname()) < sizeof(hostname)`, then `gethostname` will write `\0` by itself, otherwise the second line will overwrite the final `\0`.

jsquyres: Blarg.  That should have been `sizeof(hostname)-1`, which would have protected against the case where the hostname > `OPAL_MAXHOSTNAMELEN`.

@hppritcha unfortunately merged this PR already, so I can't fix it on this PR.  It's *almost* not worth another PR... but I'll be annoyed with Past Jeff someday if I come across this code again in the future and wonder why the `{0}` is there without the corresponding `-1`.  Arrgh.

hppritcha:
this should not had the v4.0.2 milestone set.  That's for PRs that are going into v4.0.x.  I

PR #6803, opened 2019-07-09
Fix osc sm posts when only 32 bit atomics support

guserav:


bosilca:
ok to test.

guserav:
In v2.x and v3.0.x it just assumes that 64 bit atomics work, also another atomic is used (so this patch wont apply anyway)
It seems like 7cdda242066dfb14451287f724237862af912242 has introduced the problem.
So v3.1.x and v4.0.x have the problem

PR #6811, opened 2019-07-13
fortran/use-mpi-f08: do not slurp the sentinel module files

ggouaillardet:
A sentinel is only an internal Fortran module and hence should not
be slurped into libmpi_usempif08.so

### DIFF
--- a/ompi/mpi/fortran/use-mpi-f08/Makefile.am
+++ b/ompi/mpi/fortran/use-mpi-f08/Makefile.am
@@ -7,8 +7,8 @@
 # Copyright (c) 2012-2013 Inria.  All rights reserved.
 # Copyright (c) 2013-2018 Los Alamos National Security, LLC. All rights
 #                         reserved.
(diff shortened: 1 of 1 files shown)


hppritcha:
this commit busted fortran user defined reduction ops.

jsquyres:
@#$%#!@$%

Do you have a test we can use to replicate the problem?

ggouaillardet:
It turns out

 - `ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-callbacks.F90`
 - `ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-types.F90`

contains some code (just grep the `contains` keyword) so `libforce_usempif08_internal_modules_to_be_built.la` is not only a sentinel, so the two related object files should be slurped into the Fortran library.

I'll work on a fix from Monday.

ggouaillardet:
Refs. #6861

Issue #5459, opened 2018-07-20
Fortran: MPI_Alltoallw/MPI_Ialltoallw using MPI_IN_PLACE do not ignore sendtypes


geimer:
According to the standard, `MPI_Alltoallw` and `MPI_Ialltoallw` should ignore the `sendcounts`, `sdispls`, and `sendtypes` arguments if `sendbuf` is `MPI_IN_PLACE`.  Open MPI (tested with 3.0.0) takes this into account for `sendcounts` and `sdispls`, but not for `sendtypes`.  This leads to a segmentation fault with the following test code:
```fortran
PROGRAM main
    USE mpi
    IMPLICIT NONE

    INTEGER :: size
    INTEGER :: rank
    INTEGER :: ierr
    INTEGER :: i

    INTEGER, ALLOCATABLE :: recvbuf(:)
    INTEGER, ALLOCATABLE :: recvcounts(:)
    INTEGER, ALLOCATABLE :: recvdispls(:)
    INTEGER, ALLOCATABLE :: recvtypes(:)
    INTEGER, ALLOCATABLE :: NULL(:)

    CALL MPI_Init(ierr)
    CALL MPI_Comm_size(MPI_COMM_WORLD, size, ierr)
    CALL MPI_Comm_rank(MPI_COMM_WORLD, rank, ierr)

    ALLOCATE(recvbuf(size))
    ALLOCATE(recvcounts(size))
    ALLOCATE(recvdispls(size))
    ALLOCATE(recvtypes(size))
    DO i = 1, size
        recvcounts(i) = 1
        recvdispls(i) = i * (storage_size(recvbuf) / 8)
        recvtypes(i)  = MPI_INTEGER
        recvbuf(i)    = rank
    END DO

    CALL MPI_Alltoallw(MPI_IN_PLACE, NULL, NULL, NULL, &
                       recvbuf, recvcounts, recvdispls, recvtypes, &
                       MPI_COMM_WORLD, ierr)

    CALL MPI_Finalize(ierr)
END PROGRAM
```
It seems that the loops at https://github.com/open-mpi/ompi/blob/e1728490394787a0b30c19583044d52e5efc2c03/ompi/mpi/fortran/mpif-h/alltoallw_f.c#L95-L99 and https://github.com/open-mpi/ompi/blob/e1728490394787a0b30c19583044d52e5efc2c03/ompi/mpi/fortran/mpif-h/ialltoallw_f.c#L96-L100 are the culprit.

[...]

PR #6813, opened 2019-07-13
fortran/mpif-h: fix MPI_Alltoallw() binding

ggouaillardet:
- ignore sendcounts, sendispls and sendtypes arguments when MPI_IN_PLACE is used
 - use the right size when an inter-communicator is used.

Thanks Markus Geimer for reporting this.

Refs. open-mpi/ompi#5459

### DIFF
--- a/ompi/mpi/fortran/mpif-h/alltoallw_f.c
+++ b/ompi/mpi/fortran/mpif-h/alltoallw_f.c
@@ -10,7 +10,7 @@
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2011-2012 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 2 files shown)


jsquyres:
@ggouaillardet Does this need to get cherry picked to release branches?

ggouaillardet:
@jsquyres yes, this has to be backported to the release branches.
 - good news is this can be seen as a corner case
 - bad news is the reported issue is the top of the iceberg. `mpif-h` bindings do **not** ignore parameters that should be ignored. Also, non blocking collectives use an intermediate malloc'ed array that is free'd right after the PMPI subroutine returns, which is incorrect since it can only be free'd after the non blocking operation completes (or is free'd if this is a persistent non blocking collective).

I am working on a patch that keeps growing, and I'd rather backport the fixes  all at once.

PR #6820, opened 2019-07-15
btl tcp: Fix error path memory leak

wckzhang:
After the OPAL_MODEX_RECV call, remote_addrs was not freed in the error
path. Moved the free call into cleanup to ensure we always free this
memory before leaving the function.

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_proc.c
+++ b/opal/mca/btl/tcp/btl_tcp_proc.c
@@ -17,7 +17,7 @@
  * Copyright (c) 2015-2016 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2015-2018 Cisco Systems, Inc.  All rights reserved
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

opal/mca/btl/tcp/btl_tcp_proc.c
```diff
@@ -223,6 +221,10 @@ mca_btl_tcp_proc_t* mca_btl_tcp_proc_create(opal_proc_t* proc)
         }
     }
 
+    if (remote_addrs){
```
bwbarrett: I missed this in the internal review, but we explicitly check against NULL in Open MPI.  So this should be if (NULL == remote_addrs) {.

wckzhang: Ah yes, I will change this.

wckzhang: Oops, should be if (NULL != remote_addrs)

PR #6825, opened 2019-07-18
memory/patcher: Added missed checking for returning value of SYS_brk calling

nikitaxgusev:
Signed-off-by: Nikita Gusev <nikita.gusev@intel.com>

### DIFF
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -591,6 +591,9 @@ static int patcher_open (void)
 
 #if defined (SYS_brk)
     rc = opal_patcher->patch_symbol ("brk", (uintptr_t)intercept_brk, (uintptr_t *) &original_brk);
+    if (OPAL_SUCCESS != rc) {
+        return rc;
+    }
 #endif
 
     return rc;


hoopoepg:
ok to test

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -591,6 +591,9 @@ static int patcher_open (void)
 
 #if defined (SYS_brk)
     rc = opal_patcher->patch_symbol ("brk", (uintptr_t)intercept_brk, (uintptr_t *) &original_brk);
+    if (OPAL_SUCCESS != rc) {
```
hoopoepg: rc will be returned in any case, here evaluation of returned value is not needed

nikitaxgusev: @hoopoepg ok. I would like to support the code style. Should I close it?

hoopoepg: let's wait for another opinions :)

hjelmn: The code was intentionally written this way. The return code is already passed up.

PR #6828, opened 2019-07-18
Try to prevent the compiler from optimizing out MPIR_Breakpoint().

awlauria:
Signed-off-by: Austen Lauria <awlauria@us.ibm.com>

PR #6837, opened 2019-07-23
Update minimum versions of automake and libtool

gpaulsen:
Recently we ran into a strange issue where incorrect CFLAGS were added
to makefiles for various source files when incorrect minimum version
of Automake was used (see some discussion near the end of PR #6828 -- 
specifically https://github.com/open-mpi/ompi/pull/6828#issuecomment-514315352).

Versions taken from (https://www.open-mpi.org/source/building.php)

This change will create a hard error if the user is trying to run
with too early of versions, rather than confusing runtime behavior.

This Commit should be PRed back to v3.0.x, v3.1.x and v4.0.x

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -61,9 +61,9 @@ my $include_list;
 my $exclude_list;
 
 # Minimum versions
-my $ompi_automake_version = "1.13.4";
+my $ompi_automake_version = "1.15";
 my $ompi_autoconf_version = "2.69";
-my $ompi_libtool_version = "2.4.2";
+my $ompi_libtool_version = "2.4.6";
 
 # Search paths
 my $ompi_autoconf_search = "autoconf";


gpaulsen:
I tested and when run with a bad version, instead of silently failing we see:
```
=================================================================
I could not find a recent enough copy of automake.
I need at least 1.15, but only found the following versions:

    automake: 1.13.4

I am gonna abort.  :-(

Please make sure you are using at least the following versions of the
tools:

    GNU Autoconf: 2.69
    GNU Automake: 1.15
    GNU Libtool: 2.4.6
=================================================================
```

awlauria:
Thanks!

rhc54:
I would recommend you also provide an override flag, as we have done in the `make tarball` scripts. Not every build nor developer has an issue with the flags from the earlier tools - e.g., the Jenkins testers couldn't care less - so no reason to force it.

In fact, why not just use the same logic we used there? You don't abort the build - you emit the warning, explaining why it might be an issue, and then pause for 5 seconds before proceeding to execute. If they give the override flag, then don't even bother with the pause.

I guess I'd rather not force someone to update their autotools if they don't use a compiler/tool combination that exhibits this issue.

bwbarrett:
The CI tests currently use the system installed autotools; it looks like the ARM hosts LANL helped get us access to are using an older version of Libtool than this update requires.  I guess I have trouble seeing the advantage of this patch.  I'd also really prefer not make all the CI even longer for a pause :).

jjhursey:
As long as we roll tarballs with the required level of automake then anyone picking up a release/rc/nightly will be ok (right?). I worry about developers that accidentally stumble over this in the future though.

Adding an override sounds fine. Alternatively, changing this from a hard error to an autogen time warning message would be fine (something descriptive like "Hey you have an older automake. Be careful which compiler you pick at configure time and check out issue XYZ for details").

However, this is a functional correctness issue so some form of enforcement to make sure that we are building the library correctly is needed. I'm fine backing that off to a build time warning message if it is easily recognizable.

rhc54:
Perhaps I'm misunderstanding, but is this a problem on every compiler, or just xlc/pgi? If the latter, then perhaps the enforcement needs to be compiler-specific and should be done in configure instead of autogen. If it is a problem for all compilers when using older automake, then just a warning should be fine as people using tarballs are already warned not to run autogen, and thus should never see the problem. I agree with Brian about not including a timeout - I find that really annoying.

hppritcha:
bot:ompi:retest

gpaulsen:
@jsquyres Any thoughts on reactions cited on this PR?

I'm shocked that an update in automake from 1.13.4 to 1.15 could silently change the CFLAGS being generated for Open MPI.  Apparently something in the Open-MPI makefiles are sensitive to the automake version (manifested with gcc 4.8.5 on ppc64le).

gpaulsen:
retest

gpaulsen:
bot:ompi:retest

gpaulsen:
So what's the plan for this PR?
I initially created it so that Open-MPI CODE would reflect our documentation: https://www.open-mpi.org/source/building.php
These are not new build requirements.  According to the table, Open MPI has required these levels of tools for developers since v2.x.
IBM happened to get bit by this on ppc64le gnu/xlc compilers using an old version of one of these tools because this protection was not updated in the v2.x time frame.  The problem is, that it was a silent build failure (i.e. it built okay we thought), so others might be having similar issues and not know it.

gpaulsen:
bot:ompi:retest

gpaulsen:
@hppritcha This PR is failing on `cray_cle`.  Do you want to update that cluster before merging this?

hppritcha:
bot:ompi:retest

gpaulsen:
bot:ompi:retest

gpaulsen:
In our RM meeting this week, we're concerned that this PR is too heavy of a hammer to require all developers to upgrade to these new tool versions.
 - Users who download the tarball, should not need these versions, and that they can build with older versions.
 - developers may not want to have to use these never versions of the tools if they know that they will not hit silent runtime errors like: https://github.com/open-mpi/ompi/pull/6828#issuecomment-514315352

It was also pointed out that this PR is not complete as we should also update `configure.ac` to match.

ibm-ompi:

 Gist: https://gist.github.com/55df9a2a430333d70b35a095734a65d3

awlauria:
bot:ibm:retest

awlauria:
bot:ibm:gnu:retest
bot:ibm:scale:80:test

ibm-ompi:

 Gist: https://gist.github.com/f90a8cb3b2b52b05bb0e34367b817f15

ibm-ompi:

 Gist: https://gist.github.com/b974f84d631af4bee6f7712309ec1a54

ibm-ompi:

 Gist: https://gist.github.com/adc9aae8527ead1f2cd4807c248e532e

jjhursey:
The IBM CI shows up twice because the older, passing ones were running bare metal with custom tools builds. The newer, failing ones are running in a container with the autoconf/automake/libtool from Centos 7 which are the levels below:
```
[root@25d252287c8a /]# autoconf --version
autoconf (GNU Autoconf) 2.69
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<http://gnu.org/licenses/gpl.html>, <http://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by David J. MacKenzie and Akim Demaille.
[root@25d252287c8a /]# automake --version
automake (GNU automake) 1.13.4
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl-2.0.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Tom Tromey <tromey@redhat.com>
       and Alexandre Duret-Lutz <adl@gnu.org>.
[root@25d252287c8a /]# libtool --version
libtool (GNU libtool) 2.4.2
Written by Gordon Matzigkeit <gord@gnu.ai.mit.edu>, 1996

Copyright (C) 2011 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I can install a new version in the container, but this does mean that users on Centos7 would have to do the same.

lanl-ompi:

jsquyres:
Converted this PR to draft because it still says "WIP DNM", and there are unanswered questions.

PR #6840, opened 2019-07-24
osc/ucx: Fix data corruption with non-contiguous accumulates

nysal:
This PR fixes a data corruption issue in MPI_Accumulate and MPI_Get_accumulate when using non-contiguous data types. The c_strided_acc_indexed and c_strided_acc_subarray tests in the ibm test suite now pass. The c_strided_getacc_indexed test case can be modified to use the MPI_SUM operation to expose a similar issue with MPI_Get_accumulate.

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -510,12 +510,13 @@ int ompi_osc_ucx_accumulate(const void *origin_addr, int origin_count,
             if ((op != &ompi_mpi_op_maxloc.op && op != &ompi_mpi_op_minloc.op) ||
                 ompi_datatype_is_contiguous_memory_layout(temp_dt, temp_count)) {
                 size_t temp_size;
+                char *curr_temp_addr = (char *)temp_addr;
                 ompi_datatype_type_size(temp_dt, &temp_size);
                 while (origin_ucx_iov_idx < origin_ucx_iov_count) {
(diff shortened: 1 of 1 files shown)


hppritcha:
@nysal does this need to backportrd to 4.0.x and older branches?

nysal:
@hppritcha For v4.0.x I've opened https://github.com/open-mpi/ompi/pull/6846 
The same patch fixes a crash in v3.1.x, however when I try to run with anything more than 2 procs, there is a hang. It looks like this fix exposes a different issue in v3.1.x. @janjust any thoughts about v3.1.x?

janjust:
@nysal , what's the hang's backtrace?

nysal:
With 4 ranks, 0 and 2 are in accumulate. 1 and 3 are in the subsequent barrier. 
I've posted the stack traces for ranks 2 and 3. Its the same for 0 and 1 - https://gist.github.com/nysal/52826a07f6d2d2dd8853e9342b73bf4b

janjust:
@nysal my suggestion is to backport the datacorruption and open up an issue for hang and resolve hang

nysal:
Done - https://github.com/open-mpi/ompi/pull/6848
I'll open an issue for the hang shortly.

artpol84:
@nysal, can you share your reproducer for this?

nysal:
@artpol84 Its the c_strided_acc_indexed test in the ibm test suite from ompi-tests

artpol84:
Thank you!

PR #6844, opened 2019-07-25
Do not use CMA in user namespaces

adrianreber:
Trying out to run processes via mpirun in Podman containers has shown that the CMA btl_vader_single_copy_mechanism does not work when user namespaces are involved.

Creating containers with Podman requires at least user namespaces to be able to do unprivileged mounts in a container

Even if running the container with user namespace user ID mappings which result in the same user ID on the inside and outside of all involved containers, the check in the kernel to allow ptrace (and thus process_vm_{read,write}v()), fails if the same IDs are not in the same user namespace.

One workaround is to specify '--mca btl_vader_single_copy_mechanism none' and this commit adds code to automatically skip CMA if user namespaces are detected.

Preferred implementation would have been to detect if the other local processes are running in different user namespaces, but it was not clear how get the PIDs of the other involved processes in
mca_btl_vader_check_single_copy(). This is even more complicated if some processes would be running in the same user namespace, but not all of them. If one different user namespace is detected, CMA should be disabled for all involved processes. So if one local process detects that CMA is not working it would need to communicate this information to all local processes.

This implementation now checks during the first access of mca_btl_vader_{put,get}_cma() if the destination process is running in another user namespace and switches to MCA_BTL_VADER_EMUL if this is true.

So if the first access to process_vm_{read,write}v()) fails all further accesses are automatically no longer trying to use CMA.

### DIFF
--- a/opal/mca/btl/vader/btl_vader.h
+++ b/opal/mca/btl/vader/btl_vader.h
@@ -84,7 +84,12 @@ union vader_modex_t {
         void *segment_base;
     } xpmem;
 #endif
-    opal_shmem_ds_t seg_ds;
+    struct vader_modex_other_t {
+        ino_t user_ns_id;
+        int seg_ds_size;
+        /* seg_ds needs to be the last element */
+        opal_shmem_ds_t seg_ds;
+    } other;
 };
 
 /**
@@ -269,6 +274,8 @@ int mca_btl_vader_get_knem (mca_btl_base_module_t *btl, mca_btl_base_endpoint_t 
                             mca_btl_base_registration_handle_t *remote_handle, size_t size, int flags,
                             int order, mca_btl_base_rdma_completion_fn_t cbfunc, void *cbcontext, void *cbdata);
 #endif
+
+ino_t mca_btl_vader_get_user_ns_id(void);
 
 int mca_btl_vader_get_sc_emu (mca_btl_base_module_t *btl, mca_btl_base_endpoint_t *endpoint, void *local_address,
                                uint64_t remote_address, mca_btl_base_registration_handle_t *local_handle,
--- a/opal/mca/btl/vader/btl_vader_component.c
+++ b/opal/mca/btl/vader/btl_vader_component.c
@@ -41,6 +41,10 @@
 #include "btl_vader_fbox.h"
 #include "btl_vader_xpmem.h"
 
+#ifdef HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif
+
 #include <sys/mman.h>
 #include <fcntl.h>
 
@@ -350,6 +354,25 @@ static int mca_btl_vader_component_close(void)
     return OPAL_SUCCESS;
 }
 
+/*
+ * mca_btl_vader_parse_proc_ns_user() tries to get the user namespace ID
+ * of the current process.
+ * Returns the ID of the user namespace. In the case of an error '0' is returned.
+ */
+ino_t mca_btl_vader_get_user_ns_id(void)
+{
+    struct stat buf;
+
+    if (0 > stat("/proc/self/ns/user", &buf)) {
+        /*
+         * Something went wrong, probably an old kernel that does not support namespaces
+         * simply assume all processes are in the same user namespace and return 0
+         */
+        return 0;
+    }
+
+    return buf.st_ino;
+}
 static int mca_btl_base_vader_modex_send (void)
 {
     union vader_modex_t modex;
@@ -363,8 +386,16 @@ static int mca_btl_base_vader_modex_send (void)
         modex_size = sizeof (modex.xpmem);
     } else {
 #endif
-        modex_size = opal_shmem_sizeof_shmem_ds (&mca_btl_vader_component.seg_ds);
-        memmove (&modex.seg_ds, &mca_btl_vader_component.seg_ds, modex_size);
+        modex.other.seg_ds_size = opal_shmem_sizeof_shmem_ds (&mca_btl_vader_component.seg_ds);
+        memmove (&modex.other.seg_ds, &mca_btl_vader_component.seg_ds, modex.other.seg_ds_size);
+        modex.other.user_ns_id = mca_btl_vader_get_user_ns_id();
+        /*
+         * If modex.other.user_ns_id is '0' something did not work out
+         * during user namespace detection. Assuming there are no
+         * namespaces available it will return '0' for all processes and
+         * the check later will see '0' everywhere and not disable CMA.
+         */
+        modex_size = sizeof (modex.other);
 
 #if OPAL_BTL_VADER_HAVE_XPMEM
     }
--- a/opal/mca/btl/vader/btl_vader_module.c
+++ b/opal/mca/btl/vader/btl_vader_module.c
@@ -27,6 +27,7 @@
  */
 
 #include "opal_config.h"
+#include "opal/util/show_help.h"
 
 #include "btl_vader.h"
 #include "btl_vader_endpoint.h"
@@ -78,6 +79,28 @@ mca_btl_vader_t mca_btl_vader = {
         .btl_ft_event = vader_ft_event
     }
 };
+
+/*
+ * Exit function copied from btl_usnic_util.c
+ *
+ * The following comment tells Coverity that this function does not return.
+ * See https://scan.coverity.com/tune.
+ */
+
+/* coverity[+kill] */
+static void vader_btl_exit(mca_btl_vader_t *btl)
+{
+    if (NULL != btl && NULL != btl->error_cb) {
+        btl->error_cb(&btl->super, MCA_BTL_ERROR_FLAGS_FATAL,
+                      (opal_proc_t*) opal_proc_local_get(),
+                      "The vader BTL is aborting the MPI job (via PML error callback).");
+    }
+
+    /* If the PML error callback returns (or if there wasn't one), just exit. Shrug. */
+    fprintf(stderr, "*** The Open MPI vader BTL is aborting the MPI job (via exit(3)).\n");
+    fflush(stderr);
+    exit(1);
+}
 
 static int vader_btl_first_time_init(mca_btl_vader_t *vader_btl, int n)
 {
@@ -173,6 +196,7 @@ static int vader_btl_first_time_init(mca_btl_vader_t *vader_btl, int n)
 static int init_vader_endpoint (struct mca_btl_base_endpoint_t *ep, struct opal_proc_t *proc, int remote_rank) {
     mca_btl_vader_component_t *component = &mca_btl_vader_component;
     union vader_modex_t *modex;
+    ino_t my_user_ns_id;
     size_t msg_size;
     int rc;
 
@@ -197,16 +221,57 @@ static int init_vader_endpoint (struct mca_btl_base_endpoint_t *ep, struct opal_
         } else {
 #endif
             /* store a copy of the segment information for detach */
-            ep->segment_data.other.seg_ds = malloc (msg_size);
+            ep->segment_data.other.seg_ds = malloc (modex->other.seg_ds_size);
             if (NULL == ep->segment_data.other.seg_ds) {
                 return OPAL_ERR_OUT_OF_RESOURCE;
             }
 
-            memcpy (ep->segment_data.other.seg_ds, &modex->seg_ds, msg_size);
+            memcpy (ep->segment_data.other.seg_ds, &modex->other.seg_ds, modex->other.seg_ds_size);
 
             ep->segment_base = opal_shmem_segment_attach (ep->segment_data.other.seg_ds);
             if (NULL == ep->segment_base) {
                 return OPAL_ERROR;
+            }
+
+            if (MCA_BTL_VADER_CMA == mca_btl_vader_component.single_copy_mechanism) {
+                my_user_ns_id = mca_btl_vader_get_user_ns_id();
+                if (my_user_ns_id != modex->other.user_ns_id) {
+                    mca_base_var_source_t source;
+                    int vari;
+                    rc = mca_base_var_find_by_name("btl_vader_single_copy_mechanism", &vari);
+                    if (OPAL_ERROR == rc) {
+                        return OPAL_ERROR;
+                    }
+                    rc = mca_base_var_get_value(vari, NULL, &source, NULL);
+                    if (OPAL_ERROR == rc) {
+                        return OPAL_ERROR;
+                    }
+                    /*
+                     * CMA is not possible as different user namespaces are in use.
+                     * Currently the kernel does not allow * process_vm_{read,write}v()
+                     * for processes running in different user namespaces even if
+                     * all involved user IDs are mapped to the same user ID.
+                     *
+                     * Fallback to MCA_BTL_VADER_EMUL.
+                     */
+                    if (MCA_BASE_VAR_SOURCE_DEFAULT != source) {
+                        /* If CMA has been explicitly selected we want to error out */
+                        opal_show_help("help-btl-vader.txt", "cma-different-user-namespace-error",
+                                       true, opal_process_info.nodename);
+                        vader_btl_exit(&mca_btl_vader);
+                    }
+                    /*
+                     * If CMA has been selected because it is the default or
+                     * some fallback, this falls back even further.
+                     */
+                    opal_show_help("help-btl-vader.txt", "cma-different-user-namespace-warning",
+                                   true, opal_process_info.nodename);
+                    mca_btl_vader_component.single_copy_mechanism = MCA_BTL_VADER_EMUL;
+                    mca_btl_vader.super.btl_get = mca_btl_vader_get_sc_emu;
+                    mca_btl_vader.super.btl_put = mca_btl_vader_put_sc_emu;
+                    mca_btl_vader.super.btl_put_limit = mca_btl_vader.super.btl_max_send_size - sizeof (mca_btl_vader_sc_emu_hdr_t);
+                    mca_btl_vader.super.btl_get_limit = mca_btl_vader.super.btl_max_send_size - sizeof (mca_btl_vader_sc_emu_hdr_t);
+                }
             }
 #if OPAL_BTL_VADER_HAVE_XPMEM
         }
--- a/opal/mca/btl/vader/help-btl-vader.txt
+++ b/opal/mca/btl/vader/help-btl-vader.txt
@@ -126,6 +126,25 @@ mechanism if one is available. This may result in lower performance.
 
   Local host: %s
 #
+[cma-different-user-namespace-error]
+ERROR: Linux kernel CMA support was requested via the
+btl_vader_single_copy_mechanism MCA variable, but CMA support is
+not available due to different user namespaces.
+
+Your MPI job will abort now. Please select another value for
+btl_vader_single_copy_mechanism.
+
+  Local host: %s
+#
+[cma-different-user-namespace-warning]
+WARNING: The default btl_vader_single_copy_mechanism CMA is
+not available due to different user namespaces.
+
+The vader shared memory BTL will fall back on another single-copy
+mechanism if one is available. This may result in lower performance.
+
+  Local host: %s
+#
 [xpmem-make-failed]
 WARNING: Could not generate an xpmem segment id for this process'
 address space.


rhc54:
I'm not familiar with Podman. Are you saying that at time of launch, you have no idea that the containers are going to wind up in different user namespaces? That the _only_ way to know is to search the `/proc` file system to discover what the other containers did? Are the containers setting the namespace upon `exec` - or is the kernel setting the namespace upon `fork`?

Assuming that situation is true (and I sincerely hope it isn't): Given that you are using mpirun to start the containers, and that mpirun knows the pid of each proc once it does the `fork`, and assuming it is the kernel setting the namespace, I'm wondering if it wouldn't be more efficient to have mpirun:

* fork all the local procs, but don't exec them yet
* check the namespace of each pid it just generated
* add a PMIx attribute to the local job info indicating if different namespaces are in operation
* exec the procs

If the container is setting the namespace at `exec`, then the only other option would be to `modex_send` the namespace of the process during `vader` component open, and have each process `modex_recv` the namespace of all local procs during `vader_add_procs`. This would remove the performance penalty on first access this PR introduces and ensure that _all_ local procs disable CMA if needed.

We have found in other scenarios that having every local process access `/proc` really impacts scalability of launch, so avoiding that method is highly preferable.

adrianreber:
> I'm not familiar with Podman.

Podman uses, just like Docker, runc to run the containers. runc is an implementation of the OCI spec. runc, as far as I understand it, does first a clone() and then an unshare() for the namespace setup.

> Are you saying that at time of launch, you have no idea that the containers are going to wind up in different user namespaces?

Well, if I know that I am running Podman, I probably could assume that each process will run in its own namespace. There is, however, an option to run different containers in the same user namespace. So just by saying it is a Podman container it will be in different user namespaces is not correct. There are also other container runtimes which are using user namespaces (for namespace based container user namespaces are basically required to be able to mount something in the container), so you would need to detect those also.

> That the _only_ way to know is to search the `/proc` file system to discover what the other containers did?

As far as I understand it, yes, looking at `/proc` is the only way. I will ask someone more familiar to also comment here.

> Are the containers setting the namespace upon `exec` - or is the kernel setting the namespace upon `fork`?

The namespaces are set up by runc, which is started by Podman (Podman actually starts conmon and conmon starts runc). There is a long comment how this is done at https://github.com/opencontainers/runc/blob/master/libcontainer/nsenter/nsexec.c#L645

> Assuming that situation is true (and I sincerely hope it isn't): Given that you are using mpirun to start the containers, and that mpirun knows the pid of each proc once it does the `fork`, and assuming it is the kernel setting the namespace, I'm wondering if it wouldn't be more efficient to have mpirun:
> 
>     * fork all the local procs, but don't exec them yet
> 
>     * check the namespace of each pid it just generated
> 
>     * add a PMIx attribute to the local job info indicating if different namespaces are in operation
> 
>     * exec the procs

I am pretty sure this will not work.

> If the container is setting the namespace at `exec`, then the only other option would be to `modex_send` the namespace of the process during `vader` component open, and have each process `modex_recv` the namespace of all local procs during `vader_add_procs`. This would remove the performance penalty on first access this PR introduces and ensure that _all_ local procs disable CMA if needed.

I think I tried to do something like this, but I did not manage to do the `modex_send` and `modex_recv` correctly. I can try to come a with a patch to do this, but last time I got really confused about how modex works. Maybe with this description and the clarification from the mailing list I can try it once more.

> We have found in other scenarios that having every local process access `/proc` really impacts scalability of launch, so avoiding that method is highly preferable.

Understood.

bwbarrett:
I agree with @rhc54; publishing the namespace id as part of modex send/recv seems like a much better idea.

If CMA is explicitly requested, the result of not being in the same namespace (and therefore not using CMA) should be a hard error, not a warning (ie, not the `cma-different-user-namespace` warning).  I could be reading the code wrong, but it appears that the patch would not fail in this situation.

adrianreber:
> If CMA is explicitly requested, the result of not being in the same namespace (and therefore not using CMA) should be a hard error, not a warning (ie, not the `cma-different-user-namespace` warning). I could be reading the code wrong, but it appears that the patch would not fail in this situation.

I implemented the same behaviour as requesting cma in a non-namespaced environment. vader will fall back to another single-copy mechanism. No hard error.

Just to make sure, before looking into @rhc54 recommendation, how it should fail. Same behaviour as the current cma code, or hard error if running in user namespaces?

jsquyres:
@adrianreber We have a general rule of thumb in Open MPI: if a human asks for something and we can't deliver it, that's a hard error / exit.  If Vader isn't obeying that in some cases, that's actually a bug that should be fixed.

adrianreber:
@jsquyres Perfect, thanks.

adrianreber:
I force pushed a new version based on recv() and send().

The only thing missing is the hard error if the user explicitly selected `cma`. I am not sure where to track the user selected single_copy_mechanism. There is `mca_btl_vader_component.single_copy_mechanism` which can change its value.

Should I just introduce a new field in `struct mca_btl_vader_component_t` to track the user selected single_copy_mechanism? The problem I see is, how can I differentiate if the user has not selected anything and vader is just using the first single_copy_mechanism (which can be CMA), or if the user has actually selected CMA. Is it possible to see if the default value is used or some specific option has been selected.

opal/mca/btl/vader/btl_vader.h
```diff
@@ -84,7 +84,11 @@ union vader_modex_t {
         void *segment_base;
     } xpmem;
 #endif
-    opal_shmem_ds_t seg_ds;
+    struct vader_modex_other_t {
+        int seg_ds_size;
+        opal_shmem_ds_t seg_ds;
```
hjelmn: The seg_ds should always be last. We don't modex send the entire structure as it is padded out with 0's.

opal/mca/btl/vader/btl_vader_component.c
```diff
@@ -363,8 +405,13 @@ static int mca_btl_base_vader_modex_send (void)
         modex_size = sizeof (modex.xpmem);
     } else {
 #endif
-        modex_size = opal_shmem_sizeof_shmem_ds (&mca_btl_vader_component.seg_ds);
-        memmove (&modex.seg_ds, &mca_btl_vader_component.seg_ds, modex_size);
+        modex.other.seg_ds_size = opal_shmem_sizeof_shmem_ds (&mca_btl_vader_component.seg_ds);
+        memmove (&modex.other.seg_ds, &mca_btl_vader_component.seg_ds, modex.other.seg_ds_size);
+        modex.other.user_ns_id = mca_btl_vader_parse_proc_ns_user();
+        if (0 == modex.other.user_ns_id) {
+            return OPAL_ERROR;
```
jsquyres: Is this really an error?  E.g., are there older versions of Linux that will reach this vader code but not have the namespace link (or have it in an unexpected format) as expected in `mca_btl_vader_parse_proc_ns_user()`?

I.e., is there a case where we will fail to obtain/parse the namespace ID, but that really means there is no namespace, and therefore all processes have the same lack of namespace?

opal/mca/btl/vader/btl_vader.h
adrianreber: okay, thanks, changed in the upcoming force push

opal/mca/btl/vader/btl_vader_component.c
adrianreber: Makes sense to not treat '0' as an error. Changed in upcoming force push.

adrianreber:
> You should be able to look at mca_base_var_get_value() to get the source of the MCA variable value -- e.g., whether it was just the default value, or whether a user actually set that value.

Thanks. I included some code based on this to OPAL_ERROR if CMA is not selected as part of MCA_BASE_VAR_SOURCE_DEFAULT.

I hope I addressed all open points.

jsquyres:
False cray failure: bot:ompi:retest

opal/mca/btl/vader/btl_vader_component.c
```diff
@@ -350,6 +350,48 @@ static int mca_btl_vader_component_close(void)
...
+        opal_output(0, "Error reading user namespace ID of process %d\n", pid);
+        return 0;
+    }
+
+    /* remove trailing ']' */
+    link[i - 1] = '\0';
+    tmp = strchr(link, '[');
+    if (NULL == tmp) {
+        free(link);
+        opal_output(0, "Error reading user namespace ID of process %d\n", pid);
```
jsquyres: I'm sorry -- one more nit pick...

If this error occurs, it is likely to happen N times on a single node (i.e., for all MPI procs launched on the same node).  If the cluster is homogeneous, it'll happen for every single MPI process in the job.

As such, this `opal_output()` (and the one above it) should be `opal_show_help()`, because `opal_show_help()` de-duplicates (i.e., it'll only show it in the output of `mpirun` once, even if multiple processes emit the same help message).

That being said, do we know if the format of the `/proc/self/ns/user` link has always been the same (`user:[<inode-number>]`)?  I.e., is it possible we're running on an older Linux kernel with a different sym link format, and the format is *not* actually an error?

This is a logical question to ask since you handle the case of having no sym link by simply returning 0 without emitting any kind of error message.

ggouaillardet:
@adrianreber you can simply get the namespace id via a `stat()` as in the inline patch below

can you also please explain why you had to add `struct vader_modex_other_t` with an `int seg_ds_size` and move `seg_ds` into that struct ?

```patch
diff --git a/opal/mca/btl/vader/btl_vader_component.c b/opal/mca/btl/vader/btl_vader_component.c
index dc299b1..922658c 100644
--- a/opal/mca/btl/vader/btl_vader_component.c
+++ b/opal/mca/btl/vader/btl_vader_component.c
@@ -41,6 +41,9 @@
 #include "btl_vader_fbox.h"
 #include "btl_vader_xpmem.h"
 
+#ifdef HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif
 #include <sys/mman.h>
 #include <fcntl.h>
 
@@ -351,47 +354,25 @@ static int mca_btl_vader_component_close(void)
 }
 
 /*
- * mca_btl_vader_parse_proc_ns_user() tries to parse the user namespace ID
+ * mca_btl_vader_parse_proc_ns_user() tries to get the user namespace ID
  * of the current process.
  * Returns the ID of the user namespace. In the case of an error '0' is returned.
  */
 uint64_t mca_btl_vader_parse_proc_ns_user(void)
 {
-    char *link = malloc(PATH_MAX);
-    pid_t pid = getpid();
-    uint64_t user_ns_id;
-    char *tmp;
-    int i;
-
-    i = readlink("/proc/self/ns/user", link, PATH_MAX);
-    if (-1 == i) {
-        return 0;
-    }
-
-    /*
-     * Result in link should look like 'user:[<inode-number>]', so at least
-     * 8 characters long: 'user:[?]'.
-     */
-    if (8 > i) {
-        free(link);
-        opal_output(0, "Error reading user namespace ID of process %d\n", pid);
-        return 0;
-    }
+    struct stat buf;
 
-    /* remove trailing ']' */
-    link[i - 1] = '\0';
-    tmp = strchr(link, '[');
-    if (NULL == tmp) {
-        free(link);
-        opal_output(0, "Error reading user namespace ID of process %d\n", pid);
+    if (0 > stat("/proc/self/ns/user", &buf)) {
+        /*
+         * Something went wrong, probably an old kernel that does not support namespaces
+         * simply assume all processes are in the same user namespace and return 0
+         */
         return 0;
     }
 
-    user_ns_id = strtoul(tmp + 1, NULL, 10);
-    free(link);
-
-    return user_ns_id;
+    return (uint64_t)buf.st_ino;
 }
+
 static int mca_btl_base_vader_modex_send (void)
 {
     union vader_modex_t modex;
```

adrianreber:
@ggouaillardet 

>  you can simply get the namespace id via a `stat()` as in the inline patch below

Are you sure? Using the command-line tool `stat` on `/proc/self/ns/user` this is not true. I assumed that the inode given by `stat` is the same as with `stat(2)`.

ggouaillardet:
@adrianreber That is why assumptions should be avoided :-)

this is a counter-intuitive, but `stat(1)` (e.g. CLI) uses `lstat(2)` and **not** `stat(2)`

from the CLI, you have to

    stat -L /proc/self/ns/user

that uses `stat(2)`, and use glibc's `stat()` in your C code.

adrianreber:
@ggouaillardet 
> can you also please explain why you had to add struct vader_modex_other_t with an int seg_ds_size and move seg_ds into that struct ?

I cannot really explain it. `vader_modex_t` was a union:
```
union vader_modex_t {
#if OPAL_BTL_VADER_HAVE_XPMEM
    struct vader_modex_xpmem_t {
        xpmem_segid_t seg_id;
        void *segment_base;
    } xpmem;
#endif 
    opal_shmem_ds_t seg_ds;
}
```
So I thought to extend the non xpmem part I have to create a struct just like `vader_modex_xpmem_t`. Do you have any other recommendations how to do it better?

opal/mca/btl/vader/btl_vader_component.c
adrianreber: > As such, this opal_output() (and the one above it) should be opal_show_help(), because opal_show_help() de-duplicates (i.e., it'll only show it in the output of mpirun once, even if multiple processes emit the same help message).

Will change it.

> That being said, do we know if the format of the /proc/self/ns/user link has always been the same (user:[<inode-number>])? I.e., is it possible we're running on an older Linux kernel with a different sym link format, and the format is not actually an error?

With the changes from @ggouaillardet included to use `stat()` this is probably irrelevant.

adrianreber:
@ggouaillardet Thanks for the simplification. Changed it and force pushed. As this removed all occurrences of `opal_output()` this should also solve the questions from @jsquyres.

ibm-ompi:

ibm-ompi:

jjhursey:
bot:ibm:retest

hjelmn:
It is a union because either Vader is using xpmem or or isn't. Do you know if xpmem works between namespaces? If not then additional changes are needed.

adrianreber:
> Do you know if xpmem works between namespaces?

Unfortunately I know nothing about xpmem. The vader code is probably the first time I read about it.

hppritcha:
bot:ompi:retest

adrianreber:
Are there any more changes required from my side? Not sure if all discussions are resolved or not.

adrianreber:
Any more comments, reviews, suggestions? Any chance to get this merged?

rhc54:
I don't see how we can merge it without resolving the xpmem concern raised by @hjelmn

hjelmn:
@rhc54 I can test but will need something to test with.

adrianreber:
> I don't see how we can merge it without resolving the xpmem concern raised by @hjelmn

As I cannot test xpmem I cannot implement that part. Closing this PR.

jsquyres:
Sorry for letting this fall off our radar for so long.  😦   FWIW, I see that you made the change that @hjelmn asked for forever ago; he just never re-reviewed it.

@adrianreber Can you supply us with a recipe to re-create / test the issue with multiple namespaces?

FWIW, this PR generally looks good to me.  I do not have XPMEM / cannot easily test, but I would hate to lose this work.

hppritcha:
Doesn't appear that xpmem has the same issue concerning different NS's:
See https://hpc.github.io/charliecloud/faq.html?highlight=xpmem
I've asked the container group to test this PR without the ch_run --join option.

hjelmn:
@hppritcha I am :+1: on this PR as long as we get the testing coverage. If I can get step-by-step testing instructions I can test as well.

adrianreber:
It is easy to test. This will fail:
```
# mpirun --allow-run-as-root -np 4 unshare --map-root-user --user /tmp/ring
```

This will succeed:
```
# mpirun --mca btl_vader_single_copy_mechanism none --allow-run-as-root -np 4 unshare --map-root-user --user /tmp/ring
```

It needs a kernel with user namespace support. I am testing it on RHEL8.

With the changes from this PR it will just work even if `none` is not explicitly set. Thanks for all your help getting this merged.

adrianreber:
Thanks for the pointer to the charliecloud documentation. It seems it explicitly lists the problem this PR tries to solve: https://hpc.github.io/charliecloud/faq.html#communication-between-ranks-on-the-same-node-fails

Nice to see that this does not just fix a problem I was seeing.

hjelmn:
@adrianreber Thanks for the test commands. Will give it a try with xpmem and make sure everything works.

hppritcha:
@adrianreber well it seems xpmem is busted on master, at least on a cray system.  Here's what I see with IMB using master or using this PR:

```
#----------------------------------------------------------------
# Benchmarking Reduce 
# #processes = 16 
#----------------------------------------------------------------
       #bytes #repetitions  t_min[usec]  t_max[usec]  t_avg[usec]
            0         1000         0.04         0.05         0.05
            4         1000         2.75         2.76         2.76
            8         1000         2.70         2.70         2.70
           16         1000         2.89         2.90         2.89
           32         1000         2.91         2.92         2.91
           64         1000         3.03         3.04         3.04
          128         1000         3.37         3.37         3.37
          256         1000         3.83         3.84         3.84
          512         1000         5.30         5.31         5.31
         1024         1000         5.84         5.85         5.84
         2048         1000         9.43         9.45         9.44
         4096         1000        17.37        17.39        17.38
         8192         1000        33.41        33.46        33.44
        16384         1000        71.08        71.17        71.13
        32768         1000       130.17       130.22       130.20
        65536          640       230.39       230.47       230.44
       131072          320       429.11       429.30       429.24
       262144          160       243.06       243.70       243.39
       524288           80       642.52       645.94       644.46
      1048576           40      1070.65      1077.46      1074.49
[nid00031:51917] *** Process received signal ***
[nid00031:51917] Signal: Segmentation fault (11)
[nid00031:51917] Signal code: Address not mapped (1)
[nid00031:51917] Failing at address: 0x2aaad240c010
[nid00031:51917] [ 0] /lib64/libpthread.so.0(+0x10c10)[0x2aaaab008c10]
[nid00031:51917] [ 1] /lib64/libc.so.6(+0x12c596)[0x2aaaab341596]
[nid00031:51917] [ 2] /usr/projects/hpctools/hpp/ompi/install_pr6844_m/lib/openmpi/mca_btl_vader.so(mca_btl_vader_get_xpmem+0x5f)[0x2aaabc8c743f]
[nid00031:51917] [ 3] /usr/projects/hpctools/hpp/ompi/install_pr6844_m/lib/openmpi/mca_pml_ob1.so(mca_pml_ob1_recv_request_get_frag+0x70)[0x2aaabe2f9c10]
[nid00031:51917] [ 4] /usr/projects/hpctools/hpp/ompi/install_pr6844_m/lib/openmpi/mca_pml_ob1.so(mca_pml_ob1_recv_request_progress_rget+0x39b)[0x2aaabe2fa03b]
```
and similar errors.  Not deterministic though.  CMA works nominally.

adrianreber:
@hppritcha Can you share the command-line used to start this test?

hppritcha:
build IMB-MPI1 using Open MPI  master then
mpirun -np 16 IMB-MPI1

jsquyres:
Do you know if XPMEM is broken due _to this PR_?  Or was it already broken on master?

rhc54:
Per @hppritcha note: _using master or using this PR_

hppritcha:
appears to have been broken already on master.

jsquyres:
> Per @hppritcha note: _using master or using this PR_

Heh.  Missed that.  Ok.

How do we move forward with this PR then?  Should we just go ahead and merge it?  Or do we need to wait for an XPMEM fix (independent of this PR) first?

hppritcha:
I'd go ahead and merge it.  But wait, have you actually seen what the output looks like if this feature kicks in?
I've asked charliecloud people to post what they see.  I'd suggest we wait till that is posted, you may not like it @jsquyres

jsquyres:
@hppritcha No problem with waiting for more data -- I just wanted to make sure this PR didn't get dropped / forgotten again...

adrianreber:
Ah, good to know, I though my change broke xpmem. So xpmem is also not working on master. I was already looking at code.

Definitely interested to hear if this also works for charliecloud.

heasterday:
Hello all I am one of the Charliecloud folks @hppritcha mentioned. The fallback seems to work as intended so all looks good on that end. However, when "OMPI_MCA_btl_vader_single_copy_mechanism=cma" is specified the application will print the fatal error and continue to run.
```
ERROR: Linux kernel CMA support was requested via the
btl_vader_single_copy_mechanism MCA variable, but CMA support is
not available due to different user namespaces.

Your MPI job will abort now. Please select another value for
btl_vader_single_copy_mechanism.

  Local host: ko034
#------------------------------------------------------------
#    Intel (R) MPI Benchmarks 2018 Update 1, MPI-1 part    
#------------------------------------------------------------
# Date                  : Thu Sep  5 15:04:04 2019
# Machine               : x86_64
# System                : Linux
<application output clipped>
``` 
The resulting performance is what I would get if I specified ```OMPI_MCA_btl=tcp,self```. I'm not sure how often anyone would do this but printing a fatal error and then continuing to run seems unintended.

heasterday:
Thank you to @adrianreber for working on this, this will be nice functionality to have :)

adrianreber:
I have also seen that the `ring` example I have been using continues to run if a fatal error happens. I never really looked but I thought that it just ignores the error it gets from Open MPI. In `ring.c` there are MPI_Send and MPI_Recv without ever checking the return code. (Actually not sure if that is necessary/correct, but that is what I would expect).

rhc54:
OMPI policy is to indeed abort if the user explicitly requests something we cannot do - so in the case cited by @heasterday, we should definitely abort.

adrianreber:
> OMPI policy is to indeed abort if the user explicitly requests something we cannot do - so in the case cited by @heasterday, we should definitely abort.

Is returning OPAL_ERROR not enough to abort? OPAL_ERROR is what I saw at all the places I looked how to handle an error. What is the right way to abort?

rhc54:
Well, looks pretty simple - from a different btl:
```c
void mca_btl_ofi_exit(void)
{
    BTL_ERROR(("BTL OFI will now abort."));
    exit(1);
}
```

adrianreber:
Okay, will add `exit()` for the abort cases.

rhc54:
Thx!

jsquyres:
A better example would be to look at the usNIC BTL utility function to exit upon error -- it will call the PML callback, if it exists / has been set:

https://github.com/open-mpi/ompi/blob/50db08527266f21743b0083d535a78f26584233b/opal/mca/btl/usnic/btl_usnic_util.c#L27-L62

adrianreber:
@jsquyres something like that?
```
diff --git a/opal/mca/btl/vader/btl_vader_module.c b/opal/mca/btl/vader/btl_vader_module.c
index 55b4726340..17381ba17d 100644
--- a/opal/mca/btl/vader/btl_vader_module.c
+++ b/opal/mca/btl/vader/btl_vader_module.c
@@ -80,6 +80,30 @@ mca_btl_vader_t mca_btl_vader = {
     }
 };
 
+// Exit function copied from btl_usnic_util.c
+// The following comment tells Coverity that this function does not return.
+// See https://scan.coverity.com/tune.
+
+/* coverity[+kill] */
+static void vader_btl_exit(mca_btl_vader_t *btl)
+{
+    if (NULL == btl) {
+        fprintf(stderr, "*** The Open MPI vader BTL is aborting the MPI job (via exit(3)).\n");
+        fflush(stderr);
+        exit(1);
+    }
+
+    if (NULL != btl->error_cb) {
+        btl->error_cb(&btl->super, MCA_BTL_ERROR_FLAGS_FATAL,
+                      (opal_proc_t*) opal_proc_local_get(),
+                      "The vader BTL is aborting the MPI job (via PML error callback).");
+    }
+
+    /* If the PML error callback returns (or if there wasn't one),
+       just exit.  Shrug. */
+    exit(1);
+}
+
 static int vader_btl_first_time_init(mca_btl_vader_t *vader_btl, int n)
 {
     mca_btl_vader_component_t *component = &mca_btl_vader_component;
@@ -236,7 +260,7 @@ static int init_vader_endpoint (struct mca_btl_base_endpoint_t *ep, struct opal_
                         /* If CMA has been explicitly selected we want to error out */
                         opal_show_help("help-btl-vader.txt", "cma-different-user-namespace-error",
                                        true, opal_process_info.nodename);
-                        return OPAL_ERROR;
+                        vader_btl_exit(&mca_btl_vader);
                     }
                     /*
                      * If CMA has been selected because it is the default or
```

jsquyres:
@adrianreber Minor suggestion (and I should probably do something like this to the usnic BTL, too...):

```c
static void vader_btl_exit(mca_btl_vader_t *btl)
{
    if (NULL != btl && NULL != btl->error_cb) {
        btl->error_cb(&btl->super, MCA_BTL_ERROR_FLAGS_FATAL,
                      (opal_proc_t*) opal_proc_local_get(),
                      "The vader BTL is aborting the MPI job (via PML error callback).");
    }

    /* If the PML error callback returns (or if there wasn't one),
       just exit.  Shrug. */
    fprintf(stderr, "*** The Open MPI vader BTL is aborting the MPI job (via exit(3)).\n");
    fflush(stderr);
    exit(1);
}
```

Also, github pro tip: if you put the word "patch" after the 3 tick marks for the verbatim section, github syntax highlights it as a diff.  Similarly, putting "c" after the 3 tick marks syntax highlights it as C code. ...etc.

adrianreber:
@jsquyres Thanks for the help.

I added an exit function and I am calling it now in case `cma` has been explicitly requested by the user.

hppritcha:
@jsquyres could you double check and see if these are the changes you requested?

adrianreber:
Is this ready or does it need additional changes?

jsquyres:
@hjelmn Are you cool with everything in this PR?

PR #6814, opened 2019-07-13
COLL/TUNED: Update alltoall selection rule for mellanox platform

brminich:
Use linear with sync alltoall algorithm for certain message/comm size ranges. Does not affect default fixed decision, unless HPCX (with its custom parameters) is used or corresponding mca is set.

Signed-off-by: Mikhail Brinskii <mikhailb@mellanox.com>

PR #6845, opened 2019-07-26
COLL/TUNED: Update alltoall selection rule for mellanox platform - v4.0.x

brminich:
Use linear with sync alltoall algorithm for certain message/comm size ranges. Does not affect default fixed decision, unless HPCX (with its custom parameters) is used or corresponding mca is set.

Signed-off-by: Mikhail Brinskii mikhailb@mellanox.com

Port from #6814


brminich:
@yosefe, ok to merge?

PR #6849, opened 2019-07-26
opal/util: Include net/if.h to prevent macro mismatch

wckzhang:
In cases where opal/util/if.h is included but net/if.h is not,
IF_NAMESIZE will be 32. If net/if.h is included on Linux systems,
IF_NAMESIZE will be 16. This can cause a mismatch when using the same
macro macro on a system. To avoid this error case, we include this file
in util/if.h.

### DIFF
--- a/opal/util/if.h
+++ b/opal/util/if.h
@@ -13,6 +13,7 @@
  *                         reserved.
  * Copyright (c) 2008      Sun Microsystems, Inc.  All rights reserved.
  * Copyright (c) 2013 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2019      Amazon.com, Inc. or its affiliates.  All Rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -35,6 +36,9 @@
 #endif
 #ifdef HAVE_NETINET_IN_H
 #include <netinet/in.h>
+#endif
+#ifdef HAVE_NET_IF_H
+#include <net/if.h>
 #endif
 
 #ifndef IF_NAMESIZE


rhc54:
I could swear we recently had a debate about namesize and the value we should use to ensure that our array was at least as large as whatever the system used. I thought we resolved it be picking a size that was bigger than anything we might encounter (in Linux or whatever OS flavor we encounter). Is there some reason that fix wasn't adequate (I can't immediately find the PR where all that was discussed)?

bwbarrett:
@rhc54 if we had called our internal constant `OPAL_IF_NAMESIZE` and made it bigger than any usage of `IF_NAMESIZE`, that would have been fine.  But we called it `IF_NAMESIZE`, and then in `opal/util/if.h` conditionally set its size, if it wasn't already set.  So inside a single build of Open MPI, depending on whether or not `net/if.h` had already been included, the `IF_NAMESIZE` constant could either be set to 16 or to 32.  And there is one code path in some code William is writing to clean up the TCP BTL, where that was happening, such that two functions (each in a different .c file) had different ideas of the size of a structure containing a `char name[IF_NAMESIZE]`.

rhc54:
Just doing a quick grep, it looks like `IF_NAMESIZE` is only used in a handful of places. Would it be better to just rename to `OPAL_IF_NAMESIZE` as you suggest and avoid this problem altogether? I'm willing to do that for PMIx as we also have this problem (since we copied the code from here).

Not trying to be a pain - just wondering if we should avoid chasing the setting of that value across the various systems. For example, I have no idea what value Solaris might use - but I bet it is less than 256 bytes.

wckzhang:
Since I don't know what the original reasoning for reusing the IF_NAMESIZE macro was, I chose to take a least impact fix. In my opinion, it should have probably been OPAL_IF_NAMESIZE from the beginning.

rhc54:
I'm not sure there was any reasoning, to be honest - I'd suggest just go ahead and "make it right" by replacing it with `OPAL_IF_NAMESIZE`. I will shortly post a PR with the changes in PMIx, so you just need to handle the OPAL ones.

ggouaillardet:
Would it be easier to use `IF_NAMESIZE` from `net/if.h` (e.g. remote it from `opal/util/if.h`) and add the required #include wherever needed ?
If not, should we use `OPAL_IF_NAMESIZE` everywhere and have its size set at `configure` time ? That will avoid some future issues if the value in `net/if.h` become larger than our own `OPAL_IF_NAMESIZE`

bwbarrett:
@ggouaillardet the challenge with including net/if.h everywhere is that we clearly believed that some platform didn't have a net/if.h and specify a IF_NAMESIZE, since we added compatibility code.  I personally think we should be more careful about redefining system constants, for exactly the problem we ran into here.

Much better to rip out all our uses of IF_NAMESIZE and use OPAL_IF_NAMESIZE directly.  We can either set OPAL_IF_NAMESIZE to IF_NAMESIZE at configure time or  leave it at 32 with a configure-time check.

wckzhang:
> I'm not sure there was any reasoning, to be honest - I'd suggest just go ahead and "make it right" by replacing it with `OPAL_IF_NAMESIZE`. I will shortly post a PR with the changes in PMIx, so you just need to handle the OPAL ones.

Alright, I'll post a PR soon with the macro change

wckzhang:
Replacing this PR with #6852

Issue #6853, opened 2019-07-29
mmap infinite recurse in opal/mca/memory/patcher


hkuno:
## Background information
On certain systems, __mmap and mmap point to the same function.

For example:

```
$ objdump -T  /lib/x86_64-linux-gnu/libc-2.27.so | grep mmap
000000000011b9d0  w   DF .text  00000000000000e0  GLIBC_2.2.5 mmap64
000000000011b9d0 g    DF .text  00000000000000e0  GLIBC_PRIVATE __mmap
000000000011b9d0  w   DF .text  00000000000000e0  GLIBC_2.2.5 mmap
```

This causes `intercept_mmap` to infinitely recurse when `HAVE___mmap` is defined because `intercept_mmap` directly calls `__mmap`:
```c
#ifdef HAVE___MMAP
    /* the darwin syscall returns an int not a long so call the underlying __mmap function */        
    result = __mmap (start, length, prot, flags, fd, offset);
#else       
    result = (void*)(intptr_t) memory_patcher_syscall(SYS_mmap, start, length, prot, flags, fd, offset);
#endif  
```

However, memory_patcher_component.c includes a comment that says that `intercept_mmap` calls `__mmap` in order to support darwin/Apple's internal `mmap` function, and no other code appears to use `HAVE___MMAP`.  Configure sets `opal_memory_patcher_happy=no` for MacOS/Darwin, so modifying `opal/mca/memory/patcher/configure.m4` to not define `HAVE___MMAP` should have no negative impact and would also prevent the infinite recursion.

Alternatively, perhaps the lines in configure.m4 and in memory_patcher_component.c could be removed instead of commented out.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
master, top of tree: 3c45542c51a62ff05b50a05e05ec4a0eb2129440
Note that the issue started appearing at:  https://github.com/open-mpi/ompi/pull/6531

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
`git clone https://github.com/open-mpi/ompi.git`
                                                                                                                          
### Please describe the system on which you are running

* Operating system/version:  Ubuntu bionic (4.15.0-51-generic #55-Ubuntu SMP Wed May 15 14:27:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux) 

* libc version: 2.26 and later
[...]

PR #6855, opened 2019-07-30
Fix mmap infinite recurse in memory patcher

hkuno:
This commit fixes issue #6853 by commenting out lines in configure.m4
that define HAVE___MMAP, avoiding infinite recursion.
This commit also opportunistically converts tabs to spaces.


opal/mca/memory/patcher/configure.m4
```diff
@@ -49,11 +49,15 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
...
-	AC_CHECK_FUNCS([__mmap])
-	AC_CHECK_DECLS([__syscall], [], [], [#include <sys/syscall.h>])
-	AC_CHECK_FUNCS([__syscall])
+    AC_CHECK_HEADERS([linux/mman.h sys/syscall.h])
+# Above sets opal_memory_patcher_happy=no for MacOS/Darwin.
+# Below defines HAVE___mmap, which is only used to support MacOS/Darwin.
+# See issue #6853: on certain systems, mmap will infinitely recurse.
+# if HAVE___mmap is defined.
+#    AC_CHECK_DECLS([__mmap], [], [], [#include <sys/mman.h>])
+#    AC_CHECK_FUNCS([__mmap])
```
jsquyres: Per #6853, we know that `__mmap` is defined on Linux systems with a new enough glibc.  If we know for sure that we do not ever need to intercept `__mmap` on Linux, then I'd just delete the two `AC_*` lines here (vs. commenting them out) and amend the comment to explain the situation / history.  Perhaps something like:

> Glibc version X.Y and later have `__mmap`, but the symbol has the same value as `mmap`.  So we only need to intercept one or the other, not both.  We used to intercept `__mmap` on MacOS/Darwin, but per the above logic, this component no longer supports MacOS/Darwin.

hkuno: @jsquyres : Not understanding whether or not  MacOS/Darwin would ever need to intercept __mmap was why we commented out the lines instead of deleting them. Maybe we should leave them commented out for now, then? Or else use the more conservative fix that you suggested in comments for #6853 ?
@hjelmn Thank you for catching my bad indentation changes, Nathan!

hjelmn: Just to be clear. We were not intercepting __mmap. On macOS syscall does nothing so for testing the patcher I added the check for __mmap so it could be called from the intercept function. We do not need to support patcher on macOS so the code for __mmap can be safely removed.

jsquyres: @hjelmn See https://github.com/open-mpi/ompi/issues/6853#issuecomment-516853917

hkuno: Thank you both for clarifying. I have removed the lines from both configure.m4 and memory_patcher_component.c .

jsquyres:
@hppritcha Cray CI's git seems to have gone off the rails -- can you please fix?

```
Caused by: hudson.plugins.git.GitException: Command "git init /global/homes/h/hpp/jenkins_cori/workspace/open-mpi.build.platforms/Platform/cray_cle/src" returned status code 127:
stdout: 
stderr: git: error while loading shared libraries: libcrypto.so.0.9.8: cannot open shared object file: No such file or directory

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2042)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2010)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2006)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommand(CliGitAPIImpl.java:1638)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$5.execute(CliGitAPIImpl.java:811)
	... 12 more
```

jsquyres:
@hkuno Can you squash down to one commit?  Then this PR will be good to go.

hppritcha:
had to fix up some things at nersc after upgrade to CLE 7
bot:ompi:retest

hkuno:
> @hkuno Can you squash down to one commit? Then this PR will be good to go.

Sure!  I have done that. (Please let me know if the commit message is too terse, and I'll add more detail.)  Thank you!

hppritcha:
well there was another thing with our jenkins slave prepper installing an old autotools package. try again.

bot:ompi:retest

PR #6872, opened 2019-08-07
Fix typos

rhc54:
Provide a missing header and paren

Thanks to @zerothi for the assistance


rhc54:
Verified by @zerothi

Issue #6631, opened 2019-05-02
LSF affinity files override command line arguments


zerothi:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
3.1.3

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

- GCC 8.2.0
- Configure command line (from distribution tarball) (from `ompi_info`):
```bash
'CC=gcc' 'CXX=g++' 'FC=gfortran' '--prefix=/appl/gcc/8.2.0/openmpi/3.1.3-lsf10b' '--with-lsf' '--without-tm' '--enable-mpi-cxx' '--enable-cxx-exceptions' '--with-hwloc=/appl/gcc/8.2.0/openmpi/3.1.3-lsf10b' '--with-threads=posix' '--enable-orterun-prefix-by-default'
```

### Please describe the system on which you are running

* Operating system/version: 
ScientificLinux 7.3 (should be unrelated)
* Computer hardware: 
XeonE5-2650v4, 2 sockets, 12 cores per socket.
* Network type: 
Unrelated

-----------------------------

## Details of the problem

The problem is that starting a program with explicit command line arguments detailing mapping and binding is not enforced when the `LSB_AFFINITY_HOSTFILE` is set.
This is a bit surprising to new users and I would suggest that command line has precedence.

I'll give a few examples and the resulting binding:

- LSF script:
```bash
#BSUB -n 24
export OMP_NUM_THREADS=6
mpirun -n 4 --map-by ppr:2:socket,PE=$OMP_NUM_THREADS -x OMP_NUM_THREADS --report-bindings ./run.sh
```
And I get this binding:
```bash
[n-62-21-38:19485] MCW rank 0 bound to socket 0[core 0[hwt 0]]: [B/././././././././././.][./././././././././././.]
[n-62-21-38:19485] MCW rank 1 bound to socket 0[core 1[hwt 0]]: [./B/./././././././././.][./././././././././././.]
[n-62-21-38:19485] MCW rank 2 bound to socket 0[core 2[hwt 0]]: [././B/././././././././.][./././././././././././.]
[n-62-21-38:19485] MCW rank 3 bound to socket 0[core 3[hwt 0]]: [./././B/./././././././.][./././././././././././.]
```
which is wrong. However, with
```bash
#BSUB -n 24
unset LSB_AFFINITY_HOSTFILE
[...]

PR #6873, opened 2019-08-07
v3.1.x: Allow individual jobs to set map policy, override LSF bindfile

rhc54:
Refs #6631


rhc54:
Verified by @zerothi

jsquyres:
@bwbarrett I think I can squint at this and call it a bug fix.  Thoughts?

bwbarrett:
I think I agree that this is a bugfix.

PR #6884, opened 2019-08-09
Reintroduce common ofi

guserav:
In discussions with @hppritcha it turned out that we want to have the common/ofi component back.
This PR is intended to be the first step into the direction of a common/ofi component that enables common code usage across components that use ofi.

### DIFF
--- a/opal/mca/common/ofi/Makefile.am
+++ b/opal/mca/common/ofi/Makefile.am
@@ -1,0 +1,105 @@
+#
+# Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+#                         University Research and Technology
+#                         Corporation.  All rights reserved.
+# Copyright (c) 2004-2013 The University of Tennessee and The University
+#                         of Tennessee Research Foundation.  All rights
+#                         reserved.
+# Copyright (c) 2004-2009 High Performance Computing Center Stuttgart,
+#                         University of Stuttgart.  All rights reserved.
+# Copyright (c) 2004-2005 The Regents of the University of California.
+#                         All rights reserved.
+# Copyright (c) 2011-2013 NVIDIA Corporation.  All rights reserved.
+# Copyright (c) 2014      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015      Intel, Inc. All rights reserved.
+# Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
+#                         reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+# A word of explanation...
+#
+# This library is linked against various MCA components because the
+# support for ofis is needed in various places.
+#
+# Note that building this common component statically and linking
+# against other dynamic components is *not* supported!
+
+AM_CPPFLAGS = $(opal_common_ofi_CPPFLAGS)
+
+# Header files
+
+headers = \
+        common_ofi.h
+
+# Source files
+
+sources = \
+        common_ofi.c
+
+# As per above, we'll either have an installable or noinst result.
+# The installable one should follow the same MCA prefix naming rules
+# (i.e., libmca_<type>_<name>.la).  The noinst one can be named
+# whatever it wants, although libmca_<type>_<name>_noinst.la is
+# recommended.
+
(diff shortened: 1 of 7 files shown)


gpaulsen:
ok to test

opal/mca/common/ofi/owner.txt
```diff
@@ -0,0 +1,7 @@
+#
+# owner/status file
+# owner: institution that is responsible for this package
+# status: e.g. active, maintenance, unmaintained
+#
+owner: Intel
```
jsquyres: Is Intel still ok with being listed as the owner / maintainer?

opal/mca/common/ofi/Makefile.am
```diff
@@ -0,0 +1,106 @@
...
+#
+# A word of explanation...
+#
+# This library is linked against various MCA components because the
+# support for ofis is needed in various places.
+#
+# Note that building this common component statically and linking
+# against other dynamic components is *not* supported!
+
+AM_CPPFLAGS = $(opal_common_ofi_CPPFLAGS) -I$(prefix)/include
```
jsquyres: The `-I$(prefix)/include` is weird here -- can you explain what it is for?

Specifically: we should not be relying on anything in the installation tree while building (because the installation tree does not exist yet).

opal/mca/common/ofi/configure.m4
```diff
@@ -18,15 +19,17 @@
...
     # successfully.
-    OPAL_CHECK_OFI([opal_common_ofi],
-                   [opal_common_ofi_happy=yes
-                    common_ofi_WRAPPER_EXTRA_LDFLAGS=$opal_common_ofi_LDFLAGS
-                    common_ofi_WRAPPER_EXTRA_LIBS=$opal_common_ofi_LIBS
-                    $1],
-                   [opal_common_ofi_happy=no
-                    $2])
+    AS_IF([test "$opal_ofi_happy" = "yes"],
+          [opal_common_ofi_happy=yes
```
jsquyres: Is this variable (`opal_common_ofi_happy`) used anywhere?

guserav: Not yet but it supposed to be used in every component that has the common/ofi as a dependency.

jsquyres: I'm thinking that the "revamp" commit intended `opal_ofi_happy` to be used, instead.  I.e., the One Variable to Rule Them All is the `opal_ofi_happy` -- which indicates that OFI configury itself was happy; it doesn't matter whether any individual OFI component's `happy` variable is set.

opal/mca/common/ofi/Makefile.am
guserav: No there is no good reason for it. I have included that when I wasn't understanding the build process and it fixed my problem at the time. But now I see that the problem was actually that opal_common_ofi_CPPFLAGS was renamed to opal_ofi_CPPFLAGS (as off f5e1a672ccd5db127e85e1e8f6bcfeb8a8b04527) and I missed to fix that line...

opal/mca/common/ofi/configure.m4
guserav: I was thinking along the way that if I have a component that depends upon the common/ofi component then that component should check if common/ofi is happy. I know that this check is equivalent to checking opal_ofi_happy for now, but in the future the common component could depend on something new. It would be similar to what https://github.com/open-mpi/ompi/blob/d58c59eb71e4b116030522928ac65caa8a34ed7b/opal/mca/common/ucx/configure.m4#L17-L19 does. Though the naming scheme is different.

On the other hand https://github.com/open-mpi/ompi/blob/d58c59eb71e4b116030522928ac65caa8a34ed7b/opal/mca/common/cuda/configure.m4#L20 and https://github.com/open-mpi/ompi/blob/d58c59eb71e4b116030522928ac65caa8a34ed7b/opal/mca/common/sm/configure.m4#L26 don't seem to set similar happy values.

So I believe that in the future when some new dependency comes along, already having `opal_common_ofi_happy` might be useful. Especially for new people who are not accustomed to the code base, it is a good hint. Though I understand that more code, is more code that you need to maintain. So I could take it out if  you prefer.

PR #6754, opened 2019-06-11
OSC/UCX: Fix early cleanup + race conditions

artpol84:
* Fix early initialization of the worker pool (WP). WP has to remain until the finalization of the OSC component.
* Ensure the atomicity of the OSC/ucx global state updates.

PR #6893, opened 2019-08-12
osc/ucx: Fix error path

janjust:
Missing error path change in: #6754 (6678ac0f557935b291ec2310216b7ea46e0c13b1)

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_component.c
+++ b/ompi/mca/osc/ucx/osc_ucx_component.c
@@ -356,7 +356,7 @@ static int component_select(struct ompi_win_t *win, void **base, size_t size, in
         ret = opal_progress_register(progress_callback);
         if (OMPI_SUCCESS != ret) {
             OSC_UCX_VERBOSE(1, "opal_progress_register failed: %d", ret);
-            goto error;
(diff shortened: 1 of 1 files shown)


ompi/mca/osc/ucx/osc_ucx_component.c
```diff
@@ -356,7 +356,7 @@ static int component_select(struct ompi_win_t *win, void **base, size_t size, in
         ret = opal_progress_register(progress_callback);
         if (OMPI_SUCCESS != ret) {
             OSC_UCX_VERBOSE(1, "opal_progress_register failed: %d", ret);
-            goto error;
+            goto select_unlock;
```
brminich: could just remove this line, you would end up in `select_unlock` anyway

janjust: @brminich :) I thought about doing it, but then I thought, what if code for whatever reason is added in between, this error path may go unchecked. Maybe for clarity I prefer to leave it unless there are strong feelings to remove it and rely on fallthrough.

brminich: ok

PR #6863, opened 2019-08-05
Refresh of the datatype engine from Topic/backport 6695

bosilca:
A backport of the datatype improvements on the 4.0.

Few things to mention:
- I had to pull few other commits to solve some of the conflicts, but the changes are minor and mostly necessary anyway. The 2 commits I imported in addition to the original PR are 42119254 and d141bf7912.
- I had to manually alter one of the patches to work around the atomic types added in 000f9eed4 and that have not been yet backported to the stable branch.

PR #6898, opened 2019-08-14
Fix/convertor raw

bosilca:
Fixes a datatype issue in ompi-test/mpi2basic-tests/filetests/fileview.c reported by Edgar.

In addition to the fix for the reported issue, this patch contains a renaming of the OPAL datatype internal variables used for the datatype debugging infrastructure.

Needs to be included in #6863


jsquyres:
I separated out the unrelated whitespace-only file change into its own commit.  Because that's how I roll.

Issue #6899, opened 2019-08-14
Incorrect behaviour with certain extent values for simple derived datatypes


d7919:
## Background information
I am attempting to construct a moderately complicated derived datatype but have been running into difficulties so have tried to strip back to a minimal case and in doing so I have come across what looks like incorrect behaviour. I have constructed a minimal example which can be accessed in this [gist](https://gist.github.com/d7919/72fca16c38cbc3a1e79b747a3d3509f7). 

I first make a contiguous type corresponding to the first half of the message, next I use `mpi_type_create_resized`  to change the extent, with the intention that I can make the start of the second half of the message overlap the first half. When the extent corresponds to two elements of the base type this works as expected - I see the first two elements of the first half of the message and then the entirety of the second message. If I change the extent to correspond to one element of the base type instead of seeing just the first element of the first half followed by the entirety of the second half we instead get the entire first half of the message followed by the entire second half -- in other words it is as if I hadn't set the extent at all. This test case is only using one processor.

I would have expected the output to look like

```shell
With extent =                      8
           1           2           9          10          11          12
          13          14          15          16          -1          -1
          -1          -1          -1          -1
 With extent =                      4
           1           9          10          11          12          13
          14          15          16          -1          -1          -1
          -1          -1          -1          -1
```

but instead I find

```shell
 With extent =                      8
           1           2           9          10          11          12
          13          14          15          16          -1          -1
          -1          -1          -1          -1
 With extent =                      4
           1           2           3           4           5           6
           7           8           9          10          11          12
          13          14          15          16
```
A version using `mpi_f08` demonstrates identical behaviour. Similarly changing the `base_type` from integer to real, complex etc. all behave in the same way.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
I've tested this with Open MPI v2.1.1 and Open MPI v3.1.1 and get the same output in both cases. These are on systems I don't manage so I'm not 100% sure how they were installed, but I'm fairly confident that the v2.1.1 has been installed from a package manager whilst v3.1.1 was installed from tarball. I've tried using Intel MPI (2018) and I get the expected behaviour for all extents.

### Please describe the system on which you are running

* Operating system/version: tested on both Ubuntu (16.04) and something derived from CetOS Linux 7.
* Computer hardware: Running on various rack mount nodes with two sockets populated with various modern intel server chips. Memory per node > 100 GB.
[...]

PR #6901, opened 2019-08-14
v3.1: Fix a corner case for the datatype extent computation.

bosilca:
It turns out that if a complex datatype built out of a single predefined
type has been resized to the size of its predefined type, we miscompute
the extent when creating other datatypes with the type.

Provides a fix for #6899 for the 3.x branch.


bwbarrett:
@bosilca does this also need to go to 3.0?

jsquyres:
@bwbarrett The commit cherry-picked cleanly to the v3.0.x branch, so I took the liberty of creating the v3.0.x PR for this: https://github.com/open-mpi/ompi/pull/6913

PR #6904, opened 2019-08-15
Conform MPIR_Breakpoint to MPIR standard. 

awlauria:
- Fix MPIR_Breakpoint standard violation by returning void
  instead of a void*.

### DIFF
--- a/orte/orted/orted_submit.c
+++ b/orte/orted/orted_submit.c
@@ -174,7 +174,7 @@ char MPIR_attach_fifo[MPIR_MAX_PATH_LENGTH] = {0};
 int MPIR_force_to_main = 0;
 static void orte_debugger_init_before_spawn(orte_job_t *jdata);
 
-ORTE_DECLSPEC void* __opal_attribute_optnone__ MPIR_Breakpoint(void);
+ORTE_DECLSPEC void __opal_attribute_optnone__ MPIR_Breakpoint(void);
 
 /* 
  * Attempt to prevent the compiler from optimizing out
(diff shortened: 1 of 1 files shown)


jsquyres:
bot:lanl:retest

jsquyres:
@awlauria I do not understand the commit message:

> Return a volatile void * instead of casting ...

What casting are you referring to?

awlauria:
@jsquyres the other alternative would be casting the return value to a void *. IE:
`return (void*)noop_mpir_breakpoint_ptr;`

I figured changing the function to return a volatile void * is safer so we don't run into it possibly being compiled out again.

jsquyres:
Hmm.

On the one hand, this is probably fine.

But on the other hand, [I just checked the MPIR spec](https://www.mpi-forum.org/docs/mpir-specification-03-01-2018.pdf), and `MPIR_Breakpoint()` is supposed to return `void` (not `void*`, nor `volatile void*`).

I wonder if a better approach would be to assign something to a global volatile variable (rather than try to return something from a function that is supposed to return `void`).

awlauria:
Yeah, I noticed that as well when working the last PR fix. That's why I figured that making this return volatile wouldn't be a big deal, since it already *technically* violates the standard. That said, I don't think returning non-void actually changes the functionality of MPI_Breakpoint at all, since all users do is break on it. The return value is just thrown away.

Either approach is fine with me, but we know this one *seems* to work, so it might be safer to just leave as is?

jsquyres:
> That's why I figured that making this return volatile wouldn't be a big deal, since it already _technically_ violates the standard. That said, I don't think returning non-void actually changes the functionality of MPI_Breakpoint at all, since all users do is break on it. The return value is just thrown away.

Yup; that's precisely why I said "One the one hand, this is probably fine."

But that being said, we should adhere to the spec.  Who knows what unintended consequences could occur someday if we don't adhere to the spec?

I know it's a little more testing on your part, but I can't think of a way for a compiler to optimize away a call to a function that assigns something to a public global volatile variable.  For extra double bonus points, you could even read the global volatile variable in `MPIR_Breakpoint()` and conditionally assign back to that same global volatile variable.  😉

jsquyres:
@awlauria Are you going to fix this?

awlauria:
Thanks for the poke, completely forgot about this. I'll have a patch tomorrow.

awlauria:
@jsquyres this is updated.

jsquyres:
Sorry to be nit-picky -- I think your commit message is a bit misleading.  You're not silencing a warning (where "silence" usually has the connotation of a workaround vs. a real fix).  Instead, this commit brings us back into conformance with the MPIR spec, and has a side effect of avoiding the situation where the warnings would have occurred in the first place.

awlauria:
I could have sworn someone reported the warning where this was brought up in the first place - and indeed gcc showed a warning for me when I rebuild ompi/master base.

But that said I have no problems changing the commit message/comment.

awlauria:
Updated the message.

jsquyres:
Yeah, sorry, this is probably to-MAY-to / to-MAH-to...

My only $0.02 here is that the primary cause for this commit is to bring us into compliance with the MPIR spec.  Fixing the warning is a side effect (and wouldn't have ever been there if we had been in compliance with the MPIR spec).

jsquyres:
I do not recall offhand -- does this need to be cherry-picked to any of the release branches?

rhc54:
Hmmm...if "noop_mpir_breakpoint_ptr" is a global _public_ symbol, then shouldn't you at least have prefixed it with "orte" or something?

jsquyres:
Blarg, yeah, we probably should have followed the prefix rule here.

That being said, I think library symbol visibility is enabled by default, so we should probably be ok.

awlauria:
The original MPIR fix that went in to produce the warning went into 3.0.x, 3.1.x and 4.0.x if I remember right. Of course the violation standard (returning a void *) has been around much longer.

PR #6895, opened 2019-08-12
Run-as-root env vars in orterun.c

simonbyrne:
I found that I needed to apply the same change as #5597/#5598 to orterun.c for the environment variables to work correctly.

PR #6910, opened 2019-08-19
orterun: remove duplicate code

jsquyres:
https://github.com/open-mpi/ompi/pull/6895 fixed the code in orterun.c
to allow running as root if both OMPI_ALLOW_RUN_AS_ROOT and
OMPI_ALLOW_RUN_AS_ROOT_CONFIRM env vars are set.  However, this
env-var-checking code already exists in
orte_submit.c:orte_submit_init() -- it looks like the
geteuid()/getenv()-checking code here in orterun is now duplicate
code.

So let's just get rid of the duplicate code.

### DIFF
--- a/orte/tools/orterun/orterun.c
+++ b/orte/tools/orterun/orterun.c
@@ -137,37 +137,6 @@ int orterun(int argc, char *argv[])
         exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


jsquyres:
Refs #6895

PR #6930, opened 2019-08-27
Coll/hcoll: fixes hcoll non-blocking colls support

vspetrov:
open-mpi/ompi@0fe756d4166eecf2f0ee2598da690c69a7c824c4 Introduced
    a bug in coll/hcoll component. The ompi_requests allocated by
    libhcoll would be treated as coll_base_nbc_request during
    ompi_coll_base_retain_<> call. Afterwards this would lead to a
    segv in the request cleanup.

    Fix: since libhcoll interface does not distinguish between the
    blocling/non-blocking requests use coll_base_nbc_request all the
    time and initialize it properly in
    coll/hcoll/get_coll_handle(). It is still within 2 cache lines.


amaslenn:
@gpapaure found in v4.0.x branch

vspetrov:
v4.0.x PR https://github.com/open-mpi/ompi/pull/6935

PR #6933, opened 2019-08-27
UCX osc: properly release exclusive lock to avoid lockup

### LINKED ISSUE #6931: UCX osc: mixing exclusive and shared locks leads to lockup
Looking at the code in `osc_ucx_passive_target.c` I found that the implementation of `end_exclusive` is flawed, leading to a lock-up if ranks try to take a shared lock while one rank holds an exclusive lock on the same target. Simply replacing the value in the lock's memory with `TARGET_LOCK_UNLOCKED` overwrites any changes to that value made by other processes trying to acquire a shared lock, causing the lock to get out of sync. Instead, the value of `TARGET_LOCK_EXCLUSIVE` should be subtracted from the lock to release it and to not interfere with the attempts of other ranks.

While debugging, I also found that some of the asserts in these code paths are overly strict and trigger even when they should not.

I will post PRs for `master`, `v4.0.x`, and `v3.1.x` soon.

Potentially related: #6549

devreal:
Fixes #6931 

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_passive_target.c
+++ b/ompi/mca/osc/ucx/osc_ucx_passive_target.c
@@ -76,21 +76,10 @@ static inline int start_exclusive(ompi_osc_ucx_module_t *module, int target) {
 }
 
 static inline int end_exclusive(ompi_osc_ucx_module_t *module, int target) {
-    uint64_t result_value = 0;
     uint64_t remote_addr = (module->state_addrs)[target] + OSC_UCX_STATE_LOCK_OFFSET;
-    int ret = OMPI_SUCCESS;
-
-    ret = opal_common_ucx_wpmem_fetch(module->state_mem,
-                                    UCP_ATOMIC_FETCH_OP_SWAP, TARGET_LOCK_UNLOCKED,
-                                    target, &result_value, sizeof(result_value),
-                                    remote_addr);
-    if (OMPI_SUCCESS != ret) {
-        return ret;
-    }
-
-    assert(result_value >= TARGET_LOCK_EXCLUSIVE);
-
-    return ret;
+    return opal_common_ucx_wpmem_post(module->state_mem, UCP_ATOMIC_POST_OP_ADD,
+                                      -((int64_t)TARGET_LOCK_EXCLUSIVE), target,
+                                      sizeof(uint64_t), remote_addr);
 }
 
 int ompi_osc_ucx_lock(int lock_type, int target, int assert, struct ompi_win_t *win) {


ompi/mca/osc/ucx/osc_ucx_passive_target.c
```diff
@@ -76,21 +76,10 @@ static inline int start_exclusive(ompi_osc_ucx_module_t *module, int target) {
 }
 
 static inline int end_exclusive(ompi_osc_ucx_module_t *module, int target) {
-    uint64_t result_value = 0;
     uint64_t remote_addr = (module->state_addrs)[target] + OSC_UCX_STATE_LOCK_OFFSET;
-    int ret = OMPI_SUCCESS;
-
-    ret = opal_common_ucx_wpmem_fetch(module->state_mem,
-                                    UCP_ATOMIC_FETCH_OP_SWAP, TARGET_LOCK_UNLOCKED,
```
brminich: can you please elaborate a bit about the hang? If we set `TARGET_LOCK_UNLOCKED` (which is zero), `start_shared` should succeed, because `result_value` is supposed to be less than `TARGET_LOCK_EXCLUSIVE`

devreal: Assume that process 0 holds an exclusive lock and process 1 tries to acquire a shared lock. Process 1 will retrieve the lock value (which is `TARGET_LOCK_EXCLUSIVE`) and add `1` it, making it `TARGET_LOCK_EXCLUSIVE + 1`. In the meantime, however, process 0 releases the lock by resetting it to `TARGET_LOCK_UNLOCKED` (which is 0). Since Process 1 has seen the lock as being taken exclusively, it will subtract `1` again, leading to a value of `-1` in the lock variable (the lock is thus out of sync). 

The next time Process 1 tries to take a shared lock, it will get the value -1 in a variable of type `uint64_t`, which is definitely `>=TARGET_LOCK_EXCLUSIVE`.

I realized, though, the my patch used `-((int64_t)TARGET_LOCK_UNLOCKED)` instead of `-((int64_t)TARGET_LOCK_EXCLUSIVE)`. That should be fixed now. Sorry if that led to confusion.

brminich: I see now, thanks

artpol84:
@hoopoepg @brminich, please keep me in the loop on all OSC issues.

artpol84:
@devreal thanks for the fix.
@janjust FYI

awlauria:
I notice this was cherry-picked to the 3.1.x branch. Do we want this for the 4.0 branches?

awlauria:
4.0.x: #6934

PR #6934, opened 2019-08-27
UCX osc: properly release exclusive lock to avoid lockup (v4.0.x)

devreal:
Cherry-pick of #6933 to v4.0.x

See #6931 for the issue this fixes.

(cherry picked from commit 08cb6389e034c1a70368671f745f20904c774a1e)


artpol84:
@devreal @hoopoepg @brminich I see that this PR changes abstracted use of UCX to a direct call.
Is it absolutely necessary? I'd prefer to keep abstraction.

artpol84:
Ok, I confused this with master. It is OK for it to be this way in 4.0.x

PR #6939, opened 2019-08-28
fortran: generate bindings from templates

ggouaillardet:
make the code less error prone by generating mpif-h bindings
for
 - blocking collective
 - non blocking collective
 - persistent collective
from a single template file


ggouaillardet:
@jsquyre could you please have a look at this PR ?

I was fixing some `mpif-h` bindings when I found a lot of code is duplicated in three file (e.g. `bcast_f.c`, `ibcast_f.c` and `bcast_init_f.c`) and we can simplify this by using some kind of template.

I am not fluent in perl nor python, so I simply implemented it in good old `sh`.

lanl-ompi:

jsquyres:
This is a very interesting concept.  The current form doesn't look too scalable, though -- it has a lot of particulars for these two functions.  What I mean is: how much would we gain, subjectively, if we replaced all of our current Fortran bindings C code with templated code like this?  I think the templated code is a bit harder to read / maintain.

That being said, I'm a huge believer in generating code -- I think that _someday_ we should really generate _all_ of the Fortran bindings for Open MPI.  

I think the real question is: paired with a JSON file of all the bindings from the MPI Forum itself, is there a generic way to take any given Fortran binding and map it to C in a generic way?  If the answer is "yes" -- or even close to "yes" -- this would allow us to fully generate the Fortran bindings.

ggouaillardet:
@jsquyres I started looking at how to automatically generate the Fortran bindings, based on a previous effort of yours, at https://github.com/ggouaillardet/ompi/tree/poc/new_fortran_bindings
(yep, that was 5+ years ago!)

IIRC, 90% can be achieved with a modest effort, and then it gets painful to complete the remaining 10%.

Anyway, I can try to resume that based on the JSON file instead of an ad-hoc perl abstraction of the MPI standard.

Can you please point me to (or email me privately) the JSON file?

jsquyres:
Sweet.  I sent a private email about the JSON stuff, because it's all very prototype-ish at the moment.

Maybe we can automatically generate that which is easy, and leave hard-coded the remaining 10%/hard stuff.  Honestly, anything would be an improvement -- we don't have to have a 100% solution in the first round.

Issue #6932, opened 2019-08-27
ompi/v3.x.x bug since August 21: opal_datatype_pack.c:203 and opal_datatype_unpack.c:135


ericch1:
Hi,

_EDIT: I modified the mentioned SHAs in this first message since it contains wrong info about the wrong sha_
up to commit d3587f5, everything was fine, but
as of commit 390e0bc, we have some tests that are failing with errors like this:

<pre>
[dockercentos7:18478] opal_datatype_pack.c:203
	Pointer 0xdf6c970 size 9 is outside [0xdf6c880,0xdf6c969] for
	base ptr 0xdf6c880 count 10 and data 
[dockercentos7:18478] Datatype 0xa10f7a0[] size 17 align 8 id 0 length 4 used 3
true_lb 0 true_ub 17 (true_extent 17) lb 0 ub 24 (extent 24)
nbElems 3 loops 0 flags 114 (committed contiguous )-cC----GD--[---][---]
   contain OPAL_INT1:* OPAL_INT8:* OPAL_FLOAT8:* 
--C---P-D--[---][---]    OPAL_FLOAT8 count 1 disp 0x0 (0) blen 1 extent 8 (size 8)
--C---P-D--[---][---]      OPAL_INT8 count 1 disp 0x8 (8) blen 1 extent 8 (size 8)
--C---P-D--[---][---]      OPAL_INT1 count 1 disp 0x10 (16) blen 1 extent 1 (size 1)
-------G---[---][---]    OPAL_LOOP_E prev 3 elements first elem displacement 0 size of data 17
Optimized description 
-cC---P-DB-[---][---]     OPAL_UINT1 count 1 disp 0x0 (0) blen 8 extent 8 (size 8)
-cC---P-DB-[---][---]     OPAL_UINT1 count 1 disp 0x8 (8) blen 9 extent 9 (size 9)
-------G---[---][---]    OPAL_LOOP_E prev 2 elements first elem displacement 0 size of data 17

[dockercentos7:18478] opal_datatype_unpack.c:135
	Pointer 0xeb57a98 size 9 is outside [0xeb579a8,0xeb57a91] for
	base ptr 0xeb579a8 count 10 and data 
[dockercentos7:18478] Datatype 0xa10f7a0[] size 17 align 8 id 0 length 4 used 3
true_lb 0 true_ub 17 (true_extent 17) lb 0 ub 24 (extent 24)
[...]

PR #6945, opened 2019-08-29
Small optimization on the datatype commit.

bosilca:
This patch fixes the merge of contiguous elements into larger but more
compact datatypes, and allows for contiguous elements to have thir
blocklen increasing instead of the count. The idea is to always maximize
the blocklen, aka. the contiguous part of the datatype.

Possible fix for #6932 

### DIFF
--- a/opal/datatype/opal_datatype_optimize.c
+++ b/opal/datatype/opal_datatype_optimize.c
@@ -163,6 +163,12 @@ opal_datatype_optimize_short( opal_datatype_t* pData,
             if( 0 == last.count ) {  /* first data of the datatype */
                 last = *current;
                 continue;  /* next data */
+            } else {  /* can we merge it in order to decrease count */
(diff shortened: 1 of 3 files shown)


bosilca:
I was able to track back the issue by looking at the datatype from issue #6932. It failed to merge into a single predefined element but the datatype was marked as contiguous, allowing the datatype engine to use one of the optimized pack/unpack function (contig_with_gaps). However, having a composed description mislead the contig_with_gaps function to wrongly interpret the shape of the datatype.

Anyway, long story short, this patch fixes the issue and add a test to prevent it from happening in the future. Once merged it should be backported to v3 and v4.

jsquyres:
@bosilca Any particular reason not to merge this PR?

bosilca:
It need to be backported but I don't have time to do it right now. Feel free to do so.

PR #6947, opened 2019-08-30
v3.1.x: regx/naive: add regx/naive component

sam6258:
Signed-off-by: Scott Miller <scott.miller1@ibm.com>
(cherry picked from commit 1b0cfdf264c793bb5f0fe2714cbbdfc6f54b46eb)

### DIFF
--- a/orte/mca/regx/fwd/regx_fwd.c
+++ b/orte/mca/regx/fwd/regx_fwd.c
@@ -313,5 +313,9 @@ static int nidmap_create(opal_pointer_array_t *pool, char **regex)
     free(nodenames);
     free(tmp);
     *regex = tmp2;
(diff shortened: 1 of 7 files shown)


jjhursey:
Thanks for cherry-picking this back to the v3.1.x series.

Note that the v3.0.x series does not have the `regx` framework so this change cannot be cherry-picked back to that release branch. The v3.0.x branch effectively just has the `fwd` component. We do not intend to push a fix for the v3.0.x series since users have two newer release series to move too that will have this functionality. See it as motivation to upgrade :)

gpaulsen:
Note this is cherry-picked from the v4.0.x branch because this framework doesn't exist in master.

PR #6954, opened 2019-09-04
orte/util: fix generate_ppn when multiple apps are used

ggouaillardet:
always reconstruct the bucket after it is unloaded to ensure
we always start an iteration with a correctly initialized empty buffer.

### DIFF
--- a/orte/util/nidmap.c
+++ b/orte/util/nidmap.c
@@ -1078,6 +1078,8 @@ int orte_util_generate_ppn(orte_job_t *jdata,
             }
         }
         opal_dss.unload(&bucket, (void**)&bytes, &nbytes);
+        OBJ_DESTRUCT(&bucket);
+        OBJ_CONSTRUCT(&bucket, opal_buffer_t);
 
         if (opal_compress.compress_block(bytes, (size_t)nbytes,
                                          (uint8_t**)&bo.bytes, &sz)) {


ggouaillardet:

this fixes the crashes with `MPI_Spawn_multiple()` we can see in MTT.

This fix (or a better one) should also be applied to `prrte`.

ggouaillardet:
FWIW, unloading the buffer resets `bytes_used` and `base_ptr` but **not** `bytes_allocated`. so the next time something is packed into the buffer, it crashes because the `base_ptr` is not allocated/extended since we believe there is enough free room (e.g. `bytes_allocated - bytes_used > "the space we need for packing data"`).

I do not know the full rationales, but we might also want to reset `bytes_allocated` when unloading a buffer and/or fix test test (e.g. `base_ptr == NULL`) to decide whether the buffer should be extended.

rhc54:
I believe it was mostly an oversight - when you unload a buffer, the buffer contents become invalid. I have an alternate fix in #6957 that I'd like you to look at.

As for PRRTE, this isn't a problem over there as we use PMIx buffers for packing/unpacking in that region and those unload routines already reset everything in the buffer struct.

ggouaillardet:
The right fix in (just merged) #6957, so I am closing this PR.

Issue #6568, opened 2019-04-04
Issue #6258 still occurring on 4.0.1 - When on OSX hang when message size larger than a few MB.  


steve-ord:
[Thank]() you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

4.0.1 for the release

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

I edited the brew formula to pull the 4.0.1 instead of 4.0.0 and build from source. The compiler flags are those as default for the brew formula for 4.0.0

### Please describe the system on which you are running

* Operating system/version: 
* Computer hardware: 
* Network type: 

OSX Mojave 10.14.3
-----------------------------

## Details of the problem

This is a continuation of ticket #6258 which is now closed and was advised to open a new one I still see exactly the same issue. If I drop the size of the messages the application runs fine. 

I replicated the last test and the magic number for failure is 19145881 bytes - the test code replicated from #6258 is:

```c
#include <stdio.h>

#include "mpi.h"

static const MPI_Datatype Datatype = MPI_PACKED;
static const int Tag = 42;
static const int RecvProc = 0;
static const int SendProc = 1;

// 19145880 does not hang 19145881  does hang in the Wait()s
#define MessageSize (19145880+1)
static unsigned char data[MessageSize] = {0};

int main(int argc, char *argv[])
{
	MPI_Init(&argc, &argv);
	MPI_Comm comm = MPI_COMM_WORLD;

	int myID = 0;
	int nProcs = 1;
	MPI_Comm_size(comm, &nProcs);
	MPI_Comm_rank(comm, &myID);

	if (nProcs != 2)
	{
		if (myID == 0)
			printf("Must be run on 2 procs\n");
		MPI_Finalize();
		return -1;
	}

	int result = 0;
	if (myID == RecvProc)
	{MPI_Status probeStatus;
		result = MPI_Probe(SendProc, MPI_ANY_TAG, comm, &probeStatus);
		printf("[%i] MPI_Probe => %i\n", myID, result);
		int size = 0;
		result = MPI_Get_count(&probeStatus, Datatype, &size);
		printf("[%i] MPI_Get_count => %i, size = %i\n", myID, result, size);

		MPI_Request recvRequest;
		result = MPI_Irecv(data, size, Datatype, SendProc, Tag, comm, &recvRequest);
[...]

PR #6961, opened 2019-09-06
btl/vader: when using single-copy emulation fragment large rdma

hjelmn:
This commit changes how the single-copy emulation in the vader btl
operates. Before this change the BTL set its put and get limits
based on the max send size. After this change the limits are unset
and the put or get operation is fragmented internally.

References #6568

### DIFF
--- a/opal/mca/btl/vader/btl_vader_atomic.c
+++ b/opal/mca/btl/vader/btl_vader_atomic.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2010-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019      Google, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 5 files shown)


hjelmn:
@jsquyres Seems to be working. Ran with local ompi-tests ibm test suite and it is passing. Also passing the reproducer from #6568.

jsquyres:
@hjelmn is this an alternate workaround to the OB1 PUT issue?

hjelmn:
This is a work around to issues with fragmented gets in ob1.

hjelmn:
Probably helps with puts too.

bosilca:
For the sake of completeness, this provides a nice partial solution to the PUT issue, but ensuring that the vader BTL accepts to handle very large fragments, removing the PML need to fragment messages. However, it does not fix the UT protocol, which remains broken for all multi-fragments messages.

al-rigazzi:
@bosilca can you be more specific? What protocols and components are affected?

PR #6952, opened 2019-09-03
v4.0.x: Datatype optimization and fix

jsquyres:
Refs issue #6932 (corresponding master PR: #6945).

PR #6978, opened 2019-09-13
Backport/3.1.x/6952

bosilca:
This effort started as the backport of #6952 to the 3.1 branch, but due to a large number of conflicts it expanded to include everything necessary to bring the datatype engine of 3.1 in sync with 4.0 (and master).

I lightly tested it, and everything looks OK. However, what started as a simple backport became a gigantic patch and I certainly do not plan to spend more time on it. As such, I would advice against it's inclusion in the 3.1. Moving this functionality to 3.0 will be similarly complicated.


jsquyres:
@bosilca We agree: many thanks for the effort, but let's not take this to v3.1.x.  Can you give us a README-worthy description of what is broken?  I.e., if we don't fix this in v3.0.x / v3.1.x, we at least want to describe the corner case that doesn't work (and therefore the user should upgrade to v4.0.x).

bosilca:
I don't know what bugs are or not fixed, because the datatype engine in master and 4.x is now drastically different that the one in 3.x. All the bugs the original patch fixed in master and 4.x, were the result of the new datatype engine that has not been yet backported in 3.x. The patch provided here would have done all that (bring the 3.x datatype engine in sync with the master and fix all known bugs), but as we discussed we will not get it in. So, the 3.x has the old datatype engine with its own set of bugs.

jsquyres:
Per https://github.com/open-mpi/ompi/issues/7019#issuecomment-536114609, this PR has been replaced with #7023.

Issue #6777, opened 2019-06-26
UCX seemingly causing Python MPI tests to fail with Open MPI v4.0.1


jsquyres:
[As reported on the mpi4py bitbucket](https://bitbucket.org/mpi4py/mpi4py/issues/124/test-failure-with-openmpi-401), it looks like enabling UCX support in Open MPI v4.0.1 in Fedora 30 is causing some mpi4py tests to fail.

See the link above for more details, but the short version is:

* Preliminary indications look like this is a bug in Open MPI and/or UCX, not mpi4py
* The mpi4py tests in question fail with openmpi-4.0.1-1.fc31.x86_64, but succeed with openmpi-4.0.1-5.fc31.x86_64. 
* @opoplawski, who is the Open MPI maintainer for Fedora, confirmed that:
    1. He disabled UCX in the -5 version, which enabled the tests to work.
    1. The issue was supposedly fixed in UCX v1.5.2, so Orion re-enabled UCX Open MPI support in -6, and the tests broke again.
* It *looks* like the corresponding MPI calls are writing past the buffer end for datatypes `MPI_SIGNED_CHAR` and `MPI_SHORT`.

@jladd-mlnx @artpol84 Can someone from Mellanox look into this?

FYI: @dalcinl @opoplawski

artpol84:
@jsquyres 

`openmpi-4.0.1-1.fc31.x86_64` seems to be older than `openmpi-4.0.1-5.fc31.x86_64`. Can this indicate that the problem was fixed?

artpol84:
FYI: @yosefe

jsquyres:
@artpol84 No, please re-read my summary and/or the entire thread.

artpol84:
> > He disabled UCX in the -5 version, which enabled the tests to work.
> @artpol84 No, please re-read my summary and/or the entire thread.

@jsquyres, thanks. I missed it.

jsquyres:
@artpol84 Can Mellanox check to see if this is now fixed with Open MPI v4.0.2?

artpol84:
@jsquyres I don't expect it to be fixed as of now. UCX doesn't support 1B and 2B atomics.
We are planning to fix it in the near future, but it is not yet fixed.
@janjust, @jladd-mlnx, please correct me if I am wrong.

opoplawski:
I suspect that this was resolved at some point.  At least, mpi4py test_rma is no longer failing on Fedora Rawhide.

[...]

PR #6980, opened 2019-09-16
UCX osc: add support for acc_single_intrinsic

devreal:
This PR adds support for the `acc_single_intrinsic` info key / mca parameter that allows the user to specify that only a single intrinsic type will be used in accumulate operations to avoid taking the accumulate lock for every operation. It also enables asynchronicity for accumulate operations that do not require emulation through compare-and-swap (currently only `replace` and `sum` and `compare-and-swap` itself).

Limitations:

- ~128 bit types are not supported and lead to an error if this info key is set (not sure how to handle 128 bit values that with the current UCX API)~
- ~Operations other than `replace` and `sum` in `MPI_Accumulate` require a `get` followed by a `cas` loop until successful replacement. This loop is currently executed directly but could be encapsulated in a UCX completion request, issuing a new cas if the previous one failed. Not sure whether UCX allows new operations to be issued from within a completion callback though.~
- ~After merging of #6933 there might be a few more chances to reduce latency, e.g., by fencing the release of the accumulate lock with the completion of the final put (if `acc_single_intrinsic` is not enabled).~

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -816,8 +816,9 @@ int ompi_osc_ucx_fetch_and_op(const void *origin_addr, void *result_addr,
             }
         }
 
-        ret = opal_common_ucx_wpmem_fetch(module->mem, opcode, value, target,
-                                        (void *)result_addr, dt_bytes, remote_addr);
+        ret = opal_common_ucx_wpmem_fetch_nb(module->mem, opcode, value, target,
+                                             (void *)result_addr, dt_bytes,
+                                             remote_addr, NULL, NULL);
 
         if (module->acc_single_intrinsic) {
             return ret;
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -1,5 +1,7 @@
 /*
- * Copyright (C) Mellanox Technologies Ltd. 2001-2017. ALL RIGHTS RESERVED.
+ * Copyright (C) 2001-2017 Mellanox Technologies Ltd. ALL RIGHTS RESERVED.
+ * Copyright (c) 2019-2020 High Performance Computing Center Stuttgart,
+ *                         University of Stuttgart.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -998,6 +998,7 @@ int ompi_osc_ucx_rput(const void *origin_addr, int origin_count,
                                          sizeof(uint64_t), remote_addr,
                                          req_completion, ucx_req);
     if (ret != OMPI_SUCCESS) {
+        OMPI_OSC_UCX_REQUEST_RETURN(ucx_req);
         return ret;
     }
 
@@ -1049,6 +1050,7 @@ int ompi_osc_ucx_rget(void *origin_addr, int origin_count,
                                          sizeof(uint64_t), remote_addr,
                                          req_completion, ucx_req);
     if (ret != OMPI_SUCCESS) {
+        OMPI_OSC_UCX_REQUEST_RETURN(ucx_req);
         return ret;
     }
 
@@ -1077,6 +1079,7 @@ int ompi_osc_ucx_raccumulate(const void *origin_addr, int origin_count,
     ret = ompi_osc_ucx_accumulate(origin_addr, origin_count, origin_dt, target, target_disp,
                                   target_count, target_dt, op, win);
     if (ret != OMPI_SUCCESS) {
+        OMPI_OSC_UCX_REQUEST_RETURN(ucx_req);
         return ret;
     }
 
@@ -1111,6 +1114,7 @@ int ompi_osc_ucx_rget_accumulate(const void *origin_addr, int origin_count,
                                       target, target_disp, target_count,
                                       target_datatype, op, win);
     if (ret != OMPI_SUCCESS) {
+        OMPI_OSC_UCX_REQUEST_RETURN(ucx_req);
         return ret;
     }
 
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -417,21 +417,22 @@ static int atomic_op_cswap(
     for (int i = 0; i < origin_count; ++i) {
 
         uint64_t tmp_val;
+        uint64_t target_val = 0;
+
+        // get the value from the origin
+        ret = opal_common_ucx_wpmem_putget(module->mem, OPAL_COMMON_UCX_GET,
+                                            target, &target_val, origin_dt_bytes,
+                                            remote_addr);
+        if (ret != OMPI_SUCCESS) {
+            return ret;
+        }
+
+        ret = opal_common_ucx_wpmem_flush(module->mem, OPAL_COMMON_UCX_SCOPE_EP, target);
+        if (ret != OMPI_SUCCESS) {
+            return ret;
+          }
+
         do {
-            uint64_t target_val = 0;
-
-            // get the value from the origin
-            ret = opal_common_ucx_wpmem_putget(module->mem, OPAL_COMMON_UCX_GET,
-                                               target, &target_val, origin_dt_bytes,
-                                               remote_addr);
-            if (ret != OMPI_SUCCESS) {
-                return ret;
-            }
-
-            ret = opal_common_ucx_wpmem_flush(module->mem, OPAL_COMMON_UCX_SCOPE_EP, target);
-            if (ret != OMPI_SUCCESS) {
-                return ret;
-            }
 
             tmp_val = target_val;
             // compute the result value
--- a/ompi/mca/osc/ucx/osc_ucx_comm.c
+++ b/ompi/mca/osc/ucx/osc_ucx_comm.c
@@ -323,6 +323,149 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
     return ret;
 }
 
+static int atomic_op_replace_sum(
+    ompi_osc_ucx_module_t  *module,
+    struct ompi_op_t       *op,
+    int                     target,
+    const void             *origin_addr,
+    int                     origin_count,
+    struct ompi_datatype_t *origin_dt,
+    ptrdiff_t               target_disp,
+    int                     target_count,
+    struct ompi_datatype_t *target_dt,
+    void                   *result_addr)
+{
+    int ret = OMPI_SUCCESS;
+    size_t origin_dt_bytes;
+    size_t target_dt_bytes;
+    ompi_datatype_type_size(origin_dt, &origin_dt_bytes);
+    ompi_datatype_type_size(target_dt, &target_dt_bytes);
+
+    if (origin_dt_bytes  > sizeof(uint64_t) ||
+        origin_dt_bytes != target_dt_bytes  ||
+        target_count    != origin_count) {
+        return OMPI_ERR_NOT_SUPPORTED;
+    }
+
+    uint64_t remote_addr = (module->addrs[target]) + target_disp * OSC_UCX_GET_DISP(module, target);
+
+    if (module->flavor == MPI_WIN_FLAVOR_DYNAMIC) {
+        ret = get_dynamic_win_info(remote_addr, module, target);
+        if (ret != OMPI_SUCCESS) {
+            return ret;
+        }
+    }
+
+    ucp_atomic_fetch_op_t opcode;
+    if (op == &ompi_mpi_op_replace.op) {
+        opcode = UCP_ATOMIC_FETCH_OP_SWAP;
+    } else {
+        opcode = UCP_ATOMIC_FETCH_OP_FADD;
+    }
+
+    for (int i = 0; i < origin_count; ++i) {
+        uint64_t value = 0;
+        memcpy(&value, origin_addr, origin_dt_bytes);
+        ret = opal_common_ucx_wpmem_fetch_nb(module->mem, opcode, value, target,
+                                             result_addr ? result_addr : &(module->req_result),
+                                             origin_dt_bytes, remote_addr, NULL, NULL);
+
+        // advance origin and remote address
+        origin_addr  = (void*)((intptr_t)origin_addr + origin_dt_bytes);
+        remote_addr += origin_dt_bytes;
+        if (result_addr) {
+            result_addr = (void*)((intptr_t)result_addr + origin_dt_bytes);
+        }
+    }
+
+    return ret;
+}
+
+static int atomic_op_cswap(
+    ompi_osc_ucx_module_t  *module,
+    struct ompi_op_t       *op,
+    int                     target,
+    const void             *origin_addr,
+    int                     origin_count,
+    struct ompi_datatype_t *origin_dt,
+    ptrdiff_t               target_disp,
+    int                     target_count,
+    struct ompi_datatype_t *target_dt,
+    void                   *result_addr)
+{
+    int ret = OMPI_SUCCESS;
+    size_t origin_dt_bytes;
+    size_t target_dt_bytes;
+    ompi_datatype_type_size(origin_dt, &origin_dt_bytes);
+    ompi_datatype_type_size(target_dt, &target_dt_bytes);
+
+    if (origin_dt_bytes  > sizeof(uint64_t) ||
+        origin_dt_bytes != target_dt_bytes  ||
+        target_count    != origin_count) {
+        return OMPI_ERR_NOT_SUPPORTED;
+    }
+
+    uint64_t remote_addr = (module->addrs[target]) + target_disp * OSC_UCX_GET_DISP(module, target);
+
+    if (module->flavor == MPI_WIN_FLAVOR_DYNAMIC) {
+        ret = get_dynamic_win_info(remote_addr, module, target);
+        if (ret != OMPI_SUCCESS) {
+            return ret;
+        }
+    }
+
+    for (int i = 0; i < origin_count; ++i) {
+
+        uint64_t tmp_val;
+        do {
+            uint64_t target_val = 0;
+
+            // get the value from the origin
+            ret = opal_common_ucx_wpmem_putget(module->mem, OPAL_COMMON_UCX_GET,
+                                               target, &target_val, origin_dt_bytes,
+                                               remote_addr);
+            if (ret != OMPI_SUCCESS) {
+                return ret;
+            }
+
+            ret = opal_common_ucx_wpmem_flush(module->mem, OPAL_COMMON_UCX_SCOPE_EP, target);
+            if (ret != OMPI_SUCCESS) {
+                return ret;
+            }
+
+            tmp_val = target_val;
+            // compute the result value
+            ompi_op_reduce(op, (void *)origin_addr, &tmp_val, 1, origin_dt);
+
+            // compare-and-swap the resulting value
+            ret = opal_common_ucx_wpmem_cmpswp(module->mem, target_val, tmp_val,
+                                               target, &tmp_val, origin_dt_bytes,
+                                               remote_addr);
+            if (ret != OMPI_SUCCESS) {
+                return ret;
+            }
+
+            // check whether the conditional swap was successful
+            if (tmp_val == target_val) {
+              break;
+            }
+
+        } while (1);
+
+        // store the result if necessary
+        if (NULL != result_addr) {
+          memcpy(result_addr, &tmp_val, origin_dt_bytes);
+          result_addr = (void*)((intptr_t)result_addr + origin_dt_bytes);
+        }
+        // advance origin and remote address
+        origin_addr  = (void*)((intptr_t)origin_addr + origin_dt_bytes);
+        remote_addr += origin_dt_bytes;
+    }
+
+    return ret;
+}
+
+
 int ompi_osc_ucx_put(const void *origin_addr, int origin_count, struct ompi_datatype_t *origin_dt,
                      int target, ptrdiff_t target_disp, int target_count,
                      struct ompi_datatype_t *target_dt, struct ompi_win_t *win) {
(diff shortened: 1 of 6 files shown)


artpol84:
@devreal, sorry for the delay
I'll try to review later this week and will ask @janjust to go over it next week (as h isn't available this week).

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -582,13 +818,29 @@ int ompi_osc_ucx_compare_and_swap(const void *origin_addr, const void *compare_a
...
-    ret = opal_common_ucx_wpmem_cmpswp(module->mem,*(uint64_t *)compare_addr,
-                                     *(uint64_t *)origin_addr, target,
-                                     result_addr, dt_bytes, remote_addr);
-    if (ret != OMPI_SUCCESS) {
+    if (sizeof(uint64_t) < dt_bytes) {
+        return OMPI_ERR_NOT_SUPPORTED;
+    }
+
+    uint64_t compare_val;
+    memcpy(&compare_val, compare_addr, dt_bytes);
```
artpol84: Is this to ensure 64bit alignment?

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -582,13 +818,29 @@ int ompi_osc_ucx_compare_and_swap(const void *origin_addr, const void *compare_a
...
+    memcpy(result_addr, origin_addr, dt_bytes);
+    ret = opal_common_ucx_wpmem_fetch_nb(module->mem, UCP_ATOMIC_FETCH_OP_CSWAP,
```
artpol84: Could you clarify, why we need to replace `opal_common_ucx_wpmem_cmpswp`?

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+    void                   *result_addr,
+    ompi_osc_ucx_request_t *ucx_req)
+{
+    int ret = OMPI_SUCCESS;
+    size_t origin_dt_bytes;
+    size_t target_dt_bytes;
+    ompi_datatype_type_size(origin_dt, &origin_dt_bytes);
+    ompi_datatype_type_size(target_dt, &target_dt_bytes);
+
+    if (origin_dt_bytes  > sizeof(uint64_t) ||
```
artpol84: Should we also have ompi_datatype_is_predefined(origin_dt) check?

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+    void* user_req_ptr = NULL;
+    for (int i = 0; i < origin_count; ++i) {
+        uint64_t value = 0;
+        if ((origin_count - 1) == i && NULL != ucx_req) {
+            // the last item is used to feed the request, if needed
+            user_req_cb = &req_completion;
+            user_req_ptr = ucx_req;
+            // issue a fence if this is the last but not the only element
+            if (0 < i) {
+                ret = opal_common_ucx_wpmem_fence(module->mem);
```
artpol84: Currently, this fence is a no-op.
@yosefe do we need to have something stronger to ensure the correct ordering?

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+                }
+            }
+        }
+        if (is_no_op) {
+            value = 0;
+        } else {
+            memcpy(&value, origin_addr, origin_dt_bytes);
+        }
+        ret = opal_common_ucx_wpmem_fetch_nb(module->mem, opcode, value, target,
+                                             result_addr ? result_addr : &(module->req_result),
```
artpol84: Minor: I think the following would be a bit fewer computations:

```C
void *output_ptr = &(module->req_result);
if( result_addr ) {
    output_ptr = result_addr;
}
...
ret = opal_common_ucx_wpmem_fetch_nb(module->mem, opcode, value, target,
                                             output_addr, ... );
...
if (result_addr) {
    output_addr += origin_dt_bytes;
}
```

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+
+        // get the value from the origin
+        ret = opal_common_ucx_wpmem_putget(module->mem, OPAL_COMMON_UCX_GET,
+                                            target, &target_val, origin_dt_bytes,
+                                            remote_addr);
+        if (ret != OMPI_SUCCESS) {
+            return ret;
+        }
+
+        ret = opal_common_ucx_wpmem_flush(module->mem, OPAL_COMMON_UCX_SCOPE_EP, target);
```
artpol84: You also will need to flush it per element.

artpol84:
Overall, this looks good to me. Though I'd like to double-check some details.

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+    void                   *result_addr,
+    ompi_osc_ucx_request_t *ucx_req)
+{
+    int ret = OMPI_SUCCESS;
+    size_t origin_dt_bytes;
+    size_t target_dt_bytes;
+    ompi_datatype_type_size(origin_dt, &origin_dt_bytes);
+    ompi_datatype_type_size(target_dt, &target_dt_bytes);
+
+    if (origin_dt_bytes  > sizeof(uint64_t) ||
```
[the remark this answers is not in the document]
devreal: I'd expect that this is checked upon entry to MPI. Have to check if that happens.

ompi/mca/osc/ucx/osc_ucx_comm.c
devreal: Yeah, I've seen 64bit accesses of potentially arbitrarily aligned pointers provided by the user in several places (here and elsewhere in Open MPI, iirc). I'm not sure whether that is safe on all architectures (it is on x86 and seems to be on Arm >= v7). The C language defines it as UB if a pointer references a memory location that is not suitably aligned for the referenced type (www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf, page 74)

ompi/mca/osc/ucx/osc_ucx_comm.c
devreal: There is no `nb` version of `opal_common_ucx_wpmem_cmpswp`, it calls `opal_common_ucx_atomic_cswap` (and in turn `opal_common_ucx_atomic_fetch`) and eventually waits for the operation to complete.

ompi/mca/osc/ucx/osc_ucx_comm.c
devreal: There is also this nugget hidden in `opal_common_ucx_atomic_cswap`:

```C
        if (op_size == sizeof(uint64_t)) {
            *(uint64_t*)result = tmp;
        } else {
            assert(op_size == sizeof(uint32_t));
            *(uint32_t*)result = tmp;
        }
```
https://github.com/open-mpi/ompi/blob/master/opal/mca/common/ucx/common_ucx.h#L217

I can only guess that the intention there was to avoid unaligned accesses? I don't see language in the standard that prohibits `MPI_Compare_and_swap` to be called on `MPI_SHORT`.

Since we have only four relevant data type sizes to cover (1, 2, 4, 8) we could have an inlined custom `memcpy` that issues loads on appropriately typed pointers, if there is a concern about using `memcpy`.

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +323,215 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+        }
+    }
+
+    for (int i = 0; i < origin_count; ++i) {
+
+        uint64_t tmp_val;
+        uint64_t target_val = 0;
+
+        // get the value from the origin
+        ret = opal_common_ucx_wpmem_putget(module->mem, OPAL_COMMON_UCX_GET,
```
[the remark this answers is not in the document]
devreal: I agree. Considering that the standard restricts the use of accumulate operations to either `same_op` or `same_op_no_op` we can fall back to taking the accumulate lock for operations that are not supported by UCX (like `MPI_PROD`)

ompi/mca/osc/ucx/osc_ucx_comm.c
devreal: I will change that, sounds reasonable.

devreal:
I addressed the reviewer comments and added two optimizations: the accumulate lock won't be acquired if an exclusive lock is present; and the release of the accumulate lock will be overlapped with any previous operations.

artpol84:
Thanks, I’ll go over later this week

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -235,37 +235,76 @@ static inline int ddt_put_get(ompi_osc_ucx_module_t *module,
...
+            ret = opal_common_ucx_wpmem_cmpswp(module->state_mem,
+                                            TARGET_LOCK_UNLOCKED, TARGET_LOCK_EXCLUSIVE,
+                                            target, &result_value, sizeof(result_value),
+                                            remote_addr);
+            if (ret != OMPI_SUCCESS) {
+                OSC_UCX_VERBOSE(1, "opal_common_ucx_mem_cmpswp failed: %d", ret);
+                return OMPI_ERROR;
+            }
+            if (result_value == TARGET_LOCK_UNLOCKED) {
+                return OMPI_SUCCESS;
```
devreal: break here

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +362,110 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+static inline
+bool use_ucx_op(struct ompi_op_t *op, struct ompi_datatype_t *origin_dt)
+{
+
+    if (op == &ompi_mpi_op_replace.op ||
+        op == &ompi_mpi_op_sum.op     ||
+        op == &ompi_mpi_op_no_op.op) {
+        size_t dt_bytes;
+        ompi_datatype_type_size(origin_dt, &dt_bytes);
+        if (ompi_datatype_is_predefined(origin_dt) &&
```
devreal: `assert` that `origin_dt` is a predefined datatype

devreal: Check for `module->acc_single_intrinsic` here

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -323,6 +362,110 @@ static inline int get_dynamic_win_info(uint64_t remote_addr, ompi_osc_ucx_module
...
+    void                   *result_addr,
+    ompi_osc_ucx_request_t *ucx_req)
+{
+    int ret = OMPI_SUCCESS;
+    size_t origin_dt_bytes;
+    size_t target_dt_bytes;
+    ompi_datatype_type_size(origin_dt, &origin_dt_bytes);
+    ompi_datatype_type_size(target_dt, &target_dt_bytes);
+
+    if (sizeof(uint64_t) > origin_dt_bytes  ||
```
devreal: Check if UCX supports 8 and 16bit operands

devreal: See also https://github.com/open-mpi/ompi/issues/6777

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -582,14 +745,23 @@ int ompi_osc_ucx_compare_and_swap(const void *origin_addr, const void *compare_a
     }
 
     ompi_datatype_type_size(dt, &dt_bytes);
-    ret = opal_common_ucx_wpmem_cmpswp(module->mem,*(uint64_t *)compare_addr,
-                                     *(uint64_t *)origin_addr, target,
-                                     result_addr, dt_bytes, remote_addr);
-    if (ret != OMPI_SUCCESS) {
+    if (sizeof(uint64_t) < dt_bytes) {
```
devreal: Move this check to the beginning of the function

devreal:
@artemry-mlnx I finally found the time to work on this again. I hopefully have addressed all the points we found during our review in December. Please take a look if you can spare some time :) I would be happy if we could get this into 5.0.

@yosefe One point that is not clear to me:

There is a specific function in `opal/mca/common/ucx/common_ucx.h` for compare-and-swap (https://github.com/open-mpi/ompi/blob/master/opal/mca/common/ucx/common_ucx.h#L204). I don't really understand why this code is needed (and doubt that it's correct):

```C
static inline
int opal_common_ucx_atomic_cswap(ucp_ep_h ep, uint64_t compare,
                                 uint64_t value, void *result, size_t op_size,
                                 uint64_t remote_addr, ucp_rkey_h rkey,
                                 ucp_worker_h worker)
{
    uint64_t tmp = value;
    int ret;

    ret = opal_common_ucx_atomic_fetch(ep, UCP_ATOMIC_FETCH_OP_CSWAP, compare, &tmp,
                                       op_size, remote_addr, rkey, worker);
    if (OPAL_LIKELY(OPAL_SUCCESS == ret)) {
        /* in case if op_size is constant (like sizeof(type)) then this condition
         * is evaluated in compile time */
        if (op_size == sizeof(uint64_t)) {
            *(uint64_t*)result = tmp;
        } else {
            assert(op_size == sizeof(uint32_t));
            *(uint32_t*)result = tmp;
        }
    }
    return ret;
}
```

AFAICS, the assignment from `tmp` after the operation is faulty in the 32 bit case: with `op_size == 4` the upper 4B of `tmp` will be overwritten with the previous value in `remote_addr` and the lower 4B should be left untouched by `UCP_ATOMIC_FETCH_OP_CSWAP`. However, the assignment `*(uint32_t*)result = tmp;` will downcast the value in `tmp`, cutting off the upper 4B and assigning the lower 4B. Am I missing something here? Do we really need a special function for compare-and-swap or can't we just handle it through `opal_common_ucx_atomic_fetch` with `UCP_ATOMIC_FETCH_OP_CSWAP`? See also the comment here https://github.com/open-mpi/ompi/pull/6980#discussion_r340424408

yosefe:
@devreal it works because the data is taken from first 4B of the target address and also placed in first 4B of the result. on LE arch, the lower 4B of tmp are overwritten with the previous value in remote_addr, not the upper 4B.

devreal:
> it works because the data is taken from first 4B of the target address and also placed in first 4B of tmp. on LE arch, the lower 4B of tmp are overwritten with the previous value in remote_addr, not the upper 4B.

Ahh, I see. I was not thinking in terms of LE architectures. That would not be portable to BE architectures though, right? Why not copy the `value` into `result` and pass the `result` pointer directly to `opal_common_ucx_atomic_fetch`? That would make it easier to create a nonblocking version of compare-and-swap.

In similar terms, it is not entirely clear to me what the format of the 64bit value passed to `ucp_atomic_fetch_nb` is if `op_size == 4`. Is the 64-bit value cast down to a 32-bit integer? Are either the upper or lower 4B used? I don't see that specified in the UCX standard.

yosefe:
@devreal it works on BE as well, in this case the first 4B of tmp will be tye "upper" part. However, it doesn't really matter, we don't to integer case on tmp, we just do pointer cast, which does not care about endianness.

- the input value for UCX atomic is "uint64_t" so it's int-casted to uint32 if needed.
- the output value of the atomic is "void*" so it's a direct pointer to 32-bit result

devreal:
I updated this PR to include the following:

1) a non-blocking variant of `opal_common_ucx_wpmem_cmpswp`
1) a safe and portable way to load and store 64-bit integer values to variable-size pointers without relying on `memcpy`
2) support for compare-and-swap for types that are not supported by UCX atomics (from what I understand that is 8, 16, and 128-bit types)

devreal:
@awlauria I think this PR is now in a good state again (after a long pause). It's awaiting a review now. Can we add this to the list of tracked issues for 5.0?

awlauria:
@devreal I will add it to the list. Thanks! However - it looks like there may be a commit or here missing the signed-off by, can you fix that?

devreal:
@awlauria Yes, I'm aware of that. I wasn't sure whether some of the commits should be squashed. I will fix that one unsigned today though.

devreal:
Fixed two bugs that slipped through during initial testing and squashed down some commits.

artpol84:
@janjust can you please review

janjust:
@devreal @artpol84 , guys this looks ok to me -I don't have any constructive comments. But @devreal out of curiosity, what kind of testing was done on this? Do you have any tests of your own that you sued, either way, can you share?

devreal:
@janjust I tested this on a ConnectX-3 cluster with a benchmark I wrote that covers most of the MPI RMA functions and runs with both shared and exclusive locks. The code is available at https://github.com/devreal/mpi-progress and I am happy to contribute that to any existing test suite.

awlauria:
@artpol84 you still have a change request that is preventing this from getting merged. Can you give it another look when you get a chance? Thanks!

jladd-mlnx:
@artpol84 have you and @devreal reached a consensus? If so, can you please remove the request change label, Artem?

jladd-mlnx:
@janjust @artpol84 @devreal what is the status? Seems like it's a critical fix. Can we merge it?

artpol84:
I'd like to review one more time. I'll plan to go over it this week. I am sorry for not doing this before.

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -582,20 +780,50 @@ int ompi_osc_ucx_compare_and_swap(const void *origin_addr, const void *compare_a
     }
 
     ompi_datatype_type_size(dt, &dt_bytes);
-    ret = opal_common_ucx_wpmem_cmpswp(module->mem,*(uint64_t *)compare_addr,
-                                     *(uint64_t *)origin_addr, target,
-                                     result_addr, dt_bytes, remote_addr);
+    if (ATOMIC_SIZE_SUPPORTED(dt_bytes)) {
```
artpol84, line 783:
```suggestion
    if (ATOMIC_SIZE_SUPPORTED(remote_addr, dt_bytes)) {
```

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -604,16 +832,22 @@ int ompi_osc_ucx_fetch_and_op(const void *origin_addr, void *result_addr,
         return ret;
     }
 
-    if (op == &ompi_mpi_op_no_op.op || op == &ompi_mpi_op_replace.op ||
-        op == &ompi_mpi_op_sum.op) {
+    ompi_datatype_type_size(dt, &dt_bytes);
+
+    /* UCX atomics are only supported on 32 and 64 bit values */
+    if (ATOMIC_SIZE_SUPPORTED(dt_bytes) &&
```
artpol84, line 838:
```suggestion
    uint64_t remote_addr = (module->addrs[target]) + target_disp * OSC_UCX_GET_DISP(module, target);
    if (ATOMIC_SIZE_SUPPORTED(remote_addr, dt_bytes) &&
```

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -604,16 +832,22 @@ int ompi_osc_ucx_fetch_and_op(const void *origin_addr, void *result_addr,
...
+        (op == &ompi_mpi_op_no_op.op || op == &ompi_mpi_op_replace.op ||
+         op == &ompi_mpi_op_sum.op)) {
+        uint64_t value;
         uint64_t remote_addr = (module->addrs[target]) + target_disp * OSC_UCX_GET_DISP(module, target);
```
artpol84, line 842:
```suggestion
```

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -763,17 +1009,29 @@ int ompi_osc_ucx_get_accumulate(const void *origin_addr, int origin_count,
             if (ret != OMPI_SUCCESS) {
                 return ret;
             }
-
```
artpol84, line 766: Can we remove this?

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -544,15 +708,53 @@ int ompi_osc_ucx_accumulate(const void *origin_addr, int origin_count,
...
 
-        ret = opal_common_ucx_wpmem_flush(module->mem, OPAL_COMMON_UCX_SCOPE_EP, target);
+    }
+
+    if (NULL != ucx_req) {
+        // nothing to wait for, mark request as completed
+        ompi_request_complete(&ucx_req->super, true);
+    }
+
+    return end_atomicity(module, target, lock_acquired, free_ptr);
```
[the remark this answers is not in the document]
devreal: `free_ptr` is passed to `end_atomicity` because the transfer is overlapped with the release of the atomic lock and only after that the `free_ptr` can actually be free'd.

ompi/mca/osc/ucx/osc_ucx_comm.c
devreal: What is 'this'? The whitespace? It's part of a longer section of code that was removed.

ompi/mca/osc/ucx/osc_ucx_comm.c
```diff
@@ -24,6 +26,10 @@
         return OMPI_ERROR;                                                       \
     }
 
+/* macro to check whether UCX supports atomic operation on the size the operands */
+#define ATOMIC_SIZE_SUPPORTED(_size) (sizeof(uint32_t) == _size || \
+                                      sizeof(uint64_t) == _size)
```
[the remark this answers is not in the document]
devreal: I have extended `ATOMIC_SIZE_SUPPORTED` to check for proper target memory alignment

ompi/mca/osc/ucx/osc_ucx_comm.c
artpol84: ok

artpol84:
@devreal , Do we have everything in place now?

Issue #24, opened 2014-10-01
Implement MPIX_GREQUEST_START


ompiteam:
Rob Latham proposed MPIX_GREQUEST_START as described in "Extending the MPI-2 Generalized Request Interface":http://www-unix.mcs.anl.gov/~thakur/papers/grequest-redesign.pdf (PDF).  Prototypes are in the paper (no use reproducing them here).

The main improvement is allowing generalized requests to specify their own progression function that will be invoked by MPI's progress engine.  This MPIX_GREQUEST_START function is in MPICH2 and is now used by newer versions of ROMIO.

devreal:
The infrastructure for extended generalized requests has been around for quite a while in Open MPI (protected by `--enable-grequest-extensions`). Any opposition to adding the user-facing API for extended generalized requests? It's not in the standard but has been used in a paper at EuroMPI'19 and I've been hacking my way around using it with Open MPI myself.

jsquyres:
As long as it's MPIX, I doubt there's a problem.

jjhursey:
@ggouaillardet Do you know the state of this ticket? I think it can be closed. However, after looking at the code I think that some of the protections added for the `--enable-grequest-extensions` may have been removed by accident in [this](https://github.com/open-mpi/ompi/commit/1e2019ce2a903be24361b3424d8e98d27e941c6c) commit.

I see these two changes from you so I thought I would reach out to see what you think:
 * Implement the extensions: https://github.com/open-mpi/ompi/pull/5337
 * Make them internal: https://github.com/open-mpi/ompi/pull/5353

PR #6991, opened 2019-09-18
Ensure that grequestx continuously make progress

devreal:
The handling of the grequest extensions is missing the release of the `in_progress` guard, causing requests not being polled after the first invocation.

### DIFF
--- a/ompi/request/grequestx.c
+++ b/ompi/request/grequestx.c
@@ -50,6 +50,7 @@ static int grequestx_progress(void) {
             }
             OPAL_THREAD_LOCK(&lock);
         }
+        in_progress = false;
     }
     OPAL_THREAD_UNLOCK(&lock);
 


bosilca:
Nice catch.

jsquyres:
Does this mean that grequests were, in general, broken?  (i.e., never being advanced)

bosilca:
yup.

jsquyres:
This is NEWS-worthy, then...

jsquyres:
@devreal Can you cherry-pick to the appropriate branches?

devreal:
@jsquyres :

> Does this mean that grequests were, in general, broken? (i.e., never being advanced)

I think only the extended grequests (see #24) were affected, the ones that come with a progress function. that are not in the MPI standard. AFAICS they are used internally by ROMIO but are not exposed to the user (unlike MPICH). The regular grequests should not be affected. 

I will cherry-pick to the release branches.

jsquyres:
Oh, if there's no way for the user to be exposed to this bug, then it's not worth cherry-picking.

Specifically: are we using the extended grequests functionality in ROMIO?

bosilca:
indeed, the normal grequest are having their query function called as expected. I wonder how is ROMIO working if we don't progress their requests as expected ? If this functionality used in the current code ?

devreal:
@bosilca Grepping through the romio code it seems that extended grequests are only used when a working aio library is found, for example [here](https://github.com/open-mpi/ompi/blob/master/ompi/mca/io/romio321/romio/adio/common/ad_iwrite.c#L177). Not sure how often this configuration is used.

PR #6993, opened 2019-09-19
Fix miscellaneous compiler warnings.

awlauria:

### DIFF
--- a/opal/datatype/opal_copy_functions_heterogeneous.c
+++ b/opal/datatype/opal_copy_functions_heterogeneous.c
@@ -89,13 +89,14 @@ struct bit80 {
 static inline void
 opal_dt_swap_long_double(void *to_p, const void *from_p, const size_t size, size_t count, uint32_t remoteArch)
 {
+
 #ifdef HAVE_IEEE754_H
-    size_t i;
-    long double*to = (long double *) to_p;
 
     if ((opal_local_arch&OPAL_ARCH_LDISINTEL) && !(remoteArch&OPAL_ARCH_LDISINTEL)) {
 #ifdef __x86_64
-        for (i=0; i<count; i++, to++) {
+        long double*to = (long double *) to_p;
+
+        for (size_t i=0; i<count; i++, to++) {
             union ieee854_long_double ld;
             struct bit128 * b = (struct bit128 *)to;
             ld.ieee.empty = 0;
@@ -108,7 +109,9 @@ opal_dt_swap_long_double(void *to_p, const void *from_p, const size_t size, size
 #endif
     } else if (!(opal_local_arch&OPAL_ARCH_LDISINTEL) && (remoteArch&OPAL_ARCH_LDISINTEL)) {
 #ifdef __sparcv9
-        for (i=0; i<count; i++, to++) {
+        long double*to = (long double *) to_p;
+
+        for (size_t i=0; i<count; i++, to++) {
             union ieee854_long_double ld;
             struct bit80 * b = (struct bit80 *)to;
             ld.ieee.mantissa3 = 0;
--- a/opal/mca/btl/tcp/btl_tcp_addr.h
+++ b/opal/mca/btl/tcp/btl_tcp_addr.h
@@ -68,7 +68,7 @@ struct mca_btl_tcp_addr_t {
 #if OPAL_ENABLE_IPV6
         struct in6_addr addr_inet6;     /* IPv6 listen address */
 #endif
-    };
+    } addr_union;
     in_port_t   addr_port;     /**< listen port */
     int         addr_ifkindex; /**< remote interface index assigned with
                                     this address */
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -334,12 +334,20 @@ static int intercept_brk (void *addr)
 
 #endif
 
-#define HAS_SHMDT (defined(SYS_shmdt) || \
-    (defined(IPCOP_shmdt) && defined(SYS_ipc)))
-#define HAS_SHMAT (defined(SYS_shmat) || \
-    (defined(IPCOP_shmat) && defined(SYS_ipc)))
+#if defined(SYS_shmdt) || (defined(IPCOP_shmdt) && defined(SYS_ipc))
+#define HAS_SHMDT 1
+#else
+#define HAS_SHMDT 0
+#endif
 
-#if (HAS_SHMDT || HAS_SHMAT) && defined(__linux__)
+#if defined(SYS_shmat) ||(defined(IPCOP_shmat) && defined(SYS_ipc))
+#define HAS_SHMAT 1
+#else
+#define HAS_SHMAT 0
+#endif
+
+#if HAS_SHMDT || HAS_SHMAT
+#if defined(__linux__)
 
 #include <stdio.h>
 #include <fcntl.h>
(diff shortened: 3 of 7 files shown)


opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
```diff
@@ -563,27 +563,28 @@ pmix_value_cmp_t pmix_bfrops_base_value_cmp(pmix_value_t *p,
             }
             rc = PMIX_EQUAL;
             break;
-        case PMIX_COORD:
-            rc = memcmp(p->data.coord, p1->data.coord, sizeof(pmix_coord_t));
-            if (0 > rc) {
+        case PMIX_COORD: {
```
rhc54: Ick - that sure violates a few of our coding standards. I don't really care about it here, though, as I can resolve that warning upstream and will overwrite any change made in OMPI anyway.

opal/datatype/opal_copy_functions_heterogeneous.c
```diff
@@ -89,13 +89,16 @@ struct bit80 {
 static inline void
 opal_dt_swap_long_double(void *to_p, const void *from_p, const size_t size, size_t count, uint32_t remoteArch)
 {
+
+#if defined(HAVE_IEEE754_H) && defined(__x86_64) || defined(__sparcv9)
```
rhc54: This is confusing - is there a way to make the precedence more obvious? I _think_ the "and" goes first and then the "or", but...

awlauria: Thanks - fixed.

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
awlauria: Is it the use of the `{`? I can change it if you have a preferred way.

This one might be a bug (how it could be exposed I am not entirely sure). So it would be good to get this fixed right sooner than later.

opal/datatype/opal_copy_functions_heterogeneous.c
bosilca: yeah this is wrong as && has a higher precendence than || so the new code is not equivalent to the old one. It should read  ```defined(HAVE_IEEE754_H) && (defined(__x86_64) || defined(__sparcv9))```.

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
rhc54: No worries - I'd leave it as-is. It will get overwritten next time I update PMIx which happens pretty frequently.

opal/datatype/opal_copy_functions_heterogeneous.c
bosilca: However this patch does not make the code more readable. We can remove the ```to``` (it is never used as a long double), and then we can move the declaration of i inside the for.

opal/mca/btl/tcp/btl_tcp_addr.h
```diff
@@ -68,7 +68,7 @@ struct mca_btl_tcp_addr_t {
 #if OPAL_ENABLE_IPV6
         struct in6_addr addr_inet6;     /* IPv6 listen address */
 #endif
-    };
+    } addr_union;
```
bosilca: We require C99 which accepts unnamed unions so what is the need for this ?

opal/mca/memory/patcher/memory_patcher_component.c
```diff
@@ -334,12 +334,20 @@ static int intercept_brk (void *addr)
 
 #endif
 
-#define HAS_SHMDT (defined(SYS_shmdt) || \
-    (defined(IPCOP_shmdt) && defined(SYS_ipc)))
-#define HAS_SHMAT (defined(SYS_shmat) || \
-    (defined(IPCOP_shmat) && defined(SYS_ipc)))
+#if defined(SYS_shmdt) || (defined(IPCOP_shmdt) && defined(SYS_ipc))
```
bosilca: Why ?

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
bosilca: What are we trying to fix here ?

opal/util/malloc.c
```diff
@@ -52,13 +52,14 @@ int opal_malloc_debug_level = OPAL_MALLOC_DEBUG_LEVEL;
 int opal_malloc_output = -1;
 
 
+
+#if OPAL_ENABLE_DEBUG
+
 /*
  * Private variables
  */
 static opal_output_stream_t malloc_stream;
```
bosilca: I think it is safe to completely remove malloc_stream, it is never used anywhere.

opal/mca/memory/patcher/memory_patcher_component.c
awlauria: This silences:

```warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
```

opal/util/malloc.c
awlauria: I thought about that as well, but figured this was safest for the purposes of this PR.

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
awlauria: rc was defined as a `pmix_value_cmp_t` which can't store values < 0.

opal/mca/btl/tcp/btl_tcp_addr.h
bosilca: Right, it's a GNU extension for C99 but it is officially only in C11.

opal/mca/memory/patcher/memory_patcher_component.c
bosilca: clang is awesome.

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
bosilca: indeed. why not use ret instead ?

opal/util/malloc.c
awlauria: Removed in a separate commit to make it easier to revert if anyone wants it to hang around.

opal/mca/pmix/pmix4x/pmix/src/mca/bfrops/base/bfrop_base_fns.c
awlauria: Good point - didn't notice ret, thanks. Fixed.

opal/datatype/opal_copy_functions_heterogeneous.c
awlauria: thanks, updated.

opal/datatype/opal_copy_functions_heterogeneous.c
```diff
@@ -89,13 +89,14 @@ struct bit80 {
...
 {
+
 #ifdef HAVE_IEEE754_H
-    size_t i;
-    long double*to = (long double *) to_p;
 
     if ((opal_local_arch&OPAL_ARCH_LDISINTEL) && !(remoteArch&OPAL_ARCH_LDISINTEL)) {
 #ifdef __x86_64
-        for (i=0; i<count; i++, to++) {
+        long double*to = (long double *) to_p;
```
awlauria: To fix the CI compiler failure and make sure we preserve functionality, I moved the temporary pointer `to` back inside of each #define as a new fix.

awlauria:
I rebased the conflicts out.

rhc54:
I can't speak to the remaining changes in this PR - the conflict with the PMIx change has been resolved.

PR #6998, opened 2019-09-21
v3.0.x: Do not use CMA in user namespaces

jsquyres:
Trying out to run processes via mpirun in Podman containers has shown
that the CMA btl_vader_single_copy_mechanism does not work when user
namespaces are involved.

Creating containers with Podman requires at least user namespaces to be
able to do unprivileged mounts in a container

Even if running the container with user namespace user ID mappings which
result in the same user ID on the inside and outside of all involved
containers, the check in the kernel to allow ptrace (and thus
process_vm_{read,write}v()), fails if the same IDs are not in the same
user namespace.

One workaround is to specify '--mca btl_vader_single_copy_mechanism none'
and this commit adds code to automatically skip CMA if user namespaces
are detected and fall back to MCA_BTL_VADER_EMUL.

Signed-off-by: Adrian Reber <areber@redhat.com>
(cherry picked from commit fc68d8a90fe86284e9dc730f878b55c0412f01d2)

### DIFF
--- a/opal/mca/btl/vader/btl_vader.h
+++ b/opal/mca/btl/vader/btl_vader.h
@@ -81,7 +81,12 @@ union vader_modex_t {
         void *segment_base;
     } xpmem;
 #endif
-    opal_shmem_ds_t seg_ds;
+    struct vader_modex_other_t {
+        ino_t user_ns_id;
+        int seg_ds_size;
+        /* seg_ds needs to be the last element */
+        opal_shmem_ds_t seg_ds;
+    } other;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


jsquyres:
FYI @adrianreber 

@bwbarrett Thoughts on bringing this back to v3.0.x?

ibm-ompi:

 Gist: https://gist.github.com/9bb9548c565aef8af0feab982bb05982

ibm-ompi:

 Gist: https://gist.github.com/b0175ddd3d85b4b86d4fa80ee3072fef

jsquyres:
@adrianreber It turns out that `MCA_BTL_VADER_EMUL` does not exist back on the v3.1.x/v3.0.x branches.  I updated this PR to instead fall back to `MCA_BTL_VADER_NONE`.

Can you please test that the functionality still works correctly for you?

PR #7003, opened 2019-09-23
v3.1.x: REF6976 Silent failure of OMPI over OFI with large messages sizes

mwheinz:
INTERNAL: STL-59403

(cherry-picked from commit 3aca4af548a3d781b6b52f89f4d6c7e66d379609)
Change-Id: I65ab130794574393edab4a318365c1c125e50712

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -247,13 +247,20 @@ ompi_mtl_ofi_send_start(struct mca_mtl_base_module_t *mtl,
     endpoint = ompi_mtl_ofi_get_endpoint(mtl, ompi_proc);
 
     ompi_ret = ompi_mtl_datatype_pack(convertor, &start, &length, &free_after);
-    if (OMPI_SUCCESS != ompi_ret) return ompi_ret;
+    if (OPAL_UNLIKELY(OMPI_SUCCESS != ompi_ret)) {
+        return ompi_ret;
+    }
 
     ofi_req->buffer = (free_after) ? start : NULL;
     ofi_req->length = length;
     ofi_req->status.MPI_ERROR = OMPI_SUCCESS;
 
-    if (OPAL_UNLIKELY(MCA_PML_BASE_SEND_SYNCHRONOUS == mode)) {
+    if (OPAL_UNLIKELY(length > endpoint->mtl_ofi_module->max_msg_size)) {
+        opal_show_help("help-mtl-ofi.txt",
+            "message too big", false,
+            length, endpoint->mtl_ofi_module->max_msg_size);
+        return OMPI_ERROR;
+    } else if (OPAL_UNLIKELY(MCA_PML_BASE_SEND_SYNCHRONOUS == mode)) {
         ack_req = malloc(sizeof(ompi_mtl_ofi_request_t));
         assert(ack_req);
         ack_req->parent = ofi_req;
--- a/ompi/mca/mtl/ofi/help-mtl-ofi.txt
+++ b/ompi/mca/mtl/ofi/help-mtl-ofi.txt
@@ -16,3 +16,6 @@ unusual; your job may behave unpredictably (and/or abort) after this.
   Local host: %s
   Location: %s:%d
   Error: %s (%zd)
+#
+[message too big]
+Message size %llu bigger than supported by selected transport. Max = %llu
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -466,9 +466,10 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
     }
 
     /**
-     * Save the maximum inject size.
+     * Save the maximum sizes.
      */
     ompi_mtl_ofi.max_inject_size = prov->tx_attr->inject_size;
+    ompi_mtl_ofi.max_msg_size = prov->ep_attr->max_msg_size;
 
     /**
      * Create the objects that will be bound to the endpoint.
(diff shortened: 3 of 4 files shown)


ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -247,13 +247,20 @@ ompi_mtl_ofi_send_start(struct mca_mtl_base_module_t *mtl,
     endpoint = ompi_mtl_ofi_get_endpoint(mtl, ompi_proc);
 
     ompi_ret = ompi_mtl_datatype_pack(convertor, &start, &length, &free_after);
-    if (OMPI_SUCCESS != ompi_ret) return ompi_ret;
+	if (OPAL_UNLIKELY(OMPI_SUCCESS != ompi_ret)) {
```
jsquyres: We have a very small number of style guidelines (https://github.com/open-mpi/ompi/wiki/CodingStyle), but one of them is: absolutely no tabs whatsoever.  Always spaces.  😄

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -247,13 +247,20 @@ ompi_mtl_ofi_send_start(struct mca_mtl_base_module_t *mtl,
...
+		return ompi_ret;
+	}
 
     ofi_req->buffer = (free_after) ? start : NULL;
     ofi_req->length = length;
     ofi_req->status.MPI_ERROR = OMPI_SUCCESS;
 
-    if (OPAL_UNLIKELY(MCA_PML_BASE_SEND_SYNCHRONOUS == mode)) {
+	if (OPAL_UNLIKELY(length > endpoint->mtl_ofi_module->max_msg_size)) {
```
jsquyres: Similar to libfabric, this check seems unnecessary (because it should be obeyed by higher layers).

Do you want to enable this check only in debug mode, perchance?

ompi/mca/mtl/ofi/mtl_ofi.h
```diff
@@ -247,13 +247,20 @@ ompi_mtl_ofi_send_start(struct mca_mtl_base_module_t *mtl,
...
+		opal_show_help("help-mtl-ofi.txt",
+			"message too big", false,
+			length, endpoint->mtl_ofi_module->max_msg_size);
+		return OMPI_ERROR;
+	} else if (OPAL_UNLIKELY(MCA_PML_BASE_SEND_SYNCHRONOUS == mode)) {
```
bosilca: why "else if" ?

ompi/mca/mtl/ofi/mtl_ofi.h
bosilca: It's an MTL so there are no higher layers to handle the error case. The next level here is directly the MPI user, and as a result any error returned to the higher level will automatically translate into an MPI error and thus an abort.

jsquyres: `cm` doesn't handle this case?

jsquyres: If `cm` doesn't handle this, why does setting `ompi_mtl_ofi.max_msg_size` matter?

mwheinz: max_inject_size is also referenced in ompi_mtl_ofi_send().

jsquyres: I totally missed that this is an OFI-module-specific value; it's not on a struct that `cm` groks.  Got it.

mwheinz: I am aware, however, that this is a performance sensitive path - it's the reason I used OPAL_UNLIKELY(). Speaking of which, I also wrapped an OPAL_UNLIKELY around the check for a failure from ompi_mtl_datatype_pack() - because that's only going to return an error if malloc() has failed.

ompi/mca/mtl/ofi/mtl_ofi.h
mwheinz: Mostly personal preference - I tend to prefer the form 

`if (A) {
    return fail;
} else if (B) {
   return fail;
}`

to 
`if (A) {
    return fail;
}
if (B) {
    return fail;
}`

but that's purely cosmetic. I can change it if you like.

ompi/mca/mtl/ofi/mtl_ofi.h
mwheinz: Doh! okay.

mwheinz:
Note that this patch is for 3.1.x only. The patch for 4.x/5.x/master needs to be slightly different.

hppritcha:
@mwheinz first need to open a PR on master then we cherry pick to release branches.

mwheinz:
> 
@hppritcha 

> @mwheinz first need to open a PR on master then we cherry pick to release branches.

Okay.

mwheinz:
Are there any other issues I should look at before I submit the master branch version?

gpaulsen:
@mwheinz, If you can get the v4.0.x PR done soon (presumably a cherry-pick -x from master if possible), we will get it into the next v4.0.2 rc.

jsquyres:
> Are there any other issues I should look at before I submit the master branch version?

No, I think you're good.  I should have mentioned this in my first review: we generally commit stuff to master first, and then `git cherry-pick -x ...` to get them to release branches (potentially adapting / porting the cherry-pick if necessary).  We like to see the `(cherry-picked from ...)` tagline in the release branch PRs / cherry-picks so that we can track what commits on the release branches came from where.

So once the master one is merged, can you amend this commit with a cherry-picked message (like it would have been if you `git cherry-pick -x ...`'ed)?

mwheinz:
> 
> 
> Everything looks good code-wise, now; thanks. Just need to get the cherry-pick notice in the commit message once #7004 is merged to master.

Will do.

jsquyres:
@mwheinz Do you plan to make a v3.0.x PR?  (we usually make v3.0.x, v3.1.x, and v4.0.x PRs unless there's a reason not to go back that far)

bwbarrett:
I'd prefer we not merge this into release branches until my comment on the master pull request (https://github.com/open-mpi/ompi/pull/7004) is discussed.  I don't think this was the right fix.

jsquyres:
Good to go for v3.1.x.

PR #7004, opened 2019-09-23
REF6976 Silent failure of OMPI over OFI with large messages sizes

mwheinz:
INTERNAL: STL-59403

Change-Id: I05aa71d332f2df897133b30c28bf37d98f061996

### DIFF
--- a/ompi/mca/mtl/ofi/help-mtl-ofi.txt
+++ b/ompi/mca/mtl/ofi/help-mtl-ofi.txt
@@ -75,3 +75,5 @@ recoverable and your application is likely to abort.
   Local host: %s
   Remote host: %s
   Error: %s (%d)
+[message too big]
+Message size %llu bigger than supported by selected transport. Max = %llu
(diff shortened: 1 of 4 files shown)


mwheinz:
Double checked the commit before submitting the pull request. No whitespace issues.</throws salt over shoulder></knocks on wood>

bwbarrett:
I don't think this patch is what we want to have happen.  This leads to a slightly more sane error message, but the OFI component doesn't do any fragmentation, so it still will be a fatal error message during runtime.  Wouldn't a better solution be either (1) not select providers with a small max message size or (2) add fragmentation?

jsquyres:
Ya, we discussed this exact issue on the RM call yesterday, and resolved to bring it up on the webex today (i.e.: this is a band-aid -- it resolves the "silent" part of the error, which is arguably the most nefarious part of the issue, but the more complete fix [i.e., sending arbitrary-sized messages] is more involved).

PR #7005, opened 2019-09-23
v4.0.x: REF6976 Silent failure of OMPI over OFI with large messages sizes

mwheinz:
INTERNAL: STL-59403

(cherry-picked from commit 3aca4af548a3d781b6b52f89f4d6c7e66d379609)
Change-Id: Ie50445e5edfb0f30916de0836db0edc64ecf7c60

### DIFF
--- a/ompi/mca/mtl/ofi/help-mtl-ofi.txt
+++ b/ompi/mca/mtl/ofi/help-mtl-ofi.txt
@@ -16,3 +16,5 @@ unusual; your job may behave unpredictably (and/or abort) after this.
   Local host: %s
   Location: %s:%d
   Error: %s (%zd)
+[message too big]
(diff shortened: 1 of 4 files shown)


bwbarrett:
I don't think this is the right fix; we need to discuss more (see https://github.com/open-mpi/ompi/pull/7004).

gpaulsen:
@jsquyres added WIP-DNM until we could discuss at today's web-ex.

We were in agreement that the above PR is not "complete", as it does nothing to support larger message sizes (and MPI should be able to support all message sizes).  It was noted that this PR does improve the current situation by failing with an error message rather than silently.

@hppritcha and I (as release maangers for v4.0.x) have agreed to merge this for now, and if a better solution comes along consider that at that time.

PR #7010, opened 2019-09-25
Cleanup stale code in ORTE/OOB

rhc54:
Remove code for multiple OOB progress threads as it is an optimization
nobody uses. Also turns out to have a race condition that can cause
segfault on finalize, so maybe good that nobody is using it.


jsquyres:
I confirm that this is definitely causing a problem for me and at least one of my customers on the v3.1.x branch.  A cherry-pick PR will be coming shortly...

PR #7014, opened 2019-09-25
v3.0.x: Cleanup stale code in ORTE/OOB

rhc54:
Remove code for multiple OOB progress threads as it is an optimization
nobody uses. Also turns out to have a race condition that can cause
segfault on finalize, so maybe good that nobody is using it.

(cherry picked from commit 41eb41c3f224abcacec78f4d77c138197c36f172)
(cherry picked from commit 2198d22785e0c29716c099f87c1406280fd9c70e)


jsquyres:
Tested ok.  Waiting for @rhc's final approval (i.e., does this need the same additional cleanup that #7013 did?).

rhc54:
Already committed the extra cleanup, @jsquyres so this is good to go!

PR #7021, opened 2019-09-30
orte:set orte_oob_base ev_base to non-NULL

hppritcha:
need orte_oob_base ev_base to be set to some valid address to keep
libevent happy.  This partially reverts #7010.


rhc54:
This is not the correct fix - we pushed the correct fix to the v3.x branches but forgot to apply it here. Correct fix coming.

PR #7028, opened 2019-10-01
mtl/ofi: replace OMPI_UNLIKELY with OPAL version

jsquyres:
one off patch for v4.0.x.  for some reason commit on master
didn't have this problem.

(cherry picked from commit 5f3dbdb5c8a94a4f426ecca1a3a91c83035f956c)

Note that this commit is actually a cherry-pick from the v4.0.x
branch.  This is the opposite direction than what we nornmally do: we
usually commit to master first and then cherry-pick to the release
branches (vs. the other way around).

As is probably evident from the original commit message above, through
a comedy of errors, this commit was actually applied to the v4.0.x
branch first and then cherry-picked back to master (i.e., the problem
*did* exist in the original master commit
3aca4af548a3d781b6b52f89f4d6c7e66d379609, but it was not recongized at
the time).

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -2,6 +2,8 @@
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved
  * Copyright (c) 2017      Los Alamos National Security, LLC. All rights
  *                         reserved.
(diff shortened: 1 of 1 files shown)


mwheinz:
Looks good to me - not sure how to approve. Not sure why the broken version compiles on my machine.

PR #7034, opened 2019-10-03
Cpu set v40x

markalle:
This is a redo of
https://github.com/open-mpi/ompi/pull/6755
just for 4.x.

From our phone call about affinity my understanding is we want to keep the --cpu-set option where it functions like a lightweight cgroup.  And I'd prefer to consider this a bugfix that restores the old feature rather than an enhancement that's adding a new feature.

I admittedly had to go back to a pretty old OMPI tree (2.1.6) to find one that had that feature working.

The timeline was roughly
2.x : --cpu-set could be used like a cgroup, the implementation was complex and had the whole hwloc tree present, but every mapping option that involved iterating over the tree elements had extra logic in place to skip over elements that weren't inside the allowed mask
3.x : the core functionality of skipping over disallowed hwloc tree elements was still in place, but the error checking was made too aggressive so any command line that tried to use --cpu-set along side other misc binding/mapping options was rejected
4.x : with the switch to hwloc2 a bunch of code was simplified and the core functionality of skipping over disallowed elements was removed because it was unnecessary for cases like cgroups because the hwloc tree itself was being loaded without WHOLE_SYSTEM so it would already only contain the right elements

This PR changes the error checking back so that --cpu-set is allowed with the other mapping options, and uses hwloc_topology_restrict() to make the --cpu-set mask get used the same way that a cgroup would: the hwloc tree gets whittled to only contain elements in the allowed mask.

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_frame.c
+++ b/opal/mca/hwloc/base/hwloc_base_frame.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
  * Copyright (c) 2016-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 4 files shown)


gpaulsen:
Unfortunately this is too late for v4.0.2.

gpaulsen:
@markalle Is this more of a bugfix or a new feature.

gpaulsen:
We will close this PR soon.
Please port this PR to pmix/PRRTE which will be the default launcher for Open MPI v5.0

PR #6999, opened 2019-09-21
v3.1.x: Do not use CMA in user namespaces

jsquyres:
Trying out to run processes via mpirun in Podman containers has shown
that the CMA btl_vader_single_copy_mechanism does not work when user
namespaces are involved.

Creating containers with Podman requires at least user namespaces to be
able to do unprivileged mounts in a container

Even if running the container with user namespace user ID mappings which
result in the same user ID on the inside and outside of all involved
containers, the check in the kernel to allow ptrace (and thus
process_vm_{read,write}v()), fails if the same IDs are not in the same
user namespace.

One workaround is to specify '--mca btl_vader_single_copy_mechanism none'
and this commit adds code to automatically skip CMA if user namespaces
are detected and fall back to MCA_BTL_VADER_EMUL.

Signed-off-by: Adrian Reber <areber@redhat.com>
(cherry picked from commit fc68d8a90fe86284e9dc730f878b55c0412f01d2)

PR #7040, opened 2019-10-04
Updates for v3.0.5rc1

jsquyres:
@bwbarrett Some notes:

* See the commit messages for more detail (e.g., about c:r:a)
* I included a NEWS bullet for the CMA/vader issue (#6999), because I'm assuming we'll review+merge it before making the RC
* This PR is marked WIP/DNM because comparing the NEWS items of v3.0.x and v3.1.x, I see that the v3.0.x branch is missing a few fixes that made it to the v3.1.x branch (I marked these as "NO" in the NEWS file, which needs to be cleaned up before we merge).
    * Specifically: we need to decide whether we want to chase down these fixes for the v3.0.x branch.
    * My $0.02: barring technical problems / significant reasons *not* to include these fixes in the v3.0.x branch, we should chase them down / get the v3.0.x PRs filed.
    * I did find one v3.1.x PR that can **not** be applied to v3.0.x: the new `regx/naive` component on the v3.1.x branch.  There is no `regx` framework here on the v3.0.x branch, so that PR is moot over here.

Refs #6591

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -39,6 +39,50 @@ it is possible for items to appear more than once in the list.
 3.0.5 -- October, 2019
 ----------------------
 
+- Fix OMPIO issue limiting file reads/writes to 2GB.  Thanks to
+  Richard Warren for reporting the issue.
+- At run time, automatically disable Linux cross-memory attach (CMA)
+  for vader BTL (shared memory) copies when running in user namespaces
+  (i.e., containers).  Many thanks to Adrian Reber for raising the
+  issue and providing the fix.
+- Sending very large MPI messages using the ofi MTL will fail with
+  some of the underlying Libfabric transports (e.g., PSM2 with
+  messages >=4GB, verbs with messages >=2GB).  Prior version of Open
+  MPI failed silently; this version of Open MPI invokes the
+  appropriate MPI error handler upon failure.  See
+  https://github.com/open-mpi/ompi/issues/7058 for more details.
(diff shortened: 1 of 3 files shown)


jsquyres:
Per discussions with @bwbarrett, I tried cherry-picking fixes for the "NO"-labeled items.  They all required more than trivial effort except the buffered sends with the cm PML.

The following are the list of NEWS items that were *not* cherry-picked back, and therefore will not be fixed in the v3.0.x series:

* Fix MPI one-sided 32 bit atomic support.
* Allow individual jobs to set their map/rank/bind policies when running LSF.  Thanks to Nick R. Papior for assistance in solving the issue.
* Properly propagate errors to avoid deadlocks in MPI one-sided operations.
* Fix data corruption in non-contiguous MPI accumulates over UCX.

In particular, one-sided support changed significantly between v3.0.x and v3.1.x; it would not be simple to port those OSC changes back to v3.0.x.  The solution for users needing OSC fixes is to upgrade to v3.1.x or v4.0.x.

jsquyres:
@bwbarrett README updated:

* Added OPMIO bullet
* Re-wrote bullet about CM PML fix.
* Removed "NO" bullets.
* Waiting for #7059

jsquyres:
@bwbarrett #7059 is now merged.  Both the v3.0.x and v3.1.x README/NEWS/VERSION PRs are ready for your review so that we can make the RCs.

PR #7043, opened 2019-10-05
v4.0.x: usnic fixes and optimizations

jsquyres:
See individual commit messages for more details.

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic.h
+++ b/opal/mca/btl/usnic/btl_usnic.h
@@ -206,6 +206,10 @@ typedef struct opal_btl_usnic_component_t {
     /** retrans characteristics */
     int retrans_timeout;
 
+    /** minimum number of times through component progress before
+        checking to see if standalone ACKs need to be sent */
+    int ack_iteration_delay;
+
     /** transport header length for all usNIC devices on this server
         (it is guaranteed that all usNIC devices on a single server
         will have the same underlying transport, and therefore the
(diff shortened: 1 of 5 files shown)


gpaulsen:
Sorry, too late for v4.0.2, moving to v4.0.3

jsquyres:
@gpaulsen @hppritcha Can you approve and merge?  I have no one left in the OMPI community who can review usNIC stuff, so I got a co-worker to review it for me (but he doesn't have write perms to the OMPI repo, so his review "doesn't count" in terms of github's requirements).  

Thanks.

jsquyres:
Yes, I probably could/should have made that explanation better.

The real reason was that I merged the commits on master before getting the code review.  In the code review, it came out that it really wasn't necessary to increase the timeout -- 5 miliseconds is long enough already.  The real reason for (effectively) that timeout being too short was the lack of a cap on retransmitting during a single iteration of usNIC progress.  I.e., when sending many large messages to a very large number of peers, we could get into a situation where the retransmit queue got *so* long that it effectively took longer to retransmit everything than the retrans timers on the sends.  Hence, by the time we finished retransmitting everything, at least some of the timers had already expired, which would then lead to immediately retransmitting again, ...etc.  Meaning: the retrans timer wasn't the issue.  The amount of work we allowed to happen during a single iteration of usNIC progress was the problem.

gpaulsen:
Thanks @bwbarrett for reviewing.
@hppritcha and I haven't yet opened the v4.0.x for merging, but once we open that, we'll get this in.

hppritcha:
Ready to merge imho

PR #7050, opened 2019-10-06

### LINKED ISSUE #1089: need ability to not use --enable-new-dtags option in linker scripts.

hppritcha:
causes issues.

The new configury option is

--enable-wrappers-runpath

To disable use of --enable-new-dtags in the wrappers, add

--disable-wrappers-runpath

to the Open MPI configury line.

Fixes #1089

(cherry picked from commit ebb30c15f2a3808a51c94bf7e0f382ba096ade2f)


config/opal_setup_wrappers.m4
```diff
@@ -133,7 +133,7 @@ AC_DEFUN([OPAL_SETUP_WRAPPER_INIT],[
 
     AC_MSG_CHECKING([if want wrapper compiler runpath support])
     AC_ARG_ENABLE([wrapper-runpath],
-                  [AS_HELP_STRING([--enable-wrapper-runpath],
+                  [AS_HELP_STRING([--enable--wrapper-runpath],
```
ggouaillardet: are you sure about this?

config/opal_setup_wrappers.m4
```diff
@@ -227,12 +227,41 @@ AC_DEFUN([OPAL_SETUP_RUNPATH],[
                             AC_MSG_RESULT([yes (-Wl,--enable-new-dtags)])],
                            [AC_MSG_RESULT([no])])
             AC_LANG_POP([C])])
-    m4_ifdef([project_ompi],[
-        OPAL_LIBTOOL_CONFIG([wl],[wl_fc],[--tag=FC],[])
+m4_ifdef([project_ompi],[
```
ggouaillardet: the previous indentation was correct imho

jsquyres:
The commit on this PR doesn't look much like ebb30c1 at all -- it might be better to say that this commit is based off ebb30c1, rather than being a cherry-pick of ebb30c1.

jsquyres:
Ah!  Sorry for the false alarm, then... 😦

PR #7051, opened 2019-10-07
--report_bindings on whole machine, plus numa markers

markalle:
This PR combines a few related issues and enhancements on --report-bindings
1. --report-bindings on WHOLE_SYSTEM like 3.x did
2. mark unallowed (eg cgroup) parts of the whole machine with ~
3. optional numa markers
4. allow hwloc tree to not have sockets/cores
----

1. whole machine:

The topology sent to the pretty-print functions doesn't have the WHOLE_SYSTEM
flag, and in general we don't want WHOLE_SYSTEM.  But for pretty-printing I
think it makes more sense, and should be considered a regression vs 3.x where
the topologies had been loaded with WHOLE_SYSTEM and output looked more like
the second section in the below example.

Example of what pretty-printing looks like with/without whole system:

Suppose the machine is
```
    [..../..../..../....][..../..../..../....]
     0    4    8    12    16   20   24   28
```
and we run with
```
    cgset -r cpuset.cpus=24,25,28,29 mycgroup1
    cgset -r cpuset.cpus=26,27,30,31 mycgroup2
```
to leave only these hardware threads active:
```
    mycgroup1: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/..~~/..~~]
    mycgroup2: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/~~../~~..]
```
(here "~" is outside the cgroup and "." is in the allowed mask)

Without whole-system the printout would be
```
mycgroup1 (-np 2):
    MCW rank 0 bound to socket 1[core 0[hwt 0-1]]: [][BB/..]
    MCW rank 1 bound to socket 1[core 1[hwt 0-1]]: [][../BB]
mycgroup2 (-np 2):
    MCW rank 0 bound to socket 1[core 0[hwt 0-1]]: [][BB/..]
    MCW rank 1 bound to socket 1[core 1[hwt 0-1]]: [][../BB]
```

With whole-system the output is this, which I think is more informative
```
mycgroup1 (-np 2):
    MCW rank 0 bound to socket 1[core 6[hwt 0-1]]: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/BB~~/..~~]
    MCW rank 1 bound to socket 1[core 7[hwt 0-1]]: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/..~~/BB~~]
mycgroup2 (-np 2):
    MCW rank 0 bound to socket 1[core 6[hwt 2-3]]: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/~~BB/~~..]
    MCW rank 1 bound to socket 1[core 7[hwt 2-3]]: [~~~~/~~~~/~~~~/~~~~][~~~~/~~~~/~~../~~BB]
```

2. mark unallowed (~)

When using the whole-machine option there's a bitmask available to identify
the allowed PUs, eg omitting PUs not in our cgroup.  To distinguish those
PUs I'm using "~"

On the affinity phone call there was a question if we can distinguish "outside
the cgroup" vs "hardware configured off" or other states and I don't know of
hwloc info to distinguish those.  The current code uses
hwloc_topology_get_allowed_cpuset() so anything not in the allowed mask for
whatever reason gets the "~" character.

3. numa markers (<>)

I like having numa markers as well as the existing separators between
sockets and cores.  They're a little harder to code since the numas are more
fluid, eg sockets always contain cores not vice versa, so you can hard code a
loop over sockets follwed by a loop over cores. But numas might be be above or
below sockets in the tree.

This code identifies which level should be considered the child of the
numas, and has each of the hard coded loops capable of adding numa markers.

The feature is off by default and can be turned on with
   --mca rtc_hwloc_report_bindings_use_hwview 1
which puts the numa markers <> in and also shortens the output by omitting
the first part that looks like "socket 0[core 1[hwt 0-3]]" and only includes
the graphical view of the hardware.

```
Eg it shortens
> MCW rank 0 bound to socket 0[core 1[hwt 0-3]]: [BBBB/....][..../....]
to
> MCW rank 0 bound to [BBBB/....][..../....]
```

4. allow hwloc tree to not have sockets/cores

I may be behind the times on hwloc development, but as far as I know
hwloc trees aren't guaranteed to have sockets and cores, just a MACHINE
at the top and PU at the bottom. So I added a little code to the loops
so it would still print the PUs on a hypothetical machine that lacked any
structuring of the PUs into cores/sockets.

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -1680,6 +1680,26 @@ static int build_map(int *num_sockets_arg, int *num_cores_arg,
 }
 
 /*
+ * use a cached WHOLE_SYSTEM topology in cset2str / cset2mapstr
+ * if the input topology is this system (hwloc_topology_is_thissystem())
+*/
+static hwloc_topology_t cached_whole_system_topology = NULL;
+
+static int
+conditionally_modify_the_topology_to_whole_system(hwloc_topology_t *ptopo)
+{
+    if (hwloc_topology_is_thissystem(*ptopo)) {
+        if (cached_whole_system_topology == NULL) {
+            hwloc_topology_init(&cached_whole_system_topology);
+            hwloc_topology_set_flags(cached_whole_system_topology,
+                HWLOC_TOPOLOGY_FLAG_WHOLE_SYSTEM);
+            hwloc_topology_load(cached_whole_system_topology);
+        }
+        *ptopo = cached_whole_system_topology;
+    }
+}
+
+/*
  * Make a prettyprint string for a hwloc_cpuset_t
  */
 int opal_hwloc_base_cset2str(char *str, int len,
@@ -1694,6 +1714,8 @@ int opal_hwloc_base_cset2str(char *str, int len,
     int **map=NULL;
     hwloc_obj_t root;
     opal_hwloc_topo_data_t *sum;
+
+    conditionally_modify_the_topology_to_whole_system(&topo);
 
     str[0] = tmp[stmp] = '\0';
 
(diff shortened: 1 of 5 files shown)


gpaulsen:
Please cherry-pick with `-x` flag (along with the usual `-s`) from master commit to include the "Cherry Picked from <HASH> reference for easier auditing of release branches vs master.

markalle:
Geoffrey reminded me to note that this isn't going in master or 5.x because anything orte-related is radically changing there already.  Parts of this PR might end up still being relevant over there, but it will likely be very different there

opal/mca/hwloc/base/hwloc_base_util.c
```diff
@@ -1744,24 +1764,98 @@ int opal_hwloc_base_cset2str(char *str, int len,
     return OPAL_SUCCESS;
 }
 
+// given an input obj somewhere in the hwloc tree, look for a
```
hppritcha: Generally we don't want c++ style comments in OMPI.

orte/mca/rtc/hwloc/rtc_hwloc.h
```diff
@@ -45,7 +45,7 @@ typedef struct {
 ORTE_MODULE_DECLSPEC extern orte_rtc_hwloc_component_t mca_rtc_hwloc_component;
 
 extern orte_rtc_base_module_t orte_rtc_hwloc_module;
-
+extern bool orte_report_bindings_use_hwview;
```
hppritcha: this externally visible symbol may bother the ABI compatibility crew,  but I'm okay with this.

markalle: I'm not really wanting it to become an external part of the ABI, it's just used inside the component. But its use spans multiple files within the component.

I don't use mca_base_var_get_value() much, but without putting it in a global I think that would be the only way to access it in the other file

markalle:
repushed with comments changed to /* */

gpaulsen:
We will close this PR soon.  No further action is needed as this PR is already in master.

PR #7053, opened 2019-10-08
v3.1: pml_ucx: add ompi datatype attribute to release ucp_datatype - v3.1

hoopoepg:
Fixes openucx/ucx#2921
(checked that the leak is gone with IMB Allgatherv 512b, and passed mpi_test_suite)

backport from #5878


jsquyres:
Please be sure to assign relevant labels and milestones when filing PRs on release branches -- if you don't, you risk them not being noticed by the release managers.  Thanks.

PR #7060, opened 2019-10-08
BTL usnic MCA updates

jsquyres:
See individual commit messages for more detail.


jsquyres:
These fixes have been confirmed.

PR #7061, opened 2019-10-09
pml/ucx: correctly handle zero size datatypes

ggouaillardet:
zero-size derived datatypes are now flagged as OPAL_DATATYPE_FLAG_CONTIGUOUS
so update mca_pml_ucx_init_datatype() to correctly handle them

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx_datatype.c
+++ b/ompi/mca/pml/ucx/pml_ucx_datatype.c
@@ -1,5 +1,7 @@
 /*
  * Copyright (C) Mellanox Technologies Ltd. 2001-2011.  ALL RIGHTS RESERVED.
+ * Copyright (c) 2019      Research Organization for Information Science
+ *                         and Technology (RIST).  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -155,7 +157,6 @@ ucp_datatype_t mca_pml_ucx_init_datatype(ompi_datatype_t *datatype)
         (lb == 0))
     {
         ompi_datatype_type_size(datatype, &size);
-        PML_UCX_ASSERT(size > 0);
         datatype->pml_data = ucp_dt_make_contig(size);
         return datatype->pml_data;
     }


ggouaillardet:
@yosefe can you please review this ?

If UCX has a *predefined* zero-size datatype, this PR could be improved by using it instead of invoking `ucp_dt_make_contig()`.


A simple reproducer is inlined below (it only requires one MPI task).

```c
#include <stdio.h>

#include <mpi.h>

int main(int argc, char *argv[]) {
    MPI_Datatype ddt;

    MPI_Init(&argc, &argv);
    MPI_Type_contiguous(0, MPI_INT, &ddt);
    MPI_Type_commit(&ddt);
    MPI_Sendrecv(NULL, 1, ddt, 0, 0,
                 NULL, 1, ddt, 0, 0,
                 MPI_COMM_SELF, MPI_STATUS_IGNORE);
    MPI_Type_free(&ddt);
    MPI_Finalize();
    return 0;
}
```

ggouaillardet:
@bosilca FYI (I guess this issue is a side effect of a recent open-mpi/ompi@35229169716d6ede1e172915c8bd87d02648e45c).

ompi/mca/pml/ucx/pml_ucx_datatype.c
```diff
@@ -155,7 +157,7 @@ ucp_datatype_t mca_pml_ucx_init_datatype(ompi_datatype_t *datatype)
         (lb == 0))
     {
         ompi_datatype_type_size(datatype, &size);
-        PML_UCX_ASSERT(size > 0);
+        PML_UCX_ASSERT(size >= 0);
```
brminich: size is `size_t`, so this assert does not make sense now - it will always be positive.

ggouaillardet:
thanks @brminich I updated the PR.

bosilca:
What is the purpose of the ```0 == lb``` test ?

gpaulsen:
@yosefe ^^^ ?

yosefe:
if lb!=0 then an array of that datatype may have gaps (even if a single element does not have gaps)

gpaulsen:
I PRed to v4.0.x here: https://github.com/open-mpi/ompi/pull/7161

PR #7063, opened 2019-10-09
coll/basic: fix neighbor alltoall message ordering

hjelmn:
This commit updates the coll/basic component to correctly order sends
and receives for cartesian communicators with cyclic boundaries. This
addresses an issue identified by mpi-forum/mpi-issues#153. This issue
occurs when the size in any dimension is 1. This gives the same
neighbor in the positive and negative directions. The old code was
sending and receiving in the same order so the -1 buffer contained
the +1 result and vise-versa. The problem is addressed by using
unique tags for each send. This should cover both the case where
overtaking is allowed and is not allowed. The former case will be
possible if a MPI_Cart_create_with_info() call is added to the
standard.

### DIFF
--- a/ompi/mca/coll/basic/coll_basic_neighbor_alltoall.c
+++ b/ompi/mca/coll/basic/coll_basic_neighbor_alltoall.c
@@ -15,6 +15,7 @@
  * Copyright (c) 2014-2015 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2017      IBM Corporation. All rights reserved.
+ * Copyright (c) 2019      Google, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -67,7 +68,7 @@ mca_coll_basic_neighbor_alltoall_cart(const void *sbuf, int scount, struct ompi_
         if (MPI_PROC_NULL != srank) {
             nreqs++;
             rc = MCA_PML_CALL(irecv(rbuf, rcount, rdtype, srank,
-                                    MCA_COLL_BASE_TAG_ALLTOALL,
+                                    MCA_COLL_BASE_TAG_NONBLOCKING_BASE - 2 * dim,
                                     comm, preqs++));
             if (OMPI_SUCCESS != rc) break;
         }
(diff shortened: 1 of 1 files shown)


hjelmn:
This doesn't fix the non-blocking variant. There may be additional issues with the other neighbor collectives.

hjelmn:
@bosilca Can you please review.

RolfRabenseifner:
First, the problem exists also for 2 processes in a direction (not only for 1).
Second, I would be very surprised if MPI_Cart_create_with_info() should allow this very wrong behavior.

ompi/mca/coll/basic/coll_basic_neighbor_alltoall.c
```diff
@@ -67,7 +68,7 @@ mca_coll_basic_neighbor_alltoall_cart(const void *sbuf, int scount, struct ompi_
         if (MPI_PROC_NULL != srank) {
             nreqs++;
             rc = MCA_PML_CALL(irecv(rbuf, rcount, rdtype, srank,
-                                    MCA_COLL_BASE_TAG_ALLTOALL,
+                                    MCA_COLL_BASE_TAG_NONBLOCKING_BASE - 2 * dim,
```
bosilca: This tag range was reserved (as the name indicates) for nonblocking collectives. If we are using it for something else (as we should in this particular case) I would suggest we rename it to something more meaningful.

RolfRabenseifner:
Be very careful: Two nonblocking MPI_NEIGHBOR_ALLTOALL still need for the same tags that the messages from process to process must not overtake.
In general, when implementing collective communication internally with MPI point-to-point then the overtaking must be switched off, which may be not a problem if the communicator is duplicated without duplicationg this info about overtaking is allowed.

bosilca:
@RolfRabenseifner, as indicated in Nathan's comment this patch only applies to blocking collectives.

hjelmn:
@bosilca Yup. The non-blocking fix is much more complicated. This is low hanging fruit.

bosilca:
@hjelmn we might need to have more well-defined ranges for collectives, and have a sliding window of tags to prevent the issues with concurrent collectives.

ompi/mca/coll/basic/coll_basic_neighbor_alltoall.c
hjelmn: Yeah. We can add another reserved range to be safe. I can add that to this PR.

jsquyres:
This PR was replaced by https://github.com/open-mpi/ompi/pull/7232.  Closing without merging.

PR #7065, opened 2019-10-09
oshmem: fix race condition on new contexts

janjust:
Do not add private contexts to active list, they need only be visible to user

Signed-off-by: Artem Y. Polyakov <artemp@mellanox.com>

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx.c
+++ b/oshmem/mca/spml/ucx/spml_ucx.c
@@ -573,19 +573,17 @@ static inline void _ctx_add(mca_spml_ucx_ctx_array_t *array, mca_spml_ucx_ctx_t 
     array->ctxs_count++;
 }
 
-static inline void _ctx_remove(mca_spml_ucx_ctx_array_t *array, mca_spml_ucx_ctx_t *ctx)
+static inline void _ctx_remove(mca_spml_ucx_ctx_array_t *array, mca_spml_ucx_ctx_t *ctx, int i)
 {
-    int i;
-
-    for (i = 0; i < array->ctxs_count; i++) {
+    for (; i < array->ctxs_count; i++) {
         if (array->ctxs[i] == ctx) {
             array->ctxs[i] = array->ctxs[array->ctxs_count-1];
             array->ctxs[array->ctxs_count-1] = NULL;
+            array->ctxs_count--;
             break;
         }
     }
 
-    array->ctxs_count--;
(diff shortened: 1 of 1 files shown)


janjust:
@jladd-mlnx fyi

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -698,6 +698,11 @@ int mca_spml_ucx_ctx_create(long options, shmem_ctx_t *ctx)
         opal_progress_register(spml_ucx_ctx_progress);
     }
 
+    if (options & SHMEM_CTX_PRIVATE) {
```
brminich: minor: could avoid code duplication if handle ` if (!(options & SHMEM_CTX_PRIVATE)) ` instead

hppritcha:
bot:ompi:retest

artpol84:
@brminich the race is when the private context that is not lock-protected is being progressed bo some other thread as part of opal progress while the thread that owns this private context is accessing it.

oshmem/mca/spml/ucx/spml_ucx.c
artpol84: Good point, @janjust you can put 
https://github.com/open-mpi/ompi/pull/7065/files#diff-8af288f2560951b1798fcbf1131caa78L701-L703
into the if statement as 
https://github.com/open-mpi/ompi/pull/7065/files#diff-8af288f2560951b1798fcbf1131caa78R702-R703
and https://github.com/open-mpi/ompi/pull/7065/files#diff-8af288f2560951b1798fcbf1131caa78R710-R711
are essentially the same.

artpol84:
Per discussion with @manjugv:
A private context should only be progressed by the thread that has created it. This allows avoiding locking around it to get the best performance.
This PR excludes private contexts from being included in the list of contexts to be processed during opal progress so it won't be implicitly progressed by an arbitrary thread calling progress.

janjust:
PR isn't ready yet, don't merge

artpol84:
@brminich please have a look.
This PR addresses 2 issues now:
* Race condition for PRIV contexts
* Memory leak when a large number of contexts is created/destroyed. Now we reuse released contexts when possible.

janjust:
@artpol84 pushed the latest, and tested

hppritcha:
@janjust is this PR ready to merge now?

janjust:
@hppritcha yes

PR #7077, opened 2019-10-10
Update README

tsgxcode:
signed off by Scott Georg


tsgxcode:
signed off by Scott Georg

rhc54:
Couple of things:
* all commits have to be "signed off by" according to the directions:
"Signed-off-by: Name <email>"

* I understand and applaud the "hacktoberfest", but please don't submit silly things like this.

PR #7082, opened 2019-10-10
Fix ipv6 improper address copy bug

kizill:
Found connection bug while using IPv6 - only 4 bytes were copied.


gpaulsen:
ok to test

jsquyres:
Seems like this should be cherry picked to v4.0.x, v3.1.x, and v3.0.x branches, too.

jsquyres:
@kizill Could you make a PR to pull this into the v4.0.x branch?  (and optionally PR's the v3.1.x and v3.0.x branches)   We usually use `git cherry-pick -x ...` to bring commits from master to release branches.  I don't know if this commit will apply cleanly to these release branches or not (e.g., have we brought all the relevant IPv6 fixes to v4.0.x?).

PR #7103, opened 2019-10-22
osc/ucx: Fix wpool use-after-free

### LINKED ISSUE #7101: opal/common/ucx: Simple one-sided test case fails with UCX
### Distribution used 
ompi - master and ucx - master (clone)

**UCX configuration: (master branch)**
```
./autogen.sh
./contrib/configure-release-mt --without-java --prefix=$shared_dir/ucx-install --with-cuda=/usr/local/cuda
make 
make install 
```
**OMPI configuration: (master branch)**
```
./autogen.pl
./configure --enable-mca-no-build=btl-uct --enable-mpi1-compatibility --prefix $shared_dir/install --with-cuda=/usr/local/cuda --with-ucx=$shared_dir/ucx-install 
make
make install
```

* Operating system/version: RHEL 7.6
* Computer hardware: ppc64le (POWER8NVL)
* Network type: IB, shmem

-----------------------------

## Details of the problem

Here is the test case which I'm using:
https://gist.github.com/AboorvaDevarajan/287067606ade089cc3499ab34d703311

```
# mpicc simple_win_test.c -o simple_win_test
# $MPI_ROOT/bin/mpirun --timeout 600 -np 2  -mca backtrace_lwcore_enable t --get-stack-traces --report-state-on-timeout -mca plm rsh --prefix $MPI_ROOT --mca osc ucx,sm --mca pml ucx --mca btl ^vader,tcp,openib,uct --mca opal_common_ucx_opal_mem_hooks 1 -x UCX_MEM_MMAP_HOOK_MODE=none -x LD_LIBRARY_PATH ./simple_win_test 
```

Once in a while the above test case fails with this error:

```
[1571571024.385372] [c712f6n05:127020:0] ucp_mm.c:116  UCX  ERROR failed to register address 0x10001d30adf0 length 8192 on md[5]=cuda_cpy: Input/output
```
Here is some debug info which I got so far:

During `MPI_Win_create` for the above configuration we take up the `cuda_copy` (ucx->uct) path and do a `cudaHostRegister` of the memory, and the corresponding `cudaHostUnregister` is not invoked in `MPI_Win_free` but it's deferred to a later point,
https://github.com/open-mpi/ompi/blob/master/opal/mca/common/ucx/common_ucx_wpool.c#L1008
if the same memory is alllocated during `MPI_Win_create` in the next iteration `cudaHostRegister` will fail with error no: 712 which is `cudaErrorHostMemoryAlreadyRegistered`.

I have one question:

Do we always take the cuda_copy path when we configure OMPI / UCX with CUDA enabled? Is there a way to disable it for applications with just host buffers?

yosefe:
Fixes #7101

### DIFF
--- a/ompi/mca/osc/ucx/osc_ucx_component.c
+++ b/ompi/mca/osc/ucx/osc_ucx_component.c
@@ -696,6 +696,7 @@ int ompi_osc_ucx_free(struct ompi_win_t *win) {
     }
 
     opal_common_ucx_wpctx_release(module->ctx);
+    mca_osc_ucx_component.env_initialized = false;
 
     if (module->disp_units) free(module->disp_units);
     ompi_comm_free(&module->comm);


yosefe:
already fixed in latest ompi master

artpol84:
@yosefe

> already fixed in latest ompi master

This PR was merged into the master. Can you clarify this sentence?

PR #7104, opened 2019-10-22
Topic/gpfs

edgargabriel:
Introduce a gpfs fs component. This component is based on work done a couple of years back by the HLRS, but has been updated to the new layout of the fs framework, and adjusted to more recent gpfs versions and the interface adjustements.


hppritcha:
@edgargabriel cray CI failure looks legit.

PR #7105, opened 2019-10-24
fortran/use-mpi-f08: revamp mpi_f08 constants

ggouaillardet:


ibm-ompi:

 Gist: https://gist.github.com/74a7bfe316f0faefe1cd1f86af2950be

ibm-ompi:

 Gist: https://gist.github.com/745ac02704122a3a661933be065be3d0

ompi/mpi/fortran/use-mpi-f08/mod/mpi-f08-types.F90
```diff
@@ -78,47 +79,47 @@ module mpi_f08_types
   ! Pre-defined handles
   !
 
-  type(MPI_Comm),       bind(C, name="ompi_f08_mpi_comm_world") OMPI_PROTECTED      :: MPI_COMM_WORLD
-  type(MPI_Comm),       bind(C, name="ompi_f08_mpi_comm_self") OMPI_PROTECTED        :: MPI_COMM_SELF
+  type(MPI_Comm), parameter       :: MPI_COMM_WORLD              = MPI_Comm(OMPI_MPI_COMM_WORLD)
+  type(MPI_Comm), parameter       :: MPI_COMM_SELF               = MPI_Comm(OMPI_MPI_COMM_SELF)
```
jsquyres: So is the overall idea of this PR to take what used to be instances declared in C and turn them into pure Fortran variables?

E.g., MPI_COMM_SELF becomes a parameter (which, IIRC, effectively means constant, right?), and initially define it to be equal to a new `MPI_Comm()` object, essentially initialized with `OMPI_MPI_COMM_SELF`?

If so, is it correct to assume that this is standard F08 notation, and is not specific to any particular compiler?

ggouaillardet:
- I did adjust `.gitignore`
 - I will double check the common whitelist symbols
 - the first commit preserve `constants.c` and should hence somehow preserve (read not break) ABI compatibility. the second commit does break it and cannot be backported to the release branches.
 - a MPI F08 type is an `INTEGER` wrapped into a type, and it can be constant here, so yes, I moved its definition from C to Fortran using the syntax I found in MPICH. I can only guess this is a standard F08 notation (it works with gfortran and flang, will try ifort tomorrow)

jsquyres:
My bad -- I missed the `.gitignore` update.

Oh, I didn't look at the 2 commits separately -- I just looked at the union and saw that `constants.c` was deleted.

That's a fairly sneaky way to preserve ABI.  👏  Can you add some comments about that?  E.g., if the plan is to bring the just the 1st commit to release branches, it would be good to make a note about this in the code somewhere -- Future Jeff and Future Gilles will thank you.

Also, just to be 100% sure: this doesn't appear to affect any of the "special" constants, like `MPI_BOTTOM`, `MPI_ARGV_NULL`, ...etc.  So therefore there's no need to adjust any of the C code comparing against these values (i.e., looking to see if Fortran passed in `MPI_BOTTOM` or its friends), right?

BTW, I hereby forbid anyone from running `git blame` on the code where C-style comments were put in the Fortran header file.

ggouaillardet:
I made the changes (first commit message and common symbols whitelist) and checked it works fine with `ifort`.

`constants.h` (with the `C` style comments) was used by `constants.c`.
I *duplicated* it into `mpi-f08-constants.h` (used in mpi-f08-types.h) and moved the style comments to `Fortran` (fun fact, `gfortran` was happy with `C` style comments, but not IBM compilers)

jsquyres:
While reading your latest this morning, I wondered if `OMPI_PROTECTED` should also be removed, since its use now seems to be defunct.

However, I see that the "shortfloat" MPI Extension is using the same technique (i.e., it's using `OMPI_PROTECTED` and the same "let's bind Fortran values to C instances" method).  Do you want to extend this PR to:

- Remove `OMPI_PROTECTED`
- Fix up shortfloat in the same way you fixed up the main bindings

?

ggouaillardet:
@jsquyres I fixed the `shortfloat` extension.

I also removed `OMPI_PROTECTED` since it is no more relevant ... except it is used in the `ompi_info` output, so we might not want to backport this specific commit to the release branches.

PR #7114, opened 2019-10-28
COLL/TUNED: Add linear scatter using isend for mlnx platform

brminich:
Add scatter algorithm using non-blocking sends with "pseudo" sync by blocking sends.
By default enable for Mellanox platform only. Default selection should not affect other platforms.
Provides significant performance benefits (up to 2x for certain message and comm sizes), tested up to 2048 ranks. 

### DIFF
--- a/contrib/platform/mellanox/optimized.conf
+++ b/contrib/platform/mellanox/optimized.conf
@@ -10,6 +10,7 @@
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
 # Copyright (c) 2006      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2019      Mellanox Technologies. All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -84,4 +85,8 @@ bml_r2_show_unreach_errors = 0
 coll_tuned_alltoall_large_msg              = 250000
 coll_tuned_alltoall_min_procs              = 2048
 coll_tuned_alltoall_algorithm_max_requests = 8
+coll_tuned_scatter_intermediate_msg        = 8192
+coll_tuned_scatter_large_msg               = 250000
+coll_tuned_scatter_min_procs               = 1048510
+coll_tuned_scatter_algorithm_max_requests  = 64
 
--- a/ompi/mca/coll/base/coll_base_scatter.c
+++ b/ompi/mca/coll/base/coll_base_scatter.c
@@ -14,6 +14,7 @@
  *                         reserved.
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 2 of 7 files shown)


brminich:
@hoopoepg, pls review

brminich:
@yosefe

brminich:
There is important difference between `send` and `isend` operations with UCX pml. Blocking send uses `ucp_tag_send_nbr` routine, which uses eager protocol up to 256K by default. It also does not use eager zcopy protocol. This is done this way for `send` operation to complete faster on the sender side. Nonblocking send uses `ucp_tag_send_nb` routine, which typically starts using rendezvous much earlier ( at ~16-32K). Also it uses eager zcopy protocol avoiding one memory copy on the sender.  Even though `isend` may take more time to complete on the sender, with scatter we initiate a bunch of `isend` simultaneously, overlapping their costs. 
I used regular `send` for synchronization to "flush" local HW rather than sync with any of the peers.
With UCX we post send ops to the NIC until HW(QP) is full. After that we start adding `isend` operations to the pending queue (waiting for HW resources). Calling blocking send once in while should help to reduce/minimize pending queue usage.
  
BTW, I also tried sync send and having just bunch of outgoing requests, but regular blocking send seem to perform better considering various ppns and nps (tested on 64 Broadwells + CX5)

bosilca:
It looks as if this is specialized behavior for UCX support, and thus has no place in the base collective. 

Moreover, with a eager so high I hope you have support for dynamically decreasing it when the pressure is built on the receiver side.

brminich:
Please note that this algo is selected by default for mellanox platform only (when compiled with mellanox optimization).
Also initiating many `isends` at the same time may be beneficial comparing to calling blocking sends in a row on other platforms as well (even without any sync). IMO, this algo is generic enough

bosilca:
Indeed, the alg is disabled by default. Thanks for pointing it out.

ompi/mca/coll/base/coll_base_scatter.c
```diff
@@ -273,5 +274,106 @@ ompi_coll_base_scatter_intra_basic_linear(const void *sbuf, int scount,
...
+    for (i = 0, ptmp = (char *)sbuf, preq = reqs; i < size; ++i, ptmp += incr) {
+        /* simple optimization */
+        if (i == rank) {
+            if (MPI_IN_PLACE != rbuf) {
+                err = ompi_datatype_sndrcv(ptmp, scount, sdtype, rbuf, rcount,
+                                           rdtype);
+            }
+        } else {
+            if (!max_reqs || (i % max_reqs)) {
+                err = MCA_PML_CALL(isend(ptmp, scount, sdtype, i,
```
yosefe: seems like we actually can have much more than max_reqs in flight..

bosilca: Good catch, the code writes outside the bounds of preq (the array of requests). This code works only when coll_tuned_scatter_algorithm_max_requests is less or equal to 1.

brminich: @yosefe , what do you mean? Yes, we can, we just invoke blocking send every `max_reqs` 

@bosilca, in what case does it write outside the bounds? We allocate requests for every rank, except 'blocking send' cases, which is `size/max_reqs`.   
 It works with up to `size` max_reqs (was tested up to 256 max_reqs, 2048 ranks)

brminich:
> Fix the usage of the array of requests.

what is the problem with usage?

ompi/mca/coll/base/coll_base_scatter.c
bosilca: let's assume max_reqs is 20 on a larger communicator, enough to make nreqs < (size - 1) - 1. Due to the usage of the % in the if on line 337,  all requests over max_reqs will  be sent as non-blocking and will be stored on preq (after it has been ++ by more than max_reqs). So you are writing outside the bounds you were granted on the preq array. It might happen that it  seems to be working simply because internally the request management allocates more than you requested, so you might not write outside the internally allocated array (only outside the range you were provided).

brminich: Please check request allocation:
```
    if (max_reqs <= 1) {
        max_reqs = 0;
        nreqs = size - 1; /* no send for myself */
    } else {
        nreqs = size - (size / max_reqs); /* avoid allocating too many reqs */
    }
```

brminich: Should I rename mca `coll_tuned_scatter_algorithm_max_requests` to `coll_tuned_scatter_algorithm_blocking_send_ratio`?

bosilca: Oh my, I now understand the real meaning of ```nreqs = size - (size / max_reqs)```, allocate space for all non-blocking requests and skip the one for the blocking one.

brminich: sorry for the confusion, I just tried to re-use some mca name used by other collectives.
Do you tnink better rename it?

bosilca: names are fine but I would add a comment explaining the nreqs usage.

brminich:
@bosilca, @jsquyres, how can I re-trigger `Signed-off-by checker`?  I pushed 2 signed commits, but this check is pending for several day already

jsquyres:
@brminich There's unfortunately no easy user-accessible way to trigger it again other than pushing a new commit.

I triggered it on the back-end, and now it appears to be good.

brminich:
thanks, @jsquyres!
@bosilca, @yosefe, do you have other comments?

bosilca:
nope. I will re-approve the PR.

bosilca:
That being said I would definitively be interested to see if this might help outside the very specific UCX case.

contrib/platform/mellanox/optimized.conf
```diff
@@ -84,4 +85,8 @@ bml_r2_show_unreach_errors = 0
 coll_tuned_alltoall_large_msg              = 250000
 coll_tuned_alltoall_min_procs              = 2048
 coll_tuned_alltoall_algorithm_max_requests = 8
+coll_tuned_scatter_intermediate_msg        = 8192
+coll_tuned_scatter_large_msg               = 250000
+coll_tuned_scatter_min_procs               = 2048
```
yosefe: maybe set this to <inf>? any reason we want to change to basic linear on large scale?

ompi/mca/coll/tuned/coll_tuned_scatter_decision.c
```diff
@@ -114,6 +116,38 @@ ompi_coll_tuned_scatter_intra_check_forced_init(coll_tuned_force_algorithm_mca_p
...
+                                           "scatter_min_procs",
+                                           "use basic linear algorithm for communicators larger than this value",
+                                           MCA_BASE_VAR_TYPE_INT, NULL, 0, 0,
+                                           OPAL_INFO_LVL_6,
+                                           MCA_BASE_VAR_SCOPE_READONLY,
+                                           &ompi_coll_tuned_scatter_min_procs);
+
+    (void)mca_base_component_var_register(&mca_coll_tuned_component.super.collm_version,
+                                          "scatter_algorithm_max_requests",
+                                          "Maximum number of outstanding send requests. Only has meaning for synchronized algorithms.",
```
yosefe: "Issue a blocking send every this many non-blocking requests"

brminich:
@yosefe, comments applied

yosefe:
@brminich can you pls squash?

Issue #7091, opened 2019-10-15
Wrong values for MPI_Datatype with armflang


alinask:
Thank you for taking the time to submit an issue!

## Background information
MPI_Datatype is not handled correctly with armflang.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
OMPI branch v4.0.x 
commit 9f4b7ba5087134e1848f1c6508acb2485adc969d

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
From git clone.

### Please describe the system on which you are running

* Operating system/version:  Red Hat Enterprise Linux Server release 7.6 (Maipo)
* Computer hardware: ARM
Architecture:          aarch64
Cavium ThunderX2(R) CPU CN9975 v2.1 @ 2.0GHz 
* Network type: IB

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

A Fortran simple benchmark fails to run with ompi that was compiled with an Arm Fortran compiler:
```
$./install/bin/mpifort --version
Arm C/C++/Fortran Compiler version 19.0 (build number 69) (based on LLVM 7.0.2)
Target: aarch64--linux-gnu
Thread model: posix

$./install/bin/mpifort --showme
armflang -I/path/to/ompi/install/include -I/path/to/ompi/install/lib -Wl,-rpath -Wl,/path/to/ompi/install/lib -Wl,--enable-new-dtags -L/path/to/ompi/install/lib -lmpi_usempif08 -lmpi_usempi_ignore_tkr -lmpi_mpifh -lmpi
```
Please note the armflang in the output above.

A debug of the failing benchmark led to the following:
The MPI_Datatype in the MPI_Send and MPI_Recv calls was incorrect.
The data type that was used in the test is MPI_INTEGER.
Due to this, the following fortran program is a simple reproducer of the problem:
```
$cat example.f90
program example_mpif08
    use mpi_f08

    print *, 'MPI_INTEGER = ', MPI_INTEGER, 'MPI_INTEGER4 = ', MPI_INTEGER4
end program example_mpif08
```
Compile:
```
$./install/bin/mpifort example.f90 -o example
```
Run:
```
$./example
[...]

PR #6174, opened 2018-12-11
fortran/use-mpi-f08: Add C++ datatypes and MPI_NO_OP

kawashima-fj:
The following handles were missing in Fortran 2008 bindings.

- `MPI_CXX_BOOL`
- `MPI_CXX_COMPLEX`
- `MPI_CXX_FLOAT_COMPLEX`
- `MPI_CXX_DOUBLE_COMPLEX`
- `MPI_CXX_LONG_DOUBLE_COMPLEX`
- `MPI_NO_OP`

Though the MPI standard does not have `MPI_CXX_COMPLEX`, `mpi.h`, `mpif.h`, and `mpi.mod` have it. So I added it to `mpi_f08.mod` for consistency.

Need PRs to release branches.

PR #7116, opened 2019-10-29
v4.0.x: fortran/use-mpi-f08: revamp mpi_f08 constants

ggouaillardet:


ggouaillardet:
@jsquyres can you please review this?

I incidentally noted some stuff is missing. For example `type(MPI_Op) :: MPI_NO_OP` is defined in the `master` branch but not in the `v4.0.x` branch. Is this something intentional ?

jsquyres:
Have you tested that this doesn't break ABI w.r.t. all the constants changing types?

I can't think of a reason why `MPI_NO_OP` wouldn't be here on v4.0.x.  I wonder if that's a prior mistaken conflict resolution on a cherry pick to v4.0.x, or something along those lines...?

kawashima-fj:
@ggouaillardet I added `MPI_NO_OP` of `mpi_f08` in #6174 but probablly forgot to cherry-pick it to release branches. Can you checy-pick (or manually add without an additional commit) it to this PR?

hppritcha:
Why do we want this in v4.0.x?  Aside from the discussion about missing ```MPI_NO_OP```, is it just to get the f08 bindings to work with the old llvm fortran front end?  It does look like #6174 should be pulled in however.

jsquyres:
@hppritcha This is in reference to #7091 (i.e., bug fix for the ARM complier).

hppritcha:
@shamisp what's allinea's opinion of this flang problem?

jsquyres:
@ggouaillardet We talked about this yesterday on the RM call.  Seems like it's two issues:

1. MPI_NO_OP: That should definitely come over to v4.0.x.
1. The flang fixes: Do we know if this is desired and/or necessary on the v4.0.x branch?  I.e., we know it's a problem for flang, but does anyone care (given that -- at least as we understand it -- flang is old / may not be used much longer)?
1. Do we know for sure that the changes don't impact ABI?  I.e., did you test with an app compiled against v4.0.x and then try to run it with this PR?  And what happens if we have some objects compiled with v4.0.x and some objects compiled with this PR -- does it all work properly (i.e., can an MPI_COMM_WORLD from one of the old v4.0.0 objects be passed to an object compiled with this PR)?

ggouaillardet:
1. yes, and I also backported #5811 (it seems it has been forgotten a while ago)

2. I have some hard time undertanding this argument.
 - `armflang` and maybe PGI are `flang` based
 - `f18` is still relying on `pgf90` to work

 so yes, the plan is to move away from `flang` and use `f18` instead, but we are not quite there, so I'd rather revamp Open MPI (assuming it is impact free) than waiting an unknown amount of time for `flang`, ARM and PGI to fix their compilers.

3. the plan was to be ABI backward compatible, but I will run some more tests from now.

ggouaillardet:
I made some tests and found no ABI issues.

jsquyres:
Ok, good to know.  The discussion on the Monday+Tuesday Webexes was that `flang` was no longer relevant; if that's inaccurate, then it does change the equation.

@shamisp Can someone from ARM please chime in here?

ggouaillardet:
I discussed flang vs f18 at SC, and I can confirm f18 is not quite yet ready to replace flang (f18 currently requires `pgf90` to work). Also, armflang and pgf90 are currently based on flang, and we can expect it will take some extra time before they rebase on top of f18. Last but not least, we have no guarantee that f18 will behave differently from flang on that matter.
My best bet is to move forward is merge this PR.
I can also PR vs the `v3.1.x` release branch if needed.

gpaulsen:
@hppritcha I'm ready to merge this to v4.0.x as @ggouaillardet has tested this for ABI compatibility issues, and found none.

Issue #7109, opened 2019-10-26
deadlock running netcdf test with openmpi 4.0.2


opoplawski:
### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
4.0.2

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
Fedora packages

### Please describe the system on which you are running

* Operating system/version: Fedora 31 and 32
* Network type: localhost

-----------------------------

## Details of the problem
netcdf tst_parallel3 program hangs.  Backtrace shows:
```
(gdb) bt
#0  0x00007f90c197529b in sched_yield () from /lib64/libc.so.6
#1  0x00007f90c1ac8a05 in ompi_request_default_wait () from /usr/lib64/openmpi/lib/libmpi.so.40
#2  0x00007f90c1b2b35c in ompi_coll_base_sendrecv_actual () from /usr/lib64/openmpi/lib/libmpi.so.40
#3  0x00007f90c1b2bb73 in ompi_coll_base_allreduce_intra_recursivedoubling () from /usr/lib64/openmpi/lib/libmpi.so.40
#4  0x00007f90be96e9c5 in mca_fcoll_vulcan_file_write_all () from /usr/lib64/openmpi/lib/openmpi/mca_fcoll_vulcan.so
#5  0x00007f90be9fada0 in mca_common_ompio_file_write_at_all () from /usr/lib64/openmpi/lib/libmca_common_ompio.so.41
#6  0x00007f90beb0610b in mca_io_ompio_file_write_at_all () from /usr/lib64/openmpi/lib/openmpi/mca_io_ompio.so
#7  0x00007f90c1af033f in PMPI_File_write_at_all () from /usr/lib64/openmpi/lib/libmpi.so.40
#8  0x00007f90c1627d7b in H5FD_mpio_write () from /usr/lib64/openmpi/lib/libhdf5.so.103
#9  0x00007f90c14636ee in H5FD_write () from /usr/lib64/openmpi/lib/libhdf5.so.103
#10 0x00007f90c1442eb3 in H5F__accum_write () from /usr/lib64/openmpi/lib/libhdf5.so.103
#11 0x00007f90c1543729 in H5PB_write () from /usr/lib64/openmpi/lib/libhdf5.so.103
#12 0x00007f90c144d69c in H5F_block_write () from /usr/lib64/openmpi/lib/libhdf5.so.103
[...]

PR #7122, opened 2019-10-29
common/ompio: fix calculation in simple-grouping option

edgargabriel:
This is based on a bug reported on the mailing list using a netcdf testcase.
The problem occurs if processes are using a custom file view, but on some
of them it appears as if the default file view is being used. Because of that,
the simple-grouping option lead to different number of aggregators used on different
processes, and ultimately to a deadlock. This patch fixes the problem by not using
the file_view size anymore for the calculation in the simple-grouping option,
but the contiguous chunk size (which is identical on all processes).

Fixes issue #7109


gpaulsen:
@edgargabriel Does this need to be cherry-picked back to v4.0.x ?

edgargabriel:
@gpaulsen Yes, will file the pr shortly

PR #6641, opened 2019-05-08
OSHMEM: Add support for shmemx_malloc_with_hint()

hoopoepg:
- added multiple segments processing
- added shmemx_malloc_with_hint call + set of hints

Signed-off-by: Sergey Oblomov <sergeyo@mellanox.com>

PR #7140, opened 2019-11-05
Shmem ucx: fix missing variable declaration in segment_create

devreal:
Adds a variable whose declaration was missing in #6641.

### DIFF
--- a/oshmem/mca/sshmem/ucx/sshmem_ucx_module.c
+++ b/oshmem/mca/sshmem/ucx/sshmem_ucx_module.c
@@ -235,6 +235,7 @@ segment_create(map_segment_t *ds_buf,
         uct_ib_device_mem_h dev_mem = alloc_device_mem(spml, size,
                                                        &dev_mem_address);
         if (dev_mem != NULL) {
+            int ret;
             ret = segment_create_internal(ds_buf, dev_mem_address, size, 0,
                                           hint, dev_mem);
             if (ret == OSHMEM_SUCCESS) {


awlauria:
Note: it actually looks like `ret` was removed in this commit:

https://github.com/open-mpi/ompi/commit/421a7fd47da030101be478d027a9977f0a0d49e0#diff-ae1fde1f4c92b725fdcf638c9bd2b73b

awlauria:
It was this PR for reference:

https://github.com/open-mpi/ompi/pull/6693

Note that I am not quite sure how that passed the build checks...

devreal:
@awlauria 

> Note that I am not quite sure how that passed the build checks...

It requires a version of UCX to be built with devel headers, see https://github.com/openucx/ucx/issues/4388. Maybe such a UCX version can be added to the build checks?

jsquyres:
@devreal I'm not sure I understand your comment.  How are we building against UCX if we don't have the UCX devel headers installed / available?

jsquyres:
@devreal Are you saying that we can't build Open MPI without _internal_ UCX header files?  (i.e., header files that are not normally installed)

That seems... odd...

devreal:
@jsquyres The particular feature that was introduced with  #6641 (`shmemx_malloc_with_hint`) requires the UCX devel headers to be installed. If the (internal) UCX function `uct_ib_md_alloc_device_mem` is not detected during configure the code in question is not compiled (guarded by `HAVE_UCX_DEVICE_MEM`). Unfortunately, this was not documented anywhere...

yosefe:
> @jsquyres The particular feature that was introduced with #6641 (`shmemx_malloc_with_hint`) requires the UCX devel headers to be installed. If the (internal) UCX function `uct_ib_md_alloc_device_mem` is not detected during configure the code in question is not compiled (guarded by `HAVE_UCX_DEVICE_MEM`). Unfortunately, this was not documented anywhere...

Indeed, in order to get a performance benefit with shmemx_malloc_with_hint() running over spml/ucx component, one needs to configure UCX with installing internal header files.
@devreal what do you think is the best place to document this? configure.m4/spml_ucx/other.. ?

devreal:
@yosefe You're right, my description was not exact. The function `shmemx_malloc_with_hint` itself is of course available even without the devel headers but the enhanced functionality behind it is only available with the UCX devel headers.

I am not sure what's the best place to document this dependency. As a start, maybe a comment in the source code is sufficient for developers? That won't help users though. Printing a warning for unsupported hints might be too intrusive. Adding the status of UCX devel headers in the configure summary could be helpful to point out that this dependency exists. Is there precedence in Open MPI, @jsquyres?

rhc54:
Not to be snarky, but it seems a trifle uneven to criticize/complain about the btl/uct component's reliance on UCX internals (which it uses to gain performance) and then turn around and do the exact same thing here for the same reason. To a somewhat disinterested party, it feels like the UCX team has an issue that it should address regarding what it needs to publicly expose for adopters to optimize performance. Perhaps that is where this PR needs to start (i.e., delay adoption here until UCX resolves this recurring problem)?

jsquyres:
@devreal No one reads `configure` output except us.  😄

Meaning: even if you put something there, no one will read it unless they realize they have a problem and (likely) we instructed them to go back and look at their `configure` output.

@rhc54 is right, though -- if this is an undocumented / unpublished interface from UCX, then it's pretty much in exactly the same situation as BTL/UCT.  If this is something that the UCX community feels is an important optimization, then it should be part of the public interface and then this discussion becomes moot (i.e., there should be no need for internal UCX headers to be used by anyone).

devreal:
Apologies if things got out of bounds here. I'm aware that the used internal UCX functions are an optimization that is not required for the interface implemented in #6641 to work but rather offers a potential optimization.

This PR merely fixes an issue that occurs if these UCX internal interfaces are available. I'm not arguing for or against using them, I think that is a broader argument to have elsewhere.

rhc54:
I believe the question we are raising is: should the correct fix for the issue be to remove the use of the internal UCX functions?

I don't know where else in our code we rely on internal interfaces of a software package - certainly not in libevent, hwloc, PMIx, or libfabric. It feels a tad uncomfortable to be introducing such dependencies.

I'm not sure if UCT also falls in that category - that is a separate issue that should be investigated but is outside the scope of this discussion.

awlauria:
Should we create an issue and migrate this discussion over there?

PR #7141, opened 2019-11-05
Shmem: use bitwise and instead of logical and to check for allocator capabilities

devreal:
Found this while browsing the code. I believe the flags should be checked using bitwise and (&) instead of logical and (&&).


awlauria:
@yosefe can you please review?

yosefe:
good catch!
adding @hoopoepg

Issue #7128, opened 2019-10-30
UCT BTL fails to compile using UCX 1.7.0


hppritcha:
The UCT BTL in both master and v4.0.x fail to compile against UCX 1.7.0 rc1.
```
  CC       btl_uct_tl.lo
In file included from btl_uct_device_context.h:16:0,
                 from btl_uct_endpoint.c:17:
btl_uct_rdma.h: In function 'mca_btl_uct_get_rkey':
btl_uct_rdma.h:61:5: error: too few arguments to function 'uct_rkey_unpack'
     ucs_status = uct_rkey_unpack ((void *) remote_handle, rkey);
     ^
In file included from btl_uct_device_context.h:16:0,
                 from btl_uct_rdma.c:12:
btl_uct_rdma.h: In function 'mca_btl_uct_get_rkey':
In file included from btl_uct_device_context.h:16:0,
                 from btl_uct_amo.c:12:
btl_uct_rdma.h: In function 'mca_btl_uct_get_rkey':
In file included from btl_uct_device_context.h:16:0,
                 from btl_uct_component.c:43:
btl_uct_rdma.h: In function 'mca_btl_uct_get_rkey':
btl_uct_rdma.h:61:5: error: too few arguments to function 'uct_rkey_unpack'
     ucs_status = uct_rkey_unpack ((void *) remote_handle, rkey);
     ^
btl_uct_rdma.h:61:5: error: too few arguments to function 'uct_rkey_unpack'
     ucs_status = uct_rkey_unpack ((void *) remote_handle, rkey);
     ^
btl_uct_rdma.h:61:5: error: too few arguments to function 'uct_rkey_unpack'
     ucs_status = uct_rkey_unpack ((void *) remote_handle, rkey);
     ^
In file included from btl_uct.h:41:0,
                 from btl_uct_endpoint.c:15:
/users/hpritchard/ucx_testing//ucx-1.7.0//test_install/include/uct/api/uct.h:2142:14: note: declared here
 ucs_status_t uct_rkey_unpack(uct_component_h component, const void *rkey_buffer,
              ^
In file included from btl_uct_device_context.h:16:0,
                 from btl_uct_endpoint.c:17:
btl_uct_rdma.h: In function 'mca_btl_uct_rkey_release':
btl_uct_rdma.h:72:5: warning: passing argument 1 of 'uct_rkey_release' from incompatible pointer type [enabled by default]
     uct_rkey_release (rkey);
     ^
```
[...]

PR #7142, opened 2019-11-05
btl/uct: restrict to using UCT 1.7 or older API for now

hppritcha:
related to #7128

The UCX crew is no longer guaranteeing that the UCT API is going to be frozen, so this is kind of a whack-a-mole problem trying to keep the BTL UCT working with various changing UCT APIs.

### DIFF
--- a/opal/mca/btl/uct/configure.m4
+++ b/opal/mca/btl/uct/configure.m4
@@ -16,6 +16,8 @@
 #                         All rights reserved.
 # Copyright (c) 2018      Research Organization for Information Science
 #                         and Technology (RIST).  All rights reserved.
+# Copyright (c) 2019      Triad National Security, LLC. All rights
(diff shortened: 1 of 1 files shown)


jsquyres:
You might want to list the UCX versions supported in the “failure” ACMSGRESULT message (because that will be the user’s first question upon getting a failure).

hppritcha:
> You might want to list the UCX versions supported in the “failure” ACMSGRESULT message (because that will be the user’s first question upon getting a failure).

oh good idea.

jsquyres:
Paired with #7144, do you want to bump up the checked version to v1.7, and then cherry pick both this and #7144 to v4.0.x?

hppritcha:
yes that was the plan.

hppritcha:
@jsquyres check now

PR #7149, opened 2019-11-06
Prevent overflow when dealing with datatype count.

### LINKED ISSUE #7147: Segfault in opal_generic_simple_pack when sending struct with a >4GB block

Tested with v4.0.0 and v4.0.2

Tarball from open-mpi.org. No additional parameters provided to `configure`.

* Operating system/version: Ubuntu 18.04 LTS
* Computer hardware: AMD Ryzen Threadripper 2920X, 32GB RAM
* Network type: Vader/TCP (10GE)

-----------------------------

I'm running into a segfault inside `opal_generic_simple_pack` when trying to send a custom struct that contains two members, with one block that is larger than (or equal to) `UINT_MAX` (i.e., 4 GB). I'm not sure whether this is related to the known issue mentioned in the 4.0.2 changelog, but since I'm getting a plain segfault (as opposed to some error message), I believe this is a different corner case.

Here's a minimal example:

```cpp
#include <cassert>
#include <cstdio>
#include <memory>
#include <cstdlib>
#include <vector>

#include <mpi.h>

//constexpr size_t BUFFER_SIZE = 2ull << 30; // Works
constexpr size_t BUFFER_SIZE = 2ull << 31; // Breaks

constexpr int BLOCK_SIZE = 1024;

struct Header {
    int foo;
    int bar;
};

int main(int argc, char* argv[]) {
    assert(BUFFER_SIZE % BLOCK_SIZE == 0); 

    MPI_Init(&argc, &argv);
    int size;
    if(size != 2) return EXIT_FAILURE;
    int rank;

    MPI_Datatype block_type;
    MPI_Type_contiguous(BLOCK_SIZE, MPI_BYTE, &block_type);
    MPI_Type_commit(&block_type);

    std::vector<char> buffer(BUFFER_SIZE);

    Header header;

    MPI_Datatype struct_type;
    // Remove the header block to make it work again
    std::vector<int> block_lengths = { 
        sizeof(Header), // Remove
        BUFFER_SIZE / BLOCK_SIZE};
    std::vector<MPI_Aint> block_disps = { 
        reinterpret_cast<MPI_Aint>(&header), // Remove
        reinterpret_cast<MPI_Aint>(buffer.data())};
    std::vector<MPI_Datatype> block_types = { 
        MPI_BYTE, // Remove
        block_type};
    MPI_Type_create_struct(block_lengths.size(), block_lengths.data(),
        block_disps.data(), block_types.data(), &struct_type);
    MPI_Type_commit(&struct_type);

    if(rank == 1) {
        header.foo = 3;
        header.bar = 42; 
        MPI_Send(MPI_BOTTOM, 1, struct_type, 0, 0, MPI_COMM_WORLD);
    } else {
        MPI_Recv(MPI_BOTTOM, 1, struct_type, 1, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
        printf("Received header foo %d bar %d\n", header.foo, header.bar);
    }   

    MPI_Type_free(&struct_type);
    MPI_Type_free(&block_type);
    MPI_Finalize();
    return EXIT_SUCCESS;
}
```

I tested this both using shared memory (Vader) on a single node, as well as using TCP on a single node and two nodes. All variants cause the segfault.

What's interesting is that this only happens if the header is included in the struct. If you remove the lines indicated above, the `BUFFER_SIZE` can be larger than `UINT_MAX` and everything still works!

bosilca:
This patch fixes #7147 by preventing overflow when multiplying
the count and the blocklen. The count reflects MPI count and is
therefore bound to the size of an int (it is an uint32_t) while the
blocklen can be merged together to represent the largest contiguous
memory layout and it is therefore promoted to a size_t.

### DIFF
--- a/opal/datatype/opal_datatype_internal.h
+++ b/opal/datatype/opal_datatype_internal.h
@@ -156,8 +156,8 @@ typedef struct ddt_elem_id_description ddt_elem_id_description;
  */
 struct ddt_elem_desc {
     ddt_elem_id_description common;           /**< basic data description and flags */
-    uint32_t                blocklen;         /**< number of elements on each block */
-    size_t                  count;            /**< number of blocks */
+    size_t                  blocklen;         /**< number of elements on each block */
+    uint32_t                count;            /**< number of blocks */
     ptrdiff_t               extent;           /**< extent of each block (in bytes) */
     ptrdiff_t               disp;             /**< displacement of the first block */
 };
--- a/opal/datatype/opal_datatype_internal.h
+++ b/opal/datatype/opal_datatype_internal.h
@@ -156,8 +156,8 @@ typedef struct ddt_elem_id_description ddt_elem_id_description;
  */
 struct ddt_elem_desc {
     ddt_elem_id_description common;           /**< basic data description and flags */
-    size_t                  blocklen;         /**< number of elements on each block */
     uint32_t                count;            /**< number of blocks */
+    size_t                  blocklen;         /**< number of elements on each block */
     ptrdiff_t               extent;           /**< extent of each block (in bytes) */
     ptrdiff_t               disp;             /**< displacement of the first block */
 };
--- a/opal/datatype/opal_convertor.c
+++ b/opal/datatype/opal_convertor.c
@@ -403,7 +403,7 @@ opal_convertor_create_stack_at_begining( opal_convertor_t* convertor,
         pStack[1].count = pElems[0].loop.loops;
         pStack[1].type  = OPAL_DATATYPE_LOOP;
     } else {
-        pStack[1].count = pElems[0].elem.count * pElems[0].elem.blocklen;
+        pStack[1].count = (size_t)pElems[0].elem.count * pElems[0].elem.blocklen;
         pStack[1].type  = pElems[0].elem.common.type;
     }
     return OPAL_SUCCESS;
--- a/opal/datatype/opal_convertor_raw.c
+++ b/opal/datatype/opal_convertor_raw.c
@@ -125,7 +125,7 @@ opal_convertor_raw( opal_convertor_t* pConvertor,
     if( pElem->elem.common.flags & OPAL_DATATYPE_FLAG_DATA ) {
         const ddt_elem_desc_t* current = &(pElem->elem);
 
-        if( count_desc != (current->count * current->blocklen) ) {  /* Not the full element description */
+        if( count_desc != ((size_t)current->count * current->blocklen) ) {  /* Not the full element description */
             if( (do_now = count_desc % current->blocklen) ) {
                 do_now = current->blocklen - do_now;  /* how much left in the block */
                 source_base += current->disp;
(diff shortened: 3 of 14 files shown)


opal/datatype/opal_datatype_internal.h
```diff
@@ -156,8 +156,8 @@ typedef struct ddt_elem_id_description ddt_elem_id_description;
  */
 struct ddt_elem_desc {
     ddt_elem_id_description common;           /**< basic data description and flags */
-    uint32_t                blocklen;         /**< number of elements on each block */
-    size_t                  count;            /**< number of blocks */
+    size_t                  blocklen;         /**< number of elements on each block */
+    uint32_t                count;            /**< number of blocks */
```
devreal: Why change `count` from `size_t` to `uint32_t` here? It's being cast to `size_t` in many places now, which seems error prone. It seems safer to handle everything as `size_t` inside the implementation.

bosilca: As the PR log indicates the count mimics what we receive from MPI, right now an int32_t. The only protection we need is to prevent the optimizer from merging elements if their summed count is larger than uint32_t.

Allowing both fields to be size_t would have changed the size of the structure, and this structure is critical to be as small as possible (and cache aligned).

The cast added in many places is to ensure that the compiler generates the correct coda, aka promote the result to size_t.

devreal: In that case: your change changes the size of this struct ;) By putting the 64bit `size_t` after the 2x16-bit `ddt_elem_id_description` member the compiler is required to insert padding (4B each) before `blocklen` and after `count` (before `extent`, which again is 64-bit). Swapping the two members would fix this, of course.

AFAICS, the struct was previously 48B, now it's 56B. Making both members size_t would have the same effect and the struct would still fit into one cache line.

bosilca: Good catch, I forgot to swap the 2 fields. The size of the ddt_elem_desc_t struct is 32B (aka 4 * 8B) on 64 bits architectures.

I am not sure what would be the point of having the count larger than uint32_t, but if you have a valid use case let me know.

devreal: Ahh right, it is 32bit. Not sure I how got to 48...

> I am not sure what would be the point of having the count larger than uint32_t, but if you have a valid use case let me know.

My rational was that having the internal handling in `size_t` across the board (i.e., everything behind the MPI interface) helps avoid future overflows, e.g., because of a forgotten cast somewhere. Plus, I assume this structure will have to be adapted anyway if/when BigCount lands, right?

bosilca: Maybe. It does not makes sense to have both the blocklen and the count as size_t, as together that would represent a gigantic memory span (2^128).

jsquyres:
@bosilca Any reason this isn't merged yet?  Second q: this is applicable for v4.0.x, too, right?

bosilca:
I will merged right now. And yes, it needs to go in the 3.x and 4.x.

jsquyres:
I cherry picked to v4.0.x in #7282.

I tried to cherry pick to v3.0.x and v3.1.x, but immediately got a bunch of conflicts that I didn't know how to deal with.  I think it would be ok to say that these are fixed in v4.0.x; if a user needs these fixes, they need to upgrade to v4.0.x.

PR #7151, opened 2019-11-07
COLL/TUNED: Add linear scatter using isend for mlnx platform - v4.0.x

brminich:
Add scatter algorithm using non-blocking sends with "pseudo" sync by blocking sends.
By default enable for Mellanox platform only. Default selection should not affect other platforms.
Provides significant performance benefits (up to 2x for certain message and comm sizes), tested up to 2048 ranks.

(cherry picked from commit f2cbd4806e9a38b5e58c0fc69b41624af79fb99b)

### DIFF
--- a/contrib/platform/mellanox/optimized.conf
+++ b/contrib/platform/mellanox/optimized.conf
@@ -10,6 +10,7 @@
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
 # Copyright (c) 2006      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2019      Mellanox Technologies. All rights reserved.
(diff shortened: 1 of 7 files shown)


hppritcha:
@brminich this appears to be an enhancement.  Does it actually fix an issue?  We generally do not take performance enhancements into a release branch.

yosefe:
> @brminich this appears to be an enhancement. Does it actually fix an issue? We generally do not take performance enhancements into a release branch.

@hppritcha this is indeed a performance improvement, but since it's a general improvement which could be enabled (but disabled by default), and the v5.0.0 release seems pretty far away, is it possible to consider it for v4.0.x branch?

gpaulsen:
We will close this PR soon.  No further action is needed as this PR is already in master.

PR #7152, opened 2019-11-07
Topic/btl uct fixes for v4.0.x

hppritcha:
three BTL UCT commits cherry-picked  from master:

9d345d9aa000233bec148540b071cecffc94438c
a3026c016a6a8be379f62585b6ddc070175c8106
526775dfd7ad75c308532784de4fb3ffed25458f

we fix the whitelisted UCX/UCT to 1.7 since "1.8" is really UCX master for which the UCT API could still be in flux.

### DIFF
--- a/opal/mca/btl/uct/btl_uct.h
+++ b/opal/mca/btl/uct/btl_uct.h
@@ -12,6 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2015-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2019      Google, LLC. All rights reserved.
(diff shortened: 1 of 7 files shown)


gpaulsen:
While the commits in the PR claim to fix UCT/BTL for both v1.7 and v1.8, note that the configury in this PR will only attempt to biuld UCT/BTL for versions up to v1.7 (not v1.8).  The rational for that being UCX v1.8 still hasn't released, and is just master in the UCX project, and may still need additional changes.

artpol84:
@yosefe what is the schedule for the UCX 1.7 release?
@hppritcha would like to align with respect to 4.0.3 release so this fix is out and packagers won't hit the issue.

PR #7153, opened 2019-11-10
fix misleading error message with missing #! interpreter

### LINKED ISSUE #4528: Misleading error message from mpiexec with missing #! interpreter
With openmpi 2.0.2 compiled from source I try to execute this script (test.sh):
```bash
#!/bin/nonexistent
echo works
```
and the error message from `mpiexec -n 1` is very confusing:
```
...
  Application name:  ./test.sh
  Error:             No such file or directory
```

It took me a long while to figure out what was wrong with my mpi4py script, which had a wrong #! line due to virtualenv weirdness...

mcoil1:
This change fixes the misleading error message from mpiexec with missing #! interpreter (#4528). I added a conditional to check whether the file exists, which determines whether to output the default error string (missing file) or a new string (bad interpreter).

Fixes #4528

### DIFF
--- a/orte/mca/odls/default/odls_default_module.c
+++ b/orte/mca/odls/default/odls_default_module.c
@@ -454,11 +454,22 @@ static int do_child(orte_odls_spawn_caddy_t *cd, int write_fd)
 
     /* Exec the new executable */
     execve(cd->cmd, cd->argv, cd->env);
-    getcwd(dir, sizeof(dir));
+    /* If we get here, an error has occurred. */
+    (void) getcwd(dir, sizeof(dir));
+    struct stat stats;
+    char* msg;
+    /* If errno is ENOENT, that indicates either cd->cmd does not exist, or
+     * cd->cmd is a script, but has a bad interpreter specified. */
+    if (ENOENT == errno && 0 == stat(cd->app->app, &stats)) {
+        asprintf(&msg, "%s has a bad interpreter on the first line.",
+                 cd->app->app);
+    } else {
+        msg = strdup(strerror(errno));
+    }
     send_error_show_help(write_fd, 1,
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

orte/mca/odls/default/odls_default_module.c
```diff
@@ -455,9 +455,16 @@ static int do_child(orte_odls_spawn_caddy_t *cd, int write_fd)
     /* Exec the new executable */
     execve(cd->cmd, cd->argv, cd->env);
     getcwd(dir, sizeof(dir));
+    struct stat stats;
+    char* msg;
+    if(0 == stat(cd->app->app, &stats)) {
```
ggouaillardet: What if the app is a binary with unresolved dependencies?
I am afraid the error message will be "Bad interpreter" in that case.

bwbarrett: I agree; this would batch too many failure modes (wrong architecture, missing libraries, wrong os, etc.) into one error code, and worse, it would be coming from us, not the OS>

ggouaillardet: one option could be to `stat()` the file, and if successful and it starts with `#!`, then add a hint into the error message that the MPI app is indeed a script and the error could come from the interpreter.

jsquyres: On Linux, it looks like `errno == ENOENT` means either the file was missing or it's a bad interpreter.  Looks like it's the same behavior on Linux.

So adding a check for `errno == ENOENT` would likely be sufficient for this small fix.

mcoil1:
I added the `ENOENT` check.

Issue #7157, opened 2019-11-12
wrong environment in Docker


Sunmon:
Thank you for taking the time to submit an issue!

## Background information
Found some problem while 'mpirun on docker'

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
openMPI 4.0.2

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Dockerfile
COPY ./openmpi-docker /openmpi-docker (# same as (https://github.com/open-mpi/ompi/archive/master.zip))
WORKDIR /openmpi-docker
RUN ./configure --prefix="/usr/local/openmpi"
RUN make install
RUN apt-get install -y openssh-client

### Please describe the system on which you are running

* Operating system/version: Ubuntu 18.04.3 LTS, Docker 19.03.2
* Computer hardware: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
* Network type: Not use

-----------------------------

## Details of the problem

There are some problems on mpi when running on the docker.
I think mpi does not notice container environment.
Here are not-working commands I found:
- np
- --cpu-set
- --bind-to core

### How to test

1. build image by using dockerfile
2. `docker run -it --cpuset-cpus="1,2" [openmpiImg]  //only use cpu1, cpu2
3. command
   3.1. mpirun --allow-run-as-root -np 4 [example.out]
   3.2. mpirun --allow-run-as-root --cpu-set 0,1 [example.out]
   3.3. mpirun  --allow-run-as-root --bind-to coe [example.out]

- -np option: 
np %d is bigger than cpuset. So it has to print 'slot error' but I does not.

- --cpu-set:
docker container did not allowed to use cpu0. So, when I check cpu usage by `htop`, I can found that only cpu1 is running.
But running with this command `mpirun --allow-run-as-root --cpu-set 0 [example.out], it causes error and eventually use both cpus, number cpu 1 and cpu 2.

- --bind-to core:
when I use --bind-to core option, it prints error

---

All above problems does not appear when using cgroup.. Only appears on docker.
I fixed the `-np` bug, but it is a temporary measure. (I will pull request)

rhc54:
@bgoglin 

I believe the problem here is that root may not be constrained to the designated cpus. Thus, hwloc sees all cpus as available. However, I defer to Brice.

bgoglin:
(should have posted this here instead of in the PR)
[...]

PR #7160, opened 2019-11-12
prevent not-allowed cpu recognize

Sunmon:
issue #7157
Signed-off-by : hansolpp <hansolmin3@gmail.com>
Signed-off-by: sunmon <blackmilk274@gmail.com>

This is the problem:

```bash
docker run --cpuset-cpus="1,3" [openmpi image]

# on mpi
mpirun -np 4 [example.out] 
```

mpirun must not be runned, because container has 2 cpus but np gives 4 processes. So it has to be print 'slot error', but it doesn't.

So I fixed this problem by prevent mpi recognizing unallocated cpu.
But it is only temporary solution.

### DIFF
--- a/opal/mca/hwloc/hwloc201/hwloc/hwloc/topology-linux.c
+++ b/opal/mca/hwloc/hwloc201/hwloc/hwloc/topology-linux.c
@@ -3178,6 +3178,8 @@ list_sysfsnode(struct hwloc_linux_backend_data_s *data,
   return indexes;
 }
 
+
+
 static int
 look_sysfsnode(struct hwloc_topology *topology,
 	       struct hwloc_linux_backend_data_s *data,
@@ -3240,6 +3242,7 @@ look_sysfsnode(struct hwloc_topology *topology,
 	    } else {
 	      hwloc_bitmap_or(nodes_cpuset, nodes_cpuset, cpuset);
 	    }
+
 
 	    node = hwloc_alloc_setup_object(topology, HWLOC_OBJ_NUMANODE, osnode);
 	    node->cpuset = cpuset;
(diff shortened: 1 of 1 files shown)


awlauria:
ok to test

opal/mca/hwloc/hwloc201/hwloc/hwloc/topology-linux.c
```diff
@@ -3419,6 +3422,86 @@ look_sysfsnode(struct hwloc_topology *topology,
...
+
+        last_explore = current_explore;
+        set_allowed(allowed, num1, num2);
+    }
+}
+
+void read_cpuset_cpus(char* buf, char* path)
+{
+    FILE *fd = fopen(path, "r");
+    fgets(buf, 20, fd);
```
awlauria: if you get all 20, you won't have room for a `'0' in `buf`

rhc54:
Time out, folks. We are not going to modify the embedded hwloc in OMPI. We should discuss this with @bgoglin to see if (a) there already is a solution in hwloc (I believe there is, but we may not be using it) and (b) what the correct solution should be.

opal/mca/hwloc/hwloc201/hwloc/hwloc/topology-linux.c
```diff
@@ -3419,6 +3422,86 @@ look_sysfsnode(struct hwloc_topology *topology,
...
+{
+    char *last_explore, *current_explore;
+
+    current_explore = buf;
+    last_explore = current_explore; //last_explore는 마지막으로 탐색한 숫자 첫 자리에 위치
+
+    while( *(++current_explore) != '\n' && *current_explore != '\0')
+    {
+        current_explore = find_next_deli(current_explore);
+        if( current_explore == NULL) current_explore = buf + strlen(buf) - 1;
```
awlauria: I'm not sure what the purpose of this line is.

if current_explore is NULL, why not break out of the loop?

opal/mca/hwloc/hwloc201/hwloc/hwloc/topology-linux.c
```diff
@@ -3443,6 +3526,10 @@ look_sysfscpu(struct hwloc_topology *topology,
     struct dirent *dirent;
     cpuset = hwloc_bitmap_alloc();
 
+    int allow[20];
```
awlauria: does this only allow for 20 cpus per node?

rhc54:
I've flagged Brice on the issue that was filed. Let's table this PR for the moment until we see how he wants to proceed.

ibm-ompi:

 Gist: https://gist.github.com/3ff99f228b7e99a72c6f6282038c3368

ibm-ompi:

 Gist: https://gist.github.com/5d8bd98770e950a214fd47f30a69085b

bgoglin:
@Sunmon hwloc already supports cgroups, so things should work fine. One thing that isn't supported in current releases is Linux cgroups v2 (I added it to git a couple weeks ago), but your problem seems to involve cgroups v1 only.
Can you run hwloc-gather-topology inside that docker and send the generated tarball to me (brice.goglin@inria.fr) so that I try to debug why it fails here?
Otherwise the output of lstopo when hwloc is configured with --enable-debug would also be helpful.

ibm-ompi:

 Gist: https://gist.github.com/e17d844b6e88682144e7e43a4969b7ff

PR #7167, opened 2019-11-14
Reachable documentation change

wckzhang:
We are changing the documentation and promised behavior of the reachability framework and changing the netlinks component to match.

See more details in #7134

### DIFF
--- a/opal/mca/reachable/netlink/reachable_netlink_module.c
+++ b/opal/mca/reachable/netlink/reachable_netlink_module.c
@@ -2,8 +2,8 @@
 /*
  * Copyright (c) 2014-2015 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015 Cisco Systems.  All rights reserved.
- * Copyright (c) 2017 Amazon.com, Inc. or its affiliates.
- *                    All Rights reserved.
+ * Copyright (c) 2017-2019 Amazon.com, Inc. or its affiliates.
+ *                         All Rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


opal/mca/reachable/reachable.h
```diff
@@ -75,6 +75,8 @@ typedef int (*opal_reachable_base_module_fini_fn_t)(void);
  * interfaces, but only a (comparable) reachability between any pair
  * of local/remote interfaces.
  *
+ * This function will show an identical local and remote interface as
+ * reachable through loopback.
```
jsquyres: Nit pick: it will show it as reachable -- I don't think the "reachable" framework indicates / implies which interface the traffic will actually go through, right?

wckzhang: That's true, I'll change it to just say "as reachable".

wckzhang:
@markalle Can you confirm that this patch combined with my previous one fixes your issue? Thanks.

opal/mca/reachable/netlink/reachable_netlink_module.c
```diff
@@ -112,6 +112,15 @@ static int get_weights(opal_if_t *local_if, opal_if_t *remote_if)
         remote_ip = (uint32_t)((struct sockaddr_in *)&(remote_if->if_addr))->sin_addr.s_addr;
         outgoing_interface = local_if->if_kernel_index;
 
+        /* If the ips are identical, assume reachable through loopback */
```
bwbarrett: Can you add a bit more of a comment here?  Say that we're doing something artificial because historical reasons.  Someone in 2 years is going to wonder why this comment is here.

wckzhang: I updated the comment.

PR #7168, opened 2019-11-14
v4.0.x: schizo/ompi: correctly handle the yield_when_idle option

wbailey2:
in schizo/ompi, sets the new OMPI_MCA_mpi_oversubscribe environment
variable according to the node oversubscription state.

This MCA parameter is used to set the default value of the
mpi_yield_when_idle parameter.

This two steps tango is needed so the mpi_yield_when_idle setting
is always honored when set in a config file.

Refs. open-mpi/ompi#6433

### DIFF
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -17,7 +17,7 @@
  * Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016      Research Organization for Information Science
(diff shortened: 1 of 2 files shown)


jsquyres:
ok to test

gpaulsen:
@wbailey2,  Thanks for the PR.   Normally when cherry-picking to a release branch, we require the developer to specify `-x` to the `git cherry-pick` command.  This provides a reference to the master commit, so that we can more easily audit code flow.
This is fine, this time to leave it since you left the Reference in the commit message, but I wanted to point it out for future Open MPI dev work you may do. :)

Thanks.

wbailey2:
I updated the commit to include where the cherry-pick was from.

jsquyres:
bot:ompi:retest

PR #7170, opened 2019-11-15
Add detection for JSM direct launch

jjhursey:
* Adds the `schizo/jsm` component that detects if the process was
   direct launched with IBM's Job Step Manager (JSM). JSM is a PMIx
   enhanced runtime environment so flag it as such.


jjhursey:
I did some testing with JSM + Open MPI in a JSM environment and the detection worked correctly (and so did the app :) )

hppritcha:
@jjhursey is this PR still relevant once ORTE is removed?

jjhursey:
Once ORTE is removed we shouldn't need this component anymore (It doesn't exist in PRRTE). I'll confirm with the OMPI without ORTE branch to make sure it still works.

So for a bit of time on master this functionality is available for those looking for it. But once ORTE is removed then it'll just be in history.

PR #7174, opened 2019-11-18
mtl/ofi: increase the FI_VERSION requested to 1.5 and make sure to check for OFI_LOCAL_COMM

jsquyres:
usNIC requires a minimum of v1.3.

### DIFF
--- a/config/opal_check_ofi.m4
+++ b/config/opal_check_ofi.m4
@@ -1,6 +1,6 @@
 dnl -*- shell-script -*-
 dnl
-dnl Copyright (c) 2015-2019 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2015-2020 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2016-2017 Los Alamos National Security, LLC. All rights
 dnl                         reserved.
 dnl $COPYRIGHT$
@@ -9,6 +9,45 @@ dnl Additional copyrights may follow
 dnl
 dnl $HEADER$
 dnl
+
+dnl
+dnl OPAL_CHECK_OFI_VERSION_GE
+dnl
+dnl Check that the OFI API version number is >= a specific value.
+dnl
+dnl $1: version number to compare, in the form of "major,minor"
+dnl     (without quotes) -- i.e., a single token representing the
+dnl     arguments to FI_VERSION()
+dnl $2: action if OFI API version is >= $1
+dnl $3: action if OFI API version is < $1
+AC_DEFUN([OPAL_CHECK_OFI_VERSION_GE],[
+    OPAL_VAR_SCOPE_PUSH([opal_ofi_ver_ge_save_CPPFLAGS opal_ofi_ver_ge_happy])
+
+    AC_MSG_CHECKING([if OFI API version number is >= $1])
+    opal_ofi_ver_ge_save_CPPFLAGS=$CPPFLAGS
+    CPPFLAGS=$opal_ofi_CPPFLAGS
+
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <rdma/fabric.h>]],
+[[
+#if !defined(FI_MAJOR_VERSION)
+#error "we cannot check the version -- sad panda"
+#elif FI_VERSION_LT(FI_VERSION(FI_MAJOR_VERSION, FI_MINOR_VERSION), FI_VERSION($1))
+#error "version is too low -- nopes"
+#endif
+]])],
+                      [opal_ofi_ver_ge_happy=1],
+                      [opal_ofi_ver_ge_happy=0])
+
+    AS_IF([test $opal_ofi_ver_ge_happy -eq 1],
+          [AC_MSG_RESULT([yes])
+           $2],
+          [AC_MSG_RESULT([no])
+           $3])
+
+    CPPFLAGS=$opal_ofi_ver_ge_save_CPPFLAGS
+
+    OPAL_VAR_SCOPE_POP
+])dnl
 
 dnl
 dnl _OPAL_CHECK_OFI
--- a/ompi/mca/mtl/ofi/configure.m4
+++ b/ompi/mca/mtl/ofi/configure.m4
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2013-2014 Intel, Inc. All rights reserved
 #
-# Copyright (c) 2014-2019 Cisco Systems, Inc.  All rights reserved
+# Copyright (c) 2014-2020 Cisco Systems, Inc.  All rights reserved
 # Copyright (c) 2017      Los Alamos National Security, LLC.  All rights
 #                         reserved.
 # $COPYRIGHT$
@@ -27,6 +27,12 @@ AC_DEFUN([MCA_ompi_mtl_ofi_CONFIG],[
 
     # Check for OFI
     OPAL_CHECK_OFI
+
+    # The OFI MTL requires at least OFI libfabric v1.5.
+    AS_IF([test "$opal_ofi_happy" = "yes"],
+          [OPAL_CHECK_OFI_VERSION_GE([1,5],
+                                     [],
+                                     [opal_ofi_happy=no])])
 
     AS_IF([test "$opal_ofi_happy" = "yes"],
           [$1],
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -647,9 +647,11 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                             __FILE__, __LINE__);
         goto error;
     }
+    /* Make sure to get a RDM provider that can do the tagged matching
+       interface and local communication and remote communication. */
     hints->mode               = FI_CONTEXT;
-    hints->ep_attr->type      = FI_EP_RDM;      /* Reliable datagram         */
-    hints->caps               = FI_TAGGED;      /* Tag matching interface    */
+    hints->ep_attr->type      = FI_EP_RDM;
+    hints->caps               = FI_TAGGED | FI_LOCAL_COMM | FI_REMOTE_COMM;
     hints->tx_attr->msg_order = FI_ORDER_SAS;
     hints->rx_attr->msg_order = FI_ORDER_SAS;
     hints->rx_attr->op_flags = FI_COMPLETION;
@@ -697,8 +699,13 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
      * FI_VERSION provides binary backward and forward compatibility support
      * Specify the version of OFI is coded to, the provider will select struct
      * layouts that are compatible with this version.
+     *
+     * Note: API version 1.5 is the first version that supports
+     * FI_LOCAL_COMM / FI_REMOTE_COMM checking (and we definitely need
+     * that checking -- e.g., some providers are suitable for RXD or
+     * RXM, but can't provide local communication).
      */
-    fi_version = FI_VERSION(1, 0);
+    fi_version = FI_VERSION(1, 5);
 
     /**
      * fi_getinfo:  returns information about fabric  services for reaching a
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -341,21 +341,12 @@ is_in_list(char **list, char *item)
 }
 
 static struct fi_info*
-select_ofi_provider(struct fi_info *providers)
+select_ofi_provider(struct fi_info *providers,
+                    char **include_list, char **exclude_list)
 {
-    char **include_list = NULL;
-    char **exclude_list = NULL;
     struct fi_info *prov = providers;
 
-    opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
-                        "%s:%d: mtl:ofi:provider_include = \"%s\"\n",
-                        __FILE__, __LINE__, prov_include);
-    opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
-                        "%s:%d: mtl:ofi:provider_exclude = \"%s\"\n",
-                        __FILE__, __LINE__, prov_exclude);
-
-    if (NULL != prov_include) {
-        include_list = opal_argv_split(prov_include, ',');
+    if (NULL != include_list) {
         while ((NULL != prov) &&
                (!is_in_list(include_list, prov->fabric_attr->prov_name))) {
             opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
@@ -364,8 +355,7 @@ select_ofi_provider(struct fi_info *providers)
                                 prov->fabric_attr->prov_name);
             prov = prov->next;
         }
-    } else if (NULL != prov_exclude) {
-        exclude_list = opal_argv_split(prov_exclude, ',');
+    } else if (NULL != exclude_list) {
         while ((NULL != prov) &&
                (is_in_list(exclude_list, prov->fabric_attr->prov_name))) {
             opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
@@ -375,9 +365,6 @@ select_ofi_provider(struct fi_info *providers)
             prov = prov->next;
         }
     }
-
-    opal_argv_free(include_list);
-    opal_argv_free(exclude_list);
 
     opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
                         "%s:%d: mtl:ofi:prov: %s\n",
@@ -621,7 +608,9 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
     int ret, fi_version;
     int num_local_ranks, sep_support_in_provider, max_ofi_ctxts;
     int ofi_tag_leading_zeros, ofi_tag_bits_for_cid;
-    struct fi_info *hints;
+    char **include_list = NULL;
+    char **exclude_list = NULL;
+    struct fi_info *hints, *hints_dup = NULL;
     struct fi_info *providers = NULL;
     struct fi_info *prov = NULL;
     struct fi_info *prov_cq_data = NULL;
@@ -629,6 +618,19 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
     size_t namelen;
     int universe_size;
     char *univ_size_str;
+
+    opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                        "%s:%d: mtl:ofi:provider_include = \"%s\"\n",
+                        __FILE__, __LINE__, prov_include);
+    opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                        "%s:%d: mtl:ofi:provider_exclude = \"%s\"\n",
+                        __FILE__, __LINE__, prov_exclude);
+
+    if (NULL != prov_include) {
+        include_list = opal_argv_split(prov_include, ',');
+    } else if (NULL != prov_exclude) {
+        exclude_list = opal_argv_split(prov_exclude, ',');
+    }
 
     /**
      * Hints to filter providers
@@ -708,6 +710,52 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
     fi_version = FI_VERSION(1, 5);
 
     /**
+     * The EFA provider in Libfabric versions prior to 1.10 contains a bug
+     * where the FI_LOCAL_COMM and FI_REMOTE_COMM capabilities are not
+     * advertised.  However, we know that this provider supports both local and
+     * remote communication. We must exclude these capability bits in order to
+     * select EFA when we are using a version of Libfabric with this bug.
+     *
+     * Call fi_getinfo() without those capabilities and specifically ask for
+     * the EFA provider. This is safe to do as EFA is only supported on Amazon
+     * EC2 and EC2 only supports EFA and TCP-based networks. We'll also skip
+     * this logic if the user specifies an include list without EFA or adds EFA
+     * to the exclude list.
+     */
+    if ((include_list && is_in_list(include_list, "efa")) ||
+        (exclude_list && !is_in_list(exclude_list, "efa"))) {
+        hints_dup = fi_dupinfo(hints);
+        hints_dup->caps &= ~(FI_LOCAL_COMM | FI_REMOTE_COMM);
+        hints_dup->fabric_attr->prov_name = strdup("efa");
+
+        ret = fi_getinfo(fi_version, NULL, NULL, 0ULL, hints_dup, &providers);
+
+        opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                            "%s:%d: EFA specific fi_getinfo(): %s\n",
+                            __FILE__, __LINE__, fi_strerror(-ret));
+
+        if (FI_ENODATA == -ret) {
+            /**
+             * EFA is not available so fall through to call fi_getinfo() again
+             * with the local/remote capabilities set.
+             */
+            fi_freeinfo(hints_dup);
+            hints_dup = NULL;
+        } else if (0 != ret) {
+            opal_show_help("help-mtl-ofi.txt", "OFI call fail", true,
+                           "fi_getinfo",
+                           ompi_process_info.nodename, __FILE__, __LINE__,
+                           fi_strerror(-ret), -ret);
+            goto error;
+        } else {
+            fi_freeinfo(hints);
+            hints = hints_dup;
+            hints_dup = NULL;
+            goto select_prov;
+        }
+    }
+
+    /**
      * fi_getinfo:  returns information about fabric  services for reaching a
      * remote node or service.  this does not necessarily allocate resources.
      * Pass NULL for name/service because we want a list of providers supported.
@@ -718,6 +766,11 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                      0ULL,          /* Optional flag                            */
                      hints,         /* In: Hints to filter providers            */
                      &providers);   /* Out: List of matching providers          */
+
+    opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+                        "%s:%d: fi_getinfo(): %s\n",
+                        __FILE__, __LINE__, fi_strerror(-ret));
+
     if (FI_ENODATA == -ret) {
         // It is not an error if no information is returned.
         goto error;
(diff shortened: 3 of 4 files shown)


jsquyres:
I added all relevant MTL/OFI people I could think of to this PR.  Please add anyone I forgot.

This is not an immediate need, but I wanted to start a discussion:

1. usNIC *requires* Libfabric API v1.3 to function properly.  If we pass in `FI_VERSION(1,0)`, the usNIC will not function properly (because of clarifications that occurred starting with API v1.3 -- we can't change the usnic provider <v1.3 behavior because then usNIC will stop working with older versions of Open MPI.  Rock: meet hard place).
1. That being said, the usNIC Libfabric provider is not yet ready to be part of the RXD world; we still need to add shared memory handling in the provider.  This PR is an opening to a discussion; it is by no means an urgent need.
1. Is there a reason to keep the `FI_VERSION` at 1.0?  If so, could we perhaps do an iterative approach (e.g., try `FI_VERSION(1,3)`, and if nothing is found, try again with `FI_VERSION(1,0)`)?
1. Starting somewhere around Libfabric v1.8, RXD now reports that usNIC can be used.  This means that RXD+usNIC tries to run in the MTL OFI, and things go rapidly downhill (because in Libfabric <=v1.9, the usNIC provider is not ready for this).  What is the best way to prevent this behavior?  I'm thinking I can add RXD+usNIC into the MTL OFI provider exclude list, but someday I would still like to have the (non-default) option of using RXD+usNIC in the MTL OFI.  Thoughts?

bwbarrett:
From an EFA standpoint, we're ok with this change, given the lateness in Libfabric's development where EFA was added.

I don't have any great answers for usnic/rxd, but I'm also not sure of the state of usnic in the libfabric source tree.  If RxD is blowing up, then it feels like there's a bug in some request / provide capability; can you key off of that in usnic/rxd?

jsquyres:
@shefty pointed me in the right direction (as did @bwbarrett): we need to pass down `FI_LOCAL_COMM | FI_REMOTE_COMM` in the caps query to ensure that whatever providers are returned are capable of doing local communication.  This seems to correctly exclude the Libfabric usNIC provider.  Yay!

It has a secondary effect: `FI_LOCAL_COMM` and `FI_REMOTE_COMM` were introduced in Libfabric API v1.5, so we actually need to bump the minimum Libfabric API version up to v1.5 (not v1.3).  So I amended this PR to also have `mtl/ofi/configure.m4` check the Libfabric API version in `configure.m4` and skip itself if the API version is <v1.5.

This is a big change to the original PR, so I'm canceling the two approved code reviews and requesting everyone look again.  Thanks.

gpaulsen:
Is any part of this intended for v4.0.3 ?

jsquyres:
> Is any part of this intended for v4.0.3 ?

Yes.  Timing is bad to have everyone discuss this, though.

hppritcha:
the GNI provider stops working with this PR.  I need to to investigate some, I think it has to do with the default memory model with 1.5

hppritcha:
well I did some more checking and as long as I don't build the rxm and ofi-rxd providers things appear to work with respect to GNI provider.

shefty:
Does rxm or rxd pick up gni as a core provider?  I thought gni only exported rdm ep types.  Any idea why including rxm/rxd causes issues?

jsquyres:
> This causes something to break in the EFA provider. We're still investigating exactly what breaks and why, so no actual request for change, just a request for not merging for now.

Looks like this is related to https://github.com/ofiwg/libfabric/pull/5461

bwbarrett:
> > This causes something to break in the EFA provider. We're still investigating exactly what breaks and why, so no actual request for change, just a request for not merging for now.
> 
> Looks like this is related to [ofiwg/libfabric#5461](https://github.com/ofiwg/libfabric/pull/5461)

Yes, it is.  The team's still trying to figure out a path forward; if we commit this PR, then all versions of Libfabric out in the wild stop working for EFA.  Need to think more here.  Blah.

jsquyres:
Understood. 

Maybe call fi_info() multiple times?

rwespetal:
Sure, we could call fi_getinfo() without FI_LOCAL_COMM and FI_REMOTE_COMM and look for the EFA provider with the assumption that those capabilities are supported. If EFA is not found we can call fi_getinfo() again with those capabilities to avoid usNIC. That would workaround the libfabric EFA bug we discovered but allow RxD+usNIC to be used in the future once it supports FI_LOCAL_COMM. I can write a patch for this if this is an acceptable solution.

I'm somewhat concerned about that adding a workaround like this may cause the selection logic to potentially do the wrong thing in the future. As it stands today the EFA provider is only on EC2 and on EC2 this is the only possible provider due to the default exclude list and because FI_LOCAL_COMM and FI_REMOTE_COMM are required.

Also, I'm not sure it matters but requiring FI_REMOTE_COMM will prevent users from using the shm provider for a job that is local to the node. Vader is going to be used by default in that case though.

jsquyres:
It might be slightly better not to hard-code anything for EFA/usNIC, but rather call `fi_getinfo()` more than once and just see if _anything_ is returned.  E.g., something like this (pseudocode):

```c
fi_getinfo(hints); // call without FI_LOCAL + FI_REMOTE
if (we_got_any_valid_providers) {
    use_the_first_provider();
} else {
    hints += FI_LOCAL + FI_REMOTE;
    fi_getinfo(hints);
    if (we_got_any_valid_providers) {
        use_the_first_provider();
    } else {
        return SADNESS_BECAUSE_THE_MTL_HAS_NO_PROVIDERS;
    }
}
```

I'm sure the above could be made a bit more elegant (and probably with some lengthy comments detailing to our Future Selves why we're doing this).

Fair point about the shm provider, but I think you're right: vader is likely better optimized than the Libfabric shm provider (at least for now...?).  That being said, someday if the shm Libfabric provider becomes Awesome(tm), then we could adjust the `fi_getinfo()` scheme to check and see if it's a single-node job, and if so, only call with FI_LOCAL.  That being said, multiple providers will likely be returned with just FI_LOCAL -- how will we know to get just the shm provider?  That's probably a question outside the scope of this PR...

shefty:
If you query for FI_LOCAL, then discard responses that also have FI_REMOTE set, you're at least dealing with local only providers.  Note that I don't expect provider names to change.  We have too much invested in using those names just for validation.

rwespetal:
> It might be slightly better not to hard-code anything for EFA/usNIC, but rather call `fi_getinfo()` more than once and just see if _anything_ is returned. E.g., something like this (pseudocode):
> 
> ```c
> fi_getinfo(hints); // call without FI_LOCAL + FI_REMOTE
> if (we_got_any_valid_providers) {
>     use_the_first_provider();
> } else {
>     hints += FI_LOCAL + FI_REMOTE;
>     fi_getinfo(hints);
>     if (we_got_any_valid_providers) {
>         use_the_first_provider();
>     } else {
>         return SADNESS_BECAUSE_THE_MTL_HAS_NO_PROVIDERS;
>     }
> }
> ```
> 
> I'm sure the above could be made a bit more elegant (and probably with some lengthy comments detailing to our Future Selves why we're doing this).
> 
> Fair point about the shm provider, but I think you're right: vader is likely better optimized than the Libfabric shm provider (at least for now...?). That being said, someday if the shm Libfabric provider becomes Awesome(tm), then we could adjust the `fi_getinfo()` scheme to check and see if it's a single-node job, and if so, only call with FI_LOCAL. That being said, multiple providers will likely be returned with just FI_LOCAL -- how will we know to get just the shm provider? That's probably a question outside the scope of this PR...

Right, it'd be ideal if we can avoid anything provider specific here. If the first provider is chosen without the local and remote flags wouldn't that pick up RxD and usNIC?

jsquyres:
> Right, it'd be ideal if we can avoid anything provider specific here. If the first provider is chosen without the local and remote flags wouldn't that pick up RxD and usNIC?

Yech; yeah.  I think I got it backwards.  And thinking about this a bit more, I think it _will_ be necessary to special case the older versions.  How's this?

```c
// Obviously, this is pseudocode
hints += FI_VERSION(1, 5)
hints += FI_LOCAL + FI_REMOTE;
fi_getinfo(hints);
if (we_got_any_valid_providers) {
    // This is the new/correct way: if we get one, we're good.
    // No special cases here.
    use_the_first_provider();
} else {
    // This is the old/incorrect way.  We need some special cases to handle the
    // legacy cases (which will never change, which is why it's ok -- albeit
    // icky -- to add some special cases here)
    hints -= (FI_LOCAL + FI_REMOTE);
    fi_getinfo(hints);
    if (we_got_any_valid_providers) {
        filter_out_rxd_usnic_providers();
        use_the_first_provider();
    } else {
        return SADNESS_BECAUSE_THE_MTL_HAS_NO_PROVIDERS;
    }
}
```

rwespetal:
> > Right, it'd be ideal if we can avoid anything provider specific here. If the first provider is chosen without the local and remote flags wouldn't that pick up RxD and usNIC?
> 
> Yech; yeah. I think I got it backwards. And thinking about this a bit more, I think it _will_ be necessary to special case the older versions. How's this?
> 
> ```c
> // Obviously, this is pseudocode
> hints += FI_VERSION(1, 5)
> hints += FI_LOCAL + FI_REMOTE;
> fi_getinfo(hints);
> if (we_got_any_valid_providers) {
>     // This is the new/correct way: if we get one, we're good.
>     // No special cases here.
>     use_the_first_provider();
> } else {
>     // This is the old/incorrect way.  We need some special cases to handle the
>     // legacy cases (which will never change, which is why it's ok -- albeit
>     // icky -- to add some special cases here)
>     hints -= (FI_LOCAL + FI_REMOTE);
>     fi_getinfo(hints);
>     if (we_got_any_valid_providers) {
>         filter_out_rxd_usnic_providers();
>         use_the_first_provider();
>     } else {
>         return SADNESS_BECAUSE_THE_MTL_HAS_NO_PROVIDERS;
>     }
> }
> ```

I think that may prefer RxD + EFA over DGRAM instead of the EFA RDM provider which isn't what we'd want. I can follow up later after I get a chance to look at this more closely. Otherwise, that looks good to me.

jsquyres:
> I think that may prefer RxD + EFA over DGRAM instead of the EFA RDM provider which isn't what we'd want. I can follow up later after I get a chance to look at this more closely. Otherwise, that looks good to me.

Not sure what you mean.  Do you mean you would need to customize `use_the_first_provider()` a bit (e.g., that function name may not even be right -- perhaps `find_a_suitable_provider()`?) to prefer non-RXD providers before RXD-based providers...?

rwespetal:
> > I think that may prefer RxD + EFA over DGRAM instead of the EFA RDM provider which isn't what we'd want. I can follow up later after I get a chance to look at this more closely. Otherwise, that looks good to me.
> 
> Not sure what you mean. Do you mean you would need to customize `use_the_first_provider()` a bit (e.g., that function name may not even be right -- perhaps `find_a_suitable_provider()`?) to prefer non-RXD providers before RXD-based providers...?

I was worried that since the EFA DGRAM endpoint correctly advertises `FI_LOCAL_COMM | FI_REMOTE_COMM` that we'd pick up RxD + EFA in the first block and then skip the else block where the special cases are handled, i.e. where those capability bits are skipped. It does indeed advertise the correct flags for the MTL to choose this provider:
```
$ fi_info --caps='FI_TAGGED|FI_LOCAL_COMM|FI_REMOTE_COMM' -p efa
provider: efa;ofi_rxd
    fabric: EFA-fe80::54:efff:fe51:5242
    domain: efa_0-dgrm
    version: 1.0
    type: FI_EP_RDM
    protocol: FI_PROTO_RXD
```

But testing with your patch it appears that an error is hit when it chooses RxD in the AV open path anyway so maybe we'll be okay with the libfabric versions that are already in the field.
```
[ip-172-31-12-156.us-east-2.compute.internal:29276] mtl_ofi_component.c:385: mtl:ofi:prov: efa;ofi_rxd
[ip-172-31-12-156.us-east-2.compute.internal:29276] mtl_ofi_component.c:580:fi_av_open failed: Invalid argument
[ip-172-31-12-156.us-east-2.compute.internal:29276] select: init returned failure for component ofi
[ip-172-31-12-156.us-east-2.compute.internal:29276] select: no component selected
```

jsquyres:
@rwespetal Er... I'm confused by your reply.  😄

Does the proposed pseudocode work for EFA?  I.e.:

* Does it work for EFA in existing Libfabric releases (where EFA does not advertise FI_LOCAL|FI_REMOTE)
* Does it work for EFA in future Libfabric releases (where EFA does advertise FI_LOCAL|FI_REMOTE)

?

rwespetal:
I apologize for the confusing reply on my part.

The pseudocode will work with EFA for future releases where the local/remote capability flags have been fixed.

The pseudocode will not work with EFA for existing Libfabric releases as it will choose `efa;ofi_rxd` as the provider instead of `efa` provider (using the RDM endpoint). This is because in the old versions of Libfabric the EFA RDM endpoint does not advertise the right capabilities, but the EFA DGRAM endpoint does and RxD layers on it. This assumes that the pseudocode is only selecting the provider and not attempting to open it until later which I'm guessing is the case based on the current OFI MTL implementation.

The `efa;ofi_rxd` provider fails to initialize correctly, at least with a build from Libfabric master that I tested, so I was implying that perhaps we can invoke the old getinfo path (where the local/remote flags are not set) if provider initialization fails when choosing a provider in the new getinfo path (where the local/remote flags are set). However, that seems like it could hide bugs if a different provider than `efa;ofi_rxd` is chosen in the new getinfo path and fails to initialize for some other reason. So that's probably not a great idea.

rwespetal:
Hi Jeff,

Just wondering if you had any more thoughts on this. I think the only way to work around the EFA bug in the wild is to add a provider specific check for EFA but if that's not the case let me know. I think it will look something like this and should take care of the `usnic;ofi_rxd` case as well. I can write a patch if this looks reasonable:

```c
ompi_mtl_ofi_component_init() {

    // snip

    hints += FI_VERSION(1, 5)
    
    // Special case to select EFA RDM provider if selected or available
    if (efa_in_include_list() || efa_not_in_exclude_list()) {
        hints->fabric_attr->prov_name = 'efa';
        ret = fi_getinfo(hints, &prov);
        if (0 == ret) {
            // We found EFA, use it
            goto ofi_init;
        } else {
            // Return if fatal, otherwise fall through
            if (-FI_ENODATA != ret)
                goto error;
        }
    }
    
    // Normal path
    hints += (FI_LOCAL + FI_REMOTE);
    fi_getinfo(hints, &prov);
    prov = select_ofi_provider(prov)
    
    if (!prov)
        goto error;

ofi_init:
    // rest of mtl ofi init code

}
```

bwbarrett:
I obviously don't love @rwespetal's solution, but I can't think of a better one.  However, it needs a much bigger comment about why that code exists :).

jsquyres:
@rwespetal So sorry about the delay in replying -- SC/Thanksgiving/MPI Forum/kitchen sink/etc...

Yeah, I think your solution is probably going to end up being the best one.  I agree with @bwbarrett that this probably deserves quite a bit longer comment, because this is a tangled story we've got here, and our Future Selves are going to need a description to remind us why it is there.

Since this directly affects EFA and I can't test that, would you mind updating this PR?

rwespetal:
Not a problem. I'll write a patch for this and open a PR soon.

jsquyres:
@rwespetal Cool, thanks!

Note that you can just push to my fork / this branch.  Github pro tip: when you open a PR that's from your fork to a group repo, the permissions on the branch in your fork (by default) take on the permissions of the branch in the group repo.  So if you have write perms on the main repo, you have write perms on my branch in this PR.

rwespetal:
Thanks for the tip, @jsquyres! I didn't realize that I could do that.

I've updated the PR with an additional commit to workaround the EFA bug. I also noticed that there's an issue where the 'UDP' provider gets picked up due to case sensitivity so I will open another PR for that.

jsquyres:
False cray failure: bot:ompi:retest

jsquyres:
@rwespetal Can you add some `opal_output_verbose()` statements to show the progression through the new code?

I don't immediately see why the new code would cause problems for @mwheinz -- so it might be worthwhile to add these verbose statements (that may ultimately be useful for end users, too).

rwespetal:
@jsquyres - Sure, I can add some prints to the code. For what it's worth, the patch took the right paths with and without an EFA device loaded and worked fine with the `efa` and `tcp;ofi_rxm` providers.

rwespetal:
I added some prints to the last patch so that it's easier to tell what path it's taking.

mwheinz:
I've re-confirmed to my own satisfaction that the warnings only appear in RHEL 8.0 (and possibly higher) but not RHEL 7.x. I've also re-confirmed that the problem disappears if I revert this patch.

So, next step for me is I'm going to try adding my own debug statements, see if I can scrape some more info to see what's going on.

mwheinz:
Jeff, the not-valid error message appears in a couple of places in the source, so I threw in a few debug print statements and the problem appears to be in component_find_check() where ofi isn't being listed as a component:
```
[LINUX hdsmpriv01 20200108_1339 mpi_apps]# /usr/mpi/gcc/openmpi-4.1.0/bin/mpirun -np 2 -map-by node --allow-run-as-root -machinefile /usr/src/opa/mpi_apps/mpi_hosts  -mca mtl ofi osu2/osu_bw
[hdsmpriv01:06987] MWHEINZ: Requested Name=pmi, Component=pmi
[hdsmpriv01:06987] MWHEINZ: Requested Name=pmi, Component=pmi
[hdsmpriv01:06987] MWHEINZ: mca_base_component_find.c:337
A requested component was not found, or was unable to be opened.  This
means that this component is either not installed or is unable to be
used on your system (e.g., sometimes this means that shared libraries
that the component requires are unable to be found/loaded).  Note that
Open MPI stopped checking at the first component that it did not find.

Host:      hdsmpriv01
Framework: mtl
Component: ofi
```

What's odd Is that the selected benchmark continues to execute even though I specified ofi as the only transport to use! Next up - why are "pmi" and "pmi" failing to match each other...?

Stack trace at the time the help message was displayed:

```
#2  0x00007f803c194cba in component_find_check (
    framework=framework@entry=0x7f803cff07e0 <ompi_mtl_base_framework>, 
    requested_component_names=<optimized out>) at mca_base_component_find.c:342
#3  0x00007f803c194f6a in component_find_check (requested_component_names=<optimized out>, 
    framework=0x7f803cff07e0 <ompi_mtl_base_framework>) at mca_base_component_find.c:317
#4  mca_base_component_find (directory=directory@entry=0x0, 
    framework=framework@entry=0x7f803cff07e0 <ompi_mtl_base_framework>, 
    ignore_requested=ignore_requested@entry=false, open_dso_components=open_dso_components@entry=true)
    at mca_base_component_find.c:145
#5  0x00007f803c1a0c0e in mca_base_framework_components_register (
    framework=framework@entry=0x7f803cff07e0 <ompi_mtl_base_framework>, 
    flags=flags@entry=MCA_BASE_REGISTER_DEFAULT) at mca_base_components_register.c:55
#6  0x00007f803c1a1198 in mca_base_framework_register (framework=0x7f803cff07e0 <ompi_mtl_base_framework>, 
    flags=<optimized out>) at mca_base_framework.c:133
#7  0x00007f803c1a1255 in mca_base_framework_open (
    framework=framework@entry=0x7f803cff07e0 <ompi_mtl_base_framework>, 
    flags=flags@entry=MCA_BASE_OPEN_DEFAULT) at mca_base_framework.c:168
#8  0x00007f80264071b6 in mca_pml_cm_component_open () at pml_cm_component.c:112
#9  0x00007f803c1969c3 in open_components (framework=0x7f803cff31c0 <ompi_pml_base_framework>)
    at mca_base_components_open.c:125
#10 mca_base_framework_components_open (
    framework=framework@entry=0x7f803cff31c0 <ompi_pml_base_framework>, 
    flags=flags@entry=MCA_BASE_OPEN_DEFAULT) at mca_base_components_open.c:68
#11 0x00007f803cdab4bf in mca_pml_base_open (flags=MCA_BASE_OPEN_DEFAULT) at base/pml_base_frame.c:185
#12 0x00007f803c1a12c9 in mca_base_framework_open (framework=0x7f803cff31c0 <ompi_pml_base_framework>, 
    flags=flags@entry=MCA_BASE_OPEN_DEFAULT) at mca_base_framework.c:196
#13 0x00007f803cdb8ed8 in ompi_mpi_init (argc=<optimized out>, argv=<optimized out>, requested=0, 
    provided=provided@entry=0x7ffe5e346474, reinit_ok=reinit_ok@entry=false) at runtime/ompi_mpi_init.c:620
#14 0x00007f803cd5f6b2 in PMPI_Init (argc=0x7ffe5e3464cc, argv=0x7ffe5e3464c0) at pinit.c:67
```

jsquyres:
@mwheinz Is the "not found" selection for the `osc` framework (i.e., MPI one-sided)?  I.e., it finds MTL/OFI just fine for point-to-point operations, but finds nothing for one-sided?

On whatever branch you're testing, see what `ompi_mpi_init()` is initializing at `runtime/ompi_mpi_init.c:620`.

mwheinz:
> On whatever branch you're testing, see what `ompi_mpi_init()` is initializing at `runtime/ompi_mpi_init.c:620`.

Looks like PML:

```
 620     if (OMPI_SUCCESS != (ret = mca_base_framework_open(&ompi_pml_base_framework, 0))) {
 621         error = "mca_pml_base_open() failed";
 622         goto error;
 623     }
```

mwheinz:
Okay, I've got a pair of detailed logs, one for an rhel7 run (successful) and one for an rhel8 run (unsuccessful).

TLDR, it looks like the RHEL8 version isn't loading components that the RHEL7 version can. (in the diff fragment below, "<" is RHEL 7, ">" is RHEL 8.)

    1 6,9c6
     2 < [host:1] mca: base: components_register: found loaded component sm
     3 < [host:1] mca: base: components_register: found loaded component ofi
     4 < [host:1] mca: base: components_register: component ofi register function successful
     5 < [host:1] mca: base: components_register: found loaded component self
     6 ---
     7 > [host:1] mca: base: components_register: found loaded component tcp
     8 11,13c8,9
     9 < [host:2] mca: base: components_register: found loaded component sm
    10 < [host:2] mca: base: components_register: found loaded component ofi
    11 < [host:1] mca: base: components_register: component self register function successful
    12 ---
    13 > [host:2] mca: base: components_register: found loaded component tcp
    14 > [host:1] mca: base: components_register: component tcp register function successful
    15 15,16c11,12
    16 < [host:2] mca: base: components_register: component ofi register function successful
    17 < [host:2] mca: base: components_register: found loaded component self


[rhel7.log.gz](https://github.com/open-mpi/ompi/files/4040798/rhel7.log.gz)
[rhel8.log.gz](https://github.com/open-mpi/ompi/files/4040799/rhel8.log.gz)

mwheinz:
I'm wondering if the issue isn't the massive difference between gcc version on the two machines.

RHEL 8.0 comes with GCC 8.2.1 while RHEL 7.6 comes with gcc 4.8.5. Unfortunately, installing an RPM built for RHEL 7.6 on an RHEL 8.0 machine is non-trivial. I'm exploring this issue now.

mwheinz:
Finally figured it out. Problem is in the configure step:
checking if we have OFI API version >=v1.5... ./configure: line 306738: test: -gt: unary operator expected
./configure: line 306738: test: -ge: unary operator expected
no
checking if MCA component mtl:ofi can compile... no

jsquyres:
@mwheinz and I iterated on this via webex and figured it out: his CPP (default for CentOS/RHEL 8 -- from gcc 8.2.1) is emitting stuff differently than mine (hand-compiled gcc 8.2.0) (!!!).

what

New commit pushed that should hopefully handle both his and my CPPs.

jsquyres:
@bwbarrett You requested changes on this PR before; can you review again and hopefully remove your negative review on this PR?  Thanks.

(@mwheinz still needs to certify this on his systems; I'm just proactively clearing up Brian's old review even if we're not 100% verified on this yet because I think Brian's objections are solved by @rwespetal's commit)

config/opal_check_ofi.m4
```diff
@@ -81,6 +135,30 @@ AC_DEFUN([_OPAL_CHECK_OFI],[
...
+        dnl See comment above about why we call this macro twice, once
+        dnl with nothing and once with -P.
+        _OPAL_CHECK_OFI_GET_API_VERSION([])
+        AS_IF([test -z "$OFI_MAJOR_VERSION" || test -z "$OFI_MINOR_VERSION"],
+              [_OPAL_CHECK_OFI_GET_API_VERSION([-P])])
+
+        dnl Sanity check: make sure we got values
+        AS_IF([test -z "$OFI_MAJOR_VERSION" || test -z "$OFI_MINOR_VERSION"],
+              [AC_MSG_RESULT([unknown])
+               AC_MSG_WARN([unable to run the C preprocessor])
```
bwbarrett: This warning is misleading; we're not sure why we don't have a number, but the C preprocessor might have run fine.

config/opal_check_ofi.m4
```diff
@@ -81,6 +135,30 @@ AC_DEFUN([_OPAL_CHECK_OFI],[
...
+               AC_MSG_ERROR([Unable to continue])],
```
bwbarrett: It feels very weird to error here, because this is a test that might be unnecessary (if the user is only building the OFI BTL, not the OFI MTL, for example).  I think at a high level, this all should be in its own macro that is called only by the OFI MTL.  Further, get rid of all this code and, since we don't actually care about the version, only that it's greater than some number, have a macro OFI_MIN_VERSION(1,5) which expands to something like:

```
AC_TRY_COMPILE([#include <ofi.h>
#include <rdma/fabric.h>],[
#if FI_VERSION_LT(FI_VERSION(FI_MAJOR_VERSION, FI_MINOR_VERSION), FI_VERSION($1, $2))
#error "Nopes on the version
#endif 
```

jsquyres: Yes, that is better.  I'll update.

jsquyres: @bwbarrett Done.  Can you re-review?

jsquyres:
> Note that you did change the minimum version for the usnic btl from 1.1 to 1.3 :).

Yep.  Configure was checking for 1.1, but the usnic component was checking for 1.3 at runtime.

mwheinz:
Looks good. I've verified the updated patch works compiles correctly with the GCC provided by RHEL 8.0.

jsquyres:
I just noticed that my original commit message was now a bit stale -- this most recent push was just updating the commit message to be current (no code changes).

jsquyres:
@rwespetal @bwbarrett I went to make a v4.0.x PR for this, but there's some minor conflicts in cherry picking @rwespetal's commit to the v4.0.x branch.  I don't really have a good mechanism to test the OFI MTL -- can you guys take care of making a v4.0.x PR for this PR?

rwespetal:
@jsquyres  - Sure, I can open a PR against v4.0.x.

PR #7175, opened 2019-11-18
OPAL/MCA/BTL/OPENIB: Detect ConnectX-6 HCAs

Akshay-Venkatesh:
@jsquyres @gpaulsen 

Hi Jeff, Geoff,

Some users who still rely on OpenIB support in 4.0.x and 3.\*.x ran into issues when running OpenMPI on ConnectX-6 HCAs becasue OpenIB BTL doesn't detect them. This PR adds checks to avoid `OpenFarbrics device initialization error`. If the PR is good enough, I can make equivalent PRs for 3.\*.x as well. Lmk.

### DIFF
--- a/opal/mca/common/verbs/common_verbs_port.c
+++ b/opal/mca/common/verbs/common_verbs_port.c
@@ -94,6 +94,10 @@ int opal_common_verbs_port_bw(struct ibv_port_attr *port_attr,
         /* 12x */
         *bandwidth *= 12;
         break;
+    case 16:
+        /* 16x */
+        *bandwidth *= 16;
+        break;
     default:
         /* Who knows? */
         return OPAL_ERR_NOT_FOUND;
--- a/opal/mca/btl/openib/mca-btl-openib-device-params.ini
+++ b/opal/mca/btl/openib/mca-btl-openib-device-params.ini
@@ -190,6 +190,15 @@ max_inline_data = 256
 
 ############################################################################
 
+[Mellanox ConnectX6]
+vendor_id = 0x2c9,0x5ad,0x66a,0x8f1,0x1708,0x03ba,0x15b3,0x119f
+vendor_part_id = 4123
+use_eager_rdma = 1
+mtu = 4096
+max_inline_data = 256
+
+############################################################################
+
 [IBM eHCA 4x and 12x]
 vendor_id = 0x5076
 vendor_part_id = 0


opal/mca/common/verbs/common_verbs_port.c
```diff
@@ -94,6 +94,10 @@ int opal_common_verbs_port_bw(struct ibv_port_attr *port_attr,
         /* 12x */
         *bandwidth *= 12;
         break;
+    case 16:
+        /* 12x */
+        *bandwidth *= 12;
```
gpaulsen: Are you sure you wanted the bandwidth to be *= 12 in this case?

Akshay-Venkatesh: @gpaulsen Thanks for the catch. I confirmed that it should be 16 (UCX does the same https://github.com/openucx/ucx/blob/master/src/uct/ib/base/ib_iface.c#L1033)
Will make this change now.

hppritcha:
@gpaulsen looks good to me, go ahead and merge if you have time

opal/mca/common/verbs/common_verbs_port.c
gpaulsen: @Akshay-Venkatesh ^^^ ?

jsquyres:
Why isn't this a cherry pick from master?

Akshay-Venkatesh:
> 
> 
> Why isn't this a cherry pick from master?

@jsquyres Wasn't openib wiped in master? I don't see the directory anymore here: https://github.com/open-mpi/ompi/tree/master/opal/mca/btl

Akshay-Venkatesh:
@hppritcha Should I submit similar PRs to 3.\*.x?

rhc54:
> @jsquyres Wasn't openib wiped in master?

We wiped @jsquyres brain at the same time. 😄

jsquyres:
> @jsquyres Wasn't openib wiped in master? I don't see the directory anymore here: https://github.com/open-mpi/ompi/tree/master/opal/mca/btl

Duh -- yes it was.  Oops!

To avoid dumb questions like this from me in the future 😄, it is probably good to cite why release branch commits are not cherry-picks from master in the commit message and/or the github PR description.

Yes, please file PR's for v3.0.x and v3.1.x.

Thanks!

PR #7177, opened 2019-11-19
fix ipv6 btl connection bug

hppritcha:
Signed-off-by: Stanislav Kirillov <staskirillof@yandex.ru>
(cherry picked from commit 0e0763e0066a5eb886468f2334b7af6c37453836)

Conflicts:
	opal/mca/btl/tcp/btl_tcp_proc.c

### DIFF
--- a/opal/mca/btl/tcp/btl_tcp_proc.c
+++ b/opal/mca/btl/tcp/btl_tcp_proc.c
@@ -956,7 +956,7 @@ bool mca_btl_tcp_proc_tosocks(mca_btl_tcp_addr_t* proc_addr,
             struct sockaddr_in6* inaddr = (struct sockaddr_in6*)output;
             output->ss_family = AF_INET6;
             memcpy(&inaddr->sin6_addr, &proc_addr->addr_inet,
-                   sizeof (proc_addr->addr_inet));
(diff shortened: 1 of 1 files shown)


hppritcha:
hmmm looks like we're missing some commits

hppritcha:
looks like we'd need a number of other ipv6 related changes from master to - 457f058e etc.

gpaulsen:
Hmmm...  457f058 looks quite big to bring back to v4.0.x.  What are your thoughts?

hppritcha:
I'll see about doing a different approach for 4.0.x that avoids pulling in all these other changes.

bwbarrett:
457f058 is mostly a bugfix, if that helps.  It's also really just a structure swap, so it's either going to compile and work or fail to compile and not work.

gpaulsen:
@kizill, Can you please test this change on v4.0.x?  We're hoping that this is equivalent to https://github.com/open-mpi/ompi/pull/7082 without requiring backporting all of https://github.com/open-mpi/ompi/pull/5938.

hppritcha:
this appears to work for me. 
@bwbarrett please double check when you can.

PR #7184, opened 2019-11-21
cray ftn: modify fortran module loc checker

hppritcha:
to support the Cray Fortran compiler.  Cray Fortran compiler does not
contain all symbol info in the module file, have to link with the *.o
created as part of module file compilation.

### DIFF
--- a/config/ompi_fortran_find_module_include_flag.m4
+++ b/config/ompi_fortran_find_module_include_flag.m4
@@ -11,6 +11,8 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2010-2012 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2019      Triad National Security, LLC. All rights
+dnl                         reserved.
 dnl $COPYRIGHT$
 dnl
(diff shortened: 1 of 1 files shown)


ggouaillardet:
Could a compiler choose **not** to emit a `.o` file if there is no code in it?

jsquyres:
> Could a compiler choose **not** to emit a `.o` file if there is no code in it?

The module in question has a variable instance in it -- so I think something must be emitted, right?

hppritcha:
@ggouaillardet has a point.  I'll rework to check for case of a .o or not.

jsquyres:
A few lines above your code, there's this:

```
OPAL_LOG_COMMAND([$FC $FCFLAGS $FCFLAGS_f90 -c conftest-module.f90 $LDFLAGS $LIBS] ...
```

The `-c` means that a `.o` _must_ be emitted, right?

hppritcha:
@jsquyres and @ggouaillardet I could see that there might be some weird compiler that only emits a *.mod for this simple example.  I tweaked the code some to try both with and without a *.o.

hppritcha:
cori is again discomfited
bot:ompi:retest

jsquyres:
@hppritcha I made a minor code suggestion in a new commit on this PR.  Feel free to squash or discard.

hppritcha:
bot:ompi:retest

PR #7191, opened 2019-11-24
Add the missing code to check a return code

gvallee:
Signed-off-by: Geoffroy Vallee <geoffroy.vallee@gmail.com>


jsquyres:
False cray failure.

bot:lanl:retest

Issue #7095, opened 2019-10-17
Bug or Wrong documentation for --output-filename option


ericch1:
Hi,

In the documentation of mpixec, the option "-output-filename, --output-filename <filename>" is described as:

Redirect the stdout, stderr, and stddiag of all processes to a process-unique version of the specified filename. Any directories in the filename will automatically be created. Each output file will consist of filename.id, where the id will be the processes’ rank in MPI_COMM_WORLD, left-filled with zero’s for correct ordering in listings. A relative path value will be converted to an absolute path based on the cwd where mpirun is executed. Note that this will not work on environments where the file system on compute nodes differs from that where mpirun is executed.

Refs:
https://www.open-mpi.org/doc/v4.0/man1/mpiexec.1.php
https://www.open-mpi.org/doc/v3.1/man1/mpiexec.1.php
https://www.open-mpi.org/doc/v3.0/man1/mpiexec.1.php
https://www.open-mpi.org/doc/v2.1/man1/mpiexec.1.php
https://www.open-mpi.org/doc/v2.0/man1/mpiexec.1.php
https://www.open-mpi.org/doc/v1.10/man1/mpiexec.1.php

However, as of version 3.x and 4.x, the filename is *not* generated, it is instead a *directory*.
Is it a bug or a wrong documentation?  Is there a way to have the same behavior as of 2.X and before?

to reproduce this error:

```
[eric@lorien] bug (master $ u=)> mpiexec -n 1 --output-filename out.txt echo "Hi"
Hi
[eric@lorien] bug (master $ u=)> ls -latr
total 8
drwxr-xr-x 18 eric giref 4096 Oct 17 09:59 ..
drwx------  3 eric giref   15 Oct 17 10:00 out.txt
drwx------  3 eric giref   21 Oct 17 10:00 .
[eric@lorien] bug (master $ u=)> ls -la out.txt/
total 0
drwx------ 3 eric giref 15 Oct 17 10:00 .
drwx------ 3 eric giref 21 Oct 17 10:00 ..
drwx------ 3 eric giref 20 Oct 17 10:00 1
[eric@lorien] bug (master $ u=)> ls -la out.txt/1/
total 0
drwx------ 3 eric giref 20 Oct 17 10:00 .
drwx------ 3 eric giref 15 Oct 17 10:00 ..
drwxr-x--- 2 eric giref 34 Oct 17 10:00 rank.0
[...]

PR #6578, opened 2019-04-08
Implement a MCA framework for threads

hppritcha:
Add a framework to support different types of threading models including
user space thread packages such as Qthreads and argobots:

https://github.com/pmodels/argobots

https://github.com/Qthreads/qthreads

The default threading model is pthreads.  Alternate thread models are
specificed at configure time using the --with-threads=X option.

The framework is static.  The theading model to use is selected at
Open MPI configure/build time.

PR #7200, opened 2019-11-26
Fix unsafe use of gethostname()

cpshereda:
Refs: open-mpi#6801.

Change calls to gethostname() to call new opal_gethostname() function that safely populates hostname:

`const char *opal_gethostname( void )`

Although this function returns the desired hostname and not a return code, values for errno that indicate a memory allocation problem or a truncated result are preserved and/or properly set.

Two new files:
`opal/util/opal_gethostname.c`
`opal/util/opal_gethostname.h`

All other mods are necessary changes to call the above function instead of gethostname.


cpshereda:
Refs: open-mpi#6801

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,173 @@
...
+                    break;
+                }
+                /*
+                 * The buffer is empty. In some gethostname
+                 * implementations this can be because the buffer was too 
+                 * small. But, if it is zero for some other reason, we
+                 * could get caught in the loop. So, break after
+                 * NUM_TRIES_FOR_NULL_HOSTNAME. 
+                 */
+                else if( 0 == count ) {
```
awlauria: What version of `gethostname()` returns 0 (success) yet has an empty buffer?

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,173 @@
...
+
+
+const char *opal_gethostname( void )
+{
+    size_t count;
+    size_t length = OPAL_LOCAL_MAXHOSTNAMELEN;
+    int ret_val, num_tries = 0;
+    int temp_errno = 0;
+
+    if( first_pass ) {
```
awlauria: Are we worried about thread safety here? It's conceivable that two threads can be in here at the same time for first the first_pass case. Might lead to some goofy behavior.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,173 @@
...
+                    num_tries++;
+                }
+                /* The result *may* be truncated (can't know for sure, so we increase 
+                 * the length of the buffer (happens below) and try again). */
+                else if( count == length - 1 ) {
+			; /* no op */
+                }
+                /* Unidentified case, implies some bug in this code or some unknown
+                 * return string from gethostname() */
+                else {
```
awlauria: We're in the success block (IE - a return value of 0) - so I'm not sure what case this might cover.

My understanding is if it returns 0, it succeeded - with a possibility that the hostname may have been truncated.

opal/util/opal_gethostname.c
cpshereda: I am trying to account for all possible cases. 
http://perfec.to/gethostname/
While there is not a case of 0 and an empty buffer listed here, I have no evidence that it's not possible. Certainly it does no harm.

opal/util/opal_gethostname.c
cpshereda: I'm going to let others comment on whether or not this needs to be reentrant. I am new to this effort and did not code it with that in mind. Having said that, I followed @jsquyres 's suggestion on this in the original issue - open-mpi#6801. Perhaps you should read that issue to catch up. A comment there from you earlier might have been helpful.

opal/util/opal_gethostname.c
cpshereda: See above. I am trying to account for all possible behaviors of `gethostname()`, even ones that we would not expect.

opal/util/opal_gethostname.c
awlauria: I'm not so worried about reentrantcy as we are of two threads calling this at the same time, and each thread executing the `first_pass` block at the same time. Considering the scope of this change, I'd bet that it is *possible*, but maybe not probable.

opal/util/opal_gethostname.c
awlauria: There's certainly nothing wrong with trying to account for all cases. I asked this question because in your comment you mentioned "in some implementations..." which made me think there was a case or a version or a version of `gethostname()` where this was possible. If there isn't, maybe it can be tightened up a little bit.

opal/util/opal_gethostname.c
hppritcha: I think this concern could be addressed by a statically initialized mutex to protect accessing ```first pass```

cpshereda: Ok, if statically initialized mutexes are already used elsewhere in other parts of the code, I'll make the necessary change.

opal/util/opal_gethostname.c
cpshereda: Well, I was using this writeup as a guide more than the prior one I pasted: http://www.microhowto.info/howto/determine_the_fully_qualified_hostname_of_the_local_machine_in_c.html

It's a more complicated and lengthier discussion, but the way I read it, it seemed to imply that a non-error return code and a null buffer were both possible in some implementations.

hppritcha:
@cpshereda please use the opal mutex abstractions to address thread safety.    See what's hopefully coming soon as motivation - #6578

orte/orted/orted_main.c
```diff
@@ -209,10 +211,6 @@ opal_cmd_line_init_t orte_cmd_line_opts[] = {
       &orted_globals.singleton_died_pipe, OPAL_CMD_LINE_TYPE_INT,
       "Watch on indicated pipe for singleton termination"},
 
-    { "orte_output_directory", '\0', "output-directory", "output-directory", 1,
-      NULL, OPAL_CMD_LINE_TYPE_STRING,
-      "Redirect output from application processes into filename/job/rank/std[out,err,diag]." },
-
```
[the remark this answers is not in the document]
rhc54: Please don't remove it - there are people using it (we just finished a lengthy exchange over it on #7095 and the mailing list)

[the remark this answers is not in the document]
cpshereda: No, I didn't mean to - I didn't touch this part of the code. Were commits put it recently for this?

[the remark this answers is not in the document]
jsquyres: You might want to rebase on top of current HEAD?

cpshereda:
> Ah, I see you removed some (all?) of the fprintf/printf's in the final commit.
> 
> Can you squash commits that simply fix/change things from prior commits? I.e., it would be great to get a sequence of commits that are each correct -- they don't need a subsequent commit to be correct.

I will squash when we are ready to merge, yes. There was a logic to my commits, though, and that was important for my purposes, so I was kind of surprised to see all the comments about the printf's. Knowing that I (and maybe others) will not have all commits be squashed at the point a PR is opened, can you work from most recent backward rather than from first forward? Or do you want everything to be squashed at the point the PR is opened?

> Additionally, you can probably remove all the "Changes to be committed..." lines from all your commit messages. The git commit contains that metadata; listing it in the commit message is redundant.
> 
I personally like the file list in the commit message but yes, I can remove it from the squashed commit.

> Is there a new test program to be committed with this PR (that requires -DTESTING=1)? If not, the TESTING stuff should probably be deleted.

I asked you a question about this a while ago on the original issue but I didn't get an answer. Here it is again, along with some other questions. Don't worry about the ompi/examples question, Howard told me we aren't touching anything there. But as for the rest, what do you think?

> @jsquyres - Before I open a PR for these changes, any suggestions for what can help me test all of them as thoroughly as possible? @hppritcha suggested the IBM component of MTT (Actually LAM?) for testing mpi, but I'm not really clear on how to test everything else.
> 
> Per Howard's suggestion and recent developments, I changed more than just opal/, such as some codefiles in orte/ and some in ompi/. How would I test what's been changed in orte/, for example?
> 
> Also, if I have changes to ompi/examples, how can we incorporate that since it's excluded via .gitignore?
> 
> Finally, I have a standalone version of this to put somewhere, as a tester. It's not using MPI. Where would I put it? Or do we not put that in the master branch?

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,186 @@
...
+ * 
+ * Copyright (c) 2019      Triad National Security, LLC.
+ *                         All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ * Based on pgethostname (a portable gethostname)
+ * available at http://perfec.to/gethostname/
+ * and Graham Shaw's sample code at
+ * http://www.microhowto.info/howto/determine_the_fully_qualified_hostname_of_the_local_machine_in_c.html
```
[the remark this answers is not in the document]
cpshereda: Attribution should be adequate for the first (perfec.to/gethostname) as I don't think I used a single line of code from his work, just gleaned ideas from it - and, he does not mention anything about licensing on that site.

As for the second, I think I did use some of that code, at least as a base, and here is his license:

> Licence (source code)
> 
> This is a free licence, however it applies only to the samples of source code included on this site. (For ease of identification these are enclosed within HTML <pre> elements of class code.) It does not apply to the remainder of the site, including in particular the supporting text.
> 
>     Copying and distribution of this software, with or without modification, is permitted in any medium without royalty. This software is offered as-is, without any warranty.
> 

I didn't think I needed to do more than attribute for a license like that. Do you agree?

orte/test/mpi/singleton_client_server.c
```diff
@@ -81,12 +84,18 @@ int main(int argc, char *argv[])
 
   /* get the node name */
   {
-    int retval = gethostname(hostname, sizeof(hostname));
-    if(retval == -1)
+    temp_errno = errno;
```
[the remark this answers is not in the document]
cpshereda: Because this `opal_gethostname()` function returns a char* rather than a return code, I am replicating the return code check that was here (that I removed) by zeroing out errno (which could be anything at this point) and then checking to see if there is a failure-type errno on return. If there isn't, I put it back to what it previously was because I want to avoid any unintended side effects. Remember, I'm touching a lot of code, I don't know what it's all doing, so I am trying to be cautious here.

I should probably only restore errno if it's zero on return from the function call, so that is a bug that I will fix. So thanks for bringing that to my attention. Also, because I'll be setting errno to ENOMEM if the malloc or realloc checks fail (per your suggestion), I will need to add that.

I realize this is not ideal but when it was decided that `opal_gethostname()` would return a char*, the die was cast for some clunkiness if we want to see if it succeeded or failed. One alternative is to just no longer check but that seems bad, especially in the case of a memory fail.

ompi/tools/mpisync/sync.c
```diff
@@ -99,11 +104,31 @@ int main(int argc, char **argv)
         MPI_Finalize();
         exit(1);
     }
-
-    if( gethostname(hname, sizeof(hname)) ){
-        perror("Cannot get hostname. Abort");
+    /* zero out errno so we don't mistakenly enter the condition below */
+    temp_errno = errno;
```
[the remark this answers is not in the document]
cpshereda: Same as prior explanation.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,186 @@
...
+                 * implementations this can be because the buffer was too 
+                 * small. But, if it is zero for some other reason, we
+                 * could get caught in the loop. So, break after
+                 * NUM_TRIES_FOR_NULL_HOSTNAME. 
+                 */
+                else if( 0 == count ) {
+                    if( num_tries > NUM_TRIES_FOR_NULL_HOSTNAME ) {
+                        ret_val = -1;
+                        temp_errno = EINVAL;
+                        fprintf( stderr, "hostname returned empty result\n" );
```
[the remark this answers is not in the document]
cpshereda: When commits are squashed this will go away.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+            /*
+             * hostname was truncated and there was an error. Perhaps
+             * there is something in the buffer and perhaps not. 
+             * Double the length and try again. 
+             */
+            else if( EINVAL == errno || ENAMETOOLONG == errno ) {
+                temp_errno = errno;
+                /*printf("Hostname truncated in call to gethostname(), retrying with longer buffer: length = %zu, errno = %d\n", length, errno); */
```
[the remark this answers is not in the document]
cpshereda: When commits are squashed this will go away.

awlauria:
Thanks for adding the mutex lock.

As an enhancement around that, all you really need to do is protect `first_pass`, which will significantly shorten your critical section, instead of the entire if statement like it currently does. You could do something like this instead:

```
opal_mutex_lock(lock);
bool real_first_pass = first_pass;
first_pass = false;
opal_mutex_unlock(lock);

if(real_first_pass){
... your logic.
}
```

devreal:
@awlauria Unfortunately, that is not sufficient as the second thread entering the function may attempt to use the hostname before it is populated.

Suggestion: check `first_pass` (which should be static in the scope of the function), if `true` take the lock and check again for `first_pass` to make sure there was no other thread that worked on it. If still `true`, assemble `opal_local_hostname` and only then set `first_pass` to `false`. That way you can save both the temporary (which is not stable anyway, multiple threads may copy into it) and avoid excessive locking on every call.

rhc54:
before we spend a lot of effort on this, let's just remember that ```gethostname``` is not in the critical path. Just put a lock around the entire thing and move on.

awlauria:
@devreal ah yes, you are right. Good catch!. I'm not quite following how your solution shortens the critical section though - an you illustrate? Maybe I just need some more :coffee: ...

@rhc54 you are right, here it will only really affect the first pass go around - so maybe we shouldn't bang our heads against it for minimal gain.

cpshereda:
@devreal @awlauria The additional checking seems like it has the potential for me to introduce a new race condition if I'm not really careful. I will push a squashed commit that has the final strcpy inside the lock (that was an oversight/bug anyway that my testing didn't catch). 

Plus the temporary's presence doesn't have to do with threading, it has to do with the fact that a caller could modify the return and then mess it up for subsequent calls.

devreal:
@cpshereda `opal_gethostname` returns a `const char*` so if the caller decides to ignore (cast away) the `const` constraint and alters the string it's a violation of the interface. It is an internal interface anyway so we should not assume the presence of adversarial callers. Those things should be caught during code reviews.

jsquyres:
I agree with the above sentiments: this is not a performance-critical piece of code, so simple = good.  It has to be thread safe, but it doesn't need to optimize performance.

cpshereda:
Well, I admit it turns out to be more than theoretical. I did not discover this snippet with my testing when I first opened the PR (this is inside `orte/util/proc_info/c`); I didn't catch this when changing the `gethostname` calls to `opal_gethostname()`. I supposed I could fix this here instead, and remove the temporary from within `opal_gethostname.c`, but my thinking was, if it can happen here without triggering a compiler warning, it can happen elsewhere, so I chose to protect the static variable inside `opal_gethostname`. I'm open to doing whatever others judge to be best though.

(The lack of compiler warning is at least occurring on a Mac with clang, and it wouldn't surprise me if it isn't warned about anywhere given the call to `strchr`.)

Note hostname declaration:

```
int orte_proc_info(void)
{

    int idx, i;
    char *ptr;
    const char *hostname;
    char **prefixes;
    bool match;
```

Problematic section:

```
   /* get the nodename */
    hostname = opal_gethostname();
    /* add this to our list of aliases */
    opal_argv_append_nosize(&orte_process_info.aliases, hostname);

    // Strip off the FQDN if present, ignore IP addresses
    if( !orte_keep_fqdn_hostnames && !opal_net_isaddr(hostname) ) {
        if (NULL != (ptr = strchr(hostname, '.'))) {
            *ptr = '\0';
            /* add this to our list of aliases */
            opal_argv_append_nosize(&orte_process_info.aliases, hostname);
        }
    }
```

orte/orted/orted_main.c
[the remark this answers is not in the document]
cpshereda: The re-commit of Ralph's commit just now was not intended, FYI.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,186 @@
...
+
+const char *opal_gethostname( void )
+{
+    size_t count;
+    size_t length = OPAL_LOCAL_MAXHOSTNAMELEN;
+    int ret_val, num_tries = 0;
+    int temp_errno = 0;
+
+    if( first_pass ) {
+        char *buf = malloc( length );
```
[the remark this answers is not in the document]
cpshereda: Should be resolved.

opal/util/opal_gethostname.c
cpshereda: Should be resolved.

rhc54:
> @devreal Well, I admit it turns out to be more than theoretical. I did not discover this snippet with my testing when I first opened the PR (this is inside `orte/util/proc_info.c`); I didn't catch this when changing the `gethostname` calls to `opal_gethostname()`. I suppose I could fix this here instead, and remove the temporary from within `opal_gethostname.c`, but my thinking was, if it can happen here without triggering a compiler warning, it can happen elsewhere, so I chose to protect the static variable inside `opal_gethostname`. I'm open to doing whatever others judge to be best though.
> 
> (The lack of compiler warning is at least occurring on a Mac with clang, and it wouldn't surprise me if it isn't warned about by any compiler given the call to `strchr`.)
> 
> Note hostname declaration:
> 
> ```
> int orte_proc_info(void)
> {
> 
>     int idx, i;
>     char *ptr;
>     const char *hostname;
>     char **prefixes;
>     bool match;
> ```
> 
> Problematic section:
> 
> ```
>    /* get the nodename */
>     hostname = opal_gethostname();
>     /* add this to our list of aliases */
>     opal_argv_append_nosize(&orte_process_info.aliases, hostname);
> 
>     // Strip off the FQDN if present, ignore IP addresses
>     if( !orte_keep_fqdn_hostnames && !opal_net_isaddr(hostname) ) {
>         if (NULL != (ptr = strchr(hostname, '.'))) {
>             *ptr = '\0';
>             /* add this to our list of aliases */
>             opal_argv_append_nosize(&orte_process_info.aliases, hostname);
>         }
>     }
> ```

You hit a good point here - we actually want all parts of the code to use the same hostname, which we usually truncate to remove fqdn unless requested otherwise. Perhaps the best answer is to move the "keep fqdn" flag to the opal_hostname area and have everyone else strictly access the results?

cpshereda:
> You hit a good point here - we actually want all parts of the code to use the same hostname, which we usually truncate to remove fqdn unless requested otherwise. Perhaps the best answer is to move the "keep fqdn" flag to the opal_hostname area and have everyone else strictly access the results?

Given my unfamiliarity with this code I am likely not the right person to decide this - although I can understand the argument for enforcing strict access and correcting any hidden issues like in the above snippet. 

So you are saying this 'keep_fqdn' flag would then be passed as an argument to `opal_gethostname()`? Can you think of an example of a function that was originally calling `gethostname()` (now `opal_gethostname()` in this branch) that should instead be getting the truncated hostname?

Isn't orte_process_info.nodename already functioning as the global variable for the truncated hostname? Or is the answer yes, but it's not used everywhere it should be?

rhc54:
Might be easier if I simply provide a patch to implement the change. The global version of the hostname is actually supposed to reside in OPAL so it is visible to the OMPI area without requiring that ORTE be built/used. Let us talk about this on today's telecon and consider the best way forward.

cpshereda:
Sorry if I missed the meeting, I don't think I have current information about it.

hppritcha:
@rhc54 do you mean a patch to apply to the PR's branch or add to master after this one is merged?

rhc54:
@jsquyres and I discussed this yesterday and he was going to get back to @cpshereda, though perhaps that hasn't happened yet? Upshot is that things got somewhat off-track at some point. @cpshereda correctly noted that the code base should simply be referencing the global "nodename" value instead of everyone going and getting it themselves. It is precisely that uncoordinated retrieval that is causing some of the library confusion (is the host named ```foo```, or ```foo.com```?).

The global storage is actually in the OPAL layer in opal/util/proc.c. What we need to do is:

- move the "keep fqdn" MCA parameter down to the OPAL layer, registering an ORTE synonym that breaks abstraction but retains backward compatibility
- have OPAL continue to populate that global storage during opal_init, taking the MCA param into account
- change all the other places that call "gethostname" to simply reference opal_process_info.nodename

This would fix not only the problem cited in the associated issue, but the long-standing problems we have had with hostname confusion.

cpshereda:
> 
> 
> @jsquyres and I discussed this yesterday and he was going to get back to @cpshereda, though perhaps that hasn't happened yet? Upshot is that things got somewhat off-track at some point. @cpshereda correctly noted that the code base should simply be referencing the global "nodename" value instead of everyone going and getting it themselves. It is precisely that uncoordinated retrieval that is causing some of the library confusion (is the host named `foo`, or `foo.com`?).
> 
> The global storage is actually in the OPAL layer in opal/util/proc.c. What we need to do is:
> 
>     * move the "keep fqdn" MCA parameter down to the OPAL layer, registering an ORTE synonym that breaks abstraction but retains backward compatibility
> 
>     * have OPAL continue to populate that global storage during opal_init, taking the MCA param into account
> 
>     * change all the other places that call "gethostname" to simply reference opal_process_info.nodename
> 
> 
> This would fix not only the problem cited in the associated issue, but the long-standing problems we have had with hostname confusion.

Huh. Well, ok. I wish I'd have known/understood this sooner but I guess no one was thinking about it in relation to this issue? I suppose I could have figured this out on my own earlier if I had tried taking a more holistic view from the start. And I'm waiting to hear from Jeff but I'm guessing then that responsibility for at least some of these mods is going to transfer to @rhc54 or away from me? Not sure what exactly we're going to do with this PR.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+        /*
+         * Eventually, we would like this code to allow the full string
+         * to be copied in all (or all reasonable) cases (arbitrary length)
+         */
+        strncpy( opal_local_hostname, buf, OPAL_LOCAL_MAXHOSTNAMELEN - 1 );
+        opal_local_hostname[OPAL_LOCAL_MAXHOSTNAMELEN - 1] = '\0';
+        free( buf );
+        /* restore errno if there was a prior error to preserve error condition */
```
jsquyres: Per my other comment, I think that since `gethostname()` may write a new value to `errno`, there is no expectation that a prior value of `errno` will be preserved after calling `gethostname()`.  So it's probably ok to not save a prior value of `errno`.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,178 @@
...
+        /* if there was not an error, but the result is truncated, 
+         * we set errno appropriately */
+        else if( length > OPAL_LOCAL_MAXHOSTNAMELEN ) {
+            errno = ENAMETOOLONG;
+        }
+    }
+
+    /* Re-copy hostname into a temp variable because of risk
+       of caller modifying this variable. */
+    strcpy(temp_hostname, opal_local_hostname);
```
jsquyres: Fair point that some other caller may have modified the buffer.  But since we're returning a `const char *`, is that actually a possibility?

jsquyres:
> And I'm waiting to hear from Jeff but I'm guessing then that responsibility for at least some of these mods is going to transfer to @rhc54 or away from me? Not sure what exactly we're going to do with this PR.

Sorry for the delay -- I half-typed this up yesterday and then got distracted by other shiny objects.  @rhc54's comment this morning prompted me to a) do the review, and b) finish (i.e., re-write) my comments.

jsquyres:
I'm sorry -- I forgot to mention one thing before submitting my review: can you git rebase so that the IOF commit is not necessary to be on this PR?  Thanks.

hppritcha:
@cpshereda discussed. Plan now is to move the ```opal_gethostname``` content into ```opal/runtime/opal_init.c```, rename, and make it a static function just called with a single routine ```opal_init_util``` to get the nodename to assign to ```opal_process_info.nodename``` .  There will be a new ```opal_gethostname``` which, per the suggestion, will simply return a pointer to ```opal_process_info.nodename```.  Hopefully no one will truncate this!

orte/test/mpi/singleton_client_server.c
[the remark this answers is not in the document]
cpshereda: I understand, and you're right, but gethostname() also had a return code. Anyway, it was an effort to be super-careful.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,178 @@
...
+    int ret_val, num_tries = 0;
+    int temp_errno = 0;
+
+    opal_mutex_lock(&opal_gethost_mutex);
+
+    if( first_pass ) {
+        char *buf = malloc( length );
+        if( buf == NULL ) {
+            errno = ENOMEM;
+            return MALLOC_ERROR_STRING;
```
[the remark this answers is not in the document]
cpshereda: Yes, thanks for catching this.

opal/util/opal_gethostname.c
```diff
@@ -0,0 +1,178 @@
...
+            else {
+                temp_errno = errno;
+                strcpy( buf, UNKNOWN_ERROR_STRING );
+                break;
+            }
+            length *= 2;
+            buf = realloc( buf, length );
+            if( NULL == buf ) {
+               errno = ENOMEM;
+               return MALLOC_ERROR_STRING;
```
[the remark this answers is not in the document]
cpshereda: Yes, will be fixed.

opal/util/opal_gethostname.c
cpshereda: Yes, it can still happen under certain conditions - this is in the conversation about orte/util/proc_info.c here:
https://github.com/open-mpi/ompi/pull/7200#issuecomment-566186017

jsquyres: Ah, gotcha!

Given that you've now had to explain it twice 😄, that means it's probably worth describing in a comment so that our future selves don't have this conversation again...

ibm-ompi:

ibm-ompi:

cpshereda:
bot:ibm:retest

cpshereda:
> The IBM CI (XL Compiler) build failed! Please review the log, linked below.
> 
> Gist: https://gist.github.com/ibm-ompi/88b480fb98dfd2bdac939f023e1f8fe6

bot:ibm:retest

ibm-ompi:

ibm-ompi:

hppritcha:
@jjhursey or other IBMer take a look at the CI results?

awlauria:
bot:ibm:retest

awlauria:
Ibm CI issues seem to be resolved. Some configuration issues on the nodes.

hppritcha:
@jsquyres could you check this PR again?

orte/util/proc_info.c
```diff
@@ -164,19 +168,29 @@ int orte_proc_info(void)
...
+    hostname = opal_gethostname();
+    /* copy into a local version so that we can manipulate it */
+    hostname_len = strlen(hostname);
+    if( hostname_len > 0 ) {
+        local_hostname = malloc( strlen(hostname) );
+    }
+    if( NULL == local_hostname ) {
+       return OPAL_ERR_BUFFER; 
+    }
+    strcpy(local_hostname, hostname);
```
jsquyres: You could probably replace the above block with
```c
local_hostname = strdup(opal_gethostname());
if (NULL == local_hostname) {
    return OPAL_ERR_BUFFER;
}

File: orte/util/proc_info.c [hunk 16]
```diff
@@ -164,19 +168,29 @@ int orte_proc_info(void)
...
-        if (NULL != (ptr = strchr(hostname, '.'))) {
+    if( !orte_keep_fqdn_hostnames && !opal_net_isaddr(local_hostname) ) {
+        if (NULL != (ptr = strchr(local_hostname, '.'))) {
             *ptr = '\0';
             /* add this to our list of aliases */
-            opal_argv_append_nosize(&orte_process_info.aliases, hostname);
+            opal_argv_append_nosize(&orte_process_info.aliases, local_hostname);
         }
     }
+    printf( "local_hostname after strip = %s\n", local_hostname);
```
jsquyres: Not sure if this printf was intended to be here or not, but it probably shouldn't be in the final commit.

orte/util/proc_info.c
```diff
@@ -213,13 +227,14 @@ int orte_proc_info(void)
...
-            orte_process_info.nodename = strdup(hostname);
+            orte_process_info.nodename = strdup(local_hostname);
         }
         opal_argv_free(prefixes);
     } else {
-        orte_process_info.nodename = strdup(hostname);
+        orte_process_info.nodename = strdup(local_hostname);
     }
 
+    printf( "local_hostname after assignment and free = %s\n", local_hostname);
```
jsquyres: Not sure if this printf was intended to be here or not, but it probably shouldn't be in the final commit.

ompi/tools/mpisync/sync.c
```diff
@@ -77,7 +80,9 @@ int main(int argc, char **argv)
     int rank, commsize;
     double offs = 0, rtt = 0;
     char hname[OPAL_MAXHOSTNAMELEN];
-
+    const char *local_hname;
+    int temp_errno;
```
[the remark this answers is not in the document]
cpshereda: Should be fixed in the latest.

opal/runtime/opal.h
```diff
@@ -111,6 +112,25 @@ OPAL_DECLSPEC void opal_warn_fork(void);
...
+ */
+OPAL_DECLSPEC static inline const char *opal_gethostname( void ) {
+
+/*    if( NULL != opal_process_info.nodename ) { */
+        return opal_process_info.nodename;
+/*    }
+    else {
+        return ??????
+    }
+*/
```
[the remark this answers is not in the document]
cpshereda: Thanks, yes. Should be fixed in the latest push.

opal/runtime/opal_init.c
```diff
@@ -95,6 +95,159 @@ int opal_util_initialized = 0;
...
+#define MEMORY_ERROR_STRING "Invalid address (EFAULT) in call to opal_init_gethostname(). Check memory usage."
+#define HOSTNAMELEN_ERROR_STRING "Error in call to opal_init_gethostname() - check hostname length. Excessively long hostnames are unsupported."
+#define UNKNOWN_ERROR_STRING "Unknown error in opal_init_gethostname() function. Probable code issue."
+
+static char opal_local_hostname[OPAL_LOCAL_MAXHOSTNAMELEN] =
+    HOSTNAMELEN_ERROR_STRING;
+/*
+static char temp_hostname[OPAL_LOCAL_MAXHOSTNAMELEN] =
+    HOSTNAMELEN_ERROR_STRING;
+*/
```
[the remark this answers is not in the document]
cpshereda: oops. Removed.

opal/util/Makefile.am
```diff
@@ -17,6 +17,8 @@
 # Copyright (c) 2016      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 # Copyright (c) 2016-2017 IBM Corporation. All rights reserved.
+# Copyright (c) 2019      Triad National Security, LLC.
+#                         All rights reserved.
```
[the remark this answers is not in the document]
cpshereda: Yeah, was lingering from prior state in which opal_gethostname.c was in there. Should be fixed.

opal/runtime/opal_init.c
```diff
@@ -95,6 +95,159 @@ int opal_util_initialized = 0;
...
+ * ENAMETOOLONG.
+ */
+static const char *opal_init_gethostname( void )
+{
+    size_t count;
+    size_t length = OPAL_LOCAL_MAXHOSTNAMELEN;
+    int ret_val, num_tries = 0;
+    int temp_errno = 0;
+
+    opal_mutex_lock(&opal_gethost_mutex);
```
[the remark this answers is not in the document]
cpshereda: Ok, I wasn't sure. Should be fixed. 

@jsquyres - In addition, I modified `opal_init_gethostname()` to return an int now, so there are more modifications to go with that (I know, I know, but changing it to return a return code cleaned things up a bit).

opal/runtime/opal_init.c
```diff
@@ -95,6 +95,159 @@ int opal_util_initialized = 0;
...
+         * we set errno appropriately */
+        else if( length > OPAL_LOCAL_MAXHOSTNAMELEN ) {
+            errno = ENAMETOOLONG;
+        }
+    }
+
+    opal_mutex_unlock(&opal_gethost_mutex);
+
+    return opal_local_hostname;
+
```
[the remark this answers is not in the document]
cpshereda: Should be fixed.

opal/runtime/opal_init.c
```diff
@@ -390,7 +543,24 @@ opal_init_util(int* pargc, char*** pargv)
      * that we don't bother with fqdn and prefix issues here - we let
      * the RTE later replace this with a modified name if the user
      * requests it */
-    gethostname(hostname, sizeof(hostname));
+    errno = 0;
```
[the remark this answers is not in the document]
cpshereda: Because it's returning a return code now, this should now be moot.

orte/util/proc_info.c
cpshereda: Thanks, fixed.

orte/util/proc_info.c
cpshereda: oops. fixed.

orte/util/proc_info.c
cpshereda: also oops, should be fixed.

opal/runtime/opal.h
```diff
@@ -111,6 +114,19 @@ OPAL_DECLSPEC void opal_warn_fork(void);
...
 
+/**
+ * Wrapper to return the hostname value that is in
+   opal_process_info.nodename, as opposed to calling gethostname()
+   directly, which is not guaranteed to be null-terminated and
+   varies in its behavior depending on implementation. The 
+   opal_process_info.nodename value is first populated in 
+   opal/runtime/opal_init.c
+ */
+OPAL_DECLSPEC static inline const char *opal_gethostname( void ) {
```
jsquyres:
```suggestion
static inline const char *opal_gethostname( void ) {
```

Cannot have the `OPAL_DECLSPEC` there -- that's only for non-static symbols (for me, it results in a compiler warning every time the file is included).

jsquyres:
@cpshereda I like returning the `int` from `opal_init_gethostname()` a lot.  I took it a step further and returned `OPAL_ERR_*` codes rather than just 0 or -1.  That ended up simplifying the logic quite a bit.  What do you think of this?

https://gist.github.com/jsquyres/b5a84160cbca5bc127ffbdec46f9c761

hppritcha:
bot:ompi:retest

cpshereda:
> @cpshereda I like returning the `int` from `opal_init_gethostname()` a lot. I took it a step further and returned `OPAL_ERR_*` codes rather than just 0 or -1. That ended up simplifying the logic quite a bit. What do you think of this?
> 
> https://gist.github.com/jsquyres/b5a84160cbca5bc127ffbdec46f9c761

Well, using OPAL error codes sounds good, although that's a lot of changes. But yes, I'll change to this and push it.

jsquyres:
Other than the 2 additional commits that made it in to your last push, this looks good to me!

Want to squash down your commit message for the final commit a bit?  We probably don't need the history entries -- just a decent description of the final state.

Thank you!

cpshereda:
> Other than the 2 additional commits that made it in to your last push, this looks good to me!
> 
> Want to squash down your commit message for the final commit a bit? We probably don't need the history entries -- just a decent description of the final state.
> 
> Thank you!

You're welcome and should be fixed, thanks for all the checking.

opal/runtime/opal.h
jsquyres: I'm sorry, I think my final review comment was misleading / I didn't reference this minor issue.  We can't have this OPAL_DECLSPEC here.  I took the liberty of amending the commit to fix this issue.  We should be good to go now.

cpshereda:
> Other than the 2 additional commits that made it in to your last push, this looks good to me!
> 
> Want to squash down your commit message for the final commit a bit? We probably don't need the history entries -- just a decent description of the final state.
> 
> Thank you!

My mistake about the OPAL_DECLSPEC bug.

jsquyres:
bot:ibm:retest

jjhursey:
bot:ibm:gnu:retest

jjhursey:
bot:ibm:xl:retest

Issue #7203, opened 2019-11-27
MPI_Win_allocate_shared does not conform to the standard when size=0


nschaeff:
When N processes share memory on one node, and they allocate a shared memory window with
MPI_Win_allocate_shared(), if the local size on some process is zero but the total size of the window is not, the process with zero size obtains a NULL pointer in the return parameter baseptr.
This does not seem to be compatible with the MPI-3.1 standard, which requires (line 43 of page 407) "The allocated memory is contiguous across process ranks".
Certainly, the NULL pointer will not be contiguous to other ranks...

Workaround: get the actual pointer through a call to MPI_WIN_SHARED_QUERY() + some basic computations.

This has been found in OpenMPI v2.0.4. I know it is not supported anymore, but maybe this is also an issue in more recent OpenMPI versions (I can't test because I cannot install a newer version).

devreal:
I can reproduce this on `master`. The fix should be simple, will post a PR soon.

jsquyres:
Looks like this was fixed in master in #7204, and then back ported to v4.0.x, v3.1.x, and v3.0.x.

Closing this issue.

PR #7204, opened 2019-11-28
Correctly set baseptr in contiguous shared memory window with local size zero

devreal:
The MPI standard mandates that for contiguous shared memory windows the returned `baseptr` can be used to directly calculate the remote addresses on other processes so we cannot return `NULL` if `size == 0`.


devreal:
See #7203 for the issue

awlauria:
@hjelmn ok to merge?

awlauria:
Should this be cherry-picked to release branches?

PR #7207, opened 2019-11-29
Remove unused opal_shmem_seg_hdr_t to retain alignment

### LINKED ISSUE #6950: Cache alignment in osc/sm
## Background information
I have been working on a new osc/ofi component in an internship at HPE. To get started I tried to understand the existing osc/sm component. While jumping through the code I confirmed what @hkuno and @jlbyrne-hpe already told me. Everything in the sm component gets shifted over by 4 bytes. This means that nothing is cache-aligned, even though the code tries to make everything cache-aligned.
 
I only looked at the code in master
 
## Details of the problem
The problem is that each opal/shmem component introduces a header to the shared memory segment:
https://github.com/open-mpi/ompi/blob/c4d07520368e8aaeae5d15edb6f091df124f316b/opal/mca/shmem/shmem_types.h#L101-L106
 
This causes everything else in the segment to get shifted over by the size of the header:
https://github.com/open-mpi/ompi/blob/c4d07520368e8aaeae5d15edb6f091df124f316b/opal/mca/shmem/mmap/shmem_mmap_module.c#L544-L545
 
This causes two problems:
- The returned adress is no longer cache- and page- aligned.
- The allocated memory is larger than requested, potentially mapping a page more than intended.
 
The first problem is potentially an issue because e.g. osc/sm is not aware of how much the allocated memory shifted away from a cache line boundary:
https://github.com/open-mpi/ompi/blob/c4d07520368e8aaeae5d15edb6f091df124f316b/ompi/mca/osc/sm/osc_sm_component.c#L281-L284
https://github.com/open-mpi/ompi/blob/c4d07520368e8aaeae5d15edb6f091df124f316b/ompi/mca/osc/sm/osc_sm_component.c#L337-L349
 
--------------------
### Remove the segment header
I couldn't find any evidence the segment base header is used apart from the initialisation of the lock in it. So if I am not mistaken the header could easily be removed and all problems would be solved.
 
### Move the segment header to the end of the segment
This would solve the alignment issue but not the size issue.
 
### Pad the header to the size of a cache line
This would be the method that probably requires the least amount of code change. Similar to the previous solution, this would only solve one of the "problems" described.
 
### Add a method to the framework to get the header size
Now we get into more invasive idea but I want to name it anyway.
This would shift the issue over to the user of the opal/shmem components. It would allow the user to request exactly the amount of memory he wants, taking the header size into account. This especially means that if the user is padding to a page anyway he can take the header size into account and pad with less.
 
On the other hand this would be a lot of overhead on the user side and could lead to a potential misuse of the component if the user is not aware of the fact that his segment base will not be at a page boundary.

devreal:
As pointed out in #6950, the `opal_shmem_seg_hdr_t` struct allocated at the beginning of each shared memory segment is not actually used but interferes with the alignment of the memory. This PR removes this struct, allowing the `shmem` components to return page-aligned memory.

Fixes #6950

### DIFF
--- a/opal/mca/shmem/mmap/shmem_mmap_module.c
+++ b/opal/mca/shmem/mmap/shmem_mmap_module.c
@@ -133,7 +133,7 @@ shmem_ds_reset(opal_shmem_ds_t *ds_buf)
     ds_buf->seg_id = OPAL_SHMEM_DS_ID_INVALID;
     ds_buf->seg_size = 0;
     memset(ds_buf->seg_name, '\0', OPAL_PATH_MAX);
-    ds_buf->seg_base_addr = (unsigned char *)MAP_FAILED;
+    ds_buf->seg_base_addr = MAP_FAILED;
 }
 
 /* ////////////////////////////////////////////////////////////////////////// */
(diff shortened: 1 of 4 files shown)


devreal:
Should this be cherry-picked to the release branches?

PR #7208, opened 2019-11-29
rdma osc: register shared memory locally in each process

devreal:
This PR improves the handling of shared memory backing a window allocated through `MPI_Win_allocate` in case more than one process exists per node. Two points are important:

1) The state segment and each process' data segments are put into separate pages to ensure memory is page-aligned and pages are not shared between processes (similar to what the `sm` osc component does for noncontiguous shared memory windows).
2) Each process registers its data segment with the btl separately to avoid NUMA issues. Previously, rank 0 on the node registered all data segments in one go, likely pinning all pages to one NUMA domain.

It should be noted that this creates a discrepancy in terms of alignment between having a single process per node (uses `malloc`, so only minimal alignment guaranteed) and multiple processes per node (now each process' memory is page-aligned). We thought about replacing `malloc` with `posix_memalign` in the single-process case but we are not sure how to ensure portability (there is `opal/win32/win_compat.h` but that is not included anywhere and none of us works on win32 to test it). The MPI standard does not mandate page-alignment but the discrepancy might be something users notice. Any thoughts on whether or not to extend page-alignment to the single-process case are welcome :)

Page alignment in this PR really only works in combination with #7207 

Thanks to Hatem Elshazly (@hshazly) for working on this patch.

Signed-off-by: Hatem Elshazly <hatem.elshazly@bsc.com>

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -46,6 +46,7 @@
 #include "opal/util/argv.h"
 #include "opal/util/printf.h"
 #include "opal/align.h"
+#include "opal/util/sys_limits.h"
 #if OPAL_CUDA_SUPPORT
 #include "opal/datatype/opal_datatype_cuda.h"
 #endif /* OPAL_CUDA_SUPPORT */
@@ -574,9 +575,10 @@ static int allocate_state_shared (ompi_osc_rdma_module_t *module, void **base, s
     data_base = state_base + leader_peer_data_size + module->state_size * local_size;
 
     /* ensure proper alignment */
-    data_base += OPAL_ALIGN_PAD_AMOUNT(data_base, OPAL_ALIGN_MIN);
+    int page_size = opal_getpagesize();
+    data_base += OPAL_ALIGN_PAD_AMOUNT(data_base, page_size);
     if (MPI_WIN_FLAVOR_ALLOCATE == module->flavor) {
-        size += OPAL_ALIGN_PAD_AMOUNT(size, OPAL_ALIGN_MIN);
+        size += OPAL_ALIGN_PAD_AMOUNT(size, page_size);
     }
 
     do {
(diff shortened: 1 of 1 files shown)


bosilca:
go for posix_memalign, win32 support is busted and I don't think there is anybody interested in reviving it in the short term.

hjelmn:
This will greatly increase the memory requirements for memory windows. On Cray systems the registration data is 16-bytes so by registering once per node we save 16*nprocs memory. This is not a small amount. There is a reason I did it this way.

hjelmn:
I will have to ask but I think if each process touches each page in their allocation the registration won't move the pages (I could be wrong). Have you tried that?

hjelmn:
Also, if that doesn't help then the registration should be per-NUMA not per-process.

awlauria:
@devreal have you had a chance to look at @hjelmn 's questions/comments?

devreal:
Unfortunately not yet, I'm about to finish my current project this week and will have some more time after that. Sorry about the delay.

awlauria:
@devreal no worries, there's no rush. Thanks, I am just going through open pr's for 5.0.

awlauria:
@devreal ping. I can also add this one to the 5.0 tracking if you wish. If you can get this in by April 30th that would be great. Thanks!

hjelmn:
@devreal Ping. I would like to get this issue fixed in 5.0.0. Have you tried touching the pages that will be owned by each process before the registration step? Would be good to know if that resolves the issue.

devreal:
@hjelmn Sorry, I completely lost track of this issue. I realized we already added code touching the pages (https://github.com/open-mpi/ompi/blob/master/ompi/mca/osc/rdma/osc_rdma_component.c#L652) before this PR came up. I believe it did fix the problem. However, for allocations smaller than page size, pages will still be shared in the current implementation in `master`. Maybe we should page-align each process' memory?

The idea for this PR was that we wouldn't have to write to all the pages but instead rely on the pinning to allocate them. I see your point about the memory requirement though.

hjelmn:
I think it would be reasonable to force the size to be a multiple of the page size in all cases.

PR #7211, opened 2019-12-02
Fix a few compiler warnings

mcoil1:
See individual commit messages.

### DIFF
--- a/ompi/dpm/dpm.c
+++ b/ompi/dpm/dpm.c
@@ -407,7 +407,7 @@ int ompi_dpm_connect_accept(ompi_communicator_t *comm, int root,
     }
     if (0 < opal_list_get_size(&ilist)) {
         uint32_t *peer_ranks = NULL;
-        int prn, nprn;
+        int prn, nprn = 0;
         char *val, *mycpuset;
         uint16_t u16;
         opal_process_name_t wildcard_rank;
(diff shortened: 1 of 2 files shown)


jsquyres:
ok to test

hppritcha:
bot:ompi:retest

opal/mca/mpool/base/base.h
```diff
@@ -53,7 +53,7 @@ OPAL_DECLSPEC OBJ_CLASS_DECLARATION(mca_mpool_base_selected_module_t);
 OPAL_DECLSPEC mca_mpool_base_component_t* mca_mpool_base_component_lookup(const char* name);
 OPAL_DECLSPEC mca_mpool_base_module_t* mca_mpool_base_module_lookup(const char* name);
 
-OPAL_DECLSPEC mca_mpool_base_module_t *mca_mpool_basic_create (void *base, size_t size, unsigned min_align);
+OPAL_DECLSPEC mca_mpool_base_module_t *mca_mpool_basic_create (mca_mpool_base_module_address_fn_t base, size_t size, unsigned min_align);
```
jsquyres: @bosilca @hjelmn Is there a reason this argument was `(void*)` instead of `mca_mpool_base_module_address_fn_t`)?

jsquyres: @bosilca @hjelmn ping 😄

bosilca: Let me first state that the current implementation of mca_mpool_basic_create() is incorrect because it uses the base argument both as an ```mpool_base``` (which is a function of type mca_mpool_base_module_address_fn_t) and as ```ptr``` (which is a void*).

I think the correct approach would be leave the argument of ```mca_mpool_basic_create``` as a void, but to change the implementation to store the ```base``` only as ```ptr```, and to have a local static function in ```mpool_base``` (similar to what we do to the other base members in mca_mpool_basic_template [opal/mca/mpool/base/mpool_base_basic.c:76]).

jsquyres: We'll snip the mpool "fix" off this PR -- I just created https://github.com/open-mpi/ompi/issues/7221 to track the issue separately.

jsquyres:
False AWS jenkins failure: bot:ompi:retest

PR #7215, opened 2019-12-03
v4.0.x, oshmem:ucx, fix race condition and add context recycling

janjust:
1) Race condition: Do not add private contexts to active list.
Private contexts are only visible to the user.
2) Recycled contexts: Destroyed contexts are put on an idle list until
finalize, continuous context creation will lead to oom condition.
Instead, check if context from idle list meets new context requirements
and reuse it.

Co-authored with: Artem Y. Polyakov <artemp@mellanox.com>,
                  Manjunath Gorentla Venkata <manjunath@mellanox.com>


janjust:
@gpaulsen done

hppritcha:
bot:ompi:retest

janjust:
@artpol84 , please approve if no objections

PR #7218, opened 2019-12-04
opal/util/error.c: fixed unused result message

wbailey2:
Squashed compiler warning.
Ignoring the result is fine here because it is outputing an error so it doesn't matter if the write function fails.

Signed-off-by: William Bailey <wbailey2@nd.edu>

### DIFF
--- a/opal/util/error.c
+++ b/opal/util/error.c
@@ -234,7 +234,10 @@ opal_delay_abort(void)
                      "[%s:%05d] Looping forever "
                      "(MCA parameter opal_abort_delay is < 0)\n",
                      opal_process_info.nodename, (int) pid);
-            write(STDERR_FILENO, msg, strlen(msg));
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wunused-result"
+            write(STDERR_FILENO, msg, strlen(msg)); /* this specific unused result warning gets ignored during compilation */
+        #pragma GCC diagnostic pop    
             while (1) {
                 sleep(5);
             }


jsquyres:
ok to test

rhc54:
What happens when exposed to a non-gcc compiler?

jsquyres:
Open MPI community: this is an invitation for comment.  We have a bunch of places in the code base where we ignore the return codes from various functions.  Recent versions of the gcc compiler don't allow us to just `(void)` cast away the return code -- the compiler still emits warnings.

This pragma option turned up as a possibility.  As Ralph noted, this may or may not work on non-GCC compilers (i.e., we may have to surround these pragma's with #if blocks for GCC, or somesuch).

What are people's thoughts here?

wbailey2:
Accidentally changed the spacing on a line and went back to fix it.

jsquyres:
FWIW, OMPI built just fine with the Intel 2019 compilers and these `#pragma GCC` statements.  But because of #7223, it was very difficult to tell if new warnings were emitted due to the `#pragma GCC` statements.  I don't _think_ so, but I couldn't tell for sure.

awlauria:
I think we *should* be checking the return codes of functions such as write()/read() ect as a general practice. Granted in places such as this it won't matter, but if we do this as a general practice in all areas it will save a headache when it should be checked and it causes a subtle bug. At the very least it can be logged in the component verbose output.

jsquyres:
I don't think that this one is going to be finished.

PR #7219, opened 2019-12-04
romio: Update ADIOI_R_Exchange_data function

mcoil1:
Squash compiler warning due to whitespace problems.
I updated the function to match the current version from MPICH.

### DIFF
--- a/ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
+++ b/ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
@@ -826,17 +826,17 @@ static void ADIOI_R_Exchange_data(ADIO_File fd, void *buf, ADIOI_Flatlist_node
 	    if (recv_size[i]) recv_buf[i] = 
                                   (char *) ADIOI_Malloc(recv_size[i]);
 
-	    j = 0;
-	    for (i=0; i < nprocs; i++) 
-		if (recv_size[i]) {
-		    MPI_Irecv(recv_buf[i], recv_size[i], MPI_BYTE, i, 
-			      myrank+i+100*iter, fd->comm, requests+j);
-		    j++;
+        j = 0;
+        for (i=0; i < nprocs; i++)
+            if (recv_size[i]) {
+                MPI_Irecv(recv_buf[i], recv_size[i], MPI_BYTE, i,
+                          myrank+i+100*iter, fd->comm, requests+j);
+                j++;
 #ifdef RDCOLL_DEBUG
-		    DBG_FPRINTF(stderr, "node %d, recv_size %d, tag %d \n", 
-		       myrank, recv_size[i], myrank+i+100*iter); 
+                DBG_FPRINTF(stderr, "node %d, recv_size %d, tag %d \n",
+                            myrank, recv_size[i], myrank+i+100*iter);
 #endif
-		}
+            }
     }
 
 /* create derived datatypes and send data */


ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
```diff
@@ -811,32 +810,33 @@ static void ADIOI_R_Exchange_data(ADIO_File fd, void *buf, ADIOI_Flatlist_node
...
-    }
-    else {
+        }
+    } else {
 /* allocate memory for recv_buf and post receives */
 	recv_buf = (char **) ADIOI_Malloc(nprocs * sizeof(char*));
-	for (i=0; i < nprocs; i++) 
-	    if (recv_size[i]) recv_buf[i] = 
-                                  (char *) ADIOI_Malloc(recv_size[i]);
+	recv_buf[0] = (char *) ADIOI_Malloc(memLen);
```
jsquyres: It looks like `memLen` will always be 0 when it gets to this point...?

ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
```diff
@@ -864,9 +864,13 @@ static void ADIOI_R_Exchange_data(ADIO_File fd, void *buf, ADIOI_Flatlist_node
...
 
-    statuses = (MPI_Status *) ADIOI_Malloc((nprocs_send+nprocs_recv+1) * \
+#ifdef MPI_STATUSES_IGNORE
+    statuses = MPI_STATUSES_IGNORE;
+#else
+    statuses = (MPI_Status *) ADIOI_Malloc((nprocs_send+nprocs_recv+1) *
                                      sizeof(MPI_Status)); 
-     /* +1 to avoid a 0-size malloc */
+    /* +1 to avoid a 0-size malloc */
+#endif
```
jsquyres: This new stuff about MPI_STATUSES_IGNORE isn't really necessary.

Indeed, Open MPI has MPI_STATUSES_IGNORE, but it's not a #define, so it won't even pass the #ifdef, above.

jsquyres:
ok to test

ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
mcoil1: Oops, looks like I missed a line somewhere. Do you want me to add that, or should I revert this and just keep the `{}` fixes?

jsquyres: Let's do just the `{}` fixes.

ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
```diff
@@ -811,32 +809,34 @@ static void ADIOI_R_Exchange_data(ADIO_File fd, void *buf, ADIOI_Flatlist_node
...
 		buf_idx[i] += recv_size[i];
 	    }
-    }
-    else {
+        }
+    } else {
 /* allocate memory for recv_buf and post receives */
 	recv_buf = (char **) ADIOI_Malloc(nprocs * sizeof(char*));
-	for (i=0; i < nprocs; i++) 
+	for (i=0; i < nprocs; i++) {
```
jsquyres: I think that the only _required_ pair of `{}` that is missing is on this line and between old lines 839 and 840.  I don't think we want to outdent the `j=0; for (i=0; i < nprocs; i++)...` chunk of text -- I think that's all part of the same block that should be protected by the missing `{}`.

ompi/mca/io/romio321/romio/adio/common/ad_read_coll.c
```diff
@@ -811,32 +809,34 @@ static void ADIOI_R_Exchange_data(ADIO_File fd, void *buf, ADIOI_Flatlist_node
 
     if (buftype_is_contig) {
 	j = 0;
-	for (i=0; i < nprocs; i++) 
+	for (i=0; i < nprocs; i++) {
```
jsquyres: By the Open MPI coding standards, this pair of `{}` is technically missing -- but it's not actually _needed_ here (because the body of the `for` loop is technically only 1 statement (the `if` block)).

Let's go for the minimum-distance fix here, though -- let's not add by-the-OMPI-style-it's-missing-but-it's-still-technically-correct pairs of `{}` like this one.

PR #7226, opened 2019-12-08
memory/patcher: fix compiler warning

mcoil1:
syscall() returns a long, but we are invoking shmat(), which returns
a void*.


jsquyres:
ok to test

PR #7228, opened 2019-12-10
Harmonize return values of progress callbacks

devreal:
The logic in `opal_progress` expects a callback to return "the number of events progressed" and based on that determines whether or not to yield the CPU. Most components track the number of completed operations for that matter. However, some return a fixed value instead. The problem is that if a callback always returns a positive value (`psm2` mtl, `pt2pt` osc) the process will never yield if these components are active. Similarly, if the callback always returns `OMPI_SUCCESS` (or zero; `libnbc`, `ucx` pml, `uct` btl, `grequestx`), the process will yield unconditionally even if it wasn't necessary (plus, there is a mismatch in semantics in the first place). This PR tries to adapt the components in question to try to track whether progress has been made and signal it accordingly.

These changes should have no impact unless `yield_when_idle` is enabled.

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -514,8 +514,7 @@ int mca_pml_ucx_enable(bool enable)
 
 int mca_pml_ucx_progress(void)
 {
-    ucp_worker_progress(ompi_pml_ucx.ucp_worker);
-    return OMPI_SUCCESS;
+    return ucp_worker_progress(ompi_pml_ucx.ucp_worker);
 }
 
 int mca_pml_ucx_add_comm(struct ompi_communicator_t* comm)
--- a/ompi/communicator/comm_request.c
+++ b/ompi/communicator/comm_request.c
@@ -100,6 +100,7 @@ static int ompi_comm_request_progress (void)
 {
     ompi_comm_request_t *request, *next;
     static opal_atomic_int32_t progressing = 0;
+    int completed = 0;
 
     /* don't allow re-entry */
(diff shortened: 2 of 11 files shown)


ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -514,8 +514,9 @@ int mca_pml_ucx_enable(bool enable)
 
 int mca_pml_ucx_progress(void)
 {
-    ucp_worker_progress(ompi_pml_ucx.ucp_worker);
-    return OMPI_SUCCESS;
+    unsigned ret;
+    ret = ucp_worker_progress(ompi_pml_ucx.ucp_worker);
+    return (ret > 0) ? 1 : 0;
```
yosefe: maybe just return whatever  ucp_worker_progress() returns?

devreal: That was my first approach, too. The documentation for `ucp_worker_progress` states as return value 

> Non-zero if any communication was progressed, zero otherwise. 

Since it returns an `unsigned` the value could be `(unsigned)-1`, which cast to `int` would be seen as signalling an error by the progress loop.

yosefe: @devreal sorry if the documentation of ucp_worker_progress was not clear, but it can't really return -1. It's either 0 or a small integer value which indicated the number of processed events.

devreal: In that case we can return directly, yes :)

hppritcha:
bot:ompi:retest

devreal:
I fixed a conflict in the ucx smpl. Can someone review and merge please?

PR #7229, opened 2019-12-11
v4.0.x: portals4: fix flow control bugs

tkordenbrock:
This PR fixes two flow control related bugs:
* a configure bug that caused btl-portals4 flow control to be disabled regardless of the configure options used.
* resolves an mtl-portals4 segfault by skipping the flow control fini() call if Portals4 was not initialized.


hppritcha:
@tkordenbrock were these commits cherry-picked from master?  If so, please use cherry-pick -x when cherry-picking over to release branches.

tkordenbrock:
@hppritcha Yes they were.  I just pushed new commits to all three PRs.

Issue #7212, opened 2019-12-03
openmpi 3.1.5 references private glibc symbol __mmap


opoplawski:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
3.1.5

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
This is the Fedora 30 openmpi package

## Details of the problem

I'm trying to update the Fedora 30 package from openmpi 3.1.4 to 3.1.5, but it cannot be installed:
```
Error: transaction check vs depsolve:
libc.so.6(GLIBC_PRIVATE)(64bit) is needed by openmpi-3.1.5-1.fc30.x86_64
```
This appears to be referenced by:
```
$ readelf -a usr/lib64/openmpi/lib/libopen-pal.so.40.10.5

Relocation section '.rela.plt' at offset 0x1bab0 contains 607 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
00000009f108  001400000007 R_X86_64_JUMP_SLO 0000000000000000 __mmap@GLIBC_PRIVATE + 0
```
This appears to have be caused by commit 8be0a233c87dc85871acb49a8e9c7b5c6381fbc3:
```
Author: Mark Allen <markalle@us.ibm.com>
Date:   Wed Mar 27 14:12:20 2019 -0400

    shmat/shmdt additions for patcher

    This is mostly based off recent UCX additions to their patcher:
        https://github.com/openucx/ucx/pull/2703

    They added triggers for
    * mmap when (flags & MAP_FIXED) && (addr != NULL)
    * shmat when (shmflg & SHM_REMAP) && (shmaddr != NULL)

    Beyond that I noticed they already had a trigger for
    * madvise when (advice == MADV_FREE)
    that we didn't so I added that.

    And the other main thing is we didn't really have shmat/shmdt
    active for some systems because we only had a path for
    syscall(SYS_shmdt, ) but we needed to also have a path for
    syscall(SYS_ipc, IPCOP_shmdt, ) and same for shmat.

    Signed-off-by: Mark Allen <markalle@us.ibm.com>
    (cherry picked from commit eb888118e83f56c131aff900b03eab34c92b7805)
```
which changed:
```
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
[...]

PR #7241, opened 2019-12-17
v3.1.x: Fix mmap infinite recurse in memory patcher

mwheinz:
Backports Harumi Kuno's patch from July 30th to 3.1.x. This corrects issue #7212.

### DIFF
--- a/opal/mca/memory/patcher/configure.m4
+++ b/opal/mca/memory/patcher/configure.m4
@@ -40,10 +40,6 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
 
     AC_CHECK_HEADERS([linux/mman.h sys/syscall.h])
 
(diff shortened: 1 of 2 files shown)


jsquyres:
Sweet.  Thanks!

Did you confirm that this fixes #7212?  (it _looks_ like this is the right fix, but my own quick testing was not able to *confirm* that this was the correct fix)

Can you also cherry pick this back to the v3.0.x branch, too?

mwheinz:
Yes, I have confirmed that this patch allows OMPI 3.1.x to compile on RHEL 8.0 and RHEL 7.6, and that the resulting code can compile and run the OSU micro benchmarks.

I'll look at backporting to 3.0.x now.

hkuno:
Thank you, Michael, for testing and validating this!

PR #7242, opened 2019-12-17
v3.0.x: Fix mmap infinite recurse in memory patcher

mwheinz:
Backports Harumi Kuno's patch from July 30th to 3.0.x. This corrects issue #7212.

This commit fixes issue #6853 by removing
MacOS/Darwin-specific logic from intercept_mmap.

### DIFF
--- a/opal/mca/memory/patcher/configure.m4
+++ b/opal/mca/memory/patcher/configure.m4
@@ -40,10 +40,6 @@ AC_DEFUN([MCA_opal_memory_patcher_CONFIG],[
 
     AC_CHECK_HEADERS([linux/mman.h sys/syscall.h])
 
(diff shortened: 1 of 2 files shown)


mwheinz:
Confirmed that with this patch, OMPI compiles correctly in RHEL 8.0 and RHEL 7.6 and that the resulting installation can compile and execute the OSU micro benchmarks over the PSM2 MTL. I have not exhaustively tested other transports, but that should not be relevant to this patch.

hkuno:
Thank you, Michael, for testing and validating this! I had not merged this patch into 3 because I didn't want to do that unless I could build and demonstrate that the 3.0.x branch would hit the bug without the fix, and I had trouble demonstrating that.

Issue #6501, opened 2019-03-19
Move --host ordering fix to v3.0.x, v3.1.x, v4.0.x?


jsquyres:
Per #6298, we had an accidental change in behavior of `mpirun --host aaa,bbb` between version v2.1.x and v3.0.x.  A fix just went in to master in #6493.

Here's what happened:

* v2.0.x: behavior X
* v2.1.x: behavior X
* v3.0.x: switch to behavior Y
* v3.1.x: behavior Y
* v4.0.x: behavior Y
* master (to become v5.0.x): after PR #6493, back to behavior X

The question is: should we put this fix on any of v3.0.x, v3.1.x, and/or v4.0.x?

### Behavior X

The ordering of hosts in the `--host` list matters:

```
$ mpirun --host aaa,bbb rank_test
aaa: MCW rank 0
bbb: MCW rank 1
$ mpirun --host bbb,aaa rank_test
aaa: MCW rank 1
bbb: MCW rank 0
```

### Behavior Y

The ordering of hosts in the `--host` list does not matter (**note:** this behavior was unintentional.  It was always intended that we honor the ordering of hosts in the `--host` list):

```
$ mpirun --host aaa,bbb rank_test
aaa: MCW rank 0
bbb: MCW rank 1
$ mpirun --host bbb,aaa rank_test
aaa: MCW rank 0
bbb: MCW rank 1
```

## Discussion points

We need to discuss this and decide what to do.  Points (in no particular order):

1. This is a fairly minor change in behavior.
1. Apparently no one noticed this change in behavior between v2.1.x and v3.0.x.  It was only discovered recently by @bturrubiates, a Cisco employee (while using Open MPI for other / unrelated testing).
1. The fix is probably not worth putting into v3.0.x or v3.1.x. 
1. But it *might* be worthwhile to put in to v4.0.x...?
1. That being said, even putting it in v4.0.x is at least *sorta* breaking backwards compatibility.  You could squint at this and call it a bug and therefore allow it in.  Or you could say that it was effectively the behavior of all the v3.x/v4.x releases, and they're backwards compatible with each other, so we should maintain that behavior in v4.0.x.

jjhursey:
I think we noticed in PR #4327, but never got the cycles to get back to it. I'd like to verify this change also fixes that issue (I think it should)

jsquyres:
Two good points were made on the webex:

[...]

PR #7252, opened 2019-12-19
hwloc: clarify --with-hwloc behavior

jsquyres:
Clarify in README what --with-hwloc does in its different use cases.

Also, ensure that the behavior when specifying `--with-hwloc` is the
same as if that option is not specified at all.  This is what we did
in Open MPI <= v3.x; looks like we inadvertantly caused `--with-hwloc`
to be synonymous with `--with-hwloc=external` in v4.0.0.

Thanks to @marmistrz on #6501 for the heads-up on this issue.

### DIFF
--- a/README
+++ b/README
@@ -1252,28 +1252,26 @@ MISCELLANEOUS SUPPORT LIBRARIES
   this option is only needed for special configurations.
 
 --with-hwloc(=value)
-  Build hwloc support (default: enabled).  This option specifies where
+  hwloc is a support library that provides processor and memory
+  affinity information for NUMA platforms.  It is required by Open
+  MPI.  Therefore, specifying --with-hwloc=no (or --without-hwloc) is
+  disallowed.
+
+  By default (i.e., if --with-hwloc is not specified, or if
+  --with-hwloc is specified without a value), Open MPI will first try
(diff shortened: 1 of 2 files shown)


jsquyres:
@rhc54 @gpaulsen @hppritcha You're all tagged on this PR for a review because it looks like we inadvertently caused `--with-hwloc` (with no value specified) to be equivalent to `--with-hwloc=external` in v4.0.0.

If we cherry-pick this back to v4.0.x, it means that `--with-hwloc` is equivalent to not specifying the option at all (i.e., first try external, and if that fails, fall back to internal).

I think that this is the behavior that we want.

The only question is if this breaks any installation scripts for >=v4.0.x -- i.e., are we breaking backwards compat.  I don't think we are: if you specify `--with-hwloc` in >=v4.0.x, it currently fails if external can't be used.  This change (fix!) relaxes that a bit: if external can't be used, it falls back to internal.  It falls more in the spirit of "yes, do whatever you need to do to use hwloc" -- which IMNSHO is basically what `--with-hwloc[=yes]` should mean.

Thoughts?

hppritcha:
I think this is a safe change.

gpaulsen:
I agree, I doubt this will break any scripts, is what we intended.

PR #7257, opened 2019-12-20
mca/plm/rsh: close only opened descriptors using /proc/self/fd

trnila:
It takes a long time to call 'close' on all possible descriptor numbers
when _SC_OPEN_MAX is set to a high value.

Try to close only opened descriptors with the information provided from
procfs. If it fails, fallback to closing all possible descriptors.

### DIFF
--- a/orte/mca/plm/rsh/plm_rsh_module.c
+++ b/orte/mca/plm/rsh/plm_rsh_module.c
@@ -34,6 +34,7 @@
 #include "orte/constants.h"
 
 #include <stdlib.h>
+#include <ctype.h>
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
@@ -51,6 +52,9 @@
 #include <time.h>
 #ifdef HAVE_SYS_TYPES_H
 #include <sys/types.h>
+#endif
+#ifdef HAVE_DIRENT_H
+#include <dirent.h>
 #endif
 #ifdef HAVE_SYS_STAT_H
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


jsquyres:
ok to test

orte/mca/plm/rsh/plm_rsh_module.c
```diff
@@ -709,6 +713,38 @@ static int setup_launch(int *argcptr, char ***argvptr,
     return ORTE_SUCCESS;
 }
 
+/* close all open file descriptors w/ exception of stdin/stdout/stderr */
+static int close_open_file_descriptors(void) {
+    DIR *dir = opendir("/proc/self/fd");
```
jsquyres: This won't work on non-Linux systems (e.g., MacOS).

orte/mca/plm/rsh/plm_rsh_module.c
```diff
@@ -709,6 +713,38 @@ static int setup_launch(int *argcptr, char ***argvptr,
...
+    if (NULL == dir) {
+        return ORTE_ERR_FILE_OPEN_FAILURE;
+    }
+    struct dirent *files;
+
+    /* grab the fd of the opendir above so we don't close in the 
+     * middle of the scan. */
+    int dir_scan_fd = dirfd(dir);
```
jsquyres: I'm not sure `dirfd()` exists outside of Linux, either...?  It doesn't appear to be on MacOS, at least.

orte/mca/plm/rsh/plm_rsh_module.c
```diff
@@ -743,8 +779,10 @@ static void ssh_child(int argc, char **argv)
     close(fdin);
 
     /* close all file descriptors w/ exception of stdin/stdout/stderr */
-    for(fd=3; fd<fdmax; fd++)
-        close(fd);
+    if(close_open_file_descriptors() != ORTE_SUCCESS) {
+        for(fd=3; fd<fdmax; fd++)
```
jsquyres: Minor style thing while you're editing this chunk of code: can you put the `close(fd)` in `{}`, too?  We typically enclose _all_ blocks in `{}`, even 1-line blocks.  Not sure how this one escaped... 😄

rhc54:
Might be worth comparing this to what was done in the olds/default component - that also reduced the fd closures and the algo used there works on non-Linux systems.

orte/mca/plm/rsh/plm_rsh_module.c
trnila: Is it safe to wrap function into `#ifdef __linux__`? Or should we check existence of `opendir` and `dirfd`?

trnila:
I did not find any code in openmpi that can efficiently close open descriptors on multiple systems.
Same linux-only procedure is implemented in these files:
[orte/mca/odls/default/odls_default_module.c](https://github.com/trnila/ompi/blob/1b76b482d76371cdce8f3899587b4a49d7c058f9/orte/mca/odls/default/odls_default_module.c#L298)
[orte/mca/odls/alps/odls_alps_module.c](https://github.com/trnila/ompi/blob/1b76b482d76371cdce8f3899587b4a49d7c058f9/orte/mca/odls/alps/odls_alps_module.c#L294)
[orte/mca/odls/pspawn/odls_pspawn.c](https://github.com/trnila/ompi/blob/1b76b482d76371cdce8f3899587b4a49d7c058f9/orte/mca/odls/pspawn/odls_pspawn.c#L227)
[opal/mca/pmix/pmix4x/openpmix/src/mca/pfexec/linux/pfexec_linux.c](https://github.com/trnila/ompi/blob/1b76b482d76371cdce8f3899587b4a49d7c058f9/opal/mca/pmix/pmix4x/openpmix/src/mca/pfexec/linux/pfexec_linux.c#L308)

ibm-ompi:

ibm-ompi:

rhc54:
@trnila You can use whatever method was employed in odls_default_module.c as that is used across all systems we support, so we know that method works. Since we use it in multiple places, probably should be put in orte/util and just have everyone call it.

orte/mca/plm/rsh/plm_rsh_module.c
rhc54: @jsquyres This is the same code used in the odls today, so I suspect we have a similar problem there. The issue here isn't a missing ```opendir```, but rather that the ```/proc``` directories aren't present. What we should do is fall back to the "slow" way of closing things instead of returning an error.

PR #7267, opened 2020-01-02
v3.1.x: Fix incorrect behavior with length == 0

### LINKED ISSUE #6575: Possibly, incorrect MPI_Type_indexed behavior with zero blockLenght

bosilca:
Cherry-pick f68b06e9 on the 3.1.
Fixes #6575.

### DIFF
--- a/ompi/datatype/ompi_datatype_create_contiguous.c
+++ b/ompi/datatype/ompi_datatype_create_contiguous.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2004-2006 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2013 The University of Tennessee and The University
+ * Copyright (c) 2004-2019 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2006 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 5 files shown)


hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

USRC is having a network outage.

jsquyres:
@ggouaillardet Can you please review?  This PR was just brought to my attention (it wasn't labeled / milestoned before moments ago)

bwbarrett:
@ggouaillardet can you take a look?  we'd like to get going with a 3.1 update.

jsquyres:
We need a review for this.  We're looking to make a 3.1RC1 in the immediate future.  Can someone review?

jsquyres:
Unless we get an immanent review for this, it won't be included in v3.1.6.

PR #7271, opened 2020-01-05
Fix typo and update URLs (https, redirection) [skip ci]

e-kwsm:


rhc54:
Please remove the changes made to the ORTE area as that area is about to be replaced.

bwbarrett:
ok to test

e-kwsm:
> Please remove the changes made to the ORTE area as that area is about to be replaced.

@rhc54 Excuse me, what do you mean—revert changes to the files under `orte/`?

rhc54:
I mean just remove any changes to files under ```orte/```. There is another PR pending that will remove that entire directory tree, so (a) there is no point in changing those files, and (b) this will create an unnecessary conflict with the other PR (or vice versa, depending on which one gets committed first)

bwbarrett:
He means to edit your patch in the e-kwsm:update-man branch to not include any changes in the orte/ directory.  You'll have to modify the commit and then force-push into your update-man branch.

PR #7273, opened 2020-01-06
oshmem/ucx: Fix progress in iput/iget: periodically poke progress to prevent hardware stalls when using DCT transport.

janjust:
Fixes a progress issue when excess iput/iget operations are issued when using DC, by periodically calling progress on ucx workers.

Co-authored with:
Artem Y. Polyakov <artemp@mellanox.com>,
Manjunath Gorentla Venkata <manjunath@mellanox.com>

Signed-off-by:
Tomislav Janjusic <tomislavj@mellanox.com>

### DIFF
--- a/oshmem/mca/spml/ucx/spml_ucx.c
+++ b/oshmem/mca/spml/ucx/spml_ucx.c
@@ -774,6 +774,30 @@ int mca_spml_ucx_get_nb(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_
     return ucx_status_to_oshmem_nb(status);
 }
 
+int mca_spml_ucx_get_nb_wprogress(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_addr, int src, void **handle)
+{
+    unsigned int i;
+    void *rva;
+    ucs_status_t status;
+    spml_ucx_mkey_t *ucx_mkey;
+    mca_spml_ucx_ctx_t *ucx_ctx = (mca_spml_ucx_ctx_t *)ctx;
+
+    ucx_mkey = mca_spml_ucx_get_mkey(ctx, src, src_addr, &rva, &mca_spml_ucx);
+    status = ucp_get_nbi(ucx_ctx->ucp_peers[src].ucp_conn, dst_addr, size,
+                     (uint64_t)rva, ucx_mkey->rkey);
+
+    if (++ucx_ctx->nb_progress_cnt > mca_spml_ucx.nb_get_progress_thresh) {
+        for (i = 0; i < mca_spml_ucx.nb_ucp_worker_progress; i++) {
+            if (!ucp_worker_progress(ucx_ctx->ucp_worker)) {
+                ucx_ctx->nb_progress_cnt = 0;
+                break;
+            }
+        }
+    }
+
+    return ucx_status_to_oshmem_nb(status);
+}
+
 int mca_spml_ucx_put(shmem_ctx_t ctx, void* dst_addr, size_t size, void* src_addr, int dst)
 {
     void *rva;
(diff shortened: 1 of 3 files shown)


jladd-mlnx:
@janjust @artpol84  - when do you expect this to be ready?

janjust:
bot:retest

artpol84:
@yosefe @jladd-mlnx, please review

oshmem/mca/spml/ucx/spml_ucx_component.c
```diff
@@ -132,6 +146,22 @@ static int mca_spml_ucx_component_register(void)
                                      "Use synchronized quiet on shmem_quiet or shmem_barrier_all operations",
                                      &mca_spml_ucx.synchronized_quiet);
 
+    mca_spml_ucx_param_register_ulong("nb_progress_thresh_global", 0,
+                                    "Number of nb_put or nb_get operations before ucx progress is triggered. Disabled by default (0)",
+                                    &mca_spml_ucx.nb_progress_thresh_global);
+
```
jladd-mlnx: Shouldn't the default value be non-zero? Otherwise, progress is called on every iput/iget. Right?

janjust: No this is the global value. nb_put_thresh and nb_get_thresh inherit this value during init.
If user sets global, then both will be set otherwise if users sets individual value then only those will be used. Individual values take precedence overriding global value.
During nb_put or nb_get operations, progress is triggered only if individual values are non-zero AND progress counters is larger than the value.

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -771,6 +772,16 @@ int mca_spml_ucx_get_nb(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_
     status = ucp_get_nbi(ucx_ctx->ucp_peers[src].ucp_conn, dst_addr, size,
                      (uint64_t)rva, ucx_mkey->rkey);
 
+    if (OPAL_UNLIKELY(mca_spml_ucx.nb_get_progress_thresh &&
```
brminich: maybe it is worth to define a new  get and put functions with this code and update func pointers in `mca_spml_ucx_t` if `nb_(get|put)_progress_thresh` is not zero?
Then there will be no extra branch if this feature is not enabled.

janjust: good idea, I'll change it, avoids an extra conditional

janjust: Juist fyi, I updated the code, split the functionality to a new nb put/get call with progress and update the function pointer in case progress is used.

artpol84: Note that as far as I know calling function by pointer is similar to a branch from the CPU perspective. Might be a bit faster though as it is more stable.
So not much gain probably

artpol84: But as I’ve told @janjust I’m fine either way

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -774,6 +774,30 @@ int mca_spml_ucx_get_nb(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_
...
+{
+    unsigned int i;
+    void *rva;
+    ucs_status_t status;
+    spml_ucx_mkey_t *ucx_mkey;
+    mca_spml_ucx_ctx_t *ucx_ctx = (mca_spml_ucx_ctx_t *)ctx;
+
+    ucx_mkey = mca_spml_ucx_get_mkey(ctx, src, src_addr, &rva, &mca_spml_ucx);
+    status = ucp_get_nbi(ucx_ctx->ucp_peers[src].ucp_conn, dst_addr, size,
+                     (uint64_t)rva, ucx_mkey->rkey);
```
brminich: minor: alignment

oshmem/mca/spml/ucx/spml_ucx.c
brminich: it is already called by function pointer, but we can use another function if periodical progress is needed

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -774,6 +774,30 @@ int mca_spml_ucx_get_nb(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_
...
+
+    if (++ucx_ctx->nb_progress_cnt > mca_spml_ucx.nb_get_progress_thresh) {
```
hoopoepg: here should be error handling

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -822,7 +846,33 @@ int mca_spml_ucx_put_nb(shmem_ctx_t ctx, void* dst_addr, size_t size, void* src_
...
+    }
 
+    if (++ucx_ctx->nb_progress_cnt > mca_spml_ucx.nb_put_progress_thresh) {
+        for (i = 0; i < mca_spml_ucx.nb_ucp_worker_progress; i++) {
+            if (!ucp_worker_progress(ucx_ctx->ucp_worker)) {
+                ucx_ctx->nb_progress_cnt = 0;
+                break;
+            }
+        }
+    }
```
hoopoepg: what about just add this loop into mca_spml_ucx_put_nb instead of separate function?

oshmem/mca/spml/ucx/spml_ucx.c
```diff
@@ -774,6 +774,30 @@ int mca_spml_ucx_get_nb(shmem_ctx_t ctx, void *src_addr, size_t size, void *dst_
...
+                     (uint64_t)rva, ucx_mkey->rkey);
+
+    if (++ucx_ctx->nb_progress_cnt > mca_spml_ucx.nb_get_progress_thresh) {
+        for (i = 0; i < mca_spml_ucx.nb_ucp_worker_progress; i++) {
+            if (!ucp_worker_progress(ucx_ctx->ucp_worker)) {
+                ucx_ctx->nb_progress_cnt = 0;
+                break;
+            }
+        }
+    }
```
hoopoepg: what about just add this loop into mca_spml_ucx_get_nb instead of separate function?

oshmem/mca/spml/ucx/spml_ucx.c
artpol84: Got it!

gpaulsen:
If we're targeting for v4.0.3, it'd be good to get this and https://github.com/open-mpi/ompi/pull/7274 merged this week for the v4.0.3rc2 (either 1/24 or 1/27?)

jladd-mlnx:
@gpaulsen - thank you; we will have this and #7274 ready to go before the deadline.

Issue #6914, opened 2019-08-20
RPM build options --define 'install_in_opt 1' and --define 'install_modulefile 1' don't interact properly


jschwartz-cray:
## Background information

Trying to build RPMs from the SRPM which get installed properly under /opt with an env modulefile in /etc/modulefiles.

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)

v4.0.1

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Trying to build RPMs

### Please describe the system on which you are running

* Operating system/version: CentOS Linux release 7.5.1804 (Core)
* Computer hardware: Docker
* Network type: N/A

-----------------------------

## Details of the problem

```
rpmbuild \
    --define 'install_in_opt 1' \
    --define 'install_modulefile 1' \
    --rebuild openmpi-4.0.1-1.src.rpm
```

Currently when these options are combined the modulefile gets installed under:

```/opt/openmpi/4.0.1/share/openmpi/modulefiles```

rather than the desired location of:

```/etc/modulefiles```

which defeats the purpose of installing in /opt.  Generally people who are installing in /opt are doing so because they have many different versions installed at the same time and they want to be able to switch between them using module purge/load/unload.  By putting the modulefile under /opt as well it adds a redundant post-install step to add the path to MODULEPATH.

When --define 'install_in_opt 1' is set, the spec file should not change %{_sysconfdir}.

jsquyres:
Sorry for taking forever to get to this.  Thanks for bringing this to our attention.

FWIW, although `install_in_opt` does redefine the modulefile installation location, it doesn't do it via `%{_sysconfdir}`.

That being said, the default modulefile location is `/usr/share/Modules/modulefiles` because -- at least back when this specfile was written -- there was no common, cross-distro location where modulefiles were supposed to be written.  I don't remember exactly how that default location was picked, but it was probably where at least one distro (or HPC packager?) put them.  Is `/etc/modulefiles/ standard across lots of distros / packagers these days?

Note that you can `--define modulefile_path WHATEVER` to set the installation path to be whatever you want.  Unfortunately, it looks like this gets overwritten if `install_in_opt` is set, though.

Would you consider it better behavior to not overwrite the `modulefile_path` if a) `install_in_opt==1` and b) `modulefile_path` was set to something by the user?

jschwartz-cray:
[...]

PR #7279, opened 2020-01-07
v4.0.x: openmpi.spec: update modulefile_path behavior

jsquyres:
Allow the user to override the modulefile_path (root directory to
install the Open MPI modulefile), even if install_in_opt==1.  For
example:

rpmbuild \
    --rebuild \
    --define 'install_in_opt 1' \
    --define 'modulefile_path /path/to/my/modulefiles/openmpi/%{version}' \
    openmpi-4.0.2-1.src.rpm

(cherry picked from commit 352e575e187bfc0a129c51924c244384f3303a65)

Refs #6914
FYI @jschwartz-cray 

RMs: technically, this is an enhancement.  But it's pretty much an oversight; we really should have included this functionality in the first place.

### DIFF
--- a/contrib/dist/linux/openmpi.spec
+++ b/contrib/dist/linux/openmpi.spec
@@ -9,7 +9,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
-# Copyright (c) 2006-2016 Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2006-2019 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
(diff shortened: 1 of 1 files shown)


hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

jsquyres:
> I tried to find documentation of the `%{!?var: %define var foo}` syntax (which there was already precedent for in the spec file before this PR, so I'm not concerned), but couldn't find that syntax anywhere on either `https://rpm.org/documentation.html` or `https://rpm-packaging-guide.github.io/`. I read it as "if var is not yet defined, then define it to be foo".

RPM specfile syntax is a witch's brew of nomenclature, much of which is only documented in secret places, passed on via regal family tree lines, and otherwise un-verifiable to the general public.  

Your interpretation of this syntax is correct: "if var is not defined, do this action" (and this action happens to correspond to "define var to foo").

PR #7280, opened 2020-01-07
v3.1.x: openmpi.spec: update modulefile_path behavior

jsquyres:
Allow the user to override the modulefile_path (root directory to
install the Open MPI modulefile), even if install_in_opt==1.  For
example:

rpmbuild \
    --rebuild \
    --define 'install_in_opt 1' \
    --define 'modulefile_path /path/to/my/modulefiles/openmpi/%{version}' \
    openmpi-4.0.2-1.src.rpm

(cherry picked from commit 352e575e187bfc0a129c51924c244384f3303a65)

Refs #6914
FYI @jschwartz-cray 

RMs: technically, this is an enhancement.  But it's pretty much an oversight; we really should have included this functionality in the first place.

### DIFF
--- a/contrib/dist/linux/openmpi.spec
+++ b/contrib/dist/linux/openmpi.spec
@@ -9,7 +9,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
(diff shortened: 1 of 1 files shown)


bwbarrett:
The cray timed out on testing.  Rather than run all the tests again for code that isn't actually tested by the CI builds, I'm just going to commit.

PR #7284, opened 2020-01-08
Minor cleanup in the monitoring PML.

bosilca:

### DIFF
--- a/ompi/mca/common/monitoring/common_monitoring.c
+++ b/ompi/mca/common/monitoring/common_monitoring.c
@@ -232,7 +232,7 @@ void mca_common_monitoring_finalize( void )
 {
     if( ! mca_common_monitoring_enabled || /* Don't release if not last */
         0 < opal_atomic_sub_fetch_32(&mca_common_monitoring_hold, 1) ) return;
-    
+
     OPAL_MONITORING_PRINT_INFO("common_component_finish");
     /* Dump monitoring informations */
     mca_common_monitoring_flush(mca_common_monitoring_output_enabled,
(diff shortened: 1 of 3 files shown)


ompi/mca/common/monitoring/common_monitoring.c
```diff
@@ -253,7 +253,7 @@ void mca_common_monitoring_finalize( void )
     }
 }
 
-void mca_common_monitoring_register(void*pml_monitoring_component)
+int mca_pml_monitoring_register(void)
```
gpaulsen: Seems strange to define a `mca_pml_monitoring_register()` function in the `common_monitoring.c`, but the common files are kinda a wild west, so go for it.

bosilca: I could indeed change the name to reflect the real meaning of this function. One problem remains however, all these MCA parameters are tied to the monitoring PML, as the PML is required for the entire monitoring infrastructure to work.

gpaulsen: It's fine by me as is.

bosilca: Really funny thing we got here. I tried to fix this, and move the MCA variables out of the PML scope into the common/monitoring scope. Well, that didn't work well, the common scope is removed last, after the last usage of any user of the common support, which means a moment where the common shared libraries are already dlclosed. And this, generate segfaults on all string registrations.

jsquyres: @hjelmn ...?  😄

PR #7293, opened 2020-01-08
Topic/v3.0.x/from pr7190 7192

gpaulsen:
Fix comment type-os, from master PR7190 and 7192.


jsquyres:
@gpaulsen Is this necessary on the v3.0.x branch?  I.e., does the lack of it preclude / cause conflicts with bringing in some other bug fix?

If not, I'm inclined to not take it... we should be raising the bar to bring fewer things into the v3.0.x and v3.1.x branches, and instead focusing on the v4.0.x branch.

gpaulsen:
No, not needed.  I'll close.

PR #7299, opened 2020-01-10
Fix some compiler warnings.

awlauria:
Silence unused variables, incompatible pointer types,
un-initialized variables, and signed/unsigned comparisons.

### DIFF
--- a/ompi/mca/coll/base/coll_base_util.c
+++ b/ompi/mca/coll/base/coll_base_util.c
@@ -255,7 +255,7 @@ static int free_vecs_callback(struct ompi_request_t **rptr) {
 }
 
 int ompi_coll_base_retain_datatypes_w( ompi_request_t *req,
-                                       ompi_datatype_t *stypes[], ompi_datatype_t *rtypes[]) {
+                                       ompi_datatype_t * const stypes[], ompi_datatype_t * const rtypes[]) {
     ompi_coll_base_nbc_request_t *request = (ompi_coll_base_nbc_request_t *)req;
     bool retain = false;
     ompi_communicator_t *comm = request->super.req_mpi_object.comm;
@@ -282,8 +282,8 @@ int ompi_coll_base_retain_datatypes_w( ompi_request_t *req,
         }
     }
     if (OPAL_UNLIKELY(retain)) {
-        request->data.vecs.stypes = stypes;
-        request->data.vecs.rtypes = rtypes;
+        request->data.vecs.stypes = (ompi_datatype_t **) stypes;
+        request->data.vecs.rtypes = (ompi_datatype_t **) rtypes;
         if (req->req_persistent) {
             request->cb.req_free = req->req_free;
             req->req_free = free_vecs_callback;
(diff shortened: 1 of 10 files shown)


ompi/mca/coll/base/coll_base_util.c
```diff
@@ -255,7 +255,7 @@ static int free_vecs_callback(struct ompi_request_t **rptr) {
 }
 
 int ompi_coll_base_retain_datatypes_w( ompi_request_t *req,
-                                       ompi_datatype_t *stypes[], ompi_datatype_t *rtypes[]) {
+                                       ompi_datatype_t * const stypes[], ompi_datatype_t * const rtypes[]) {
```
jsquyres: Are you sure that this is correct -- add `const` to the param type in the declaration, only to cast it away in its usage in the definition?

Is there a reason to do that?  It seems like a no-op...?

awlauria: It is a no-op in this case. It's essentially a hack to make gcc happy. At the highest level API we're passing in a `const MPI_Datatype[]`, and somewhere down the stack (I suppose here), that const needs to be lost somewhere. I didn't want to touch the higher level api, so I figured here is the best place to fix it.

The gcc warning itself was converting a const pointer to a non-const pointer when using this function.

jsquyres: @bosilca Can you chime in on this?

bosilca: Yeah, this warning was there for quite some time, but I was not sure if it is worth fixing it. We are altering the datatype inside the ```ompi_coll_base_retain_datatypes_w``` function (by increasing the refcount), so the datatype is not const, but the pointer to it should be. Thus, I think the proposed fix is correct.

However, we do need to cast it away below, as we cant change the type of ```vecs.stypes```and ```vecs.rtypes``` because these will be used later by the ```release_datatype``` function, where nothing shall be constant (the types can disappear if the user called ```MPI_Type_free``` on them).

awlauria: Yeah, that is correct. I tried changing the types of vect.types to also be constant and not have the cast. That not working, I settled for the current "fix".

ompi/mca/coll/tuned/coll_tuned_decision_fixed.c
```diff
@@ -806,8 +806,10 @@ int ompi_coll_tuned_scatter_intra_dec_fixed(const void *sbuf, int scount,
...
                (communicator_size > intermediate_comm_size) &&
-               (block_size >= ompi_coll_tuned_scatter_intermediate_msg) &&
-               (block_size < ompi_coll_tuned_scatter_large_msg)) {
+               (ompi_coll_tuned_scatter_intermediate_msg > -1) &&
+               (block_size >= (size_t) ompi_coll_tuned_scatter_intermediate_msg) &&
+               (ompi_coll_tuned_scatter_large_msg > -1) &&
+               (block_size < (size_t) ompi_coll_tuned_scatter_large_msg)) {
         return ompi_coll_base_scatter_intra_linear_nb(sbuf, scount, sdtype,
                                                       rbuf, rcount, rdtype,
                                                       root, comm, module,
```
bosilca: It makes little sense for these 2 cut-of MCA params (ompi_coll_tuned_scatter_intermediate_msg and ompi_coll_tuned_scatter_large_msg) to be negative. Maybe we should fix those.

awlauria: I agree.

Changing them both to be a `size_t` or an `unsigned` seems to make sense. That way if the user does specify -1 at the command line, it would essentially force this path.

awlauria:
bot:ibm:retest

awlauria:
bot:retest

hppritcha:
bot:ompi:retest

PR #7304, opened 2020-01-14
btl/vader: modify how the max attachment address is determined

hjelmn:
This PR removes the constant defining the max attachment address and
replaces it with the largest address that shows up in /proc/self/maps.
This should address issues found on AARCH64 where the max address
may differ based on the configuration.

Since the calculated max address may differ between processes the
max address is sent as part of the modex and stored in the endpoint
data.

### DIFF
--- a/opal/mca/btl/vader/btl_vader_xpmem.c
+++ b/opal/mca/btl/vader/btl_vader_xpmem.c
@@ -22,7 +22,47 @@
 
 int mca_btl_vader_xpmem_init (void)
 {
-    mca_btl_vader_component.my_seg_id = xpmem_make (0, VADER_MAX_ADDRESS, XPMEM_PERMIT_MODE, (void *)0666);
+    /* Any attachment that goes past the Linux TASK_SIZE will always fail. To prevent this we need to
+     * determine the value of TASK_SIZE. On x86_64 the value was hard-coded in vader to be
+     * 0x7ffffffffffful but this approach does not work with AARCH64 (and possibly other architectures).
+     * Since there is really no way to directly determine the value we can (in all cases?) look through
+     * the mapping for this process to determine what the largest address is. This should be the top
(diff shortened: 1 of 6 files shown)


opal/mca/btl/vader/btl_vader_xpmem.c
```diff
@@ -22,14 +22,37 @@
...
 
+    /* Any attachment that goes past the Linux TASK_SIZE will always fail. To prevent this we need to
+     * determine the value of TASK_SIZE. On x86_64 the value was hard-coded in vader to be
+     * 0x7ffffffffffful but this approach does not work with AARCH64 (and possibly other architectures).
+     * Since there is really no way to directly determine the value we can (in all cases?) look through
+     * the mapping for this process to determine what the largest address is. This should be the top
+     * of the stack. No heap allocations should be larger than this value. Since the largest address
+     * may differ between processes the value must be shared as part of the modex and stored in the
+     * endpoint. */
+    FILE *fh = fopen("/proc/self/maps", "r");
```
jsquyres: Just to be totally safe, you should check for failure here.

hjelmn: Sure. Will do.

hjelmn: Done.

hjelmn:
@sekifjikkatsu This should fix the vader issue for you. I opted with a more dynamic approach as the TASK_SIZE can vary on AARCH64.

sekifjikkatsu:
@hjelmn Thank you very much, I will try to patch "https://patch-diff.githubusercontent.com/raw/open-mpi/ompi/pull/7304.diff" to my test environment.

hjelmn:
@sekifjikkatsu Thank you for verifying. I plan to PR this and the other vader fixes to all active releases.

Issue #7303, opened 2020-01-14
Potential deadlock in rcache?


devreal:
I'm trying to run a PaRSEC benchmark (`testing_dgeqrf`) on a Cray XC40 with the latest v4.0.x compiled with the Intel compiler. For larger problem sizes (16 nodes, 24 threads per node, 20k^2 elements per process) the run locks up after all tasks have been processed and I see the following aggregated backtrace in DDT:

```
Threads,Function
1,progress_engine (opal_progress_threads.c:105)
1,  opal_libevent2022_event_base_loop (event.c:1630)
1,    poll_dispatch (poll.c:165)
1,      poll
1,progress_engine (pmix_progress_threads.c:232)
1,  opal_libevent2022_event_base_loop (event.c:1630)
1,    epoll_dispatch (epoll.c:407)
1,      epoll_wait
22,__parsec_thread_init
22,  __parsec_context_wait
18,    nanosleep
1,    sched_lfq_select
1,      parsec_hbbuffer_pop_best
3,    __parsec_task_progress
3,      complete_hook_of_dgeqrf_dtsmqr
3,        intercept_munmap (memory_patcher_component.c:167)
3,          _intercept_munmap (memory_patcher_component.c:153)
3,            opal_mem_hooks_release_hook (memory.c:129)
3,              mca_rcache_base_mem_cb (rcache_base_mem_cb.c:63)
3,                mca_rcache_grdma_invalidate_range (rcache_grdma_module.c:496)
3,                  mca_rcache_base_vma_iterate (rcache_base_vma.c:153)
3,                    mca_rcache_base_vma_tree_iterate (rcache_base_vma_tree.c:105)
3,                      opal_interval_tree_traverse (opal_interval_tree.c:706)
3,                        inorder_traversal (opal_interval_tree.c:670)
3,                          inorder_traversal (opal_interval_tree.c:654)
3,                            inorder_traversal (opal_interval_tree.c:670)
3,                              inorder_traversal (opal_interval_tree.c:670)
3,                                inorder_traversal (opal_interval_tree.c:670)
3,                                  inorder_traversal (opal_interval_tree.c:670)
3,                                    inorder_traversal (opal_interval_tree.c:670)
3,                                      inorder_traversal (opal_interval_tree.c:670)
3,                                        inorder_traversal (opal_interval_tree.c:670)
2,                                          inorder_traversal (opal_interval_tree.c:654)
1,                                            inorder_traversal (opal_interval_tree.c:654)
[...]

PR #7307, opened 2020-01-15
Release the vma_lock before deregistration to avoid deadlock

devreal:
Don't hold the `vma_lock` while deregistering memory in `mca_rcache_grdma_evict_lru_local` as that may lead to a deadlock with other threads attempting to take the same lock while releasing memory.

See #7303 for more details.


awlauria:
Added the WIP label until #7319 is resolved.

PR #7315, opened 2020-01-17
Handle error cases in TCP BTL (v2)

abouteiller:
This set of patches corrects the behavior of the TCP BTL under some error scenarios.

In a previous iteration of that patch, the reporting of errors from the BTL has caused MPI_FINALIZE to abort/spew useless error messages during normal shutdown of the TCP sockets. Relevant history: https://github.com/open-mpi/ompi/pull/5916 . This patch adds a mechanism for discriminating between failures and normal closing of sockets.


abouteiller:
bot:recheck

awlauria:
@abouteiller looks like you didn't sign off your first commit. I think after you do that this is ok to merge.

PR #7317, opened 2020-01-18
Check the return from "chdir" to avoid infinite loop

rhc54:
When autogen attempts to change to a new directory while processing a
subdirectory, it can get into an infinite loop if that directory
doesn't exist as it will remain in the top-level directory, see itself
there (as "autogen.pl"), and re-execute itself. Check the return code on
"chdir" and error out if it fails.

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -4,7 +4,7 @@
 # Copyright (c) 2010      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
-# Copyright (c) 2013-2014 Intel, Inc.  All rights reserved.
+# Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
 # Copyright (c) 2015-2019 Research Organization for Information Science
 #                         and Technology (RIST).  All rights reserved.
 # Copyright (c) 2015      IBM Corporation.  All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
@hppritcha It appears the Cray is having another bad day. It made it all the way thru build and the unit tests, only to "timeout" at the end. As this PR has nothing to do with execution, and the build verified this change didn't break anything, I'm going ahead to commit it.

rhc54:
@artemry-mlnx Just an FYI. I noticed that the new Mellanox CI took over an hour on this PR. It looks like much of that time may have been spent waiting for an Azure allocation?

Might be worth noting this: when I installed the app, I was required to open an Azure account and I actually read the terms (gasp!). Azure makes no guarantees of responsiveness, and "penalizes" those who generate frequent requests by lowering their priority so that others can get access (only a set number of resources are set aside for this purpose). Given the number of PRs we file, OMPI may well fall into that category, especially when one or more PRs are frequently being updated.

It is therefore possible, albeit unlikely, that a PR could sit for days waiting for Mellanox CI to get an allocation. I would suggest you add a timeout in your script, and if you don't get an allocation within a certain period of time, then put a comment on the PR that Mellanox CI was "unable to run" (don't mark it as failed!). We'll then know it is okay to merge and we'll catch any problems later when an allocation can be obtained.

artemry-mlnx:
@rhc54 
Currently Mellanox CI takes ~1h 10min (mostly spent within Open MPI building). We have TODOs to decrease this time (with faster CPU with greater number of cores to use parallel building more, use SSD, maybe some improvements in the scripts, etc.)

For Mellanox CI we use dedicated CI hosts located in the Mellanox lab so Azure policies and limitations should not impact Open MPI CI. With current average Open MPI CI load (several CI runs per a day) there should not bottlenecks for now. Overall Mellanox Open MPI CI timeout is set to 4h now.

rhc54:
It would be good if you can reduce the time - that is long enough to make it appear to be "stuck". Good to hear you have a timeout, though most developers will wonder if something hasn't completed in less than 4 hours 😄

PR #7323, opened 2020-01-21
Trap wrong parameters to MPI_Init_thread.

### LINKED ISSUE #7320: Weird error message from MPI_Init_thread
I am experimenting with TAMPI [1], which for some reason introduces a new thread-level `MPI_TASK_MULTIPLE`. Due to wrong library linking order, I managed to bypass the TAMPI library and pass `MPI_TASK_MULTIPLE` directly to Open MPI's `MPI_Init_thread`. Of course, Open MPI barked at it, which lead to this quite humorous (albeit misleading) error message:

```
*** The MPI_Init_thread() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```

Obviously my application calling *the real* `MPI_Init_thread` this way is erroneous but maybe Open MPI can provide a more helpful error message in case the `required` argument to `MPI_Init_thread` is off?

[1] https://github.com/bsc-pm/tampi/

bosilca:
Instead of triggering the errhandler early in the initialization process, do
a serialized initialization and report the error once all the supporting
infrastructure is up and running.

Fixes #7320 

### DIFF
--- a/ompi/mpi/c/init_thread.c
+++ b/ompi/mpi/c/init_thread.c
@@ -47,26 +47,36 @@ static const char FUNC_NAME[] = "MPI_Init_thread";
 int MPI_Init_thread(int *argc, char ***argv, int required,
                     int *provided)
 {
-    int err;
+    int err, safe_required = MPI_THREAD_SERIALIZED;
 
     ompi_hook_base_mpi_init_thread_top(argc, argv, required, provided);
 
-    if ( MPI_PARAM_CHECK ) {
-        if (required < MPI_THREAD_SINGLE || required > MPI_THREAD_MULTIPLE) {
-            ompi_mpi_errors_are_fatal_comm_handler(NULL, NULL, FUNC_NAME);
-        }
+    /* Detect an incorrect thread support level, but dont report until we have the minimum
+     * infrastructure setup.
+     */
+    if( (MPI_THREAD_SINGLE == required) || (MPI_THREAD_SERIALIZED == required) ||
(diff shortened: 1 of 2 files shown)


ompi/mpi/c/init_thread.c
```diff
@@ -47,28 +47,36 @@ static const char FUNC_NAME[] = "MPI_Init_thread";
...
     ompi_hook_base_mpi_init_thread_top(argc, argv, required, provided);
 
-    if ( MPI_PARAM_CHECK ) {
-        if (required < MPI_THREAD_SINGLE || required > MPI_THREAD_MULTIPLE) {
-            ompi_mpi_errors_are_fatal_comm_handler(NULL, NULL, FUNC_NAME);
-        }
+    /* Detect an incorrect thread support level, but dont report until we have the minimum
+     * infrastructure setup.
+     */
+    if( (MPI_THREAD_SINGLE == required) && (MPI_THREAD_SERIALIZED == required) &&
```
hppritcha: i don't think I understand this logic.  shouldn't it be ```||```

bosilca: You're absolutely right, || shall be.

awlauria:
Instead of having the safe_required logic, couldn't we move the checks to `ompi_mpi_init`?

bosilca:
That could be another solution, but in general we avoid checking arguments except on the top level functions. The second reason I preferred to have the check here is the fact that we are in the initialization function and if we want to be able to nicely report the error, we need the OMPI infrastructure to be up and running, in other terms we need the ```ompi_mpi_init``` to succeed.

PR #7324, opened 2020-01-21
hwloc2: advance hwloc git submodule

jsquyres:
Advance to hwloc-2.1.0rc2-33-g38433c0f, which includes a .gitignore
update that we want here in Open MPI.

***Be warned:*** this is actually 33 commits beyond the hwloc v2.1.0 tag.

Signed-off-by: Jeff Squyres <jsquyres@cisco.com>

Fixes #7316.
FYI @bgoglin

PR #7326, opened 2020-01-21
opal: ensure opal_gethostname() always returns a value

### LINKED ISSUE #7322: opal_gethostname segfaults
We have always permitted the use of ```opal_output``` prior to successfully executing ```opal_init```, with the output simply being printed to stderr. However, if you hit that scenario now, the program segfaults because ```opal_gethostname``` accesses ```opal_process_info.nodename```, which hasn't been defined yet.

What you need to do is have ```opal_gethostname``` check for ```NULL```, and if it finds it, then go ahead and call ```gethostname``` instead. Or do the protection in ```opal_output```.

I don't care which option gets used, but we shouldn't be segfaulting for no reason.

jsquyres:
We initially thought it was a safe bet that opal_gethostname() would
never be called before opal_init().  However, it turns out that there
are some cases -- e.g., developer debugging -- where it is useful to
call opal_output() (which calls opal_gethostname()) before
opal_init().

Hence, we need to guarantee that opal_gethostname() always returns a
valid value.  If opal_gethostname() finds NULL in
opal_process_info.nodename, simply call the internal function to
initialize opal_process_info.nodename.

Fixes #7322.

### DIFF
--- a/opal/runtime/opal.h
+++ b/opal/runtime/opal.h
@@ -115,15 +115,31 @@ OPAL_DECLSPEC void opal_warn_fork(void);
 OPAL_DECLSPEC int opal_register_params(void);
 
 /**
- * Wrapper to return the hostname value that is in
-   opal_process_info.nodename, as opposed to calling gethostname()
-   directly, which is not guaranteed to be null-terminated and
-   varies in its behavior depending on implementation. The 
-   opal_process_info.nodename value is first populated in 
-   opal/runtime/opal_init.c
+ * Internal function.  Should not be called directly (should only be
+ * invoked internally by opal_init() and opal_gethostname()).
  */
-static inline const char *opal_gethostname( void ) {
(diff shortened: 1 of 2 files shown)


jsquyres:
@hppritcha Something is unfortunately borked in the Cray Jenkins.  It's failing to checkout the hwloc git submodule (and this PR doesn't touch the hwloc submodule): https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.platforms/5870/console

Can you have a look?

([also see my comment about the Cray CI/hwloc submodule](https://github.com/open-mpi/ompi/pull/7324#issuecomment-576806493) on #7324)

bwbarrett:
bot:ompi:retest

I turned off shallow copies for submodules.  I thought Jenkins cloned from the target commit, but that appears not to be the case, so shallow copies won't work.

PR #7333, opened 2020-01-23
oshmem/spml/ucx Fix memory leak and size mismatch

janjust:
Address coverity issues:
CID 1457920
CID 1457921
CID 1457922
CID 1457923


janjust:
closing since there is already a PR addressing the same

Issue #7309, opened 2020-01-15
OMPIO symbol abstraction violation in v3.0.x/v3.1.x; only shows up with GCC 8.2.5, 8.3.1. (RHEL 8.0, RHEL 8.1)


mwheinz:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

OMPI 3.1.4, 3.1.5, 3.1.6

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Building from source using the IFS/OPA build scripts.

### Please describe the system on which you are running

* Operating system/version: RHEL 8.1 / GCC 8.3.1-4
* Computer hardware: Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz
* Network type: OPA

-----------------------------

## Details of the problem

After building OMPI 3.1.4 for RHEL 8.1, I encountered the following warnings when I ran ompi_info. Problem does not seem to occur when building OMPI 4.0.2.

Is this expected or a known issue?
```
[LINUX hds1fna2271 20200115_1130 MPI]# /usr/mpi/gcc/openmpi-3.1.4/bin/ompi_info | grep btl
[hds1fna2271:81579] mca_base_component_repository_open: unable to open mca_fcoll_static: /usr/mpi/gcc/openmpi-3.1.4/lib64/openmpi/mca_fcoll_static.so: undefined symbol: mca_common_ompio_register_print_entry (ignored)
[hds1fna2271:81579] mca_base_component_repository_open: unable to open mca_fcoll_two_phase: /usr/mpi/gcc/openmpi-3.1.4/lib64/openmpi/mca_fcoll_two_phase.so: undefined symbol: mca_common_ompio_register_print_entry (ignored)
[hds1fna2271:81579] mca_base_component_repository_open: unable to open mca_fcoll_dynamic_gen2: /usr/mpi/gcc/openmpi-3.1.4/lib64/openmpi/mca_fcoll_dynamic_gen2.so: undefined symbol: mca_common_ompio_register_print_entry (ignored)
[hds1fna2271:81579] mca_base_component_repository_open: unable to open mca_fcoll_individual: /usr/mpi/gcc/openmpi-3.1.4/lib64/openmpi/mca_fcoll_individual.so: undefined symbol: mca_common_ompio_file_write (ignored)
[...]

PR #7336, opened 2020-01-24
ompio: fix abstraction violation in 3.0.x series

edgargabriel:
This commit is the minimum set required to fix the abstraction
violation in the ompio components for the 3.0.x series. This commit
does not change names of the structues (as it was done in the masters
and 4.x series), does also not address the compilation part (i.e.,
including headers of components). It should, however, take care of the
runtime loading problem of data from different components.

It is, to some extent, a simplified subset of pull request 5263.

Fixes issue https://github.com/open-mpi/ompi/issues/7309

### DIFF
--- a/ompi/mca/common/ompio/Makefile.am
+++ b/ompi/mca/common/ompio/Makefile.am
@@ -23,10 +23,15 @@ if OMPI_PROVIDE_MPI_FILE_INTERFACE
 
 headers = \
 	common_ompio_print_queue.h \
+	common_ompio_aggregators.h \
+	common_ompio_request.h \
+	common_ompio_callbacks.h \
 	common_ompio.h
 
 sources = \
 	common_ompio_print_queue.c \
+	common_ompio_aggregators.c \
+	common_ompio_request.c \
 	common_ompio_file_open.c   \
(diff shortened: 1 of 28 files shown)


edgargabriel:
@jsquyres and @mwheinz could you please test whether this commit makes the 3.0 series work on redhat 8? 

I did have to do a couple of more manual changes compared to the v3.1.x series, since a number of things changed between 3.0.x and 3.1.x in ompio (mca parameters, aggregator selection algorithm, etc), but also in ompi itself (ompi_info_t vs. opal_info_t for example).

mwheinz:
Looking at it now, but my makesrpm script seems to be having an issue with 3.0.x that's unrelated to your patch.

mwheinz:
@edgargabriel - I'm noticing a discrepency between the pull request and the commit message on 9f93743ad816689bed7ccae32be48ccf390d7feb - the commit says "3.1.x". am I looking at the wrong commit?

edgargabriel:
@mwheinz, that was probably because I cherry picked the commit from the v3.1.x branch.

jsquyres:
bot:ompi:retest

Issue #7330, opened 2020-01-23
[[INVALID],INVALID] FORCE-TERMINATE AT Not found:-13 - error plm_rsh_component.c(327)


adrianreber:
While trying to build petsc for OpenHPC I get a core dump in mpiexec on CentOS 8 with the Open MPI 4.0.2 build from OpenHPC:

```
[   91s] + mpiexec --oversubscribe -n 1 printenv
[   91s] --------------------------------------------------------------------------
[   91s] The value of the MCA parameter "plm_rsh_agent" was set to a path
[   91s] that could not be found:
[   91s] 
[   91s]   plm_rsh_agent: rsh
[   91s] 
[   91s] Please either unset the parameter, or check that the path is correct
[   91s] --------------------------------------------------------------------------
[   91s] [ohpc1:04221] [[INVALID],INVALID] FORCE-TERMINATE AT Not found:-13 - error plm_rsh_component.c(327)
[   91s] [ohpc1:04221] *** Process received signal ***
[   91s] [ohpc1:04221] Signal: Segmentation fault (11)
[   91s] [ohpc1:04221] Signal code: Address not mapped (1)
[   91s] [ohpc1:04221] Failing at address: (nil)
[   91s] [ohpc1:04221] [ 0] /lib64/libpthread.so.0(+0x12dd0)[0x7f70eb00add0]
[   91s] [ohpc1:04221] *** End of error message ***
```
At first I was not able to reproduce it outside of our build system, but setting `OMPI_MCA_plm_rsh_agent` to `does-not-exist` I am able to reproduce this with the OS provided Open MPI 4.0.2 on RHEL 8:
```
[mpi@sms ~]$ export OMPI_MCA_plm_rsh_agent=does-not-exist
[mpi@sms ~]$ mpiexec --oversubscribe -n 1 printenv 
--------------------------------------------------------------------------
The value of the MCA parameter "plm_rsh_agent" was set to a path
that could not be found:

  plm_rsh_agent: does-not-exist

Please either unset the parameter, or check that the path is correct
--------------------------------------------------------------------------
[sms:16573] [[INVALID],INVALID] FORCE-TERMINATE AT Not found:-13 - error plm_rsh_component.c(327)
[sms:16573] *** Process received signal ***
[sms:16573] Signal: Segmentation fault (11)
[sms:16573] Signal code: Address not mapped (1)
[sms:16573] Failing at address: (nil)
[...]

PR #7337, opened 2020-01-24
plm/rsh: Fix segv on missing agent.

jjhursey:
* Additionally, fixes the `NULL` option to `OMPI_MCA_plm_rsh_agent`
   would would also lead to a segv. Now it operates as intended by
   disqualifying the `rsh` component and falling back onto the `isolated`
   component.
 * Refs #7330


rhc54:
Oh - please be sure to bring this over to PRRTE!

PR #7355, opened 2020-02-02
v4.0.x: Make C and Fortran types for MPI sentinels agree in size

jsquyres:
Fix the C types for the following:

* MPI_UNWEIGHTED
* MPI_WEIGHTS_EMPTY
* MPI_ARGV_NULL
* MPI_ARGVS_NULL
* MPI_ERRCODES_IGNORE

There is lengthy discussion on
https://github.com/open-mpi/ompi/pull/7210 describing the issue; the
gist of it is that the C and Fortran types for several MPI global
sentenial values should agree (specifically: their sizes must(**)
agree).  We erroneously had several of these array-like sentinel
values be "array-like" values in C.  E.g., MPI_ERRCODES_IGNORE was an
(int *) in C while its corresponding Fortran type was "integer,
dimension(1)".  On a 64 bit platform, this resulted in C expecting the
symbol size to be sizeof(int*)==8 while Fortran expected the symbol
size to be sizeof(INTEGER, DIMENSION(1))==4.

That is incorrect -- the corresponding C type needed to be (int).
Then both C and Fortran expect the size of the symbol to be the same.

(**) NOTE: This code has been wrong for years.  This mismatch of types
typically worked because, due to Fortran's call-by-reference
semantics, Open MPI was comparing the *addresses* of these instances,
not their *types* (or sizes) -- so even if C expected the size of the
symbol to be X and Fortran expected the size of the symbol to be Y
(where X!=Y), all we really checked at run time was that the addresses
of the symbols were the same.  But it caused linker warning messages,
and even caused errors in some cases.

Specifically: due to a GNU ld bug
(https://sourceware.org/bugzilla/show_bug.cgi?id=25236), the 5 common
symbols are incorrectly versioned VER_NDX_LOCAL because their
definitions in Fortran sources have smaller st_size than those in
libmpi.so.

This makes the Fortran library not linkable with lld in distributions
that ship openmpi built with -Wl,--version-script
(https://bugs.llvm.org/show_bug.cgi?id=43748):

  % mpifort -fuse-ld=lld /dev/null
  ld.lld: error: corrupt input file: version definition index 0 for symbol
  mpi_fortran_argv_null_ is out of bounds
  >>> defined in /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi_usempif08.so
  ...

If we fix the C and Fortran symbols to actually be the same size, the
problem goes away and the GNU ld bug does not come into play.

This commit also fixes a minor issue that MPI_UNWEIGHTED and
MPI_WEIGHTS_EMPTY were not declared as Fortran arrays (not fully fixed
by commit 107c0073dd11fb90d18122c521686f692a32cdd8).

Fixes open-mpi/ompi#7209

Signed-off-by: Fangrui Song <i@maskray.me>
(cherry picked from commit 5609268e90cb0ff7b2431d29041c10a700fd6996)

### DIFF
--- a/ompi/mpi/fortran/base/gen-mpi-mangling.pl
+++ b/ompi/mpi/fortran/base/gen-mpi-mangling.pl
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2015      Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
-# Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
+# Copyright (c) 2015-2020 Cisco Systems, Inc.  All rights reserved.
 # $COPYRIGHT$
 #
 # Subroutine to generate a bunch of Fortran declarations and symbols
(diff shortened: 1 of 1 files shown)


jsquyres:
@gpaulsen @hppritcha Did we miss the boat for v4.0.3?

It would be nice -- but not critical -- to get this in there.

NEWS item:

* Fixed an issue with a mismatch of Fortran and C symbol sizes that triggered an error in certain versions of GNU ld.  Thanks to Fangrui Song for identifying and fixing the issue.

hppritcha:
@gpaulsen lets pull this in and do a rc4

hppritcha:
@jsquyres will this break ABI compatibility with 4.0.x where x <=2?

jsquyres:
@hppritcha Good question.  I don't think so, but let me test.

jsquyres:
My tests indicate that ABI should be preserved.

Note that I just updated the NEWS item in a prior comment on this PR -- I forgot that we also fixed `MPI_UNWEIGHTED` and `MPI_WEIGHTS_EMPTY` in the `mpi_f08` module.  Previously, this wouldn't have compiled:

```fortran
  call MPI_DIST_GRAPH_CREATE(MPI_COMM_WORLD, n, sources, degrees, &
       destinations, MPI_UNWEIGHTED, info, reorder, intercomm)
  call MPI_DIST_GRAPH_CREATE(MPI_COMM_WORLD, n, sources, degrees, &
       destinations, MPI_WEIGHTS_EMPTY, info, reorder, intercomm)
```

PR #7373, opened 2020-02-07
OSHMEM/SEGMENTS: increase number of max segments

hoopoepg:
- increase number of max segments to allow application be launched
  on some Ubuntu configurations


yosefe:
bot:retest

yosefe:
@hoopoepg pls backport to 4.0.x

PR #7390, opened 2020-02-12
v3.0.x: Fix pgcc18 support.

awlauria:
  it for gcc.


ibm-ompi:

 Gist: https://gist.github.com/4afefc8ca6d7c712527d816492e58e08

jjhursey:
I think it is safe to ignore the IBM failure on this PR. I'll clear it and leave the Gist.
I suspect that the issue is with the hostnames being in a hash format which messes up the regex parser in this generation of the code. Given just the right set of hashes the regex can be incorrect. This is a known issue in these older branches that is fixed in the newer ones.

PR #7391, opened 2020-02-12
v3.1.x: Fix pgcc18 support.

awlauria:
  it for gcc.


gpaulsen:
looks like ARMv8 had some non-PR related issue.

bwbarrett:
bot:ompi:retest

PR #7392, opened 2020-02-12
v4.0.x: Fix pgcc18 support.

awlauria:
  it for gcc.


hppritcha:
@gpaulsen I think we should take this if there aren't MTT/CI issues.

gpaulsen:
awlauria is from IBM team.  Removing NEWS label.

hppritcha:
removing news again because pgcc18 is not ready time.

Issue #7384, opened 2020-02-11
failure during MPI_Win_detach


naughtont3:
## Background information

Application failure with Open MPI with one sided communication (OSC).   

Reporting on behalf of user to help track problem.  

The test works fine with MPICH/3.x, Spectrum MPI, and Intel MPI.

### What version of Open MPI are you using? 

 - Fails with latest master (c6831c59c8)
 - Need to check status with v4.0.x and v3.0.x

### Describe how Open MPI was installed 

Standard tarball build can reproduce, using gcc/8.x compiler suite (gcc-8, g++-8, gfortran-8).  Need gcc > 8.x to avoid past gfortran bugs.

### Please describe the system on which you are running

* Operating system/version: Linux
* Computer hardware:  
* Network type: 

Reproducible on any Linux workstation (Need gcc > 8.x to avoid past gfortran bugs)

-----------------------------

## Details of the problem

You should be able to reproduce this failure on any Linux workstation. The only
thing you need to make sure is to use the gcc/8.x compiler suite (gcc-8, g++-8,
gfortran-8) since other versions are buggy in their gfortran part.

 - 1. git clone https://gitlab.com/DmitryLyakh/ExaTensor.git
 - 2. git checkout openmpi_fail
 - 3. export PATH_OPENMPI=PATH_TO_OPENMPI_ROOT_DIR
 - 4. make
 - 5. Copy the produced binary Qforce.x into some directory and place both attached scripts there as well
- 6. Run run.exatensor.sh (it runs 4 MPI processes, which is the minimal configuration; each process runs up to 8 threads, which is also mandatory,but all these can be run on a single node)

I added `.txt` extension to attach to github ticket.
 - [run.exatensor.sh.txt](https://github.com/open-mpi/ompi/files/4185922/run.exatensor.sh.txt)
 - [exec.sh.txt](https://github.com/open-mpi/ompi/files/4185923/exec.sh.txt)

Normally `run.exatensor.sh` invokes `mpiexec` with the binary directly, but for some reason the `mpiexec` from the latest GIT master branch fails to load some dynamic libraries (libgfortran), so I introduced a workaround where `run.exatensor.sh` invokes `mpiexec` with `exec.sh`, which in turn executes the binary `Qforce.x`. Previous OpenMPI versions did not have this issue by the way. But all of them fail in `MPI_Win_detach` as you can see below:

```shell
Destroying tensor dtens ... [exadesktop:32108] *** An error occurred in MPI_Win_detach
[...]

PR #7415, opened 2020-02-17
v4.0.x: osc/rdma: modify attach to check for region overlap

hjelmn:
This commit addresses two issues in osc/rdma:
    
  1) It is erroneous to attach regions that overlap. This was being
      allowed but the standard does not allow overlapping attachments.
    
  2) Overlapping registration regions (4k alignment of attachments)
       appear to be allowed. Add attachment bases to the bookeeping
       structure so we can keep better track of what can be detached.

It is possible that the standard did not intend to allow 2). If that
is the case then 2) should fail in the same way as 1). There should
be no technical reason to disallow 2) at this time.
    
References #7384

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma.h
+++ b/ompi/mca/osc/rdma/osc_rdma.h
@@ -13,6 +13,7 @@
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012-2013 Sandia National Laboratories.  All rights reserved.
  * Copyright (c) 2016-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2020      Google, LLC. All rights reserved.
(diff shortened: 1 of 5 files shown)


jjhursey:
bot:ibm:gnu:retest

ibm-ompi:

 Gist: https://gist.github.com/f98bd5e7f856529bccf4fac3cb09f5f0

jjhursey:
bot:ibm:gnu:retest

gpaulsen:
@hppritcha This seems like a lot of changes for this late in the v4.0.3 cycle.  Do we need this for v4.0.3?

hppritcha:
@gpaulsen I agree.

gpaulsen:
@hppritcha I think this is ready to review and merge when you're ready.

PR #7416, opened 2020-02-17
opal/mca: check if the user home directory is NULL

hjelmn:
This commit fixes an issue in the MCA base variable system. The
code was retrieving the user home directory (from HOME) and
attempting to use it to build a search path for config files.
In this case user-level configuration directories have been
enabled so the appropriate thing to do is to print an error
message and return. This commit makes that change. It does not
ensure that HOME is set correctly.


hjelmn:
@rhc54 FYI. HOME is not set in the environment when running with master. That is probably a bug.

PR #7268, opened 2020-01-02
make mpifort obey disable-wrapper-runpath

hppritcha:
related to #6539

Signed-off-by: Howard Pritchard <howardp@lanl.gov>

PR #7420, opened 2020-02-18
Topic/pr7268 back to v3.1.x

hppritcha:
back port of PR #7268 plus a fix to v3.1.x.


hppritcha:
actually don't think this is needed.  closing.

PR #7421, opened 2020-02-18
osc/rdma: fix bug in attach for non-debug builds

hjelmn:
This commit fixes an issue with non-debug builds where adding an
attachment to the attachment list doesn't actually happen. This
causes all MPI_Win_detach calls to fail. The call was within an
assert which is optimized out in optimized builds.


naughtont3:
FWIW, my build/test w/o debug worked for dmitry test on my desktop.

PR #7425, opened 2020-02-19
opal/atomic: clean out s390(x)

hjelmn:
This commit removes the CMA support for s390 and s390x. These
architectures have been unsupported for awhile and no one has
verified that CMA actually works with Open MPI on these systems.

s390 and s390x will continue to work with Open MPI without CMA.


hjelmn:
Ran across this while cleaning out the __sync built-in atomics. This is an untested architecture so there is no reason to try and support CMA on it. @gpaulsen Does IBM care? I have to assume no.

gpaulsen:
It's safe to remove support for the IBM 390 and 390x, as they're over 20 years out of support.
Thanks.

PR #7426, opened 2020-02-19
opal/asm: remove MIPS

hjelmn:
This commit removes the code specific to MIPS. This architecture
has been unsupported for some time. Open MPI will continue to work
on MIPS with C11 and __atomic but will not longer use CMA for
shared memory.


hjelmn:
Doesn't exist anymore. Thats how untested MIPS was.

PR #7431, opened 2020-02-19
Deprecate the --with-pmi option

rhc54:
Per the developer's meeting, add detection of the deprecated --with-pmi
(and its associated --with-pmi-libdir) configure option and error out
with a polite note of the change in support

### DIFF
--- a/config/ompi_deprecated_options.m4
+++ b/config/ompi_deprecated_options.m4
@@ -1,0 +1,37 @@
+# -*- shell-script -*-
+#
+# Copyright (c) 2020      Intel, Inc.  All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+AC_DEFUN([OMPI_CHECK_DEPRECATED_OPTIONS],[
+    OPAL_VAR_SCOPE_PUSH(with_pmi_given with_pmi_libdir_given)
+
+    AC_ARG_WITH([pmi],
+                [AC_HELP_STRING([--with-pmi(=DIR)],
+                                [*DEPRECATED* Build PMI support, optionally adding DIR to the search path (default: no)])],
+                [with_pmi_given=yes])
+
+    AC_ARG_WITH([pmi-libdir],
+                [AC_HELP_STRING([--with-pmi-libdir=DIR],
+                                [*DEPRECATED* Look for libpmi or libpmi2 in the given directory DIR, DIR/lib or DIR/lib64])],
+                [with_pmi_libdir_given=yes])
+
+    if test "$with_pmi_given" = "yes" || test "$with_pmi_libdir_given" = "yes"; then
+        AC_MSG_WARN([Open MPI no longer supports PMI-1 or PMI-2 libraries.])
+        AC_MSG_WARN([PMIx is now required. Either the internal version or an])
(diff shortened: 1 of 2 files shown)


config/ompi_deprecated_options.m4
```diff
@@ -0,0 +1,37 @@
...
+                [AC_HELP_STRING([--with-pmi(=DIR)],
+                                [Build PMI support, optionally adding DIR to the search path (default: no)])],
+                                [with_pmi_given=yes])
+
+    AC_ARG_WITH([pmi-libdir],
+                [AC_HELP_STRING([--with-pmi-libdir=DIR],
+                                [Look for libpmi or libpmi2 in the given directory DIR, DIR/lib or DIR/lib64])],
+                                [with_pmi_libdir_given=yes])
+
+    if test "$with_pmi_given" = "yes" || test "$with_pmi_libdir_given" = "yes"; then
```
jsquyres: Minor nit: I'd change these tests to:

```suggestion
    if test -n "$with_pmi_given" || test -n "$with_pmi_libdir_given"; then
```

I.e., _any_ value given to these CLI params will trigger this warning.

rhc54: That isn't how AC_ARG_WITH works:
```text
AC_ARG_WITH (package, help-string, [action-if-given], [action-if-not-given])

If the user gave configure the option --with-package or
 --without-package, run shell commands action-if-given. 
```

jsquyres: The user can do:

* `--with-foo`: then `$foo` will be `yes`
* `--without-foo`: then `$foo` will be `no`
* `--with-foo=VALUE`: then `$foo` will be `VALUE`
* do not specify any form of `--with-foo`: then `$foo` will be empty

rhc54: Sigh - you aren't reading the code. I am using the "action-if-given" field to set a variable ```with_pmi_given``` to be ```yes```.  I am _not_ using the value of ```with_pmi```.

jsquyres: Laugh -- you're right, sorry about that.

In my defense, the indenting is wrong on the lines where you have that -- I eyeballed that code and thought it was part of AC_HELP_STRING.  😄

rhc54: okay, just for you - i fixed the indenting to make it more obvious. Wouldn't want to stress that little brain too much!

rhc54:
@jsquyres Ick - this means that ```--with-pmi``` now shows as a `configure` option. I don't see a way to trap it without adding it - do you? At the least, I think we need a big ```DEPRECATED``` in the help message.

Also, I'm seeing ```--with-pmix``` showing up twice in ```configure --help```. We may be picking up the `PRRTE``` configure option as well as the `OMPI` one?

jsquyres:
> @jsquyres Ick - this means that `--with-pmi` now shows as a `configure` option. I don't see a way to trap it without adding it - do you? At the least, I think we need a big `DEPRECATED` in the help message.

Erf.  Ya, I agree -- start the help string with "DEPRECATED" or something else blindingly obviously.

> Also, I'm seeing `--with-pmix` showing up twice in `configure --help`. We may be picking up the `PRRTE` configure option as well as the `OMPI` one?

Mmph.  I don't see an obvious reason why this is happening.  Let me dig.

PR #7432, opened 2020-02-19
Add a -j option to autogen.pl to set AUTOMAKE_JOBS

jjhursey:
* `AUTOMAKE_JOBS` can improve the performance to `autogen.pl`
 * The user can set this envar in the environment before calling
   `autogen.pl` or use the new `-j #` option to set it.

### DIFF
--- a/autogen.pl
+++ b/autogen.pl
@@ -56,6 +56,7 @@ my $platform_arg = 0;
 my $include_arg = 0;
 my $exclude_arg = 0;
 my $force_arg = 0;
+my $automake_jobs;
 
 # Include/exclude lists
 my $include_list;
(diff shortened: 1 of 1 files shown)


jjhursey:
Setting `AUTOMAKE_JOBS` can improve the performance of `autogen.pl` by shaving off about a minute in our experiments (seen below on a 20 core machine):
```
-j    time
---------------
off   5m26.965s
  2   4m48.214s
  4   4m30.321s
  8   4m24.806s
 16   4m19.187s
 20   4m18.771s
```

This optimization has been suggested in the past (see [this](https://www.mail-archive.com/devel@lists.open-mpi.org/msg08574.html) thread) from 2010). To address the concerns raised this PR does the following:
 * Default to not setting this parameter. Per "principle of least surprise"
 * Add a `-j` option to set the value similar to `make -j #`
 * User can always still just set `AUTOMAKE_JOBS` in their environment.

jjhursey:
Note that I flagged the two reviewers because they were actively discussing in the old thread. So would like the :+1: from them before merging.

PR #7434, opened 2020-02-20
fix an issue with configuring with external pmix

hppritcha:
External pmix installs are frequently in non-standard locations and
the path to their shared libraries are not ldconfig'd in because
there may be multiple pmix installs.

The way the configury was set up prior to this patch, the configuration
would fail soon after the PMIX config stuff was called because it
added some pmix lib stuff to the LDFLAGS, resulting in configury tests
for things that require running the configure test to fail.

This patch avoids this problem by resetting the LDFLAGS and LIBS back
to what they were prior to the run of the external PMIX detection.

The CFLAGS setting is left because there are many places in the ompi
and opal source code where pmix_common.h needs to be included.

### DIFF
--- a/config/opal_check_pmi.m4
+++ b/config/opal_check_pmi.m4
@@ -17,6 +17,9 @@
 # Copyright (c) 2014-2018 Research Organization for Information Science
 #                         and Technology (RIST).  All rights reserved.
 # Copyright (c) 2016      IBM Corporation.  All rights reserved.
+# Copyright (c) 2020      Triad National Security, LLC. All rights
+#                         reserved.
+#
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -180,14 +183,17 @@ AC_DEFUN([OPAL_CHECK_PMIX_LIB],[
           [$3
            # add the new flags to our wrapper compilers
            AS_IF([test "$pmix_ext_install_incdir" != "/usr" && test "$pmix_ext_install_incdir" != "/usr/include"],
-                 [pmix_pmix4x_WRAPPER_EXTRA_CPPFLAGS="-I$pmix_ext_install_incdir"])
+                 [pmix_external_WRAPPER_EXTRA_CPPFLAGS="-I$pmix_ext_install_incdir"])
            AS_IF([test "$pmix_ext_install_libdir" != "/usr" && test "$pmix_ext_install_libdir" != "/usr/include"],
-                 [pmix_external_WRAPPER_EXTRA_LDFLAGS="-L$pmix_ext_install_libdir"])
-           pmix_external_WRAPPER_EXTRA_LIBS=-lpmix],
-          [CPPFLAGS=$opal_external_pmix_save_CPPFLAGS
-           LDFLAGS=$opal_external_pmix_save_LDFLAGS
-           LIBS=$opal_external_pmix_save_LIBS
-           $4])
+                 [pmix_external_WRAPPER_EXTRA_LDFLAGS="-L$pmix_ext_install_libdir"
+                  pmix_external_WRAPPER_EXTRA_LIBS="-lpmix"])],
+          [$4])
+
+dnl swap back in original LDFLAGS, LIBS to avoid messing up subsequent configury checks
+dnl don't swap back in orig CFLAGS as there are lots of places where the external pmix
+dnl header file location needs to be known
+    LDFLAGS=$opal_external_pmix_save_LDFLAGS
+    LIBS=$opal_external_pmix_save_LIBS
 
     OPAL_VAR_SCOPE_POP
 ])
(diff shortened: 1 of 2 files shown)


hppritcha:
@artpol84 please see if this fixes your configure issue

config/opal_check_pmi.m4
```diff
@@ -180,14 +183,17 @@ AC_DEFUN([OPAL_CHECK_PMIX_LIB],[
...
-          [CPPFLAGS=$opal_external_pmix_save_CPPFLAGS
-           LDFLAGS=$opal_external_pmix_save_LDFLAGS
-           LIBS=$opal_external_pmix_save_LIBS
-           $4])
+                  pmix_external_WRAPPER_EXTRA_LIBS=-lpmix],
+          [$4])
+
+    AC_SUBST(pmix_external_WRAPPER_EXTRA_LDFLAGS)
+    AC_SUBST(pmix_external_WRAPPER_EXTRA_LIBS)
+dnl swap back in original LDFLAGS, LIBS to avoid messing up subsequent configury checks
```
artpol84: Should we also restore CFLAGS?

hppritcha: I tried that but you end up having to change several makefiles.  Didn't seem worth it.

artpol84:
@hppritcha, I have confirmed that the observed failures (`fail to determine the sizeof(long long)`) are now fixed with this branch.

rhc54:
Be sure to check static builds - you may need to add those flags back in at some point to make it work.

hppritcha:
good point.  things hold together with ```--disable-dlopen``` and ```--enable-static```

config/opal_check_pmi.m4
jsquyres: Yes, it is worth it.  We can't pollute the top-level CPPFLAGS.

You need to set PMIx-related CPPLAGS only where it's relevant to use them.

Sorry.  😦

opal/mca/pmix/external/Makefile.am
```diff
@@ -19,8 +22,10 @@ noinst_LTLIBRARIES = libmca_pmix_external.la
 libmca_pmix_external_la_SOURCES = \
         external.h \
         pmix_external_component.c
+
+libmca_pmix_external_la_LIBADD = $(pmix_external_WRAPPER_EXTRA_LIBS)
 libmca_pmix_external_la_LDFLAGS = \
-    -module -avoid-version
+    -module -avoid-version $(pmix_external_WRAPPER_EXTRA_LDFLAGS)
```
jsquyres: The wrapper flags are for the wrappers.  You probably shouldn't be using them in internal Makefile.am's.

artpol84:
@hppritcha I found that my build was not using the right branch.
I'm re-testing now. Please, wait with merging.

artpol84:
@hppritcha I've double-checked. This PR fixes the issue I had with configure.

jsquyres:
This PR still needs to fix its issue with CPPFLAGS before merging.

awlauria:
I did notice that when I built an external pmix, I did not need to point ompi to the same libevent. If I don't specify `--with-libevent` at the ompi configure line, it'll use the libevent embedded in ompi. So it's possible to be using two different libevent libraries. 

@hppritcha mentioned there used to be a check there for this case - did that go away some how?

rhc54:
Must have gone away - you also need to check for external hwloc and abort if all three (libevent, hwloc, and pmix) are not external if external pmix is requested.

hppritcha:
> Must have gone away - you also need to check for external hwloc and abort if all three (libevent, hwloc, and pmix) are not external if external pmix is requested.

this will be dealt with in another PR.

jsquyres:
"big fat change" isn't much of a commit message.

I'm sorry I didn't understand that the change you were looking to make.  Specifically: if the PMIx CPPFLAGS are used extensively throughout the Makefile.am space, then they _should_ go into the top-level CPPFLAGS.  My (flawed) understanding of your prior comments was that these were component-local flags, but you didn't think it was worth it to do the proper cleanup and didn't mind polluting the top-level CPPFLAGS.  That is apparently not the case.

Specifically: this "big fat change" commit does not seem like the right solution.

hppritcha:
That was a temporary commit just to get stuff up on GitHub so I could work on plane

hppritcha:
I did not know about top level cop flags I’ll use those instead

hppritcha:
@artpol84 when you have a chance could you double check this PR?  I made a few minor changes since you last tested.

hppritcha:
@jsquyres should be adjusted per your comments now

hppritcha:
@jsquyres could you please double check?  I think your concerns are addressed now.

hppritcha:
@artpol84 please retest when you have a chance

PR #7440, opened 2020-02-20
Correctly set baseptr in contiguous shared memory window with local size zero (v3.0.x)

devreal:
The MPI standard mandates that for contiguous shared memory windows the returned `baseptr` can be used to directly calculate the remote addresses on other processes so we cannot return `NULL` if `size == 0`.

Cherry-pick of #7204 to the v3.0.x branch.

(cherry picked from commit 06bbcf4fd63dd184cf22f8bcad007c4b8b991a3c)

### DIFF
--- a/ompi/mca/osc/sm/osc_sm_component.c
+++ b/ompi/mca/osc/sm/osc_sm_component.c
@@ -331,7 +331,7 @@ component_select(struct ompi_win_t *win, void **base, size_t size, int disp_unit
             }
 
             module->sizes[i] = rbuf[i];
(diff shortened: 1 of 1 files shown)


jsquyres:
I think @hjelmn would be better to give a review than me.  😄

jsquyres:
@hjelmn Ping.

bwbarrett:
Decided to pull this into 3.1.x because (1) it's simple and easy to understand and (2) we need to fire up a new MTT run.

PR #7443, opened 2020-02-21
Fix incorrect argument in manual page.

tjahns:
The documentation of MPI_Type_create_hindexed_block has an incorrect type specification:

Only the F2008 version of MPI_Type_create_hindexed_block is described as having an MPI_Aint argument for displacements, but this should also be true of the C and Fortran version.


tjahns:
I'll redo this with the Signed-off-by-line.

gpaulsen:
Redone in https://github.com/open-mpi/ompi/pull/7444

PR #7445, opened 2020-02-21
Save the old ORTE simple tests

rhc54:
Useful when debugging RTE-related issues

Not for inclusion in the tarball - just added to git repo for use by
developers.


rhc54:
Looks like Cray is having problems again.

bot:ompi:retest

rhc54:
Hmmm...Cray says it is done and completed successfully, but the status on this PR isn't being updated because the PR checker isn't getting updated with the completion notice.

rhc54:
I surrender. This PR changes nothing in the code. It simply adds some tests that aren't even compiled nor executed. This is insane that we block for hours on something this mundane.

PR #7446, opened 2020-02-21
Remove "compress" OPAL framework

jsquyres:
This framework is no longer used.


jsquyres:
@artemry-mlnx @artpol84 Per our discussion yesterday, here's a PR that -- as of this writing -- has been queued for 2.25 hours so far.  This is... really, really long.  😢

artemry-mlnx:
@jsquyres 
I see. The CI itself takes 1h 10m - 1h 30m. Additional delay is mainly caused by the situation when there are multiple CI jobs submitted simultaneously (for example, right now there are 4 jobs in the queue: 1 - running, 3 - waiting in the queue).
Currently there is one host configured for CI, taking into account average load - ~10 runs per a day. I can raise a topic about involving additional HW for CI if there are such demands.
Regarding to decreasing 1h+ CI duration - next week we should get new faster building host (this will speed up CI). Moreover I'm reworking the CI script itself - it should speed up the CI as well (I'm going to publish it within next 1-2 weeks).

jsquyres:
Not to sound too much like a jerk, but you speed up that publication timeframe?

This is actually a fairly urgent issue -- yesterday, there were queued up for ***6 hours***.  This is really problematic, because developers are starting to disregard mellanox CI because they cannot wait that long.  It's causing serious logistical problems.

artemry-mlnx:
@jsquyres 
Here're interim Mellanox Open MPI CI optimization results:
1. Optimized Open MPI source checkout (saves 10-20min): https://github.com/open-mpi/ompi/pull/7456 (master), https://github.com/open-mpi/ompi/pull/7457 (v4.0.x)
2. Refactored CI scripts (disabled some useless stuff, saves 30-40 min): https://github.com/mellanox-hpc/jenkins_scripts/pull/93
3. Add more powerful hosts into the Azure self-hosted agent pool (faster CPU + SSD) - WIP
4. (optional) We can disable CI triggering for master/release branches and keep only triggering within PR. This will reduce CI queue load.

With already implemented optimizations (which are under review, PRs submitted) total Mellanox Open MPI CI duration is 35-40min.

artemry-mlnx:
@jsquyres 
UPD: enabled new Melannox self-hosted agent with faster CPU and SSD - CI speedup ~40%.

jsquyres:
@artemry-mlnx That's great -- many thanks!

PR #7447, opened 2020-02-21
Pickup change in check for deleted vs deprecated options and add deprecated ones

rhc54:
Tweak: rename "deprecated" --> "deleted"
We're dealing with CLI options that have been deleted, not
deprecated.

(cherry picked from commit 5aea444)

--------------------

 Deprecate the enable-orterun-prefix-by-default options
Mark the --enable-orterun-prefix-by-default and
--enable-mpirun-prefix-by-default options as deprecated, but continue to
honor them by translating them to the new
--enable-prte-prefix-by-default option.

### DIFF
--- a/config/ompi_deleted_options.m4
+++ b/config/ompi_deleted_options.m4
@@ -8,17 +8,18 @@
 # $HEADER$
 #
 
-AC_DEFUN([OMPI_CHECK_DEPRECATED_OPTIONS],[
-    OPAL_VAR_SCOPE_PUSH(with_pmi_given with_pmi_libdir_given)
+AC_DEFUN([OMPI_CHECK_DELETED_OPTIONS],[
+    OPAL_VAR_SCOPE_PUSH([with_pmi_given with_pmi_libdir_given cxx])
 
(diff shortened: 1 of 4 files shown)


rhc54:
Cray passed again but failed to update AWS, so the PR can't be updated either.

rhc54:
@jsquyres Just an FYI: the way we fork/exec `prte` causes us to automatically force the prefix onto it. So whether we configure with `prefix-by-default` or not, you are going to get the exact same behavior. There is no way around it as we have to ensure we are getting the `prte` generated by the embedded build and not some other `prte` sitting around on the system.

jsquyres:
> @jsquyres Just an FYI: the way we fork/exec `prte` causes us to automatically force the prefix onto it. So whether we configure with `prefix-by-default` or not, you are going to get the exact same behavior. There is no way around it as we have to ensure we are getting the `prte` generated by the embedded build and not some other `prte` sitting around on the system.

That's probably ok, but worth documenting somewhere.

rhc54:
Let Mellanox catch up another day. This is passing everywhere else.

Issue #7429, opened 2020-02-19
Error when writing file on more than one node


romainhild:
I use Open MPI v4.0.2 from the tar.gz on a cluster running Ubuntu 16.04 with 6 compute nodes (24 cores each). The jobs are managed by slurm.

I have the following error when I try to write some files on disk when running on multiple nodes:
```
[atlas6:242817] mca_sharedfp_individual_file_open: Error during datafile file open
```
This error is displayed for each proc of one of the two nodes I use, not the other one.

The problem is that it does not happen all the times I launch the same program.
I could not find a way to reproduce the issue consistently.

Is there anyone that have an idea about this ?

edgargabriel:
what filesystem are you using, and would it be possible to provide a reproducer for this problem?

romainhild:
the compute nodes are in ext4 and where I'm trying to write is in nfs.
And I'm sorry but I'm still trying to find a way to reproduce in a consistent way the problem, as I said, it does not happen every time...

edgargabriel:
is your code actually using shared file pointer operations? If not, it might be easiest to tell Open MPI to not try to initialize a sharedfp component at all

romainhild:
We use HDF5 to export some fields and we don't know exactly what is done by it.
But otherwise, I don't think we use them.

prudhomm:
@edgargabriel we don't use shared file pointers anymore only explicit offsets version for MPI IO. is there a way to disable the sharedfp component (`individual`)?

edgargabriel:
@prudhomm it turns out its not that simple. The individual sharedfp component is only loaded, if the other sharedfp components disqualify themselves, i.e. its the last resort. Now, in the 3.x series ompi used to be able to operate without a sharedfp component, but that is not the case in the 4.x series. So if we would disable the individual component, your application would segfault.

I am currently looking into this, from both aspects: 
1. why does the sharedfp_individual_file_open raise an error ( which I can not yet reproduce), and 
2. whether its possible to fix the sequence of running without a sharedfp component. 

Will try to keep you posted, I hope to have a fix for at least one of the two items very soon.

edgargabriel:
actually, it might be worth a trial: can you see whether adding the following parameter to your command line fixes the issue:

mpirun --mca sharedfp ^individual <whatever other arguments you set>

prudhomm:
Thanks a lot. We will test asap with @romainhild

edgargabriel:
[...]

PR #7448, opened 2020-02-21
sharedfp/individual: defer error when not being able to open datafile

edgargabriel:
This commit changes the behavior of the individual sharedfp component. If
the component cannot create either the datafile or the metadatafile during File_open,
no error is being raised going forward. This allows applications that do not use shared
file pointer operations to continue execution without any issue.

If the user however subsequently calls MPI_File_write_shared or similar operations, an error
will be raised.

Fixes issue #7429


hppritcha:
@edgargabriel should this be PR'd back to release branches?  If so, which ones?

edgargabriel:
@hppritcha we should definitely bring it to the 4.0.x series, the behavior was slightly different in the 3.x series, so this patch does not apply there. I *suspect* that the issue could however also arise in the 3.x series.

edgargabriel:
bot:retest

PR #7451, opened 2020-02-21
Fix Slurm process name

rhc54:
Ensure that we truncate the local jobid to 15-bits


rhc54:
@artpol84 I believe this will fix the Slurm direct launch issue.

PR #7462, opened 2020-02-24
sharedfp/individual: defer error when not being able to open datafile

edgargabriel:
This commit changes the behavior of the individual sharedfp component. If
the component cannot create either the datafile or the metadatafile during File_open,
no error is being raised going forward. This allows applications that do not use shared
file pointer operations to continue execution without any issue.

If the user however subsequently calls MPI_File_write_shared or similar operations, an error
will be raised.

Fixes issue #7429

(cherry picked from commit df6e3e503aee6954807a6bdc3a73dfa9a7d030af)


hppritcha:
@raafatfeki please review promptly as we want to get an rc out tomorrow or Wednesday this week.

raafatfeki:
@hppritcha Checking it right now.

PR #7464, opened 2020-02-24
Disabled Mellanox Open MPI per-commit CI (as redundant).

artemry-mlnx:
The CI is triggered only upon a PR creation or by special PR comments.

Signed-off-by: Artem Ryabov <artemry@mellanox.com>


jsquyres:
ok to test

PR #7466, opened 2020-02-25
pml/ucx: Add MCA_PML_BASE_FLAG_REQUIRE_WORLD flag.

awlauria:
In MPI_Init(), the UCX pml calls PMIx_Get()
on rank 0 to verify that it is using the UCX pml in
mca_pml_ucx_add_proc(). However, without this flag, remote ranks
will be performing this on their local rank 0, which doesn't
have this information - so they will be stuck until PMIx_Get()
hits a timeout.

This patch should give the ranks the correct information,
and most importantly eliminate the time spent idling in MPI_Init()
when using UCX.

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx_component.c
+++ b/ompi/mca/pml/ucx/pml_ucx_component.c
@@ -98,6 +98,8 @@ mca_pml_ucx_component_init(int* priority, bool enable_progress_threads,
         return NULL;
     }
 
+    ompi_pml_ucx.super.pml_flags |= MCA_PML_BASE_FLAG_REQUIRE_WORLD;
+
     *priority = ompi_pml_ucx.priority;
     return &ompi_pml_ucx.super;
 }


yosefe:
@karasevb @artpol84 does it mean this patch disables direct modex?

karasevb:
@yosefe yes, it seems that this flag will, in any case, performed a full modex for UCX EPs.

yosefe:
@awlauria then it's a problem since since this disables an important functionality in UCX

jsquyres:
Also, the word "rank" is not correct here.  Please use "process" to distinguish what you mean.  Thanks!

awlauria:
@jsquyres  will do.

@yosefe  I see two other ways around this other than setting this flag:

- Modify `mca_pml_base_select()` to call `mca_pml_base_pml_selected()` on process 0 on each node. I think this should be fine and will probably work, but I am not familiar with the ramifications of it.

- Possibly eliminating this PMIx_Get() ucx - seems dangerous, but...?

-  Decreasing the timeout PMix to mask this, but it won't fix the core issue.

Open to other ideas, but just what I saw when debugging this - a simple 2 process + multinode  mpi-hello program takes about a minute on ompi/pmix master right now due to this issue.

awlauria:
For what it is worth, `ob1` does not make this check of the pml as far as I can tell.

bosilca:
The ob1 PML will make this decision dynamically, depending on the capabilities of the selected BTLs (flag MCA_BTL_FLAGS_SINGLE_ADD_PROCS). For the UCX PML, as it supports lazy setup, the flag should not be present, and the first add_procs should be only done with locally known processes. I will check that in **ompi_mpi_init**, the function **ompi_proc_get_allocated** always returns a subset of the world.

jjhursey:
Maybe @karasevb @artpol84 can take the lead on addressing the UCX slowdown on master. It seems to be cycling in the dmodex - so maybe this needs to be chased down inside of PMIx master.

Austen confirmed that this slowdown behavior does not happen on OMPI v4.0.x and PMIx 3.1.5.

karasevb:
@awlauria could you give a little clarification on this issue? What are the steps to reproducing the timeout you mentioned?

awlauria:
@karasevb 

take ompi/pmix master, and run an MPI_Init()/Finalize() program across node:

`mpirun -np 2 --hosts X:1,Y:1 ./hello`

If you enable PMIX_DEBUG you'll see it cycling over this:

```[f10n18:128686] HASH:FETCH rank 5 key pml.base.2.0
[f10n18:128686] HASH:FETCH data for key pml.base.2.0 not found
[f10n18:128686] HASH:FETCH rank 5 key pml.base.2.0
[f10n18:128686] HASH:FETCH proc data for rank 5 not found
[f10n18:128686] HASH:FETCH rank 5 key pml.base.2.0
...
```

rhc54:
Remember that PMIx master has changed the dmodex behavior to more closely align with the PMIx Standard. I suspect you may have to update whatever UCX and/or OMPI is doing in that regard.

jjhursey:
This [thread](https://github.com/openpmix/openpmix/issues/1586#issuecomment-581070965) may or may not be related. But it sounds like it might be something to look at when investigating this.

awlauria:
Testing this by raising the priority of `ob1` to be higher than ucx - it also happens in that case. The remote ranks are stuck in the same PMIx_Get() loop and will block until released. And explicitly requesting ob1 will avoid it, like ucx.

I think a more general solution is needed - this isn't just ucx.

awlauria:
Closing this. I opened issue #7475 for any further discussion.

PR #7470, opened 2020-02-25
btl/sm: remove the deprecation-notice shell

jsquyres:
The SM BTL was effectively removed a long time ago.  All that was left
was a shell that warned people if they tried to use the SM BTL.  For
v5.0, we plan to finally remove this ancient shell (and possibly
replace it with vader).


jsquyres:
@hjelmn Please review (can't merge without a review 😉)

PR #7480, opened 2020-02-26
Remove the pml check_selected code.

### LINKED ISSUE #7475: Simple multiprocess mpirun taking > 1 minute on master
## Background information

Reproducer:

`$. mpirun -np 2 --host X:1,Y:1 ./hello`

If you enable PMIX_DEBUG you'll see it cycling over this:

```
[f10n18:128686] HASH:FETCH data for key pml.base.2.0 not found
[f10n18:128686] HASH:FETCH rank 1 key pml.base.2.0
[f10n18:128686] HASH:FETCH proc data for rank 5 not found
[f10n18:128686] HASH:FETCH rank 1 key pml.base.2.0
...
```

It'll take over 1 minute to complete until it gets unwedged trying to fetch the pml base.

What happening is the following:

- If the pml selection logic determines that there > 1 viable pml, it will have __global rank 0__ expose it using `PMIx_Put(PMIX_WORLD...)`.
- At `mca_pml_*_add_procs()`  time, if > 1 pml was determined viable during selection, it will attempt to query its __local rank 0__ to see what pml it used to verify.
- Since only the __global rank 0__ exposed its pml, all ranks outside of the first node will become wedeged in the PMIx_Get() call until it gives up.
- If only one pml is deemed viable (as is the case when using --mca pml ...), all of this is avoided.

- I confirmed that ucx and pml can hit this bug, haven't checked others.
  - to hit with ob1, I manually raised its priority to be higher than ucx.
- Neither will hit it if explicitly asked for at mpirun time (`--mca pml ucx`/`--mca pml ob1`). Because there is only one viable pml, neither will attempt to query.

An easy fix that could work for all pmls is to register the pml on the 0th local process on each node. That way all processes on their local node has a reference to query and check the pml being used. 

I'm not sure of the performance impact of that, but testing ompi/pmix master with and without the change showed no discernible performance difference on a MPI_Init/Finalize program.

awlauria:
The current logic is broken - this only works from the
node which has rank 0 on it. All other processes will fail
trying to fetch the data from rank 0 on a PMIx_Get().

Since this has been broken for what is probably a long time with no
complaints, it is probably safe to remove completely
and gain back some of the lost performance.

Fixes #7475

### DIFF
--- a/ompi/mca/pml/base/base.h
+++ b/ompi/mca/pml/base/base.h
@@ -53,12 +53,6 @@ OMPI_DECLSPEC extern mca_base_framework_t ompi_pml_base_framework;
 OMPI_DECLSPEC  int mca_pml_base_select(bool enable_progress_threads,
                                        bool enable_mpi_threads);
 OMPI_DECLSPEC  int mca_pml_base_progress(void);
-    /* share in modex the name of the selected component */
-OMPI_DECLSPEC int mca_pml_base_pml_selected(const char *name);
-    /* verify that all new procs are using the currently selected component */
-OMPI_DECLSPEC int mca_pml_base_pml_check_selected(const char *my_pml,
-                                                  struct ompi_proc_t **procs,
-                                                  size_t nprocs);
 
 OMPI_DECLSPEC int mca_pml_base_finalize(void);
 
--- a/ompi/mca/pml/base/pml_base_select.c
+++ b/ompi/mca/pml/base/pml_base_select.c
@@ -44,8 +44,6 @@ typedef struct opened_component_t {
   mca_pml_base_component_t *om_component;
 } opened_component_t;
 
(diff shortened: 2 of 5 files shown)


rhc54:
Just a thought: there may be occasions where someone is having a problem like what @bwbarrett said on the the issue. Would it make sense to provide a "PML debug" MCA param that would:

* push the selected PML into the modex
* force a full modex to be performed
* check PML selection by each proc and generate a big error if it doesn't match

This would fix the current problem but still provide a debugging capability if explicitly requested.

awlauria:
@rhc54 we could do that. But as a counter-point, the user can enable pml base logging in this situation which would give them the component selected for each process. While not nearly as pretty, it would work to debug a rare situation like this.

PR #7487, opened 2020-02-28
Make sure the PML selection is consistent across the world.

bosilca:
Temporary solution for the PML inconsistency issue discussed in #7475.
This patch address 2 things: first it make the PMIx key optional so that
if we are not in a full modex mode we don't do a direct modex, and
second it get the PML info from the vpid 0 instead of from the local
rank.


ibm-ompi:

 Gist: https://gist.github.com/bf3de9a9757da081e6d8030ea14fa8a1

rhc54:
bot:ibm:scale:retest

PR #7488, opened 2020-02-29
Update the PRRTE and PMIx pointers

rhc54:


rhc54:
Included in this PR:
   -bind-to-core|--bind-to-core 
                         Bind processes to cores
   -bind-to-socket|--bind-to-socket 
                         Bind processes to sockets
   -bycore|--bycore      Whether to map and rank processes round-robin by
                         core
   -bynode|--bynode      Whether to map and rank processes round-robin by
                         node
   -byslot|--byslot      Whether to map and rank processes round-robin by
                         slot
   -cpu-list|--cpu-list <arg0>  
                         List of processor IDs to bind processes to
                         [default=NULL]
   -cpu-set|--cpu-set <arg0>  
                         Comma-separated list of ranges specifying logical
                         cpus allocated to this job [default: none]
   -cpus-per-proc|--cpus-per-proc <arg0>  
                         Number of cpus to use for each process [default=1]
   -cpus-per-rank|--cpus-per-rank <arg0>  
                         Synonym for cpus-per-proc
   -nolocal|--nolocal    Do not run any MPI applications on the local node
   -nooversubscribe|--nooversubscribe 
                         Nodes are not to be oversubscribed, even if the
                         system supports such operation
   -npernode|--npernode <arg0>  
                         Launch n processes per node on all allocated nodes
   -npersocket|--npersocket <arg0>  
                         Launch n processes per socket on all allocated
                         nodes
   -pernode|--pernode    Launch one process per available node
   --ppr <arg0>          Comma-separated list of number of processes on a
                         given resource type [default: none]
   -use-hwthread-cpus|--use-hwthread-cpus 
                         Use hardware threads as independent cpus

Also includes:
Drop support for aggregated single char cmd line options 
No longer parse "-abc" as "-a -b -c" - instead treat it as "-abc"
Check for deprecated single-dash OMPI options
Warn and show corrected option


There are a fair number of problems where the map/bin/rank of the prior job throws off the next job or become the defaults for the next job.
    --cpus-per-rank issue
        --map-by :pe=X works, and X becomes the default for the next job (even when not specified)
        --map-by core:pe=X fails to map
    --nolocal not implemented in the backend
    --use-hwthread-cpus
        Should we also set --map-by hwthread? Probably.
        There is some investigation about intended behavior here

All of these CLI options are represented in that PR:

PR #7491, opened 2020-02-29
Tweak the C++ binding deprecation check

rhc54:
Some of us have platform files that expressly disabled C++ support.
While it is true that v5 no longer supports C++ and thus no longer needs
us to disable it, there seems no reason to make us create platform files
that differentiate based on OMPI version just for that reason.

So if someone asks to "disable" the no-longer-existing support, just
ignore it.

### DIFF
--- a/ompi/mca/pml/base/pml_base_frame.c
+++ b/ompi/mca/pml/base/pml_base_frame.c
@@ -16,6 +16,7 @@
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2018 IBM Corporation. All rights reserved.
+ * Copyright (c) 2020      Intel, Inc.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


rhc54:
@jsquyres Would you mind taking another peek at this?

ompi/mca/pml/base/pml_base_frame.c
```diff
@@ -214,7 +216,6 @@ static int mca_pml_base_open(mca_base_open_flag_t flags)
         if( (NULL == default_pml || NULL == default_pml[0] ||
              0 == strlen(default_pml[0])) || (default_pml[0][0] == '^') ) {
             opal_pointer_array_add(&mca_pml_base_pml, strdup("ob1"));
-            opal_pointer_array_add(&mca_pml_base_pml, strdup("yalla"));
```
jsquyres: This is definitely a correct fix, but it begs the question: why on earth do we have a hard-coded list of PMLs here?

@hjelmn Don't we have a list of autogen/configure-found PMLs somewhere that we can use instead?

PR #7492, opened 2020-02-29
Fix singleton operation

rhc54:
OpenPMIx fills in a variety of info when it detects that we are in
singleton mode. Best way of detecting it is to look for the "singleton"
at the beginning of the returned nspace.

Make the modex recvs optional so we don't bounce up to the server and
then to the host trying to retrieve job-level info that must be given to
us at job start.


ibm-ompi:

 Gist: https://gist.github.com/b03ce58d1ee5f4af4c0bbb39755c18f1

rhc54:
bot:ibm:scale:retest

Issue #10, opened 2014-10-01
MPI_Pack_external_size is returning the wrong values in 64-bit applications


ompiteam:
We have a program that tests for the size returned from MPI_Pack_external_size with the external32 data representation.  It should return the same value for both 32-bit and 64-bit applications, but it is returning different values.

```
 burl-ct-v40z-0 65 =>mpicc ext32.c -o ext32
"ext32.c", line 105: warning: shift count negative or too big: << 32
 burl-ct-v40z-0 66 =>mpirun -np 2 ext32
First test passed
Second test passed
Third test passed
ext32: PASSED
```

```
 burl-ct-v40z-0 67 =>mpicc -xarch=amd64 ext32.c -o ext32_amd64
 burl-ct-v40z-0 68 =>mpirun -np 2 ext32_amd64 
First test passed
Second test failed. Got size of 80, expected 40
Third test failed. Got size of 6400, expected 3200
[burl-ct-v40z-0:13864] *** An error occurred in MPI_Pack_external
[burl-ct-v40z-0:13864] *** on communicator MPI_COMM_WORLD
[burl-ct-v40z-0:13864] *** MPI_ERR_TRUNCATE: message truncated
[burl-ct-v40z-0:13864] *** MPI_ERRORS_ARE_FATAL (goodbye)
 burl-ct-v40z-0 69 =>
```

bosilca:
few years old and no replicator.

PR #7497, opened 2020-03-03
Fix segv in btl/vader.

awlauria:
Keep track of the connected procs in vader_add_procs().
Otherwise, the same rank will reconnect the same shmem
segment (rank 0+...) multiple times instead of the next
one as intended.

### DIFF
--- a/opal/mca/btl/vader/btl_vader.h
+++ b/opal/mca/btl/vader/btl_vader.h
@@ -125,6 +125,7 @@ struct mca_btl_vader_component_t {
     char *my_segment;                       /**< this rank's base pointer */
     size_t segment_size;                    /**< size of my_segment */
     int32_t num_smp_procs;                  /**< current number of smp procs on this host */
+    int32_t local_rank;                     /**< current rank index at add_procs() time */
     opal_free_list_t vader_frags_eager;     /**< free list of vader send frags */
     opal_free_list_t vader_frags_max_send;  /**< free list of vader max send frags (large fragments) */
     opal_free_list_t vader_frags_user;      /**< free list of small inline frags */
--- a/opal/mca/btl/vader/btl_vader_component.c
+++ b/opal/mca/btl/vader/btl_vader_component.c
@@ -560,6 +560,8 @@ static mca_btl_base_module_t **mca_btl_vader_component_init (int *num_btls,
     /* no fast boxes allocated initially */
     component->num_fbox_in_endpoints = 0;
 
+    component->local_rank = 0;
+
     mca_btl_vader_check_single_copy ();
 
     if (MCA_BTL_VADER_XPMEM != mca_btl_vader_component.single_copy_mechanism) {
--- a/opal/mca/btl/vader/btl_vader_module.c
+++ b/opal/mca/btl/vader/btl_vader_module.c
@@ -349,7 +349,7 @@ static int vader_add_procs (struct mca_btl_base_module_t* btl,
         }
     }
 
-    for (int32_t proc = 0, local_rank = 0 ; proc < (int32_t) nprocs ; ++proc) {
+    for (int32_t proc = 0; proc < (int32_t) nprocs; ++proc) {
         /* check to see if this proc can be reached via shmem (i.e.,
            if they're on my local host and in my job) */
         if (procs[proc]->proc_name.jobid != my_proc->proc_name.jobid ||
@@ -367,8 +367,10 @@ static int vader_add_procs (struct mca_btl_base_module_t* btl,
         }
 
         /* setup endpoint */
-        peers[proc] = component->endpoints + local_rank;
-        rc = init_vader_endpoint (peers[proc], procs[proc], local_rank++);
+        int rank = opal_atomic_fetch_add_32(&component -> local_rank, 1);
+           
+        peers[proc] = component->endpoints + rank;
+        rc = init_vader_endpoint (peers[proc], procs[proc], rank);
         if (OPAL_SUCCESS != rc) {
             break;
         }


bosilca:
I don't think this is correct. The local procs need to be indexed with the local rank (which is small) and not with the jobid vpid (which might be very large).

awlauria:
Right now on ompi master, a simple:

`mpirun -np 2 ./1sided` test is failing in mtt. This may not fix that specific failure, but it does fix a segv on my local box when running the above. What appears to be happening is that rank 0 is not creating a shmem segment with rank 1, resulting in an eventual segv in `mca_btl_vader_get_cma()`

ibm-ompi:

 Gist: https://gist.github.com/8dd02338c2860498fb7cb5bca1a7d5e4

ibm-ompi:

 Gist: https://gist.github.com/9f6078ab931657f81dff0d75bedc46b8

awlauria:
@bosilca ah yes, you are correct, was thinking that was a local rank..How is this? Keeps the same logic, but preserves the value of `local_procs` between calls so we aren't reconnecting the same shmem segment over and over. Not sure how important thread safety is here, so I added it in anyway.

```-    for (int32_t proc = 0; proc < (int32_t) nprocs ; ++proc) {
+    static int32_t local_rank = 0;
+
+    bool using_threads = opal_using_threads();
+
+    for (int32_t proc = 0; proc < (int32_t) nprocs; ++proc) {
         /* check to see if this proc can be reached via shmem (i.e.,
            if they're on my local host and in my job) */
         if (procs[proc]->proc_name.jobid != my_proc->proc_name.jobid ||
@@ -367,8 +371,17 @@ static int vader_add_procs (struct mca_btl_base_module_t* btl,
         }
 
         /* setup endpoint */
-        peers[proc] = component->endpoints + procs[proc]->proc_name.vpid;
-        rc = init_vader_endpoint (peers[proc], procs[proc], procs[proc]->proc_name.vpid);
+        int rank = local_rank;
+        if(OPAL_UNLIKELY(using_threads)) {
+          rank = opal_atomic_fetch_add_32(&local_rank, 1);
+        }
+           
+        peers[proc] = component->endpoints + rank;
+        rc = init_vader_endpoint (peers[proc], procs[proc], rank);
+        if(OPAL_LIKELY(false == using_threads)) {
+          local_rank++;
+        }
+
```

bosilca:
I will always use opal_atomic_fetch_add_32  as this part of the code is not really that performance critical.

Instead of making this the local_rank static I would add it to the BTL, so that we can have multiple vader BTL in same time (once we decide to support shared memory between spawnees or between local processes in different sessions).

rhc54:
> Instead of making this the local_rank static I would add it to the BTL, so that we can have multiple vader BTL in same time (once we decide to support shared memory between spawnees or between local processes in different sessions).

Probably safer to use `node_rank` for this purpose as it will prevent collisions. Local rank is only within the current job. Node rank is unique across all jobs on the node

bosilca:
I though we had some info on the job related to local rank, but an 'mpirun env' didn't show anything valuable so I discarded that idea.

Anyway, in this particular instance we do not need we know the process name (as it is guaranteed that a peer will not be add_proc multiple time), so I will take the simplest route, one that does not make us dependent on information provided by an external software.

rhc54:
> one that does not make us dependent on information provided by an external software

Not sure I understand. We already do a modex recv to get the local rank and the number of local peers - all I suggested is that we incorporate the node rank to avoid collisions. We require that any RM provide us with that info, and now that OMPI is locked to PMIx, they all meet that requirement. So getting the node rank doesn't invoke any added dependency.

rhc54:
@hjelmn Would you mind taking a look at this PR?

Nathan and I had done some work a little while ago that enabled vader communication between procs of different jobs - e.g., comm_spawn parents and their children. I'd like to ensure we aren't doing anything here that breaks that capability.

bosilca:
While such a capability was on our wishlist for a while, OMPI never supported it. Indeed, according to opal/mca/btl/vader/btl_vader_module.c line 355, different jobid are prevented to connect to each other, and that code was untouched for the last 6 years.

rhc54:
No, Nathan and I enabled it a short while back - however it may still be on the sessions work. All I know is that it was put into play and verified by at least one user, if only on a branch.

@hjelmn Can you speak to this?

awlauria:
Little late notice - I updated the pr per your suggestions @bosilca . 

If this change will step on an upcoming features toes, I can try to re-work it to not do so.

bosilca:
We should not delay a bug fix for a future features (especially one that does not have a PR yet). When the new future will become public / ready to be PR we will reassess and if needed find a more suitable solution.

awlauria:
bot:ibm:retest

hjelmn:
Wait. What? How did this break? This fix would only change anything if add_procs is called multiple times on btl/vader. The version of Vader in master does not support this. Why is add_procs being called multiple times in this scenario? Is pml/ob1 not in use?

hjelmn:
The correct fix is to ensure add_procs is called on the btl's for all local processes.

hjelmn:
In all cases.

rhc54:
@hjelmn Can you also please address my question regarding comm_spawn parent/child vader support?

awlauria:
@hjelmn in this case it is osc /rdma - not pml/ob1, adding each local endpoint one at a time:

```#2  0x00002000013e5b20 in vader_add_procs (btl=0x200001410230 <mca_btl_vader>, nprocs=1, 
    procs=0x7fffccdd37f0, peers=0x7fffccdd37e8, reachability=0x0) at btl_vader_module.c:336
#3  0x00002000013b3270 in mca_bml_r2_add_proc (proc=0x16cad9d0) at bml_r2.c:406
#4  0x0000200002866f30 in mca_bml_base_get_endpoint (proc=0x16cad9d0)
    at ../../../../ompi/mca/bml/base/base.h:72
#5  0x0000200002869e7c in ompi_osc_rdma_query_btls (comm=0x2000002d1ed0 <ompi_mpi_comm_world>, 
    btl=0x0) at osc_rdma_component.c:860
#6  0x0000200002868330 in ompi_osc_rdma_component_query (win=0x171ae0b0, base=0x7fffccdd3b00, 
    size=1000000, disp_unit=4, comm=0x2000002d1ed0 <ompi_mpi_comm_world>, 
    info=0x2000002da820 <ompi_mpi_info_null>, flavor=1) at osc_rdma_component.c:380
#7  0x00002000002413a4 in ompi_osc_base_select (win=0x171ae0b0, base=0x7fffccdd3b00, 
    size=1000000, disp_unit=4, comm=0x2000002d1ed0 <ompi_mpi_comm_world>, 
    info=0x2000002da820 <ompi_mpi_info_null>, flavor=1, model=0x7fffccdd3af0)
    at base/osc_base_init.c:57
#8  0x000020000010afa0 in ompi_win_create (base=0x200013140010, size=1000000, disp_unit=4, 
    comm=0x2000002d1ed0 <ompi_mpi_comm_world>, info=0x2000002da820 <ompi_mpi_info_null>, 
    newwin=0x100201c8 <win>) at win/win.c:246
#9  0x00002000001cc648 in PMPI_Win_create (base=0x200013140010, size=1000000, disp_unit=4, 
    info=0x2000002da820 <ompi_mpi_info_null>, comm=0x2000002d1ed0 <ompi_mpi_comm_world>, 
    win=0x100201c8 <win>) at pwin_create.c:81
#10 0x0000000010003020 in main_test_fn (comm=0x2000002d1ed0 <ompi_mpi_comm_world>, tid=1)
    at 1sided.c:408
#11 0x0000000010004ae4 in runtest (comm=0x2000002d1ed0 <ompi_mpi_comm_world>, tid=1)
    at 1sided.c:762
#12 0x0000000010004b58 in main () at 1sided.c:772```

hjelmn:
Yup. That is not supported. There was supposed to be code somewhere that ensured r2 had all the local ranks. That is the break. This PR just papers over the real bug.

hjelmn:
@rhc54 I can't remember if I ever added the comm spawn support for vader. I think there are changes in the sessions branch but I can not recall what they were. Was going to take a look in the next week or so.

awlauria:
When I explored doing that it seemed to be a large refactor. osc/rdma is getting one endpoint at a time which has not changed in years, so r2 is calling vader_add_procs() one at a time (also not changed in years). Honestly I'm not sure how this ever worked for osc/rdma.

It might be due for a refactor all the way up to the rdma layer if this is unintended functionality - but I'm not sure, have to dig a bit deeper. When I get some time I can explore how to refactor to get it to work the way it was intended if nobody beats me to it. In the meantime this patch fixes a segv, and on some machines (mtt) it fixes a wrong answer for some 1sided tests.

bosilca:
The root cause might be the recent changes in the modex, which then stopped adding all local procs via the PML early in the process (so that OSC now has to add them directly instead of just getting them via the ompi_proc). Anyway, this change allows for vader add_proc to be called as many time as necessary, even with a single peer at the time, with as far as I see no negative side effect.

hjelmn:
@bosilca Need to verify the ordering. It might be working now because the endpoints are added in the same order on all processes. What happens if the order if different on one ore more processes. At some point that would have caused problems for btl/vader but I can't remember if I fixed that.

bosilca:
@hjelmn I thought about that when answering above, I checked the vader code and I can't see anything that uses the local rank as an index. All peers address the same target process concurrently through a single shared message fifo attached to the local ep, and a message source is computed back in the PML using the OB1 header info (comm id + rank). I think we're good.

PR #7503, opened 2020-03-04
pml/ucx: Warn if UCX is built in a single-threaded mode

artpol84:

### DIFF
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -289,8 +289,8 @@ int mca_pml_ucx_init(int enable_mpi_threads)
     if (enable_mpi_threads && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
         /* UCX does not support multithreading, disqualify current PML for now */
         /* TODO: we should let OMPI to fallback to THREAD_SINGLE mode */
-        PML_UCX_VERBOSE(1, "UCP worker does not support MPI_THREAD_MULTIPLE. "
-                           "PML UCX could not be selected");
+        PML_UCX_WARN("UCP worker does not support MPI_THREAD_MULTIPLE. "
+                     "PML UCX could not be selected");
         rc = OMPI_ERR_NOT_SUPPORTED;
         goto err_destroy_worker;
     }
--- a/ompi/mca/pml/ucx/pml_ucx.h
+++ b/ompi/mca/pml/ucx/pml_ucx.h
@@ -26,7 +26,7 @@
 #define PML_UCX_ASSERT  MCA_COMMON_UCX_ASSERT
 #define PML_UCX_ERROR   MCA_COMMON_UCX_ERROR
 #define PML_UCX_VERBOSE MCA_COMMON_UCX_VERBOSE
-
+#define PML_UCX_WARN    MCA_COMMON_UCX_WARN
 
 typedef struct mca_pml_ucx_module           mca_pml_ucx_module_t;
 typedef struct pml_ucx_persistent_request   mca_pml_ucx_persistent_request_t;


ompi/mca/pml/ucx/pml_ucx.c
```diff
@@ -289,8 +289,8 @@ int mca_pml_ucx_init(int enable_mpi_threads)
     if (enable_mpi_threads && (attr.thread_mode != UCS_THREAD_MODE_MULTI)) {
         /* UCX does not support multithreading, disqualify current PML for now */
         /* TODO: we should let OMPI to fallback to THREAD_SINGLE mode */
-        PML_UCX_VERBOSE(1, "UCP worker does not support MPI_THREAD_MULTIPLE. "
-                           "PML UCX could not be selected");
+        PML_UCX_WARN("UCP worker does not support MPI_THREAD_MULTIPLE. "
```
yosefe: i think we had some objections for this from MPI community: there should be a warning if no MPI component can support MT. Otherwise, there should be a graceful fallback.

artpol84:
I experienced an issue when I was trying to use UCX pml in the multi-threaded mode.
The message wasn't clear enough to understand the problen:
```
No components were able to be opened in the pml framework.

This typically means that either no components of this type were
installed, or none of the installed components can be loaded.
Sometimes this means that shared libraries required by these
components are unable to be found/loaded.

  Host:      clx-hercules-034
  Framework: pml
[clx-hercules-034:03868] PML ucx cannot be selected
[mngx-orion-01:19516] 1 more process has sent help message help-mca-base.txt / find-available:none found
[mngx-orion-01:19516] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

With the fix, the error is more articulated:
```
[clx-hercules-034:15409] ../../../../../ompi/mca/pml/ucx/pml_ucx.c:289  Warning: UCP worker does not support MPI_THREAD_MULTIPLE. PML UCX could not be selected
No components were able to be opened in the pml framework.

This typically means that either no components of this type were
installed, or none of the installed components can be loaded.
Sometimes this means that shared libraries required by these
components are unable to be found/loaded.

  Host:      clx-hercules-034
  Framework: pml
[clx-hercules-034:15409] PML ucx cannot be selected
[clx-hercules-035:18213] ../../../../../ompi/mca/pml/ucx/pml_ucx.c:289  Warning: UCP worker does not support MPI_THREAD_MULTIPLE. PML UCX could not be selected
[mngx-orion-01:26640] 1 more process has sent help message help-mca-base.txt / find-available:none found
[mngx-orion-01:26640] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

yosefe:
is it possible to push this warning somehow into the "[clx-hercules-034:03868] PML ucx cannot be selected" message?
like "[clx-hercules-034:03868] PML ucx cannot be selected because it does not support MT"

artpol84:
This comes from pml/base.
We can only do this if we provide some string output from the pml_init callback.

IMO if users built with UCX they probably intend to use it. Though I don't have all of the contexts here.

artpol84:
My further 2 cents

It is very confusing when you explicitly select UCX pml and it doesn't work.
It's even more confusing when you have ring_c working, but not the app you are running.
I was building everything myself and I can tell you I couldn't match those facts and had to GDB it.
The regular user won't be able to solve it at all.

This error, in most cases, will indicate that OMPI is misconfigured. If you don't want a warning - it is possible to disable UCX component.

yosefe:
@artpol84 if users have multiple PMLs (and don't select UCX explicitly) this warning will also be printed, right?

artpol84:
Yes

yosefe:
> Yes

This is the reason OMPI community rejected such warning when we tried to added it a while back..

jladd-mlnx:
@artpol84  - See https://github.com/open-mpi/ompi/pull/6209 for historical context.

rhc54:
You could  check to see if UCX was specifically requested and only emit it then. It is just an MCA param, so you can always get the value of the "pml" param and check it.

jsquyres:
@artpol84 You're right that the error messages you initially got probably weren't 100% clear.  Got any PML-neutral suggestions to make them better?

artpol84:
One option could be that PML returns the string error description from the init callback (if it wants to).
Then in case no PML was found, PML base can report the errors from all PMLs attempted in this error message.

Other way would be to follow @rhc54 advice.
However this wouldn’t work if, say, OMPI was build with “disable-building “ ob1 and all other PMLs except UCX. Then UCX would be implicitly selected without pml option.
In this case, maybe PML can pass the flag to the init callback saying “this is the only component I consider, warn badly of you are going to fail”

artpol84:
> One option could be that PML returns the string error description from the init callback (if it wants to).
Then in case no PML was found, PML base can report the errors from all PMLs attempted in this error message.

The other incarnation of this way would be to introduce one more callback "get_fail_reason", that PML may invoke to get a string representation of a failure reason. If it's NULL, the component doesn't wish to provide one and PML base will skip it.

This might apply to other frameworks though, but can be implemented on per-need basis.

yosefe:
@can we check if ucx is the only PML which was both (built and selected by mca) ?
for example that it's the only element in the list of PML frameworks?

artpol84:
I’ll check that out

lanl-ompi:

gpaulsen (approved):

Issue #7504, opened 2020-03-05
fi_close(domain) EBUSY in mca_btl_ofi_finalize


hkuno:
### Please describe the system on which you are running

* Operating system/version:  CentOS
* Computer hardware: dl385
* Network type: N/A (for example, sockets) 

-----------------------------

## Details of the problem

When running osu\_put\_latency with a basic provider (in this case, sockets):

```shell
${bin}/mpirun --bind-to none --mca btl ^openib,tcp,vader --mca mtl ^psm,psm2,portals4 --mca pml ^ucx --mca mtl_ofi_provider_include sockets --mca btl_ofi_provider_include sockets --mca btl_ofi_disable_sep true --mca mtl_ofi_enable_sep 0 
${libexec}/osu-micro-benchmarks/mpi/one-sided/osu_put_latency -m 4194304:4194304
```

... fi\_close returns -FI\_EBUSY when mca\_btl\_ofi\_finalize tries to close the domain because dom->ref > 0 .

```shell
(gdb) step
sock_dom_close (fid=0x241e4b8) at prov/sockets/src/sock_dom.c:180
180             dom = container_of(fid, struct sock_domain, dom_fid.fid);
(gdb) list
175     }
176
177     static int sock_dom_close(struct fid *fid)
178     {
179             struct sock_domain *dom;
180             dom = container_of(fid, struct sock_domain, dom_fid.fid);
181             if (ofi_atomic_get32(&dom->ref))
182                     return -FI_EBUSY;
183
184             sock_conn_stop_listener_thread(&dom->conn_listener);
(gdb) n
181             if (ofi_atomic_get32(&dom->ref))
(gdb)
182                     return -FI_EBUSY;
```

When OFI Libfabric binds objects to endpoints, before the object can
be successfully closed, the endpoint must first be freed.  For scalable
endpoints, objects can also be bound to transmit and receive contexts,
and for objects that are bound to contexts, we need to first free the
contexts before freeing the endpoint. We also need to clear the memory
registration cache before closing the domain.

If we don't clean up properly, then fi\_close may not be able to close
the domain because its ref count will be non-zero.

awlauria:
@hkuno did #7505 fix this? If so, we can close this issue.

hkuno:
**> @hkuno did #7505 fix this? If so, we can close this issue.

@awlauria Yes, #7505 fixes this. We can close this isssue.

PR #7505, opened 2020-03-05
Fix btl ofi clean-up logic

hkuno:
This fix is from John L. Byrne (john.l.byrne@hpe.com) and addresses issue https://github.com/open-mpi/ompi/issues/7504

When OFI Libfabric binds objects to endpoints, before the object can
be successfully closed, the endpoint must first be freed.  For scalable
endpoints, objects can also be bound to transmit and receive contexts,
and for objects that are bound to contexts, we need to first free the
contexts before freeing the endpoint. We also need to clear the memory
registration cache.

If we don't clean up properly, then fi\_close may not be able to close
the domain because the dom will have a non-zero ref count.

Signed-off-by: harumi kuno <harumi.kuno@hpe.com>

### DIFF
--- a/opal/mca/btl/ofi/btl_ofi_component.c
+++ b/opal/mca/btl/ofi/btl_ofi_component.c
@@ -584,6 +584,7 @@ fail:
     /* close basic ep before closing av */
     if (NULL != ep && !module->is_scalable_ep) {
         fi_close(&ep->fid);
+        ep = NULL;
     }
 
     /* if the contexts have not been initiated, num_contexts should
--- a/opal/mca/btl/ofi/btl_ofi_component.c
+++ b/opal/mca/btl/ofi/btl_ofi_component.c
@@ -581,6 +581,7 @@ static int mca_btl_ofi_init_device(struct fi_info *info)
 fail:
     /* clean up */
 
+    /* close basic ep before closing av */
     if (NULL != ep && !module->is_scalable_ep) {
         fi_close(&ep->fid);
     }
@@ -592,10 +593,12 @@ fail:
     }
     free(module->contexts);
 
+    /* check for NULL ep to avoid double-close */
     if (NULL != ep) {
         fi_close(&ep->fid);
     }
 
+    /* close av after closing basic ep */
     if (NULL != av) {
         fi_close(&av->fid);
     }
--- a/opal/mca/btl/ofi/btl_ofi_component.c
+++ b/opal/mca/btl/ofi/btl_ofi_component.c
@@ -581,6 +581,10 @@ static int mca_btl_ofi_init_device(struct fi_info *info)
 fail:
     /* clean up */
 
+    if (NULL != ep && !module->is_scalable_ep) {
+        fi_close(&ep->fid);
+    }
+
     /* if the contexts have not been initiated, num_contexts should
      * be zero and we skip this. */
     for (int i=0; i < module->num_contexts; i++) {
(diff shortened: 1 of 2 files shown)


opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -581,21 +581,25 @@ static int mca_btl_ofi_init_device(struct fi_info *info)
 fail:
     /* clean up */
 
+    if (NULL != ep && !module->is_scalable_ep) {
+        fi_close(&ep->fid);
```
awlauria: Looks like this could be double closed on line 596?

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -581,21 +581,25 @@ static int mca_btl_ofi_init_device(struct fi_info *info)
...
-    if (NULL != av) {
-        fi_close(&av->fid);
-    }
-
     if (NULL != ep) {
         fi_close(&ep->fid);
     }
 
+    if (NULL != av) {
+        fi_close(&av->fid);
```
awlauria: Seems like there's an ordering necessity here. It would be good to comment why this needs to come after freeing `ep->fid`

opal/mca/btl/ofi/btl_ofi_module.c
```diff
@@ -277,20 +277,32 @@ int mca_btl_ofi_finalize (mca_btl_base_module_t* btl)
...
-    if (NULL != ofi_btl->av) {
-        fi_close(&ofi_btl->av->fid);
-    }
-
     if (NULL != ofi_btl->ofi_endpoint) {
         fi_close(&ofi_btl->ofi_endpoint->fid);
     }
 
+    if (NULL != ofi_btl->av) {
+        fi_close(&ofi_btl->av->fid);
```
awlauria: Ordering problem again it seems. Good to have a comment here for that.

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

opal/mca/btl/ofi/btl_ofi_component.c
hkuno: I added a comment about checking for NULL to avoid a double-close. If the endpoint is scalable, this will not be a double-close. If the endpoint is basic, then we needed to close it before closing anything else.

Please let me know if I should be less succinct in comments.

opal/mca/btl/ofi/btl_ofi_component.c
hkuno: Thank you. I've added a brief comment.

opal/mca/btl/ofi/btl_ofi_module.c
hkuno: I added a brief comment. Should I instead have added a block comment ?

opal/mca/btl/ofi/btl_ofi_component.c
awlauria: The comments look good to me, thanks.

Couldn't `fi_close(&ep->fid);` be called again on line 603 though still? Should ep be set to NULL here? (or that line on 603 removed?)

hkuno: Good catch. Thank you. I will set ep to NULL.

hkuno: I've made the change and pushed it .
Before pushing it, I tested with both a scalable and a basic provider, forcing it into the fail code by using gdb to set a ret to a non-zero value in mca_btl_ofi_init_device and putting a breakpoint on fi_close to verify that without the fix there will be a double-close but with the fix, fi_close functions properly during the fail.

PR #7510, opened 2020-03-06
v3.0.x: Fix segv in btl/vader.

awlauria:
Keep track of the connected procs in vader_add_procs().
Otherwise, the same rank will reconnect the same shmem
segment (rank 0+...) multiple times instead of the next
one as intended.

(cherry picked from commit f69c8d6819dcc14f471cea90b50dc8ca98de12d4)

Conflicts:
	opal/mca/btl/vader/btl_vader_component.c

### DIFF
--- a/opal/mca/btl/vader/btl_vader.h
+++ b/opal/mca/btl/vader/btl_vader.h
@@ -118,6 +118,7 @@ struct mca_btl_vader_component_t {
     size_t segment_size;                    /**< size of my_segment */
     size_t segment_offset;                  /**< start of unused portion of my_segment */
     int32_t num_smp_procs;                  /**< current number of smp procs on this host */
+    int32_t local_rank;                     /**< current rank index at add_procs() time */
     opal_free_list_t vader_frags_eager;     /**< free list of vader send frags */
     opal_free_list_t vader_frags_max_send;  /**< free list of vader max send frags (large fragments) */
     opal_free_list_t vader_frags_user;      /**< free list of small inline frags */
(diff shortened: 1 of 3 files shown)


awlauria:
I can't hit this segv, but it appears to not be using shmem for on-node osc. Just running the `1sided` test and eyeballing it with pstack, it seems to be going over rdma:
```#0  0x00002000002cf4a4 in pthread_spin_lock () from /lib64/libpthread.so.0
#1  0x000020000254c128 in mlx5_poll_cq_1 () from /lib64/libmlx5-rdmav2.so
#2  0x000020000342295c in ibv_poll_cq (cq=0x276ee090, num_entries=256, wc=0x7fffd8f056b0) at /usr/include/infiniband/verbs.h:1458
#3  0x00002000034321ac in poll_device (device=0x27600ef0, count=0) at btl_openib_component.c:3609
#4  0x0000200003432894 in progress_one_device (device=0x27600ef0) at btl_openib_component.c:3742
#5  0x00002000034329b8 in btl_openib_component_progress () at btl_openib_component.c:3766
#6  0x00002000006a21bc in opal_progress () at runtime/opal_progress.c:222
#7  0x000020000cbbbd0c in ompi_osc_rdma_sync_rdma_complete (sync=0x27a1d0f8) at osc_rdma.h:493
#8  0x000020000cbbe868 in ompi_osc_rdma_complete_atomic (win=0x27f4f9d0) at osc_rdma_active_target.c:449
```
So we may not need this for the 3.x branches - unless that gets fixed.

jladd-mlnx:
I suspect this is due to the fact that atomic operations will use HCA atomics if available.  In general, you cannot mix CPU and HCA hardware atomics.

jsquyres:
Upon further testing, this problem was not reproducible.  Since v3.0.6 is out, going to close this without merging.

PR #7511, opened 2020-03-06
v4.0.x: schizo/slurm: Fix binding detection

artpol84:

### DIFF
--- a/orte/mca/schizo/slurm/schizo_slurm.c
+++ b/orte/mca/schizo/slurm/schizo_slurm.c
@@ -93,33 +93,12 @@ static orte_schizo_launch_environ_t check_launch_environment(void)
     opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"ess");
     opal_argv_append_nosize(&pushed_vals, "pmi");
 
-    /* if we are direct launched by SLURM, then we want
-     * to ensure that we do not override their binding
-     * options, so set that envar */
-    if (NULL != (bind = getenv("SLURM_CPU_BIND_TYPE"))) {
-        if (0 == strcmp(bind, "none")) {
-            opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"hwloc_base_binding_policy");
-            opal_argv_append_nosize(&pushed_vals, "none");
-            /* indicate we are externally bound so we won't try to do it ourselves */
-            opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"orte_externally_bound");
-            opal_argv_append_nosize(&pushed_vals, "1");
-        } else if (bind == strstr(bind, "mask_cpu")) {
-            /* if the bind list is all F's, then the
-             * user didn't specify anything */
-            if (NULL != (list = getenv("SLURM_CPU_BIND_LIST")) &&
-                NULL != (ptr = strchr(list, 'x'))) {
-                ++ptr;  // step over the 'x'
-                for (i=0; '\0' != *ptr; ptr++) {
-                    if ('F' != *ptr) {
-                        /* indicate we are externally bound */
-                        opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"orte_externally_bound");
-                        opal_argv_append_nosize(&pushed_vals, "1");
-                        break;
-                    }
-                }
-            }
-        }
-    }
+    /* if we are direct-launched by SLURM, then disable binding */
+    opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"hwloc_base_binding_policy");
+    opal_argv_append_nosize(&pushed_vals, "none");
+    /* indicate we are externally bound so we won't try to do it ourselves */
+    opal_argv_append_nosize(&pushed_envs, OPAL_MCA_PREFIX"orte_externally_bound");
+    opal_argv_append_nosize(&pushed_vals, "1");
 
   setup:
       opal_output_verbose(1, orte_schizo_base_framework.framework_output,


artpol84:
This fix seemed to work for my case.
I'm checking with SchedMD if this covers all of the cases

rhc54:
Could you explain what "case" you are trying to fix?

artpol84:
In my experiment with Slurm 18.08 —cpu_bind=rank is used. In this case envar will be = rank.

orte/mca/schizo/slurm/schizo_slurm.c
```diff
@@ -97,17 +97,11 @@ static orte_schizo_launch_environ_t check_launch_environment(void)
      * to ensure that we do not override their binding
      * options, so set that envar */
     if (NULL != (bind = getenv("SLURM_CPU_BIND_TYPE"))) {
-        if (0 == strcmp(bind, "none")) {
```
artpol84: @rhc54 this also doesn’t make sense to me.
It’s testing for bind == “none” and concludes that it is externally bound.
I think it’s a mistake

rhc54: `bind=none` is what you get from Slurm when they don't bind you to anything. Thus, if it is being bound at all, it is because something external to Slurm (e.g., `mpirun`) has done the binding.

rhc54:
> In my experiment with Slurm 18.08 —cpu_bind=rank is used. In this case envar will be = rank.

This has worked for quite some time. However, Slurm changed their cpu binding options not long ago - you can see that we had to make adjustments in the plm/slurm component. You may be hitting some other impacts from those changes, so you may need to be careful to differentiate between the Slurm versions being employed.

orte/mca/schizo/slurm/schizo_slurm.c
artpol84: The way I understood the term “externally” is that it is external to OMPI.
Thus if Slurm did the binding - this process is “externally bound”.
From your response, I read that “external” means external to Slurm.
 What should be the expected env outcome if Slurm did the binding and you want to avoid doing the binding in OMPI?

artpol84:
> you may need to be careful to differentiate between the Slurm versions being employed.

Yes, that’s why I set the PR “work in progress”

orte/mca/schizo/slurm/schizo_slurm.c
rhc54: Sorry - had to go back and look at this code again as I haven't had reason to do so for quite a few years. Let me clarify a bit.

The comment you are referring to relates to the setting of an MCA parameter indicating that OMPI should not attempt its own internal binding because the binding policy was externally set - in this case, by Slurm. The fact that the policy is "do not bind" is irrelevant. We are simply marking that the binding was set by something external to this process and should not be overrriden.

artpol84: So for my experiment, ompi was trying to bind while procs were already bound.

And I believe if a Slurm indicates that it did the binding (by having a non-“none” value) means we wanted to avoid binding at ompi level.

To be clear
‘ 0 == strcmp(bind, "none")‘ means that bind is equal to “none” and we are interested in cases where it is not equal.

artpol84: So for my experiment, ompi was trying to bind while procs were already bound.

And I believe if a Slurm indicates that it did the binding (by having a non-“none” value) means we wanted to avoid binding at ompi level.

To be clear
‘ 0 == strcmp(bind, "none")‘ means that bind is equal to “none” and we are interested in cases where it is not equal.

rhc54: `none` means the user specifically directed that their procs not be bound - thus, we should not bind them.

If the bind list is all `F`'s, then that means the user didn't specify anything, which means we should go ahead and bind the procs.

If the bind list is _not_ all `F`'s, then the user _did_ specify a binding pattern, and we need to leave it alone.

Hope that helps

hppritcha:
@artpol84 this missed the boat for the v4.0.4.  moving to v4.1.0.

orte/mca/schizo/slurm/schizo_slurm.c
angainor: @rhc54 I think that it should be ok for me to specify all `F`'s at `srun` argument and expect that OMPI would not force it's own binding. It should be also OK to set a default binding scheme for slurm to `none` (all `F`'s), and expect OMPI to not change it. In fact, unless explicitly specified by MCA environment variables, I would expect that if I run using `srun` then OMPI doesn't do any binding in any case.

rhc54: It's a complex issue. We went with the above policy at the request of users who were unhappy at the reduced performance they saw when they didn't specify a binding pattern. They expect OMPI to make things work well "out-of-the-box" without having to specify a bunch of things.

Bottom line is that we cannot please everyone 😄 . This policy was instantiated quite a long time ago. Slurm may have changed, as I indicated above, but all `F`'s used to mean that you didn't specify _anything_, which means OMPI is free to do what it thinks best. If you don't want either Slurm or OMPI to bind, then you should pass the `none` directive to Slurm and we should obey it too.

angainor: > Bottom line is that we cannot please everyone
sure not.. And I would never even find it an issue, had it not been for the hybrid MPI-OpenMP application, which suddenly ended up running all threads on the same core. Maybe that could be somehow helpful in deciding on the binding?

rhc54: Yeah, it's a problem. As I said on the other note, it probably merits some thought. Setting "none" isn't the right answer either, but we may need some other option to deal with the multi-NUMA-per-socket situation for multi-thread procs.

hppritcha:
@artpol84 where are we on this PR?

rhc54:
See #7806 - it's more complicated than this PR addresses.

artpol84:
@hppritcha @rhc54 @jsquyres I was unable to attend this week's telecon.
But plan to attend next week's one. I suggest discussing this.

artpol84:
Per today's telecon, I've checked Slurm v18.08.9 with the following result:

```
$./srun --version
slurm 18.08.9
artemp@hpchead /tmp/slurm_deploy/slurm/bin 
$./srun -n 1 --cpu_bind=rank /labhome/artemp/test_script.sh  | grep BIND
SLURM_CPU_BIND=quiet,rank
SLURM_CPU_BIND_LIST=
SLURM_CPU_BIND_TYPE=rank
SLURM_CPU_BIND_VERBOSE=quiet
```

The important thing there to see is that `SLURM_CPU_BIND` is set to `quiet,rank`.
Current code only disables binding in case this envar is set to "none" and this is exactly what I was trying to fix with this PR.

Bottom line: Slurm 18.08 has the same issue.

artpol84:
Per our agreement, I'm disabling the bindings

hppritcha:
@artpol84 could you also open a PR on v4.1.x?

artpol84:
@hppritcha yes, will do

PR #7518, opened 2020-03-10
Update PRRTE and PMIx

rhc54:
- Avoid modifying single-dash options of applications
- Fix fetch of node/app-level info
- Don't convert the return of the exited rank to a pmix status.
- Don't blow away the return code in prun main.


rhc54:
Somehow, I don't think this has anything to do with this PR:
```text
No components were able to be opened in the pml framework.

This typically means that either no components of this type were
installed, or none of the installed components can be loaded.
Sometimes this means that shared libraries required by these
components are unable to be found/loaded.

  Host:      swx-rdmz-vm-ompi
  Framework: pml
```
@artemry-mlnx Any ideas?

rhc54:
Nice - all of AWS was unable to download the repo. Sigh.

bot:ompi:retest

rhc54:
The Mellanox failure is due to an error in the CI script:
```text
Running /__w/1/ompi/ompi_install/thread_tests/thread-tests-1.1/overlap 
+ timeout -s SIGSEGV 17m /__w/1/ompi/ompi_install/bin/mpirun --np 4 --bind-to none --report-state-on-timeout --get-stack-traces --timeout 900 --mca coll '^hcoll' -x UCX_NET_DEVICES=mlx5_0:1 --mca pml ucx /__w/1/ompi/ompi_install/thread_tests/thread-tests-1.1/overlap ''
No components were able to be opened in the pml framework.

This typically means that either no components of this type were
installed, or none of the installed components can be loaded.
Sometimes this means that shared libraries required by these
components are unable to be found/loaded.

  Host:      swx-rdmz-vm-ompi
  Framework: pml
```
You are requiring the pml/ucx component on a threaded test and ucx is rejecting it. Most likely, ucx wasn't built with thread support.

@artemry-mlnx Can you please correct the script?

PR #7522, opened 2020-03-10
Update PRRTE and PMIx

rhc54:
- re-enable the timeout, state, and stacktrace testing options


rhc54:
@artemry-mlnx I checked other PRs and the Mellanox CI is failing everywhere due to the PML/UCX problem with multi-threaded apps. Can you please address it?

artemry-mlnx:
@rhc54 
Disabled thread tests within Mellanox Open MPI CI (https://github.com/mellanox-hpc/jenkins_scripts/pull/95).
Indeed the thread test scenarios should be revisit and reworked.
Thanks for the catch - it seems previously they're passed despite of the error (potentially wrong exit code).

rhc54:
Ah @#$@# - this PR never got reviewed and sat here for nothing.

PR #7525, opened 2020-03-11
Correct fence logic in MPI_Init

rhc54:
The fence logic in MPI_Init got messed up somehow such that we were
always executing a fence, which is not desirable. The logic is supposed
to be:

* if async fence is requested and we are not collecting data, then do
not fence at all

* if async fence is requested and we are collecting data, then execute
the fence in the background - wait for completion at the end of MPI_Init.

* if async fence is not requested, then execute a blocking fence at that
point, collecting data as directed. Note that we cannot actually do a
blocking fence as we need to cycle the event library via opal_progress
as the PMIx progress thread is tied to the OMPI event base.

### DIFF
--- a/ompi/runtime/ompi_mpi_init.c
+++ b/ompi/runtime/ompi_mpi_init.c
@@ -679,38 +679,39 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided,
 #endif
 
     if (!ompi_singleton) {
-        /* If we have a non-blocking fence:
-         * if we are doing an async modex, but we are collecting all
-         * data, then execute the non-blocking modex in the background.
-         * All calls to modex_recv will be cached until the background
-         * modex completes. If collect_all_data is false, then we skip
-         * the fence completely and retrieve data on-demand from the
-         * source node.
-         *
-         * If we do not have a non-blocking fence, then we must always
-         * execute the blocking fence as the system does not support
-         * later data retrieval. */
         if (opal_pmix_base_async_modex) {
-            /* execute the fence_nb in the background to collect
-             * the data */
-            background_fence = true;
-            active = true;
-            OPAL_POST_OBJECT(&active);
-            PMIX_INFO_LOAD(&info[0], PMIX_COLLECT_DATA, &opal_pmix_collect_all_data, PMIX_BOOL);
-            if( PMIX_SUCCESS != (rc = PMIx_Fence_nb(NULL, 0, NULL, 0,
-                                                    fence_release,
-                                                    (void*)&active))) {
-                ret = opal_pmix_convert_status(rc);
-                error = "PMIx_Fence_nb() failed";
-                goto error;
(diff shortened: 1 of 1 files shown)


rhc54:
More infrastructure failures:
```text
FATAL: Remote call on EC2 (Open MPI AWS Production) - Ubuntu 16.04 (i-09d1964918a421f34) failed
Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to EC2 (Open MPI AWS Production) - Ubuntu 16.04 (i-09d1964918a421f34)
		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1788)
		at hudson.remoting.UserRequest$ExceptionResponse.retrieve(UserRequest.java:356)
		at hudson.remoting.Channel.call(Channel.java:998)
		at hudson.Launcher$RemoteLauncher.kill(Launcher.java:1083)
		at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:510)
		at hudson.model.Run.execute(Run.java:1856)
		at hudson.matrix.MatrixRun.run(MatrixRun.java:153)
		at hudson.model.ResourceController.execute(ResourceController.java:97)
		at hudson.model.Executor.run(Executor.java:428)
java.lang.NoClassDefFoundError: Could not initialize class hudson.slaves.SlaveComputer
```

rhc54:
bot:ompi:retest

ompi/runtime/ompi_mpi_init.c
```diff
@@ -679,38 +679,43 @@ int ompi_mpi_init(int argc, char **argv, int requested, int *provided,
...
-        } else if (!opal_pmix_base_async_modex) {
-            /* we want to do the modex */
+        } else {
+            /* we want to do the modex - we block at this point, but we must
+             * do so in a manner that allows us to call opal_progress so our
+             * event library can be cycled as we have tied PMIx to that
+             * event base */
             active = true;
             OPAL_POST_OBJECT(&active);
-            PMIX_INFO_LOAD(&info[0], PMIX_COLLECT_DATA, &opal_pmix_collect_all_data, PMIX_BOOL);
```
artpol84: Can you explain a bit more why the old way doesn't work?

rhc54: Yeah, it's a tough choice. Technically, people are supposed to check not only for the existence of a `boolean` attribute, but also for the value of the `boolean` flag. Initially, we failed to do that in OpenPMIx and just equated existence with `true` - we made a concerted effort to go back and correct those places, and indeed it is handled correctly for this case.

The negative of _not_ providing it is that we are assuming the default condition will be "do not collect", which again may not be correct.

So I guess maybe we should always provide it as technically that would be more correct. If someone isn't checking the value of the flag and just equates existence with `true`, then we can argue that their implementation has a bug (though it may be hard to sniff that out). I can make that adjustment.

PR #7536, opened 2020-03-17
Support hwloc retrieval using legacy key

### LINKED ISSUE #7507: HWLOC information exported by Slurm/PMIx is not honored starting from v4.x
The Slurm plugin is still using legacy PMIx key to export hwloc topology:
```C
#define OPAL_PMIX_LOCAL_TOPO "pmix.ltopo"
```
It works fine in v3.x.
https://github.com/open-mpi/ompi/blob/v3.0.0rc1/opal/mca/hwloc/base/hwloc_base_util.c#L262

For some reason, in v.4 querying of this variable was removed. This causes up to 5 seconds of delays during job startup. 

I would like to return the support of this variable back in v4.
@hppritcha do you think it is possible?

rhc54:
Re-enable support for Slurm plugin using earlier PMIx_LOCAL_TOPO key

Fixes #7507 

### DIFF
--- a/opal/mca/hwloc/base/hwloc_base_util.c
+++ b/opal/mca/hwloc/base/hwloc_base_util.c
@@ -13,10 +13,10 @@
  * Copyright (c) 2011-2018 Cisco Systems, Inc.  All rights reserved
  * Copyright (c) 2012-2017 Los Alamos National Security, LLC.
  *                         All rights reserved.
- * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


gpaulsen:
Isn't this desirable on master as well?  Or is it intended that master won't support this legacy key and will only allow the newer approach?  Is there a description somewhere of what changes are needed for Slurm or any other RM using the legacy key to update to the newer way of doing this?

rhc54:
It is already on master, @gpaulsen

gpaulsen:
I see:  `ed42f568a` is the commit on master.

gpaulsen:
@hppritcha looks fine to me.

PR #7540, opened 2020-03-18
opal_check_pmi.m4: properly save top-level flags

jsquyres:
CPPFLAGS, LDFLAGS, and LIBS were only being saved conditionally, but
restored unconditionally.  This could result in wiping out
CPPFLAGS/LDFLAGS/LIB.

Make sure to *always* save these flags so that when they are restored,
they are restored to their proper value.

**NOTE:** This is only an issue on master: v4.0.x and older are not affected.


jsquyres:
@hppritcha This is what got me on MacOS today while reviewing the thread PR (https://github.com/open-mpi/ompi/pull/6578): it hosed my flags when compiling with Fortran on MacOS.

PR #7542, opened 2020-03-18
v3.0.x: Backport a fix for pgcc18 to pmix.

awlauria:
Refs openpmix/openpmix#1628


jjhursey:
PMIx 2.x will likely not have another release. However, @awlauria Can you file a PR against the v2.2 branch of OpenPMIx so if anyone goes looking for the fix they can find it in the repo?

I'd suggest that the RM's take this patch to the branch. This patch should have come over with https://github.com/open-mpi/ompi/pull/7390

jsquyres:
Per developer discussions, it sounds like this is not a regression against v3.0.x.  So we're not including it in v3.0.6.  It will be merged after the formal v3.0.6 release, meaning that it'll be in the nightly snapshot tarball for any users who need it.

awlauria:
@jjhursey opened https://github.com/openpmix/openpmix/pull/1678

PR #7545, opened 2020-03-19
README: Document what compiler atomics support is necessary

rhc54:
Apparently, recent changes to atomics causes OMPI to fail to configure
on Solaris machines:

Tried:
Studio 12.5 Sun C 5.14 SunOS_sparc 2016/05/31
GCC 4.0.2

Error:
configure: error: No atomic primitives available for sparc-sun-solaris2.11

Thanks to @jdelsign for the report

### DIFF
--- a/README
+++ b/README
@@ -15,7 +15,7 @@ Copyright (c) 2007      Myricom, Inc.  All rights reserved.
 Copyright (c) 2008-2018 IBM Corporation.  All rights reserved.
 Copyright (c) 2010      Oak Ridge National Labs.  All rights reserved.
 Copyright (c) 2011      University of Houston. All rights reserved.
-Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
+Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
 Copyright (c) 2015      NVIDIA Corporation.  All rights reserved.
 Copyright (c) 2017-2018 Los Alamos National Security, LLC.  All rights
                         reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
@gpaulsen @hppritcha @bwbarrett 

I'm not sure how far down the release branches the atomics changes were propagated - you might need to update this in your branch README's and on the web site

jsquyres:
@gpaulsen We should probably bring this up on the Tuesday webex -- do we care if we don't support Solaris any more?

jsquyres:
Tagging @hjelmn and @bosilca, too.

jdelsign:
According to the blurb here <https://docs.oracle.com/cd/E77782_01/html/E77789/gqhbq.html> we'd need GCC 4.8 or later, or Studio 12.6. Even then, there are some incompatibilities between the in the atomics libraries, which means that GCC and Studio code that use the atomics cannot be mixed together in the same process.

hppritcha:
I think the big changes for c11 atomics is only in master.

jsquyres:
Talking to @hjelmn, we decided at the Portland Feb 2020 meeting that we only support:

1. hand-written atomics on x86, ARM, and PPC
1. C11 atomics (i.e., "the future") wherever they exist
1. GCC's `__atomic` atomics (i.e., gcc >= v4.7.2)
     * **NOTE:** these were specifically added to replace the "legacy" GCC `sync` atomics

Specifically:

1. We do not want to support any other hand-written atomics (e.g., for Solaris) -- everything other than x86/ARM/PPC hand-written atomics have been removed from master / won't be in v5.0.x
1. We are not supporting GCC's `sync` builtin atomics any more (they have been removed from master / won't be in v5.0.x)

We should make a PR to ensure that README is correct.

jsquyres:
Meaning: I will tweak this README/PR a little more to be a bit more specific.

bwbarrett:
Um, there's no reason the hand written atomics shouldn't work on Solaris.  Perhaps not work on UltraSPARC, but Solaris on x86 should be fine.  Let's not conflate OS and architecture :).

PR #7546, opened 2020-03-20
grequestx: fix racy initialization and premature destruction

devreal:
This PR fixes two issues I found:

1) Upon completion, the request object may be destroyed before it is removed from the list of active requests. The object should be retained until it is removed from the list.

2) The initialization has been shaky if two threads attempt to register "the first" request, potentially leading to one request getting lost.


awlauria:
@devreal did you want to pr for release branches? Thanks.

PR #7547, opened 2020-03-20
Ensure we get our local topology

rhc54:
Restore missing call to get_topology - others were doing it in their
components as repeated calls just return success, but let's ensure it is
always present.

### DIFF
--- a/ompi/runtime/ompi_rte.c
+++ b/ompi/runtime/ompi_rte.c
@@ -747,6 +747,12 @@ int ompi_rte_init(int *pargc, char ***pargv)
         val = NULL;
     }
 
+    /* get our topology */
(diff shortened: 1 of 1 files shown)


rhc54:
Urrrr....I believe the discussion on the devel mailing list had concluded that we didn't want to commit this one. The mtl/ofi component was going to call `opal_hwloc_get_topology` to ensure the topology was present just as all the other components who need the topo do, thus preserving the "only load the topology if someone will actually use it" philosophy.

awlauria:
@rhc54 ah - I mused have missed that email...My bad. I can revert this.

awlauria:
https://github.com/open-mpi/ompi/pull/7558

@rhc54 ^

PR #7548, opened 2020-03-20
usnic: remove typo

jsquyres:
Remove an amusing -- but harmless -- typo.

### DIFF
--- a/opal/mca/btl/usnic/btl_usnic_hwloc.c
+++ b/opal/mca/btl/usnic/btl_usnic_hwloc.c
@@ -201,7 +201,7 @@ int opal_btl_usnic_hwloc_distance(opal_btl_usnic_module_t *module)
                         "btl:usnic:filter_numa: filtering devices by NUMA distance");
 
     /* ensure we have the topology */
-    if (OPAL_SUCCESS !=- opal_hwloc_base_get_topology()) {
+    if (OPAL_SUCCESS != opal_hwloc_base_get_topology()) {
         opal_output_verbose(5, USNIC_OUT,
                             "btl:usnic:filter_numa: not sorting devices by NUMA distance (topology not available)");
         return OPAL_SUCCESS;


jsquyres:
@bwbarrett @wckzhang CI failed on this in the IPv6 build, but this PR has nothing to do with IPv6.

Could one of you look at that build and see what went wrong?

wckzhang:
I don't believe I have edit access or hardware access to the jenkins, but we should fix the -hostfile in the command line.

The relevant error occurred during ring_c:
```
Hello, world, I am 1 of 2, (Open MPI v4.1.0a1, package: Open MPI ubuntu@ip-172-31-10-242 Distribution, ident: 4.1.0a1, repo rev: v2.x-dev-7599-ga361b80, Unreleased developer copy, 151)
Hello, world, I am 0 of 2, (Open MPI v4.1.0a1, package: Open MPI ubuntu@ip-172-31-10-242 Distribution, ident: 4.1.0a1, repo rev: v2.x-dev-7599-ga361b80, Unreleased developer copy, 151)
[warn] Epoll MOD(1) on fd 20 failed.  Old events were 6; read change was 0 (none); write change was 2 (del): Bad file descriptor
[warn] Epoll MOD(4) on fd 20 failed.  Old events were 6; read change was 2 (del); write change was 0 (none): Bad file descriptor
--> Running example: ring_c
---------------------------
```

Doesn't seem related to your change.

jsquyres:
@wckzhang Ya, I know you don't have usNIC hardware.  😄  I should have been more clear: I was asking because the IPV6 job is running on AWS, and that's where the error occurred.

Hey @rhc54 -- is the error that William is showing a PMIX or PRRTE error, perchance?

rhc54:
Not sure - it's certainly possible that something in the PMIx IPv6 support isn't right as nobody checks it on our side. Is the test for IPv6 something newly added to the CI? If not, I can't think of anything we have done that relates to IPv6, and I would have thought it would have shown up in an update PR for PMIx and/or PRRTE.

bwbarrett:
The IPv6 test has been there for at least a year, so it's definitely not new.

The first crash looks like it was during finalize/cleanup.  These are single core instances; I wonder if it didn't clean up, causing the other test to timeout.  either way, it looks like we might have a problem with IPv6 support.

bwbarrett:
bot:aws:retest

PR #7556, opened 2020-03-23
v3.1.x: memory/patcher: fix compiler warning

nmorey:
syscall() returns a long, but we are invoking shmat(), which returns
a void*.

### DIFF
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -440,7 +440,7 @@ static void *_intercept_shmat(int shmid, const void *shmaddr, int shmflg)
 
     if (!original_shmat) {
 #if defined(SYS_shmat)
-        result = memory_patcher_syscall(SYS_shmat, shmid, shmaddr, shmflg);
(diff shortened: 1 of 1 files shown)


nmorey:
Backport of patch 52a9cce6f3dcd87e9ae66177398b60b9317e9339 in master branch

jsquyres:
ok to test

jsquyres:
We just released v3.1.6, and we really, really don't want to do another v3.1.x release unless it is absolutely necessary.  As such, we'd kinda like the bar to be fairly high before accepting fixes to the v3.1.x branch.

What's your level of need here -- do you need this fix for some SuSE build/QA automation, perchance?

nmorey:
SUSE needs this patch as we have a 64bit portability checker that prevents 3.1.6 from building because of this int to void* cast.
However I've manually applied the patch to the SUSE package so we are not stuck.
If the 3.1.x branch is dead, feel free to drop this PR.

bwbarrett:
We're going to merge this on the off chance we do another 3.1.x release.  I don't think this is worthy of a release itself, but we'll see what happens.

PR #7566, opened 2020-03-25
Update PMIx and PRRTE

### LINKED ISSUE #7564: master: mpirun does not emit error message with non-existent executable
On master HEAD, `mpirun` (i.e., `prun`), does not seem to emit an error message if there is no executable found:

```
$ mpirun -n 1 uptime
 15:03:51 up 200 days, 10:36, 80 users,  load average: 0.59, 1.66, 1.29
$ mpirun -n 1 uptimefasdfgaskfasdlkfsadf
$ echo $?
78
```

### LINKED ISSUE #7565: "mpirun -np 1 ..." emits warning
I could be wrong, but I thought we decided to whitelist `-np` as an allowable "single dash" option, simply because of its massive prevalence out in the user community.

I notice that it's still emitting a warning:

```
$ mpirun -np 1 uptime
WARNING: All multi-character command line options must be prefixed with double
dashes, not a single dash:

  Deprecated option:   -np
  Corrected option:    --np

We have updated this for you and will proceed. However, this will be treated
as an error in a future release. Please update your command line.

******* Corrected cmd line: mpirun --np 1 uptime


 15:04:20 up 200 days, 10:36, 80 users,  load average: 0.48, 1.53, 1.26
```

Can we whitelist it to not emit this warning?

If someone points me in the right direction, I'd be happy to submit a PR.

rhc54:
Deprecate --am and --amca options

Avoid default param files on backend nodes
Any parameters in the PRRTE default or user param files will have been
picked up by prte and included in the environment sent to the prted, so
don't open those files on the backend.

Avoid picking up MCA param file info on backend
Avoid the scaling problem at PRRTE startup by only reading the system
and user param files on the frontend.

Complete revisions to cmd line parser for OMPI
Per specification, enforce following precedence order:

1. system-level default parameter file
1. user-level default parameter file
1. Anything found in the environment
1. "--tune" files. Note that "--amca" goes away and becomes equivalent to "--tune". Okay if it is provided more than once on a cmd line (we will aggregate the list of files, retaining order), but an error if a parameter is referenced in more than one file with a different value
1. "--mca" options. Again, error if the same option appears more than once with a different value. Allowed to override a parameter referenced in a "tune" file
1. "-x" options. Allowed to overwrite options given in a "tune" file, but cannot conflict with an explicit "--mca" option
1. all other options

Fix special handling of "-np"
Fixes #7565 

Get agreement on jobid across the layers
Need all three pieces (PRRTE, PMIx, and OPAL) to agree on the nspace
conversion to jobid method

Ensure prte show_help messages get output
Print abnormal termination messages
Fixes #7564 

Add extra libs to PRRTE binaries for external deps
libevent, hwloc, and pmix can be external and may require that their
libs be explicitly linked into the PRRTE binaries

Fix scalable launch with rsh - resolve several issues in tree-spawn

### DIFF
--- a/config/ompi_setup_prrte.m4
+++ b/config/ompi_setup_prrte.m4
@@ -24,7 +24,7 @@
 #
 
 AC_DEFUN([OMPI_SETUP_PRRTE],[
-    OPAL_VAR_SCOPE_PUSH([opal_prrte_save_CPPFLAGS opal_prrte_save_CFLAGS opal_prrte_save_LDFLAGS opal_prrte_save_LIBS opal_prrte_args opal_prrte_save_enable_dlopen opal_prrte_save_enable_mca_dso opal_prrte_save_enable_mca_static])
+    OPAL_VAR_SCOPE_PUSH([opal_prrte_save_CPPFLAGS opal_prrte_save_CFLAGS opal_prrte_save_LDFLAGS opal_prrte_save_LIBS opal_prrte_args opal_prrte_save_enable_dlopen opal_prrte_save_enable_mca_dso opal_prrte_save_enable_mca_static opal_prrte_extra_libs opal_prrte_extra_ltlibs opal_prrte_extra_ldflags])
 
     opal_prrte_save_CFLAGS=$CFLAGS
     opal_prrte_save_CPPFLAGS=$CPPFLAGS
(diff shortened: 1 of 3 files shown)


rhc54:
@artemry-mlnx I fully expect this to fail Mellanox CI due to the changes in cmd line parsing rules when it comes to "--tune", "--amca", "--mca", and "-x" options. I'll need your help to work thru the updates to your CI

ibm-ompi:

 Gist: https://gist.github.com/abd088a29a277ad9887d94a10eda74c0

artemry-mlnx:
@rhc54 sure, I'll deal with the CI script update.

ibm-ompi:

 Gist: https://gist.github.com/b77a4297874f9da7e8e23744ebeda892

ibm-ompi:

 Gist: https://gist.github.com/e5dd7d0b1ddaf3edb44c7f54a1c23022

ibm-ompi:

 Gist: https://gist.github.com/9112b4655efa24bc07f8b800e1cf7141

ibm-ompi:

 Gist: https://gist.github.com/4c997ba8d94165b5a7876c2a39025e46

gpaulsen:
Well, host `c656f6n03` is up, but address `172.18.0.3` must be a docker overlay network.
@jjhursey can you please take a look?

rhc54:
No, it isn't anything to do with IBM CI - the MCA params need to be properly prefixed in the OMPI case and they aren't. I'm fixing it

jjhursey:
> Well, host `c656f6n03` is up, but address `172.18.0.3` must be a docker overlay network.

Just to follow up on this question: Yeah the 172.18.0.x network is the overlay inside the virtual cluster. It is private and routable between only the "virtual nodes" in that cluster. It shouldn't be getting in the way. It sounds like Ralph is working the real issue.

jsquyres:
Inside an salloc job, I see that mpirun of a non-MPI job works fine, but mpirun of an MPI job fails:

```
$ mpirun hostname | sort | uniq
mpi028
mpi029

$ mpirun ring_c
------------------------------------------------------------
A process or daemon was unable to complete a TCP connection
to another process:
  Local host:    savbu-usnic-a
  Remote host:   mpi028
This is usually caused by a firewall on the remote host. Please
check that any firewall (e.g., iptables) has been disabled and
try again.
------------------------------------------------------------
------------------------------------------------------------
A process or daemon was unable to complete a TCP connection
to another process:
  Local host:    savbu-usnic-a
  Remote host:   mpi029
This is usually caused by a firewall on the remote host. Please
check that any firewall (e.g., iptables) has been disabled and
try again.
------------------------------------------------------------
```

rhc54:
Not sure I can understand that one - did the daemon on the remote host segfault? Otherwise, it's hard to understand why the PRRTE infrastructure would care about MPI vs non-MPI apps

jsquyres:
I am now unable to reproduce the error.  Weird, because it was 100% reproducible before.  🤷‍♂ 

Sooo... let's ignore my report, I guess!

jsquyres:
What's the failure in the Mellanox CI?  Is that some in-flight CLI change that the Mellanox CI needs to be updated for?

rhc54:
Yes and no. Yes, the specific test should return an error because they included the same envar in two different tune files, and assigned it a different value. This is no longer permitted.

No in that there are still some bugs in the parser as it didn't error out as it should. I'm working on that one, and then we'll have to address the test itself.

rhc54:
@artemry-mlnx I believe I have this working now and reporting cmd line errors as it should. There are two tests that are going to fail:
```text
echo "--mca mca_base_env_list \"XXX_A=1;XXX_B=2;XXX_C;XXX_D;XXX_E\"" > "$WORKSPACE/test_tune.conf"
# shellcheck disable=SC2086
val=$("${OMPI_HOME}/bin/mpirun" $mca -np 2 --tune "$WORKSPACE/test_tune.conf" --mca mca_base_env_list \
     "XXX_A=7;XXX_B=8" "${abs_path}/env_mpi" | sed -n -e 's/^XXX_.*=//p' | sed -e ':a;N;$!ba;s/\n/+/g' | bc)
# precedence goes left-to-right.
# A is set to 1 in "tune", and then reset to 7 in the --mca parameter
# B is set to 2 in "tune", and then reset to 8 in the --mca parameter
# C, D, E are taken from the environment as 3,4,5
# return (7+8+3+4+5)*2=54

val=$("${OMPI_HOME}/bin/mpirun" $mca --np 2 --tune "$WORKSPACE/test_tune.conf" \
    --am "$WORKSPACE/test_amca.conf" "${abs_path}/env_mpi" | sed -n -e 's/^XXX_.*=//p' | sed -e ':a;N;$!ba;s/\n/+/g' | bc)
# precedence goes left-to-right.
# A is first set to 1 in "tune", and then reset to 7 in "amca".
# B is first set to 2 in "tune", but then reset to 8 in "amca"
# C, D, E are taken from the environment as 3,4,5
# return (7+8+3+4+5)*2=54
```
These will fail because they "reset" values, which is no longer allowed. You can override values from a file by putting a new value on the cmd line itself, but values from within two files (or even within a single file) are not allowed to override each other. I'm not sure how you would like to fix those - we could just turn them off, I suppose, or you could fix the override. Up to you.

rhc54:
@artemry-mlnx Any estimate on when you will update the tests? We are hitting a bit of a problem in that PRRTE is now ahead of OMPI and cannot launch OMPI jobs without this PR being committed.

artemry-mlnx:
@rhc54 working on the script update.
ETA: today/tomorrow.

artemry-mlnx:
@rhc54 
Could you please elaborate a bit on the new logic based on the scenarios above - who is not allowed to overwrite whom?
I've temporary disabled 2 tests you mentioned above but there are still some failed test scenarios: [log file](https://dev.azure.com/mlnx-swx/9a715c71-515f-4a6f-a324-ca65880a05ca/_apis/build/builds/690/logs/7) - could you please take a look?

rhc54:
@artemry-mlnx Perhaps the easiest approach is for me to give you a pull request with the required changes? That will give me a chance to manually check each test and determine the required changes

rhc54:
@artemry-mlnx I believe https://github.com/mellanox-hpc/jenkins_scripts/pull/96 will fix the CI tests. Can you please review and we can recheck here?

rhc54:
@jladd-mlnx

artemry-mlnx:
@rhc54 
Have you verified the changes?

rhc54:
@artemry-mlnx I ran the updated tests by hand and they all worked, so I believe it should pass. If not, I can deal with any failures once I see them.

artemry-mlnx:
@rhc54 
Can you please update Azure Pipeline file in your feature branch to the new CI scripts (your topic branch in jenkins_scripts repo) to verify before merging them: https://github.com/open-mpi/ompi/blob/5dcd1f4e6c8006d886a0b31ee81d15eb443d02fb/.ci/mellanox/azure-pipelines.yml#L13
https://github.com/open-mpi/ompi/blob/5dcd1f4e6c8006d886a0b31ee81d15eb443d02fb/.ci/mellanox/azure-pipelines.yml#L14

rhc54:
@artemry-mlnx Done - let's see how it does!

artemry-mlnx:
@rhc54 it's passed, please revert back Azure Pipeline and I'll proceed with the merge for jenkins_scripts.

rhc54:
@artemry-mlnx Done!

artemry-mlnx:
@rhc54 
Ideally we need to merge these both PRs in the same time - are you going to merge this Open MPI PR after CI is passed?

rhc54:
@artemry-mlnx Yes - though it isn't clear that this PR is going to pass without my local CI change unless you commit the PR, true?

artemry-mlnx:
That's true - merged first.

rhc54:
@artemry-mlnx Thanks!

PR #7571, opened 2020-03-27
ofi: Call add_procs through PML

bwbarrett:
Change ompi_mtl_ofi_get_endpoint() to call the active PML's
add_procs() rather than the OFI MTL add_procs() directly when
discovering a new process during operation.

Functionally, this has no impact in correct operation.  However,
the current behavior means that the heterogenous and active PML
checks are not being executed in the dynamic discovery case.

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -4,8 +4,9 @@
  *                         reserved.
  * Copyright (c) 2019      Triad National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2018-2020 Amazon.com, Inc. or its affiliates. All rights
+ *                         reserved.
  *
- * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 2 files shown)


bwbarrett:
@hjelmn, @bosilca, and @rhc54 I'm sure one of you worked on the dynamic add_procs() code.  Can you sanity check this patch does what the original design expected?

@regrant, @mdosanjh, @mwheinz It looks like the same issue exists in the Portals 4 and PSM2 MTLs, and I think the change is exactly the same.  Today, nothing bad will happen if you don't include this change, other than you won't get some correctness checks that the same PML is used on both processes.  If we ever needed a proc table in the CM PML, that could be a problem.

rhc54:
Afraid I didn't actively work on the MPI side of the change - I believe that was mostly @hjelmn

bwbarrett:
> We only missed the heterogeneity check, the PML was already checked or you would not be down in the OFI MTL discovering new procs.

This is true on the receive side.  But on the send side, nothing has called pml.add_procs() when this function is called to add the endpoint data.  The CM PML does not store anything in the proc, so it doesn't call add_procs() on itself on the send path.  Unless I'm missing something...

PR #7572, opened 2020-03-27
Add .mailmap entry for Harumi Kuno

rhc54:
Avoid the problem of someone using caps in their name from one machine
and not using caps in their name when committing from another. This
generates a false error when making the tarball.


rhc54:
@jsquyres @bwbarrett This fixes the nightly tarball build for master

jsquyres:
Updated the PR to simply add a `.mailmap` entry to @hkuno

PR #7577, opened 2020-03-29
mtl/ofi: Added multi-NIC support to provider selection

dancejic:
changed logic for selecting provider to take into account
the hardware distance between the device and process NUMA nodes.
This is in preparation for multi NIC support.


awlauria:
ok to test

dancejic:
Retracting for now, needs some internal discussion. will re-post after internal consensus.

PR #7579, opened 2020-03-30
Harmonize return values of progress callbacks (v4.0.x)

devreal:
The logic in opal_progress expects a callback to return "the number of events progressed" and based on that determines whether or not to yield the CPU. Most components track the number of completed operations for that matter. However, some return a fixed value instead. The problem is that if a callback always returns a positive value (psm2 mtl, pt2pt osc) the process will never yield if these components are active. Similarly, if the callback always returns OMPI_SUCCESS (or zero; libnbc, ucx pml, uct btl, grequestx), the process will yield unconditionally even if it wasn't necessary (plus, there is a mismatch in semantics in the first place). This PR tries to adapt the components in question to try to track whether progress has been made and signal it accordingly.

These changes should have no impact unless yield_when_idle is enabled.

Cherry-pick of #7228 to v4.0.x

### DIFF
--- a/ompi/communicator/comm_request.c
+++ b/ompi/communicator/comm_request.c
@@ -100,6 +100,7 @@ static int ompi_comm_request_progress (void)
 {
     ompi_comm_request_t *request, *next;
     static int32_t progressing = 0;
+    int completed = 0;
 
     /* don't allow re-entry */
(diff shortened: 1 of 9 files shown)


hppritcha:
@devreal please assign one or more reviewers

jladd-mlnx:
@hoopoepg please take a look.

Issue #7570, opened 2020-03-26
UCX PML add_procs bug


bwbarrett:
While digging into an odd add_procs behavior for the ofi mtl and the cm pml, I ran across what I'm pretty sure is a bug (or two) in the UCX PML on master.  The issue is in mca_pml_ucx_add_proc() on [line 403](https://github.com/open-mpi/ompi/blob/c704ed4cc51baf3ee0926454476b2bee0148581f/ompi/mca/pml/ucx/pml_ucx.c#L403) where the UCX PML is calling pml_check_selected().  First issue is that pml_check_selected() expects the proc being added, not proc 0, to be the one passed into the function.  The second issue is that the third argument is supposed to be the number of entries in the procs array, not the remote rank (the comment about not using the third argument isn't guaranteed, but is true today).

The impact of the first issue is that you've essentially disabled the PML is the same check when using late-binding proc creation today, and will cause a return of the rank 0 direct call flood issue in the near future.  Today, the code got better because George added a temporary hack to the check_selected call to return success if the key wasn't already locally loaded.  Soon, we are finishing the fix and replacing that with a blocking call to the added rank (which has already been filled in and therefore is a quick call).  However, this change will cause a return of the "everyone pounds on rank 0" problem for the UCX PML, as it is passing the wrong data to check_selected().  Worse, because the UCX PML is passing an erroneous procs array length, the fixes mentioned above are likely to start causing segfaults.

I have no way of testing changes to the UCX PML; can someone who does please take a look at this issue?

bosilca:
I have a patch, but right now I am unable to run any OMPI based code so I can't validate it. I also allowed compilations with newer versions of UCX, I've been using them (including UCX master) successfully for quite some time.

```diff
diff --git a/ompi/mca/pml/ucx/pml_ucx.c b/ompi/mca/pml/ucx/pml_ucx.c
index 93a2760..fb97299 100644
--- a/ompi/mca/pml/ucx/pml_ucx.c
+++ b/ompi/mca/pml/ucx/pml_ucx.c
@@ -393,22 +393,6 @@ static ucp_ep_h mca_pml_ucx_add_proc_common(ompi_proc_t *proc)
     return ep;
 }

-static ucp_ep_h mca_pml_ucx_add_proc(ompi_communicator_t *comm, int dst)
-{
-    ompi_proc_t *proc0     = ompi_comm_peer_lookup(comm, 0);
[...]

PR #7582, opened 2020-03-30
mca/pml: PML check for direct modex

dipti-kothari:
For direct modex, all procs publish the selected pml module
and then at add_procs pml module for each proc is checked
against every other proc in the add_proc call.
For full modex, there is no change in functionality. Only Rank0
publishes its selected pml, all other procs in the add_proc call
check their selected pml against Rank0.
If pml's do not match, throw error and exit.
Issue: https://github.com/open-mpi/ompi/issues/7475

Signed-off-by: Dipti Kothari <dkothar@amazon.com>

### DIFF
--- a/ompi/mca/pml/base/pml_base_select.c
+++ b/ompi/mca/pml/base/pml_base_select.c
@@ -14,6 +14,8 @@
  *                         reserved.
  * Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2020      Amazon.com, Inc. or its affiliates.  All Rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -44,8 +46,6 @@ typedef struct opened_component_t {
   mca_pml_base_component_t *om_component;
 } opened_component_t;
 
-static bool modex_reqd=false;
-
 /**
  * Function for selecting one component from all those that are
  * available.
@@ -59,7 +59,7 @@ static bool modex_reqd=false;
 int mca_pml_base_select(bool enable_progress_threads,
                         bool enable_mpi_threads)
 {
-    int i, priority = 0, best_priority = 0, num_pml = 0;
+    int i, priority = 0, best_priority = 0, num_pml = 0, ret = 0;
     opal_list_item_t *item = NULL;
     mca_base_component_list_item_t *cli = NULL;
     mca_pml_base_component_t *component = NULL, *best_component = NULL;
@@ -186,13 +186,6 @@ int mca_pml_base_select(bool enable_progress_threads,
                          "selected %s best priority %d\n",
                          best_component->pmlm_version.mca_component_name, best_priority);
 
-    /* if more than one PML could be considered, then we still need the
-     * modex since we cannot know which one will be selected on all procs
-     */
-    if (1 < num_pml) {
-        modex_reqd = true;
-    }
-
     /* Save the winner */
 
     mca_pml_base_selected_component = *best_component;
@@ -287,13 +280,11 @@ int mca_pml_base_select(bool enable_progress_threads,
     }
 
     /* register winner in the modex */
-    if (modex_reqd && 0 == OMPI_PROC_MY_NAME->vpid) {
-        mca_pml_base_pml_selected(best_component->pmlm_version.mca_component_name);
-    }
+    ret = mca_pml_base_pml_selected(best_component->pmlm_version.mca_component_name);
 
     /* All done */
 
-    return OMPI_SUCCESS;
+    return ret;
 }
 
 /* need a "commonly" named PML structure so everything ends up in the
@@ -307,50 +298,55 @@ static mca_base_component_t pml_base_component = {
 };
 
 
+/*
+ * If direct modex, then publish PML for all procs. If full modex then
+ * publish PML for rank 0 only. This information is used during add_procs
+ * to perform PML check.
+ * During PML check, for direct modex, compare our PML with the peer's
+ * PML for all procs in the add_procs call. This does not change the
+ * connection complexity of modex transfers, since adding the proc is
+ * going to get the peer information in the MTL/PML/BTL anyway.
+ * For full modex, compare our PML with rank 0.
+ * Direct Modex is performed when collect_all_data is false, as we do
+ * not perform a fence operation during MPI_Init if async_modex is true.
+ * If async_modex is false and collect_all_data is false then we do a
+ * zero-byte barrier and we would still require direct modex during
+ * add_procs
+ */
 int
 mca_pml_base_pml_selected(const char *name)
 {
-    int rc;
+    int rc = 0;
 
-    OPAL_MODEX_SEND(rc, PMIX_GLOBAL, &pml_base_component, name, strlen(name) + 1);
+    if (!opal_pmix_collect_all_data || 0 == OMPI_PROC_MY_NAME->vpid) {
+        OPAL_MODEX_SEND(rc, PMIX_GLOBAL, &pml_base_component, name,
+                        strlen(name) + 1);
+    }
     return rc;
 }
 
-int
-mca_pml_base_pml_check_selected(const char *my_pml,
-                                ompi_proc_t **procs,
-                                size_t nprocs)
+static int
+mca_pml_base_pml_check_selected_impl(const char *my_pml,
+                                     opal_process_name_t proc_name)
 {
     size_t size;
-    int ret;
+    int ret = 0;
     char *remote_pml;
-    opal_process_name_t rank0 = {.jobid = ompi_proc_local()->super.proc_name.jobid, .vpid = 0};
 
-    /* if no modex was required by the PML, then
-     * we can assume success
-     */
-    if (!modex_reqd) {
+    /* if we are proc_name=OMPI_PROC_MY_NAME, then we can also assume success */
+    if (0 == opal_compare_proc(ompi_proc_local()->super.proc_name, proc_name)) {
         opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
-                            "check:select: modex not reqd");
+                            "check:select: PML check not necessary on self");
         return OMPI_SUCCESS;
     }
-
-    /* if we are rank=0, then we can also assume success */
-    if (0 == OMPI_PROC_MY_NAME->vpid) {
-        opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
-                            "check:select: rank=0");
-        return OMPI_SUCCESS;
-    }
-
-    /* get the name of the PML module selected by rank=0 */
-    OPAL_MODEX_RECV_STRING_OPTIONAL(ret, mca_base_component_to_string(&pml_base_component),
-                                    &rank0, (void**) &remote_pml, &size);
-
-    /* if this key wasn't found, then just assume all is well... */
+    OPAL_MODEX_RECV_STRING(ret,
+                           mca_base_component_to_string(&pml_base_component),
+                           &proc_name, (void**) &remote_pml, &size);
     if (PMIX_ERR_NOT_FOUND == ret) {
         opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
-                            "check:select: PML modex for vpid 0 data not found");
-        return OMPI_SUCCESS;
+                            "check:select: PML modex for process %s not found",
+                            OMPI_NAME_PRINT(&proc_name));
+        return OMPI_ERR_NOT_FOUND;
     }
 
     /* the remote pml returned should never be NULL if an error
@@ -359,22 +355,26 @@ mca_pml_base_pml_check_selected(const char *my_pml,
      */
     if (NULL == remote_pml) {
         opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
-                            "check:select: got a NULL pml from rank=0");
+                            "check:select: got a NULL pml from process %s",
+                            OMPI_NAME_PRINT(&proc_name));
         return OMPI_ERR_UNREACH;
     }
 
     opal_output_verbose( 10, ompi_pml_base_framework.framework_output,
-                        "check:select: checking my pml %s against rank=0 pml %s",
-                        my_pml, remote_pml);
+                        "check:select: checking my pml %s against process %s"
+                        " pml %s", my_pml, OMPI_NAME_PRINT(&proc_name),
+                        remote_pml);
 
     /* if that module doesn't match my own, return an error */
     if ((size != strlen(my_pml) + 1) ||
         (0 != strcmp(my_pml, remote_pml))) {
-        char *errhost = opal_get_proc_hostname(&procs[0]->super);
+        char *errhost;
+        OPAL_MODEX_RECV_VALUE_OPTIONAL(ret, PMIX_HOSTNAME, &proc_name,
+                                       &(errhost), PMIX_STRING);
         opal_output(0, "%s selected pml %s, but peer %s on %s selected pml %s",
                     OMPI_NAME_PRINT(&ompi_proc_local()->super.proc_name),
-                    my_pml, OMPI_NAME_PRINT(&procs[0]->super.proc_name),
-                    errhost,
+                    my_pml, OMPI_NAME_PRINT(&proc_name),
+                    (NULL == errhost) ? "unknown" : errhost,
                     remote_pml);
         free(remote_pml);
         free(errhost);
@@ -384,4 +384,39 @@ mca_pml_base_pml_check_selected(const char *my_pml,
 
     free(remote_pml);
     return OMPI_SUCCESS;
+}
+
+int
+mca_pml_base_pml_check_selected(const char *my_pml,
+                                ompi_proc_t **procs,
+                                size_t nprocs)
+{
+    int ret = 0;
+    size_t i;
+
+    if (!opal_pmix_collect_all_data) {
+        /*
+         * If direct modex, then compare our PML with the peer's PML
+         * for all procs
+         */
+        for (i = 0; i < nprocs; i++) {
+            ret = mca_pml_base_pml_check_selected_impl(
+                                                 my_pml,
+                                                 procs[i]->super.proc_name);
+            if (ret) {
+                return ret;
+            }
+        }
+    } else {
+        /* else if full modex compare our PML with rank 0 */
+        opal_process_name_t proc_name = {
+                           .jobid = ompi_proc_local()->super.proc_name.jobid,
+                           .vpid = 0
+        };
+        ret = mca_pml_base_pml_check_selected_impl(
+                                                 my_pml,
+                                                 proc_name);
+    }
+
+    return ret;
 }


ompi/mca/pml/base/pml_base_select.c
```diff
@@ -359,22 +348,26 @@ mca_pml_base_pml_check_selected(const char *my_pml,
...
+                        "check:select: checking my pml %s against process %s"
+                        " pml %s", my_pml, OMPI_NAME_PRINT(&proc_name),
+                        remote_pml);
 
     /* if that module doesn't match my own, return an error */
     if ((size != strlen(my_pml) + 1) ||
         (0 != strcmp(my_pml, remote_pml))) {
-        char *errhost = opal_get_proc_hostname(&procs[0]->super);
+        char *errhost;
+        OPAL_MODEX_RECV_VALUE(ret, PMIX_HOSTNAME, &proc_name,
```
rhc54: This needs to be an "optional" recv as the hostname is not always available.

dipti-kothari: Yes, ill make it optional. Thank you

bwbarrett:
There was an issue with UCX and this patch (https://github.com/open-mpi/ompi/issues/7570), but George has fixed it already.

jladd-mlnx:
@artpol84 @karasevb guys, please review this. Need to ensure it doesn't break direct modex performance for UCX PML.

jladd-mlnx:
@bwbarrett - I believe your concerns have been addressed with https://github.com/open-mpi/ompi/pull/7575

artpol84:
Update:
With @karasevb we were trying to evaluate this fix at runtime and got hold by https://github.com/openpmix/prrte/issues/505.
We will apply a temp fix in the PRRTE to fix the key in question so we can proceed with GDS/ds21 component and will post the results later.

ompi/mca/pml/base/pml_base_select.c
```diff
@@ -385,3 +378,38 @@ mca_pml_base_pml_check_selected(const char *my_pml,
...
+{
+    int ret = 0;
+    size_t i;
+
+    if (opal_pmix_base_async_modex) {
+        /*
+         * If direct modex, then compare our PML with the peer's PML
+         * for all procs
+         */
+        for (i = 0; i < nprocs; i++) {
```
artpol84, line 395: I don't think this is a good solution. With this, you are hitting the negative sides of both approaches:
* still all the blobs are collected as you are requesting the data from all of them. Technically this will be a full modex thus all of the memory footpring concerns.
* on the other hand instead of doing it in an efficient collective fashion, it will be done using slow one-by-one direct modex fetching.

Given all of that, here are my thoughts:
* With this change and prior, in case of a full modex, everyone compares to `rank == 0`. So some of procs might be ok, some not.
* What if, in case of direct modex, the comparison will be performed prior to connection establishment? This will delay the detection of an error to some undefined time, but will ensure that we only fetch the data we need for a particular communication graph.

rhc54:
> It is crystal clear that each process will fetch a particular key from every other rank which will result in a full modex implemented over direct modex.

Urrr...I don't think that is correct. The "nprocs" in this function is the number of procs that are being added by the "add_procs" call, not the number of procs in the entire job. We only add procs as we initialize communication with them, and so this function is looking at the keys returned by a direct modex call for information on a proc we want to talk to.

If someone is going to call "add_procs" on the entire job, then they should be using the full modex, not direct modex.

karasevb:
I've got a hang and then exit some processes by timeout when I tried to enable the direct-modex (`opal_pmix_base_async_modex=1` and `opal_pmix_collect_all_data=0`).  The probability of reproduction increases with ppn. It reproduces with force enabled `ob1` pml.

Async modex removes the barrier between PMIx_Put and PMIx_Get of pml keys.
I observe that in case of the hang the hanging process requests the key prior to it's being published. We see that this client sends a request to the server, but never gets the response. That leads to exit by timeout.

I'm still investigating.

artpol84:
Per telecon,

* @dipti-kothari we don't think that the issue @karasevb is reporting is related to this patch. We are going to track this on our end and update here. We still want to finish the runtime evaluation before providing the final assessment.
* My comment about "full modex over direct modex" is incorrect as (as @rhc54 pointed out) only a handy of processes are added in add_proc and not all of them.

artpol84:
@karasevb please proceed with the analysis we discussed:
* Have one client forced to be "fast"
* Slowdown another client
* See how the server process this situation.

dipti-kothari:
Any update @artpol84 @karasevb

karasevb:
@dipti-kothari I found at least two issues which do not give to evaluate this PR regarding direct modex. The first one is investigated: https://github.com/openpmix/openpmix/pull/1715, but the second is in progress. Will update the results.

karasevb:
I compared the timings values of the issued changes. I used OMPI timing framework (`--enable-timing`) and added extra timings to identify possible influences on direct-modex. Here is results for 4 nodes 28 ppn with direct-modex:

  | master | PR7582 | master + patch(PR7582)
-- | -- | -- | --
[ompi_mpi_init.c:ompi_mpi_init:mca_pml_base_select] | 0.485694 | 0.514275 | 0.491299
[ompi_mpi_init.c:ompi_mpi_init:modex-add_procs] | 0.229029 | 0.22014 | 0.218848
[ompi_mpi_init:total] | 0.790873 | 0.833407 | 0.797475

All results were obtained in 50 iterations and the average value was calculated. If we compare master branch results against this PR ( #7582 ) the latter is worse by 0.028 sec in `mca_pml_base_select`. But I tried to apply this PR as a patch for the master and got the same results as with the original master.

dipti-kothari:
@karasevb For direct modex this PR ( #7582 ) will publish PML for all procs in mca_pml_base_select.

artpol84:
@jladd-mlnx we double-checked with a scaling test and it looks like everything works as expected and we don't see much of a deviation between the master with and without this PR:

image

@karasevb thinks that it is good to go, I feel the same.

jladd-mlnx:
@dipti-kothari - is this the expected scaling?

rhc54:
I think the answer depends on what this test is actually measuring. Can someone perhaps clarify what is being reported? The meaning of those curves plus some info on what the MPI app is doing (are all procs connecting to every other proc, or to only one, or...) would help.

I ask because that timing looks awfully linear...

artpol84:
@rhc54 I will need to check with @karasevb

artpol84:
In the meantime, @dipti-kothari what test are you using to obtain this?
My guess for the plots from @karasevb would be that it is either hello_c or ring_c example from OMPI repo. But I'll double check.

NOTE that on the first plot it is PPN scaling, so it is probably expected to be linear as all procs on the node will be added.
On the second plot though, the nodes are on x-axis and it's not very good as well:
image

artpol84:
One more comment:
what scales bad is "pml_base_select". IIUC the optimization supposed to improve add_procs, right?

artpol84:
Bad scaling of "pml_base_select" could be the result of NFS storm when every process tries to open components files.

dipti-kothari:
@artpol84 I have tested both full modex and direct modex with MPI_Comm_spawn test from mpich, osu_alltoall and pt2pt tests (osu_bw  and osu_latency). I can post the numbers.
Also, regarding the above results, they are for direct modex, right?

rhc54:
For me, it would help if I knew what the curves actually meant. The curve labeled "mca_pml_base_select" - what is that actually measuring? The time it took to execute that function? If so, then sure - it will scale linearly with the number of procs being "added". The "hello_c" code shouldn't be adding any procs, so the curve makes no sense if that was used - it should be zero. On the other hand, "ring_c" only adds the procs on either side - again, the curve would make no sense for that case (should be a constant). So something doesn't add up unless the curve is measuring something else.

Likewise, what precisely is the curve labeled "modex-add_procs" representing?

karasevb:
Plots which are presented were obtained for direct-modex with `hello_c` program.

rhc54:
@karasevb Could you please clarify what you mean by the "mca_pml_base_select" time? Is that the time spent in that one function? Likewise, what does the "modex-add_proc" time mean?

I'm having trouble understanding the graph because (as I noted above) hello_c does not add anything other than the local procs. So there is no dmodex operations happening in that scenario. I would expect the time spent doing the pml check to be linear in ppn but independent of the number of nodes. Likewise for add_procs.

Yet that isn't what your graphs show. So either I'm not understanding what was measured, or we aren't measuring what we think we are, or there is something wrong with the code. Regardless, running hello_c cannot tell us anything about dmodex scaling.

karasevb:
@rhc54 
> Could you please clarify what you mean by the "mca_pml_base_select" time? Is that the time spent in that one function? Likewise, what does the "modex-add_proc" time mean?

Yes, this is time spent in the function.

The main idea of this measurements is to check influence of this PR on the direct modex. Since PR changes the logic of the functions `mca_pml_base_select` and `modex-add_proc` in relation to direct-modex, I added timings for this and checked how it scales with the number of nodes and ppn.

artpol84:
@rhc54 
In order to collect the data, @karasevb updated the timings we have in OMPI (when configured with "--enable-timing" ) to measure 
1. The time it takes to "select PML" [[see here](configured)]. This portion wasn't really affected by this PR and (as I've said before) IMO has linear scaling, because PML selection logic opens PML DSO's on the NFS, thus we observe linear scaling.
1. The time it takes to execute PML add_proc [[see here](https://github.com/karasevb/ompi/blob/timings_7582/ompi/runtime/ompi_mpi_init.c#L841)]. This function was the most affected by this PR and on the plots (labeled as modex-add_proc) it has a pretty flat scaling, so IMO they act as expected.

karasevb:
@dipti-kothari could you please clarify, what scale and ppn did you test? And what is speedup benefit did you got for that scale? Do you use `pmix_base_async_modex=1` with `pmix_base_collect_data=0` to ensure direct modex?

rhc54:
@karasevb @artpol84 Thank you for the explanations of the measurements. However, I must point out again that your use of "hello" as the test program means that you actually didn't measure anything relating to this PR. Let me try to explain why.

When we added support for async modex, OMPI was changed so that:

* it only calls add_procs on local procs during MPI_Init
* it calls add_procs on each remote process on first communication with that process

The "hello" program does not include any communication. Thus, OMPI will not call "add_procs" on any remote processes (i.e., any process not on the local node).

The dmodex operation is _only_ called when OMPI requests information about a _remote_ process. Thus, since "hello" never attempts to communicate with a remote process, the dmodex operation is never executed.

Your graph therefore only shows what one would expect - that the calling of "add_procs" on local procs should scale roughly linearly with the ppn (with a very small slope as this is a fast operation), but not be at all affected by the number of nodes.

However, this doesn't address the question: does this PR cause an impact on dmodex? Does it force additional overhead and/or data retrieval?

I fully expect the answer to be "no" if this was implemented correctly as the PML information should be included in the dmodex return blob and be checked after receiving it. There should specifically not be any extra dmodex to a specific proc to obtain its PML selection as there was before this PR. However, that is what we were seeking to confirm.

Hope that helps.

rhc54:
After spending some time at the whiteboard, I _think_ this data may be telling us enough to commit this PR and move on, even though it is an indirect measurement. Let me talk thru it and let's see if folks concur with the explanation.

Prior to this PR on master, OMPI was not checking the PML selection between procs at all unless someone set the appropriate flag. The original form of the PR was simply to force that check to occur for all procs. Unfortunately, at that time, only rank=0 was posting its PML selection. This was okay for full modex operations as the data was always there, but caused a major problem for async modex operations as it made every process execute a dmodex request back to rank=0.

The PR was subsequently modified to its current form. In async modex mode, every proc posts its PML selection. When proc A first communicates to a remote peer (call it B), a dmodex request is generated for that peer's posted information. The returned blob includes B's PML selection, which is then checked against A's PML selection during add_procs.

Thus, with this PR, the PML selection is always cross-checked during add_procs. True, it is delayed until first communication (instead of occurring upon MPI_Init), but this was judged to be an acceptable situation as different PML selections are rare errors. The modified PR should not introduce any additional dmodex operations.

If we look at the graphs, we don't see "hello" showing any new dmodex operations during MPI_Init. The behavior of master and PR are identical, indicating that the PR is not causing add_procs of the local procs to seek the PML info posted by rank=0.

The graphs don't address the question of what happens on first message - does everyone flood back to rank=0 to get the PML info? Or do they correctly just get the PML info of the proc they are trying to reach? This could be tested by running a communicating app (e.g., ring) and verifying that master and PR are identical (as the PML info should be present in the dmodex from the target proc).

I think we can probably verify that by simple inspection of the code - I'm not sure we really need to rerun these tests. Hopefully, though, this clarifies the questions and interpretation of the data.

karasevb:
I checked the PML check feature deeper, the current master does not PML check correctly opposite this PR: [OPAL_MODEX_RECV_STRING_OPTIONAL](https://github.com/open-mpi/ompi/blob/master/ompi/mca/pml/base/pml_base_select.c#L346) uses to request PML from rank 0. Master checks PML as `OPTIONAL` that means for direct modex will not send a request to rank 0, for remote ranks. This PR is fixes that.
I also confirm that this PR has the same behavior as the master in relation to the direct modex. I checked, at the first communication to the node A, the proc at node B receives the PML and other process data within a blob that received through a single direct modex request. Thus `add_procs` at the first communication performs a single dmodex req that shows direct modex works correctly for this PR.

dipti-kothari:
Time taken by mca_pml_base_select and mca_pml_cm_add_procs collected during ring_c test on 2, 4, 8, and 16 nodes respectively with 4ppn, averaged over 50 iteration. @karasevb I have collected the below data with ```pmix_base_async_modex=1``` and ```pmix_base_collect_data=0```
@jladd-mlnx: I can see similar results, as produced by @karasevb. There is not much deviation between the master with and without this PR.
@rhc54: Are the numbers as expected? Do we need more data?

Screen Shot 2020-04-20 at 12 05 28 AM

Function \ Number of nodes | 2 | 4 | 8 | 16
-- | -- | -- | -- | --
mca_pml_base_select_with_PR | 0.2508 | 0.487 | 1.0032 | 2.0026
mca_pml_base_select_without_PR | 0.2506 | 0.4998 | 1.006 | 1.9752
mca_pml_add_procs_with_PR | 0.561 | 1.246 | 3.1138 | 8.5374
mca_pml_add_procs_without_PR | 0.56 | 1.2614 | 3.1056 | 8.5346

I wasn't able to test for higher scale, as I was hitting failure in pmix with ompi master with/without the PR.

rhc54:
That is...absolutely awful. You are showing essentially quadratic scaling. I think someone needs to take a deeper look to determine exactly what is going on in these operations - I don't see how we can release an OMPI with quadratic launch scaling.

If you think about it, a ring_c program means that each proc (let's call it rank N) should only be adding the procs to either side of it (N+1 and N-1). So "add_procs" should at most require a proc to do two dmodex requests, regardless of the size of the job.

What you also should see is that all but two procs (the lowest rank on a node, and the highest rank on the node) will resolve their "add_procs" without doing a dmodex at all because the N-1 and N+1 procs are on the same node. Thus, looking at this from the node level, each node should be executing exactly 2 dmodex requests, regardless of the ppn being launched.

Thus, the curve should be a constant - there should be no dependency on the number of nodes or the total number of procs in the job. An individual proc will either get its "add_proc" info immediately (as it will be local), or it will require a single dmodex operation (as the lowest rank needs to get info about the highest rank on the preceding node, and the highest rank needs to get info on the lowest rank of the next node).

What looks like is happening is that every process is going after info for _every_ proc in the job. Definitely an error somewhere in the code.

rhc54:
It would also be good to know more about this error in PMIx you hit - can you please provide more info on it?

jladd-mlnx:
@rhc54 - fully agree. This is not an acceptable. @dipti-kothari - would it be possible to make the same set of measurements with OMPI v4.0.3?

dipti-kothari:
> It would also be good to know more about this error in PMIx you hit - can you please provide more info on it?

Created an issue for the error https://github.com/open-mpi/ompi/issues/7647

dipti-kothari:
@rhc54 I revisited my testing scripts and found an issue. Total Time taken in mca_pml_cm_add_procs is quite less. 

Total time \ Number   of Nodes | 2 | 4 | 8 | 16
-- | -- | -- | -- | --
Total time in mca_pml_cm_add_procs with   PR | 0.0026 | 0.0058 | 0.0112 | 0.021
Total time in mca_pml_cm_add_procs   without PR | 0.0032 | 0.0064 | 0.0122 | 0.027
image

Also, I should have been more specific earlier, I have collected the Total time in mca_pml_cm_add_procs summed over all nodes. For 2 nodes
```
Total time taken by mca_pml_cm_add_procs = time taken by mca_pml_cm_add_procs Node 1
                                           +time taken by mca_pml_cm_add_procs Node 2
```
Taking an average over the nodes, gives constant time.

Time per Node \   Number of Nodes | 2 | 4 | 8 | 16
-- | -- | -- | -- | --
Time in mca_pml_cm_add_procs per Node   with PR | 0.0013 | 0.00145 | 0.0014 | 0.0013125
Time in mca_pml_cm_add_procs per Node   without PR | 0.0016 | 0.0016 | 0.001525 | 0.0016875
image

ompi/mca/pml/base/pml_base_select.c
```diff
@@ -385,3 +378,38 @@ mca_pml_base_pml_check_selected(const char *my_pml,
...
+
+int
+mca_pml_base_pml_check_selected(const char *my_pml,
+                                ompi_proc_t **procs,
+                                size_t nprocs)
+{
+    int ret = 0;
+    size_t i;
+
+    if (opal_pmix_base_async_modex) {
```
karasevb, line 390: The direct modex means that we will not exchange all data at startup, so the correct condition here is to compare `opal_pmix_collect_all_data` instead of `opal_pmix_base_async_modex`. From my vision we need ro replace the all cases for this PR to use `opal_pmix_collect_all_data`.

rhc54: I'm not sure I understand what you are suggesting. Async modex means "don't perform the fence", which means we will use the direct modex method for obtaining endpts.

Collect all data tells the fence operation to include endpt data instead of being a zero-byte barrier operation. Thus, _if_ you perform a fence (i.e., `async_modex` is false) _and_ you tell the fence not to collect the data, then you will still have to perform a direct modex to get the endpt info on first communication. You just forced a synchronizing barrier into MPI_Init. Some of the transports need that barrier even if they still have to do the direct modex.

So given that understanding, are you suggesting that this test needs to be changed because it really depends on whether or not data was exchanged, and not on whether the fence was executed? I'm just seeking clarity, not questioning the suggestion.

rhc54:
@dipti-kothari I confess to still being a little confused about what you are saying. IIUC, you are saying that the original graphs were reporting the total time (added up across all procs) spent in add_procs as opposed to the time any one proc spent in it - is that correct?

If so, then yes - that would explain the dependencies shown in the graphs.

IIUC, your new graph shows the average time spent in add_procs by any one process - is that correct? If so, then this is much, much better and exactly what I would have expected to see.

ompi/mca/pml/base/pml_base_select.c
karasevb: My point is we shouldn't rely on `opal_pmix_base_async_modex` to identify that direct modex is on. As @rhc54 mentioned `async_modex=0` and `collect_data=0` then direct modex will be performed. So my suggestion is to use `if (!opal_pmix_collect_all_data)` to identify dmodex will be used.

dipti-kothari:
@rhc54 Yes, I had reported the total time added over all procs. In the second graph I have averaged over number of nodes rather than number of procs. But averaging the time over number of nodes or number of procs would give constant output as expected. I have also verified that its not all to all.

ompi/mca/pml/base/pml_base_select.c
dipti-kothari: @rhc54 @karasevb Do we need to use opal_pmix_collect_all_data to decide whether to use dmodex or not?

rhc54: Let me provide you with the decision tree that guides behavior:

* if async_modex is true:
    * if collect_all_data is false, then we will _not_ perform a fence operation during MPI_Init and we need to use dmodex during add_procs
    * if collect_all_data is true, then we perform a non-blocking fence operation during MPI_Init, waiting for it to complete only at the very end of MPI_Init. In this case, all data has been collected and we do _not_ need to use dmodex

* if async_modex is false, then we _always_ perform a blocking fence operation during MPI_Init
    * if collect_all_data is false, then this is a zero-byte barrier and we need to use dmodex during add_procs
    * if collect_all_data is true, then all data will be collected during the fence and we do _not_ need to use dmodex

So the answer to your question is that there are two times we have to use dmodex:
* async_modex is true and collect_all_data is false
* async_modex is false and collect_all_data is false

Thus, checking for collect_all_data being false should suffice to tell you that dmodex is going to be used.

dipti-kothari: Thank you @rhc54. I have updated the patch

Issue #7583, opened 2020-03-31
Cannot compile on Mac with Nag 6.2


pbrady:
I'm on MacOS Catalina and attempting to use Nag6.2 as the fortran compiler to build Open MPI v4.0.3 (source tarball).  I have tried using the system clang compilers as well as gcc/g++ 9.3 from homebrew.  If I use gfortran from homebrew then open mpi compiles without issue so Nag appears to be the culprit here. 

The configure step appears to be successful and I've attached the [config.log](https://github.com/open-mpi/ompi/files/4411489/config.log).

`make` is able to run for quite some time but I keep getting the following error:
```shell
Making all in mpi/fortran/use-mpi-tkr
  PPFC     mpi.lo
NAG Fortran Compiler Release 6.2(Chiyoda) Build 6255
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi.F90, line 23: Invalid #line directive
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi.F90, line 25: Syntax error
       detected at <invalid token>@MODULE
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi.F90, line 46: Invalid #line directive
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi.F90, line 47: Invalid #line directive
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi.F90, line 47: Syntax error
Error: ../../../../../ompi/mpi/fortran/use-mpi-tkr/mpi-f90-interfaces.h, line 27: Syntax error
       detected at END@FUNCTION
```

I get the same errors when trying to compile Open MPI v3.1.5.

Any advice on how to fix this would be greatly appreciated.

ggouaillardet:
On Linux, `.f90` files are compiled with the "standard" flags, but `.F90` files are preprocessed before being compiled.
Both error message and the OS X case insensitive filesystem points to the `.F90` not being preprocessed on OS X.

Assuming any fortran file can be preprocessed by NAG compiler via the `-fpp` flag, I would suggest you try
```
configure FCFLAGS=-fpp ...
```
and see if it helps

pbrady:
Was able to compile and successfully run `make check` with the added `-fpp` option.  I'll leave this issue open because this seems like something that should be handled by the build system.  Feel free to close if you disagree.

ggouaillardet:
I agree this is something that should be handled automatically.

[...]

PR #7584, opened 2020-04-01
configury: try if -fpp flag is needed to preprocess .F90 files

ggouaillardet:
.F90 files are preprocessed by gfortran and other compilers.
NAG compilers only preprocess .{ff,ff90,ff95} files, and the -fpp
flag is required to process .F90 files.

Fixes open-mpi/ompi#7583


jsquyres:
LGTM.  Let's see what @pbrady says on #7583 before merging.

PR #7593, opened 2020-04-02
Fortran: fix the F90 compiler preprocessor check

jsquyres:
Only check the if the Fortran compiler needs additional CLI flags for
preprocessing .F90 files if we actually have an F90 compiler.

Also fix a the AC_MSG_* usage.


jsquyres:
This will fix the failing Cray CI (because at least in some cases, it's not compiling with a Fortran compiler).  I'm not sure how Cray CI passed when we did the original PR -- it must have run with a Fortran compiler that time...?  🤷‍♂

PR #7599, opened 2020-04-05
MPI_T general docs

### LINKED ISSUE #7519: Document some Open MPI-specific MPI_T behavior
Need to document the following:

* Values set via MPI_T are the highest precedence (e.g., even higher that values set by CLI).
* If you set framework values (e.g., set `btl` before `MPI_INIT`, it'll actually affect component selection).
* We have a CVAR per MCA framework that will tell you which component(s) are currently selected / available (after `MPI_INIT`).
* After `MPI_INIT`, most CVARs become read-only.
* There is no CVAR that lists all the frameworks.
* Groups are related projects, frameworks, components.
* MCA variables *are* CVARs.
* As of this version, very few PVARs.

This can probably go in a man page somewhere, and/or the FAQ...?

FYI @hjelmn

jsquyres:
Document Open MPI's MPI_T implementation.  Also do it in Markdown rather than nroff -- it's so much easier to read/write.

Fixes #7519

### DIFF
--- a/ompi/mpi/man/man3/MPI_T_init_thread.3.md
+++ b/ompi/mpi/man/man3/MPI_T_init_thread.3.md
@@ -1,0 +1,65 @@
+# NAME
+
+MPI_T_init_thread - Initializes the MPI Tool information interface
+
+# SYNTAX
+
+## C Syntax
+
+```c
+#include <mpi.h>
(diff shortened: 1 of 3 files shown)


jsquyres:
Shoot -- I meant to make this a draft PR...

Issue #5115, opened 2018-04-27
attach fifo deleted on exit and not recreated on reattach


lee218llnl:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v1.10.3, v2.1.0, git branch name and hash, etc.)
v 2.1.3, 3.0.1, and git master at commit 0552d127714b4b2b101236eebcf6e551238f68cd

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
from source

### Please describe the system on which you are running

* Operating system/version: RHEL 7
* Computer hardware: x86
* Network type: psm2

-----------------------------

## Details of the problem

@rhc54 and @dongahn, I thought we had all the debugging kinks worked out, but in testing I notice that I have trouble when launching a job, attaching a debugger, detaching a debugger, and then reattaching the debugger. From what I can see, orterun deletes the debugger_attach_fifo when the debugger daemon exits, but then does not recreate it upon reattach.

Here's how you can reproduce

```
git clone https://github.com/llnl/launchmon.git
cd launchmon/
# modify test/src/be_kicker to set the "kill_detach_shutdown_time = 6"
CFLAGS="-g -O0" CXXFLAGS="-g -O0" ./configure --prefix=/nfs/tmp2/lee218/prefix/launchmon-1.0.3b --with-test-rm=orte --with-test-rm-launcher=mpirun --with-test-installed --with-test-nnodes=1 && make clean && make -j 8 install && make -j 8 check
cd test/src
# modify test.attach_4_detach to re-run fe_attach_smoketest $PID `pwd`/be_kicker $SIGNUM 1
./test.attach_4_detach
```

The additional fe_attach_smoketest will hang.

rhc54:
Related to #5501

gpaulsen:
We should retest this with new MPIR fixes that went in lately.

@lee218llnl Could you please retest this with latest master and bundled pmix.

lee218llnl:
I was able to attach STAT to an OpenMPI job based on the master branch. I get this warning after attaching, which I imagine is expected:

Open MPI has detected that you have attached a debugger to this MPI
job, and that debugger is using the legacy "MPIR" method of
attachment.

Please note that Open MPI has deprecated the "MPIR" debugger
attachment method in favor of the new "PMIx" debugger attchment
mechanisms.

*** This means that future versions of Open MPI may not support the
[...]

PR #7609, opened 2020-04-07
debuggers: don't remove session dirctory

hppritcha:
when the exiting process is a debugger process.

relates to #5115


rhc54:
Although we don't support MPIR, I have ported this to PRRTE for correctness going forward.

PR #7621, opened 2020-04-13
Remove OSC pt2pt component

jsquyres:
Per https://github.com/open-mpi/ompi/wiki/5.0.x-FeatureList, remove
the OSC pt2pt component.


mwheinz:
Jeff, I thought Brendan Cunningham was going to do this?

jsquyres:
@mwheinz Quite possibly so -- happy to have him do it, if you prefer.  We were talking through RM stuff today and saw that this wasn't done, and there was no other PR filed for it, and at least this part  was trivial to do.  So I filed this PR.  😄 

No need to merge it until we get closer to April 30, so if there's other cleanup that needs to be done for PSM/PSM2, that came either come in first and/or replace this PR.

hjelmn:
@bosilca The datatype caching is not used by osc/rdma because I never bothered using adding and active-message support. I disagreed with the assertion some have made that it is useful. We can probably clean out all the datatype caching code.

PR #7632, opened 2020-04-15
UCX osc: make progress on idle worker if none are active

### LINKED ISSUE #7631: UCX osc: missing progress in shared memory
I encountered the following problem when forcing UCX with `--mca osc ucx` on a shared memory system (mainly for testing; I know it's not optimal but it shouldn't hang):

```C
MPI_Win win;
MPI_Win_allocate(..., &win);
if (myrank == 1) {
  MPI_Win_lock_all(win); // <- hangs
}
MPI_Barrier(MPI_COMM_WORLD);
```

Process 1 will hang in `MPI_Win_lock_all` because `MPI_Barrier` will not progress UCX (presumably because UCX is not used for collectives in shared memory?). The UCX osc's progress callback gets called in the barrier call but since there are no active workers in process 0 it will not progress the operations required by `MPI_Win_lock_all` on process 1.

I have a possible fix that I will PR soon.

devreal:
When using UCX in shared memory progress is missing if one process waits in a barrier while the another process attempts to perform RMA operations. This PR adds progress on the first inactive workers if there are no active workers available.

Fixes #7631

### DIFF
--- a/opal/mca/common/ucx/common_ucx_wpool.c
+++ b/opal/mca/common/ucx/common_ucx_wpool.c
@@ -59,7 +59,7 @@ _winfo_create(opal_common_ucx_wpool_t *wpool)
         goto exit;
     }
 
-    winfo = calloc(1, sizeof(*winfo));
+    winfo = OBJ_NEW(opal_common_ucx_winfo_t);
     if (NULL == winfo) {
         MCA_COMMON_UCX_ERROR("Cannot allocate memory for worker info");
         goto release_worker;
@@ -194,9 +194,10 @@ opal_common_ucx_wpool_init(opal_common_ucx_wpool_t *wpool,
         rc = OPAL_ERROR;
         goto err_worker_create;
     }
-    wpool->dflt_worker = winfo->worker;
+    wpool->dflt_winfo = winfo;
+    OBJ_RETAIN(wpool->dflt_winfo);
 
-    status = ucp_worker_get_address(wpool->dflt_worker,
+    status = ucp_worker_get_address(wpool->dflt_winfo->worker,
                                     &wpool->recv_waddr, &wpool->recv_waddr_len);
     if (status != UCS_OK) {
         MCA_COMMON_UCX_VERBOSE(1, "ucp_worker_get_address failed: %d", status);
@@ -214,8 +215,10 @@ opal_common_ucx_wpool_init(opal_common_ucx_wpool_t *wpool,
 err_wpool_add:
     free(wpool->recv_waddr);
 err_get_addr:
-    if (NULL != wpool->dflt_worker) {
-        ucp_worker_destroy(wpool->dflt_worker);
+    if (NULL != wpool) {
+        OBJ_RELEASE(winfo);
+        OBJ_RELEASE(wpool->dflt_winfo);
+        wpool->dflt_winfo = NULL;
     }
  err_worker_create:
     ucp_cleanup(wpool->ucp_ctx);
@@ -233,7 +236,7 @@ void opal_common_ucx_wpool_finalize(opal_common_ucx_wpool_t *wpool)
 
     /* Release the address here. recv worker will be released
      * below along with other idle workers */
-    ucp_worker_release_address(wpool->dflt_worker, wpool->recv_waddr);
+    ucp_worker_release_address(wpool->dflt_winfo->worker, wpool->recv_waddr);
 
     /* Go over the list, free idle list items */
     if (!opal_list_is_empty(&wpool->idle_workers)) {
@@ -258,6 +261,9 @@ void opal_common_ucx_wpool_finalize(opal_common_ucx_wpool_t *wpool)
     }
     OBJ_DESTRUCT(&wpool->active_workers);
 
+    OBJ_RELEASE(wpool->dflt_winfo);
+    wpool->dflt_winfo = NULL;
+
     OBJ_DESTRUCT(&wpool->mutex);
     ucp_cleanup(wpool->ucp_ctx);
     return;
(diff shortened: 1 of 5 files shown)


opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -308,6 +309,13 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
             }
             opal_mutex_unlock(&winfo->mutex);
         }
+        if (active_workers == 0 && opal_list_get_size(&wpool->idle_workers)) {
+            // make sure to progress at least some
+            // no need to lock the winfo here, it is idle and we have the wpool mutex
+            _winfo_list_item_t *item = (_winfo_list_item_t*)opal_list_get_first(&wpool->idle_workers);
+            opal_common_ucx_winfo_t *winfo = item->ptr;
+            ucp_worker_progress(winfo->worker);
```
artpol84, line 317: Does it make sense to call opal progress instead?

devreal: This is in the progress function that is called by `opal_progress`. It's important to progress UCX itself here.

artpol84: Can you elaborate more on why it is important to progress idle worker?

devreal: Here is what I believe is going on: when I run with `-mca osc ucx` on a shared memory system, some operations require involving the target process, such as for AMOs used when locking a window. The target window is not aware of its window being locked, so all workers at the target are idle if the target does not have active operations going on itself. However, the AMO needs to be processed somehow (since it is not offloaded to any third-party entity such as the NIC) so occasionally the target process has to call into UCX to find the AMO request and process it. The only way to do so without unintended side effects was to progress an idle worker. There might be better ways to handle unexpected incoming RMA operations but I am not aware of them.

artpol84: Then you need to progress a worker that is being used as a target and not a random worker.
I think it just happens to be the first (and only) worker in the idle list.
But this may not be the case always.
See [[link](https://github.com/open-mpi/ompi/blob/master/ompi/mca/osc/ucx/osc_ucx_comm.c#L256)] for the worker I believe you really want to progress

devreal: Got it, thanks! I changed the commit to progress the `dflt_worker` instead.

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -308,6 +309,10 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
             }
             opal_mutex_unlock(&winfo->mutex);
         }
+        if (active_workers == 0 && opal_list_get_size(&wpool->idle_workers)) {
```
artpol84, line 312: Actually, I think we need to progress it in any case.
It might happen that it's in the "idle" list even if we have other workers non-idling.

artpol84: We are currently doing a significant refactoring of this code BTW. 
But this is a good hint for us. So thank you very much.
For now, I think, we can just always progress this one. But with locking precautions.

devreal: The `wpool->mutex` is held at this point, that should be sufficient, right?

devreal:
@artpol84 Is there anything left to do here?

lanl-ompi:

devreal:
@artpol84 @janjust I rebased this PR onto current master. I was surprised to see that the `wpool` mutex is now take unconditionally (https://github.com/open-mpi/ompi/pull/7632/files#diff-4277b36854f67c21224bff1f16f6aa090af4e3905c9e00ce96dc735f05c90200R275). This seems like an unnecessary synchronization point, threads could do other things if another thread is already progressing, no?

opal/mca/common/ucx/common_ucx_wpool.c
artpol84: @devreal we need to progress the default worker.

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -273,6 +273,7 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
      * TODO: may want to have some partitioning to progress only part of
      * workers */
     opal_mutex_lock(&wpool->mutex);
+    int active_workers = opal_list_get_size(&wpool->active_workers);
```
artpol84, line 276:
```suggestion
    int need_dflt_progress = 0;
```

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -282,6 +283,10 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
         } while (progressed);
         opal_mutex_unlock(&winfo->mutex);
     }
+    if (active_workers == 0 && opal_list_get_size(&wpool->idle_workers)) {
+        /* make sure to progress at least some */
+        ucp_worker_progress(wpool->dflt_worker);
+    }
```
artpol84, line 289:
```suggestion
    ucp_worker_progress(wpool->dflt_worker);
```

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -282,6 +283,10 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
         } while (progressed);
```
artpol84, line 283: We can skip the default worker here to avoid double-progress.

But we can't skip progressing the default worker if there are active workers. The default worker might still not be active, but you want to progress it.

artpol84 (changes requested):

artpol84:
@devreal lets sync if you still would like to discuss this.
I'm planning to merge this ASAP.

devreal:
@artpol84 I made a couple of changes: 

1) Store the `dflt_winfo` in the `wpool` instead of the `dflt_worker` to allow for locking it in case the default worker has to be progressed.
2) Avoid blocking on a taken mutex (which actually lead to deadlocks otherwise, presumably due to lock inversion).
3) Avoid progressing the `dflt_worker` directly in `osc/ucx` but instead call `opal_common_ucx_wpool_progress` due to ensure proper locking.
4) I found that the `winfo` objects are allocated using `calloc` but then released using `OBJ_RELEASE`. I don't see how that ever can have worked... I tired to fix it, please review carefully.

bosilca:
Calloc/Mallove + OBJ_CONSTRUCT is morally equivalent to OBJ_NEW. It can therefore be used with OBJ_RELEASE.

devreal:
@bosilca I am aware of that, but there was no `OBJ_CONSTRUCT` as far as I can see. Maybe I am just blind here...?

artpol84:
@devreal as I mentioned we have a PR ready with a lot of cleanups and stability fixes. Including more active use of opal objects.
@janjust please see if you’ll be able to rebase on top of this one.
If not- maybe you can port these changes instead keeping the original authorship of @devreal

opal/mca/common/ucx/common_ucx_wpool.c
```diff
@@ -272,17 +278,33 @@ opal_common_ucx_wpool_progress(opal_common_ucx_wpool_t *wpool)
     /* Go over all active workers and progress them
      * TODO: may want to have some partitioning to progress only part of
      * workers */
-    opal_mutex_lock(&wpool->mutex);
+    if (0 != opal_mutex_trylock(&wpool->mutex)) {
```
artpol84, line 281: We have this in @janjust pr already.
Maybe we should have merged it first :(

artpol84 (approved):
Overall looks ok.

Though please let @janjust to check what’s the easier sequence of commits in the context of upcoming refactoring.

janjust:
@devreal @artpol84 rebasing is fine, let's get it merged

janjust:
> @janjust please see if you’ll be able to rebase on top of this one.
> If not- maybe you can port these changes instead keeping the original authorship of @devreal

@devreal @artpol84 rebasing is fine, let's get it merged

artpol84:
Thanks a lot!

PR #7640, opened 2020-04-16
adding NUMA_RANK to process metadata

dancejic:
adding PMIX_NUMA_RANK info to process metadata so that the local NUMA
rank can be accessed through the opal_process_info object.

### DIFF
--- a/ompi/runtime/ompi_rte.c
+++ b/ompi/runtime/ompi_rte.c
@@ -12,6 +12,8 @@
  *                         reserved.
  * Copyright (c) 2019      Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
+ * Copyright (c) 2020      Amazon.com, Inc. or its affiliates.  All Rights
+ *                         reserved.
  * $COPYRIGHT$
  */
 #include "ompi_config.h"
@@ -70,6 +72,7 @@ pmix_process_info_t pmix_process_info = {
     .proc_session_dir = NULL,
     .my_local_rank = 0,
     .my_node_rank = 0,
+    .my_numa_rank = UINT16_MAX,     /* Assume invalid NUMA rank, set to UINT16_MAX */
     .num_local_peers = 0,
     .num_procs = 0,
     .app_num = 0,
@@ -777,6 +780,28 @@ int ompi_rte_init(int *pargc, char ***pargv)
         pmix_proc_is_bound = false;
     }
 
+    /* get our numa rank from PMIx */
+    if (pmix_proc_is_bound) {
+        OPAL_MODEX_RECV_VALUE_OPTIONAL(rc, PMIX_NUMA_RANK,
+                                       &pmix_process_info.my_name, &u16ptr, PMIX_UINT16);
+        if (PMIX_SUCCESS != rc) {
+            if (ompi_singleton) {
+                /* just assume the numa_rank is invalid, set to UINT16_MAX */
+                u16 = UINT16_MAX;
+            } else {
+                ret = opal_pmix_convert_status(rc);
+                error = "numa rank";
+                goto error;
+            }
+        }
+        pmix_process_info.my_numa_rank = u16;
+    } else {
+        /* If processes are not bound, the numa_rank is not available
+         * Assign UINT16_MAX to the numa_rank to indicate an invalid value
+         */
+        pmix_process_info.my_numa_rank = UINT16_MAX;
+    }
+
     /* get our local peers */
     if (0 < pmix_process_info.num_local_peers) {
         /* if my local rank if too high, then that's an error */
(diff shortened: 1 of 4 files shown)


bwbarrett:
ok to test

rhc54:
@dancejic PRRTE only assigns a NUMA_RANK if the procs are bound as otherwise they have no NUMA locality. These tests are failing because they specify `--bind-to none` and therefore the modex_recv for `PMIX_NUMA_RANK` fails to return a value.

You'll need to change your `if-else` logic after the modex_recv (perhaps assign a value of `UINT16_MAX` to represent unassigned) and then detect the numa_rank wasn't provided and do an alternative multi-NIC selection for those cases.

rhc54:
Alternative solutions occurred to me:

* I could always assign a numa_rank, setting the value to `UINT16_MAX` if the proc isn't bound to a NUMA domain
* I could always assign a numa_rank, setting the value equal to the `LOCAL_RANK` if the proc isn't bound

The latter would allow you to always use the same algo for assigning NICs, but might seem a little contrary to the definition of `numa_rank`. Any thoughts on how you'd like to go?

bwbarrett:
My preference would be to let Open MPI, rather than PMIx, figure out what to do if processes aren't bound.  I'm a little worried about using the lack of a value for the numa_rank key to directly make that decision.  Is there a more direct indicator that OMPI can use?  I'm not sure what I'd propose the OMPI code looks like yet; just trying to understand what data we have.

rhc54:
Hmmm...well, we do already query the proc's locality, so you should see that come back as "unbound":
```c
    /* identify our location */
    val = NULL;
    OPAL_MODEX_RECV_VALUE_OPTIONAL(rc, PMIX_LOCALITY_STRING,
                                   &pmix_process_info.my_name, &val, PMIX_STRING);
    if (PMIX_SUCCESS == rc && NULL != val) {
        pmix_process_info.cpuset = val;
        pmix_proc_is_bound = true;
    } else {
        pmix_process_info.cpuset = NULL;
        pmix_proc_is_bound = false;
    }
```
Would that suffice?

rhc54:
I should have clarified: that code snippet is taken from ompi_rte_init, and thus every process knows whether or not it is bound from that point forward (the `pmix_proc_is_bound` flag is a global value, but only defined at the ompi level - perhaps it should be moved to opal?).

dancejic:
I just tried using `pmix_proc_is_bound`. I think this should work for my purposes, I'll add a check for `pmix_proc_is_bound` before the modex_recv, and assign the numa_rank to 0 if it's not bound.

dancejic:
Looks like it grabbed another commit when I pushed, I'll fix that really quickly

rhc54:
Just want to clarify something. I approved the PR because it is technically correct so far as the integration to PMIx is concerned. However, I do believe that assigning the numa_rank to 0 if the proc isn't bound is a mistake.

The problem with that approach is two-fold. First, from a correctness perspective, assigning the numa_rank to a specific value implies that the proc is bound to that NUMA domain. In fact, it is not, and that might generate confusion elsewhere down the road as developer's won't realize that the value is bogus.

Second, assigning numa_rank to 0 in this case means you can no longer distinguish between an unbound proc (which can be anywhere on the node) and procs that are all bound to NUMA0. In the latter case, the user has clearly specified they want all the procs for this application to be bound to that specific location, likely because they are doing (or planning to do) something on the other NUMA domains. Thus, you probably don't want to round-robin the NIC assignments in that use-case as they may well interfere with an app using one or more of the other NUMA domains.

I would suggest that you instead set the numa_rank to `UINT16_MAX`, which is our default method for indicating an "invalid" value. This would server as an easy marker that the numa_rank was not set and distinguish it from the case where it is actually set to some value (including 0). It would also make it consistent with what we do with other values (e.g., vpid, jobid, ...) where an "invalid" value indicates that it was not set.

ompi/runtime/ompi_rte.c
```diff
@@ -777,6 +778,28 @@ int ompi_rte_init(int *pargc, char ***pargv)
...
+                error = "numa rank";
+                goto error;
+            }
+        }
+        pmix_process_info.my_numa_rank = u16;
+    } else {
+        /* if processes are not bound, numa_rank is not available
+         * assume numa_rank is 0
+         */
+        pmix_process_info.my_numa_rank = 0;
```
bwbarrett, line 800: I agree with Ralph, only stronger.  We should make this MAX_<datatype size>, not 0.

opal/util/proc.c
```diff
@@ -35,6 +35,7 @@ opal_process_info_t opal_process_info = {
     .proc_session_dir = NULL,
     .num_local_peers = 0,  /* there is nobody else but me */
     .my_local_rank = 0,    /* I'm the only process around here */
+    .my_numa_rank = 0,     /* Only process on the NUMA node */
```
bwbarrett, line 38: fix

ompi/runtime/ompi_rte.h
```diff
@@ -253,6 +253,7 @@ typedef struct {
     char *proc_session_dir;
     uint16_t my_local_rank;
     uint16_t my_node_rank;
+    uint16_t my_numa_rank;
```
bwbarrett, line 256: Add some documentation comment about this being MAX_? when not set.

opal/util/proc.h
```diff
@@ -108,6 +108,7 @@ typedef struct opal_process_info_t {
     char *proc_session_dir;             /**< Session directory for the process */
     int32_t num_local_peers;            /**< number of procs from my job that share my node with me */
     int32_t my_local_rank;              /**< local rank on this node within my job */
+    int32_t my_numa_rank;               /**< rank on this processes NUMA node */
```
bwbarrett, line 111: Add comment in here about being MAX_? when not set.

dancejic:
adding copyright info

PR #7641, opened 2020-04-16
Added multi-NIC support to provider selection

dancejic:
Adds the capability to select a NIC based on hardware locality.
Creates a list of NICs on the same NUMA node as the process,
then selects the NIC based on the (NUMA rank) % (number of NICs).
If no NICs are available on the same NUMA node, the selection process
will create a list of all available NICs and select based on
(NUMA rank) % (number of NICs)

### DIFF
--- a/opal/mca/common/ofi/common_ofi.c
+++ b/opal/mca/common/ofi/common_ofi.c
@@ -11,6 +11,7 @@
 
 #include "opal_config.h"
 #include "opal/constants.h"
+#include "opal/mca/hwloc/base/base.h"
 
 #include <errno.h>
 #include <unistd.h>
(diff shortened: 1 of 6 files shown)


awlauria:
ok to test

bwbarrett:
This is going to fail CI (and, actually, fail to compile) until https://github.com/open-mpi/ompi/pull/7640 is merged.  Putting the WIP marker on this until we merge 7640.

dancejic:
@bwbarrett could you remove the WIP label since 7640 is now merged

hppritcha:
related to #7646

opal/mca/common/ofi/common_ofi.c
```diff
@@ -25,3 +26,291 @@ int mca_common_ofi_register_mca_variables(void)
...
+        !check_ep_attr(provider_info->ep_attr, provider->ep_attr) &&
+        !(provider_info->caps & ~(provider->caps)) &&
+        !(provider_info->mode & ~(provider->mode))) {
+        return 0;
+    } else {
+        return OPAL_ERROR;
+    }
+}
+
+/* Takes a device BDF and returns it's NUMA node's logical index.
```
hkuno, line 91: TRIVIAL nit that may very well not be worth doing anything about -- typo in the comment. Should be "its".

opal/mca/common/ofi/common_ofi.c
```diff
@@ -25,3 +26,291 @@ int mca_common_ofi_register_mca_variables(void)
...
+ *                                  are found under the same NUMA node but multiple
+ *                                  valid providers are available.
+ *
+ * returns:
+ *     @param provider (OUT)        struct fi_info* object with the selected
+ *                                  provider if the selection succeeds
+ *                                  if the selection fails, returns the fi_info
+ *                                  object that was initially provided.
+ *
+ * If there is more than one provider that shares the same NUMA node, we use
```
mwheinz, line 209: What's the rationale behind using this approach?

bgoglin:
I am not sure looking for NUMA nodes is a good idea nowadays because there can be multiple NUMA nodes local to a PCI bus on recent machines.

Also get_process_numanode() assumes it will find an object with a NUMA node child by getting the largest object contained in the binding, this only works if you're bound at the NUMA level (or Package level if there's exactly one NUMA per package).

A better way would be: from the pcidev, use a hwloc_get_non_io_ancestor_obj() to get a CPU-side parent. From there, compare the cpuset of that object with the binding of your process. I'd say you should check whether the process binding intersects with the PCI device cpuset:
* binding can be smaller since procs are often bound to a single core
* binding can be larger if procs are bound to entire sockets while PCI devices are attached to subpackages on current Intel and AMD platforms.

By the way hwloc_get_pcidev_by_busid() only works if PCI devices is enabled, I don't remember if that's the case in OMPI.

rhc54:
Sigh - yeah, I realized after our discussing this week that assigning `PMIX_NUMA_RANK` really doesn't make much sense in today's CPU packages, especially now that we are completely removing the `bind-to numa` option. We can certainly enable PCI devices in the topology, and even generate something in the proc's job-level info that might be useful for this purpose.

Sadly, `numa rank` likely isn't it 😢

rhc54:
@bgoglin I've been searching, but I don't immediately see what I have to do to enable PCI devices in hwloc - can you steer me?

rhc54:
Duh - never mind! 🤷

rhc54:
FWIW: hwloc defaults to enabling PCI devices, and OMPI doesn't explicitly disable them. PRRTE is the key going forward as it is the one that does the discovery - and PRRTE uses the hwloc provided to it (PRRTE does not itself embed a copy of hwloc). Since PCI is enabled by default, we should be okay if we wind up using an external hwloc - though we definitely need to check the return code if we call PCI-related APIs.

bgoglin:
Actually hwloc disables PCI by default (but lstopo enables everything by default).

rhc54:
Really? Hmmm...might want to adjust the configure help output, then, because it only says this:
```text
  --disable-pci           Disable the PCI device discovery
```
which implies that it is enabled by default.

opal/mca/common/ofi/common_ofi.c
dancejic: I'll fix this, thanks!

bgoglin:
Oh thanks. It's enabled at configure time but disabled at execution time :/
So many minor things to tweak these days.

rhc54:
> So many minor things to tweak these days.

I know what you mean 😄  How do I indicate that we want PCI device info in the topology when we load it? I didn't find a flag that explicitly deals with PCI.

bgoglin:
Instead of topo flags, 2.x uses type-specific filtering:
```
hwloc_topology_set_type_filter(topo, HWLOC_OBJ_PCIDEV, HWLOC_TYPE_FILTER_KEEP_ALL);
```
or even
```
hwloc_topology_set_io_types_filter(topo, HWLOC_TYPE_FILTER_KEEP_IMPORTANT);
```
"Important" is somehow specific to bridges, it means keep all except bridges that have nothing behind them.

rhc54:
Ah, okay - we already set the io_types_filter to "keep important", so it sounds like we should be good. Thanks!

opal/mca/common/ofi/common_ofi.c
dancejic: My general goal was to distribute the processes across all available NIC as much as possible while choosing a local NIC when available. Is there a specific aspect that you wanted me to comment on?

dancejic:
Thanks for the feedback! I'll work on fixing `get_process_numanode()` for now. A few other questions that I have in the meantime:

- Should I go back to using the local rank instead of numa rank?

- Should I still try and search for NUMA nodes? And if not, what would the alternative be? My main goal was to use them to somewhat determine locality of the device to the process, but it sounds like this might not always work that easily.

bgoglin:
No need to go up to object for determining locality, otherwise you need to determine what kind of object, and that's not easy to decide. You may rely on cpuset which is just a locality without caring about what kind of object it corresponds too. Objects have a cpuset, process binding is a cpuset.

rhc54:
Wouldn't it make more sense just to use the distance values from the proc to the device, and then pick the minimum one? It's what we do in other BTLs.

bgoglin:
Yes, anything that doesn't assume a specific hierarchy (NUMA under package, etc) is better.

rhc54:
@dancejic You might want to look at the btl/usnic code as it assigns to the proc the NIC that is closest to it based on the hwloc distance matrix values.

bwbarrett:
I think Nikola's question about NUMA rank was what to do when there are two NICs with the same distance (which will happen in the cases of a single Intel Cascade Lake socket with two NICs).  We need some mechanism for picking one.  Given a single process, this is probably not a big deal.  Given (say) 24 processes, you'd like a mechanism where you generally end up with 12 processes picking NIC A and 12 processes picking NIC B.  We were using NUMARANK as a way of doing that...

rhc54:
What if you turn the problem on its head? If you have each NIC look at its distance to a given proc, you can collect the NICs that are closest - and then have the one with the fewest number of procs assigned to it take that proc. This will always load balance the NICs while ensuring that you are using the min-dist connection, and doesn't rely on any rank-based value.

bwbarrett:
but does require inter-process communication before the modex :/.

rhc54:
Hmmm...I guess that means each process would have to get some info from the NIC. Never mind - probably would take too much effort to create that integration. Still, you could turn that approach around and compute it from the proc's perspective using the local or node rank, so maybe that would work? In other words, collect the group of NICs that are closest and then just `my-local-rank % num-nics` to select the one to use?

Guess that is what you are already doing, just replacing noderank with localrank...so ignore my ruminations 😄

bwbarrett:
bot:aws:retest (testing the new CI images)

bwbarrett:
bot:aws:retest (testing the new CI images)

bwbarrett:
bot:aws:retest (testing the new CI images)

bwbarrett:
bot:aws:retest (testing the new CI images)

dancejic:
Changed to use cpuset to determine locality instead of NUMA nodes, I use hwloc_bitmap_intersects to determine if a pci device and process have related cpusets.

rhc54:
@bgoglin I'm trying to understand what this means (from within this PR):
```text
 *     Uses a pci device to find an ancestor that contains a cpuset, and
 *     determines if it intersects with the cpuset that the process is bound to.
 *     if the process is not bound, or if a cpuset is unavailable for whatever
 *     reason, returns false. Otherwise, returns the result of
```
Why would a PCI device have a cpuset? I'm trying to figure out what it means when a process has a cpuset that intersects with a PCI device - is that supposed to mean something about proximity?

bgoglin:
PCI device don't have a CPU set in hwloc. However it walk up parents until one has a cpuset, you find the locality of your PCI device in terms of local PUs.

rhc54:
Interesting. Let me see if I understand this correctly as I am getting ready to write the code that provides each process with a list of NICs on their node ordered in terms of proximity to the proc. It sounds like you are saying that I can take each NIC, climb the topology tree until I find a parent with a cpuset, and then test that cpuset against the procs on the node. Any proc with an intersecting cpuset is "local" to that NIC - procs without an intersection are "remote".

The higher I have to go to find the intersection, the further away the NIC is from me? Guess I'm searching for an understanding of how using "distance" compares to this climbing of the tree - are the two equivalent?

bgoglin:
The higher you have to go from PCI to object with cpuset, the deeper the PCI tree is. But even if you have to traverse multiple bridges, it's not clear they are shared, hence it's not clear performance will decrease. So I wouldn't use that "distance" to sort NICs.

What you could do instead is measure a distance from the parent of the PCI (the one with a cpuset) to the object that corresponds to the binding (if I am bound to a single core, that object is that core. if I am bound to two cores, that object is the container package). Distance would be the number of hops from the PCI parent up to the common ancestor + number of hops down to the binding object.

The distance I see in btl/usnic is quite different, it's the NUMA distance. NUMA distance==10 (NIC is local) if the cpuset intersection isn't empty. However using cpusets won't let you compare two non-local NICs (because both have empty cpuset intersections). In short, cpuset is good for finding local NICs, not for sorting non-local NICs.

Once thing to remember is that NUMA distance is good for comparing NICs whose locality is exactly equal to a single NUMA node (pretty much always the case these days) and for processes whose binding isn't larger than a NUMA node.

What annoys me is that machines where there are multiple non-local NICs are often machines with lots of NUMA nodes (more than 4), and NUMA distances are often dumb there. For instance I may be at the same distance from a NUMA node in the same socket and from a NUMA node in a different socket but connected to me through a direct link.

dancejic:
That's the approach that I used in this PR. I used `hwloc_get_non_io_ancestor_obj(topology, obj);` to find a parent, then use the cpuset of that parent and the process cpuset to determine if the device is local.

PR #7646, opened 2020-04-20
add a common ofi whitelist/blacklist

hppritcha:

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -6,6 +6,8 @@
  * Copyright (c) 2015-2016 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights reserved.
+ * Copyright (c) 2020      Triad National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -28,8 +30,6 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
                             bool enable_mpi_threads);
 
 static int param_priority;
-static char *prov_include;
-static char *prov_exclude;
 static int control_progress;
 static int data_progress;
 static int av_type;
@@ -130,24 +130,6 @@ ompi_mtl_ofi_component_register(void)
                                     OPAL_INFO_LVL_9,
                                     MCA_BASE_VAR_SCOPE_READONLY,
                                     &param_priority);
-
-    prov_include = NULL;
-    mca_base_component_var_register(&mca_mtl_ofi_component.super.mtl_version,
-                                    "provider_include",
-                                    "Comma-delimited list of OFI providers that are considered for use (e.g., \"psm,psm2\"; an empty value means that all providers will be considered). Mutually exclusive with mtl_ofi_provider_exclude.",
-                                    MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
-                                    OPAL_INFO_LVL_1,
-                                    MCA_BASE_VAR_SCOPE_READONLY,
-                                    &prov_include);
-
-    prov_exclude = "shm,sockets,tcp,udp,rstream";
-    mca_base_component_var_register(&mca_mtl_ofi_component.super.mtl_version,
-                                    "provider_exclude",
-                                    "Comma-delimited list of OFI providers that are not considered for use (default: \"sockets,mxm\"; empty value means that all providers will be considered). Mutually exclusive with mtl_ofi_provider_include.",
-                                    MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
-                                    OPAL_INFO_LVL_1,
-                                    MCA_BASE_VAR_SCOPE_READONLY,
-                                    &prov_exclude);
 
     ompi_mtl_ofi.ofi_progress_event_count = MTL_OFI_MAX_PROG_EVENT_COUNT;
     opal_asprintf(&desc, "Max number of events to read each call to OFI progress (default: %d events will be read per OFI progress call)", ompi_mtl_ofi.ofi_progress_event_count);
@@ -268,6 +250,8 @@ ompi_mtl_ofi_component_register(void)
                                     MCA_BASE_VAR_SCOPE_READONLY,
                                     &ompi_mtl_ofi.num_ofi_contexts);
 
+    opal_common_ofi_register_mca_variables(&mca_mtl_ofi_component.super.mtl_version);
+
     return OMPI_SUCCESS;
 }
 
@@ -312,6 +296,7 @@ ompi_mtl_ofi_component_query(mca_base_module_t **module, int *priority)
 static int
 ompi_mtl_ofi_component_close(void)
 {
+    opal_common_ofi_mca_deregister();
     return OMPI_SUCCESS;
 }
 
@@ -350,7 +335,7 @@ select_ofi_provider(struct fi_info *providers,
     if (NULL != include_list) {
         while ((NULL != prov) &&
                (!is_in_list(include_list, prov->fabric_attr->prov_name))) {
-            opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+            opal_output_verbose(1, opal_common_ofi.output,
                                 "%s:%d: mtl:ofi: \"%s\" not in include list\n",
                                 __FILE__, __LINE__,
                                 prov->fabric_attr->prov_name);
@@ -359,7 +344,7 @@ select_ofi_provider(struct fi_info *providers,
     } else if (NULL != exclude_list) {
         while ((NULL != prov) &&
                (is_in_list(exclude_list, prov->fabric_attr->prov_name))) {
-            opal_output_verbose(1, ompi_mtl_base_framework.framework_output,
+            opal_output_verbose(1, opal_common_ofi.output,
                                 "%s:%d: mtl:ofi: \"%s\" in exclude list\n",
                                 __FILE__, __LINE__,
                                 prov->fabric_attr->prov_name);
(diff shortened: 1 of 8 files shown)


ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -619,19 +604,21 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
     int universe_size;
     char *univ_size_str;
 
+
```
wckzhang, line 607: NIT: Rogue whitespace addition

ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -619,19 +604,21 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
...
-        include_list = opal_argv_split(prov_include, ',');
-    } else if (NULL != prov_exclude) {
-        exclude_list = opal_argv_split(prov_exclude, ',');
+    if (NULL != *opal_common_ofi.prov_include) {
+        include_list = opal_argv_split(*opal_common_ofi.prov_include, ',');
+    } else if (NULL != *opal_common_ofi.prov_exclude) {
+        exclude_list = opal_argv_split(*opal_common_ofi.prov_exclude, ',');
     }
 
+
```
wckzhang, line 621: NIT: Rogue whitespace addition

opal/mca/btl/ofi/Makefile.am
```diff
@@ -58,8 +60,10 @@ mca_btl_ofi_la_SOURCES = $(component_sources)
 mca_btl_ofi_la_LDFLAGS = -module -avoid-version \
                          $(opal_ofi_LDFLAGS)
 mca_btl_ofi_la_LIBADD = $(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la \
+    $(OPAL_TOP_BUILDDIR)/opal/mca/common/ofi/lib@OPAL_LIB_PREFIX@mca_common_ofi.la \
     $(opal_ofi_LIBS)
 
+
```
wckzhang, line 66: NIT: Rogue whitespace addition

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -268,8 +259,12 @@ static mca_btl_base_module_t **mca_btl_ofi_component_init (int *num_btl_modules,
...
     }
 
-    /* Select the provider */
-    fabric_attr.prov_name = prov_include;
+    fabric_attr.prov_name = NULL;
+    /* Select the provider - sort of.  we just take first element in list for now */
+    if (NULL != *opal_common_ofi.prov_include) {
+        include_list = opal_argv_split(*opal_common_ofi.prov_include, ',');
+        fabric_attr.prov_name = include_list[0];
+    }
```
wckzhang, line 267: This doesn't seem like the correct behavior...It ignores all providers except the first one and ignores the exclude list as well. Not saying it needs to be done in this patch, but can you make an issue so we can track this?

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -314,9 +309,16 @@ static mca_btl_base_module_t **mca_btl_ofi_component_init (int *num_btl_modules,
     rc = fi_getinfo(FI_VERSION(1, 5), NULL, NULL, 0, &hints, &info_list);
     if (0 != rc) {
         BTL_VERBOSE(("fi_getinfo failed with code %d: %s",rc, fi_strerror(-rc)));
+        if (NULL != include_list) {
+            opal_argv_free(include_list);
+        }
         return NULL;
     }
 
+    if (NULL != include_list) {
```
wckzhang, line 318: Can put this after line 209 if we always want to free include_list after fi_getinfo, then we don't have to duplicate this code.

opal/mca/btl/ofi/btl_ofi_component.c
hkuno: @wckzhang, @hppritcha  ignoring all providers except the first one seems to match the original btl/ofi code and the pre-existing README, which says that btl/ofi will initialize a module with ONLY the first compatible info returned from OFI.  I might be reading it wrong, but it seems that currently there is no btl/ofi exclude list? Should there be? (Or William, are you saying that the btl/ofi should look a the mtl/ofi exclude list?)

hppritcha: @hkuno I think the answer to your question is yes.  That's part of why I marked this as a WIP.  If we'd like to have the BTL work like the MTL in terms of include/exclude I'd be fine with adding that cap.  I'd do that as  a separate commit to this PR.

Anyone object to adding an ofi mtl like include/exclude behavior to the ofi btl?

wckzhang: I think that's a good idea Jeff

hkuno: Thank you for both the current commit and for volunteering to do the new one!

hppritcha: @jsquyres this PR already includes the aliasing.  See the common_ofi.c

hppritcha:
@wckzhang @hkuno @mwheinz i've update the PR with changes from William and also add a preliminary common provider selection code.

opal/mca/btl/ofi/btl_ofi_component.c
mwheinz: What happens if opal_common_ofi.prov_include is NULL? I assume that mean we can choose from any BTL?

opal/mca/common/ofi/common_ofi.c
```diff
@@ -16,12 +18,62 @@
...
+    static int exclude_index;
+
+    if (fi_version() < FI_VERSION(1,0)) {
         return OPAL_ERROR;
     }
+
+    if (!registered) {
+        /*
+         * this monkey business is needed because of the way the MCA VARs stuff tries to handle pointers to strings when
+         * when destructig the MCA var database.  If you don't do something like this,the MCA var framework will try
```
mwheinz, line 68: typo.

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -192,7 +192,8 @@ static int mca_btl_ofi_component_open(void)
  */
 static int mca_btl_ofi_component_close(void)
 {
-    /* If we don't sleep, sockets provider freaks out. */
+    opal_common_ofi_mca_deregister();
+    /* If we don't sleep, sockets provider freaks out. Ummm this is a scary comment */
```
mwheinz, line 196: This is a bit scary. Should that sleep be in the sockets provider instead?

opal/mca/common/ofi/common_ofi.c
```diff
@@ -73,7 +103,95 @@ int opal_common_ofi_register_mca_variables(const mca_base_component_t *component
...
+    }
+
+    if (NULL != *opal_common_ofi.prov_include) {
+        include_list = opal_argv_split(*opal_common_ofi.prov_include, ',');
+    } else if (NULL != *opal_common_ofi.prov_exclude) {
+        exclude_list = opal_argv_split(*opal_common_ofi.prov_exclude, ',');
+    }
+
+    if (NULL != include_list) {
+        while ((NULL != prov) &&
```
mwheinz, line 164: Maybe this is existing practice, but it seems that by traversing the providers and checking against the include list we're ignoring any preference the user may have intended. In other words, if I specified an include list of psm2,openib (as an example) and openib happened to appear in the provider list ahead of psm2 we would choose openib when psm2 would have been preferred. 

Again, this might be how it works now, but is that the desirable behavior?

opal/mca/btl/ofi/btl_ofi_component.c
hppritcha: yes that's the way the logic currently works for both MTL and BTL.

opal/mca/common/ofi/common_ofi.c
hppritcha: We may want to come up with something more complex later but let's stick with current method for this PR.  Anyway I'm thinking of not including the select provider logic change as it looks like there's something like that already in the works - see PR #7641

opal/mca/btl/ofi/Makefile.am
hppritcha: fixed in additional commit

opal/mca/common/ofi/common_ofi.c
wckzhang: I agree with hprritcha, if we want to do that later, we can. I don't think we specify any ordering priority for include lists, so that is a user assumption right now. We can change that, but I think it'd be more appropriate in a subsequent patch.

opal/mca/btl/ofi/btl_ofi_component.c
wckzhang: Sounds like a sockets bug, is there something filed in ofiwg libfabric? Is this comment outdated now - could have been fixed? I'm fairly certain sockets has been mostly replaced by the tcp provider.

hppritcha:
@mwheinz @hkuno @wckzhang 
I'm thinking to remove the common select provider method from this PR given the activity going on with PR #7641

opal/mca/btl/ofi/btl_ofi_component.c
hppritcha: this is going to change if PR #7641 common ofi provider selection method is pulled in.

opal/mca/btl/ofi/btl_ofi_component.c
hppritcha: yes that's my understanding as well, that sockets is/has been replaced by tcp

hppritcha:
@wckzhang @hkuno @mwheinz  took a git trickery but backed out the component select pieces from this PR.  This PR now only adds a common blacklist/whitelist as well as a common verbose parameter.  PR #7641 is where the action is for the provider selector logic.

Please double check.  Thanks

opal/mca/btl/ofi/btl_ofi_component.c
hppritcha: this will be replace with other stuff when PR #7641 comes in.

hppritcha:
i checked merging in PR #7641 with this and except for a trivial fix to common_ofi.c it merges cleanly.

hppritcha:
bot:ompi:retest

Issue #7471, opened 2020-02-25
Bring back "component alias" functionality


jsquyres:
Per the [Feb 2020 face-to-face meeting](https://github.com/open-mpi/ompi/wiki/Meeting-2020-02), we're going to bring the "component alias" functionality back (i.e., @hjelmn opened a PR about this forever ago, but it never got merged).

For v5.0, we'd like to:

1. Delete the existing "sm" BTL
2. Rename "vader" to be "sm"
3. Make "vader" be an alias to "sm"

This requires bringing back / finishing the "component alias" functionality.

hjelmn:
Still working on this. Estimate ~ 2 weeks development time given my work/home schedule this month.

jsquyres:
Ok.  Planned branch (i.e., feature completeness) for v5.0 is April 30.

awlauria:
Completed with #7671 merged to master.

PR #7671, opened 2020-04-29
Add support for component aliasing

hjelmn:
mca/base: add support for component aliasing
    
This commit adds support for aliasing component names. A component
name alias is created by calling: mca_base_alias_register. The name
of the project and framework are optional. The component name and
component alias are required. Once an alias is registered all
variables registered after the alias creation will have synonyms
also registered. For example:
    
```c
mca_base_alias_register("opal", "btl", "vader", "sm", false);
```
    
would cause all of the variables registered by btl/vader to have
aliases that start with btl_sm. Ex: btl_vader_single_copy_mechanism
would have the synonym: btl_sm_single_copy_mechanism.
    
If aliases are registered before component filtering the alias
can also be used for component selection. For example, if sm is
registered as an alias to vader in the btl framework register
function then ```--mca btl self,sm``` would be equivalent to
```--mca btl self,vader```.
    
### CODE REVIEW

**Comment by bwbarrett** [comment]
bot:aws:retest

**Comment by hjelmn** [comment]
Need to move some changes from one commit to another.

**Comment by hjelmn** [comment]
Fixed. Also updated the header with documentation.

File: opal/mca/btl/base/btl_base_frame.c [hunk 1]
```diff
@@ -154,6 +156,8 @@ static int mca_btl_base_register(mca_base_register_flag_t flags)
     (void) mca_base_var_enum_register("opal", "btl", "base", "btl_atomic_flags",
                                  &mca_btl_base_atomic_enum);
 
+    (void) mca_base_alias_register ("opal", "btl", "sm", "vader", MCA_BASE_ALIAS_FLAG_NONE);
```
jsquyres, line 159: This seems like an odd place to put this (i.e., in the base, not in the `sm` component itself).

Is it not sufficient to put this at the very beginning of the `sm` component?

hjelmn: That is too late for selection. We don't always dlopen the components before selection so the only place to register the synonym is in the base. My other approach had that issue. Had to make vader static for it to work.

hjelmn:
They are synonyms....

jsquyres:
> They are synonyms....

You're right: I totally missed that in the `ompi_info` output.

opal/mca/btl/base/btl_base_frame.c
jsquyres: Ok.  I added a comment here in the code explaining this.

jsquyres:
@hjelmn I squashed my commits into your "rename vader" commit and pushed here.

awlauria:
Fixes https://github.com/open-mpi/ompi/issues/7471

PR #7677, opened 2020-04-30
openmpi.spec: Add override install for openmpi-mca-params.conf

wckzhang:


wckzhang:
There's probably a better way to do this, don't have much experience writing spec files

contrib/dist/linux/openmpi.spec
```diff
@@ -66,6 +66,10 @@
 %{!?shell_scripts_path: %define shell_scripts_path %{_bindir}}
 # type: string (base name of the shell scripts)
 %{!?shell_scripts_basename: %define shell_scripts_basename mpivars}
+# type: bool (0/1)
+%{!?install_params_file: %define install_params_file 0}
+# type: string (full path of params file)
+%{!?params_file: %define params_file openmpi-mca-params.conf}
```
jsquyres, line 72: Where does this default file come from?

Or is it an error to specify `install_params_file==1` without also specifying a valid `params_file`?

wckzhang: Yeah it's an error

jsquyres: You should make it fail early, then.  Otherwise, it doesn't fail until deep during `make install`, and the error isn't obvious -- it'll say `openmpi-mca-params.conf not found`, or something.  And that will be uber confusing, because people will think that that file _does_ exist.

wckzhang: I was thinking instead to get rid of "install_params_file" and just have "params_file" and a check to see if it exists in the make install <-- just haven't had the time to test if I can do this

jsquyres: @wckzhang Ping.  Any progress, perchance?

wckzhang: No sorry, I will put this on my TODO list

wckzhang: I was having a few issues when testing so I put it off for later and never got back to it

lanl-ompi:

contrib/dist/linux/openmpi.spec
jsquyres: @wckzhang Do you think you'll finish this PR?

wckzhang: I don't see it in the short term. If you want to close it, I can always re-open if I see the need in the future.

jsquyres: You can close it if you like, or perhaps put it in draft mode.

wckzhang:
Crap, pushed to master branch caused this PR to be updated

bwbarrett:
AWS isn't really interested in this PR at this point, so closing.

PR #7698, opened 2020-05-05
Add checks for libevent.so conflict with LSF

jjhursey:
* LSF ships a `libevent.so` that is no related to the `libevent.so`
   shipped with Libevent.
 * Add some checks to the configure logic to detect scenarios where this
   conflict can be detected, and provide the user with a descriptive
   warning message.
   - When detected by `event/external` this is just a warning since
     the internal component may be able to be used instead.
     - This happens when the user supplies the LSF path via the
       `LDFLAGS` envar instead of via `--with-lsf-libdir`.
   - When detected by a LSF component and LSF was explicitly requested
     then this becomes an error. Otherwise, it will just print the warning
     and that component will fail to build.

### DIFF
--- a/config/orte_check_lsf.m4
+++ b/config/orte_check_lsf.m4
@@ -144,21 +144,25 @@ AC_DEFUN([ORTE_CHECK_LSF],[
           # If the external libevent component used 'event_core' instead of 'event'
           orte_check_lsf_event_conflict=na
           # Split libs into an array, see if -levent is in that list
-          orte_check_lsf_libevent_present=`echo $LIBS | awk '{split([$]0, a, " "); {for (k in a) {if (a[[k]] == "-levent") {print a[[k]]}}}}' | wc -l`
+          orte_check_lsf_libevent_present=`echo "$LIBS" | awk '{split([$]0, a, " "); {for (k in a) {if (a[[k]] == "-levent") {print a[[k]]}}}}' | wc -l | tr -d '[[:space:]]'`
+          # (1) LSF check must have failed above. We need to know why...
           AS_IF([test "$orte_check_lsf_happy" = "no"],
-                [AS_IF([test "$opal_event_external_support" = "yes" && test "$orte_check_lsf_libevent_present" != 0],
+                [# (2) If there is a -levent in the $LIBS then that might be the problem
+                 AS_IF([test "$opal_event_external_support" = "yes" && test "$orte_check_lsf_libevent_present" != "0"],
                        [AS_IF([test "$orte_check_lsf_libdir" = "" ],
                               [],
                               [LDFLAGS="$LDFLAGS -L$orte_check_lsf_libdir"])
                         # Note that we do not want to set LIBS here to include -llsf since
                         # the check is not for an LSF library, but for the conflict with
                         # LDFLAGS.
+                        # (3) Check to see if the -levent is from Libevent (check for a symbol it has)
                         AC_CHECK_LIB([event], [evthread_set_condition_callbacks],
                                      [AC_MSG_CHECKING([for libevent conflict])
                                       AC_MSG_RESULT([No. The correct libevent.so was linked.])
                                       orte_check_lsf_event_conflict=no],
-                                     [AC_MSG_CHECKING([for libevent conflict])
-                                      AC_MSG_RESULT([Yes. A wrong libevent.so was linked.])
+                                     [# (4) The libevent.so is not from Libevent. Warn the user.
+                                      AC_MSG_CHECKING([for libevent conflict])
+                                      AC_MSG_RESULT([Yes. Detected a libevent.so that is not from Libevent.])
                                       orte_check_lsf_event_conflict=yes])
                        ],
                        [AC_MSG_CHECKING([for libevent conflict])
@@ -175,6 +179,8 @@ AC_DEFUN([ORTE_CHECK_LSF],[
                  AC_MSG_WARN([A system-installed Libevent library was detected and the Open MPI])
                  AC_MSG_WARN([build system chose to use the 'external' component expecting to])
                  AC_MSG_WARN([link against the Libevent in the linker search path.])
+                 AC_MSG_WARN([If LSF is present on the system and in the default search path then])
+                 AC_MSG_WARN([it _may be_ the source of the conflict.])
                  AC_MSG_WARN([LSF provides a libevent.so that is not from Libevent in its])
                  AC_MSG_WARN([library path. At this point the linker is attempting to resolve])
                  AC_MSG_WARN([Libevent symbols using the LSF library because of the lack of])
--- a/config/orte_check_lsf.m4
+++ b/config/orte_check_lsf.m4
@@ -15,7 +15,7 @@ dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
 dnl Copyright (c) 2016      Los Alamos National Security, LLC. All rights
 dnl                         reserved.
-dnl Copyright (c) 2017      IBM Corporation.  All rights reserved.
+dnl Copyright (c) 2017-2020 IBM Corporation.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


jjhursey:
Note: This is filed against v4.0.x first because the configury change spans ORTE and OMPI. We will be able to pick a part of this patch back to OMPI master, and the other part to PRRTE master (thus two commits).

For the first round of review, it was easier to keep them together in a single commit.

jjhursey:
I'm keeping this as a draft for the moment. I'm looking at a libevent library linking advancement to tack onto this ticket.

opal/mca/event/external/configure.m4
```diff
@@ -115,11 +116,41 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
...
+                 [AC_CHECK_LIB([event], [event_fini],
+                               [AC_MSG_WARN([===================================================================])
+                                AC_MSG_WARN([Possible conflicting libevent.so libraries detected on the system.])
+                                AC_MSG_WARN([])
+                                AC_MSG_WARN([LSF provides a libevent.so that is not from Libevent in its])
+                                AC_MSG_WARN([library path. It is possible that you have installed Libevent])
+                                AC_MSG_WARN([on the system, but the linker is picking up the wrong version.])
+                                AC_MSG_WARN([])
+                                AC_MSG_WARN([Configure will continue and attempt to use the 'internal' libevent])
+                                AC_MSG_WARN([instead of the 'external' libevent.])
```
jsquyres, line 133: This part of the message isn't true if the user specified `--with-event=external`

opal/mca/event/external/configure.m4
```diff
@@ -115,11 +116,41 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
...
+                                AC_MSG_WARN([===================================================================])
+                                opal_event_external_support=no
+                               ],
+                               [])
+                 ])
+
+           AS_IF([test "$opal_event_external_support" = "yes"],
+                 [LDFLAGS="$opal_event_external_LDFLAGS $LDFLAGS"
+                  CPPFLAGS="$opal_event_external_CPPFLAGS $CPPFLAGS"],
+                 [])
```
jsquyres, line 148: You don't need to put the empty `[]` here.

opal/mca/event/external/configure.m4
```diff
@@ -116,6 +116,32 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
...
+                                AC_MSG_WARN([instead of the 'external' libevent.])
+                                AC_MSG_WARN([])
+                                AC_MSG_WARN([If your intention was to use the 'external' libevent then you need])
+                                AC_MSG_WARN([to address this linker path ordering issue. One way to do so is])
+                                AC_MSG_WARN([to make sure the libevent system library path occurs before the])
+                                AC_MSG_WARN([LSF library path.])
+                                AC_MSG_WARN([===================================================================])
+                                opal_event_external_support=no
+                               ],
+                               [])
```
jsquyres, line 142: You don't need the empty `[]` here.

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+          #
+          # So look for a function that libevent_pthreads is looking for from libevent.so
+          # If it does not appear here when adding the same option then we have
+          # a conflict.
+          orte_check_lsf_event_conflict=na
+          AS_IF([test "$orte_check_lsf_happy" = "no"],
+                [AS_IF([test "$opal_event_external_support" = "yes"],
+                       [AS_IF([test "$orte_check_lsf_libdir" = "" ],
+                              [],
+                              [LDFLAGS="$LDFLAGS -L"$orte_check_lsf_libdir])
```
jsquyres, line 149: That's a weird placement for the closing quote -- is there a reason to have it there?

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                 AC_MSG_RESULT([No. LSF checks passed.])
+                 orte_check_lsf_event_conflict=na])
+
+          AS_IF([test "$orte_check_lsf_event_conflict" = "yes"],
+                [AC_MSG_WARN([===================================================================])
+                 AC_MSG_WARN([Conflicting libevent.so libraries detected on the system.])
+                 AC_MSG_WARN([])
+                 AC_MSG_WARN([A system installed Libevent library was detected and the Open MPI])
+                 AC_MSG_WARN([build system chose to use the 'external' component expecting to])
+                 AC_MSG_WARN([link against the Libevent in the default search path.])
```
jsquyres, line 174: It may not be the default search path -- you might just want to say "in the linker search path."

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                        orte_check_lsf_event_conflict=na])],
+                [AC_MSG_CHECKING([for libevent conflict])
+                 AC_MSG_RESULT([No. LSF checks passed.])
+                 orte_check_lsf_event_conflict=na])
+
+          AS_IF([test "$orte_check_lsf_event_conflict" = "yes"],
+                [AC_MSG_WARN([===================================================================])
+                 AC_MSG_WARN([Conflicting libevent.so libraries detected on the system.])
+                 AC_MSG_WARN([])
+                 AC_MSG_WARN([A system installed Libevent library was detected and the Open MPI])
```
jsquyres, line 172:
```suggestion
                 AC_MSG_WARN([A system-installed Libevent library was detected and the Open MPI])
```

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                 AC_MSG_WARN([build system chose to use the 'external' component expecting to])
+                 AC_MSG_WARN([link against the Libevent in the default search path.])
+                 AC_MSG_WARN([LSF provides a libevent.so that is not from Libevent in its])
+                 AC_MSG_WARN([library path. At this point the linker is attempting to resolve])
+                 AC_MSG_WARN([Libevent symbols using the LSF library because of the lack of])
+                 AC_MSG_WARN([an explicit linker path pointing to the system installed Libevent.])
```
jsquyres, line 178:
```suggestion
                 AC_MSG_WARN([an explicit linker path pointing to the system-installed Libevent.])
```

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                 AC_MSG_WARN([])
+                 AC_MSG_WARN([To resolve this issue either (1) explicitly pass the Libevent])
```
jsquyres, line 180:
```suggestion
                 AC_MSG_WARN([To resolve this issue either (A) explicitly pass the Libevent])
```

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,71 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                 AC_MSG_WARN([library path on the configure line (--with-libevent-libdir), or])
+                 AC_MSG_WARN([(B) use the internal libevent by requesting it from configure ])
+                 AC_MSG_WARN([with the --with-libevent=internal option.])
+                 AC_MSG_WARN([===================================================================])
+                ],
+                [])
```
jsquyres, line 186: You don't need the empty `[]` here.

config/orte_check_lsf.m4
jjhursey: nope. I can fix.

opal/mca/event/external/configure.m4
jjhursey: Good point. I'll generalize the text. If they did specify `--with-libevent=external` then this message will print, but it will error out at the bottom of the configure check.

jjhursey: Pushed change. Take a look at the new wording.

jjhursey:
I just re-pushed with the changes that Jeff suggested.

#### libevent vs libevent_core

I added 1 new commit that moves from using `-levent` to `-levent_core` for the core libevent functionality. I found some notes [here](http://www.wangafu.net/~nickm/libevent-book/Ref0_meta.html) that discuss the legacy nature of `-levent`. `-levent_core` has the functionality that we need. `-levent` has all of core plus a few things we don't need.

`libevent_core.so` exists as far back as Libevent 2.0.5 which is the oldest supported Libevent that Open MPI supports.

I kept that particular change out as a separate commit to isolate it. We may need to change the internal component, but I wanted to hold off on that given the changes to libevent happening in @bwbarrett 's branch. But I'd like Brian to comment on the libevent_core part of this change before deciding if we should include it or not. We might decide not to do it for v4, but to move to it for v5.

opal/mca/event/external/configure.m4
jsquyres: I'll look at the new wording.  You could also just selectively emit this paragraph.

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,70 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+                              [],
+                              [LDFLAGS="$LDFLAGS -L$orte_check_lsf_libdir"])
+                        # Note that we do not want to set LIBS here to include -llsf since
+                        # the check is not for an LSF library, but for the conflict with
+                        # LDFLAGS.
+                        AC_CHECK_LIB([event], [evthread_set_condition_callbacks],
+                                     [AC_MSG_CHECKING([for libevent conflict])
+                                      AC_MSG_RESULT([No. The correct libevent.so was linked.])
+                                      orte_check_lsf_event_conflict=no],
+                                     [AC_MSG_CHECKING([for libevent conflict])
```
jjhursey, line 157: (Note from code review with Geoff)
We use `event` here instead of `event_core` because we are looking to see if the `-levent` that was picked up was from LSF (false condition) or Libevent (true condition) to detect if there is a conflict.

If we use `-levent_core` instead of `-levent` then this condition will fire if we built with an external libevent on a system without LSF installed. Thus triggering a warning that is inaccurate. So if we decide to go with `-levent_core` we should probably remove this check since it will be for an always false case.

Note that if we use `-levent` instead on a system with an external libevent without LSF installed then we get the 'true' condition and no warning is issued.

config/orte_check_lsf.m4
```diff
@@ -137,7 +201,7 @@ AC_DEFUN([ORTE_CHECK_LSF],[
            $1_WRAPPER_EXTRA_LDFLAGS=[$]$1_LDFLAGS
            $1_WRAPPER_EXTRA_LIBS=[$]$1_LIBS
            $2],
-          [AS_IF([test ! -z "$with_lsf" && test "$with_lsf" != "no"],
+          [AS_IF([test ! -z "$with_lsf" && test "$with_lsf" != "no" && test "$orte_check_lsf_event_conflict" = "yes"],
```
jjhursey, line 204: (Note from code review with Geoff)
This check is not doing what I originally thought and should be removed.

The scenario that I was considering is if LSF and Libevent are the default search path (no specified with --with-lsf) then if we detect a conflict then we error out after printing the warning. When actually it is probably ok to just print the warning and not error out here since the user did not explicitly say `--with-lsf`. They can notice it was not compiled in the final list of runtime components reported, then scroll up to see the warning message indicating the possible issue.

config/orte_check_lsf.m4
jjhursey: Actually I take that back. I got the logic backwards for the `libevent_core` scenario.
 - On a system with no LSF and an external libevent, the 'true' path is followed and no warning is reported. That's because it'll link in the libevent.so from Libevent.
 - On a system with LSF in the default path and an external libevent, the check would only fire if LSF couldn't compile. In that case it could run the AC_CHECK_LIB and the 'false' path is followed (since the only libevent.so it could pick up is from LSF).

In the second case the warning message is misleading. I think I can protect this a bit better. I'll tinker with it. At the very least I need to add a clarifying comment.

gpaulsen:
It'd be nice if the non-orte commits were cherry-picked from an equivalent master PR.

config/orte_check_lsf.m4
jjhursey: I just pushed a change to add protection for the `libevent_core` case. The check now only runs if it detects a `-levent` in the `$LIBS` - thus exposing us to picking up the wrong library.

config/orte_check_lsf.m4
jjhursey: done

jjhursey:
@gpaulsen I wanted to get eyes on it all together before breaking it apart between OMPI and PRRTE for `master`. I'll prepare the `master` versions next week. For now, I think this is ready for review.

Note that the last commit might be too much for v4 (changing from `-levent` to `-levent_core`). So the v4 RMs can make a call on that, and we can drop that last commit. The prior commits will work independently of it.

jjhursey:
bot:ibm:pgi:retest

config/orte_check_lsf.m4
```diff
@@ -120,6 +120,73 @@ AC_DEFUN([ORTE_CHECK_LSF],[
...
+          #   lib64/libevent_pthreads.so: undefined reference to `evthread_set_lock_callbacks'
+          # Because it picked up -levent from LSF, but -levent_pthreads from Libevent.
+          #
+          # So look for a function that libevent_pthreads is looking for from libevent.so.
+          # If it does appears then we have the correct libevent.so, otherwise then we picked
+          # up the LSF version and a conflict has been detected.
+          # If the external libevent component used 'event_core' instead of 'event'
+          orte_check_lsf_event_conflict=na
+          # Split libs into an array, see if -levent is in that list
+          orte_check_lsf_libevent_present=`echo $LIBS | awk '{split([$]0, a, " "); {for (k in a) {if (a[k] == "-levent") {print a[k]}}}}' | wc -l`
```
jjhursey, line 147: In further testing, this awk is not being processed quite right. I'm investigating.

jjhursey: Fixed. Needed bracketing protection.

jjhursey:
I'm not sure why the "Pull Request Build Checker" is stuck. It looks like the sub-jobs passed, but the main job status is still flashing. It's been stuck like that for `1 day 19 hr ago` so far :(

opal/mca/event/external/configure.m4
jsquyres: Nit: you still have the empty `[]` there.  It's not a deal-breaker, but I think you meant to take it out and missed it.  😇

jjhursey: It was removed. you need to refresh the browser.

gpaulsen:
We discussed on today's call, and we cam to consensus that this change _WOULD_ be appropriate for a v4.0.x release (changing dependency from `libevent` to `libevent_core` is okay, as it doesn't break any compatibility guarantees).

rhc54:
There is a problem with this patch - it causes spurious warnings to be emitted if there is a system-level libevent (a good one) and you point at a custom one. Please don't merge this until that gets fixed!

jjhursey:
The issue that @rhc54 mentioned is described here: https://github.com/openpmix/prrte/issues/560
I'll take a look.

jjhursey:
I just pushed to commits that should address Ralph's concerns.

Related PRs:
 * Open MPI `master` (cherry-picked here) PR #7749
 * PRRTE `master` (Ported version of the second commit in this PR) https://github.com/openpmix/prrte/pull/571

gpaulsen:
@bwbarrett This looks like the (transient?) gcc6 issue: https://jenkins.open-mpi.org/jenkins/job/open-mpi.build.compilers/4220/Compiler=gcc6/console

gpaulsen:
bot:retest

PR #7707, opened 2020-05-06
Updating README to include WARNING about ABI break

gpaulsen:
This commit adds a blurb to the README for v4.0.4, suggesting
that users of libmpi_usempif08.so skip over v4.0.3.

### DIFF
--- a/README
+++ b/README
@@ -8,11 +8,11 @@ Copyright (c) 2004-2008 High Performance Computing Center Stuttgart,
                         University of Stuttgart.  All rights reserved.
 Copyright (c) 2004-2007 The Regents of the University of California.
                         All rights reserved.
-Copyright (c) 2006-2018 Cisco Systems, Inc.  All rights reserved.
+Copyright (c) 2006-2020 Cisco Systems, Inc.  All rights reserved.
 Copyright (c) 2006-2011 Mellanox Technologies. All rights reserved.
 Copyright (c) 2006-2012 Oracle and/or its affiliates.  All rights reserved.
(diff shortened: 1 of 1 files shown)


gpaulsen:
Any wording suggestions are appreciated.

README
```diff
@@ -66,11 +66,16 @@ Much, much more information is also available in the Open MPI FAQ:
...
+base as of this writing (May 2020):
 
 General notes
 -------------
 
+- WARNING - Open MPI v4.0.3 accidentally dropped some external constant
+  definitions from the libmpi_usempif08.so library.  This is remedied
+  in v4.0.4.  It is recommended that users of libmpi_usempif08.so skip
+  over v4.0.3, and upgrade to v4.0.4 or later.
+
```
jsquyres, line 78: Per https://github.com/open-mpi/ompi/pull/7696#issuecomment-624857209, should probably also add something here at the new F08 symbols that were (effectively) added in 4.0.4, and what it means to the backwards compatibility guarantee for those symbols.

gpaulsen: I didn't list the constants out as the list appears to be ALL of the MPI constants, and is quite long.

jsquyres: No, that's not what I meant.  I mean that you should mention that if an app uses these symbols and compiles/links against 4.0.4 (i.e., the symbols that were erroneously left out of 4.0.0-4.0.3), they will not be able to link against the Fortran libraries from 4.0.0-4.0.3.

jsquyres: @gpaulsen and I chatted in Slack and came up with better wording.  He will push shortly.

PR #7708, opened 2020-05-06
PSM2 update to use PRRTE instead of ORTE

### LINKED ISSUE #7700: PSM2 is not compiling with OMPI master
master can't compile the PSM2 transport due to a missing #define. This is apparently related to the replacement of ORTE with PRRTE.

At the moment, it's only a single #define, but I'm still looking into it.

mwheinz:
Replace obsolete references to OMPI_NODE_RANK_INVALID and orte_precondition_transports.

Fixes #7700 

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2.c
@@ -101,7 +101,7 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     char env_string[256];
     int rc;
 
-    generated_key = getenv(OPAL_MCA_PREFIX"orte_precondition_transports");
+    generated_key = getenv("PMIX_MCA_opa_precondition_transports");
     memset(uu, 0, sizeof(psm2_uuid_t));
 
     if (!generated_key || (strlen(generated_key) != 33) ||
--- a/ompi/mca/mtl/psm2/mtl_psm2.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2.c
@@ -101,7 +101,7 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     char env_string[256];
     int rc;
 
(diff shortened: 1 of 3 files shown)


ompi/mca/mtl/psm2/mtl_psm2.c
```diff
@@ -101,7 +101,7 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
     char env_string[256];
     int rc;
 
-    generated_key = getenv(OPAL_MCA_PREFIX"orte_precondition_transports");
+    generated_key = getenv("PMIX_MCA_opa_precondition_transports");
```
rhc54, line 104: I don't know why this is being prefixed with `PMIX_MCA_` as it really isn't a parameter that controls PMIx. Would it make more sense to prefix/name it something else, perhaps more OPA-specific?

mwheinz: > I don't know why this is being prefixed with `PMIX_MCA_` as it really isn't a parameter that controls PMIx. 

The prefix is being attached by ``pmix_mca_base_var_env_name()``. I don't mind if we change it, but the prefix seems to be used entirely because it's PMIx that's generating the value.

rhc54: I understand _how_ the name gets that way - just not sure if we _want_ it to be that way as it isn't an MCA param that PMIx will recognize. Might confuse people who see it in their environment and think that PMIx should do something with it, or try to manually set it in an attempt to influence PMIx behavior.

Maybe just use `OPA_TRANSPORT_KEY`? Something that makes it clear it is solely intended for OPA's use?

mwheinz: That works for me. OPA is the only remaining user of the "preconditioned transports" key, so it seems appropriate.

Issue #7699, opened 2020-05-06
PSM2: Call add_procs through PML


mwheinz:
Shamelessly cribbed from Brian Barrett's issue #7571 - Update PSM2 MTL to use the correct add_procs call. 

> Change ompi_mtl_ofi_get_endpoint() to call the active PML's add_procs() rather than the OFI MTL add_procs() directly when discovering a new process during operation.
> Functionally, this has no impact in correct operation. However, the current behavior means that the heterogenous and active PML checks are not being executed in the dynamic discovery case.

mwheinz:
Catching up on my backlog.

PR #7713, opened 2020-05-07
PSM2: Call add_procs through PML

mwheinz:
Fixes issue #7699 .
Change ompi_mtl_ofi_get_endpoint() to call the active PML's add_procs()
rather than the OFI MTL add_procs() directly when discovering a new
process during operation.

Functionally, this has no impact in correct operation. However, the
current behavior means that the heterogenous and active PML checks
are not being executed in the dynamic discovery case.

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
+++ b/ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
@@ -10,7 +10,7 @@
  * Copyright (c) 2004-2006 The Regents of the University of California.
  *                         All rights reserved.
  * Copyright (c) 2006      QLogic Corporation. All rights reserved.
- * Copyright (c) 2015      Intel, Inc. All rights reserved
+ * Copyright (c) 2015-2020 Intel, Inc. All rights reserved
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -58,7 +58,10 @@ OBJ_CLASS_DECLARATION(mca_mtl_psm2_endpoint);
 static inline mca_mtl_psm2_endpoint_t *ompi_mtl_psm2_get_endpoint (struct mca_mtl_base_module_t* mtl, ompi_proc_t *ompi_proc)
 {
     if (OPAL_UNLIKELY(NULL == ompi_proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_MTL])) {
-	ompi_mtl_psm2_add_procs (mtl, 1, &ompi_proc);
+        int rc;
+        if (OPAL_UNLIKELY(OMPI_SUCCESS != (rc = MCA_PML_CALL(add_procs(&ompi_proc, 1))))) {
+            ompi_rte_abort(rc,"Error in ompi_mtl_psm2_get_endpoint.\n");
+        }
     }
 
     return ompi_proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_MTL];


ompi/mca/mtl/psm2/mtl_psm2.h
```diff
@@ -14,6 +14,8 @@
  * Copyright (c) 2015      Intel, Inc. All rights reserved
  * Copyright (c) 2015-2017 Los Alamos National Security, LLC. All rights
  *                         reserved.
+ * Copyright (c) 2020      Intel, Inc. All rights reserved.
```
jsquyres, line 17: Did you mean to put this copyright change on a file you didn't otherwise edit?

ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
```diff
@@ -58,7 +58,13 @@ OBJ_CLASS_DECLARATION(mca_mtl_psm2_endpoint);
 static inline mca_mtl_psm2_endpoint_t *ompi_mtl_psm2_get_endpoint (struct mca_mtl_base_module_t* mtl, ompi_proc_t *ompi_proc)
 {
     if (OPAL_UNLIKELY(NULL == ompi_proc->proc_endpoints[OMPI_PROC_ENDPOINT_TAG_MTL])) {
-	ompi_mtl_psm2_add_procs (mtl, 1, &ompi_proc);
+        if (OPAL_UNLIKELY(OMPI_SUCCESS != MCA_PML_CALL(add_procs(&ompi_proc, 1)))) {
+            /* Fatal error. exit() out */
+            opal_output(0, "%s:%d: *** The Open MPI PSM2 MTL is aborting the MPI job (via exit(3)).\n",
+                           __FILE__, __LINE__);
+            fflush(stderr);
```
jsquyres, line 65: FWIW, `opal_output()` uses an out-of-band mechanism to send the output back to `mpirun` (etc.).  Calling `fflush()` here may not have any effect.

Additionally, if you want to print a detailed error message, `opal_show_help()` is usually a better choice.

rhc54: Yeah, `fflush()` won't do anything as the output didn't go to `stderr`.

ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
```diff
@@ -58,7 +58,13 @@ OBJ_CLASS_DECLARATION(mca_mtl_psm2_endpoint);
...
+            exit(1);
```
jsquyres, line 66: Forgot to mention: I don't think you're supposed to call `exit()` directly; aren't you supposed to call `module->pml_error_callback` (assuming `module` and `module->pml_error_callback` are non-NULL)?

ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
mwheinz: So, I'm open to changing it, but I'd like to point out this is exactly the same code fragment that was approved for #7571 .

ompi/mca/mtl/psm2/mtl_psm2.h
mwheinz: Yeah, there was a change, then I realized it wasn't needed.

mwheinz: Done.

ompi/mca/mtl/psm2/mtl_psm2_endpoint.h
mwheinz: Same as my previous comment, this matches the code in #7571 .

rhc54: I don't see anything like that in the referenced PR - was it previously in that code area?

mwheinz: Look at the commit for #7571, or lines 45-47 of mtl_ofi_endpoint.h. This is literally a cut-and-paste of Brian's code.

mwheinz: (well, brian's code plus existing code in that function.)

rhc54: That is what I was asking, Mike - when you say this is something previously reviewed and approved, we only looked at the code that was actually being changed. Nobody looked at the rest of it or we would have flagged it there. Libraries are _not_ allowed to call `exit` under any circumstances. The `fflush` call is likely some old cruft.

mwheinz: Well, like I said, I'm willing to change it.

So, the previous code always just assumed add_procs() couldn't fail. Looking at other MTLs, OFI calls exit() (as we've discussed) and portals4 doesn't call add_procs(). Looking at the OFI and PSM2 MTLs, both of them will crash if add_procs() fails. 

I searched the code for references to pml_error_callback and the only place it's used is in usenic.  There are, however, a number of places in portals4 and osc/pt2pt that call abort() directly and dump core.

I mean, I could call ompi_mpi_abort()?

mwheinz:
Okay, for now I've mostly restored the old behavior, which was to ignore a failure in add_procs(). I've kept the error message, however.

mwheinz:
@rhc54 , @jsquyres , I believe this version fits what we discussed at the weekly meeting. It should be ready for review and merge.

PR #7714, opened 2020-05-08
Fix potential out-of-bounds write in opal_progress_unregister

devreal:
Fix a potential OOB write that may occur if the number of registered callbacks equals the size of the array holding the callbacks. The chances of this occurring in the wild seem fairly slim and probably only affects finalization.

The `callback_array_len` holds the number of active callbacks in the array and thus if `len == size` points one past the last element.

### DIFF
--- a/opal/runtime/opal_progress.c
+++ b/opal/runtime/opal_progress.c
@@ -474,8 +474,8 @@ static int _opal_progress_unregister (opal_progress_callback_t cb, volatile opal
         (void) opal_atomic_swap_ptr ((opal_atomic_intptr_t *) (callback_array + i), (intptr_t) callback_array[i+1]);
     }
 
-    callback_array[*callback_array_len] = fake_cb;
     --*callback_array_len;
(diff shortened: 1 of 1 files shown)


opal/runtime/opal_progress.c
```diff
@@ -474,7 +474,7 @@ static int _opal_progress_unregister (opal_progress_callback_t cb, volatile opal
         (void) opal_atomic_swap_ptr ((opal_atomic_intptr_t *) (callback_array + i), (intptr_t) callback_array[i+1]);
     }
 
-    callback_array[*callback_array_len] = fake_cb;
+    callback_array[*callback_array_len - 1] = fake_cb;
```
bosilca, line 477: I would just swap the 2 lines, to avoid replicating the decrement operation.

devreal: That was my first solution but I went for this in the interest of readability. Changed it back to just swapping the two lines.

Issue #7721, opened 2020-05-11
Add check for PSM2 reference counting to PSM2 MTL


mwheinz:
As discussed, a feature is being added to libpsm2 to correctly handle the case where the library is opened by multiple OMPI transports in the same process. (For example, the OFI BTL and the PSM2 MTL).

hppritcha:
@mwheinz can this issue be closed?

mwheinz:
Oops. Yeah, sorry.

PR #7722, opened 2020-05-11
Add check for PSM2 reference counting to PSM2 MTL #7721

mwheinz:
As discussed, a feature is being added to libpsm2 to correctly handle
the case where the library is opened by multiple OMPI transports in the same
process. (For example, the OFI BTL and the PSM2 MTL).

### DIFF
--- a/ompi/mca/mtl/psm2/mtl_psm2_component.c
+++ b/ompi/mca/mtl/psm2/mtl_psm2_component.c
@@ -323,6 +323,18 @@ ompi_mtl_psm2_component_open(void)
 static int
 ompi_mtl_psm2_component_query(mca_base_module_t **module, int *priority)
 {
+
+    /*
+     * Mixing the PSM2 MTL with the OFI BTL (using PSM2) 
+     * can cause an issue when they both call psm2_finalize
+     * in older versions of libpsm2.
+     */
+    if (!psm2_get_capability_mask(PSM2_LIB_REFCOUNT_CAP)) {
+        opal_output_verbose(2, ompi_mtl_base_framework.framework_output, 
+            "This version of the PSM2 MTL needs version 11.2.173 or later of the libpsm2 library for correct operation.\n");
+        return OMPI_ERR_FATAL;
+    }   
+
     /*
      * if we get here it means that PSM2 is available so give high priority
      */
--- a/config/ompi_check_psm2.m4
+++ b/config/ompi_check_psm2.m4
@@ -76,7 +76,15 @@ AC_DEFUN([OMPI_CHECK_PSM2],[
                [AC_MSG_WARN([glob.h not found.  Can not build component.])
                ompi_check_psm2_happy="no"])])
 
-	OPAL_SUMMARY_ADD([[Transports]],[[Intel Omnipath (PSM2)]],[$1],[$ompi_check_psm2_happy])
+        AS_IF([test "$ompi_check_psm2_happy" = "yes"],
+              [AC_CHECK_DECL([PSM2_LIB_REFCOUNT_CAP],
+                        [],
+                        [AC_MSG_WARN([PSM2 needs to be version 11.2.173 or later. Disabling MTL.])
+                         ompi_check_psm2_happy="no"],
+                        [#include <psm2.h>])
+              ])
+
+        OPAL_SUMMARY_ADD([[Transports]],[[Intel Omnipath (PSM2)]],[$1],[$ompi_check_psm2_happy])
     fi
 
     AS_IF([test "$ompi_check_psm2_happy" = "yes"],


ompi/mca/mtl/psm2/mtl_psm2.c
```diff
@@ -115,6 +115,12 @@ int ompi_mtl_psm2_module_init(int local_rank, int num_local_procs) {
 
     }
 
+    if (!psm2_get_capability_mask(PSM2_LIB_REFCOUNT_CAP)) {
```
jsquyres, line 118: Should this be in component selection (instead of module initialization)?

mwheinz:
Might as well get this out in public now: This change should not be merged until the matching version of libpsm2 is released. I'm doing this now because I want to get this into review before the branch is cut.

jsquyres:
> Might as well get this out in public now: This change should not be merged until the matching version of libpsm2 is released. I'm doing this now because I want to get this into review before the branch is cut.

Cool.  I converted this PR to "Draft" status, then.  Feel free to move it out of Draft status once you're ready.

mwheinz:
Okay. Updated based on previous feedback. Leaving in draft until libpsm2 update is published.

hppritcha:
bot:ompi:retest

mwheinz:
The matching version of libpsm2 has been released as [11.2.173](https://github.com/intel/opa-psm2/releases/tag/PSM2_11.2.173). This should be good to go.

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -323,6 +323,18 @@ ompi_mtl_psm2_component_open(void)
...
 {
+
+    /*
+     * Mixing the PSM2 MTL with the OFI BTL (using PSM2) 
+     * can cause an issue when they both call psm2_finalize
+     * in older versions of libpsm2.
+     */
+    if (!psm2_get_capability_mask(PSM2_LIB_REFCOUNT_CAP)) {
+        opal_output_verbose(2, ompi_mtl_base_framework.framework_output, 
+            "Please install version 11.2.173 version or later of the libpsm2 library.\n");
```
jsquyres, line 334: A suggestion (which you are free to use or not):

> The PSM2 MTL component will not work unless you have at least version 11.2.173 of the libpsm2 library installed.

This slightly changes the tone of the message to be: if you want to use the PSM2 MTL, then you have to install this library (vs. "you must install this library, no matter what you're doing").

mwheinz:
Guys? It looks like the build checker failed because of a problem with the build checker?

> Unsupported compiler gcc6.  Aborting
> Build step 'Execute shell' marked build as failure
> Finished: FAILURE

ompi/mca/mtl/psm2/mtl_psm2_component.c
mwheinz: I changed it to

> This version of the PSM2 MTL needs version 11.2.173 or later of the libpsm2 library for correct operation.

hppritcha:
I tried to build this on one of our Intel HFI1 (PSM2) systems but compile failed:
```
mtl_psm2_component.c: In function 'ompi_mtl_psm2_component_query':
mtl_psm2_component.c:332:35: error: 'PSM2_LIB_REFCOUNT_CAP' undeclared (first use in this function)
     if (!psm2_get_capability_mask(PSM2_LIB_REFCOUNT_CAP)) {
                                   ^
mtl_psm2_component.c:332:35: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [mtl_psm2_component.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory `/yellow/usr/projects/hpctools/hpp/ompi_pr7721/ompi/mca/mtl/psm2'
make[1]: *** [install-recursive] Error 1
make[1]: Leaving directory `/yellow/usr/projects/hpctools/hpp/ompi_pr7721/ompi'
make: *** [install-recursive] Error 1
hpp@gr-fey1[]:/usr/projects/hpctools/hpp/ompi_pr7721> (mwheinz-mwheinz-7721)
```

ompi/mca/mtl/psm2/mtl_psm2_component.c
```diff
@@ -323,6 +323,18 @@ ompi_mtl_psm2_component_open(void)
 static int
 ompi_mtl_psm2_component_query(mca_base_module_t **module, int *priority)
 {
+
+    /*
+     * Mixing the PSM2 MTL with the OFI BTL (using PSM2) 
+     * can cause an issue when they both call psm2_finalize
+     * in older versions of libpsm2.
+     */
+    if (!psm2_get_capability_mask(PSM2_LIB_REFCOUNT_CAP)) {
```
hppritcha, line 332: looks like this symbol isn't defined for our Intel HFI1 clusters psm2 install.  Need to protect this code with ifdef's.

mwheinz: > looks like this symbol isn't defined for our Intel HFI1 clusters psm2 install. Need to protect this code with ifdef's.

Howard, that function has been in libpsm2 for 3 years now. If your cluster is running a PSM2 that old it can't run the OFI MTL properly and has other problems as well.

mwheinz: Okay, so now I'm not sure what to do - if OMPI 5.x is built with a version of PSM2 that doesn't have PSM2_LIB_REFCOUNT_CAP defined, the PSM2 MTL isn't going to operate correctly.

The whole point of this patch is to require the correct version of libpsm2...

mwheinz: I mean, I can add a #warning or #error...

hppritcha: Which version of PSM2 defines PSM2_LIB_REFCOUNT_CAP?
We'd rather have these kind of restrictions enforced in the configury stage, so if one configured OMPI with --with-psm2 and the psm2 lib installed on the system is too old, the configure will fail with a hopefully informative error message.

mwheinz: I've pushed a patch that #defines the constant if it isn't defined. This should allow compilation to complete with the older libpsm2, but it will not execute until the libpsm2 is updated.

mwheinz: > We'd rather have these kind of restrictions enforced in the configury stage,

Sorry, I didn't see your reply before I pushed the previous commit.

I would agree but I know nothing about how to modify m4 files. This might take a while...

mwheinz: If the version of PSM2 is too old, configure will how report:

> PSM2 needs to be version 11.2.173 or later. Disabling MTL.

This is comparable to the configure error that occurs if you try to enable progress threads with PSM2.

mwheinz:
Still looking for a review for this. @hppritcha - did the changes to  config/ompi_check_psm2.m4 address you concerns?

jsquyres:
I pushed an additional commit to this PR with very minor updates -- it should be squashed appropriately before merging.

mwheinz:
> I pushed an additional commit to this PR with very minor updates -- it should be squashed appropriately before merging.

Squashed. Thanks for the assist.

rhc54:
Took myself off of this as I'm not familiar enough with this code are to comment on it.

bwbarrett:
bot:aws:retest

bwbarrett:
We should probably update the README and NEWS with this change.

hppritcha:
works for me now - at least giving an error in configure stage

mwheinz:
> works for me now - at least giving an error in configure stage

@hppritcha - the corresponding version of libpsm is now available at https://github.com/intel/opa-psm2/releases/tag/PSM2_11.2.173

PR #7734, opened 2020-05-13
Move to `libevent_core` and add checks for libevent.so conflict with LSF

jjhursey:
* LSF ships a `libevent.so` that is no related to the `libevent.so`
   shipped with Libevent.
 * Add some checks to the configure logic to detect scenarios where this
   conflict can be detected, and provide the user with a descriptive
   warning message.
   - When detected by `event/external` this is just a warning since
     the internal component may be able to be used instead.
     - This happens when the user supplies the LSF path via the
       `LDFLAGS` envar instead of via `--with-lsf-libdir`.
   - When detected by a LSF component and LSF was explicitly requested
     then this becomes an error. Otherwise, it will just print the warning
     and that component will fail to build.

### DIFF
--- a/opal/mca/event/external/configure.m4
+++ b/opal/mca/event/external/configure.m4
@@ -108,7 +108,7 @@ AC_DEFUN([MCA_opal_event_external_CONFIG],[
 
            OPAL_CHECK_PACKAGE([opal_event_external],
                               [event2/event.h],
-                              [event],
+                              [event_core],
                               [event_config_new],
                               [-levent_pthreads],
                               [$opal_event_dir],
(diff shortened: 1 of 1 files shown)


jjhursey:
Cherry Pick from https://github.com/open-mpi/ompi/pull/7698
 * This is the OMPI side of the patch. I'm working on the PRRTE/PMIx side of the patch for their mater branches.
 * I'll move this out of "draft" once all of the PRs are posted and linked together.

jjhursey:
* OpenPMix PR: https://github.com/openpmix/openpmix/pull/1766
 * PRRTE PR: https://github.com/openpmix/prrte/pull/554

PR #7751, opened 2020-05-19
update NEWS for 4.0.4rc2

hppritcha:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -59,6 +59,9 @@ included in the vX.Y.Z section and be denoted as:
 
 4.0.4 -- May, 2020
 -----------------------
+- Add checks to avoid conflicts with a libevent library shipped with LSF.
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -59,6 +59,8 @@ included in the vX.Y.Z section and be denoted as:
 
 4.0.4 -- May, 2020
 -----------------------
+- Add checks to avoid conflicts with a libevent library shipped with LSF.
```
gpaulsen, line 62: So this PR is more than just checks.  It actually changes our dependency from `-levent` to `-levent_core`.  That SHOULDn't matter, but perhaps we should mention that?  If you agree, I can update if you'd like.

hppritcha: I think you're right.  adding now.

PR #7758, opened 2020-05-22
coll/tuned: Fix dynamic message size for gather and scatter

wckzhang:
The gather and scatter operations did not use the correct message size
(Only did datatype size * com size). This did not correctly reflect the
total message size and prevents fine tuning within a com size. This
patch multiplies the value by the number of elements sent.


wckzhang:
I didn't look into changing scan or exscan in this patch.

Issue #7755, opened 2020-05-20
Open MPI master fails to initialize ofi MTL component


dancejic:
### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

master branch, but the issue began on commit 9f1081a07ac3c7b7277a27277ed970ed713207c9.

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

installed from git clone

### If you are building/installing from a git clone, please copy-n-paste the output from `git submodule status`.

 4a43c39c89037f52b4e25927e58caf08f3707c33 opal/mca/hwloc/hwloc2/hwloc (hwloc-2.1.0rc2-53-g4a43c39c)
 4996218a7a13a0ec59c94ffd72ead6d95d6acb4e opal/mca/pmix/pmix4x/openpmix (v1.1.3-2366-g4996218a)
 9cf321a38d36900702fd9c0862f633f4981b740d prrte (dev-30542-g9cf321a38d)

### Please describe the system on which you are running

* Operating system/version: Amazon Linux 2
* Computer hardware: 4.14.138-114.102.amzn2.x86_64
* Network type: 

-----------------------------

## Details of the problem
MTL ofi component is failing initialization when running examples/ring_c. This seems to be due to the include list being filled with values when it should be NULL.

```
$ mpirun --np 8 --mca mtl ofi --mca mtl_base_verbose 10 -v examples/ring_c
[ip-172-31-10-52:71727] Spawning job
[ip-172-31-10-52:71727] JOB [6899,1] EXECUTING
include: �
include: �
include: �
include: �
include: �
include: �
include: �
include: �
[ip-172-31-10-52.us-west-2.compute.internal:71738] mca: base: components_register: registering framework mtl components
[ip-172-31-10-52.us-west-2.compute.internal:71738] mca: base: components_register: found loaded component ofi
[ip-172-31-10-52.us-west-2.compute.internal:71738] mca: base: components_register: component ofi register function successful
[ip-172-31-10-52.us-west-2.compute.internal:71732] mca: base: components_register: registering framework mtl components
[ip-172-31-10-52.us-west-2.compute.internal:71732] mca: base: components_register: found loaded component ofi
[ip-172-31-10-52.us-west-2.compute.internal:71732] mca: base: components_register: component ofi register function successful
[...]

PR #7761, opened 2020-05-23
OFI common: set include list explicitly to NULL

hppritcha:
related to #7755


wckzhang:
I'll try this patch and see if it fixes the issue

wckzhang:
I think it resolved the issue for me. Was able to run with this patch. Thanks!

Issue #7644, opened 2020-04-17
OpenMPI ./configure --with-hwloc=<dir> fails in opal/mca/threads/pthreads/configure.m4


BrendanCunningham:
## Background information
Am building ompi/master and required hwloc, libevent, and pmix libraries under my `~/` to be able to develop without overwriting my distro's installed ompi. To do this, I configure hwloc, libevent, pmix, and ompi with `--prefix=~/projects/ompi_master_build_issue`.

Am using the following versions of hwloc, libevent, and pmix:
- libevent, heads/master, commit 0d2f1700
- hwloc, heads/master, commit 8249f30d
- pmix, heads/master, commit c9e85ea3

simple_build.sh that I have used to reproduce both failure and successes detailed below is include at end of thsi bug description. Script can and should be run as non-root user.

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)
ompi, master, commit 1a46e95b

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
git clone of ompi/master at hash mentioned above. Did `git submodule init && git submodule update` afterward.

### If you are building/installing from a git clone, please copy-n-paste the output from `git submodule status`.
```
git submodule status
4a43c39c89037f52b4e25927e58caf08f3707c33 opal/mca/hwloc/hwloc2/hwloc (hwloc-2.1.0rc2-53-g4a43c39c)
427dd6b738da5075620fbc4555e7c97b3a65fdb6 opal/mca/pmix/pmix4x/openpmix (v1.1.3-2321-g427dd6b7)
b624eae87078e32b3fb60a5dbe18557c2d276aa4 prrte (dev-30504-gb624eae8)
```

### Please describe the system on which you are running
* Operating system/version: RHEL 8.1, kernel 4.18.0-80.el8.x86_64
* Computer hardware: Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz, 
* Network type: OPA HFI in back-to-back configuration

## Details of the problem
Configuring and building ompi with internal {hwloc, libevent, pmix} succeeds (see simple_build.sh below). ompi internal-library ./configure output includes `checking if threads have different pids (pthreads on linux)... no`.

[...]

PR #7764, opened 2020-05-26
fix a problem with pthreads configure

hppritcha:
when using external hwloc, etc.

related to issue #7644


hppritcha:
@jsquyres will do, want to see if this actually helps Brandon first.

hppritcha:
closing this.  going to try a different approach.

PR #7771, opened 2020-05-29
common/ofi: Fixing compilation issue with ofi versions that do not support fi_info.nic

dancejic:
Added the flag OPAL_OFI_PCI_DATA_AVAILABLE to remove accessing the nic object in
fi_info when the ofi version does not support that structure.

### DIFF
--- a/config/opal_check_ofi.m4
+++ b/config/opal_check_ofi.m4
@@ -89,10 +89,11 @@ AC_DEFUN([_OPAL_CHECK_OFI],[
     OPAL_CHECK_WITHDIR([ofi-libdir], [$with_ofi_libdir],
                        [libfabric.*])
 
-    OPAL_VAR_SCOPE_PUSH([opal_check_ofi_save_CPPFLAGS opal_check_ofi_save_LDFLAGS opal_check_ofi_save_LIBS])
+    OPAL_VAR_SCOPE_PUSH([opal_check_ofi_save_CPPFLAGS opal_check_ofi_save_LDFLAGS opal_check_ofi_save_LIBS opal_check_fi_info_pci])
     opal_check_ofi_save_CPPFLAGS=$CPPFLAGS
     opal_check_ofi_save_LDFLAGS=$LDFLAGS
     opal_check_ofi_save_LIBS=$LIBS
+    opal_check_fi_info_pci=0
 
     opal_ofi_happy=yes
     AS_IF([test "$with_ofi" = "no"],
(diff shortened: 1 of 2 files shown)


bwbarrett:
ok to test

bwbarrett:
Since this is @dancejic's second PR this month, I added him to the ompi org so that we wouldn't have to approve his PRs for testing.

opal/mca/common/ofi/common_ofi.c
```diff
@@ -357,17 +378,18 @@ opal_mca_common_ofi_select_provider(struct fi_info *provider_list, int local_ind
...
 
-#if OPAL_DEBUG_ENABLE
+#if OPAL_OFI_PCI_DATA_AVAILABLE
     if (NULL != provider->nic) {
         pci = provider->nic->bus_attr->attr.pci;
         cpusets_match = compare_cpusets(opal_hwloc_topology, pci);
     }
+#endif
 
+#if OPAL_DEBUG_ENABLE
```
jsquyres, line 388: Sorry, I didn't notice these `opal_output()`s the first time around.  There should be on the `opal_common_ofi.output` stream -- not on fixed number streams like 1 or 10.

There's two `opal_output()`s earlier in this module that seem to be emitting error messages (vs. helpful / info messages) -- should they be converted to `opal_show_help()`?

config/opal_check_ofi.m4
```diff
@@ -120,6 +121,16 @@ AC_DEFUN([_OPAL_CHECK_OFI],[
                               [],
                               [opal_ofi_happy=no])])
 
+    AS_IF([test $opal_ofi_happy = yes],
+          [AC_CHECK_MEMBER([struct fi_info.nic],
+                           [opal_check_fi_info_pci=1],
+                           [opal_check_fi_info_pci=0],
+                           [[#include "$with_ofi/include/rdma/fabric.h"]])
+
+           AC_DEFINE_UNQUOTED([OPAL_OFI_PCI_DATA_AVAILABLE],
```
jsquyres, line 130: Minor nit: it doesn't actually matter (because Autoconf does the Right Thing), but technically the `AC_DEFINE_UNQUOTED` should probably outside of the `AS_IF` block (because you want to define the value regardless of the result of the test).

opal/mca/common/ofi/common_ofi.c
wckzhang: I don't think that needs to go into this patch though. I think we should get this merged so we can get your build running and then we can fix the output issues. I'd say that this is a critical blocker so we should get it merged ASAP.

jsquyres: Can you please make another commit / PR / whatever ASAP to fix these fixed-stream `opal_output()`s?  They're actually wrong -- you're writing to streams without having any idea what those streams are.  I.e., those outputs could show up for unexpected reasons.

dancejic: yes I pushed some of the other changes, but I'm working on addressing that as well. I think it'll need internal CR since I'll have to add a help-common-ofi.txt file and we'll have to agree on the error messages/info

jsquyres: If you want to take a little more time to fully convert to `opal_show_help()`, no problem.  But you could change the `opal_output(x, ...)` to `opal_output_verbose(...)` in the immediate future or somesuch.

I.e., I agree that converting to `opal_show_help()` is secondary: my primary concern was the hard-coded stream IDs.

dancejic: ok I switched them to `opal_output_verbose(...)` for now, I'll work on adding `opal_show_help()`

jsquyres: Thanks!

PR #7777, opened 2020-06-03
adding op-codes for syscall ipc for shmat/shmdt

markalle:

(the opcodes being undefined on the system made it select the


bwbarrett:
It looks like your commit message got truncated before the signed-off-by part

gpaulsen:
@markalle can you please check if this affects v3.0.x or v3.1.x on RHEL8, and if so, please cherry-pick there as well?

PR #7778, opened 2020-06-03
v4.0.x: adding op-codes for syscall ipc for shmat/shmdt

markalle:

(the opcodes being undefined on the system made the #define HAS_SHMDT
evaluate to false so intercept_shmat / intercept_shmdt were

(cherry picked from commit e8fab058dac7300569cb54b08e5500115f8bab8f)

### DIFF
--- a/opal/mca/memory/patcher/memory_patcher_component.c
+++ b/opal/mca/memory/patcher/memory_patcher_component.c
@@ -15,7 +15,7 @@
  *                         reserved.
  * Copyright (c) 2016-2017 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016-2019 IBM Corporation.  All rights reserved.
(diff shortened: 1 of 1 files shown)


gpaulsen:
@markalle Please update the commit comment.  As @bwbarrett pointed out on the master PR #7777, it looks like the commit message was truncated.

markalle:
I updated the commit, I was just meaning for it to describe how IPCOP_shmat/shmat not being defined in the system causes the macro to be false so intercept_shmat/shmdt go undefined

markalle:
I've submitted a bug report to gcc to see if they agree that the suspicious early store of r2 should be considered a bug, and gave a simple reproducer.  I'm also submitting a bug report to xlc because our compiler can do the same thing.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95692

PR #7786, opened 2020-06-04
dist: Update NEWS

bwbarrett:
It looks like we've been lazy updating the NEWS file in master.
Sync to reality.


bwbarrett:
Arg!  Not sure why the unicode got messed up.  Grumble.  marking do not merge until I fix that.

PR #7788, opened 2020-06-04
adding op-codes for syscall ipc for shmat/shmdt

markalle:

(the opcodes being undefined on the system made the #define HAS_SHMDT
evaluate to false so intercept_shmat / intercept_shmdt were

(cherry picked from commit e8fab058dac7300569cb54b08e5500115f8bab8f)
(cherry picked from commit 57c7d682334a997e302712842d1e4348bb03f2b6)


jsquyres:
Can you refresh this to cherry pick from the master commit?

markalle:
Sure, v4.0.x and v3.0.x are both cherry picks with a tiny merge edit.  So for v3.0.x I cherry picked from the v4.0.x, but logically it's a cherry pick from master.  I'll edit the checking message so it just has the e8fab05

gpaulsen:
CI failed with these two configure options:
```
open-mpi.build.autogen_options » --no-oshmem completed with result FAILURE
open-mpi.build.autogen_options » --no-orte completed with result FAILURE
```
I don't see how those are related to this PR though.

jsquyres:
bot:ompi:retest

jsquyres:
> CI failed with these two configure options:

Looks like transient SSL certificate issues.

jsquyres:
@markalle Is there a v3.1.x version of this PR?

PR #7802, opened 2020-06-10
mtl/ofi: break from progress loop when events are read

badgerious:
Once any number of events are read, return immediately, rather than
waiting for fi_cq_read() to return FI_EAGAIN or an error. This can
improve observed latency if the user application is in a blocking call
waiting for us to return. Breaking here also means
ofi_progress_event_count serves as an upper bound for the total number
of events read in a single call (without it, we could read far more, as
long as new events continue to arrive).

Signed-off-by: Eric Badger <eric@badgerio.us>

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -109,64 +109,58 @@ ompi_mtl_ofi_context_progress(int ctxt_id)
      * From the completion's op_context, we get the associated OFI request.
      * Call the request's callback.
      */
-    while (true) {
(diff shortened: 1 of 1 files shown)


bwbarrett:
ok to test

badgerious:
> However, given the rest of the loop, the right fix should be to remove the `while (true) {`, rather than just adding a `break`.

Good point. I've reorganized it to drop the while loop.

bwbarrett:
bot:aws:retest - looks like clang37 had a hang running the connectivity test.

awlauria:
bot:ompi:retest

awlauria:
bot:ompi:retest

awlauria:
bot:ompi:retest

awlauria:
bot:ompi:retest

PR #7807, opened 2020-06-11
Backport OFI changes from master to v4.1.x

bwbarrett:
This patch series backports all the significant changes from the OFI BTL, common, and MTL components.  The diff between master and v4.1.x are all repository-wide code cleanups that weren't worth pulling in.


rhc54:
Is there any data we provide in master that you need backported? If so, I can provide a commit to add some support to ORTE (don't think it is much).

hppritcha:
I'd like  9f1081a07ac3c7b7277a27277ed970ed713207c9 and 45b643d0cfa46f1abb9a5f43cf0ff304cf6a5fea brought over as well.  no sense to leave them out.
I think we also need ae2a447b0eddaac057beecdba99e10903051a2a7.

hppritcha:
looks good.   I will look at possibly adding the include/exclude common list in a way that doesn't mess up backwards compatibility but not a high priority.

gpaulsen:
@bwbarrett  This PR may have caused this MTT build failure last night: https://mtt.open-mpi.org/index.php?do_redir=3240

PR #7811, opened 2020-06-11
openmpi.spec: Remove libtool archive files in packaging

zhngaj:
Libtool archive files (.la files) create an unnecessary dependency
between linked applications and the development versions of packages
upon which Open MPI depends (to get the .so.1 -> .so symlink).

Remove .la libtool archive files to keep the best practice in
package builders.

Signed-off-by: Jie Zhang <zhngaj@amazon.com>

### DIFF
--- a/contrib/dist/linux/openmpi.spec
+++ b/contrib/dist/linux/openmpi.spec
@@ -150,6 +150,10 @@
 %global _binary_filedigest_algorithm 1
 %global _source_filedigest_algorithm 1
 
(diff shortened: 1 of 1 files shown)


bwbarrett:
ok to test

jsquyres:
I have a dim recollection that at least some users actually *want* the `.la` files.

Should this be a configurable option?  Perhaps it would be ok to default it to "off" (i.e., to remove the `.la` files)...?

zhngaj:
Updated with configure option.

PR #7823, opened 2020-06-16
Fix typos in OSC RDMA BTL allowlist

jsquyres:
OSC rdma had a reference to `openib`, which no longer exists on master.  It also had a typo for the `tcp` BTL (but even after fixing that typo, OSC rdma does not activate itself when the TCP BTL is used).

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -245,7 +245,7 @@ static int ompi_osc_rdma_component_register (void)
                                             MCA_BASE_VAR_SCOPE_GROUP, &mca_osc_rdma_component.locking_mode);
     OBJ_RELEASE(new_enum);
 
-    ompi_osc_rdma_btl_names = "openib,ugni,uct,ucp";
+    ompi_osc_rdma_btl_names = "ugni,uct,tcp";
     opal_asprintf(&description_str, "Comma-delimited list of BTL component names to allow without verifying "
              "connectivity. Do not add a BTL to to this list unless it can reach all "
              "processes in any communicator used with an MPI window (default: %s)",


jsquyres:
@bwbarrett @awlauria @rhc54 This is the cause of so many of the Cisco MTT failures over the past month or two.

bwbarrett:
Either someone is going to put in effort to fix the PT2PT component or fix the RDMA component.  Why would we put effort into the PT2PT component?

rhc54:
I believe nobody disputes that statement - the issue is: who is going to put in the effort? Simply removing p2p isn't the answer. To date, nobody has been willing to make the effort. Perhaps providing a clean mechanism by which BTLs can reject OSC operations with an appropriate error would solve it - but somebody would have to make the effort to create that too. 🤷‍♂️

hjelmn:
I will take a look at btl/tcp.

jsquyres:
Even if BTL tcp is fixed, do users running on a single node (e.g., a laptop) have to run `mpirun --mca btl sm,self,tcp` to have MPI_WIN_CREATE work properly?

jsquyres:
Per discussion on the weekly OMPI call today, I changed this PR to solely remove `openib` and fix the typo in the OSC pt2pt allowlist.

ompi/mca/osc/rdma/osc_rdma_component.c
```diff
@@ -245,7 +245,7 @@ static int ompi_osc_rdma_component_register (void)
                                             MCA_BASE_VAR_SCOPE_GROUP, &mca_osc_rdma_component.locking_mode);
     OBJ_RELEASE(new_enum);
 
-    ompi_osc_rdma_btl_names = "openib,ugni,uct,ucp";
+    ompi_osc_rdma_btl_names = "ugni,uct,tcp";
```
bwbarrett, line 248: should ofi be in this list?

jsquyres: Good question.  I defer to @hjelmn to answer that...

awlauria:
bot:ompi:retest

awlauria:
bot:ompi:retest

awlauria:
One more time..
bot:ompi:retest

awlauria:
bot:ompi:retest

PR #7826, opened 2020-06-16
More carefully test "alternate" short float type in configure

jsquyres:
Out of the box, Clang 6.0.x and 7.0.x nominally support the `_Float16` type.  Hence, checks like `AC_CHECK_TYPE(_Float16)` will pass.

But if you try to do any mathematical operations on variables of type `_Float16`, Clang 6 and 7 will fail to link the application, citing missing symbols such as `__gnu_h2f_ieee` and `__gnu_f2h_ieee`.  For example (on x86_64):

```
$ clang --version
clang version 7.0.0 (tags/RELEASE_700/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /cm/shared/apps/clang/7.0.0/bin
$ cat foo.c
#include <stdio.h>

{
    _Float16 a, b, c;

    a = 3.8;
    b = 2.5;
    c = a - b;

    return 0;
}
$ clang foo.c
/tmp/foo-3c5f99.o: In function `main':
foo.c:(.text+0x2c): undefined reference to `__gnu_h2f_ieee'
foo.c:(.text+0x3a): undefined reference to `__gnu_h2f_ieee'
foo.c:(.text+0x4b): undefined reference to `__gnu_f2h_ieee'
clang-7: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm not sure what those functions are, or if Clang optionally supports using some GNU library for half-precision mathematics, but the point is that there are cases where Clang 6 and 7 either require some kind of external library for half-precision math or just don't work.

So when the compiler passes checks like `AC_CHECK_TYPE(_Float16)`, also try linking a trivial program to make sure that that also works.  If it does, then go ahead and use the type.

This PR fixes Cisco MTT runs with Clang 6 and 7 ([see these sample MTT results](https://mtt.open-mpi.org/index.php?do_redir=3236)).

### DIFF
--- a/config/opal_check_alt_short_float.m4
+++ b/config/opal_check_alt_short_float.m4
@@ -1,6 +1,7 @@
 dnl -*- shell-script -*-
 dnl
-dnl Copyright (c) 2018      FUJITSU LIMITED.  All rights reserved.
+dnl Copyright (c) 2018-2020 FUJITSU LIMITED.  All rights reserved.
+dnl Copyright (c) 2020 Cisco Systems, Inc.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


ggouaillardet:
a quick google search suggests passing `--rtlib=compiler-rt` could help.
(not sure `compiler-rt` is built by default on all arch though).

note this is basically an emulation of short float, so it could be better not to support them at all on platforms with no native support such as `x86_64`.

jsquyres:
@ggouaillardet Your Google-fu was stronger than mine -- thanks!  😄   I confirmed with my small test that `clang foo.c --rtlib=compiler-rt` works like a champ for Clang 6 and 7.

We talked about this issue on the weekly call today.  No one really had any opinions, so I'm going to do this:

1. Adhere to the principle of "Open MPI shouldn't do behind-the-scenes magic unless it needs to," because that tends to lead to non-obvious interactions in the future.   I.e., let's _not_ silently add this `--rtlib` flag.
1. I'll add some verbiage in README, and possibly even here in `opal_check_alt_short_float.m4` that the user _may_ want to add `--rtlib` flag if they want software emulation of half precision floats.

I'll re-jigger this PR to do that.

jsquyres:
Re-jiggering done.  Please review.

kawashima-fj:
@jsquyres I'll review today. I'm sorry for the inconvenience.

kawashima-fj:
@jsquyres I added one more commit. Feel free to squash it into your commit.

I've checked the PR with several platforms/compilers/options and all worked as intended.

I also think that adding `--rtlib` silently is not a good idea.

FYI: Clang 6 and 7 supported `_Float16` for all platforms but Clang 8 and later amended it and supported it only for 64-bit Arm (AArch64), 32-bit Arm, and SPIR probablly because other platforms don't standardize the ABI (yet).

jsquyres:
@kawashima-fj I squashed your commit in -- thanks!

Issue #7809, opened 2020-06-11
MPI_File_read_at_all hanging


brtnfld:
Thank you for taking the time to submit an issue!

## Background information

Using CGNS with HDF5, we've run into a couple of use cases that cause the program to hang in collective H5Dread, specifically in MPI_File_read_at_all. For example, when running on 16 processes, 8 processes will complete MPI_File_read_at_all, but  the other 8 hang inside the MPI_File_read_at_all. We've tried it on two completely different systems.

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

Failed versions:

openmpi-4.02, 4.03, or 4.04rc1 with gcc-7.2.0 on linux-rhel7-skylake.
openmpi-4.0.2, master (3accffcb6edf55f457238e62f1bfde292e801463)  gcc=7.2.0 on centos 7

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

OpenMPI was installed from source.

## Details of the problem

Trace from the hanging processes:

```shell
0  0x00007f7fc5b02ebd in mca_btl_vader_component_progress () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libopen-pal.so.0
#1  0x00007f7fc5abf3ac in opal_progress () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libopen-pal.so.0
#2  0x00007f7fc62cec85 in mca_pml_ob1_recv () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libmpi.so.0
#3  0x00007f7fc6235c2b in mca_fcoll_vulcan_file_read_all () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libmpi.so.0
#4  0x00007f7fc61dced9 in mca_common_ompio_file_read_at_all () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libmpi.so.0
#5  0x00007f7fc6282a06 in mca_io_ompio_file_read_at_all () from /opt/pkgs/software/OpenMPI/master-GCC-7.2.0-2.29/lib/libmpi.so.0
[...]

PR #7828, opened 2020-06-16
common/ompio: use avg. file view size in the aggregator selection logic

edgargabriel:
This is a fix  based on a bugreport on github/mailing list from CGNS.
The core of the problem was that different processes entered different branches of
our aggregator selection logic, due to the fact that in some cases processes had
a matching file_view size and contiguous chunk size (thus assuming 1-D distribution),
and some processes did not (thus assuming 2-D distribution). The fix is to calculate
the avg. file view size across all processes and use this value, thus ensuring that
all processes enter the same branch.

Fixes issue #7809

(cherry picked from commit 4a8a330bbaf9fe5ea07cd01146afb83b569f3138)


jsquyres:
bot:aws:retest

PR #7834, opened 2020-06-17
tests/asm/run_tests: fix basename usage

### LINKED ISSUE #7804: Improper use of 'basename' in regression-tests
I'm building OpenMPI 4.0.2 & 4.0.4rc2 using the PGI 20.5 compiler.
Running the "make check" command, i get these messages
```
basename: extra operand ‘--test-name’
Try 'basename --help' for more information.
```

jsquyres:
Looks like this script was left over from quite a long time ago, and
was expecting CLI params from the "old"-style Automake test engine.
Update it to look for `--test-name` to get the test name, and update a
few other minor style things.

----

This commit fixes two minor output issues in `make check` output in `test/asm`:

1. The use of `basename` was incorrect and emitted a warning.
2. As a result, `$progname` was not populated correctly, and the test name was not emitted.

Fixes #7804.

**NOTE:** This commit could be cherry-picked back to v3.0.x and v3.1.x branches, but it's not worth it.  So I only listed master, v4.0.x, v4.1.x as the targets.

### DIFF
--- a/test/asm/run_tests
+++ b/test/asm/run_tests
@@ -1,26 +1,39 @@
 #!/bin/sh
+# Copyright (c) 2020 Cisco Systems, Inc.  All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
 
 retval=-1
+argv=$*
 
-progname="`basename $*`"
+set $*
+while shift; do
+    if test "$1" = "--test-name"; then
+        progname=`basename $2`
+        break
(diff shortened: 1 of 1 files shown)


jsquyres:
@bwbarrett Tagged you for review because you originally wrote+committed this script in https://github.com/open-mpi/ompi/commit/63bd314a0b6#diff-fe3e08d9ef3a6c2b80dfcc4271c8b495 back in 2005. 😲

Issue #7738, opened 2020-05-14
mpiexec -wdir and other MPI standard options are reported deprecated


abouteiller:
`mpiexec -wdir` is reported as deprecated, in favor of `--wdir`. 

This is problematic because the MPI standard, in "Chapter 8.8 Portable MPI Startup" states that mpiexec is optional, but if mpiexec is provided, it must be of a fixed form that accepts standardized arguments exactly (then provides a list of arguments that correspond to the MPI_INFO that are standardized to pass to MPI_COMM_SPAWN). 

All these parameters have a fixed, single dash form.

Full text of the standard for reference. 
```
While a standardized startup mechanism improves the usability of MPI, the range of
environments is so diverse (e.g., there may not even be a command line interface) that
MPI cannot mandate such a mechanism. Instead, MPI specifies an mpiexec startup
command and recommends but does not require it, as advice to implementors. However,
if an implementation does provide a command called mpiexec, it must be of the form
described below.

It is suggested that be at least one way to start <program> with an initial
MPI_COMM_WORLD whose group contains <numprocs> processes. Other arguments to
mpiexec may be implementation-dependent.


Analogous to MPI_COMM_SPAWN, we have <maxprocs>
mpiexec -n <numprocs> <program>
Advice to implementors. Implementors, if they do provide a special startup command for
MPI programs, are advised to give it the following form. The syntax is chosen in 
order that mpiexec be able to be viewed as a command-line version of MPI_COMM_SPAWN
(See Section 10.3.4).

mpiexec 
  -n 
  -soft
  -host
  -arch
  -wdir
  -path
  -file
...
       <command line>
```

Three options are available: 

1. do not deprecate the single-dash form of these arguments
2. work on the standard side to make double-dash form acceptable
3. I am wrong, and line 20 above "if an implementation does provide a command called mpiexec, it must be of the form described below" does not mandate that we use single-dash arguments to be compliant. 

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

master

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

not relevant.

markalle:
I'm not sure specifically about ```-wdir```; my reading of the standard makes it sound optional.  I do however see ```mpirun -configfile``` that doesn't look optional, so that may be an argument against the universal ban on single hyphen long command line options.

[...]

PR #7840, opened 2020-06-17
MPI-4: Initial error handler

abouteiller:
MPI-4 will introduce the concept of the 'initial error handler'. This is the error handler  that is set before MPI is initialized and after MPI is finalized, and can be selected from mpiexec/SPAWN info key parameters.

This also adds  MPI_ERROR_CLASS, and MPI_ERROR_STRING to the functions that can be called before/after MPI_INIT and are always thread safe.

More details from the MPI Forum site:
https://github.com/mpi-forum/mpi-standard/pull/50
https://github.com/mpi-forum/mpi-issues/issues/102 (if you do not have access to the standard repo).

### DIFF
--- a/ompi/dpm/dpm.c
+++ b/ompi/dpm/dpm.c
@@ -782,6 +782,7 @@ int ompi_dpm_spawn(int count, const char *array_of_commands[],
     int flag=0;
     char cwd[OPAL_PATH_MAX];
     char host[OPAL_MAX_INFO_VAL];  /*** should define OMPI_HOST_MAX ***/
+    char init_errh[OPAL_MAX_INFO_VAL];
     char prefix[OPAL_MAX_INFO_VAL];
     char stdin_target[OPAL_MAX_INFO_VAL];
     char params[OPAL_MAX_INFO_VAL];
@@ -814,6 +815,7 @@ int ompi_dpm_spawn(int count, const char *array_of_commands[],
        - "file": filename, where additional information is provided.
        - "soft": see page 92 of MPI-2.
        - "host": desired host where to spawn the processes
+       - "mpi_initial_errhandler": the error handler attached to predefined communicators.
       Non-standard keys:
        - "hostfile": hostfile containing hosts where procs are
        to be spawned
@@ -967,6 +969,15 @@ int ompi_dpm_spawn(int count, const char *array_of_commands[],
                 continue;
             }
 #endif
+
+            /* check for 'mpi_initial_errhandler' */
+            ompi_info_get (array_of_info[i], "mpi_initial_errhandler", sizeof(init_errh) - 1, init_errh, &flag);
+            if ( flag ) {
+                /* this is set as an environment because it must be available
+                 * before pmix_init */
+                opal_setenv("OMPI_MCA_mpi_initial_errhandler", init_errh, true, &app->env);
+                continue;
+            }
 
             /* 'path', 'arch', 'file', 'soft'  -- to be implemented */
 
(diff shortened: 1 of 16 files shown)


ompi/runtime/ompi_rte.c
```diff
@@ -743,6 +743,14 @@ int ompi_rte_init(int *pargc, char ***pargv)
         val = NULL;
     }
 
+    /* get our initial errhandler */
+    val = NULL;
+    OPAL_MODEX_RECV_VALUE_OPTIONAL(rc, "MPI_INITIAL_ERRHANDLER",
```
rhc54, line 748: I don't understand where this gets set - it appears to be some MPI-specific value, but that means PRRTE will never set it, and there is no way to successfully receive it here. Can you add some comments explaining who and where this gets set, and what the values are supposed to be?

abouteiller: Thanks, I removed it, it was a leftover from a prior version.

The env variable that replaces it is set from PRUN (ompi_schizo)

rhc54: If ompi/schizo knows what the value should be, then there is no issue with getting it from PMIx and prun - it just wasn't clear to me how it was being generated and how prrte would know the right value.

ompi/dpm/dpm.c
```diff
@@ -1058,6 +1060,13 @@ int ompi_dpm_spawn(int count, const char *array_of_commands[],
             }
 #endif
 
+            /* check for 'mpi_initial_errhandler' */
+            ompi_info_get (array_of_info[i], "mpi_initial_errhandler", sizeof(init_errh) - 1, init_errh, &flag);
```
rhc54, line 1064: We should revise this MPI_Info key name - if you look at the rest of them, you'll see they are all PMIx attributes. It will be somewhat confusing if we break that rule by introducing a lone exception. If we don't have an existing attribute we can use, then we should define one.

abouteiller: This could be a problem:, these keyS (wdir, host, SOFT, etc)  are MPI standard forms (just like the single dash arguments). MPI standard chapter 10.3.4. 

why has the PMIX prefix been added to these in the first place?

rhc54: Largely because I was unaware that MPI had "standardized" some info keys. Feel free to add the standard forms, but keep the PMIx ones as well - makes it simpler for people who are using those for other infos.

abouteiller: pr #7738

abouteiller:
Updated, ready for review.

ompi/errhandler/errhandler.c
```diff
@@ -78,6 +78,55 @@ ompi_predefined_errhandler_t ompi_mpi_errors_throw_exceptions = {{{0}}};
...
 
+static opal_mutex_t errhandler_init_lock = OPAL_MUTEX_STATIC_INIT;
+ompi_errhandler_t* ompi_initial_error_handler_eh = NULL;
+void (*ompi_initial_error_handler)(struct ompi_communicator_t **comm, int *error_code, ...) = NULL;
+
+/*
+ * Initialize the initial errhandler infrastructure only.
+ * This does not allocate any memory and does not require a corresponding fini.
+ */
+int ompi_initial_errhandler_init(void) {
```
bosilca, line 89: inconsistent indentation (please stay with 4 spaces)

abouteiller:
bot:retest

PR #7843, opened 2020-06-19
Some fixups for issues detected by clang-tidy

devreal:
Out of curiosity, I applied `clang-tidy` to Open MPI and found a bunch of issues. Many are false positives, quite a few are dead stores that likely will be optimized by DSE, some warnings are related to unsafe buffer handling functions such as `strcat`, and some are related to memory management. This PR tries to fix most of the detected potential memory leaks. All of them are missing free/release calls in error handling branches so the impact should be minimal. Still I thought this could be a useful patch.

Please review thoroughly as I am hunting in unknown territory here. I've split the changes into several commits for clarity but I'm happy to squash them before the merge.

### DIFF
--- a/ompi/mca/common/ompio/common_ompio_aggregators.c
+++ b/ompi/mca/common/ompio/common_ompio_aggregators.c
@@ -1333,14 +1333,12 @@ int mca_common_ompio_prepare_to_group(ompio_file_t *fh,
        if (NULL == aggr_bytes_per_group_tmp) {
           opal_output (1, "OUT OF MEMORY\n");
           ret = OMPI_ERR_OUT_OF_RESOURCE;
-          free(end_offsets_tmp);
           goto exit;
        }
     decision_list_tmp = (int* )malloc (fh->f_init_num_aggrs * sizeof(int));
     if (NULL == decision_list_tmp) {
         opal_output (1, "OUT OF MEMORY\n");
         ret = OMPI_ERR_OUT_OF_RESOURCE;
-        free(end_offsets_tmp);
         if (NULL != aggr_bytes_per_group_tmp) {
             free(aggr_bytes_per_group_tmp);
         }
--- a/ompi/mca/common/ompio/common_ompio_aggregators.c
+++ b/ompi/mca/common/ompio/common_ompio_aggregators.c
@@ -1303,12 +1303,14 @@ int mca_common_ompio_prepare_to_group(ompio_file_t *fh,
                                            fh->f_comm);
     if ( OMPI_SUCCESS != ret ) {
         opal_output (1, "mca_common_ompio_prepare_to_group: error in ompi_fcoll_base_coll_allgather_array\n");
+        free(start_offsets_lens_tmp);
         goto exit;
     }
     end_offsets_tmp = (OMPI_MPI_OFFSET_TYPE* )malloc (fh->f_init_procs_per_group * sizeof(OMPI_MPI_OFFSET_TYPE));
     if (NULL == end_offsets_tmp) {
         opal_output (1, "OUT OF MEMORY\n");
-        goto exit;
+        free(start_offsets_lens_tmp);
+        return OMPI_ERR_OUT_OF_RESOURCE;
     }
     for( k = 0 ; k < fh->f_init_procs_per_group; k++){
         end_offsets_tmp[k] = start_offsets_lens_tmp[3*k] + start_offsets_lens_tmp[3*k+1];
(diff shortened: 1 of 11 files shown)


ompi/mca/common/ompio/common_ompio_aggregators.c
```diff
@@ -1303,12 +1303,14 @@ int mca_common_ompio_prepare_to_group(ompio_file_t *fh,
...
+        free(start_offsets_lens_tmp);
         goto exit;
     }
     end_offsets_tmp = (OMPI_MPI_OFFSET_TYPE* )malloc (fh->f_init_procs_per_group * sizeof(OMPI_MPI_OFFSET_TYPE));
     if (NULL == end_offsets_tmp) {
         opal_output (1, "OUT OF MEMORY\n");
-        goto exit;
+        free(start_offsets_lens_tmp);
+        return OMPI_ERR_OUT_OF_RESOURCE;
     }
```
edgargabriel, line 1314: the same issue here. The arrays are actually freed in the layer above, even in case of an error. THis now would lead to a double free.

ompi/mca/common/ompio/common_ompio_aggregators.c
```diff
@@ -1333,14 +1333,12 @@ int mca_common_ompio_prepare_to_group(ompio_file_t *fh,
...
           opal_output (1, "OUT OF MEMORY\n");
           ret = OMPI_ERR_OUT_OF_RESOURCE;
-          free(end_offsets_tmp);
           goto exit;
        }
     decision_list_tmp = (int* )malloc (fh->f_init_num_aggrs * sizeof(int));
     if (NULL == decision_list_tmp) {
         opal_output (1, "OUT OF MEMORY\n");
         ret = OMPI_ERR_OUT_OF_RESOURCE;
-        free(end_offsets_tmp);
```
edgargabriel, line 1343: This is not correct. THere is an explicit comment at the end of the file:
/* Do not free aggr_bytes_per_group_tmp,
** start_offsets_lens_tmp, and end_offsets_tmp
** here. The memory is released in the layer above.
*/
Did you run any tests to verify this? I just looked at the code, and I am pretty sure that the layer above it frees these arrays even in the case of an error correctly.

ompi/mca/common/ompio/common_ompio_aggregators.c
devreal: I believe this is correct. The `free` and return happens before the out parameters are assigned so there is no way the upper layers may release that memory (the assignment happens on line 1323).

ompi/mca/common/ompio/common_ompio_aggregators.c
devreal: These `free` calls were erroneous as they would have led to double frees. at this point, `end_offsets_tmp` has been assigned to the out parameter and would have been free'd again in the caller.

devreal:
@edgargabriel These things are hard to test. Unless I can provoke the corresponding error conditions it is impossible to test these code paths. That's why cautious reviews are needed :)

ompi/mca/common/ompio/common_ompio_aggregators.c
edgargabriel: ok, you are actually correct in this case.

opal/mca/base/mca_base_alias.c
```diff
@@ -143,10 +143,12 @@ int mca_base_alias_register (const char *project, const char *framework, const c
 
         opal_hash_table_set_value_ptr (alias_hash_table, name, strlen(name), alias);
         free (name);
+        name = NULL;
     }
 
     mca_base_alias_item_t *alias_item = OBJ_NEW(mca_base_alias_item_t);
     if (NULL == alias_item) {
+        if (NULL != name) free(name);
```
jsquyres, line 151: FWIW, it's safe to `free(NULL)`.  I think we typically are more conservative in the OMPI code base and `if (foo != NULL) free(foo);`, but it's technically not necessary.

devreal: Yes, I was a bit hesitant to add the checks but then figured I'd stick to the coding style used in throughout the code base,

awlauria:
Good cleanup - thanks!

awlauria:
@devreal feel free to merge into master when ready. Thanks

devreal:
@awlauria It looks like I'm not authorized to merge PRs ;)

jsquyres:
@devreal Doh!  Well, I can fix that... done!

devreal:
@jsquyres Thanks, will merge :)

PR #7844, opened 2020-06-19
Some spawn info keys are standard mandated and cannot be deprecated

abouteiller:
This reverts deprecation of  spawn keys that are standard mandated (a sub-case of issue #7738).

This also lets info keys that directly correspond to `mpiexec` parameters stand as non-deprecated.

In any case, all info keys are converted to PMIX info values, and accept the PMIX prefixed version of the MPI info key. 


rhc54:
bot:aws:retest

PR #7845, opened 2020-06-19
Fix unexpected optimizations detected by STACK

devreal:
I applied the STACK tool to Open MPI to detect unintended or unexpected optimizations the compiler may apply based on undefined behavior [1]

1) `hook_comm_method_fns.c`: the check for `majority_method_offhost == -1` is always true because if it was `-1` before it was assigned the value of `this_method`, which has previously been used as index to `method_count` (https://github.com/open-mpi/ompi/compare/master...devreal:stack-fixes?expand=1#diff-f6b14d507479e3f54ae4f4d35681de08L785)

2) `osc_rdma_accumulate.c`: The check for `request` is assumed always true because of its previous dereference in line 217 (https://github.com/open-mpi/ompi/compare/master...devreal:stack-fixes?expand=1#diff-afdeeadc6df49f4ca2feee4ff3bda9a3L217)

3)  `osc_ucx_component.c`: Similar to the above, the check for `module` is always true since `module` has previously been dereferenced (https://github.com/open-mpi/ompi/compare/master...devreal:stack-fixes?expand=1#diff-f66f881fe0e7f40ffad2c3027e42c637L540) 

~4) `opal_cr.c`: This is the most interesting one. The check for `event >= OPAL_CR_INC_MAX` is always true because `event` is previously being used as index into a static array of size `OPAL_CR_INC_MAX`. Swapping the two checks fixes this.~ Removed in favor of https://github.com/open-mpi/ompi/pull/7848

5) `tm_malloc.c`: The pointer arithmetic on `ptr` causes the compiler to drop the following check for `NULL`.

[1] https://css.csail.mit.edu/stack/

### DIFF
--- a/ompi/mca/hook/comm_method/hook_comm_method_fns.c
+++ b/ompi/mca/hook/comm_method/hook_comm_method_fns.c
@@ -789,8 +789,7 @@ ompi_report_comm_methods(int called_from_location) // 1 = from init, 2 = from fi
                     if (majority_method_offhost == -1) {
                         majority_method_offhost = this_method;
                     }
-                    if (majority_method_offhost != -1 &&
-                        this_method != majority_method_offhost)
+                    if (this_method != majority_method_offhost)
                     {
                         uniformity_offhost = 0;
                     }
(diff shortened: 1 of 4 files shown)


devreal:
I removed the fix in `opal_cr.c` in favor of https://github.com/open-mpi/ompi/pull/7848 and moved the deregistration of the progress callback in `osc_ucx_component.c` so that it is always done in case of error, not just if the allocation of the `module` variable failed.

jsquyres:
This PR is a victim of #7847.

bot:ompi:retest

awlauria:
bot:ompi:retest

PR #7846, opened 2020-06-19
VERSION: Bump to 5.0.0a1

jsquyres:
Now that we have an actual 4.1.x branch, we shouldn't be calling
master "v4.1.0a1" any more.


jsquyres:
bot:ompi:retest

PR #7850, opened 2020-06-20
add a common ofi whitelist/blacklist

hkuno:
also add common verbose variable.

Note the verbosity thing is a little tricky owing to the way the MCA frameworks and components are registered and
and initialized.  The BTL's are registered/initialized prior to the MTL components even getting registered.

(cherry picked from commit 9f1081a07ac3c7b7277a27277ed970ed713207c9)


hkuno:
Closing because re-reading email, I realized that Brian said this change can't be made until a major version change.

bwbarrett:
@hkuno Just to be clear, we *could* make a change like this in 4.1, but it would have to still honor the MTL OFI include/exclude list, as well as the common OFI include/exclude list.

PR #7829, opened 2020-06-16
osc rdma: check for outstanding transfers before completing a request

devreal:
With non-contiguous RMA transfers, the operation is chunked up in multiple transfers. So far, osc/rdma did not check that all outstanding transfers have completed before releasing the associated request. This PR decrements the transfer counter and conditionally releases the request if all are completed.

See #7813 for the test case that triggers it.

Signed-off-by: Joseph Schuchart <schuchart@hlrs.de>

PR #7852, opened 2020-06-22
osc rdma: check for outstanding fragments before completing a request (v4.0.x)

devreal:
With non-contiguous RMA transfers, the operation is chunked up in multiple transfers. So far, osc/rdma did not check that all outstanding transfers have completed before releasing the associated request. This PR decrements the transfer counter and conditionally releases the request if all are completed.

See #7813 for the test case that triggers it.

Cherry-pick of https://github.com/open-mpi/ompi/pull/7829 to the v4.0.x branch.


gpaulsen:
@hppritcha I think we'll want this in v4.0.x once it's reviewed.

jsquyres:
> @hppritcha I think we'll want this in v4.0.x once it's reviewed.

...and v4.1.x.  😄

PR #7855, opened 2020-06-22
osc/rdma: fix bug in attach for non-debug builds

hjelmn:
This commit fixes an issue with non-debug builds where adding an
attachment to the attachment list doesn't actually happen. This
causes all MPI_Win_detach calls to fail. The call was within an
assert which is optimized out in optimized builds.

(cherry picked from commit 8ee80d885561c4349ab97725416613614d63c77f)


naughtont3:
Confirmed on linux desktop w/o debug enabled using ExaTensor reproducer.  Confirmed error w/o and no error w/ this patch. 👍

naughtont3:
bot:ompi:restart

naughtont3:
bot:ompi:retest

Issue #7763, opened 2020-05-26
mtl-portals4: Call add_procs through PML


tkordenbrock:
Shamelessly cribbed from Michael Heinz' cribbing of Brian Barrett's issue #7571 - Update Portals4 MTL to use the correct add_procs call.

> Change ompi_mtl_ofi_get_endpoint() to call the active PML's add_procs() rather than the OFI MTL add_procs() directly when discovering a new process during operation.
> Functionally, this has no impact in correct operation. However, the current behavior means that the heterogenous and active PML checks are not being executed in the dynamic discovery case.

jsquyres:
@tkordenbrock Are you going to back-port this to v4.0.x and v4.1.x?

tkordenbrock:
@jsquyres This was back-ported in #7890 and #7891, but the issue was never closed.  Sorry about that.  Closing now.

PR #7858, opened 2020-06-22
mtl-portals4: use the active PML to call add_procs()

tkordenbrock:
ompi_mtl_portals4_get_endpoint() was incorrectly making a direct
call to ompi_mtl_portals4_add_procs().  Instead use the actve PML
to call add_procs().  If add_procs() fails, call ompi_rte_abort()
to terminate the job.

Closes https://github.com/open-mpi/ompi/issues/7763


awlauria:
bot:ompi:retest

PR #7871, opened 2020-06-25
OSC UCX: make sure no-op fetch in rget/rput is properly aligned

devreal:
Starting with version 1.8.0, UCX barks at misaligned atomic operations so make sure the fetch-and-op used in rget/rput to acquire a UCX request uses properly aligned target memory. Also add a comment on whether this can be replaced with non-blocking flushes.

This fix should go into the next 4.0.x and 4.1.x releases as this issue breaks `MPI_Rput` and `MPI_Rget` for 32-bit integer when using osc/ucx, hence the quick fix.

See #7813 for the initial report.


awlauria:
bot:ompi:retest

jsquyres:
bot:aws:retest

PR #7873, opened 2020-06-25
OSC UCX: make sure no-op fetch in rget/rput is properly aligned (v4.1.x)

devreal:
Starting with version 1.8.0, UCX barks at misaligned atomic operations so make sure the fetch-and-op used in rget/rput to acquire a UCX request uses properly aligned target memory. Also add a comment on whether this can be replaced with non-blocking flushes.

Cherry-pick of #7871 to v4.1.x

See #7813 for the initial report.

(cherry picked from commit c1f7776341254b825a1b5f12fb93088c73c2c2ef)


jsquyres:
bot:aws:retest

Issue #7874, opened 2020-06-25
Compile: error (undeclared variable) in UCX osc component


tonycurtis:
Thank you for taking the time to submit an issue!

## Background information

Error in UCX component during compilation

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

git master @ 634f67b216fdcce841867f15ecd75538e65702e8

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### If you are building/installing from a git clone, please copy-n-paste the output from `git submodule status`.

```
 4a43c39c89037f52b4e25927e58caf08f3707c33 opal/mca/hwloc/hwloc2/hwloc (hwloc-2.1.0rc2-53-g4a43c39c)
 8283e81d1c0fd078e0f7fa85a383b633c328254b opal/mca/pmix/pmix4x/openpmix (v1.1.3-2431-g8283e81d)
 66c73f74cc4afd4ead5454771d98b5f199b7fe0e prrte (dev-30650-g66c73f74cc)
```

### Please describe the system on which you are running

* Operating system/version: CentOS 7.5.1804
* Computer hardware:  x86
* Network type:  IBV

-----------------------------

## Details of the problem

```
make[2]: Entering directory `/gpfs/projects/ChapmanGroup/src/openmpi/build/ompi/mca/osc/ucx'
depbase=`echo osc_ucx_comm.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
[...]

PR #7875, opened 2020-06-25
OSC UCX: Remove stale free from merge conflict

devreal:
There was a merge conflict between https://github.com/open-mpi/ompi/pull/6980 and https://github.com/open-mpi/ompi/pull/7843 that inserted a stray call to `free`.

Fixes https://github.com/open-mpi/ompi/issues/7874


jsquyres:
bot:aws:retest

artpol84:
@devreal
I’m not sure this is recommended approach in OMPI, but I always rebase on top of the branch and force push.
I’m this case we would see it in CI for #6980 instead of later on a random PR.
When merging #6980, for some reasons, I assumed it was done

devreal:
@artpol84 I agree, in this case I didn't realize that the two PRs modified the same file and both got merged (one by me) without a warning about the merge conflict. Otherwise I would have rebased before the merge. Not sure why that went through...

PR #7876, opened 2020-06-25
Fix UCX OSC compilation.

bosilca:


jsquyres:
Check out #7875

bosilca:
I missed #7875.

PR #7877, opened 2020-06-26
Change naming for thread_create 

janjust:
Since `opal_tsd_key_create()` adds the created key to the global data structure that is not accessible by the API user, the name seems confusing.
Especially in the presence of `opal_tsd_key_delete()` that directly calls `pthread_key_delete()`.
In order to resolve that and give some cases the flexibility to track the keys themselves, we suggest to create 2 distinct functions:
* `opal_tsd_key_create()` that is a wrapper around `pthread_key_create()`
* `opal_tsd_key_create_tracked()` that implements the current logic

### DIFF
--- a/opal/mca/threads/qthreads/threads_qthreads_module.c
+++ b/opal/mca/threads/qthreads/threads_qthreads_module.c
@@ -90,3 +90,8 @@ int opal_tsd_keys_destruct(void)
 {
     return OPAL_ERR_NOT_IMPLEMENTED;
 }
+
+int opal_tls_key_create(opal_tsd_key_t *key, opal_tsd_destructor_t destructor)
+{
+    return OPAL_ERR_NOT_IMPLEMENTED;
+}
--- a/opal/mca/threads/argobots/threads_argobots_module.c
+++ b/opal/mca/threads/argobots/threads_argobots_module.c
@@ -162,3 +162,11 @@ int opal_tsd_keys_destruct(void)
     opal_mutex_unlock(&opal_tsd_lock);
     return OPAL_SUCCESS;
 }
+
+int opal_tls_key_create(opal_tsd_key_t *key, opal_tsd_destructor_t destructor)
+{
+    opal_threads_argobots_ensure_init();
+    int rc;
+    rc = ABT_key_create(destructor, key);
+    return (ABT_SUCCESS == rc) ? OPAL_SUCCESS : OPAL_ERROR;
+}
--- a/opal/mca/threads/pthreads/threads_pthreads_module.c
+++ b/opal/mca/threads/pthreads/threads_pthreads_module.c
@@ -137,3 +137,9 @@ int opal_tsd_keys_destruct(void)
 void opal_thread_set_main(void)
 {
 }
+
+int opal_tls_key_create(opal_tsd_key_t *key, opal_tsd_destructor_t destructor)
+{
+    int ret = pthread_key_create(key, destructor);
+    return 0 == ret ? OPAL_SUCCESS : OPAL_ERR_IN_ERRNO;
+}
--- a/opal/mca/threads/tsd.h
+++ b/opal/mca/threads/tsd.h
@@ -142,6 +142,35 @@ OPAL_DECLSPEC int opal_tsd_key_create(opal_tsd_key_t *key,
 
 
 /**
+ * Create thread-specific data key
+ *
+ * Create a thread-specific data key visible to all threads in the
+ * current process.  The returned key is valid in all threads,
+ * although the values bound to the key by opal_tsd_setspecific() are
+ * allocated on a per-thread basis and persist for the life of the
+ * calling thread.
+ *
+ * Upon key creation, the value NULL is associated with the new key in
+ * all active threads.  When a new thread is created, the value NULL
+ * is associated with all defined keys in the new thread.
+ *
+ * The destructor parameter may be NULL.  At thread exit, if
+ * destructor is non-NULL AND the thread has a non-NULL value
+ * associated with the key, the function is called with the current
+ * value as its argument.
+ *
+ * @param key[out]       The key for accessing thread-specific data
+ * @param destructor[in] Cleanup function to call when a thread exits
+ *
+ * @retval OPAL_SUCCESS      Success
+ * @retval OPAL_ERR          Error
+ * @retval OPAL_ERR_IN_ERRNO Error
+ */
+OPAL_DECLSPEC int opal_tls_key_create(opal_tsd_key_t *key,
+                                      opal_tsd_destructor_t destructor);
+
+
+/**
  * Destruct all thread-specific data keys
  *
  * Destruct all thread-specific data keys and invoke the destructor


ibm-ompi:

 Gist: https://gist.github.com/39ef8ca2b20ea2e8e115b4a0112dd0dc

ibm-ompi:

 Gist: https://gist.github.com/3506091d90086cf011665fe4ee7fc4c1

ibm-ompi:

 Gist: https://gist.github.com/7803cc8b69e49a26bffb89ab77681744

artpol84:
bot:retest

opal/mca/threads/qthreads/threads_qthreads_module.c
```diff
@@ -81,6 +81,11 @@ int opal_thread_start(opal_thread_t *t)
 
 OBJ_CLASS_DECLARATION(opal_thread_t);
 
+int opal_tsd_key_create_tracked(opal_tsd_key_t *key, opal_tsd_destructor_t destructor)
```
artpol84, line 84: @hppritcha is there a way to implement this functionality with qthreads?

hppritcha: pinging the qthreads and argobots crews

npe9: Any documentation on the tracked keys anywhere? @hppritcha @artpol84

shintaro-iwasaki: @janjust Could you please apply this patch? It will fix the implementation in the Argobots module.

```diff
diff --git a/opal/mca/threads/argobots/threads_argobots_module.c b/opal/mca/threads/argobots/threads_argobots_module.c
index 106549a..f166698 100644
--- a/opal/mca/threads/argobots/threads_argobots_module.c
+++ b/opal/mca/threads/argobots/threads_argobots_module.c
@@ -141,7 +141,10 @@ int opal_tsd_key_create_tracked(opal_tsd_key_t *key, opal_tsd_destructor_t destr
 
 int opal_tsd_key_create(opal_tsd_key_t *key, opal_tsd_destructor_t destructor)
 {
-    return OPAL_ERR_NOT_IMPLEMENTED;
+    opal_threads_argobots_ensure_init();
+    int rc;
+    rc = ABT_key_create(destructor, key);
+    return (ABT_SUCCESS == rc) ? OPAL_SUCCESS : OPAL_ERROR;
 }
 
 int opal_tsd_keys_destruct(void)
```

bosilca:
I wanted to understand the reason behind this patch so I looked a little in the TSD code and I think it is incorrect. The only reasons to have this tracking mechanism would be to help remove all OMPI specific keys from the master thread (the one that initialized the library). That would make sense except that the cleanup (aka opal_tsd_keys_destruct) is called during OPAL finalization (more specifically opal_finalize_util), which is the last call before completely returning from the MPI_Finalize. Thus, at that moment all OMPI threads have been turned off (so pthread already called the keys destructors on their context), and more importantly all OMPI components have already called the opal_tsd_key_delete on the keys they created. Thus, according the the man page, it is undefined behavior to use the keys for any purpose (and especially for the getspecific call we are using in the opal_tsd_keys_destruct.

Am I missing something, or this is working because pthread is extremely lenient with calls to getspecific after the key has been deleted ?

Then coming back to the purpose of this patch, without the tracking we would leak OMPI specific keys into the application space after the MPI finalization. Why ?

artpol84:
@bosilca
1. We keep the old way of managing keys, we just rename it to have "_tracked" suffix to indicate, that the key created was placed into the internal bookkeeping structures.
1. The reason we want this is to have flexibility to use OPAL abstraction but manage keys ourselves. i.e. release them before MPI_Finalize.

artpol84:
P.S. this patch doesn't (or at list do not intend to) introduce any semantic changes - all the keys management logic should remain intact. We are going to leverage `opal_tsd_key_create()` in subsequent work.

opal/mca/threads/qthreads/threads_qthreads_module.c
artpol84: @shintaro-iwasaki what is the counterpart for this? the analog of pthread_key_delete?

shintaro-iwasaki: Sorry for the late reply. Yes, it is. This patch implements "`OPAL_ERR_NOT_IMPLEMENTED`" in `opal_tsd_key_create()` of the Argobots implementation.

janjust: @shintaro-iwasaki done, but just fyi, the _tracked version is left as unimplemented.

devreal: @janjust The description of https://github.com/open-mpi/ompi/pull/7877/commits/a4ce523d51ce4525687e308b046a05a84a3ca352 is confusing. How can it implement `OPAL_ERR_NOT_IMPLEMENTED`?

janjust: It's a copy paste of @shintaro-iwasaki description, since it's his patch that was applied. But I can change it - I'm easy :)

shintaro-iwasaki: I would like to clarify this situation:
- There are three MCA thread modules: Pthreads, Qthreads, and Argobots
- The first commit by @janjust only implemented Pthreads. The Argobots/Qthreads counterparts were marked as `OPAL_ERR_NOT_IMPLEMENTED`.
- I created a patch that fixes the "Argobots" counterpart. (`opal/mca/threads/argobots/threads_argobots_module.c`)
- The Qthreads counterpart is still marked as `OPAL_ERR_NOT_IMPLEMENTED` because the Qthreads module has not implemented TSD yet (`opal/mca/threads/qthreads/threads_qthreads_module.c`).

shintaro-iwasaki: @janjust If you don't mind, please squash-merge my patch. This just complements your original commit.

bosilca:
> 1. We keep the old way of managing keys, we just rename it to have "_tracked" suffix to indicate, that the key created was placed into the internal bookkeeping structures.

I see that, but I still don't think this is correct. Looking at the execution path where these keys are used, it seems that some of them might spill over to user threads, in which case destroying them before MPI_Finalize would not be correct.

> 2. The reason we want this is to have flexibility to use OPAL abstraction but manage keys ourselves. i.e. release them before MPI_Finalize. So we would use `opal_tsd_key_create()` (as opposed to `opal_tsd_key_create_tracked()`) and ensure we call `opal_tsd_key_destroy()`

You already are among the largest users of the TSD in the current codebase. Why suddenly a new API is mandatory remain an open question.

opal/mca/threads/qthreads/threads_qthreads_module.c
janjust: @shintaro-iwasaki done

shintaro-iwasaki: @janjust Thank you very much!

devreal: Looking at the way TSDs are used in the UCX integration I wonder if the patch provided by @shintaro-iwasaki really does what we want/expect. Provided I build my Open MPI installation with Argobots support and use tasks that are mapped to Argobots (using BOLT for example) and then issue only one or two MPI calls in each task, the thread-specific data has to be created over and over again. It seems that `_tlocal_add_ctx` and `_tlocal_ctx_connect` are pretty expensive and that would have to be done for *each task*, which seems like a prohibitively large overhead. @shintaro-iwasaki please correct me if I'm wrong here. From a UCX standpoint, pthread-based TSDs should be sufficient, no?

artpol84:
@bosilca 

>I see that, but I still don't think this is correct. Looking at the execution path where these keys are used, it seems that some of them might spill over to user threads, in which case destroying them before MPI_Finalize would not be correct.

It might be. And I can see 2 reasons for it: 
1. incomplete API (see my thoughts below).
2. incorrect API usage, in this case it's outside of the scope of this PR as we are not changing the logic, its an API change.

>> The reason we want this is to have flexibility to use OPAL abstraction but manage keys ourselves. i.e. release them before MPI_Finalize. So we would use opal_tsd_key_create() (as opposed to opal_tsd_key_create_tracked()) and ensure we call opal_tsd_key_destroy()

>You already are among the largest users of the TSD in the current codebase. Why suddenly a new API is mandatory remain an open question.

My initial understanding was that our use case differs from the cases in the existing codebase. 
We want to create/destroy the keys arbitrary number of times throughout the runtime of the application.
With the existing API implementation, `opal_tsd_key_create()` will append the new key to the internal list, and `opal_tsd_key_delete()` will not remove it. The removal will only happen at the MPI_Finalize, which in our case will result in the accumulation of unused keys.

Based on your responses, I understand that you are leaning towards fixing the existing API.
Does that mean to make `opal_tsd_delete_key() to remove the keys from the internal list? If that's the case it's fine with us as well.

opal/mca/threads/qthreads/threads_qthreads_module.c
artpol84: @janjust I'd prefer to split our changes and changes provided by @shintaro-iwasaki so the contribution is properly distributed.

shintaro-iwasaki: @devreal Thanks! I am not fully sure how UCX uses TLS, but it could be expensive if all tasks/threads create UCX contexts as you wrote.
In terms of correctness, I do not know if tasks can safely share a UCX context that is associated with Pthreads. For example, if a UCX context has a state and each "thread" that owns it needs to remember its state, yielding a task after changing the state can cause a problem.
If there is a scheduling limitation it might work, but this is not trivial to me so far. In any case, this should be a performance problem that needs to be solved.

devreal: Thanks, I believe this should be investigated separately. I agree with @artpol84 that this patch should be separated out into its own PR as this is an orthogonal issue to the proposal here.

artpol84:
@bosilca can you comment on the above?

@jsquyres @jladd-mlnx @gpaulsen Can we discuss this on the upcoming telecon? (I'll be there next week).
We would like to have some clarity on how this will be resolved.

janjust:
Closing in favor of #7936 which has a redesign of some of the discussion points raised in this PR as well on ompi calls.

PR #7882, opened 2020-06-26
osc rdma: check for outstanding fragments before completing a request (II)

devreal:
This PR fixes the same issue as https://github.com/open-mpi/ompi/pull/7829 but for a different code path (`ompi_osc_rdma_put_complete_flush`, taken when using `btl/uct` for example).


awlauria:
Thanks - please cherry-pick to release branches as well.

hjelmn:
A clear indication we need more tests for the R variety of RMA operations :).

PR #7883, opened 2020-06-27
UCX/WPOOL: fixed potential deadlock

hoopoepg:
- fixed potential deadlock in error processing

### DIFF
--- a/opal/mca/common/ucx/common_ucx_wpool.h
+++ b/opal/mca/common/ucx/common_ucx_wpool.h
@@ -470,7 +470,6 @@ opal_common_ucx_wpmem_cmpswp_nb(opal_common_ucx_wpmem_t *mem, uint64_t compare,
     rc = _periodical_flush_nb(mem, winfo, target);
     if(OPAL_UNLIKELY(OPAL_SUCCESS != rc)){
         MCA_COMMON_UCX_VERBOSE(1, "_incr_and_check_inflight_ops failed: %d", rc);
-        return rc;
     }
 
     opal_mutex_unlock(&winfo->mutex);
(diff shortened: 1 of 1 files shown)


hoopoepg:
bot:retest

hoopoepg:
curl: (77) Problem with the SSL CA cert (path? access rights?)
Build step 'Execute shell' marked build as failure

bot:retest

opal/mca/common/ucx/common_ucx_wpool.h
```diff
@@ -470,7 +470,6 @@ opal_common_ucx_wpmem_cmpswp_nb(opal_common_ucx_wpmem_t *mem, uint64_t compare,
     rc = _periodical_flush_nb(mem, winfo, target);
     if(OPAL_UNLIKELY(OPAL_SUCCESS != rc)){
         MCA_COMMON_UCX_VERBOSE(1, "_incr_and_check_inflight_ops failed: %d", rc);
-        return rc;
```
brminich, line 473: should not we also `goto` line 474 if `opal_common_ucx_atomic_cswap_nb` completed immediately (which is possible in case of error)?

hoopoepg: in case if more actions are added after this check - yes, we have to jump, but for now it is not needed

brminich: it does `_periodical_flush_nb`
i meant checking result of `opal_common_ucx_atomic_cswap_nb`, not the line you fixed

hoopoepg: not sure if such jump doesn't break anything else

devreal:
Good catch! This problem exists in the same place in multiple functions in this file. Please fix all of them:

- `opal_common_ucx_wpmem_putget`
- `opal_common_ucx_wpmem_cmpswp`
- `opal_common_ucx_wpmem_post`
- `opal_common_ucx_wpmem_fetch`
- `opal_common_ucx_wpmem_fetch_nb`

hoopoepg:
will do

hoopoepg:
@devreal I updated PR, but it seems something wrong with github - changes are not reflected in WEB UI

hoopoepg:
@artpol84 back it port into 4.x branches?

Issue #7742, opened 2020-05-16
`mpirun` failed with `--map-by core`


karasevb:
### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)
git clone -b v4.0.x ...

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
./configure --prefix=`pwd`/install
make install

### Please describe the system on which you are running

* Operating system/version: RHEL 7.6
* Computer hardware: x86_64
* Network type: 

-----------------------------

## Details of the problem

`--map-by core` looks broken, while `--map-by node` works well:
```shell
$mpirun --map-by node -np 2 ./hello_c
Hello, world, I am 0 of 2, (Open MPI v4.0.4rc1, package: Open MPI user@hpchead Distribution, ident: 4.0.4rc1, repo rev: v4.0.4rc1-2-g09b57c8, Unreleased developer copy, 143)
Hello, world, I am 1 of 2, (Open MPI v4.0.4rc1, package: Open MPI user@hpchead Distribution, ident: 4.0.4rc1, repo rev: v4.0.4rc1-2-g09b57c8, Unreleased developer copy, 143)

$mpirun --map-by core -np 2 ./hello_c
[jazz17:167156] PMIX ERROR: OUT-OF-RESOURCE in file client/pmix_client.c at line 231
[jazz17:167156] OPAL ERROR: Error in file pmix3x_client.c at line 112
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[jazz17:167156] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[5437,1],0]
  Exit code:    1
--------------------------------------------------------------------------
```
```shell
$mpirun --map-by core -np 2 hostname
jazz17
<hanging....>
```

awlauria:
@karasevb do you only see this on the 4.0.x branch? Or does it also happen on master?

[...]

PR #7887, opened 2020-06-28
Fix rank-by computation

### LINKED ISSUE #7867: v4.x: Using --map-by core:span causes error in MPI_Init
`4.0.4` from release tarball. ~~Can't test the git version since it requires `pandoc`.~~

### Describe how Open MPI was installed
OpenMPI configuration:
`./configure --prefix=$PREFIX --without-verbs --with-hwloc=internal`

* Operating system/version: CentOS Linux release 7.5.1804
* Computer hardware: Intel Xeon CPU E5-2690 v3
* Network type: Mellanox Technologies MT27600 [Connect-IB]

A few more systems were tested (laptops, ARM machines) and they all have the same issue.

-----------------------------

## Details of the problem

Compile the following MPI program:

```
#include "mpi.h"

int main() {
    MPI_Init(NULL, NULL);
    MPI_Finalize();
}
```

Run it with
`mpirun -n 4 --map-by core:span ./a.out`

Error message:
```
[login2:55588] PMIX ERROR: NOT-FOUND in file dstore_base.c at line 2866
[login2:55588] PMIX ERROR: NOT-FOUND in file server/pmix_server.c at line 3408
[login2:55592] PMIX ERROR: OUT-OF-RESOURCE in file client/pmix_client.c at line 231
[login2:55592] OPAL ERROR: Error in file pmix3x_client.c at line 112
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[login2:55592] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
mpirun detected that one or more processes exited with non-zero status, thus causing

  Process name: [[21660,1],0]
  Exit code:    1
[login2:55588] PMIX ERROR: ERROR in file gds_ds21_lock_pthread.c at line 99
[login2:55588] PMIX ERROR: ERROR in file gds_ds21_lock_pthread.c at line 99
```

I tested this on both the login node and the compute node, the error persists as long as I pass `--map-by core:span` or `--map-by numa:span` to `mpirun`.

OpenMPI 4.0.3 works fine.

rhc54:
Correctly increment vpid upon assignment

Fixes #7867 

### DIFF
--- a/orte/mca/rmaps/base/rmaps_base_ranking.c
+++ b/orte/mca/rmaps/base/rmaps_base_ranking.c
@@ -456,7 +456,7 @@ static int rank_by(orte_job_t *jdata,
                             continue;
                         }
                         /* assign the vpid */
-                        proc->name.vpid = vpid;
+                        proc->name.vpid = vpid++;
                         if (0 == cnt) {
                             app->first_rank = proc->name.vpid;
                         }


karasevb:
Fixes #7742

rhc54:
@hppritcha It depends on what you want to accomplish. This PR fixes the specific problem.

However, there are some additional issues having to do with inheritance of map/rank/bind options from the parent job (on comm_spawn) that remain. I didn't try to address those here as it would introduce a change in behavior from earlier v4.0.x releases, and nobody has raised an issue over them. One could argue that it would be fixing a bug as the inheritance really isn't as intended in the current code in this branch - but one could also argue that it is now the expected behavior.

Up to you guys.

PR #7896, opened 2020-06-30
v4.1.x: Increment the vpid after assignment

rhc54:
Fix the rank-by operation


gpaulsen:
Fixes: https://github.com/open-mpi/ompi/issues/7867

PR #7905, opened 2020-07-02
Fix shifting signed integer

cniethammer:
Shifting signed int will lead to undefined behavior in case of 1<<31 here

Signed-off-by: Christoph Niethammer niethammer@hlrs.de


cniethammer:
Extended commit to fix a second location with the same problem

jsquyres:
bot:ompi:retest

PR #7906, opened 2020-07-02
common/ofi: added address format check to fix provider selection

dancejic:
bugfix for:
    commit 167d75b42ac3ca4770d59c796c011d72e0fffde3
    common/ofi: Added multi-NIC support to provider selection

Adding a check for address format to provider selection to
ensure that all nodes use the same address format. provider
selection would not differentiate between ipv4 and ipv6
addresses which would cause some nodes to be unable to communicate.


dancejic:
Changed commit message to:

common/ofi: added address format check to fix provider selection

    bugfix: provider selection would not differentiate between ipv4
    and ipv6 addresses which would cause some nodes to be unable
    to communicate between each other. Adding a check for address
    format to provider selection to ensure that all nodes use the
    same address format.

PR #7907, opened 2020-07-03
Fix null pointer arithmetic resulting in potential undefined behavior

cniethammer:
NULL pointer arithmetic is undefined behaviour in c.
The payload_ptr can be NULL in the moment when mpool is not initialized.

References from the c11 standard:
- 6.5.6 Additive operators
- 6.3.2.3 Pointers

Signed-off-by: Christoph Niethammer <niethammer@hlrs.de>


jsquyres:
bot:aws:retest

PR #7910, opened 2020-07-06
Trivial helper script to git clean submodules

jsquyres:
Since we added the use of git submodules recently, this trivial script has been helpful to me to "git clean" not only the top-level Open MPI repo, but also all the included submodules, too.

NOTE: this script does the (harsh) "git clean -dfx" command, which deletes everything that git does not know about.  Use with care!

This is just a script that is helpful to me (as a developer of Open MPI).  I open it here as a PR on the off chance that it might be helpful to other developers, too.


jsquyres:
Weird SSL verification error in AWS CI.  Try again...

bot:aws:retest

PR #7914, opened 2020-07-07
Fix icc 2330 warning for stdatomic.h

bosilca:
The root of the issue is the the Intel compiler makes a strong distinction between _Atomic and volatile variables, in the sense that these two qualified are conflicting. According to the C11 standard this should be correct, except that atomic operations can be applied to _Atomic variables (they should be inherently volatile).

This is a potential path to fix this issue.

### DIFF
--- a/ompi/communicator/comm_request.c
+++ b/ompi/communicator/comm_request.c
@@ -99,7 +99,7 @@ int ompi_comm_request_schedule_append (ompi_comm_request_t *request, ompi_comm_r
 static int ompi_comm_request_progress (void)
 {
     ompi_comm_request_t *request, *next;
-    static opal_atomic_int32_t progressing = 0;
+    static volatile int32_t progressing = 0;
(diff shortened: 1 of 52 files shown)


hjelmn:
But this is going the wrong way. We are trying to go from volatile to _Atomic with the goal of switching from internal atomics to C11 atomics in all cases at some point.

hjelmn:
Hmm, looks like we just need to add volatile to the opal_atomic types.

hjelmn:
Hmm, no. That doesn't make it go away. For some reason we are getting volatile void * as the argument type always for the opal_atomic functions even though they are defined to take opal_atomic types.

bosilca:
After discussing with Nathan we decided to not go this route, because it is technically incorrect from the C11 perspective. I will leave this branch around for future use, but will close the PR and will propose a simpler approach.

PR #7915, opened 2020-07-08
Second take on fixing the Intel _Atomic atomic operation warning

bosilca:
We completely disable C11 atomic op support for _Atomic for
all Intel compiler prior to 20200310 (which is currently the
latest released), by switching to our pre-C11 atomic operations.


jsquyres:
@hjelmn @bosilca Is this good to go?

PR #7916, opened 2020-07-08
config: add support for setting the wrapper C compiler

hjelmn:
This PR adds a new configure option: --with-wrapper-cc.
This option allows the user to set the compiler that will
be invoked by mpicc, shmemcc, etc. This allows the user
to build Open MPI with one compiler (a C standards compliant
compiler like clang or gcc) and have the wrapper use
another compiler (icc for example). This allows building
Open MPI with the best available compiler while still
supporting compiling Open MPI for a specific compiler
suite.


hjelmn:
I am more than a little annoyed at the Intel compiler suite so I am reviving this PR. It makes sense to allow an Open MPI installation to be compiled with one C compiler then used with another C compiler. This is the norm in the Linux world where the installed libraries are often compiled with gcc but can be used with icc, llvm, etc.

PR #7917, opened 2020-07-08
Fix buildrpm.sh "-r" option used for RPM options specification

jmciver:
The "-r" option now concatenates using the rpmbuild_options
variable. The "-r" option in prior versions of buildrpm.sh
concatenated using the configure_options variable, which uses special
delineation for Autoconf options (first word of argument string is
"configure_options"). This resulted in an RPM build failure as the
Autoconf options would contain nested RPM option statements.

Signed-off-by: John K. McIver III <john.mciver.iii@gmail.com>


bwbarrett:
ok to test

jmciver:
Hello Brian,

I am new to the OpenMPI project. Is there anything I need to do on my end at this point?

Thanks!
John

Issue #7918, opened 2020-07-08
Problem with MPI_Status conversion when C has 4-byte ints and Fortran has 8-byte INTEGERs


ahaichen:
### Open MPI version: 4.0.1 
- installed from source
- build with: ./configure F77=gfortran FC=gfortran  FFLAGS=-fdefault-integer-8 FCFLAGS=-fdefault-integer-8 
- Operating system: Linux version 4.9.0-12-amd64 (debian-kernel@lists.debian.org)
- CPU: Intel(R) Xeon(R)

Hi all, 

I get an issue when running in 64-bit MPI with the rank initialized 0, the code is 

```fortran
Program test
USE MPI_F08
IMPLICIT NONE

TYPE(MPI_Status)                            :: MPI_stat
Integer(kind=MPI_INTEGER_KIND) :: MPI_err
Integer(kind=MPI_INTEGER_KIND) :: MPI_id=0
Integer(kind=MPI_INTEGER_KIND) :: MPI_np
Integer                                              :: tempint,i_MPI

CALL MPI_Init(MPI_err)
CALL MPI_Comm_rank(MPI_COMM_WORLD,MPI_id,MPI_err)
CALL MPI_Comm_size(MPI_COMM_WORLD,MPI_np,MPI_err)

if(MPI_id==0) tempint=1
if(MPI_id/=0)  tempint=-1

write(*,*) 'MPI_np check1: MPI_id=',MPI_id,'MPI_np=',MPI_np,'tempint',tempint
if(MPI_id/=0) call MPI_Send(tempint,1,MPI_integer8,0,MPI_id,MPI_COMM_WORLD,MPI_err)
DO i_MPI=1,MPI_np-1
  if(MPI_id==0) call MPI_Recv(tempint,1,MPI_integer8,i_MPI,i_MPI,MPI_COMM_WORLD,MPI_stat,MPI_err)
ENDDO
write(*,*) 'MPI_np check2: MPI_id=',MPI_id,'MPI_np=',MPI_np,'tempint',tempint

CALL MPI_Finalize(MPI_err)
End program test
```

It is compiled with 

mpifort -fdefault-integer-8 test.f90

run with 

mpirun -np 3 ./a.out

the output for 3 threads will be 

```
 MPI_np check1: MPI_id=                    0 MPI_np=                    3 tempint                    1
 MPI_np check1: MPI_id=                    2 MPI_np=                    3 tempint                   -1
[...]

PR #7921, opened 2020-07-09
mpi_f08: fix Fortran-8-byte-INTEGER vs. C-4-byte-int issue

jsquyres:
It is important to have the mpi_f08 Type(MPI_Status) be the same
length (in bytes) as the mpif.h status (which is an array of
MPI_STATUS_SIZE INTEGERs).  The reason is because MPI_Status_ctof()
basically does the following:

```
MPI_Fint *f_status = ...;
int *s = (int*) &c_status;
for i=0..sizeof(MPI_Status)/sizeof(int)
   f_status[i] = c_status[i];
```

Meaning: the Fortran status needs to be able to hold as many INTEGERs
are there are C int's that can fit in sizeof(MPI_Status) bytes.

This is because a Fortran INTEGER may be larger than a C int (e.g.,
Fortran 8 bytes vs. C 4 bytes).  Hence, the assignment on the Fortran
side will take sizeof(INTEGER) bytes for each sizeof(int) bytes in the
C MPI_Status.

This commit pads out the mpi_f08 Type(MPI_Status) with enough INTEGERs
to make it the same size as an array of MPI_TYPE_SIZE INTEGERs.
Hence, MPI_Status_ctof() will work properly, regardless of whether it
is assinging to an mpi_f08 Type(MPI_Status) or an mpif.h array of
MPI_STATUS_SIZE INTEGERs.

Refs #7918

### DIFF
--- a/config/ompi_setup_mpi_fortran.m4
+++ b/config/ompi_setup_mpi_fortran.m4
@@ -244,12 +244,28 @@ AC_DEFUN([OMPI_SETUP_MPI_FORTRAN],[
 
     # How big should MPI_STATUS_SIZE be?  (i.e., the size of
     # MPI_STATUS, expressed in units of Fortran INTEGERs).  The C
-    # equivalent of MPI_Status contains 4 C ints and a size_t.
+    # MPI_Status struct contains 4 C ints and a size_t.
     OMPI_FORTRAN_STATUS_SIZE=0
-    AC_MSG_CHECKING([for the value of MPI_STATUS_SIZE])
+
+    # Calculate how many C int's can fit in sizeof(MPI_Status).  Yes,
+    # I do mean C ints -- not Fortran INTEGERS.  The reason is because
+    # an mpif.h MPI_Status is an array of INTEGERS.  But these
+    # sizeof(INTEGER) may be larger than sizeof(int).  Hence,
+    # MPI_Status_ctof() basically does this:
+    #
+    # MPI_Fint *f_status = ...;
+    # int *s = (int*) &c_status;
+    # for i=0..sizeof(MPI_Status)/sizeof(int)
+    #    f_status[i] = c_status[i];
+    #
+    # Meaning: we have to have as many Fortran INTEGERs in the array
+    # as int's will fit in a C MPI_Status (vs. just having a Fortran
+    # array of INTEGERs that has enough bytes to hold a C MPI_Status).
     bytes=`expr 4 \* $ac_cv_sizeof_int + $ac_cv_sizeof_size_t`
-    num_integers=`expr $bytes / $ac_cv_sizeof_int`
-    sanity=`expr $num_integers \* $ac_cv_sizeof_int`
+    AC_MSG_NOTICE([C MPI_Status is $bytes bytes long])
+    AC_MSG_CHECKING([for the value of MPI_STATUS_SIZE])
+    num_ints=`expr $bytes / $ac_cv_sizeof_int`
+    sanity=`expr $num_ints \* $ac_cv_sizeof_int`
     AS_IF([test "$sanity" != "$bytes"],
           [AC_MSG_RESULT([unknown!])
            AC_MSG_WARN([WARNING: Size of C int: $ac_cv_sizeof_int])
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@jsquyres on top of my head
 - have you checked this new definition of `type(MPI_Status)` is valid w.r.t. the MPI standard?
 - this PR will break the ABI, right?
 - my first intent was to preserve the current definition of `type(MPI_Status)` (and do not break the ABI), and update the `use-mpi-f08` subroutines to something like this

```
subroutine MPI_Recv_f08(buf,count,datatype,source,tag,comm,status,ierror)
   use :: mpi_f08_types, only : MPI_Datatype, MPI_Comm, MPI_Status
   use :: ompi_mpifh_bindings, only : empi_recv_f
   implicit none
   OMPI_FORTRAN_IGNORE_TKR_TYPE :: buf
   INTEGER, INTENT(IN) :: count, source, tag
   TYPE(MPI_Datatype), INTENT(IN) :: datatype
   TYPE(MPI_Comm), INTENT(IN) :: comm
   TYPE(MPI_Status), INTENT(OUT) :: status
   INTEGER :: f_status(MPI_STATUS_SIZE)
   INTEGER, OPTIONAL, INTENT(OUT) :: ierror
   integer :: c_ierror

   call ompi_recv_f(buf,count,datatype%MPI_VAL,source,tag,comm%MPI_VAL,f_status,c_ierror)
   if (c_ierror.EQ.MPI_SUCCESS .AND. status.NE.MPI_STATUS_IGNORE) call PMPI_Status_ftof08(f_status, status, c_ierror)
   if (present(ierror)) ierror = c_ierror

end subroutine MPI_Recv_f08
```

note we could/should use macros to skip the `PMPI_Status_ftof08()` invocation when the Fortran and Fortran2008 statuses have the same size.

jsquyres:
> @jsquyres on top of my head
> 
> * have you checked this new definition of `type(MPI_Status)` is valid w.r.t. the MPI standard?

Yes, it is.  The 2 members that I removed were private members, anyway.  I think we added them here in OMPI just as a matter of course (i.e., parity with the C struct).  But this doesn't take into account the "Integers larger than ints" case.

> * this PR will break the ABI, right?

Yes.  ☹️ 

Perhaps we should do this simple solution on master only, and do something like your proposal on the v4.x branches...?

> * my first intent was to preserve the current definition of `type(MPI_Status)` (and do not break the ABI), and update the `use-mpi-f08` subroutines to something like this
> 
> ```
> subroutine MPI_Recv_f08(buf,count,datatype,source,tag,comm,status,ierror)
>    use :: mpi_f08_types, only : MPI_Datatype, MPI_Comm, MPI_Status
>    use :: ompi_mpifh_bindings, only : empi_recv_f
>    implicit none
>    OMPI_FORTRAN_IGNORE_TKR_TYPE :: buf
>    INTEGER, INTENT(IN) :: count, source, tag
>    TYPE(MPI_Datatype), INTENT(IN) :: datatype
>    TYPE(MPI_Comm), INTENT(IN) :: comm
>    TYPE(MPI_Status), INTENT(OUT) :: status
>    INTEGER :: f_status(MPI_STATUS_SIZE)
>    INTEGER, OPTIONAL, INTENT(OUT) :: ierror
>    integer :: c_ierror
> 
>    call ompi_recv_f(buf,count,datatype%MPI_VAL,source,tag,comm%MPI_VAL,f_status,c_ierror)
>    if (c_ierror.EQ.MPI_SUCCESS .AND. status.NE.MPI_STATUS_IGNORE) call PMPI_Status_ftof08(f_status, status, c_ierror)
>    if (present(ierror)) ierror = c_ierror
> 
> end subroutine MPI_Recv_f08
> ```

Ugh, yes, this is probably a good solution for the release branches.  This will touch a LOT of code, though (i.e., everywhere we use a status).

...alternatively, we could state that this is known not to work on the v4.x release.  We could add a configure check for INTEGER=8 bytes + int=4 bytes and refuse to build (i.e., abort configure).  If this has never worked in the v4.x series, then this is technically not a regression -- we would just be making the failure occur sooner (i.e., during configure).

> note we could/should use macros to skip the `PMPI_Status_ftof08()` invocation when the Fortran and Fortran2008 statuses have the same size.

If we actually fix this in v4.x, 👍

jsquyres:
bot:aws:retest

ggouaillardet:
that looks fair to me
 - merge this PR in `master`
 - do not build the `use mpi_f08` bindings in the release branches if the `INTEGER` size is larger than 32 bits.

jsquyres:
I updated the commit message to indicate that this is an ABI break.

@ggouaillardet Can you review so that I can merge?

(we keep getting weird SSL failures in AWS CI... trying again...)

bot:aws:retest

jsquyres:
bot:aws:retest

PR #7923, opened 2020-07-10
v4.1.x: fortran.m4: disallow when sizeof(int) != sizeof(INTEGER)

jsquyres:
NOTE: This is intentionally not a cherry pick from master.  Instead,
this is a cherry-pick from the equivalent commit on the v4.0.x branch.
See below.

There is a problem with the mpi_f08 module when sizeof(int) !=
sizeof(INTEGER): the size of TYPE(MPI_Status) is too small.  This
causes buffer overruns when Open MPI is configured with (for example)
sizeof(int)==4 and sizeof(INTEGER)==8, and then you call the mpi_f08
MPI_RECV subroutine.  This will end up copying the resulting C
MPI_Status to the buffer pointing to the Fortran status, but the code
does not know if the Fortran status is an mpif.h status or a
TYPE(MPI_Status) -- it just blindly copies over as if the Fortran
status is an INTEGER array of length MPI_STATUS_SIZE.  Unfortunately,
TYPE(MPI_Status) is actually smaller than this, so we overrun the
buffer.  Hilarity ensues.

The simple fix for this is to make TYPE(MPI_Status) the same size as
INTEGER(MPI_STATUS_SIZE), but we can't do that here on the release
branch because it will break ABI.

- checks to see if we're in a sizeof(int) != sizeof(INTEGER) scenario
- if so, if the user has not specifically excluded building the
  mpi_f08 module, display a Giant Error Message (GEM) and abort
  configure.

This is unusual; we don't usually abort configure when feature XYZ
can't be built -- if the user didn't specifically ask for XYZ, we
just emit a notice that we won't build XYZ and continue.

This situation is a little different because we're on a release
branch: prior releases have built mpi_f08 by default -- even in this
"bad" scenario.  Hence, in this case, we explicitly tell the user that
this is now a known-bad scenario and abort.  In the GEM, we give the
user two options:

1. Change their compiler flags so that sizeof(int) == sizeof(INTEGER)
   and re-run configure, or
2. Explicitly disable the mpi_f08 module via --enable-mpi-fortran=usempi

Note: the proper fix has been implemented on master (i.e., what will
become v5.0.0), but since that breaks ABI, we can't cherry pick it
back here to an existing release branch series. Hence, we
cherry-picked this fix from the v4.0.x branch.

(cherry picked from commit 27836a614b9c29d7636cdf1a9b838b1532281a8a)

Refs #7918

### DIFF
--- a/config/ompi_setup_mpi_fortran.m4
+++ b/config/ompi_setup_mpi_fortran.m4
@@ -373,6 +373,72 @@ AC_DEFUN([OMPI_SETUP_MPI_FORTRAN],[
            OMPI_FORTRAN_F08_TYPE=$OMPI_FORTRAN_IGNORE_TKR_TYPE
           ])
 
+    # The mpi_f08 module in this version of Open MPI does not properly
+    # handle if sizeof(INTEGER) != sizeof(int) with the mpi_f08
+    # bindings.  As of July 2020, this issue is fixed on master / what
+    # will eventually become Open MPI v5.0.x, but the fix causes an
+    # ABI break.  Hence, we're not going to fix it here on this
+    # release branch.
+    #
+    # This is a bit of a quandry, however, because up until v4.0.4, we
+    # built the Open MPI F08 bindings by default (if the compiler
+    # supported them), even in this problematic scenario.  Meaning: a
+    # user could use the exact same compilers + configure command line
+    # between v4.0.4 and v4.0.5 and suddenly -- probably unexpectedly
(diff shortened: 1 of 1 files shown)


jsquyres:
Do not merge until the discussion on the corresponding v4.0.x PR #7922 (from where this PR was cherry picked) is complete.

jsquyres:
Strange AWS CI error.

bot:aws:retest

PR #7924, opened 2020-07-10
mpi/c: fix param checks in [I]Neighbor_alltoall{v,w}

hkuno:
do not check some input parameters when an {in,out}degree is zero

Thanks Junchao Zhang for analyzing and reporting this issue.

(cherry picked from commit 5655d64bd3e064cdd6925b994e555e75d00d0f08)


jsquyres:
@hkuno Don't forget to set a target label and a milestone; otherwise, PRs on release branches like this may get missed.

Issue #7928, opened 2020-07-13
[btl/smcuda] OMPI build fails when cuda is enabled 


AboorvaDevarajan:
### OMPI version

master (git clone)

### If you are building/installing from a git clone, please copy-n-paste the output from `git submodule status`.

```
 4a43c39c89037f52b4e25927e58caf08f3707c33 opal/mca/hwloc/hwloc2/hwloc (hwloc-2.1.0rc2-53-g4a43c39)
 b18f8ff3f62bd93f17e01c4230e7eca4a30aa204 opal/mca/pmix/pmix4x/openpmix (v1.1.3-2460-gb18f8ff)
 17ace07e7d41f50d035dc42dfd6233f8802e4405 prrte (dev-30660-g17ace07)
```
-----------------------------

## Details of the problem

OMPI CUDA enabled build fails,

```
./autogen.pl 
./configure --prefix=/nfs_smpi_ci/abd/pr/ompi-install --with-cuda=/usr/local/cuda
make -j 20
```

**Error:**

```
make[2]: Entering directory `/nfs_smpi_ci/abd/pr/ompi/opal/mca/btl/smcuda'
  CC       mca_btl_smcuda_la-btl_smcuda_component.lo
btl_smcuda_component.c: In function ‘mca_btl_smcuda_send_cuda_ipc_ack’:
btl_smcuda_component.c:632:9: warning: variable ‘rc’ set but not used [-Wunused-but-set-variable]
     int rc;
         ^
btl_smcuda_component.c: In function ‘mca_btl_smcuda_component_init’:
btl_smcuda_component.c:938:68: warning: assignment from incompatible pointer type [enabled by default]
     mca_btl_base_active_message_trigger[MCA_BTL_TAG_SMCUDA].cbfunc = btl_smcuda_control;
                                                                    ^
btl_smcuda_component.c: In function ‘mca_btl_smcuda_component_progress’:
btl_smcuda_component.c:1071:29: warning: passing argument 2 of ‘reg->cbfunc’ makes pointer from integer without a cast [enabled by default]
                             reg->cbdata);
                             ^
btl_smcuda_component.c:1071:29: note: expected ‘const struct mca_btl_base_receive_descriptor_t *’ but argument is of type ‘mca_btl_base_tag_t’
btl_smcuda_component.c:1071:29: error: too many arguments to function ‘reg->cbfunc’
make[2]: *** [mca_btl_smcuda_la-btl_smcuda_component.lo] Error 1
[...]

PR #7931, opened 2020-07-13
Fix the BTL API conversion for the SMCUDA BTL

bosilca:
This is a first step, there are still changes to be made in the OB1 PML.


hjelmn:
Maybe we can add an additional context pointer to the receive struct so smcuda can store the frag pointer.

bosilca:
I need to refresh my memory on how this works, it is possible we don't need it.

bosilca:
Short story the smcuda BTL has a bunch of crippled and untested capabilities. I disabled some of it, that I found unreliable. I think this BTL only works in a pt2pt scenario, with the send operation. Not sure it is worth saving.

gpaulsen:
I believe this PR is intended to address the issue raised in https://github.com/open-mpi/ompi/issues/7928

hjelmn:
Probably should add an no build to it for now. Then commit the ob1 fixes. If Nvida wants to fix it they can do so. I would prefer CUDA support in sm.

bosilca:
It is currently working with these changes for all the pt2pt tests I run on it (not many support GPU memory thou). At this point I am mostly concerned about the HWLOC changes.

PR #7933, opened 2020-07-13
v4.1.x: schizo/slurm: Fix binding detection

artpol84:
Signed-off-by: Artem Polyakov <artpol84@gmail.com>
(cherry picked from commit c72f295dfaced7e9f879b1eba3eebb7c35cb5bf5)

Fixes #7932

PR #7934, opened 2020-07-13
v4.1.x: schizo/jsm: Disable binding when direct launched

jjhursey:


jjhursey:
The `schizo/jsm` component only exists on the v4.1.x branch.

After reviewing PR #7933 it looks like we should do the same for the JSM component.

jsquyres:
Per this:

> The `schizo/jsm` component only exists on the v4.1.x branch.

@gpaulsen Why did you approve?

gpaulsen:
My understanding is that this _IS_ the work on the jsm schizo.  I think he just meant that we need the same change that slurm needs, and this is it.

Also this is only needed on `v4.1.x`.  `v4.0.x` never accepted the `schizo:jsm` component (new feature after `v4.0.0`), and it's not needed on master as PMIx handles these things (or so I'm told).

@jjhursey can you please confirm?

rhc54:
FWIW: neither this nor the Slurm PR are actually correct. There is more going on here than we originally realized. I'll try to provide a broader PR that addresses all the systems.

jjhursey:
A few notes:
 * `schizo/jsm` is only on the v4.1.x branch. It was not accepted to v4.0.x (it was late and a new feature). `master` does not need these kind of shims anymore.
 * The _only_ thing that we need the schizo component to do is detect that it was direct launched with `jsrun` and acknowledge that OMPI is talking via PMIx to the JSM resource manager.
 * We did not explicitly do anything with binding originally, but looking at the slurm version it seemed reasonable that we would need the same patch to prevent OMPI from doing any binding. But if @rhc54 thinks that that patch isn't sufficient then we should wait until we have the proper fix and merge that.

PR #7942, opened 2020-07-14
Ensure we init and protect values

rhc54:
Scrub the entire ompi_rte.c file to initialize and protect values
received from PMIx.

Closes #7939 


rhc54:
Closes #7939 

@abouteiller Let's fix all the initializations and protections - I agree your PR fixes the one spot, but there are other places with the same/similar problems.

PR #7943, opened 2020-07-14
v4.1.x: common/ofi: added address format check to fix provider selection

dancejic:
bugfix: provider selection would not differentiate between ipv4
and ipv6 addresses which would cause some nodes to be unable
to communicate between each other. Adding a check for address
format to provider selection to ensure that all nodes use the
same address format.

(cherry picked from commit 7e463713014ae58f7e78d7a5b49e9e63d62e374e)


bwbarrett:
Please see step 8 of https://github.com/open-mpi/ompi/wiki/SubmittingPullRequests.  You need to include the cherry-pick from line generated by "cherry-pick -x".

jsquyres:
@bwbarrett Can you review this PR?

PR #7948, opened 2020-07-16
osc/rdma: fail query_btls if no endpoint for non-local peer is found

devreal:
This PR makes sure that the osc/rdma btl querying logic explicitly errors out if it finds a non-local peer that has no valid endpoint.

The underlying issue has manifested itself in a crash in `MPI_Win_lock_all` because the module's leader was `NULL`. This happened if a) running with more than one rank per node; b) the window was allocated with a sub-communicator with 2 ranks on 2 nodes; and c) no valid btl was available other than `tcp` and `vader`. In that case, the selection logic assumed that the `MCA_BTL_ATOMIC_SUPPORTS_GLOB` flag of `vader` qualified it for use in this case even though it doesn't (the check before the increment in https://github.com/open-mpi/ompi/blob/master/ompi/mca/osc/rdma/osc_rdma_component.c#L932 is not specific enough). With this PR, the check is still there (in case a BTL does not support local access) but the selection fails early if no suitable btl is found for any non-local peer.

The issue is also present in 4.0.x and 4.1.x and I will backport this patch once it is approved.

### DIFF
--- a/ompi/mca/osc/rdma/osc_rdma_component.c
+++ b/ompi/mca/osc/rdma/osc_rdma_component.c
@@ -813,6 +813,7 @@ static int ompi_osc_rdma_query_btls (ompi_communicator_t *comm, struct mca_btl_b
 {
     struct mca_btl_base_module_t **possible_btls = NULL;
     int comm_size = ompi_comm_size (comm);
+    int comm_rank = ompi_comm_rank (comm);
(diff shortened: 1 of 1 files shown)


hjelmn:
I will be reworking this code shortly to work with a combination of tcp/vader. Should be ready in the next couple of weeks but will not be going back to 4.x.

PR #7953, opened 2020-07-21
Fix memory leak in configure, which prevents leak sanitizer usage

cniethammer:
Signed-off-by: Christoph Niethammer <niethammer@hlrs.de>

### DIFF
--- a/config/c_get_alignment.m4
+++ b/config/c_get_alignment.m4
@@ -44,6 +44,7 @@ AC_DEFUN([OPAL_C_GET_ALIGNMENT],[
     FILE *f=fopen("conftestval", "w");
     if (!f) exit(1);
     diff = ((char *)&p->x) - ((char *)&p->c);
+    free(p);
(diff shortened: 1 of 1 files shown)


jsquyres:
@cniethammer I'm not entirely understanding what problem it is that you're fixing here.  Yes, the `free()` was missing from the configure program, but why does it matter?

cniethammer:
This matters if I build Open MPI with sanitizers, e.g. `configure CC=clang CFLAGS=-fsanitize=address ...`.
In this case configure test programs area also build with the sanitizers and will report errors resulting in configure to fail.

jsquyres:
Got it.  Can you put that in the commit message and/or in a comment in the code?

cniethammer:
@jsquyres Updated the commit message

devreal:
Should this be backported to the release branches?

cniethammer:
Yes, would be good to have this in at least 4.1.x for testing

jsquyres:
Might as well do v4.0.x, too.

PR #7956, opened 2020-07-21
v4.1.x: BTL/OFI: Fix missing include file.

tomhers:
The missing include file causes an error when using an external version of LibEvent.

Signed-off-by: tomhers <tom.herschberg@gmail.com>
(cherry picked from commit 88f9d2c90f1730f2e0b5bc4951893f60ff5a1332)


jsquyres:
ok to test

Issue #7963, opened 2020-07-24
How to query whether CUDA awareness is actually turned on or not at runtime?


leofang:
Hi, I am aware of the build time and runtime checks outlined in https://www.open-mpi.org/faq/?category=runcuda#mpi-cuda-aware-support, so please bear with me til the end.

We are looking for a runtime way to check whether CUDA awareness is actually turned on in Open MPI, see [the original discussion](https://bitbucket.org/mpi4py/mpi4py/pull-requests/18) in the mpi4py repo. It turns out that the existing API `MPIX_Query_cuda_support()` is useless. I quote @jsquyres: 
> Honestly, I don’t think we thought anyone was using it.  🙂

The reason is it only tells us if `--with-cuda` is set at build time, but the smcuda btl could still be ejected (as defaulted in conda-forge's `openmpi` package: https://github.com/conda-forge/openmpi-feedstock/pull/56) and thus no CUDA awareness. 

The situation becomes even more complicated when UCX is in use. As pointed out by @jsquyres, Open MPI could be built without CUDA while UCX is, and in this situation we still get CUDA awareness (but `MPIX_Query_cuda_support()` would return `false`)! For this, I am requesting the UCX side to support a runtime query (https://github.com/openucx/ucx/issues/5471), which should be incorporated here if UCX is in use by Open MPI.

In short, it would be great if Open MPI could set up a mechanism (hopefully another public API, but leave the existing `MPIX_Query_cuda_support()` intact to avoid confusion) for us to query, at runtime:
- If Open MPI's CUDA support is actually in effect or not, this needs to take into account
  1. When Open MPI is not built against UCX: whether the smcuda btl is ejected or not
  2. When UCX is used: whether UCX has CUDA support (https://github.com/openucx/ucx/issues/5471)

I hope I do not misunderstand the situation or miss any critical pieces of information. Thanks.

cc: @dalcinl

leofang:
I didn't expect I could follow up here so quickly 😁 @yosefe and @bureddy kindly pointed out in https://github.com/openucx/ucx/issues/5471 that `ucp_context_query()` can be used to query the CUDA support from UCX, and that Open MPI already uses this API during initialization (#7898). I think we can just take the information recorded there when UCX is in use, so half of the problem is resolved!

bureddy:
[...]

PR #7970, opened 2020-07-28
CUDA: Enhance MPIX_Query_cuda_support() with runtime support check 

bureddy:
fixes: https://github.com/open-mpi/ompi/issues/7963 & https://github.com/openucx/ucx/issues/5471#

@jsquyres @bosilca @yosefe @Akshay-Venkatesh

### DIFF
--- a/ompi/mpiext/cuda/c/MPIX_Query_cuda_support.3in
+++ b/ompi/mpiext/cuda/c/MPIX_Query_cuda_support.3in
@@ -24,8 +24,33 @@ There is no C++ binding for this function.
 .SH DESCRIPTION
 .ft R
 
+This routine return 1 if MPI library is build with CUDA and runtime supports CUDA buffers.
+This routine must be called after MPI is initialized by a call to MPI_Init or MPI_Init_thread.
+
 .SH Examples
 .ft R
+.nf
+
+#include <stdio.h>
+#include "mpi.h"
+
+#include "mpi-ext.h" /* Needed for CUDA-aware check */
+
+int main(int argc, char *argv[])
+{
+
+    MPI_Init(&argc, &argv);
+
+    if (MPIX_Query_cuda_support()) {
+        printf("This MPI library has CUDA-aware support.\n");
+    } else {
+        printf("This MPI library does not have CUDA-aware support.\n");
+    }
+    MPI_Finalize();
+
+    return 0;
+}
+.fi
 
 .SH See Also
 .ft R
--- a/ompi/mpiext/cuda/c/mpiext_cuda.c
+++ b/ompi/mpiext/cuda/c/mpiext_cuda.c
@@ -21,11 +21,12 @@
 #include <string.h>
 
 #include "opal/constants.h"
+#include "opal/runtime/opal_params.h"
 #include "ompi/mpiext/cuda/c/mpiext_cuda_c.h"
 
 /* If CUDA-aware support is configured in, return 1. Otherwise, return 0.
  * This API may be extended to return more features in the future. */
 int MPIX_Query_cuda_support(void)
 {
-    return OPAL_CUDA_SUPPORT;
+    return opal_built_with_cuda_support && opal_cuda_runtime_initialized;
 }
--- a/opal/mca/common/cuda/common_cuda.c
+++ b/opal/mca/common/cuda/common_cuda.c
@@ -481,6 +481,7 @@ int mca_common_cuda_stage_one_init(void)
 #if OPAL_CUDA_GET_ATTRIBUTES
     OPAL_CUDA_DLSYM(libcuda_handle, cuPointerGetAttributes);
 #endif /* OPAL_CUDA_GET_ATTRIBUTES */
+    opal_cuda_runtime_initialized = true;
     return 0;
 }
 
--- a/opal/runtime/opal_params.c
+++ b/opal/runtime/opal_params.c
@@ -62,6 +62,7 @@ bool opal_timing_overhead = true;
 #endif
 
 bool opal_built_with_cuda_support = OPAL_INT_TO_BOOL(OPAL_CUDA_SUPPORT);
+bool opal_cuda_runtime_initialized = false;
 bool opal_cuda_support = false;
 bool opal_warn_on_missing_libcuda = true;
 #if OPAL_ENABLE_FT_CR == 1
--- a/opal/runtime/opal_params.h
+++ b/opal/runtime/opal_params.h
@@ -49,6 +49,11 @@ OPAL_DECLSPEC extern bool opal_built_with_cuda_support;
 OPAL_DECLSPEC extern bool opal_cuda_support;
 
 /**
+ * Whether cuda runtime support is initialized or not.
+ */
+OPAL_DECLSPEC extern bool opal_cuda_runtime_initialized;
+
+/**
  *  * Whether we want to warn the user when libcuda is missing.
  *   */
 OPAL_DECLSPEC extern bool opal_warn_on_missing_libcuda;


bureddy:
example test:
```
#include <stdio.h>
#include "mpi.h"

#include "mpi-ext.h" /* Needed for CUDA-aware check */

{

    MPI_Init(&argc, &argv);

    if (MPIX_Query_cuda_support()) {
        printf("This MPI library has CUDA-aware support.\n");
    } else {
        printf("This MPI library does not have CUDA-aware support.\n");
    }
    MPI_Finalize();

    return 0;
}
```

UCX with cuda
```
mpirun -np 2 --mca pml ucx   ./a.out
This MPI library has CUDA-aware support.
This MPI library has CUDA-aware support.
```

UCX w/o cuda TLs
```
mpirun -np 2 --mca pml ucx  -x UCX_TLS=rc ./a.out
This MPI library does not have CUDA-aware support.
This MPI library does not have CUDA-aware support.

```

bosilca:
@bureddy what is happening is you do not specify "--mca pml ucx" ? The flag you added shows when/if the OMPI side of the CUDA infrastructure has been initialized, and you made the assumption that this means the underlying communication infrastructure is able to seamlessly handle CUDA buffers. Is this always the case ?

leofang:
Thanks for quickly putting together a solution and for tagging me, @bureddy. Forgive my tons of questions as I am not very familiar with Open MPI's internal:
1. What's the expected outcome for the following matrix (looks like 8 combinations, but really only 6 is meaningful):
   - Open MPI built without and with `--with-cuda`
   - Open MPI built without and with `--with-ucx`
   - UCX built without and with `--with-cuda`
2. I was under the impression from openucx/ucx#5471 that we must call `ucp_context_query()` to ensure CUDA support, is this already done somewhere not included in this PR?
3. What happens when we call this prior to calling `MPI_Init()`? Does it fail like `MPI_Query_thread()` as it should?
4. Can we please not change the semantics of existing API and create a new function for this query? This would break backward compatibility...

Tagging MVAPICH2 people @chingking to comment, as they recently followed Open MPI and supported the same API in v2.3.4.

bureddy:
> @bureddy what is happening is you do not specify "--mca pml ucx" ? The flag you added shows when/if the OMPI side of the CUDA infrastructure has been initialized, and you made the assumption that this means the underlying communication infrastructure is able to seamlessly handle CUDA buffers. Is this always the case ?

@bosilca added more test results above with  ob1 pml.   OMPI side runtime cuda Intra is initialized(`mca_common_cuda_stage_one_init()` called) only when underlying communication (UCX, TCP, smcuda) can handles the CUDA buffers.

bureddy:
> Thanks for quickly putting together a solution and for tagging me, @bureddy. Forgive my tons of questions as I am not very familiar with Open MPI's internal:
> 
> 1. What's the expected outcome for the following matrix (looks like 8 combinations, but really only 6 is meaningful):
>    
>    * Open MPI built without and with `--with-cuda`
>    * Open MPI built without and with `--with-ucx`
>    * UCX built without and with `--with-cuda`
```

OMPI build   --with-cuda | ompi build --with-ucx | ucx with --with-cuda | MPIX_Query_cuda_support
-- ----------------------| ----------------------| -------------------- | ------------------------------
                    0    |          x            |             x        |      0
                    0    |         x             |             x        |      0
                   1     |         0             |           N/A        |      1 - with tcp,smcuda,    0 -  other wise
                   1     |        1              |            0         |      0
                   1     |        1              |            1         |      1

```

  
            
> 2. I was under the impression from [openucx/ucx#5471](https://github.com/openucx/ucx/issues/5471) that we must call `ucp_context_query()` to ensure CUDA support, is this already done somewhere not included in this PR?

Yes. with UCX, OMPI side runtime support is called only after `ucp_context_query()` and veryfied that it support cuda buffer

> 3. What happens when we call this prior to calling `MPI_Init()`? Does it fail like `MPI_Query_thread()` as it should?

Returns False because the runtime is not initialized.

> 4. Can we please not change the semantics of existing API and instead create a new function for this query? Otherwise, it would break backward compatibility...
> 

I'm not sure it is worth complicating the things given that it is not standard API and we sort of fixing the semantics in a correct way. 

> Tagging MVAPICH2 people @chingking to comment, as they recently followed Open MPI and supported the same API in v2.3.4.

leofang:
@bureddy Thanks. Please confirm the below table, in particular the 4th and 7th row.
 ompi --with-cuda | ompi --with-ucx | ucx --with-cuda | output | comment
------------------ | ---------------- | ----------------- | -------| ---------
0 | 0 | 0 | 0 | (trivial, no one has cuda)
0 | 0 | 1 | 0 | (trivial, ucx has cuda but unused)
0 | 1 | 0 | 0 | (trivial, no one has cuda)
0 | 1 | 1 | 0 or 1 | 1: ucx; 0: otherwise
1 | 0 | 0 | 0 or 1 | 1: tcp, smcuda; 0: otherwise
1 | 0 | 1 |  0 or 1 | 1: tcp, smcuda; 0: otherwise
1 | 1 | 0 | 0 or 1 | 1: tcp, smcuda; 0: otherwise
1 | 1 | 1 | 1 | tcp, smcuda, ucx, etc

leofang:
> I'm not sure it is worth complicating the things given that it is not standard API and we sort of fixing the semantics in a correct way.

Well, the whole CUDA thing is not in the MPI standard, but we do it anyway 🙂

Perhaps @dalcinl has a different opinion. I imagine it's good to separate the two pieces of information, quoting from https://github.com/open-mpi/ompi/issues/7963#issuecomment-663746350:
> 1. Is the library compiled with CUDA support.
> 2. Is CUDA support enabled "right now" (for some definition of "right now").

In addition to being able to maintaining the same semantics, it also allows us to hint the users. For example, it could be that Open MPI/UCX is built with CUDA support, but needs to be turned on or configured at launch time. If the runtime query returns 0 but the compile time query returns 1, at least we can raise with a nice message like "the MPI library is built with CUDA support but is not enabled, please consult the user manual of your MPI library".

bureddy:
> 0	1	1	0 or 1

4th, it will be 0. because OMPI not compiled with CUDA.  you might get limited support UCX/CUDA for pt2pt.

bureddy:
> > I'm not sure it is worth complicating the things given that it is not standard API and we sort of fixing the semantics in a correct way.
> 
> Well, the whole CUDA thing is not in the MPI standard, but we do it anyway 🙂
> 
> Perhaps @dalcinl has a different opinion. I imagine it's good to separate the two pieces of information, quoting from [#7963 (comment)](https://github.com/open-mpi/ompi/issues/7963#issuecomment-663746350):
> 
> > 1. Is the library compiled with CUDA support.
> > 2. Is CUDA support enabled "right now" (for some definition of "right now").
> 
> In addition to being able to maintaining the same semantics, it also allows us to hint the users. For example, it could be that Open MPI/UCX is built with CUDA support, but needs to be turned on or configured at launch time. If the runtime query returns 0 but the compile time query returns 1, at least we can raise with a nice message like "the MPI library is built with CUDA support but is not enabled, please consult the user manual of your MPI library".

is Open MPI/UCX is built with CUDA support, you don't need to turn on anything at run time to enable cuda support. its default.

chingking:
@leofang - thanks for tagging me. Yes, MVAPICH2 currently follows the same logic as Open MPI. That is, `MPIX_Query_cuda_support` returns 1 as long as the runtime is built with CUDA support (even not enabled at runtime). MVAPICH2 will print other warning messages if the CUDA feature is not properly enabled at runtime.

We are happy to work together and have a unified standard for this support. I personally agree with the enhancement proposed here and we could print a warning message if users call `MPIX_Query_cuda_support` before `MPI_Init`.

Adding the MVAPICH2 core developer @harisubramoni.

leofang:
Thank you @chingking for engaging in this discussion! Would be great to have input from you and @harisubramoni.

bureddy:
> Hi @bureddy Sorry I dropped the ball.
> 
> > 4th, it will be 0. because OMPI not compiled with CUDA. you might get limited support UCX/CUDA for pt2pt.
> 
> OK I would like to drag in @jsquyres for confirming this, because my impression from Jeff was that for this case Open MPI could still be CUDA-aware. @bureddy Are you saying collectives won't work but point-to-point could? Interesting...

Yes.

> 
> > if Open MPI/UCX is built with CUDA support, you don't need to turn on anything at run time to enable cuda support. its default.
> 
> I don't think this is true. In fact this is one of the motivations for the whole thing. In Conda-Forge we built Open MPI with the `--with-cuda` flag, but we shut it down by default by setting `OMPI_MCA_opal_cuda_support=0`. In this case, the compile time flag is 1 but the runtime flag should be 0.

I mean, default is enabled. you can shut it down by OMPI_MCA_opal_cuda_support=0.  With this PR (compile & runtime check), it returns support 0 because you disabled it run time. 

> 
> Equivalently, (without building against ucx) we can launch MPI progresses via `--mca btl ^smcuda ...` I hope these two cases make it clear why we are bothered by the status quo and seeking for improvements 🙂

leofang:
> Yes. it might change in the future when UCX handle all datatypes pack/unpack and collectives

Great, thanks @bureddy! It's good to know. I am looking forward to it. (Is there a time table for this? 😇)

> > I don't think this is true. In fact this is one of the motivations for the whole thing. In Conda-Forge we built Open MPI with the `--with-cuda` flag, but we shut it down by default by setting `OMPI_MCA_opal_cuda_support=0`. In this case, the compile time flag is 1 but the runtime flag should be 0.
> 
> I mean, default is enabled. you can shut it down by OMPI_MCA_opal_cuda_support=0. With this PR (compile & runtime check), it returns support 0 because you disabled it run time.

Right, I also expect with this PR we get 0 for this case. But this is precisely one of the use cases we want to differentiate among all possible scenarios that I mentioned above:
>> In addition to being able to maintaining the same semantics, it also allows us to hint the users. For example, it could be that Open MPI/UCX is built with CUDA support, but needs to be turned on or configured at launch time. If the runtime query returns 0 but the compile time query returns 1, at least we can raise with a nice message like "the MPI library is built with CUDA support but is not enabled, please consult the user manual of your MPI library".

by knowing compile time support 0 and runtime support 1, we can print out nice error messages to the users. 

Actually MVAPICH2 is already kinda doing this: when using MVAPICH2-GDR (cuda support is built) but `MV2_USE_CUDA` is not set, they print out a nice warning. I expect to build a guard similar to that with the inclusion of this PR.

bureddy:
> > Yes. it might change in the future when UCX handle all datatypes pack/unpack and collectives
> 
> Great, thanks @bureddy! It's good to know. I am looking forward to it. (Is there a time table for this? 😇)

These topics are in very early stage of design phase. don't have timelines yet.

> 
> > > I don't think this is true. In fact this is one of the motivations for the whole thing. In Conda-Forge we built Open MPI with the `--with-cuda` flag, but we shut it down by default by setting `OMPI_MCA_opal_cuda_support=0`. In this case, the compile time flag is 1 but the runtime flag should be 0.
> > 
> > 
> > I mean, default is enabled. you can shut it down by OMPI_MCA_opal_cuda_support=0. With this PR (compile & runtime check), it returns support 0 because you disabled it run time.
> 
> Right, I also expect with this PR we get 0 for this case. But this is precisely one of the use cases we want to differentiate among all possible scenarios that I mentioned above:
> 
> > > In addition to being able to maintaining the same semantics, it also allows us to hint the users. For example, it could be that Open MPI/UCX is built with CUDA support, but needs to be turned on or configured at launch time. If the runtime query returns 0 but the compile time query returns 1, at least we can raise with a nice message like "the MPI library is built with CUDA support but is not enabled, please consult the user manual of your MPI library".
> 
> by knowing compile time support 1 and runtime support 0, we can print out nice error messages to the users.
> 
> Actually MVAPICH2 is already kinda doing this: when using MVAPICH2-GDR (cuda support is built) but `MV2_USE_CUDA` is not set, they print out a nice warning, so apparently they have a way to detect the support at runtime. I expect to build a guard similar to that with the inclusion of this PR.
> 
> **(UPDATE)**: I fixed a few typos.

gpaulsen:
@bureddy ?

bureddy:
@jsquyres updated documentation. can you check it?

leofang:
Hi @bureddy @jsquyres Could you confirm this PR would enable us to differentiate the case that the support is built but not enabled at runtime? See our earlier discussion:

> > > I don't think this is true. In fact this is one of the motivations for the whole thing. In Conda-Forge we built Open MPI with the `--with-cuda` flag, but we shut it down by default by setting `OMPI_MCA_opal_cuda_support=0`. In this case, the compile time flag is 1 but the runtime flag should be 0.
> > 
> > 
> > I mean, default is enabled. you can shut it down by OMPI_MCA_opal_cuda_support=0. With this PR (compile & runtime check), it returns support 0 because you disabled it run time.
> 
> Right, I also expect with this PR we get 0 for this case. But this is precisely one of the use cases we want to differentiate among all possible scenarios that I mentioned above:
> 
> > > In addition to being able to maintaining the same semantics, it also allows us to hint the users. For example, it could be that Open MPI/UCX is built with CUDA support, but needs to be turned on or configured at launch time. If the runtime query returns 0 but the compile time query returns 1, at least we can raise with a nice message like "the MPI library is built with CUDA support but is not enabled, please consult the user manual of your MPI library".
> 
> by knowing compile time support 1 and runtime support 0, we can print out nice error messages to the users.

leofang:
@chingking @harisubramoni (MVAPICH2) @dalcinl (mpi4py) @pavanbalaji (MPICH): Any chance you could take a look at this PR? The runtime query for CUDA awareness in Open MPI is being done here. Thanks 🙏

lanl-ompi:

jsquyres:
@bureddy @Akshay-Venkatesh Can you guys answer @leofang's question from 2 Sep?

awlauria:
@bureddy @Akshay-Venkatesh @leofang's re-ping. There is a v5.0.x branch now, so please cherry-pick there if wanted once this is merged to master. @gpaulsen

janjust (approved):

janjust:
@awlauria ready to merge

awlauria:
Thanks @janjust .

Unfortunately I don't think we can merge until that checker runs. @bureddy can you rebase and force push to retrigger it? I don't think there's a manual way to trigger that commit checker.

awlauria:
Merging for @janjust

PR #7973, opened 2020-07-30
btl/ofi: Disable EFA provider in versions earlier than libfabric 1.12.0

wckzhang:
EFA incorrectly implements FI_DELIVERY_COMPLETE in earlier libfabric
versions. While FI_DELIVERY_COMPLETE would be advertised by the
provider, completions would return too early by not accounting for
bounce buffers on the receive side. This would cause the BTL
to receive early completions that lead to correctness issues.

This is not an issue in the mtl/ofi as it does not require
FI_DELIVERY_COMPLETE.

### DIFF
--- a/opal/mca/btl/ofi/btl_ofi_component.c
+++ b/opal/mca/btl/ofi/btl_ofi_component.c
@@ -59,6 +59,17 @@ static int validate_info(struct fi_info *info, uint64_t required_caps)
 
     BTL_VERBOSE(("validating device: %s", info->domain_attr->name));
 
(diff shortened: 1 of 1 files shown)


wckzhang:
Tweaking the logic to avoid any combination of layered providers as well

wckzhang:
Switched to using: !strncasecmp(info->fabric_attr->prov_name, "efa", 3) instead of a direct strcmp, this will also exclude provider combinations such as efa;ofi_rxd

bwbarrett:
Re-adding the WIP label.  We are having an internal discussion about utility provider behaviors.

PR #7975, opened 2020-07-30
btl/ofi: Use common provider include/exclude list

wckzhang:
The btl/ofi does not currently utilize the common ofi include/exclude           
list. Added verification code similar to the mtl/ofi that will check if         
the info object is in the include or exclude list. If it isn't in the           
include list or is in the exclude list, validate_info will return               
OPAL_ERROR. The btl/ofi will no longer pass a provider name as a hint           
when calling getinfo, instead filtering the provider during                     
validate_info.                                                                  
                                                                                
This patch also moves the is_in_list MTL function into common code and          
adds additional debugging output to the BTL to match the MTL standard.          
                                                                                
### DIFF
--- a/opal/mca/btl/ofi/btl_ofi_component.c
+++ b/opal/mca/btl/ofi/btl_ofi_component.c
@@ -53,9 +53,24 @@ static int mca_btl_ofi_init_device(struct fi_info *info);
 
 /* validate information returned from fi_getinfo().
  * return OPAL_ERROR if we dont have what we need. */
-static int validate_info(struct fi_info *info, uint64_t required_caps)
+static int validate_info(struct fi_info *info, uint64_t required_caps,
+                         char **include_list, char **exclude_list)
 {
     int mr_mode;
+
+    if (NULL != include_list && !opal_common_ofi_is_in_list(include_list, info->fabric_attr->prov_name)) {
+        opal_output_verbose(1, opal_common_ofi.output,
+                            "%s:%d: btl:ofi: \"%s\" not in include list\n",
+                            __FILE__, __LINE__,
+                            info->fabric_attr->prov_name);
+        return OPAL_ERROR;
+    } else if (NULL != exclude_list && opal_common_ofi_is_in_list(exclude_list, info->fabric_attr->prov_name)) {
+        opal_output_verbose(1, opal_common_ofi.output,
+                            "%s:%d: btl:ofi: \"%s\" in exclude list\n",
+                            __FILE__, __LINE__,
+                            info->fabric_attr->prov_name);
+        return OPAL_ERROR;
+    }
 
     BTL_VERBOSE(("validating device: %s", info->domain_attr->name));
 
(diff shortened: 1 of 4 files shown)


opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -51,12 +51,47 @@ static char *ofi_progress_mode;
...
 {
     int mr_mode;
 
+    if (NULL != include_list && !is_in_list(include_list, info->fabric_attr->prov_name)) {
+        opal_output_verbose(1, opal_common_ofi.output,
+                            "%s:%d: btl:ofi: \"%s\" not in include list\n",
+                            __FILE__, __LINE__,
+                            info->fabric_attr->prov_name);
+        return OPAL_ERROR;
+    } else if (NULL != exclude_list && is_in_list(exclude_list, info->fabric_attr->prov_name)) {
```
hkuno, line 87: Where does info->fabric_attr->prov_name get set now? I saw that it's no longer set directly in mca_btl_ofi_component_init . Does it get set somewhere else?

hppritcha: Looks like the OFI BTL doesn't use that attribution in the info arg to fi_getinfo with these changes.  I think that's okay, we're now just filtering what fi_getinfo returns for provider info and only picking the one that fits with the specified include/exclude provider lists.

hkuno: Okay. Thank you for clarifying, Howard!

opal/mca/btl/ofi/btl_ofi_component.c
```diff
@@ -51,12 +51,47 @@ static char *ofi_progress_mode;
 static bool disable_sep;
 static int mca_btl_ofi_init_device(struct fi_info *info);
 
+static int
```
bwbarrett, line 54: This code is exactly the same in the MTL and BTL.  This should be moved to common and only exist once.

wckzhang: Makes sense, will do that.

opal/mca/common/ofi/common_ofi.h
```diff
@@ -38,6 +38,7 @@ OPAL_DECLSPEC void opal_common_ofi_mca_register(void);
 OPAL_DECLSPEC void opal_common_ofi_mca_deregister(void);
 OPAL_DECLSPEC struct fi_info* opal_common_ofi_select_ofi_provider(struct fi_info *providers, 
                                                                   char *framework_name);
+OPAL_DECLSPEC int opal_common_ofi_is_in_list(char **list, char *item);
```
bwbarrett, line 41: i'd love some documentation :).

wckzhang: I tried my best, it's kind of a funky function. Maybe the provider list logic that currently exists should be changed to explicitly call out layered providers (ie. tcp, tcp;ofi_rxm) as separate providers.

bwbarrett:
bot:ompi:retest

looks like one of the testers failed in calling in its success.

wckzhang:
Can anyone with merge permissions merge this patch? I will create a backport PR for 4.1.x branch after that.

PR #7977, opened 2020-07-31
Change env var name and behavior to reflect the OpenSHMEM spec

manjugv:
* Change name from SHMEM_SYMMETRIC_HEAP_SIZE to SHMEM_SYMMETRIC_SIZE
* Either SHMEM_SYMMETRIC_SIZE or SMA_SYMMETRIC_SIZE can be set
* When both SHMEM_SYMMETRIC_SIZE and SMA_SYMMETRIC_SIZE variables are set, SHMEM_SYMMETRIC_SIZE’s value takes precedence.

### DIFF
--- a/oshmem/include/shmem.h.in
+++ b/oshmem/include/shmem.h.in
@@ -84,7 +84,7 @@ enum {
 /* size of symmetric heap in bytes.
  * Can be qualified with the letter 'K', 'M', 'G' or 'T'
  */
-#define SHMEM_HEAP_SIZE "SHMEM_SYMMETRIC_HEAP_SIZE"
+#define SHMEM_HEAP_SIZE "SHMEM_SYMMETRIC_SIZE"
 
 /*
  * Type of allocator used by symmetric heap
--- a/oshmem/info/info.c
+++ b/oshmem/info/info.c
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2015      Mellanox Technologies, Inc.
- *                         All rights reserved.
- * Copyright (c) 2018      Amazon.com, Inc. or its affiliates.  All Rights reserved.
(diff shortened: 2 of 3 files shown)


manjugv:
@janjust Can you please review ? Please add additional reviewers, if required.

janjust:
@manjugv 👍 , but looks like you forgot to sign-off your commit

jsquyres:
Note that you will not be able to bring these commits back to existing release branches because you will be breaking backwards compatibility.

manjugv:
> Note that you will not be able to bring these commits back to existing release branches because you will be breaking backwards compatibility.

Thanks for the heads up, Jeff. 

Unfortunately, I don't see another way without breaking backward compatibility or causing more confusion (with three variables). OpenSHMEM users specifically requested to avoid using multiple variables (even with this change, we have one deprecated and one spec required variable).

jsquyres:
> Unfortunately, I don't see another way without breaking backward compatibility or causing more confusion (with three variables). OpenSHMEM users specifically requested to avoid using multiple variables (even with this change, we have one deprecated and one spec required variable).

The only way you can do it and preserve backwards compat is to make aliases for the old names.

awlauria:
@manjugv please repush using `git commit -x`.

lanl-ompi:

awlauria:
@manjugv @janjust please rebase this with the signedoff so it can be merged...Then cherry-pick to v5.0.x. @gpaulsen

awlauria:
@manjugv @janjust ping. This is just missing the `git commit --amend -s` + force push.

janjust:
@manjugv can you please do it so we can merge this?

awlauria:
v5.0.x is a good time for this since it's established it will break abi.

Issue #7955, opened 2020-07-21
Question: obtain/specify the memory alignement in MPI_Win_allocate and MPI_Alloc_mem


thomasgillis:
Thank you for taking the time to answer my question

## My question:
I am using `MPI_Win_allocate` to allocate RMA memory.
I need to access it using a standard for loop and I would like to help its vectorization.
Therefore I wish to inform the compiler about the alignment of the memory, allocated using `MPI_Win_allocate`.

I have then two associated questions:
- is it possible to retrieve the alignment of the memory (using a function, I can always compute it afterward of course)? The ideal situation would be to have a macro for it.
- is it possible to request a given alignment when using `MPI_Win_allocate` and/or `MPI_Alloc_mem`? Of course, compatibility with the MPI internal alignment has to be ensured.

-> I use Open MPI v4.0.3

devreal:
@thomasgillis Unfortunately, MPI 3.1 had no guarantee on alignment and/or any facility to query or control the alignment of window memory. However, MPI 4.0 will provide info keys for `MPI_Alloc_mem` and `MPI_Win_allocate` to hint at a desired alignment. I don't think this has been implemented in Open MPI but I take this issue as a reminder to implement this feature soon.

For the time being, you can allocate aligned memory using `posix_memalign` and create the window using `MPI_Win_create`. That may negatively impact performance implications for on-node communication but should give you the same performance for inter-node communication IIRC.

thomasgillis:
@devreal thank you for your feedback!

I saw recently that the standard MPI-4 is not complete yet
Do you have an estimated time of arrival for its implementation? 2022?

devreal:
IIRC, the 4.0 standard is slated for publication by the end of this year / early next year. I started looking into implementing this and hope to put up a PR early August.

devreal:
@thomasgillis Support or the `mpi_minimum_alignment` info key has been merged into `master` and will be part of the 5.0 release (https://github.com/open-mpi/ompi/pull/7985). Does that solve your issue? Can we close this ticket?

thomasgillis:
Yes! Thx a lot.

PR #7985, opened 2020-08-05
Add support for mpi_minimum_alignment info key

devreal:
This PR adds supports for the `mpi_minimum_alignment` info key that will likely be part of MPI 4.0 and will be available for `MPI_Alloc_mem`, `MPI_Win_allocate`, and `MPI_Win_allocate_shared`. The key is meant to request a specific alignment for the allocated memory. Since info keys are mere hints, they can be ignored if the value is out of range (larger than the page size for example) or the underlying implementation does not support it.

This has been requested in https://github.com/open-mpi/ompi/issues/7955

The current state adds support for this key to `base/mpool` to handle it in `MPI_Alloc_mem`. It also adds support for the info key to `osc/rdma`, `osc/sm`, and `osc/portals4`. The `osc` integration uses `mca_mpool_base_default_module->mpool_alloc` to allocate aligned memory as it already properly chooses `posix_memalign` and falls back to a custom implementation if that is not available.

Notably, `osc/ucx` is missing from the list because I'm not sure how to properly work that with UCX. I considered allocating aligned memory and passing it to `ucp_mem_map` instead of `UCP_MEM_MAP_ALLOCATE` but I'm worried that it may negatively impact performance of intra-node communication (because UCX cannot allocate shared memory in that case). Maybe some of UCX folks can comment on that?


rhc54:
I can't review this as I have no knowledge/understanding of this code area.

devreal:
@rhc54 no problem, I just thought I tag you because I saw your name in `git blame` :)

devreal:
I pushed a fix required for the `mpool_memkind` test to pass. The test passes `NULL` as the `info` pointer to `mca_mpool_base_alloc`. There seems to be a discrepancy between the documentation and the code in that the info object has not been used so far but the documentation states that the mpool hint can be contained in the info object: https://github.com/open-mpi/ompi/blob/master/opal/mca/mpool/mpool.h#L150 Which one is correct then?

Also, is it correct to assume that `NULL` will never be passed as `info` pointer to the `osc` components? (because it is called from the MPI layer and `MPI_INFO_NULL` points to an actual object)

gpaulsen:
@hjelmn , @markalle ?

lanl-ompi:

ibm-ompi:

 Gist: https://gist.github.com/0656a308e6acf35577f89c78d412d725

ibm-ompi:

 Gist: https://gist.github.com/44724a0fd67943f107ebc7cfe29cf9f1

ibm-ompi:

 Gist: https://gist.github.com/39be1e7ef86be35586829fdc537bd90f

hppritcha (approved):

PR #7986, opened 2020-08-05
threads: fix a problem hit when using ex-hwloc

hppritcha:
problem was that adding --with-hwloc option pointing at a non-default location for an hwloc install
can add hwloc lib info to the global LDFLAGS, which results
in some of the runtime checkes in the pthread configury to fail.  This commit resolves this problem
by using pthreads-only LDFLAGS for building the runtime checks in the pthreads configury.

related to #7644


hppritcha:
bot:ompi:restart

hppritcha:
bot:ompi:retest

hppritcha:
bot:ompi:retest

PR #7991, opened 2020-08-10
v4.1.x: btl/ofi: Use common provider include/exclude list

wckzhang:
The btl/ofi does not currently utilize the common ofi include/exclude
list. Added verification code similar to the mtl/ofi that will check if
the info object is in the include or exclude list. If it isn't in the
include list or is in the exclude list, validate_info will return
OPAL_ERROR. The btl/ofi will no longer pass a provider name as a hint
when calling getinfo, instead filtering the provider during
validate_info.

This patch also moves the is_in_list MTL function into common code and
adds additional debugging output to the BTL to match the MTL standard.

(cherry picked from commit 9b8f463a768206a26e5b1dcea0612a403462b1d0)

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -306,26 +306,6 @@ ompi_mtl_ofi_progress_no_inline(void)
 	return ompi_mtl_ofi_progress();
 }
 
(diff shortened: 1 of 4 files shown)


wckzhang:
Original PR: https://github.com/open-mpi/ompi/pull/7975

Issue #8001, opened 2020-08-13
OFI MTL is not happy with GNI provider


hppritcha:
Need to patch the OFI MTL for use with the GNI provider when using FI_VERSION 1.5 or higher.

See https://github.com/ofiwg/libfabric/issues/6194.

hppritcha:
closed via #8002 and #8027

PR #8002, opened 2020-08-13
OFI: patch OFI MTL for GNI provider

hppritcha:
Uncovered a problem using the GNI provider with the OFI MTL.
See https://github.com/ofiwg/libfabric/issues/6194.

Related to #8001

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_component.c
+++ b/ompi/mca/mtl/ofi/mtl_ofi_component.c
@@ -941,6 +941,17 @@ select_prov:
     }
 
     /**
+     * Unfortunately the attempt to implement FI_MR_SCALABLE in the GNI provider
+     * doesn't work, at least not well.  Since we're asking for the 1.5 libfabric
+     * API now, we have to tell GNI we want to use Mr. Basic.  Using FI_MR_BASIC
+     * rather than FI_MR_VIRT_ADDR | FI_MR_ALLOCATED | FI_MR_PROV_KEY to stay
+     * compatible with older libfabrics.
+     */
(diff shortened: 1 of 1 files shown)


ompi/mca/mtl/ofi/mtl_ofi_component.c
```diff
@@ -940,6 +940,17 @@ ompi_mtl_ofi_component_init(bool enable_progress_threads,
...
     }
 
+    /**
+     * Unfortunately the attempt to implement FI_MR_SCALABLE in the GNI provider
+     * doesn't work, at least not well.  Since we're asking for the 1.5 libfabric
+     * API now, we have to tell GNI we want to use Mr. Basic.  Using FI_MR_BASIC
+     * rather than FI_MR_VIRT_ADDR | FI_MR_ALLOCATED | FI_MR_PROV_KEY to stay
+     * compatible with older libfabrics.
+     */
+    if (!strcmp(prov->fabric_attr->prov_name,"gni")) {
```
wckzhang, line 950: Do you care about any utility providers?

wckzhang: The suggestion here would be to do a strncasecmp so it'll look at the leftmost provider for "gni" - future proofing for any layering that will happen on top.

hppritcha: good idea I'll change to strncasecmp.
I'm nervous to make changes for other providers, hence the check only for gni.

hppritcha: pushed change to strncmp

hppritcha:
@wckzhang updated PR with your suggested change

PR #8007, opened 2020-08-14
Improve predefined pack/unpack performance using mpich methods.

markalle:
For the original mpich implementation, see:
https://github.com/pmodels/mpich/blob/9ab5fd06af2a648bf24214f0d9cff0ee77ee3e7d/src/mpi/datatype/veccpy.h

Small testcase to demonstrate the performance difference:
https://gist.github.com/markalle/9f92e9facbd71136bcfb9f0e0305a1da
% mpicc -o x packperf_nc.c
% mpirun -np 1 ./x

```
Before:
> pack dtbig    :    943    863    863    862    862  (avg 879)  usec
> unpack dtbig  :    919    955    917    917    917  (avg 925)  usec
> pack dtsmall  :    810    810    810    831    810  (avg 814)  usec
> unpack dtsmall:    947    954    996    941    969  (avg 962)  usec
After:
> pack dtbig    :    205    124    120    118    118  (avg 137)  usec
> unpack dtbig  :    122    120    120    120    120  (avg 120)  usec
> pack dtsmall  :    133    122    122    121    121  (avg 124)  usec
> unpack dtsmall:    124    124    123    123    123  (avg 123)  usec
```

Having lots of small memcpy() was slower than the mpich code that
uses blocks of array assignments.

Notes about what changed:
* at the top-level mpi/c/pack.c and unpack.c it now sometimes
  turns (count, dtype) into (1, newdtype) with an newdtype made
  by MPI_Type_contiguous(count, dtype).  This is because the lower
  level pack/unpack always iterates over description elements and
  when it sees (count,dtype) there's no possibility of a single
  description element describing the whole data.

  I'm triggering that code only when the count is >=250 and
  the type is non-contiguous.  It likely only needs to be triggered
  if the datatype has a single element such that the
  element.count * element.extent == dtype.extent but that would
  be more code to detect.
* in Datatype_internal.h I moved the macros around a little so
  I could reuse them in the new unrolled array assignments code.
  That way I don't have to figure out that INT4 is int32_t, because
  those macros already have that info.  The diff probably looks
  large but there isn't that much going on there.
* in opal_datatype_pack/unpack.h there's an extra section to call
  the mpich vector copying code for a description element if
  it's a certain size, and continue with the regular code if
  the mpich call rejects it (due to not recognizing the element.id,
  or due to it being cuda memory, or due to alignment)
* the new opal_datatype_pack_unpack_predefined.h largely copied
  from mpich.  The macros boil down to unrolled array assignments.
  I recycled the opal_datatype_internal.h macros to get the values
  for TYPE.  That way I don't have to figure out whether
  SHORT_FLOAT_COMPELX is short float _Complex or opal_short_float_complex_t
  or unavailable for example.

Extra notes about the new pack/unpack routine:
* For checking cuda memory I didn't check every item in the vector,
  only the first and possibly the last, since I don't think individual
  description elements should be spanning gpu and system memory.
* I didn't use the unaligned-stride code from mpich, instead just
  rejecting anything unaligned

Licensing:
https://github.com/pmodels/mpich/blob/9ab5fd06af2a648bf24214f0d9cff0ee77ee3e7d/src/mpi/datatype/veccpy.h
where the code came from says
```
> /*
>  *  (C) 2001 by Argonne National Laboratory.
>  *      See COPYRIGHT in top-level directory.
>  */
```
And I pasted the above mentioned COPYRIGHT at the top of
opal_datatype_pack_unpack_predefined.h


markalle:
This came from Austen btw, then I made some updates to it

awlauria:
@bosilca can you please take a look when you get the chance?

Thanks.

ompi/mpi/c/pack.c
```diff
@@ -77,10 +77,62 @@ int MPI_Pack(const void *inbuf, int incount, MPI_Datatype datatype,
 
     OPAL_CR_ENTER_LIBRARY();
 
+/*
```
bwbarrett, line 80: indentation?

ompi/mpi/c/pack.c
```diff
@@ -77,10 +77,62 @@ int MPI_Pack(const void *inbuf, int incount, MPI_Datatype datatype,
...
+    /*
+     *  This magic number 250 is just empirical from one datapoint on one
+     *  machine, using [int,space] as the datatype.  There's some overhead
+     *  in constructing a new datatype, so just walking a sequence of
+     *  description elements is better for a short list of elements vs
+     *  creating a potentially shorter list and hoping the vector-walking
+     *  of the new elements is faster.  This could maybe be tuned dynamically
+     *  but it doesn't really seem worth it.
+     */
+    if (incount > 250) {
```
bwbarrett, line 107: Can you at least make the magic constant a #define with this comment?

ompi/mpi/c/unpack.c
```diff
@@ -79,12 +79,64 @@ int MPI_Unpack(const void *inbuf, int insize, int *position,
...
+    /*
+     *  This magic number 250 is just empirical from one datapoint on one
+     *  machine, using [int,space] as the datatype.  There's some overhead
+     *  in constructing a new datatype, so just walking a sequence of
+     *  description elements is better for a short list of elements vs
+     *  creating a potentially shorter list and hoping the vector-walking
+     *  of the new elements is faster.  This could maybe be tuned dynamically
+     *  but it doesn't really seem worth it.
+     */
+    if (outcount > 250) {
```
bwbarrett, line 109: This code is identical to the pack case.  Maybe an inline function that is shared between the two?

opal/datatype/opal_datatype_pack_unpack_predefined.h
```diff
@@ -0,0 +1,532 @@
+/*
+ * Copyright (c) 2020      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2002      University of Chicago
+ * Copyright (c) 2001      Argonne National Laboratory
```
bwbarrett, line 4: I think we also need to add a note in our COPYING / COPYRIGHT file.

markalle:
@bwbarrett Thanks for the review, I repushed with the changes you suggested.
* fixed the indentation
* moved the 250 constant to a header
* made static inline for a couple ```ompi_datatype_consolidate_*()``` functions, and used those in pack/unpack.c

For the LICENSE file I added the two lines
```
Copyright (c) 2002      University of Chicago
Copyright (c) 2001      Argonne National Laboratory
```
But I'm uncertain where to put the part that says
```
 * The following is a notice of limited availability of the code, and disclaimer
 * which must be included in the prologue of the code and in all source listings
 * of the code.
```
Can I put roughly the same wording I used in the new ```.h``` file at the bottom of LICENSE?
```
This software includes source based on MPICH code which contained the following
notice in their top-level COPYRIGHT file:
<their copyright here>
```

markalle:
I think this PR is ready, but would like to get resolution on licensing text (@jsquyres maybe?).  I added the MPICH blurb to LICENSE even though I'm not sure it's needed.  Geoffrey thinks it is, while I think just putting the blurb in ```opal_datatype_pack_unpack_predefined.h``` satisfies the requirement.

MPICH blurb is here:
https://github.com/pmodels/mpich/blob/9ab5fd06af2a648bf24214f0d9cff0ee77ee3e7d/COPYRIGHT

I pasted it at the top of ```opal_datatype_pack_unpack_predefined.h``` and the bottom of LICENSE

gpaulsen:
@jsquyres I think you've dealt with Open-MPI LICENSE wording in the past.  Can you please take a look?

lanl-ompi:

jsquyres:
I am not a lawyer and this is not legal advice, but it looks like we followed the terms of their license.

markalle:
repushing with no changes, just a rebase

markalle:
Anyway if this LICENSE file with MPICH text pasted under the previous OMPI license is acceptable then I'd like to get this in

markalle:
bot:ompi:retest

markalle:
repushed just to rebase, no changes

awlauria:
@rajachan - @bwbarrett said there was some legal stuff amazon had to check regarding this PR - do you know where that stands?

rajachan:
Let me go figure out where that's that. WIll report back.

gpaulsen:
@rajachan Any progress?

ompi/datatype/ompi_datatype.h
```diff
@@ -418,5 +418,94 @@ OMPI_DECLSPEC int ompi_datatype_pack_external_size( const char datarep[], int in
         }                                                               \
     }
 
+/*
+ * Sometimes it's faster to operate on a (count,datatype) pair if it's
```
bosilca, line 422: This is true in MPICH but not in OMPI, because the optimization step is supposed to automatically consolidate such cases. I am talking about the code in opal_datatype_optimize_short (more precisely opal_datatype_optimize.c around line 158). It is possible that code fails to detect the pattern, and I would love to see such an example. Can you provide an example for me to look into ?

bosilca: Let me answer my own question. The code I was pointing, identify and optimize datatypes that exhibit the pattern you mention in your comment, except when the pattern is not part of the datatype itself, but it is part of the MPI calls (and therefore has a count around a single data entry pointing to a non contiguous element).

gpaulsen (approved):
The LICENSE change looks good.
similar code has gone through a full IBM Spectrum MPI Quality cycle, and is production ready.

ompi/datatype/ompi_datatype.h
markalle: Yeah, if it was all contained in a single MPI datatype I think the pattern would have been detected and optimized.  It's just that a (count,dt) pair from an MPI like pack call never gets analyzed together by that code.

So in the gist testcase from the top, the two "dtbig" lines are nicely optimized, but in the two "dtsmall" cases it needs something like this datatype_consolidate structure in order to trigger the same optimization

markalle:
rebased, no changes other than I previously forgot the "Copyright IBM Corp..." stuff so I added that

PR #8012, opened 2020-08-18
ofi mtl: fix problem with mrecv

hppritcha:
the ofi mtl mrecv was not properly setting the message in/out
arg to MPI_MRECV to MPI_MESSAGE_NULL.


hppritcha:
related to #8013

Issue #8010, opened 2020-08-18
reduce_scatter and reduce_scatter_block algorithms incorrectly handling noncommutative ops


wckzhang:
OMPI v4.1.x installed from git clone

Ran ompi_tests repository tests:

ibm/collective/reduce_scatter_block_nocommute_stride 
ibm/collective/reduce_scatter_block_nocommute_stride_in_place 
ibm/collective/reduce_scatter_nocommute_stride 
ibm/collective/reduce_scatter_nocommute_stride_in_place 

These tests fail with the new default tuned algorithms. I manually selected algorithms to test and found that for RS, algorithms 2, 3, 4 (recursive_halving, ring, butterfly) fail these tests and for RSB, algorithms 2 and 4 (recursive_doubling, butterfly) fail these tests.

The new fixed code only says algorithm 3 (recursive halving) for RSB and algorithm 2 and 3 (recursive halving, ring) do not support non-commute ops. This doesn't match the test results and these algorithms need to be labelled as commute only or fixed if they are supposed to be non commute capable.

wckzhang:
@bosilca Do you know if any of these algorithms are incorrectly labelled?

bwbarrett:
@wckzhang is there a reason this issue is still open?

wckzhang:
Need to investigate these algorithms and relabel them being commute only or determine if they have bugs in their behavior.

PR #8019, opened 2020-08-25
coll/tuned: Revert RSB and RS default algorithms

wckzhang:
Reduce scatter block and reduce scatter algorithms were hitting
correctness issues for non commutative strided tests. We will revert to
the original default algorithms for those two collectives (basic linear
and non overlapping respectively) in the non commutative op case.

See #8010


wckzhang:
Tested the fix and it passes the mtt tests that are failing:

ibm/collective/reduce_scatter_block_nocommute_stride
ibm/collective/reduce_scatter_block_nocommute_stride_in_place
ibm/collective/reduce_scatter_nocommute_stride
ibm/collective/reduce_scatter_nocommute_stride_in_place

This fix will also be backported into v4.1.x branch after it is merged.

PR #8026, opened 2020-08-28
suppress icc long double message

hppritcha:
improve configury to check whether icc is handling no long double.
This prevents seeing 100s of messages like this:

icc: command line warning #10148: option '-Wno-long-double' not supported

A similar patch will be needed for pmix.

(cherry picked from commit 6df0e53421b285eef2dc0b442d818026301d29f1)


hppritcha:
@gpaulsen nothing urgent, just going in to the queue for if we do 4.0.6

PR #8029, opened 2020-08-31
OFI: patch OFI MTL for GNI provider

hppritcha:
Uncovered a problem using the GNI provider with the OFI MTL.
See https://github.com/ofiwg/libfabric/issues/6194.

Related to #8001

(cherry picked from commit d6ac41cbbd8dbfdbf1a87533e2e292f2508a85ff)


hppritcha:
@gpaulsen this can wait till v4.0.5 has been out for a bit - a week or so.

hppritcha:
I forgot we don't need this patch for v4.0.x since its using the 1.0 libfabric api.  The problem this patch addresses is only needed when using the 1.5 or more recent libfabric api.

Issue #6987, opened 2019-09-18
Regression: UCX osc: dynamic windows broken on master


devreal:
I noticed that support for dynamic windows is broken in the UCX osc on `master`. The module's worker pool memory (`module->mem`) is `NULL` if it's a dynamic windows but upon flush and window-free this pointer is accessed unconditionally, inadvertently leading to a Segfault. Upon communication, the type of window is checked and the dynamic window info is populated (in `get_dynamic_win_info`) but never used anywhere, so that will eventually Segfault, too. I'm aware that the UCX osc in `master` is under development so I just open this issue for tracking purposes.

artpol84:
@devreal, thanks for the heads up

devreal:
I just came back to this issue and found that it is still broken. This is actually a regression that exists in 5.0 and master.

Consider the following example:

```C
#include <mpi.h>
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

#define NUM_REPS 1000

static void
allocate_dynamic(size_t elsize, size_t count, MPI_Win *win, MPI_Aint *disp_set, char **b)
{
    char *base;
    MPI_Aint disp;
    int rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
    
    if (MPI_Win_create_dynamic(MPI_INFO_NULL, MPI_COMM_WORLD, win) != MPI_SUCCESS) {
        printf("Failed to create dynamic window!\n");
        exit(1);
    }

    if (MPI_Alloc_mem(elsize*count, MPI_INFO_NULL, &base) != MPI_SUCCESS) {
        printf("Failed to allocate memory!\n");
        exit(1);
    }


    if (MPI_Win_attach(*win, base, elsize*count) != MPI_SUCCESS) {
        printf("Failed to attach memory to dynamic window!\n");
        exit(1);
    }


    MPI_Get_address(base, &disp);
    printf("Offset at process %i: %p (%lu)\n", rank, base, disp);
    MPI_Allgather(&disp, 1, MPI_AINT, disp_set, 1, MPI_AINT, MPI_COMM_WORLD);

    *b = base;
}

int main(int argc, char **argv) 
{
  int *baseptr;
  MPI_Win win;
  int rank, size;

  MPI_Init(&argc, &argv);

  MPI_Comm_size(MPI_COMM_WORLD, &size);
  MPI_Comm_rank(MPI_COMM_WORLD, &rank);

[...]

PR #8035, opened 2020-09-04
UCX: do not dereference NULL pointer in wpmem_[free|flush]

devreal:
Flushing or freeing a newly created dynamic window leads to NULL being passed to the wpmem functions.

Addresses part of what was reported in https://github.com/open-mpi/ompi/issues/6987. This allows at least to flush/free an empty dynamic window without Segfault. It's a bandaid, not a full fix, but I ran into this multiple times with some code that happens to create/flush/free empty dynamic windows so I need a fix for that at least.


hoopoepg (approved):

janjust (approved):

artpol84 (commented):
Looks good!

Issue #7702, opened 2020-05-06
Argobots integration vs UCX PML


devreal:
I tried to run a rather simple benchmark (`osu_latency_mt` ported to use Argobots) to test the Argobots integration on an IB cluster using UCX. It appears that the UCX PML does not use the sync objects in blocking operations but instead hammers the UCX library, with occasional calls to `opal_progress`. Of course, there will never be a switch to another argobot in that case, causing deadlocks if there are more argobots than execution streams.

One way to work around this I guess is to call `ABT_thread_yield` in `opal_progress`. So I replaced the call to `sched_yield` with `ABT_thread_yield`. Curiously, it seems that running with `--mca mpi_yield_when_idle true` does not actually cause `opal_progress_yield_when_idle` to be set to `true` so that code path isn't taken (I am not familiar with the mca parameter code and couldn't spot anything wrong on a first glance). I can open a separate issue for this if people think that this is fishy.

If I remove the check for  `opal_progress_yield_when_idle` in `opal_progress` things start to work. Now the latency seems sensitive to `pml_ucx_progress_iterations`, the smaller the better (as the UCX PML calls into `opal_progress` more often).

hppritcha:
@shintaro-iwasaki - how did you handle this in MPICH?

shintaro-iwasaki:
In MPICH, all the possible busy-wait calls in the MPICH runtime call a yield function (i.e., `ABT_thread_yield()` for Argobots or `sched_yield()` for Pthreads). Blocking system call might be problematic in Open MPI, though I haven't encountered this issue in MPICH.

There are a few pieces to fix:

#### libevent
We should need ULT-aware (in this case, Argobots-aware) libevent, which has been dropped from #6578 and I am separately making it. It is still being debugged now (https://github.com/shintaro-iwasaki/libevent/tree/2.0.22-abt) -- this will be pushed to somewhere other than my personal repository else once this prototype gets stable enough.

#### MCA/thread yield
If such a busy wait calls `sched_yield()`, this can be replaced with another MCA/thread yield function (it needs a PR). I suspect there are busy waits that call neither `sched_yield()` nor a `pause`-equivalent operation; we need to identify them and see the solution.

Issue #8036, opened 2020-09-07
Argobots and Qthread configure detection need fixing/improvements


devreal:
Working with current master (`g560ebc5780`), I am trying to compile Open MPI with either Argobots or Qthread support and noticed the following:

1) There had been a patch to improve the detection of Argobots in `configure` (https://github.com/open-mpi/ompi/pull/7675/files) but the current state does not seem to work. I am passing `--with-argobots=$HOME/opt/argobots-1.0/` to configure but all I am getting is the following message in `config.log`:
```
configure:5285: +++ Configuring MCA framework threads
configure:153773: checking for no configure components in framework threads
configure:153775: result: 
configure:153777: checking for m4 configure components in framework threads
configure:153779: result: argobots, pthreads, qthreads
configure:5300: --- MCA component threads:argobots (m4 configuration macro, priority 30)
configure:153867: checking for MCA component threads:argobots compile mode
configure:153870: result: static 
configure:156050: checking if MCA component threads:argobots can compile
configure:156052: result: no
```
Any ideas how to get more information on what is going wrong? I don't see a failed compilation attempt anywhere in the logs...

2) There is no way to provide `configure` with a path to a custom qthread installation. Something along the lines of `--with-qthreads=/path/to/qthreads/` should be provided.

shintaro-iwasaki:
Thank you for taking a look at this issue (I am also keeping track of #8037).

For Argobots, although the Argobots version has some issues on multithreading tests, which is still under investigation, at least compilation and configuration succeeded on our CI environment last weekend (https://jenkins-pmrs.cels.anl.gov/job/ompi-weekly-argobots/compiler=gnu,label=argobots_centos64/24/consoleFull).

The configuration line is as follows:
```sh
/tmp/jenkins5540533621955127193.sh:52> ./configure '--with-threads=argobots' \
'--with-argobots=/var/lib/jenkins-slave/workspace/ompi-weekly-argobots/compiler/gnu/label/argobots_centos64/argobots' \
'--with-libevent=/var/lib/jenkins-slave/workspace/ompi-weekly-argobots/compiler/gnu/label/argobots_centos64/libevent' \
'--prefix=/var/lib/jenkins-slave/workspace/ompi-weekly-argobots/compiler/gnu/label/argobots_centos64/ompi' \
'--with-cray-xpmem=no'
```
[...]

PR #8037, opened 2020-09-07
Move yield capability to opal thread component

devreal:
This PR is based on a discussion at https://github.com/open-mpi/ompi/issues/6433 and aims to introduce alternative yielding mechanisms using `usleep` and `nanosleep`, to be chosen at runtime through MCA parameters. In doing so, it moves the yielding implementation into the opal thread components and thus provide the correct yielding capabilities for Argobots/Qthreads.

This adds two new mca parameters:
threads_pthreads_yield_strategy to choose the strategy (sched_yield and nanosleep, pthreads only), and
threads_pthreads_nanosleep_time (time passed to nanosleep, pthreads only)

A thread component may also signal that yield-when-idle should be the default. This is true for Argobots/Qhtreads because they are cooperatively scheduled and thus have to rely on MPI to yield the underlying thread and allow for all communication operations to be initiated eventually.

Marking this as WIP as I am unable to properly test the Argobots/Qthreads integration at this time, see https://github.com/open-mpi/ompi/issues/8036. I still wanted to put it out for people to comment and in case someone else was following the discussion linked above.

Fixes https://github.com/open-mpi/ompi/issues/7702

### DIFF
--- a/opal/mca/threads/pthreads/threads_pthreads_yield.c
+++ b/opal/mca/threads/pthreads/threads_pthreads_yield.c
@@ -1,0 +1,89 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2020      High Performance Computing Center Stuttgart,
+ *                         University of Stuttgart.  All rights reserved.
+ *
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "opal_config.h"
+#include <time.h>
+#ifdef HAVE_SCHED_H
+#include <sched.h>
+#endif
+
+#include "opal/constants.h"
+#include "opal/mca/threads/thread.h"
+#include "opal/mca/threads/pthreads/threads_pthreads.h"
+
+static void opal_thread_pthreads_yield_sched_yield(void);
+static void opal_thread_pthreads_yield_nanosleep(void);
+
+typedef enum {
+    OPAL_PTHREADS_YIELD_SCHED_YIELD = 0,
+    OPAL_PTHREADS_YIELD_NANOSLEEP
+} opal_threads_pthreads_yield_strategy_t;
+
+static mca_base_var_enum_value_t yield_strategy_values[] = {
+    {OPAL_PTHREADS_YIELD_SCHED_YIELD, "sched_yield"},
+    {OPAL_PTHREADS_YIELD_NANOSLEEP,   "nanosleep"},
+    {0, NULL}};
+
+
+
+/* Number of nanoseconds to nanosleep, if enabled */
+static uint64_t yield_nsleep_nanosecs;
+/* The time to nanosleep, if enabled */
+static struct timespec yield_nsleep_time = {.tv_sec = 0, .tv_nsec = 1};
+static opal_threads_pthreads_yield_strategy_t yield_strategy = OPAL_PTHREADS_YIELD_SCHED_YIELD;
+
+opal_threads_pthreads_yield_fn_t *opal_threads_pthreads_yield_fn = &opal_thread_pthreads_yield_sched_yield;
+
+int opal_threads_pthreads_yield_init(const mca_base_component_t *component)
+{
+    mca_base_var_enum_t *yield_strategy_enumerator;
+    mca_base_var_enum_create("pthread_yield_strategies", yield_strategy_values, &yield_strategy_enumerator);
+
+    (void) mca_base_component_var_register(component, "yield_strategy",
+                                           "Pthread yield strategy to use",
+                                           MCA_BASE_VAR_TYPE_INT, yield_strategy_enumerator, 0, 0, OPAL_INFO_LVL_3,
+                                           MCA_BASE_VAR_SCOPE_LOCAL, &yield_strategy);
+    switch(yield_strategy) {
+        case OPAL_PTHREADS_YIELD_NANOSLEEP:
+            opal_threads_pthreads_yield_fn = &opal_thread_pthreads_yield_nanosleep;
+            break;
+        default:
+            /* use initial value */
+            break;
+    }
+
+    OBJ_RELEASE(yield_strategy_enumerator);
+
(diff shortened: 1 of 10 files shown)


ibm-ompi:

 Gist: https://gist.github.com/a2564f8c8ba9bdb2332b6d1fe978e3b7

ibm-ompi:

 Gist: https://gist.github.com/949782597258de0c7ad48b8833f04858

ibm-ompi:

 Gist: https://gist.github.com/6fd4bdb517db3a0833ed3c590fa23024

rhc54:
@devreal I suspect you are aware of this, but just as a reminder: calling something via an MCA interface costs about 5ns. So doing that to execute a 1ns `nanosleep` doesn't make a lot of sense. I realize that the point of the function call is to force a context change, but I'm just pointing out that the time cost in the `opal_progress` critical path is non-zero.

devreal:
I think it boils down to a direct function call. Given that all pthread options either require a syscall or at least activate some low-level user-space thread scheduler the added function call should be negligible. I agree that a 1ns sleep is unrealistic but I wasn't sure what a good value actually is so this was the conservative choice. In part that is why it's configurable :) For Argobots and Qthreads the calls to `opal_thread_yield` can be inlined since they are directly dispatching to the ULT libraries.

opal/mca/threads/pthreads/threads_pthreads_yield.c
```diff
@@ -0,0 +1,54 @@
...
+#include "opal/mca/threads/pthreads/threads_pthreads.h"
+#include "opal/sys/atomic.h"
+
+
+opal_threads_pthreads_yield_strategy_t opal_threads_pthreads_yield_strategy = OPAL_PTHREADS_YIELD_SCHED_YIELD;
+uint64_t opal_threads_pthreads_yield_usleep = 1;
+uint64_t opal_threads_pthreads_yield_nsleep = 100;
+
+
+void opal_thread_yield(void)
```
bwbarrett, line 29: It seems like we can do a little better here.  If opal_thread_yield() were a function pointer to the right case, we could get 1 function call and the case statement out of the critical path.

opal/mca/threads/pthreads/threads_pthreads_yield.c
```diff
@@ -0,0 +1,54 @@
...
+        sched_yield();
+        break;
+#endif
+        /* fall-through if sched_yield is not available */
+    case OPAL_PTHREADS_YIELD_USLEEP:
+        usleep(opal_threads_pthreads_yield_usleep);
+        break;
+    case OPAL_PTHREADS_YIELD_NANOSLEEP:
+    {
+        static struct timespec sleeptime = {.tv_sec = -1, .tv_nsec = -1};
```
bosilca, line 43: There is little potential for race conditions here, but the potential is inside. You have added a registration function, why don't you compute this value there?

opal/mca/threads/pthreads/threads_pthreads_yield.c
devreal: Done, the yield strategy in use is now a function pointer and the call to `opal_thread_yield` can be inlined so all that's left is a call through that pointer.

opal/mca/threads/pthreads/threads_pthreads_yield.c
devreal: Agreed, it's now set up after reading the mca parameter.

devreal:
I moved some around some of the pthread-specific code to reduce the number of non-static symbols and address the comments above. I also made the `threads_pthreads_yield_strategy` mca parameter an enum to catch invalid values and allow for text-based selection.

bosilca:
We end up doing a function call into a function call, for nothing. Why not having opal_thread_yield the function pointer that gets called only if non NULL, and then the threading layer set it to whatever it needs ?

devreal:
I think we should avoid function pointers where possible because indirect calls are more costly and prevent some optimizations (sibling call optimization and LTO, for example). Right now, `opal_thread_yield` is marked `static inline` so I expect compilers to directly issue calls to the underlying yield function where possible (Argobots, Qthreads) and the function pointer for pthreads. We could make it a macro but I don't think it makes a difference.

bosilca:
OMPI embraced function pointers long ago, we evaluated the approach and it was well performing.

bwbarrett:
> We end up doing a function call into a function call, for nothing. Why not having opal_thread_yield the function pointer that gets called only if non NULL, and then the threading layer set it to whatever it needs ?

It would be cheaper to have an empty function than an if and a call...  function pointers are cheap(ish).  Branches in front of function pointers, not so much.

hjelmn:
Is there a good reason to have both usleep and nanosleep? From what I remember nanosleep is considered superior to usleep in most respects.

devreal:
None in particular, other than giving people options to experiment with. After looking around the internet it looks like `nanosleep` is the way to go over `usleep` now. I guess we could remove `usleep`.

devreal:
I removed the `usleep` option and will squash the commit before merge (kept it separate in case someone still wants it to be there)

lanl-ompi:

devreal:
Ping :) Could someone please review and let me know if this needs more changes?

bosilca (approved):

devreal:
I squashed and updated the commit messages. @rajachan We might want to have that in v4.1.x as this fixes issues with the combination of Argobots/Qthreads and the UCX integration (see #7702).

devreal:
Oops, I realized that the Argobots integration is not part of 4.1.x. We don't need to port this one back then.

hppritcha (approved):

jjhursey:
Looks like some results from IBM CI failed to report back. Let's re-run.
bot:ibm:retest

hjelmn:
:+1: We can probably replace the release CPU function in opal_lifo (https://github.com/open-mpi/ompi/blob/master/opal/class/opal_lifo.h#L92) with this.

hjelmn (approved):

hppritcha:
Any objections to merging this PR?

yasushi-saito:
Can we get this merged? We really want this feature in

PR #8038, opened 2020-09-09
Use correct conditional variable initializer in opal/mca/pmix/base

devreal:
`opal_pmix_lock_t` has a member of type `opal_cond_t` so use `OPAL_CONDITION_STATIC_INIT` for static initialization instead of using `PTHREAD_COND_INITIALIZER` directly.

### DIFF
--- a/opal/mca/pmix/base/base.h
+++ b/opal/mca/pmix/base/base.h
@@ -46,7 +46,7 @@ typedef struct {
 
 extern opal_pmix_base_t opal_pmix_base;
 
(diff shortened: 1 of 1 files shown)


bosilca (approved):

jsquyres:
Does this need to be taken to any of the release branches?

devreal:
I believe this is only relevant when compiling with non-pthread support (Argobots/Qthreads) and we don't have that in any release yet.

jsquyres:
This seems like a no-brainer; I'm merging.

@rhc54 FYI.

PR #8043, opened 2020-09-14
mpif-h: fix a typo in MPI_Status_f2f08()

ggouaillardet:


jsquyres (approved):
Wow.  How did that escape CI?

ggouaillardet:
this only affects environments where weak symbols are not available (e.g. OS X), and I did not see OS X when I checked jenkins platforms.

Issue #8051, opened 2020-09-17
ofi_create_recv_tag mask hides ssend's "ack" bit


hkuno:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)
master branch top of tree: commit eca00a7a3b179

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
git clone

### If you are building/installing from a git clone, please copy-n-paste the output from `git submodule status`.
$ git submodule status
+299d2a489aa53546e1320eb3fd7e8d726f16b251 opal/mca/hwloc/hwloc2/hwloc (dev-3067-g299d2a4)
+ee72a2b65b1b6480753fc12d500c51ebe4fc23aa opal/mca/pmix/pmix4x/openpmix (v1.1.3-2505-gee72a2b)
+545863e6dc055233456116da6dc85be2b307f8e2 prrte (dev-30707-g545863e)

### Please describe the system on which you are running
N/A
* Operating system/version: 
* Computer hardware: 
* Network type: 

-----------------------------

## Details of the problem

The upper 2 bits of an ompi tag encode the synchronize send and synchronize send ack. Because the mtl_ofi_create_recv_tag_CQD and mtl_ofi_create_recv_tag functions both use ompi_mtl_ofi.sync_proto_mask instead of
ompi_mtl_ofi.sync_send when generating their "ignore" masks, they hide the ack bit, turning the tag into an "any tag receive"

This is an issue because ssend is implemented by doing a send and receive internally.  So if there happens to be an outstanding posted receive posted before the ssend, that receive will end up consuming the message intended for the ssend's internal receive.

Updating mtl_ofi_create_recv_tag_CQD and mtl_ofi_create_recv_tag functions to use ompi_mtl_ofi.sync_send fixes this.

For example, consider the following: 
```
    if (my_rank == 0) {
        MPI_CALL(MPI_Irecv, NULL, 0,MPI_CHAR, MPI_ANY_SOURCE, MPI_ANY_TAG,
                 MPI_COMM_WORLD, &rx_req);
        MPI_CALL(MPI_Ssend, NULL, 0, MPI_CHAR, 1, 1, MPI_COMM_WORLD);
        MPI_CALL(MPI_Wait, &rx_req, &rx_status);
        if (rx_status.MPI_TAG != 2)
[...]

PR #8052, opened 2020-09-17
use sync_send mask for ofi_create_recv_tag

hkuno:
This commit fixes issue https://github.com/open-mpi/ompi/issues/8051 .

The upper 2 bits of an ompi tag encode the synchronize send and synchronize send ack. Because the mtl_ofi_create_recv_tag_CQD and mtl_ofi_create_recv_tag functions both use ompi_mtl_ofi.sync_proto_mask instead of
ompi_mtl_ofi.sync_send when generating their "ignore" masks, the recv tag-matching logic will disregard the ack bit so that it may match a tag that has the ack bit set.

This is an issue because ssend is implemented by doing a send and receive internally. So if there happens to be an outstanding receive posted by a user before an ssend, that user's receive may end up consuming the internal message intended for the ssend's internal receive.

Updating mtl_ofi_create_recv_tag_CQD and mtl_ofi_create_recv_tag functions to use ompi_mtl_ofi.sync_send fixes this.

Authored-by: John L. Byrne <john.l.byrne@hpe.com>

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi_types.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi_types.h
@@ -160,7 +160,7 @@ __opal_attribute_always_inline__ static inline void
 mtl_ofi_create_recv_tag_CQD(uint64_t *match_bits, uint64_t *mask_bits,
                             int comm_id, int tag)
 {
-    *mask_bits  = ompi_mtl_ofi.sync_proto_mask;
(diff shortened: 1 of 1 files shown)


bwbarrett (approved):

hppritcha (approved):

lanl-ompi:

hkuno:
Thank you!

acgoldma:
Can this issue be backported to 4.0.x (and 4.1.x)?

PR #8053, opened 2020-09-18
opal/mca/threads/qthreads: Fix #8036

### LINKED ISSUE #8036: Argobots and Qthread configure detection need fixing/improvements

shintaro-iwasaki:
This PR fixes #8036. Specifically, this PR provides the following:
1. Add `--with-qthreads` (the same as Argobots, which has been already added by #7675. I'd like to thank @hppritcha )
2. Fix Qthreads compilation issues reported by @devreal in #8036
3. Implement several not-implemented features for MCA/threads/Qthreads backend.

This PR has been developed together with the Qthreads developers (@olivier-snl and @janciesko). I checked it with the latest Qthreads (you need the latest master branch).

Note that this PR does not affect the behavior of Open MPI by default; this change is effective only when one enables Qthreads.


devreal (changes requested):
Thanks for working on that :+1: Some small comments inline

jsquyres:
ok to test

opal/mca/threads/qthreads/Makefile.am
```diff
@@ -19,12 +19,41 @@
...
-    threads_qthreads_module.c
+    threads_qthreads_module.c \
+    threads_qthreads_mutex.c \
+    threads_qthreads_wait_sync.c \
+    threads_qthreads_mutex.h \
+    threads_qthreads_threads.h \
+    threads_qthreads_tsd.h \
+    threads_qthreads_wait_sync.h
+
+#lib = libmca_threads_qthreads.la
```
jsquyres, line 36: Nit: no need to commit commented-out code.

opal/mca/threads/qthreads/Makefile.am
```diff
@@ -19,12 +19,41 @@
...
+mca_threads_qthreads_la_SOURCES = $(sources)
+mca_threads_qthreads_la_LDFLAGS = -module -avoid-version
+mca_threads_qthreads_la_LIBADD = $(top_builddir)/opal/lib@OPAL_LIB_PREFIX@open-pal.la \
+        $(opal_qthreads_LIBS)
+
+noinst_LTLIBRARIES = $(component_noinst)
+libmca_threads_qthreads_la_SOURCES = $(sources)
+libmca_threads_qthreads_la_LDFLAGS = -module -avoid-version $(opal_qthreads_LDFLAGS)
+libmca_threads_qthreads_la_LIBADD = $(opal_qthreads_LIBS)
+
```
jsquyres, line 58: Nit: blank lines at the end of the file.

opal/mca/threads/qthreads/configure.m4
```diff
@@ -25,13 +25,67 @@
...
+    AS_IF([test $opal_qthreads_happy = yes && test -n "$opal_qthreads_dir"],
+          [OPAL_QTHREADS_INCLUDE_PATH="$opal_qthreads_dir/include/"],
+          [OPAL_QTHREADS_INCLUDE_PATH=""])
+
+    AS_IF([test $opal_qthreads_happy = yes],
+          [AC_CONFIG_FILES([opal/mca/threads/qthreads/threads_qthreads.h])
+           AC_SUBST([OPAL_QTHREADS_INCLUDE_PATH])
+           AC_SUBST([opal_qthreads_CPPFLAGS])
+           AC_SUBST([opal_qthreads_LDFLAGS])
+           AC_SUBST([opal_qthreads_LIBS])
```
jsquyres, line 75: I believe that AC_CONFIG_FILES and AC_SUBST are always executed, regardless of the result of the AS_IF.  I.e., these should probably be outside the AS_IF so as not to imply that they are actually conditionally executed.

opal/mca/threads/qthreads/configure.m4
```diff
@@ -60,6 +114,21 @@ AC_DEFUN([MCA_opal_threads_qthreads_POST_CONFIG],[
...
+           THREAD_CPPFLAGS="$TPKG_CPPFLAGS"
+           THREAD_CXXCPPFLAGS="$TPKG_CXXCPPFLAGS"
+           THREAD_LDFLAGS="$TPKG_LDFLAGS"
+           THREAD_LIBS="$TPKG_LIBS"
+           AC_SUBST(THREAD_CFLAGS)
+           AC_SUBST(THREAD_FCFLAGS)
+           AC_SUBST(THREAD_CXXFLAGS)
+           AC_SUBST(THREAD_CPPFLAGS)
+           AC_SUBST(THREAD_LDFLAGS)
+           AC_SUBST(THREAD_LIBS)
```
jsquyres, line 129: Same as above.

I didn't check the `threads` framework, but it's a little odd that a component is responsible for AC_SUBST'ing `TPKG_*` and `THREAD_*` values.  Specifically:

1. Are both `TPKG_*` and `THREAD_*` used by the framework?
1. If so, why isn't the framework doing the `AC_SUBST` (vs. the components)?

opal/mca/threads/qthreads/threads_qthreads_module.c
```diff
@@ -34,40 +29,82 @@ struct opal_tsd_key_value {
     opal_tsd_destructor_t destructor;
 };
 
-static int opal_main_thread;
-struct opal_tsd_key_value *opal_tsd_key_values = NULL;
-static int opal_tsd_key_values_count = 0;
+/* 0: uninitialized, 1: initialized. */
+static opal_atomic_lock_t opal_thread_self_key_lock = OPAL_ATOMIC_LOCK_INIT;
+static int opal_thread_self_key_init = 0;
```
jsquyres, line 34: Nit: is there a reason to use `int` here instead of `bool`?

opal/mca/threads/qthreads/threads_qthreads_mutex.c
```diff
@@ -56,36 +58,81 @@ OBJ_CLASS_INSTANCE(opal_recursive_mutex_t,
...
+    while (1) {
+        opal_mutex_unlock(lock);
+        qthread_yield();
+        opal_mutex_lock(lock);
+        /* Check if someone woke me up. */
+        opal_atomic_lock(&cond->m_lock);
+        int signaled = waiter.m_signaled;
+        opal_atomic_unlock(&cond->m_lock);
+        if (1 == signaled)
+            break;
```
jsquyres, line 98: Nit: please use `{}` around all blocks, even 1-line blocks.

opal/mca/threads/qthreads/threads_qthreads_module.c
```diff
@@ -34,40 +29,82 @@ struct opal_tsd_key_value {
...
 }
 
-bool opal_thread_self_compare(opal_thread_t *t)
+opal_thread_t *opal_thread_get_self(void)
 {
-    return OPAL_ERR_NOT_IMPLEMENTED;
+    opal_threads_ensure_init_qthreads();
+    opal_thread_t *t = OBJ_NEW(opal_thread_t);
+    t->t_thread_ret_ptr = opal_thread_get_qthreads_self();
+    return NULL;
```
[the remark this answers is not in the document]
shintaro-iwasaki: Thank you for reviewing. I fixed it.

opal/mca/threads/qthreads/threads_qthreads_module.c
```diff
@@ -34,40 +29,82 @@ struct opal_tsd_key_value {
...
+
+    t->t_ret = ((void *(*)(void *))t->t_run)(t);
+    return 0;
 }
 
-bool opal_thread_self_compare(opal_thread_t *t)
+opal_thread_t *opal_thread_get_self(void)
 {
-    return OPAL_ERR_NOT_IMPLEMENTED;
+    opal_threads_ensure_init_qthreads();
```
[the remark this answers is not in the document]
shintaro-iwasaki: You are right. I removed it. This module relies on lazy initialization, so I did not need to initialize Qthreads here.

opal/mca/threads/qthreads/threads_qthreads_mutex.c
```diff
@@ -56,36 +58,81 @@ OBJ_CLASS_INSTANCE(opal_recursive_mutex_t,
...
+    opal_atomic_unlock(&cond->m_lock);
+
+    while (1) {
+        opal_mutex_unlock(lock);
+        qthread_yield();
+        opal_mutex_lock(lock);
+        /* Check if someone woke me up. */
+        opal_atomic_lock(&cond->m_lock);
+        int signaled = waiter.m_signaled;
+        opal_atomic_unlock(&cond->m_lock);
```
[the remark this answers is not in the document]
shintaro-iwasaki: This lock (or any memory barrier that can order accesses to `m_signaled`) is necessary.
`m_signaled` can be updated by another thread that runs on another worker (=Pthreads), but in my understanding, the C standard does not guarantee when a reader can see `m_signaled` updated by a writer without atomic operations.

Since this `waiter` object is allocated in a function stack, it can cause SEGV if the writer accesses this object after `m_signaled` is read by a reader.  It should not happen in `opal_cond_signal()` if all the memory write operations are issued in the order written in the code.

If acquire-release or seq-cst load/store can remove this lock, but it makes the code a bit more complicated.

opal/mca/threads/qthreads/threads_qthreads_wait_sync.c
```diff
@@ -0,0 +1,116 @@
...
+static opal_atomic_int32_t num_thread_in_progress = 0;
+
+#define WAIT_SYNC_PASS_OWNERSHIP(who)                  \
+    do {                                               \
+        opal_mutex_lock(&(who)->lock);                 \
+        opal_cond_signal(&(who)->condition);           \
+        opal_mutex_unlock(&(who)->lock);               \
+    } while (0)
+
+int ompi_sync_wait_mt(ompi_wait_sync_t *sync)
```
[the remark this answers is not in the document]
shintaro-iwasaki: Basically both Qthreads and Argobots "copy" the implementation of Pthreads, so once #8037 is merged, maybe we can declare them in a common header.

This unification needs some work, though.  Specifically, this structure (`ompi_wait_sync_t`) is different https://github.com/open-mpi/ompi/blob/master/opal/mca/threads/pthreads/threads_pthreads_wait_sync.h#L30-L38.  To abstract these operations, we need to change the Pthreads code, which needs more careful examination.

opal/mca/threads/qthreads/configure.m4
```diff
@@ -25,13 +25,67 @@
...
+                 [opal_qthreads_dir=$with_qthreads
+                  AC_MSG_RESULT([($opal_qthreads_dir)])],
+                 [AC_MSG_RESULT([(default search paths)])])
+           AS_IF([test ! -z "$with_qthreads_libdir" && \
+                         test "$with_qthreads_libdir" != "yes"],
+                 [opal_qthreads_libdir=$with_qthreads_libdir])
+          ])
+
+    AS_IF([test $opal_qthreads_happy = yes],
+          [OPAL_CHECK_PACKAGE([opal_qthreads],
```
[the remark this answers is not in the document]
shintaro-iwasaki: Thanks.  It relies on the latest master, but this change cannot be observed at a C level.  Specifically, the latest Qthreads changes the way of "#include", which is hard to detect programmatically.  Symbol/version check is promising once Qthreads has an official release that includes this update.

opal/mca/threads/qthreads/Makefile.am
shintaro-iwasaki: Thank you for reviewing.  I removed it.

opal/mca/threads/qthreads/Makefile.am
shintaro-iwasaki: I removed the extra line.

opal/mca/threads/qthreads/configure.m4
shintaro-iwasaki: I moved them outside `AS_IF()`.

> AC_CONFIG_FILES and AC_SUBST are always executed

Yes they seem so. I did not know that. Thank you!

opal/mca/threads/qthreads/configure.m4
shintaro-iwasaki: I moved `AC_SUBST()` outside `AS_IF()`.

This implementation is basically a copy from #7675.

1. `TPKG_*` is used only in `opal/mca/thread`. `THREAD_` is used in the top `configure.ac` (https://github.com/open-mpi/ompi/blob/master/configure.ac#L1114-L1117).
2. I believe it is to propagate this "thread" information to the top `configure`, but I do not fully understand the modularity/design of Open MPI. @hppritcha would you have any idea about this?

opal/mca/threads/qthreads/threads_qthreads_module.c
shintaro-iwasaki: Thank you. `bool` is used here in the revised version.

opal/mca/threads/qthreads/threads_qthreads_mutex.c
shintaro-iwasaki: I fixed it.  I checked all of my changes and I believe I fixed all of one-liner `if`/`for`/`while`/....

opal/mca/threads/qthreads/threads_qthreads_module.c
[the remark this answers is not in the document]
shintaro-iwasaki: For this line, I need `opal_threads_ensure_init_qthreads()` since a static internal function, `opal_thread_get_qthreads_self()`, calls a Qthreads function.

opal/mca/threads/qthreads/threads_qthreads_mutex.c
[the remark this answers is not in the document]
devreal: Ahh I see what you mean. The problem is that you are issuing a lot of atomic memory operations, esp. if `qthread_yield` is a no-op (i.e., if there are no other qthreads available for execution). In that short snippet, it's four atomic memory ops already. If you have multiple threads (not qthreads) waiting on this conditional variable these locks will be quite contended. Plus, `opal_mutex_unlock` calls `qthread_yield` again (which I am also not convinced is necessary).

Anyway, what is the `lock` argument in this function good for? It doesn't seem to protect anything so it can be re-taken right before exiting the function (the `cond` has its own lock). If you make `waiter.m_signaled` `volatile` (to prevent the compiler from optimizing away the loads) and insert a memory barrier (`opal_atomic_wmb()` should do) in the signalling code before setting `m_signaled` then the tail of the function could look like this:

```C
    while (1) {
        qthread_yield();
        /* Check if someone woke me up. */
        if (waiter.m_signaled) {
            break;
        }
    }
    opal_mutex_lock(lock);
    return OPAL_SUCCESS;
```

Yes, the signalling will become a bit more expensive due to the memory barrier but having several threads hammering on atomic locks is by far more problematic than a few more cycles spent by one thread in the signalling code.

opal/mca/threads/qthreads/threads_qthreads_module.c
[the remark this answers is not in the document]
devreal: What is the cost of qthread_initialize()? I wonder if it's worth calling it on each and every operation. Sooner or later qhtreads will be initialized if this module is used so why not make sure it is properly initialized from the beginning

opal/mca/threads/qthreads/threads_qthreads_mutex.c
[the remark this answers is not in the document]
olivier-snl: Indeed this type of polling is probably not the best use of Qthreads.  Using FEBs instead would make for an event-driven approach.  Not sure if its best to leave it for now and reimplement later or go ahead and fix now. @janciesko

opal/mca/threads/qthreads/threads_qthreads_module.c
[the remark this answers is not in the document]
olivier-snl: Reinitializing Qthreads on every operation is not cheap and it would basically result in not being able to take advantage of the benefits of using a user-level threading library rather than pthreads.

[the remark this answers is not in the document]
shintaro-iwasaki: In any case, since there's no good timing to initialize Qthreads (specifically, it is hard to forbid developers from using mutex before initializing Qthreads and after finalizing Qthreads), this check is necessary. (MPICH took another approach that is ideal for ULTs; as far as I remember, all the components in MPICH must be aware of from when to when mutex/TLS can be used, but it needs tremendous engineering efforts to maintain.)

In terms of cost, I believe `qthread_initialize()` will take this `redundant call` path, so it should not be very expensive (of course, calling an external function needs some cycles).
https://github.com/Qthreads/qthreads/blob/master/src/qthread.c#L814-L816

Maybe I miss something, however. If this function calling overhead would kill the benefits of using user-level threads, I can introduce another flag to check if Qthread has been already initialized.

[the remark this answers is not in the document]
olivier-snl: Yes, I forgot that we have that check for already being initialized.

shintaro-iwasaki:
I have updated the commits to address problems pointed out by @devreal and @jsquyres. Could someone please review this PR again?

devreal (approved):
I think the PR is OK. The performance issues in the conditional variable implementation is unresolved but that is not a hill I will die on.

shintaro-iwasaki:
@jsquyres Could you please take a look at this PR? Is the issue of `TPKG_*` and `THREAD_*` blocking this PR?

opal/mca/threads/qthreads/configure.m4
jsquyres: > This implementation is basically a copy from #7675.

Understood.  I guess this was missed in that PR.  ☹️ 

I guess my question still stands: if `TPKG_*` and `THREAD_*` are being used by the framework (and higher -- all the way up in the top-level `configure.ac`), why is it up to the component to `AC_SUBST` them?  That seems weird.

Can you add a commit on there to remove the `AC_SUBSTs` of these values from the components, and instead, do them in `opal/mca/threads/configure.m4`?

olivier-snl:
> I think the PR is OK. The performance issues in the conditional variable implementation is unresolved but that is not a hill I will die on.

The Qthreads team can spend some time working out a better performing solution in the near future.  The changes by Shintaro at least give us something correct for now.

opal/mca/threads/qthreads/configure.m4
shintaro-iwasaki: I added a commit, which fixed the issue (I also rebased my branch to the master):
- `TPKG_*` are not `AC_SUBSTs`; they appear only in `opal/mca/threads/<THREAD_PKG_NAME>/configure.m4`.
- `THREAD_*` are `AC_SUBSTs`. in `opal/mca/threads/configure.m4`.

As far as I checked, there is no compilation issue with Pthreads, Qthreads, and Argobots.

jsquyres (approved):

shintaro-iwasaki:
I'd like to thank @devreal and @jsquyres for reviewing this PR! Could someone merge this PR?

Since this incomplete MCA/thread/Qthread has been in the main branch of Open MPI, we'd be extremely happy if this PR is merged before the upcoming 5.0 release. (Basically this PR does not affect the existing Pthreads implementation.)
If this PR should be rebased, more tested, or checked by additional reviewers, please let me know.

jsquyres:
Done!  Thanks for all the work on this!

PR #8054, opened 2020-09-18
patcher: remove the linux component

hjelmn:
The Linux component was an attempt to hook calls by patching the dynamic
symbol table. It, unfortunately, does not work as it will always miss
calls made internally by glibc. For example, it might catch a user call
directly to munmap but will miss the chain free -> munmap. Since the
later is the common case we were trying to hook this made the component
unusable. This PR finally kills the component.


bwbarrett (approved):

awlauria:
bot:ompi:retest

PR #8057, opened 2020-09-22
MPI_T Events

cchambreau:
This PR includes Nathan Hjelm's reworking of PERUSE functionality to implement the MPI_T Events API as specified in the MPI 4.0 proposal. Subsequent minor changes to the MPI_T Events API were made and included here. The API and an initial set of events were tested against Marc-André Hermann's MEL tracing tool and other test cases. History, a description of the API and experience with tools can be found in "Enabling callback-driven runtime introspection via MPI_T."

### DIFF
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.c
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.c
@@ -13,7 +13,7 @@
  * Copyright (c) 2008      UT-Battelle, LLC. All rights reserved.
  * Copyright (c) 2011      Sandia National Laboratories. All rights reserved.
  * Copyright (c) 2012-2015 NVIDIA Corporation.  All rights reserved.
- * Copyright (c) 2011-2017 Los Alamos National Security, LLC. All rights
+ * Copyright (c) 2011-2018 Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2012      FUJITSU LIMITED.  All rights reserved.
  * Copyright (c) 2014-2016 Research Organization for Information Science
@@ -77,8 +77,9 @@ static int mca_pml_ob1_recv_request_free(struct ompi_request_t** request)
     assert (false == recvreq->req_recv.req_base.req_free_called);
 
     recvreq->req_recv.req_base.req_free_called = true;
-    PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_NOTIFY,
-                             &(recvreq->req_recv.req_base), PERUSE_RECV );
+    mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_REQUEST_FREE].event,
+                          MCA_BASE_CB_REQUIRE_MPI_RESTRICTED,
+                          recvreq->req_recv.req_base.req_comm, NULL, &recvreq);
 
     if( true == recvreq->req_recv.req_base.req_pml_complete ) {
         /* make buffer defined when the request is compeleted,
@@ -121,8 +122,10 @@ static int mca_pml_ob1_recv_request_cancel(struct ompi_request_t* ompi_request, 
         opal_list_remove_item(&proc->specific_receives, (opal_list_item_t*)request);
     }
 #endif
-    PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_REMOVE_FROM_POSTED_Q,
-                             &(request->req_recv.req_base), PERUSE_RECV );
+
+    void *req = &(request->req_recv.req_base);
+    mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_RECEIVE_CANCELED].event,
+                          MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &req);
     /**
      * As now the PML is done with this request we have to force the pml_complete
      * to true. Otherwise, the request will never be freed.
@@ -411,6 +414,7 @@ static void mca_pml_ob1_rget_completion (mca_btl_base_module_t* btl, struct mca_
 static int mca_pml_ob1_recv_request_put_frag (mca_pml_ob1_rdma_frag_t *frag)
 {
     mca_pml_ob1_recv_request_t *recvreq = (mca_pml_ob1_recv_request_t *) frag->rdma_req;
+    ompi_communicator_t *comm = recvreq->req_recv.req_base.req_comm;
 #if OPAL_ENABLE_HETEROGENEOUS_SUPPORT
     ompi_proc_t* proc = (ompi_proc_t*)recvreq->req_recv.req_base.req_proc;
 #endif
@@ -450,9 +454,10 @@ static int mca_pml_ob1_recv_request_put_frag (mca_pml_ob1_rdma_frag_t *frag)
 
     recvreq->req_ack_sent = true;
 
-    PERUSE_TRACE_COMM_OMPI_EVENT( PERUSE_COMM_REQ_XFER_CONTINUE,
-                                  &(recvreq->req_recv.req_base), frag->rdma_length,
-                                  PERUSE_RECV);
+    mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_TRANSFER].event,
+                          MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL,
+                          &((mca_pml_ob1_transfer_event_t) {.request = recvreq, .length = frag->rdma_length}));
+
 
     /* send rdma request to peer */
     rc = mca_bml_base_send (bml_btl, ctl, MCA_PML_OB1_HDR_TYPE_PUT);
@@ -491,9 +496,9 @@ int mca_pml_ob1_recv_request_get_frag (mca_pml_ob1_rdma_frag_t *frag)
         local_handle = recvreq->local_handle;
     }
 
-    PERUSE_TRACE_COMM_OMPI_EVENT(PERUSE_COMM_REQ_XFER_CONTINUE,
-                                 &(((mca_pml_ob1_recv_request_t *) frag->rdma_req)->req_recv.req_base),
-                                 frag->rdma_length, PERUSE_RECV);
+    mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_TRANSFER].event,
+                          MCA_BASE_CB_REQUIRE_ASYNC_SIGNAL_SAFE, recvreq->req_recv.req_base.req_comm, NULL,
+                          &((mca_pml_ob1_transfer_event_t){.request = frag->rdma_req, .length = frag->rdma_length}));
 
     /* queue up get request */
     rc = mca_bml_base_get (bml_btl, frag->local_address, frag->remote_address, local_handle,
@@ -507,9 +512,6 @@ int mca_pml_ob1_recv_request_get_frag (mca_pml_ob1_rdma_frag_t *frag)
 
     return OMPI_SUCCESS;
 }
-
-
-
 
 /*
  * Update the recv request status to reflect the number of bytes
@@ -1086,16 +1088,15 @@ static inline void append_recv_req_to_queue(opal_list_t *queue,
 {
     opal_list_append(queue, (opal_list_item_t*)req);
 
-#if OMPI_WANT_PERUSE
     /**
      * We don't want to generate this kind of event for MPI_Probe.
      */
     if (req->req_recv.req_base.req_type != MCA_PML_REQUEST_PROBE &&
         req->req_recv.req_base.req_type != MCA_PML_REQUEST_MPROBE) {
-        PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_REQ_INSERT_IN_POSTED_Q,
-                                &(req->req_recv.req_base), PERUSE_RECV);
+        ompi_communicator_t *comm = req->req_recv.req_base.req_comm;
+        mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_POSTED_INSERT].event,
+                              MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &req);
     }
-#endif
 }
 
 /*
@@ -1258,8 +1259,8 @@ void mca_pml_ob1_recv_req_start(mca_pml_ob1_recv_request_t *req)
      * The laps of time between the ACTIVATE event and the SEARCH_UNEX one include
      * the cost of the request lock.
      */
-    PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_SEARCH_UNEX_Q_BEGIN,
-                            &(req->req_recv.req_base), PERUSE_RECV);
+    mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_SEARCH_UNEX_BEGIN].event,
+                          MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &req);
 
     /* assign sequence number */
     req->req_recv.req_base.req_sequence = ob1_comm->recv_sequence++;
@@ -1297,8 +1298,8 @@ void mca_pml_ob1_recv_req_start(mca_pml_ob1_recv_request_t *req)
     }
 
     if(OPAL_UNLIKELY(NULL == frag)) {
-        PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_SEARCH_UNEX_Q_END,
-                                &(req->req_recv.req_base), PERUSE_RECV);
+        mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_SEARCH_UNEX_END].event,
+                              MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &req);
         /* We didn't find any matches.  Record this irecv so we can match
            it when the message comes in. */
         if(OPAL_LIKELY(req->req_recv.req_base.req_type != MCA_PML_REQUEST_IPROBE &&
@@ -1314,18 +1315,13 @@ void mca_pml_ob1_recv_req_start(mca_pml_ob1_recv_request_t *req)
         OB1_MATCHING_UNLOCK(&ob1_comm->matching_lock);
     } else {
         if(OPAL_LIKELY(!IS_PROB_REQ(req))) {
-            PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_REQ_MATCH_UNEX,
-                                    &(req->req_recv.req_base), PERUSE_RECV);
-
             hdr = (mca_pml_ob1_hdr_t*)frag->segments->seg_addr.pval;
-            PERUSE_TRACE_MSG_EVENT(PERUSE_COMM_MSG_REMOVE_FROM_UNEX_Q,
-                                   req->req_recv.req_base.req_comm,
-                                   hdr->hdr_match.hdr_src,
-                                   hdr->hdr_match.hdr_tag,
-                                   PERUSE_RECV);
 
-            PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_SEARCH_UNEX_Q_END,
-                                    &(req->req_recv.req_base), PERUSE_RECV);
+            mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_UNEX_REMOVE].event,
+                                  MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &hdr->hdr_match);
+
+            mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_SEARCH_UNEX_END].event,
+                                  MCA_BASE_CB_REQUIRE_MPI_RESTRICTED, comm, NULL, &req);
 
 #if MCA_PML_OB1_CUSTOM_MATCH
             custom_match_umq_remove_hold(req->req_recv.req_base.req_comm->c_pml_comm->umq, hold_prev, hold_elem, hold_index);
--- a/ompi/mca/pml/ob1/pml_ob1_recvreq.h
+++ b/ompi/mca/pml/ob1/pml_ob1_recvreq.h
@@ -124,8 +124,9 @@ do {                                                                \
  */
 #define MCA_PML_OB1_RECV_REQUEST_MPI_COMPLETE( recvreq )                              \
     do {                                                                              \
-        PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_COMPLETE,                            \
-                                 &(recvreq->req_recv.req_base), PERUSE_RECV );        \
+        mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_REQUEST_COMPLETE].event, \
+                              MCA_BASE_CB_REQUIRE_ASYNC_SIGNAL_SAFE,             \
+                              (recvreq)->req_recv.req_base.req_comm, NULL, &(recvreq)); \
         ompi_request_complete( &(recvreq->req_recv.req_base.req_ompi), true );        \
     } while (0)
 
@@ -162,8 +163,9 @@ recv_request_pml_complete(mca_pml_ob1_recv_request_t *recvreq)
     if(false == recvreq->req_recv.req_base.req_pml_complete){
 
         if(recvreq->req_recv.req_bytes_packed > 0) {
-            PERUSE_TRACE_COMM_EVENT( PERUSE_COMM_REQ_XFER_END,
-                    &recvreq->req_recv.req_base, PERUSE_RECV );
+            mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_TRANSFER_END].event,
+                                  MCA_BASE_CB_REQUIRE_ASYNC_SIGNAL_SAFE,
+                                  recvreq->req_recv.req_base.req_comm, NULL, &recvreq);
         }
 
         for(i = 0; i < recvreq->req_rdma_cnt; i++) {
@@ -258,8 +260,9 @@ static inline void recv_req_matched(mca_pml_ob1_recv_request_t *req,
             prepare_recv_req_converter(req);
         }
 #endif  /* OPAL_ENABLE_HETEROGENEOUS_SUPPORT */
-        PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_REQ_XFER_BEGIN,
-                                &req->req_recv.req_base, PERUSE_RECV);
+        mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_TRANSFER_BEGIN].event,
+                              MCA_BASE_CB_REQUIRE_ASYNC_SIGNAL_SAFE,
+                              req->req_recv.req_base.req_comm, NULL, &req);
     }
 }
 
@@ -268,7 +271,7 @@ static inline void recv_req_matched(mca_pml_ob1_recv_request_t *req,
  *
  */
 
-#define MCA_PML_OB1_RECV_REQUEST_UNPACK( request,                                 \
+#define MCA_PML_OB1_RECV_REQUEST_UNPACK( req,                                     \
                                          segments,                                \
                                          num_segments,                            \
                                          seg_offset,                              \
@@ -277,7 +280,7 @@ static inline void recv_req_matched(mca_pml_ob1_recv_request_t *req,
                                          bytes_delivered)                         \
 do {                                                                              \
     bytes_delivered = 0;                                                          \
-    if(request->req_recv.req_bytes_packed > 0) {                                  \
+    if((req)->req_recv.req_bytes_packed > 0) {                                    \
         struct iovec iov[MCA_BTL_DES_MAX_SEGMENTS];                               \
         uint32_t iov_count = 0;                                                   \
         size_t max_data = bytes_received;                                         \
@@ -295,18 +298,21 @@ do {                                                                            
                 offset = 0;                                                       \
             }                                                                     \
         }                                                                         \
-        OPAL_THREAD_LOCK(&request->lock);                                         \
-        PERUSE_TRACE_COMM_OMPI_EVENT (PERUSE_COMM_REQ_XFER_CONTINUE,              \
-                                      &(request->req_recv.req_base), max_data,    \
-                                      PERUSE_RECV);                               \
-        opal_convertor_set_position( &(request->req_recv.req_base.req_convertor), \
+        OPAL_THREAD_LOCK(&(req)->lock);                                           \
+                                                                                  \
+        mca_base_event_raise (mca_pml_ob1_events[MCA_PML_OB1_EVENT_TRANSFER].event, \
+                              MCA_BASE_CB_REQUIRE_ASYNC_SIGNAL_SAFE,              \
+                              (req)->req_recv.req_base.req_comm, NULL,            \
+            &((mca_pml_ob1_transfer_event_t){.request = req, .length = max_data})); \
+                                                                                  \
+        opal_convertor_set_position( &((req)->req_recv.req_base.req_convertor),   \
                                      &data_offset );                              \
-        opal_convertor_unpack( &(request)->req_recv.req_base.req_convertor,       \
+        opal_convertor_unpack( &(req)->req_recv.req_base.req_convertor,           \
                                iov,                                               \
                                &iov_count,                                        \
                                &max_data );                                       \
         bytes_delivered = max_data;                                               \
-        OPAL_THREAD_UNLOCK(&request->lock);                                       \
+        OPAL_THREAD_UNLOCK(&(req)->lock);                                         \
     }                                                                             \
 } while (0)
 
--- a/ompi/mpi/tool/event_callback_get_info.c
+++ b/ompi/mpi/tool/event_callback_get_info.c
@@ -1,0 +1,52 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2012-2013 Los Alamos National Security, LLC. All rights
+ *                         reserved.
+ * Copyright (c) 2014 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2017      IBM Corporation. All rights reserved.
+ * Copyright (c) 2019      Google, LLC. All rights reserved.
+ * Copyright (c) 2019      Triad National Security, LLC. All rights
+ *                         reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+#include "ompi/mpi/tool/mpit-internal.h"
+
+#if OPAL_HAVE_WEAK_SYMBOLS && OMPI_PROFILING_DEFINES
+#pragma weak MPI_T_event_callback_get_info = PMPI_T_event_callback_get_info
+#endif
+
+#if OMPI_PROFILING_DEFINES
+#include "ompi/mpi/tool/profile/defines.h"
+#endif
+
+
+int MPI_T_event_callback_get_info (MPI_T_event_registration event_registration,
+                                   MPI_T_cb_safety cb_safety, MPI_Info *info_used)
+{
+    ompi_info_t *info;
+    int ret;
+
+    if (!mpit_is_initialized ()) {
+        return MPI_T_ERR_NOT_INITIALIZED;
+    }
+
+    info = OBJ_NEW(ompi_info_t);
+    if (NULL == info) {
+        return ompit_opal_to_mpit_error(OMPI_ERR_OUT_OF_RESOURCE);
+    }
+
+    /* mca_base_cb_safety_t and MPI_T_cb_safety must be kept in sync for this to work */
+    ret = mca_base_event_callback_get_info (event_registration, (mca_base_cb_safety_t) cb_safety, &info->super);
+    if (OPAL_SUCCESS != ret) {
+        OBJ_RELEASE(info);
+    } else {
+        *info_used = info;
+    }
+
+    return ompit_opal_to_mpit_error(ret);
+}
--- a/ompi/mpi/tool/event_get_info.c
+++ b/ompi/mpi/tool/event_get_info.c
@@ -61,7 +61,30 @@ int MPI_T_event_get_info (int event_index, char *name, int *name_len,
         mpit_copy_string (name, name_len, event->event_name);
         mpit_copy_string (desc, desc_len, event->event_description);
 
-        max_datatypes = num_elements ? (*num_elements < (int) (event->event_datatype_count + 1) ? *num_elements - 1 : event->event_datatype_count) : 0;
+        // num_elements is INOUT
+        //
+        // Can query number of datatypes, returned in num_elements
+        //   if array_of_datatypes or displacements are NULL, just return data_type_count.
+        // Otherwise, if array_of_datatypes or displacements are not NULL, use num_elements
+        //   as maximum datatypes or displacements returned.
+        //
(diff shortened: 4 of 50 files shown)


jsquyres:
ok to test

gpaulsen:
@cchambreau, Hi.  Is this desirable for Open MPI v5.0 (not MPI-4.0 compliant)?  If so please, rebase it and get it into master and then cherry-pick to v5.0.x branch after it's in master.

ompi/mpi/tool/event_get_info.c
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        if (NULL != extent) {
+            *extent = event->event_extent;
+        }
+
+        if (NULL != info) {
+            *info = OBJ_NEW(ompi_info_t);
+        }
+    } while (0);
```
jsquyres, line 135: Is there a reason to put this in a `do { ... } while(0)`?  I've usually seen that kind of pattern inside preprocessor macros.

jsquyres (commented):
Can you rebase and fix all current conflicts?

Do we have any tests for this MPI_T event functionality?

hppritcha:
@cchambreau could you please rebase?  thanks.

hppritcha:
Howard will rebase and merge unless objections are raised.

opal/mca/base/mca_base_event.c
```diff
@@ -0,0 +1,648 @@
...
+int mca_base_event_register (const char *project, const char *framework, const char *component, const char *name,
+                             const char *description, mca_base_var_info_lvl_t verbosity, opal_datatype_t **datatypes,
+                             unsigned long *offsets, size_t num_datatypes, mca_base_var_enum_t *enumerator, int extent, int bind,
+                             int source, uint32_t flags, mca_base_notify_fn_t notify, void *ctx, mca_base_event_t **event_out)
+{
+    int ret, group_index;
+    mca_base_event_t *event;
+
+    /* ensure the caller did not set an invalid flag */
+    assert (!(flags & 0x3f));
```
jsquyres, line 190: What is 0x3f?  If there's a specific enum that should be used, use that enum -- a hard-coded hex value here seems dangerous (even if you use a `(FLAG_MAX-1)` kind of value).

opal/mca/base/mca_base_event.c
```diff
@@ -0,0 +1,648 @@
...
+
+    /* update this assert if more MPIT verbosity levels are added */
+    assert (verbosity >= OPAL_INFO_LVL_1 && verbosity <= OPAL_INFO_LVL_9);
+
+
```
jsquyres, line 195: Nit: remove extra blank line.

opal/mca/base/mca_base_source.c
```diff
@@ -0,0 +1,238 @@
...
+int mca_base_source_default_source = -1;
+
+static uint64_t mca_base_source_default_time_source (void)
+{
+    uint64_t time_value;
+
+#if OPAL_HAVE_CLOCK_GETTIME
+    struct timespec current;
+
+    clock_gettime (CLOCK_MONOTONIC, &current);
```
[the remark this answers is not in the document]
jsquyres: A larger question: given that MPI_WTICK() and MPI_TIME() exist, should MPI_T events use the same back-end time handling?  And if so, should that be based on the OPAL timer framework?

For the second question, it looks like Open MPI has a complicated history regarding time/Wtime/Wtick/the OPAL timer framework.  See, for example:

* https://github.com/open-mpi/ompi/issues/3003
* https://github.com/open-mpi/ompi/wiki/Timers

There was a lot of discussion about RTDSC timers, frequency scaling, whether RTDSC was monotonic, etc.  Currently (Dec 2021), Wtick/Wtime:

1. On Linux, use `clock_gettime()` (not the OPAL timer framework)
2. Use `gettimeofday()` everywhere else.

Is there a reason we're not using `clock_gettime()` everywhere it's available?  (and potentially `CLOCK_MONOTONIC_RAW` if it's available)?  This may be a naive question -- I haven't done any perf testing of `CLOCK_MONOTONIC_RAW` on Linux, for example.

FWIW, the RHEL 6 `clock_gettime(2)` says this:

```
       CLOCK_MONOTONIC
              Clock  that  cannot be set and represents monotonic time since
              some unspecified starting point.  This clock is  not  affected
              by discontinuous jumps in the system time (e.g., if the system
              administrator manually changes the clock), but is affected  by
              the incremental adjustments performed by adjtime(3) and NTP.

       CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
              Similar to CLOCK_MONOTONIC, but provides access to a raw hard-
              ware-based  time that is not subject to NTP adjustments or the
              incremental adjustments performed by adjtime(3).
```

which directly states that `CLOCK_MONOTONIC` is affected by `adjtime(3)` and NTP, which means we probably shouldn't be using it anywhere we expect actually-monotonic time.  Instead, it seems we should be using `CLOCK_MONOTONIC_RAW`...?  That being said, I don't know if `CLOCK_MONOTONIC_RAW` is subject to a per-hardware-core value, and may change if your process' locality changes to a different core.  This would need to be verified.

ibm-ompi:

 Gist: https://gist.github.com/d182f369f64bb66a239cbf4b56b1e3ce

ibm-ompi:

 Gist: https://gist.github.com/20f1009df186bd52f322544e39a3595c

jsquyres:
@cchambreau can you fix the conflicts on this PR?  Thanks!

jsquyres (changes requested):
In addition to the other comments on this PR, please update this PR to use `opal_clock_gettime()` (and, if necessary, `opal_clock_getres()`) from #9798.

hppritcha:
bot:ompi:retest

opal/mca/base/Makefile.am
```diff
@@ -39,7 +41,10 @@ headers = \
 	mca_base_var_enum.h \
         mca_base_var_group.h \
         mca_base_vari.h \
-	mca_base_framework.h
+	mca_base_framework.h \
+        mca_base_event.h \
+        mca_base_source.h
+
```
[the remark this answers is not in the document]
hppritcha: should be fixed now

opal/mca/base/mca_base_source.c
```diff
@@ -0,0 +1,238 @@
...
+/* mca_base_source_t class */
+static void mca_base_source_contructor (mca_base_source_t *source)
+{
+    memset ((char *) source + sizeof (source->super), 0, sizeof (*source) - sizeof (source->super));
+}
+
+static void mca_base_source_destructor (mca_base_source_t *source)
+{
+    free (source->source_name);
+    free (source->source_description);
```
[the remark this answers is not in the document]
hppritcha: should be fixed now.

opal/mca/base/mca_base_var_group.c
```diff
@@ -484,7 +503,37 @@ int mca_base_var_group_add_enum(const int group_index, const void *storage)
...
+            return i;
+        }
+    }
+
+    if (OPAL_SUCCESS !=
+        (ret = opal_value_array_append_item (&group->group_events, &event_index))) {
+        return ret;
+    }
+
+    mca_base_var_groups_timestamp++;
```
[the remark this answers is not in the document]
hppritcha: question for @hjelmn

opal/mca/base/mca_base_event.h
```diff
@@ -0,0 +1,230 @@
...
+
+    /** size of the event data */
+    size_t event_extent;
+
+
+    /** Enumerator for integer values */
+    mca_base_var_enum_t *event_enumerator;
+
+    /** Type of object to which this variable must be bound or MCA_BASE_VAR_BIND_NULL */
+    int event_bind;
```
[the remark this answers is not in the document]
hppritcha: trying to change this to an enum type leads to other problems.  the event_bind is mapped sometimes to MPI_T values defined as part of the standard as well.  this is also the case with the analogous pvar structure.  @hjelmn will need to help here.

opal/mca/base/mca_base_source.c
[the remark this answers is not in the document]
hppritcha: switched to using the opal_clock facility

opal/mca/base/mca_base_source.c
```diff
@@ -0,0 +1,242 @@
...
+int mca_base_source_init (void)
+{
+    int ret = OPAL_SUCCESS;
+
+    OPAL_THREAD_LOCK(&mca_base_source_lock);
+    if (false == mca_base_source_initialized) {
+        mca_base_source_initialized = true;
+
+        OBJ_CONSTRUCT(&registered_sources, opal_pointer_array_t);
+        opal_pointer_array_init(&registered_sources, 16, 512, 16);
```
jsquyres, line 70: Should we make these constants be MCA params (perhaps level 9)? I don't have _too_ strong of an opinion here, but it is odd to see hard-coded constants in the middle of C code.

opal/mca/base/mca_base_source.c
```diff
@@ -0,0 +1,242 @@
...
+    OPAL_THREAD_UNLOCK(&mca_base_source_lock);
+
+    return OPAL_SUCCESS;
+}
+
+/***************************************************************************************************/
+
+mca_base_source_t *mca_base_source_get (int source_index)
+{
+    return opal_pointer_array_get_item (&registered_sources, source_index);
```
jsquyres, line 110: In some places in this file, accessing `registered_sources` is protected by `mca_base_source_lock`, but other places -- like here -- it's not.  Is there a reason for the discrepancy?

opal/mca/base/mca_base_source.c
```diff
@@ -0,0 +1,242 @@
...
+int mca_base_source_get_count (int *count)
+{
+    *count = source_count;
+    return OPAL_SUCCESS;
+}
+
+static inline int mca_base_source_get_by_name (const char *name, mca_base_source_t **source_out)
+{
+    /* there are expected to be a relatively small number of sources so a linear search should be fine */
+    for (int i = 0 ; i < source_count ; ++i) {
```
jsquyres, line 143: Shouldn't this routine be protected by the lock?  E.g., what if another thread changes `source_count`, or `registered_sources`? (or can that never happen?)

jsquyres:
Hey @hppritcha Are there any plans to continue this work (and delete PERUSE)?

hppritcha:
@jsquyres this is indeed kind of stalled out.  We need assistance here from @cchambreau and @hjelmn

cchambreau:
@jsquyres, @hppritcha : I will take a pass at the requested changes and contact @hjelmn or others with questions.

ibm-ompi:

ibm-ompi:

ompi/mpi/tool/source_get_num.c
```diff
@@ -0,0 +1,35 @@
...
+{
+    if (!mpit_is_initialized ()) {
+        return MPI_T_ERR_NOT_INITIALIZED;
+    }
+
+    if (MPI_PARAM_CHECK && NULL == num_source) {
+        return MPI_ERR_ARG;
+    }
+
+    return mca_base_source_get_count (num_source);
```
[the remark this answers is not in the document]
cchambreau: Sources are initialized by the mca variable system outside of mpit and should not change, so I don't believe this needs the mpit lock.

ibm-ompi:

 Gist: https://gist.github.com/1956ee0eecf44a1368bfc1c598aad17c

opal/mca/base/mca_base_source.c
cchambreau: `mca_base_source_lock` is used during source init and finalize, but the source data should not change after initialization.

opal/mca/base/mca_base_source.c
cchambreau: Same here, too, source info won't change.

ompi/mpi/tool/event_handle_alloc.c
```diff
@@ -0,0 +1,59 @@
...
+    mca_base_event_t * const event;
+    int ret;
+
+    if (!mpit_is_initialized ()) {
+        return MPI_T_ERR_NOT_INITIALIZED;
+    }
+
+    ompi_mpit_lock ();
+
+    /* Find the performance variable. mca_base_event_get() handles the
```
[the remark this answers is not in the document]
cchambreau: Addressed with [efaedce](https://github.com/open-mpi/ompi/pull/8057/commits/efaedce19b4c0ffa9c9e32f67ea5eb68f6349c67)

ompi/mpi/tool/event_set_dropped_handler.c
```diff
@@ -0,0 +1,40 @@
...
+    if (!mpit_is_initialized ()) {
+        return MPI_T_ERR_NOT_INITIALIZED;
+    }
+
+    /* Check that this is a valid handle */
+    if (MPI_T_EVENT_REGISTRATION_NULL == handle) {
+        return MPI_T_ERR_INVALID_HANDLE;
+    }
+
+    mca_base_event_registration_set_dropped_handler (handle, (mca_base_event_dropped_cb_fn_t) dropped_cb_function);
```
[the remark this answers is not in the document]
cchambreau: Addressed with [001fe76](https://github.com/open-mpi/ompi/pull/8057/commits/001fe766257aa74a18b3ac20407630f131e664c8)

cchambreau:
bot:ibm:retest

cchambreau:
bot:ibm:gnu:retest

jsquyres:
@cchambreau This PR has conflicts that must be resolved; it's unlikely that (re)running tests will be useful without fixing those conflict errors first.

ibm-ompi:

 Gist: https://gist.github.com/8f221d51807e996900724ae757e9168c

cchambreau:
@jsquyres I addressed the include file conflicts.  The copyright conflict looks like a typo.  
The PGI check is failing with:

```
  PPFC     profile/pcomm_agree_f08.lo
pgfortran-Error-Unknown switch: -iquote../../../..
make[2]: *** [Makefile:1464: comm_revoke_f08.lo] Error 1
```
I am unable to recreate this with my local PGI build.

The Pull Request Build Checker fails with:

```
make[3]: Entering directory '/home/ubuntu/workspace/open-mpi.build.distcheck/src/openmpi-gitclone/_build/sub/opal/mca/common/sm'
  CC       common_sm.lo
  CC       common_sm_mpool.lo
FATAL: command execution failed
```

Are you comfortable resolving the conflicts?  Do you have any advice for the check failures?

ibm-ompi:

 Gist: https://gist.github.com/09dbaa6cd75a1ac69398d14942a9018b

ompi/mpi/tool/event_callback_get_info.c
```diff
@@ -0,0 +1,53 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2012-2013 Los Alamos National Security, LLC. All rights
```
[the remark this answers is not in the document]
cchambreau: Addressed with [e07bc47](https://github.com/open-mpi/ompi/pull/8057/commits/e07bc47fd3d23e344ce9217a5fa363639fdcf836)

jjhursey:
Can you rebase your branch? It looks like there are conflicts detected by GH. If the IBM PGI CI still fails after that I can look in the backend to see if I can get more information for you.

jsquyres:
> Can you rebase your branch? It looks like there are conflicts detected by GH. If the IBM PGI CI still fails after that I can look in the backend to see if I can get more information for you.

@cchambreau I think @jjhursey's advice is the best: rebase your PR branch on top of HEAD of `main`, and you'll end up fixing the conflicts.  All the CI tests build/test an implicit merge from your PR branch to `main`.  Hence, the tests are failing where the git conflicts between your branch and `main` are ending up in compile errors, etc.

ibm-ompi:

 Gist: https://gist.github.com/c60af88444836261a9a58e7c74a11d65

ibm-ompi:

 Gist: https://gist.github.com/b86ae4eaf9707b7e9b79a1deed491c2e

ibm-ompi:

 Gist: https://gist.github.com/d96b8bf838baa5c99c575572979ac2c8

cchambreau:
Missed a merge conflict.  Rebuilding locally to identify any other missed conflicts before correcting.

jjhursey:
It looks like the rebase went wrong and instead picked up a merge commit (and a bunch of unrelated commits).

I'm not sure what is the easiest way to sort this out. Possibly `git format-patch HASH` to extract your commits from the branch, then hard rebase on main, then `git am patchfile` to bring back in your changes. Basically creating a new branch in place with just your changes - a manual rebase.

cchambreau:
@jsquyres OK, will look into that.

jsquyres:
bot:ompi:retest

jsquyres:
@cchambreau and I worked in a Webex to straighten out the commits on this PR, and rebase it to `main` HEAD.

@cchambreau is going to go through and make sure we didn't miss any commits, etc.

cchambreau:
@jsquyres I successfully built from the updated PR branch and ran against the MPI_T Events tests in ompi_tests_public.  I believe all previous changes are accounted for.

opal/mca/base/mca_base_var_group.c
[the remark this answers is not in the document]
jsquyres: I don't know if we can count on @hjelmn's response at this point...

ompi/mpi/tool/source_get_num.c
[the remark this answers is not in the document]
jsquyres: I don't think that's quite right -- the value can change over the course of the run, and therefore could be changed from a different thread.  MPI-4.0 p758 21:22 says:

> An MPI implementation is allowed to increase the number of sources during the execution of an MPI process.

opal/mca/base/mca_base_source.c
jsquyres: Per MPI-4.0 p758 21:22, the number of sources is allowed to increase during the execution of an MPI process.

jsquyres (commented):
There's been a huge amount of progress -- huzzah!

There's still a few open items, though.

Also, do you need to add LLNL copyrights to the files you've touched?

ompi/mpi/tool/source_get_num.c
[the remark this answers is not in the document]
cchambreau: I am adding locking to functions in mca_base_source.c, so locking here is unnecessary.

opal/mca/base/mca_base_source.c
cchambreau: I am adding locking to protect source access and prevent potential future source access issues.

ompi/mpi/tool/source_get_num.c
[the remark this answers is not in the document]
cchambreau: Addressed with https://github.com/open-mpi/ompi/pull/8057/commits/c9b35e2ac9a780d16a05bfc9b82b4aef5ab0a970

opal/mca/base/mca_base_source.c
cchambreau: Addressed with https://github.com/open-mpi/ompi/commit/c9b35e2ac9a780d16a05bfc9b82b4aef5ab0a970

opal/mca/base/mca_base_source.c
cchambreau: Addressed with https://github.com/open-mpi/ompi/commit/c9b35e2ac9a780d16a05bfc9b82b4aef5ab0a970

opal/mca/base/mca_base_source.c
```diff
@@ -36,29 +37,23 @@ int mca_base_source_default_source = -1;
...
 
 static uint64_t mca_base_source_default_time_source_ticks (void)
 {
-#if OPAL_HAVE_CLOCK_GETTIME
-    return 1000000000;
-#else
-    return 1000000;
-#endif
+    struct timespec spec;
+    if (0 == opal_clock_getres(&spec)){
```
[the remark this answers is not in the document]
cchambreau: Addressed with https://github.com/open-mpi/ompi/pull/8057/commits/5fae8cf1d7eb735918c45d354d9605471db3e960

cchambreau:
@jsquyres I do not need to add LLNL copyrights as per the LLNL incremental contribution conditions.

It looks like the checks are stalled after my most recent commits.  Perhaps I used an incorrectly configured access token.

jsquyres:
> It looks like the checks are stalled after my most recent commits. Perhaps I used an incorrectly configured access token.

Since you don't have commit access to the repo -- and this might be your first PR to OMPI...? -- github makes us click a button to enable the GitHub Action CI every time there's a change on this PR.  I probably didn't click it after the last push; I just did now after the "Correct OSC event description" 732c117 commit; it ran and passed.  The other CI's take longer to run.

opal/mca/base/mca_base_source.c
cchambreau: This appears to be an issue with several of these arrays, such as in opal/mca/base/mca_base_pvar.c :
```
opal_pointer_array_init(&registered_pvars, 128, 2048, 128);
```
The current source values would definitely address current use.
Would it make sense to address this more generally in a different PR?

opal/mca/base/mca_base_var_group.c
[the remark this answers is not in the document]
cchambreau: I'm not sure whether this should be atomic or not, but it happens elsewhere in this file at lines 242, 307, 405, 444, and 473 from previous commits.

ompi/mpi/tool/event_callback_get_info.c
```diff
@@ -0,0 +1,53 @@
...
+    if (!mpit_is_initialized ()) {
+        return MPI_T_ERR_NOT_INITIALIZED;
+    }
+
+    info = OBJ_NEW(ompi_info_t);
+    if (NULL == info) {
+        return ompit_opal_to_mpit_error(OMPI_ERR_OUT_OF_RESOURCE);
+    }
+
+    /* mca_base_cb_safety_t and MPI_T_cb_safety must be kept in sync for this to work */
```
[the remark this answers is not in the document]
cchambreau: @jsquyres Addressed with https://github.com/open-mpi/ompi/pull/8057/commits/3be65a4c7abc807d63a0e40e34d9143f9e45a1ac
It appears that AC_CONFIG_HEADERS doesn't support substitution (mpi.h), so I used defines for both cases.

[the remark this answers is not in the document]
jsquyres: Not sure why I thought `AC_SUBST` would be required.  Looks like the current fix is good.

opal/mca/base/mca_base_source.c
jsquyres: Yeah, I guess if this is something systemic and outside the scope of just MPI_T, then it's probably worth a future PR.

opal/mca/base/mca_base_var_group.c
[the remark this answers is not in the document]
jsquyres: I'm pretty sure that the `++` operator is not atomic in C.  The question is: can multiple threads be reading this value and/or writing this value at the same time?

ompi/mpi/tool/event_get_info.c
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+
+#include "ompi/mpi/tool/mpit-internal.h"
+
+/* needed to convert between opal and ompi datatypes until a function is provided */
+#include "ompi/datatype/ompi_datatype_internal.h"
```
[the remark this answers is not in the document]
cchambreau: Is the bad feeling from including the internal header?  It looks like this can be removed as there is a duplicate hard-coded value of OMPI_DATATYPE_MAX_PREDEFINED in ompi_datatype.h which is previously included through ompi/mpi/tool/mpit-internal.h.

Would deleting these two lines address your concern?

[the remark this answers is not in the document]
cchambreau: Also different defines : OMPI_DATATYPE_MAX_PREDEFINED vs OMPI_DATATYPE_MPI_MAX_PREDEFINED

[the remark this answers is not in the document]
cchambreau: Sorry, needs ompi_datatype_basicDatatypes from ompi_datatypes_internal.h, so you can ignore the above.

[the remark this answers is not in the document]
cchambreau: Alternate proposal as suggested by the comment:

Add a datatype function, such as ompi_datatype_lookup_by_opal_id that provides the ompi_datatype without needed to include ompi_datatype_internal.h.

opal/mca/base/mca_base_var_group.c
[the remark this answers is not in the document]
cchambreau: Addressed with https://github.com/open-mpi/ompi/pull/8057/commits/c2627ab5ef549d3abca3ec6217659960396e40fa

ompi/datatype/ompi_datatype_lookup_by_opal_id.c
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * $HEADER$
+ */
+
+#include "ompi_config.h"
+#include "ompi/datatype/ompi_datatype.h"
+#include "ompi/datatype/ompi_datatype_internal.h"
+
+
+ompi_datatype_t* ompi_datatype_lookup_by_opal_id( uint16_t opal_id )
```
bosilca, line 15: OPAL and OMPI datatypes are different things, and multiple OMPI datatypes will be mapped unto the same underlying OPAL datatype. Thus, looking for an OMPI datatype that matches a specific OPAL id, will give you the first OMPI datatype that matches that id, and that might be an OMPI datatype that is equivalent to your original one, but not necessarily the same. Why are you not always manipulating OMPI datatype id (which would be unique for each datatype).

ompi/mca/pml/ob1/pml_ob1_recvreq.c
```diff
@@ -449,6 +454,7 @@ static void mca_pml_ob1_rget_completion (mca_btl_base_module_t* btl, struct mca_
 static int mca_pml_ob1_recv_request_put_frag (mca_pml_ob1_rdma_frag_t *frag)
 {
     mca_pml_ob1_recv_request_t *recvreq = (mca_pml_ob1_recv_request_t *) frag->rdma_req;
+    ompi_communicator_t *comm = recvreq->req_recv.req_base.req_comm;
```
bosilca, line 457: This will generate warning when MCA_BASE_EVENT are disabled.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
```diff
@@ -1124,16 +1128,15 @@ static inline void append_recv_req_to_queue(opal_list_t *queue,
 {
     opal_list_append(queue, (opal_list_item_t*)req);
 
-#if OMPI_WANT_PERUSE
```
bosilca, line 1127: If nothing is optional, what is the impact on performance for supporting these MPI_T events ?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
```diff
@@ -1373,18 +1376,13 @@ void mca_pml_ob1_recv_req_start(mca_pml_ob1_recv_request_t *req)
...
             hdr = (mca_pml_ob1_hdr_t*)frag->segments->seg_addr.pval;
-            PERUSE_TRACE_MSG_EVENT(PERUSE_COMM_MSG_REMOVE_FROM_UNEX_Q,
-                                   req->req_recv.req_base.req_comm,
-                                   hdr->hdr_match.hdr_src,
-                                   hdr->hdr_match.hdr_tag,
-                                   PERUSE_RECV);
 
-            PERUSE_TRACE_COMM_EVENT(PERUSE_COMM_SEARCH_UNEX_Q_END,
-                                    &(req->req_recv.req_base), PERUSE_RECV);
+            MCA_BASE_EVENT_RAISE (mca_pml_ob1_events[MCA_PML_OB1_EVENT_UNEX_REMOVE].event,
```
bosilca, line 1381: Just to be sure: there is no equivalent to PERUSE REQ_MATCH_UNEX ?

ompi/mca/pml/ob1/pml_ob1_recvreq.h
```diff
@@ -268,7 +271,7 @@ static inline void recv_req_matched(mca_pml_ob1_recv_request_t *req,
  *
  */
 
-#define MCA_PML_OB1_RECV_REQUEST_UNPACK( request,                                 \
```
bosilca, line 271: Why this change ?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: There is not an implemented equivalent.  In this case, MCA_PML_OB1_EVENT_UNEX_REMOVE functions as a similar event, as it is only used here.  Do you think it would be good to add a specific match event?

hppritcha:
bot:ibm:retest

awlauria:
@cchambreau any chance this gets in the next few weeks? Have you been able to address @bosilca 's comments?

If it can get in soonish, we can consider it for a v5.0.0 release at the end of summer. Thanks!

ompi/datatype/ompi_datatype_lookup_by_opal_id.c
cchambreau: @bosilca I see what you are saying.  The underlying event support is implemented in opal/mca/base/mca_base_event.* with opal_datatype_t, which is a different size than ompi_datatype_t (uint16_t / uint32_t).  The currently implemented events are in OMPI modules.  Do you have a suggestion for tracking OMPI datatypes with the OPAL support?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: @bosilca I'm assuming that by default we would like the MPI_T Events functions visible, but would like to control whether the implemented events are supported, such as activation with a configure option like with the software performance counters.  Is this correct?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: To compute the duration of the unexpected search one needs to first know if the search was successful or not. If successful then the duration should be computed as `MCA_PML_OB1_EVENT_UNEX_REMOVE - MCA_PML_OB1_EVENT_SEARCH_UNEX_BEGIN` while if the search was not successful it should be computed as `MCA_PML_OB1_EVENT_SEARCH_UNEX_END - MCA_PML_OB1_EVENT_SEARCH_UNEX_BEGIN`. As long as it is well documented, it should be good.

ompi/mca/pml/ob1/pml_ob1_recvreq.h
bosilca: It looks there might be a conflict with the .request field in the mca_pml_ob1_transfer_event_t struct.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: @bosilca Here is some preliminary benchmarking I did on the LLNL quartz cluster, comparing OMB MPI microbenchmark times from my PR master branch with the events branch build and running without registered callbacks and then with registered callbacks.  The callbacks do not include reading event data.

Registered callback after MPI_Init for ob1 : ompi_pml_ob1_message_arrived		
Registered callback after MPI_Init for osc : ompi_osc_rdma_get_started		
Reported times are best of 10 runs each.

I'm not sure why the event osc/get timing is so much higher.  I tried to build and test with ompi/main, but ran into PMIx errors.

Are these the results you are interested in?  Let me know if you have specific requests for tests.

<head>



</head>

<body link="#0563C1" vlink="#954F72">



  | PR Master | Events build, no callbacks | Events w/Registered Callbacks
-- | -- | -- | --
2ppn-1N ob1 pt2pt osu_bw | 3.38 sec | 3.30 sec | 5.09 sec
1ppn-2N ob1 pt2pt osu_bw | 7.08 sec | 7.05 sec | 8.63 sec
  |   |   |  
2ppn-1N osc RMA osu_get_bw | 5.33 sec | 5.00 sec | 6.75 sec
1ppn-2N osc RMA osu_get_bw | 6.14 sec | 7.16 sec | 9.22 sec



</body>

</html>

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: @bosilca Sorry, I'm having difficulty disabling MCA_BASE_EVENT.  Can you provide me with some guidance here?

cchambreau:
@awlauria I have some additional changes to commit, but am leaving on vacation on Aug 8th.  When do you need this by?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: That `event build, no callback`is faster than `no events` is strange and points to some quirks in the measurements. what is the standard deviation of the measurements ?

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: Well, I don't think it is possible to turn them off. Going back to look into the code I have a new set of questions:
1. How to you turn the events off ? I see we can turn them on/off with the callbacks, but there seem to be no way to completely turn them off. 
2. Why are they implemented in opam/mca/base/mca_base_event*. These are not MCA base type of things, what goes in the base is related to the MCA params and aliases, the events looks too low level for me. In fact why are they not at the MPI level ?

bosilca: Looking again at the code I don't see any way to turn these events off at compile time (such that we totally compile them off from). Which means they will always be in, and we have to pay the cost, even when they are not in use. 

But looking at the code also raised additional questions. Why are the events MCA base level ? Why are they even related to the MCA ? I don't think this is the right place to have them in, I don't even think they deserve to be at the OPAL level (because they are only used in OMPI).

cchambreau: For 1 : I have implemented an --enable-mpi_t_events configure flag that disables implemented events in pml/ob1 and osc.  The MPI_T Events functions are still accessible, there just aren't any available events.  I'm hoping to make this available shortly.

For 2 : It looks to me like Nathan used the existing opal pvar code to implement MPI_T Events. My understanding is that events will be marked invalid if modules are not used.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: I see a standard deviation of .093 for the 1-node master measurements and .035 for the 1-node event-no-callbacks measurements.  I re-built all of these without debugging and am seeing even more dramatic differences for the measurements from master.  I am building the 4.1.4 release to see if I see better performance.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
bosilca: When your `--enable-mpi_t_events` will be turned off the comm variable you declare here will be unused, triggering a compiler complaint.

MCA stands for Modular Component Architecture, and I'm not sure how these MPI events could become part of that. We tried hard to separate things by scope, to facilitate the code readability, and from this perspective these MPI events do not belong into the MCA base. I'm sure others in the community have comments/suggestions.

cchambreau: OK, sounds like this is going to need input from others, such as @jsquyres or @hppritcha.  I will hold off on my update until we address this.  @awlauria I'm also assuming that this will not be able to be implemented in time for 5.0.

awlauria:
@cchambreau as soon as you can.  The official v5.0.0 release will be around the end of September. So sometime before then, allowing a week or two for bug fixes and other issues.

Thanks!

cchambreau:
@gpaulsen I was hoping to get some guidance regarding George's concern with this implementation adding code in OPAL.  It is based on OPAL code for MPI_T PVARs, but it sounds like George would prefer that this not be added in OPAL.  I'd like to get this resolved before making additional changes.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
cchambreau: @jsquyres @hppritcha @bosilca Any comments or suggestions on moving forward with this?  The MPI_T Events backend functionality has been implemented in OPAL, similar to the MPI_T Performance variable implementation, to allow registration of events based on available modules.

bosilca: This is exactly my point, why an `MPI_*` level functionality exists in OPAL ? There is no portability issues there, and these events are extremely MPI-facing (anybody else would not care about calling a specialized callback allowing an upper layer to track events).

ompi/datatype/ompi_datatype_lookup_by_opal_id.c
jsquyres: I think the intent here is:

* OPAL provides a basic datatype engine.  It's not specifically tied to the MPI API or MPI semantics.  The OPAL datatype engine does generic datatype-like things.
* The datatype functionality in MPI layer adds further semantics and data structures on top of the OPAL datatype engine to implement the MPI-specific datatype functionality.

I think @bosilca is asking to maintain the clean separation between the two.  Hence, you'll need to look up MPI datatypes as MPI datatypes -- not OPAL datatypes.

The same concept applies to the base MPI_T event support: it should be tracking MPI events and MPI datatypes and other MPI objects.  If desirable, there could be an underlying OPAL "event tracking" engine that does things in an MPI-agnostic way (wholly independent of the MPI layer APIs and MPI semantics).  This would be useful, for example, if we want to be able to track OPAL-level events that have no knowledge of the MPI layer (the BTLs seem like an obvious candidate here, for example).

The MPI_T functionality in the MPI layer should be able to call down into the OPAL "event tracking" engine to track MPI-level events.

The tricky part will then be how to correlate MPI-level events with OPAL-level events (e.g., how to correlate BTL events to specific MPI communicators, for example).

awlauria:
Removing the v5.0.x label, as these labels are used in PR's for the target branch - this target is `main`.

@cchambreau any update on this?

cchambreau:
@jquyres I'm hoping to get back to this in April.

hppritcha:
@cchambreau i'm trying to get this PR fixed up for merging so did a rebase and pushed to your repo.

ompi/mca/pml/ob1/pml_ob1_recvreq.c
hppritcha: I'm looking at seeing if there's a chance that we can rescue this PR.  From @bosilca comments it sounds like there are objections to the basic design for MPI_T events - using the mca variable registration infrastructure in opal.  It looks to me like the architect of most of the MPT_T pvar and now events decided to make use of the mca variable management mechanism since the inception of MPI_T support in OMPI.  Do we want to redesign this?  I'm noticing use of these pvar in both opal and ompi.

jsquyres:
Is it really a question of who needs this?

I.e., isn't this part of conformance to the MPI standard?

hppritcha:
Some tools folks were pinged about use of MPI_T events as part of Open MPI devel f2f meeting 4/2024.  ScoreP developer said its on their todo list to use this feature in some future release.  Tau may already use events if the MPI being used supports it.  Possibly some unreleased version of MPI Advisor can use them.

kingshuk00:
It would be nice to have this standard-compliant interface from tools perspective.
I have posted another PR [PR-6](https://github.com/cchambreau/ompi/pull/6) on the source branch of this PR. This fixes a potential problem that is captured in this test [events_callbacks.c](https://github.com/open-mpi/ompi-tests-public/blob/309a3dce6ef2aa8b68e9bdb16a28daf42a51f163/events/events_callbacks.c#L69). This should be there for correctness.

hppritcha:
this effort has been moved to #13133

PR #8041, opened 2020-09-12
SHMEM/SCOLL: Fix inplace reductions

brminich:
According to shmem spec source and dest arrays can be the same for reductions:
`The source and dest arrays may be the same array, but they may not be overlapping arrays.`
If this is the case and MPI scoll is used, need to specify MPI_IN_PLACE instead of real source buffer for allreduce

Fixes openucx/ucx#5627

Signed-off-by: Mikhail Brinskii <mikhailb@nvidia.com>

PR #8062, opened 2020-09-26
SHMEM/SCOLL: Fix inplace reductions - v4.1.x

brminich:
According to shmem spec source and dest arrays can be the same for reductions:
The source and dest arrays may be the same array, but they may not be overlapping arrays.
If this is the case and MPI scoll is used, need to specify MPI_IN_PLACE instead of real source buffer for allreduce

Porting from #8041 

Signed-off-by: Mikhail Brinskii <mikhailb@nvidia.com>
(cherry picked from commit dfe20e0472dff1f1911d72ede2b20913be6bd2e3)


brminich:
@yosefe, please review

yosefe (approved):

PR #8063, opened 2020-09-26
Check the AVX512_BW flag before using _mm512_add_epi*

bosilca:


bosilca:
This PR is wrong. Let's not spend time on it.

PR #8065, opened 2020-09-28
SPC: allow starting through MPI_T without explicitly setting MCA parameter

devreal:
So far, reading SPC counter through MPI_T was conditioned on the counters being attached via the MCA parameter `mpi_spc_attach`. This PR makes the following changes:

1) Remove the `MCA_BASE_PVAR_FLAG_CONTINUOUS` flag to enable explicit starting and stopping of counters and track the number of sessions that attach to each variable.
2) Inline calls that may be in the critical path (`ompi_spc_record`, `ompi_spc_user_or_mpi`, `ompi_spc_update_watermark`) to reduce overhead.
3) Clean up some code (move includes, remove unused attributes, remove unused variable)
4) Fix the SPC test to succeed even if an invalid pvar is encountered (e.g., a pvar that was de-registered by a component)
5) Centralize counter property management and consolidate counter information into `spc_event_t` structure to ensure that only a single cache-line is touched during an update.
6) Some additional minor changes for things I noticed (correct handling of timer and counter types, moved timer start/stop into header and fixed timer/watermark update functions)

### DIFF
--- a/ompi/runtime/ompi_mpi_params.c
+++ b/ompi/runtime/ompi_mpi_params.c
@@ -334,9 +334,10 @@ int ompi_mpi_register_params(void)
                                  MCA_BASE_VAR_SCOPE_READONLY,
                                  &ompi_mpi_compat_mpi3);
 
+#if SPC_ENABLE == 1
     ompi_mpi_spc_attach_string = NULL;
     (void) mca_base_var_register("ompi", "mpi", NULL, "spc_attach",
-                                 "A comma delimeted string listing the software-based performance counters (SPCs) to enable.",
+                                 "A comma-delimeted list of software-based performance counters (SPCs) to enable (\"all\" enables all counters).",
                                  MCA_BASE_VAR_TYPE_STRING, NULL, 0, 0,
                                  OPAL_INFO_LVL_4,
                                  MCA_BASE_VAR_SCOPE_READONLY,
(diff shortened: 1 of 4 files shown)


devreal:
It looks like one of the test bots timed out.

lanl-ompi:

jsquyres:
@devreal Looks like there's now a conflict on this PR.  Can you resolve it?

devreal:
@jsquyres The conflict should be resolved. 

@bosilca Can you please review?

bosilca (changes requested):
We designed this component such that the internals of the SPC management remains hidden instead of being exposed through the static inline functions in the header file. The rest of the PR is OK, but not moving all update functions in the header.

devreal:
The intention was that inlining these calls would reduce the overhead when SPC is enabled at compile time but disabled at runtime by cutting out the unconditional call. Reverting that is fine with me.

bosilca:
Last time we measured the cost of a function call it was significantly lower than the cost of an atomic operation. Plus SPC updates are supposed to be rare events, few times per request at most.

devreal:
A added more commits to make sure that a counter update only touches a single cache line and fixed a few more issues I came across while working on that (correct types used for counter and timer updates, made watermark update thread-safe, and removed a race in timers). It's gotten quite big and I'm happy to rip out the latest changes into another PR if that's wanted.

bosilca (commented):
Most of these changes looks good and have the potential to ease the use of SPC while also decreasing the overheads. But before we merge this into master I would like to evaluate the new overheads and compare with the original version.

devreal:
After digging through the generated assembly code I made some more changes to the code. I will squash everything down before merging.

Here are some measurements (10M rank-local 4B recv+isend+wait in a tight loop, time taken through `MPI_Wtime` on my local box, with `pml/ob1`):

1) With all counters enabled (`--mca mpi_spc_attach all`):
  a) Master: 3.763253s
  b) This branch: 3.498579s (~7% speedup)
2) With no counters attached:
  a) Master: 2.410163s
  b) This branch: 2.260756s (~6% speedup)
3) SPC disabled during configure: 2.198656s

There is still some small overhead compared to when SPC is completely disabled at configure time. This for the worst case though where there is no transfer of data between ranks and we're looking at about 220ns. This is mostly due to the additional loads and conditional jumps. It turns out that reading the cycle counter is relatively costly in this case (~6% of the time spent with all counters enabled, mostly due to the `lfence` before `rdtsc`). The rest is additional loads, conditional jumps, and stores. Overall I think the overhead is reasonable and the changes in this branch yield a slight reduction over what is currently in master.

bosilca (approved):

devreal:
I have squashed everything down into one commit. Should be ready to go once the tests have passed.

PR #8067, opened 2020-09-28
PML OB1: Fix potential overcounting of SPC sent/received message sizes and account for fin messages

devreal:
This PR fixes a potential over-counting of sent/received messages sizes in `pml/ob1`: the two functions `mca_pml_ob1_recv_request_ack_send_btl` and `mca_pml_ob1_recv_request_put_frag` do not transfer payload but manage the underlying protocol, so count the transferred data accordingly. Also add instrumentation for sending `fin` messages in `mca_pml_ob1_send_fin`.


rhc54:
@devreal I have nothing to do with the PML and cannot review that code

bosilca (approved):

Issue #7560, opened 2020-03-23
Fail to link MPI apps with "--enable-static --disable-shared" master builds; missing zlib symbols


jsquyres:
Building master with `--enable-static --disable-shared` builds and installs ok, but when you try to build an MPI application, it fails to link, complaining that it can't find zlib symbols.

For example: https://mtt.open-mpi.org/index.php?do_redir=3212 -- the behavior is basically this:

```
$ mpicc hello_c.c -o hello
/home/jsquyres/bogus/lib/libopen-pal.a(compress_zlib.o): In function `pmix_compress_zlib_compress_block':
compress_zlib.c:(.text+0x8e): undefined reference to `deflateInit_'
compress_zlib.c:(.text+0xa8): undefined reference to `deflateBound'
compress_zlib.c:(.text+0xc9): undefined reference to `deflateEnd'
compress_zlib.c:(.text+0xf9): undefined reference to `deflateEnd'
compress_zlib.c:(.text+0x14c): undefined reference to `deflate'
compress_zlib.c:(.text+0x15e): undefined reference to `deflateEnd'
/home/jsquyres/bogus/lib/libopen-pal.a(compress_zlib.o): In function `pmix_compress_zlib_uncompress_block':
compress_zlib.c:(.text+0x349): undefined reference to `inflateInit_'
compress_zlib.c:(.text+0x39b): undefined reference to `inflate'
compress_zlib.c:(.text+0x3aa): undefined reference to `inflateEnd'
collect2: error: ld returned 1 exit status
```

The issue is that we recently removed the "compress" framework from OPAL in 66da0c63 (because nothing was using it).  Therefore, the wrapper compiler flags no longer get `-lz` (or whatever "compress" library was selected).  But that compress framework effectively migrated over to PMIx, where it _is_ still used.  But there's nothing propagating wrapper compiler flags up from PMIx's `configure` script to Open MPI's `configure` script.

gpaulsen:
Today @jsquyres reported that he and @bwbarrett discussed and that Brian is working on.
Adding Severity: critical, as this may impact both static and non-static builds.

gpaulsen:
@bwbarrett Is this work, part of the current configury work you're working on, or is this separate?

bwbarrett:
I'm not sure why this bug got assigned to me?  No, this is not something I intend to fix as part of my patch series.  I'm not entirely sure what the fix will be.  With PMIx as a stand-alone project, dependency management of libraries just got really hard.

[...]

PR #8069, opened 2020-09-29
Rework how Open MPI integrates with 3rd party packages

bwbarrett:
With Open MPI 5.0, the decision was made to stop building 3rd-party packages, such as Libevent, HWLOC, PMIx, and PRRTE as MCA components and instead 1) start relying on external libraries whenever possible and 2) Open MPI builds the 3rd party libraries (if needed) as independent libraries, rather than linked into libopen-pal.  This patch series implements this behavior.


bwbarrett:
This final patch, "libevent: Remove opal prefixes", is completely optional.  It does clean up the code a little bit, but could also be done later as a separate patch series.

ibm-ompi:

 Gist: https://gist.github.com/cadc75c5056b606d2c202d2a52726e1d

ibm-ompi:

 Gist: https://gist.github.com/2f1d0c31e87e35436221c743753866fd

ibm-ompi:

 Gist: https://gist.github.com/6a1ca3272fce404f150268568dedba47

ibm-ompi:

 Gist: https://gist.github.com/51a040d9a45c163ef66c5e803ea04b55

bwbarrett:
Rebase on top of master, fixed a VPATH design issue, fixed an issue with "make check" that was causing "make distcheck" to fail, and removed the libevent rename removal patch.  @jsquyres pointed out that my cleanup didn't work right in the usnic code, so I want to treat that with more care.

rhc54:
FWIW: I just sync'd my local clone of your branch with PMIx and PRRTE and things built/work fine. I'm going to run a local MTT on it, though I'm limited on resources.

ibm-ompi:

 Gist: https://gist.github.com/2e0f4d55a804580acc64399374784f2b

rhc54:
Looks like you are still hitting the platform file propagation problem - PMIx doesn't have the "mellanox" platform file

rhc54:
@bwbarrett You know, there is zero value in the `platform` option for PMIx or PRRTE any more. Would it help if we simply removed them?

rhc54:
FWIW: libevent spews hundreds of warning messages with these latest changes, all due to poor matching of data types with format options in print statements. 😦

rhc54:
Local run of the IBM test suite looks good (except for one-sided, which totally failed on my Mac). Let me know if you agree about removing the platform options, and if you want me to push the PMIx/PRRTE updates to your branch.

bwbarrett:
@rhc54 why do you say that there's no value in the platform file for PMIx/PRRTE?

rhc54:
Perhaps I should say "not that much value". It is true that we are adding more things that have dependencies (fabric, storage), so maybe it will change enough over time. If you can support it, then that would be best - if not, we can find a way to live with it. If nothing else, we could change the name of the option so it doesn't pickup the top-level one!

bwbarrett:
@rhc54 For now, I just prevent the --with-platform argument from being passed to PMIx and PRRTE.  We can revisit that later, if we decide it is important.

Also fixed a PRRTE linking issue on recent GNU linkers (like that found on Ubuntu 20.04).

Still need to figure out what is going on with the build flags and PGI compilers.

ibm-ompi:

 Gist: https://gist.github.com/86efd7bff75cc755ce36da2a5592cf2b

awlauria:
Does this resolve https://github.com/open-mpi/ompi/issues/7560 ?

rhc54:
I changed the platform options on both PMIx and PRRTE, so no worries there any more.

bwbarrett:
bot:aws:retest

It looks like Jenkins crashed on one of the builders.

bwbarrett:
@rhc54 Let's get this commit in with the current submodule pointers, and then we can follow up with another bump.  Trying to get this over the line :).

config/ompi_setup_prrte.m4
```diff
@@ -16,130 +16,153 @@
...
 
     AC_ARG_ENABLE([internal-rte],
                   [AC_HELP_STRING([--enable-internal-rte],
                                   [Enable internal runtime support and provide mpiexec/mpirun (default: enabled)])])
 
-    AC_ARG_WITH([prrte-platform],
-                [AC_HELP_STRING([--with-prrte-platform],
-                                [Platform file to use when building the internal PRRTE runtime support])])
+    # TODO: BWB: Check with Ralph if we need seperate files for OMPI and PRTE (which seems annoying)
+dnl    AC_ARG_WITH([prrte-platform],
```
rhc54, line 61: I made this option with only one 'r' - i.e., "prte"

config/opal_config_pmix.m4
```diff
@@ -0,0 +1,200 @@
...
+
+         # Pass all our compiler/linker flags to PMIx, so that it
+         # picks up how to build an internal HWLOC and libevent, plus
+         # picks up any user-specified compiler flags from the master
+         # configure run.
+         export CFLAGS CPPFLAGS LDFLAGS
+         PAC_CONFIG_SUBDIR_ARGS([3rd-party/openpmix], [$internal_pmix_args],
+                           [[--with-libevent=internal], [--with-hwloc=internal],
+                            [--with-libevent=external], [--with-hwloc=external],
+                            [--with-pmix=.*], [--with-platform=.*]],
```
rhc54, line 94: PMIx is now looking for `--with-pmix-platform`

config/opal_config_pmix.m4
```diff
@@ -0,0 +1,200 @@
...
+           # need these set for the tests below.  If things fail, will undo at the end.
+           CPPFLAGS="$opal_pmix_CPPFLAGS_save $opal_pmix_CPPFLAGS"
+           LDFLAGS="$opal_pmix_LDFLAGS_save $opal_pmix_LDFLAGS"
+           LIBS="$opal_pmix_LIBS_save $opal_pmix_LIBS"
+
+           AS_IF([test "$opal_pmix_external_support" = "yes"],
+                 [AC_MSG_CHECKING([if external PMIx version is 3.0.0 or greater])
+                    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <pmix_version.h>]],
+                          [[
+#if PMIX_VERSION_MAJOR < 3L
```
rhc54, line 166: I'm not sure OMPI (setting aside PRRTE) can support v3.0.0 - it may well require at least v3.1.5. I wouldn't hold this PR over it, but is something we should check prior to release.

rhc54 (changes requested):
This is a nit and can wait, but I would suggest moving the "base" code of the hwloc and pmix frameworks into the "opal/util" directory like you did for libevent to avoid confusion. Can be done later if you prefer - I just think it will confuse people down the line to see a "defunct" framework for those two projects.

Otherwise, it looked good. Minor changes to the "platform" file names for PRRTE and PMIx.

config/ompi_setup_prrte.m4
bwbarrett: Is the project PRRTE or PRTE?  I'd like to be consistent through the whole system.  In terms of the actual argument, Open MPI will pass all arguments not explicitly called out down to PRTE.  So we don't actually have to do anything here; customers can do `./configure --with-platform=<foo> --with-prte-platform=<bar>` and the right things will happen.

But I should remove this code, since it isn't needed at this point.

config/opal_config_pmix.m4
bwbarrett: Yup; we don't actually have to do anything here.  The 3rd argument is an array of options that PAC_CONFIG_SUBDIR_ARGS will strip out from the top-level configure options when passing down to the sub-configure.  This will avoid passing the `--with-platform` argument to PMIx, as belt and suspenders type thing.

config/opal_config_pmix.m4
bwbarrett: Good call out; I think testing is the right path.  3.0.0 was what the external component had been checking for, which is why I left it.

config/ompi_setup_prrte.m4
rhc54: The project is "PRRTE", but the commands and variables are all with a single "r". I don't know how we want to expose the PRRTE and PMIx configure options at the user level for OMPI - I guess that is an open question? I grok that the right thing will happen - just not sure how the user can see what options exist.

bwbarrett:
> This is a nit and can wait, but I would suggest moving the "base" code of the hwloc and pmix frameworks into the "opal/util" directory like you did for libevent to avoid confusion. Can be done later if you prefer - I just think it will confuse people down the line to see a "defunct" framework for those two projects.
> 
> Otherwise, it looked good. Minor changes to the "platform" file names for PRRTE and PMIx.

I tried to pull out the HWLOC and PMIx code into util.  It was a big project and I gave up.  We can always do it later.  I left some notes in the mca/<framework>/ directories as a result.

config/ompi_setup_prrte.m4
bwbarrett: I'm not sure there's a great answer here, other than eventually getting `./configure --help=recursive` to work properly.  I should figure out why that doesn't.  Otherwise, we end up exposing way too many options (like why is `--with-platform` special compared to all the others?).

rhc54: I agree - I think we can set it aside for now as it may become too complex and confusing to the user. Perhaps all we need do at some point is add a comment to configure help indicating that these subprojects are included and direct them on how to find configure options specific to each subproject.

bwbarrett: Removed the commented out platform argument from the top level.  Also figured out how to make `./configure --help=recursive` work correctly, which I think is the answer to your question about how customers find arguments for sub-projects.

bwbarrett:
Removed an un-needed todo that Ralph found and fixed `./configure --help=recursive` so that sub-configure options could be found.

rhc54 (approved):
Looks go9od - thx!

PR #8076, opened 2020-10-01
build: Ignore use-mpi's types file

bwbarrett:
This file is autogenerated, so git should not try to track it.


ggouaillardet (approved):
good catch, thanks!

PR #8077, opened 2020-10-03
Fix 3rd-party package CPPFLAGS (and others) handlinge

bwbarrett:
This patch series started as a fix for George's report that we broke support for `--enable-visibility`, but expanded to a significant cleanup of the Autoconf precious flags (`CC`, `CXX`, `CPPFLAGS`, etc.; the ones listed at the bottom of `./configure --help`).  The core issue with `--enable-visibility` was that the top-level configure adds `-Wmissing-prototypes` to `CFLAGS`.  The visibility tests use `-Werror` in `CFLAGS` to make the compiler puke if it does not support visibility features.  However, the test for visibility features uses the standard Autoconf C program template, which does not fully specify the `main()` prototype.  That test, plus `-Wmissing-prototypes`, plus `-Werror` resulted in a compiler error.

With this patch series, the sub-configure for the 3rd party packages will not use the current environment to determine which precious variables are passed down (nor their values).  Instead, precious variables will be passed down if either they were in the environment at the start of configure, they were on the configure command line, or the caller to the subconfigure explicitly sets a new value for the precious variable and asks that it is passed to the subconfigure.  In this way, the subconfigure scripts will have the exact same precious variable values as the top level, modulo any explicitly requested changes (like adding the -Is for co-built packages).  This fixes the `--enable-visibility` issue and should lead to fewer subpackage surprises in the future.

### DIFF
--- a/config/aclocal_subcfg.m4
+++ b/config/aclocal_subcfg.m4
@@ -73,31 +73,45 @@ AC_DEFUN([PAC_CONFIG_SUBDIR_ARGS],[
           --disable-option-checking)
             ;;
           *)
-            # MPICH note: this is a more robust version of the "precious
-            # variable" propagation code that was present in the previous
-            # incarnation of this macro
+            # strip out precious variables from ac_configure_args,
+            # which will include precious variables that are currently
+            # set and were set on the command line or in the
+            # environment at the time configure was invoked.  Instead,
+            # we add all precious variables which have been tagged as
+            # set, so that we can more closely control the environment
+            # of sub-configures.
+            is_precious=0
             for pac_pvar in $ac_precious_vars ; do
                 # check if configure argument token contains the
                 # precious variable, i.e. "name_of_prec_var=".
                 if ( echo $pac_arg | grep "^$pac_pvar=" >/dev/null 2>&1 ) ; then
-                    # check if current precious variable is set in env
-                    eval pvar_set=\${$pac_pvar+set}
-                    if test "$pvar_set" = "set" ; then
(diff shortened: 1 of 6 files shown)


rhc54 (approved):
Gees, this 3rd-party thing turned out to be a lot of work! Thanks for tackling it!

One thing I've always wondered: why do we not enable-orterun-prefix-by-default by default? I can't imagine the benefit of _not_ setting that flag. Is there some reason why the user has to specifically ask for it?

rhc54:
Looks like Mellanox has an issue with the PMIx path:
```text
============================================================================
== Event libraries
============================================================================
checking --with-libev value... simple ok (unspecified)
checking --with-libev-libdir value... simple ok (unspecified)
checking will libev support be built... no
checking for libevent... cobuild
checking if co-built libevent includes thread support... no
configure: error: No thread support in co-build libevent.  Aborting
configure: ===== done with 3rd-party/openpmix configure =====
configure: error: Could not find viable pmix build.
```
However, I suspect it may have to do with the lack of thread support in libevent - at least in PMIx and PRRTE, that is required.

bwbarrett:
The Mellanox issue is a real issue, although I do not entirely understand what is different about their configuration.  Libevent was built with thread support, so that is not the problem.  It appears that CPPFLAGS were not pushed down to PMIx properly, so PMIx did not find the internal libevent headers.

bwbarrett:
The Mellanox issue was caused by my poor regex to not pass `--with-platform=<arg>` to sub-configure scripts.  Fixed now (this would have eventually bitten us, even without this weird visibility bug).

PR #8091, opened 2020-10-13
fix SPC example

naughtont3:


naughtont3:
The counter names were missing `_SPC_` in name.  Also, add comment with pointer to full list of the SPC counters (`ompi_spc_counters_t`).

Note: Same problem may exist in `tests/spc/`, but did not look closely at that test.

bosilca (approved):

PR #8093, opened 2020-10-14
Adding ofi include to CPPFLAGS so that configure can check fabric.h

dancejic:
configure was previously failing to check for the fi_info.nic struct because
fabric.h relied on other header files in the ofi/include dir. This adds that
include to CPPFLAGS before running that check so that configure can check for
the struct.

### DIFF
--- a/config/opal_check_ofi.m4
+++ b/config/opal_check_ofi.m4
@@ -121,11 +121,13 @@ AC_DEFUN([_OPAL_CHECK_OFI],[
                               [],
                               [opal_ofi_happy=no])])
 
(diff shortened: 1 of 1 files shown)


jsquyres (approved):

jsquyres:
@dancejic Does this commit also need to be ported to the v4.0.x and v4.1.x branches?

dancejic:
yes, I'll open PRs for those as well.

jsquyres:
@dancejic Make sure to register for [Hacktoberfest](https://hactoberfest.com/) 😉

dancejic:
Looks like the multi-NIC code isn't in 4.0.x, but I did open a PR for 4.1.x: #8094.

Issue #7717, opened 2020-05-10
--map-by with PE=n not working


cniethammer:
I see this issue with Open MPI 4.0.3 installed with pure `./configure --prefix=... --enable-shared --enable-static` under Centos 8 running on a dual socket  AMD EPYC 7742 64-Core Processor or single CPU Intel I5-8350U system.

The logic of `--map-by`, `PE=n`, and `--bind-to` does not work as intended.

The `man mpirun` example for the usage of `--map-by`includes the following:
```
 % mpirun ... --map-by core:PE=2 --bind-to core
[...] ... binding child [...,0] to cpus 0003
[...] ... binding child [...,1] to cpus 000c
[...] ... binding child [...,2] to cpus 0030
[...] ... binding child [...,3] to cpus 00c0
```

However running this I get:
```
$ mpirun -np 2 --report-bindings --map-by core:PE=2 --bind-to core /bin/true
--------------------------------------------------------------------------
A request for multiple cpus-per-proc was given, but a directive
was also give to map to an object level that cannot support that
directive.

Please specify a mapping level that has more than one cpu, or
else let us define a default mapping that will allow multiple
cpus-per-proc.
--------------------------------------------------------------------------
```

cniethammer:
A quick test with latest master (0a21a58f08196bc9e2717bbac33f10aedaca41bd) shows the same problem.

rhc54:
I believe I explained this elsewhere, but I'll do so again here. That message is not an error - it is telling you precisely the correct thing. You cannot map-by core and bind the result to two cores as that is logically impossible to do. You are telling us to assign a process to a single core, and then bind it to two cores.

What you really want is to `map-by slot:pe=2` so that we will map processes sequentially within the socket, binding each proc to two cores. This can logically work as a "slot" has no defined number of cores in it.

rhc54:
In other words, the example in the man page is wrong, not the code implementing the mapping.

jsquyres:
@cniethammer I've opened PRs for v4.0.x and v4.1.x with new wording.

Did you want to have a look?  Check out #8099 (v4.0.x) and #8192 (a straight cherry-pick of #8099 for v4.1.x).

cniethammer:
The man page fixes are fine. closing this

PR #8099, opened 2020-10-17
v4.0.x: orterun.1in: fix minor mistake in :PE=2 example and add more descriptions/explanations

jsquyres:
See individual commit messages for detail

These are not cherry-picks from master because PRRTE has replaced ORTE on master, and orterun.1in no longer exists in master.

Refs #7717

### DIFF
--- a/orte/tools/orterun/orterun.1in
+++ b/orte/tools/orterun/orterun.1in
@@ -397,15 +397,17 @@ To map processes:
 .
 .TP
 .B --map-by \fR<foo>\fP
-Map to the specified object, defaults to \fIsocket\fP. Supported options
-include slot, hwthread, core, L1cache, L2cache, L3cache, socket, numa,
-board, node, sequential, distance, and ppr. Any object can include
-modifiers by adding a \fR:\fP and any combination of PE=n (bind n
-processing elements to each proc), SPAN (load
-balance the processes across the allocation), OVERSUBSCRIBE (allow
-more processes on a node than processing elements), and NOOVERSUBSCRIBE.
-This includes PPR, where the pattern would be terminated by another colon
-to separate it from the modifiers.
+Map to the specified object, defaults to \fIsocket\fP. Supported
+options include \fIslot\fP, \fIhwthread\fP, \fIcore\fP, \fIL1cache\fP,
+\fIL2cache\fP, \fIL3cache\fP, \fIsocket\fP, \fInuma\fP, \fIboard\fP,
+\fInode\fP, \fIsequential\fP, \fIdistance\fP, and \fIppr\fP. Any
+object can include modifiers by adding a \fI:\fP and any combination
+of \fIPE=n\fP (bind n processing elements to each proc), \fISPAN\fP
+(load balance the processes across the allocation),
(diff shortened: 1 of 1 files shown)


jsquyres:
@rhc54 If this is correct, I'll make a PR to cherry-pick this to the v4.1.x branch.

orte/tools/orterun/orterun.1in
```diff
@@ -1213,8 +1213,13 @@ the masks 0001, 0002, 0004, and 0008.  In the second case, processes bind
 to all cores on successive sockets as indicated by the masks 000f and 00f0.
 The processes cycle through the processor sockets in a round-robin fashion
 as many times as are needed.  In the third case, the masks show us that
-2 cores have been bound per process.  In the fourth case, binding is
-turned off and no bindings are reported.
+2 cores have been bound per process.  Note that the mapping by slot
+implied that hardware threading was enabled, and the mapping proceeded
```
rhc54, line 1217: Mapping by slot doesn't have anything to do with HTs - a "slot" is just an allocation that means you are allowed to run a process. It has nothing to do with hardware.

What that cmd line actually means is to start however many processes were requested, mapping them sequentially across each node before moving to the next node. Each process is to be assigned two processors (cores by default, or HTs if --use-hwthreads-as-cpus is given). Note that "bind-to core" may well get an error for including it for the case where cores are being used as PUs since the mapping algorithm includes a conflicting directive. I'd have to look at the mapping code to see how it handles that conflict.

Regardless, this text is incorrect in implying a connection between "slot" and hardware. The two concepts are unrelated.

lanl-ompi:

gpaulsen:
@jsquyres Can you please update or comment on @rhc54's comment please?

jsquyres:
@rhc54 Per our phone conversation, I updated the text.  I actually made 3 separate commits that you may want to look at individually:

1. Fix the example and add a little further explanation of that example.
1. Add sections to define "slot" and "processor element".
1. Add a little more markup (no content changed, just markup).

orte/tools/orterun/orterun.1in
```diff
@@ -107,6 +107,105 @@ using an appropriate binding level or specific number of processing elements per
...
+be informed of the number of slots for each node by the resource
+manager.  For example:
+.
+.TP 4
+mpirun ./a.out
+launches one process for every slot (on every node) as dictated by
+the resource manager job specification.
+.
+.P
+Also note that the one-process-per-slot restriction can be overridden
```
rhc54, line 164: I suppose this is a nit, but we don't allow oversubscription in RM-managed environments. If the RM allocates you 2 slots, then you can only run 2 processes. Period.

jjhursey:
Note: We need to make sure to sync the language in this PR with the PRRTE version being developed (I'll do that - just dropping a note to link the two)
 * https://github.com/openpmix/prrte/pull/557

orte/tools/orterun/orterun.1in
jsquyres: Ok; I amended that text.

rhc54 (approved):

PR #8103, opened 2020-10-20
v4.0.x: fortran.m4: reword error message when sizeof(int) != sizeof(INTEGER)

ggouaillardet:
Reword the error message to suggest only the Fortran INTEGER size
can be changed via adhoc compiler flags.

This is a one-off commit for the release branches, master does
it differently (and breaks ABI).


jsquyres (approved):

lanl-ompi:

gpaulsen (approved):

PR #8113, opened 2020-10-21
mtl/ofi: Do not fail if error CQ is empty

rajachan:
In multi-threaded scenarios, any thread that attempts to read a CQ
when there's a pending error CQ entry gets an -FI_EAVAIL. Without
any serialization here (which is okay, since libfabric will protect
access to critical CQ objects), all threads proceed to read from the
error CQ, but only one thread fetches the entry while others get
-FI_EAGAIN indicating an empty queue, which is not erroneous.


bwbarrett (approved):

hppritcha:
@rajachan could you cherry-pick this to the 4.0.x branch as well?  thanks.

rajachan:
@hppritcha done https://github.com/open-mpi/ompi/pull/8125

PR #8116, opened 2020-10-22
configury: enhance the check for ISO_FORTRAN_ENV module

ggouaillardet:
When real*16 is supported, test with real128 kind.
Test with real32 kind otherwise.

Thanks Alan Wild for reporting this issue.

### DIFF
--- a/config/ompi_fortran_check_iso_fortran_env.m4
+++ b/config/ompi_fortran_check_iso_fortran_env.m4
@@ -11,6 +11,8 @@ dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2010-2014 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright (c) 2020      Research Organization for Information Science
+dnl                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


jsquyres (approved):
The code changes look correct, but can you describe what was wrong with the previous test?  I.e., why is it better to check for ISO_FORTRAN_ENV with the real128 kind vs. the real32 kind?

ggouaillardet:
We previously implicitly assumed that when `real*16` is supported, then `real128` is usable in the `ISO_FORTRAN_ENV` module. It turned out that was wrong since `f18` does support `real*16` but `real128` is defined as `-1` in `ISO_FORTRAN_ENV` and is hence an invalid kind and cannot be used.
`real*16` is used in `mca/ops` and does not require `real128` here, that is why these two things have to be tested separately.
`ISO_FORTRAN_ENV` is only used by `MPI_Sizeof()`, so this is a good place to test `real128` (but only when `real*16` is supported).

ggouaillardet:
FWIW, an other approach would be to generate `MPI_Sizeof()`, but without support for `real*16`.
That looked half baked to me (not suggesting here `MPI_Sizeof()` is fully baked whatsoever) so I opted not to support `MPI_Sizeof()` with `f18` instead.

lanl-ompi:

PR #8119, opened 2020-10-22
[v4.1.x] mtl/ofi: Do not fail if error CQ is empty

rajachan:
In multi-threaded scenarios, any thread that attempts to read a CQ
when there's a pending error CQ entry gets an -FI_EAVAIL. Without
any serialization here (which is okay, since libfabric will protect
access to critical CQ objects), all threads proceed to read from the
error CQ, but only one thread fetches the entry while others get
-FI_EAGAIN indicating an empty queue, which is not erroneous.

(cherry picked from commit 415dddb9afbed28dc93f65ead867ef5d245d94c4)

### DIFF
--- a/ompi/mca/mtl/ofi/mtl_ofi.h
+++ b/ompi/mca/mtl/ofi/mtl_ofi.h
@@ -137,6 +137,17 @@ ompi_mtl_ofi_context_progress(int ctxt_id)
                                 &error,
                                 0);
             if (0 > ret) {
+                /*
+                 * In multi-threaded scenarios, any thread that attempts to read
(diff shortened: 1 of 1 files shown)


jsquyres:
@rajachan Is this fix also on master?  If so, we need it to be cherry-picked from master.

rajachan:
I don't always backport fixes, but when I do, I only cherry-pick from mainine:
https://github.com/open-mpi/ompi/pull/8113

awlauria:
You'll want to add the `-x` to the cherry-pick and force-push

bwbarrett (approved):

PR #8122, opened 2020-10-23
Fix cpu-list for non-uniform nodes

jjhursey:
* The `--cpu-list` argument restricts the hwloc topology. When that topology is
   sent from the remote daemon back to the HNP it is packed as XML. This packing
   results in the loss of the applied cpu-list. Mapping will then be using
   the full topology instead of the restricted topology when mapping the
   processes. When the launch command reaches the remote node it will not
   be congruent with the remote node's view of the topology leading to a
   launch failure especially if the HNP and remote node have differing
   topologies.
   - The solution is to make sure that the HNP re-applies the cpu-list to
     the incomming topology from the remote node. This way the HNP and the
     remote node are using the same restricted topology.

### DIFF
--- a/orte/mca/plm/base/plm_base_launch_support.c
+++ b/orte/mca/plm/base/plm_base_launch_support.c
@@ -16,7 +16,7 @@
  * Copyright (c) 2013-2018 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014-2018 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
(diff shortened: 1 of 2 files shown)


jjhursey:
This is going directly into v4.1.x since master does not have this code path (since it is an ORTE fix).

lanl-ompi:

gpaulsen (approved):

gpaulsen:
@jjhursey, I assume we also want this in v4.0.x.

PR #8124, opened 2020-10-23
mtl/ofi: Fix erroneous FI_PEEK/FI_CLAIM usage

rajachan:
The current iprobe/improbe implementations merely checks the return
code on the posted receive operation to tell if there is a match or
not. This commit moves the check to the probe's error callback
instead. Per the semantics defined in libfabric, the peek operation is
asynchronous and the results are to be fetched from the completion
queue. If no message is found matching the tags specified in the peek
request, then a completion queue error entry with err field set to
FI_ENOMSG will be available.


rajachan:
Tested with probe, iprobe, mprobe, and improbe tests from the IBM suite and they pass.

bwbarrett (approved):

hppritcha (approved):

lanl-ompi:

PR #8132, opened 2020-10-25
Change MCA component build style default to static

bwbarrett:
This patch series primarily changes the default component build style from dso to static.  At the same time, it cleans up some minor nits and clarifies the selection logic when there are conflicting --enable-mca-dso and --enable-mca-static arguments.

### DIFF
--- a/config/opal_mca.m4
+++ b/config/opal_mca.m4
@@ -166,7 +166,7 @@ AC_DEFUN([OPAL_MCA],[
     # resolution (prefer static) is done in the big loop below
     #
     AC_MSG_CHECKING([which components should be run-time loadable])
-    if test "$enable_static" != "no"; then
+    if test "$enable_static" != "no" || test "$OPAL_ENABLE_DLOPEN_SUPPORT" = 0; then
         DSO_all=0
         msg=none
     elif test -z "$enable_mca_dso" || test "$enable_mca_dso" = "yes"; then
@@ -175,7 +175,6 @@ AC_DEFUN([OPAL_MCA],[
     elif test "$enable_mca_dso" = "no"; then
         DSO_all=0
         msg=none
-        enable_dlopen=no
     else
         DSO_all=0
         ifs_save="$IFS"
--- a/config/opal_mca.m4
+++ b/config/opal_mca.m4
@@ -705,15 +705,12 @@ AC_DEFUN([MCA_COMPONENT_COMPILE_MODE],[
         [str="STATIC_COMPONENT=\$STATIC_$2_$3"
          eval $str])
 
-    shared_mode_override=static
-
     # Setup for either shared or static
     if test "$STATIC_FRAMEWORK" = "1" || \
        test "$STATIC_COMPONENT" = "1" || \
        test "$STATIC_all" = "1" ; then
         $4="static"
-    elif test "$shared_mode_override" = "dso" || \
-         test "$SHARED_FRAMEWORK" = "1" || \
+    elif test "$SHARED_FRAMEWORK" = "1" || \
          test "$SHARED_COMPONENT" = "1" || \
          test "$DSO_all" = "1"; then
         $4="dso"
--- a/config/opal_mca.m4
+++ b/config/opal_mca.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2010-2016 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
-dnl Copyright (c) 2018      Amazon.com, Inc. or its affiliates.
+dnl Copyright (c) 2018-2021 Amazon.com, Inc. or its affiliates.
 dnl                         All Rights reserved.
 dnl $COPYRIGHT$
 dnl
@@ -68,8 +68,7 @@ AC_DEFUN([OPAL_MCA],[
                         type-component pairs that will be built as
                         run-time loadable components (as opposed to
                         statically linked in), if supported on this
-                        platform.  The default is to build all components
-                        as DSOs.])])
+                        platform.])])
     AC_ARG_ENABLE([mca-static],
         [AS_HELP_STRING([--enable-mca-static=LIST],
                        [Comma-separated list of types and/or
@@ -77,7 +76,8 @@ AC_DEFUN([OPAL_MCA],[
                         linked into the library.  The default (if DSOs are
                         supported) is to build all components as DSOs.
                         Enabling a component as static disables it
-                        building as a DSO.])])
+                        building as a DSO.  The default is to build all
+                        components staticly.])])
     AC_ARG_ENABLE([mca-direct],
         [AS_HELP_STRING([--enable-mca-direct=LIST],
                        [Comma-separated list of type-component pairs that
@@ -166,15 +166,21 @@ AC_DEFUN([OPAL_MCA],[
     # resolution (prefer static) is done in the big loop below
     #
     AC_MSG_CHECKING([which components should be run-time loadable])
-    if test "$enable_static" != "no" || test "$OPAL_ENABLE_DLOPEN_SUPPORT" = 0; then
+    if test "$enable_static" != "no"; then
         DSO_all=0
-        msg=none
-    elif test -z "$enable_mca_dso" || test "$enable_mca_dso" = "yes"; then
-        DSO_all=1
-        msg=all
+        msg="none (static libraries built)"
+    elif test "$OPAL_ENABLE_DLOPEN_SUPPORT" = 0; then
+        DSO_all=0
+        msg="none (dlopen disabled)"
+    elif test -z "$enable_mca_dso"; then
+        DSO_all=0
+        msg=default
     elif test "$enable_mca_dso" = "no"; then
         DSO_all=0
         msg=none
+    elif test "$enable_mca_dso" = "yes"; then
+        DSO_all=1
+        msg=all
     else
         DSO_all=0
         ifs_save="$IFS"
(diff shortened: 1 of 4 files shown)


bwbarrett:
@jsquyres You had brought this up a couple months ago.  I'm not sure I have strong preferences on whether or not we should take this patch, but here's how we should make the change in component build style.

config/opal_mca.m4
```diff
@@ -706,20 +714,23 @@ AC_DEFUN([MCA_COMPONENT_COMPILE_MODE],[
...
-       test "$STATIC_COMPONENT" = "1" || \
-       test "$STATIC_all" = "1" ; then
-        $4="static"
-    elif test "$shared_mode_override" = "dso" || \
-         test "$SHARED_FRAMEWORK" = "1" || \
-         test "$SHARED_COMPONENT" = "1" || \
-         test "$DSO_all" = "1"; then
-        $4="dso"
+    # Look for the most specific specifier between static/dso.  If
+    # there is a tie (either none specified or eual specified), prefer
```
jsquyres, line 718:
```suggestion
    # there is a tie (either neither or both specified), prefer
```

jsquyres (commented):
I do remember bringing this up a few months ago (I think after talking with @rhc54 about it...?).  I don't remember if there was a reason that I didn't end up filing a PR like this (e.g., a technical or UX reason).  ...or were we waiting for The Great 3rd Party updates to be completed before doing this?

lanl-ompi:

config/opal_mca.m4
bwbarrett: fixed.

bwbarrett:
@jsquyres my recollection was that I said I'd look into it and wasn't sure how hard it would be.  Turns out, not very :).

README
```diff
@@ -968,18 +968,21 @@ INSTALLATION OPTIONS
 --enable-dlopen
   Build all of Open MPI's components as standalone Dynamic Shared
```
bwbarrett, line 969: This line has never actually been true.  --enable-dlopen has always just allowed for component loading at runtime.  It was never the cause of the default behavior.  It's definitely not true now.

bwbarrett:
@jsquyres your readme changes look good to me.  Thanks!

jsquyres (approved):

rhc54:
Do we need/want to port this to PMIx and PRRTE?

yosefe:
does it mean that if one component depends on external shared library which is not available, then OpenMPI would not be able to ignore just this one component?

bwbarrett:
> does it mean that if one component depends on external shared library which is not available, then OpenMPI would not be able to ignore just this one component?

Yes.  Most of our customers don't seem to do this.  For those that do, adding --enable-mca-dso=<blah> is probably the recommended course of action.

bwbarrett:
> Do we need/want to port this to PMIx and PRRTE?

If we think this is the right default behavior, yes.  I think there's an open question about whether we want to do this.  There was a question about ability to make the change, and I wanted to show that it wasn't a hard change...

bosilca:
Both of these concerns are legit (reducing the IO traffic and not breaking loading the OMPI library on missing external dependencies). We can take a middle-ground approach, where we allow components with no external dependencies (BTL sm, tcp, some mpools) to be statically linked in, while everything that has external dependencies remains as dynamic libraries (except when specificaly overwritten with the option described above). We can select the desired behavior either in the configure.m4 or as a file in the component directory (similar to .ignore).

jsquyres:
> Both of these concerns are legit (reducing the IO traffic and not breaking loading the OMPI library on missing external dependencies). We can take a middle-ground approach, where we allow components with no external dependencies (BTL sm, tcp, some mpools) to be statically linked in, while everything that has external dependencies remains as dynamic libraries (except when specificaly overwritten with the option described above). We can select the desired behavior either in the configure.m4 or as a file in the component directory (similar to .ignore).

This is a good idea.  I don't know if I want this to *always* be the behavior -- e.g., for the Cisco distribution of Open MPI, I'm fine if the core Open MPI libraries depend on libfabric.  But a generic Linux distribution may make a different choice, for example.

Perhaps `--enable-mca-dso` can have another value: `auto`.  If the value is `auto`, then it'll look to the component for guidance on whether it should be compiled as a DSO or static (via an optional m4 directive in the component's `configure.ac` to indicate which way it should go, and if it's not specified, use a global default).  This will allow components to make their own decision if they want to.

yosefe:
@jsquyres [suggestion ](https://github.com/open-mpi/ompi/pull/8132#issuecomment-718694885) sounds good to me

markalle:
TL;DR : I'd propose keeping treematch in its own private ```mca_topo_treematch.so``` since it exports a lot of generic symbol names that could cause problems if they appeared in ```libmpi.so```.

I have a testcase proposal
https://github.com/open-mpi/ompi-tests-public/pull/3
that looks at exported symbol names in files like libmpi.so and reports errors if the symbols don't have some recognized prefix that would segregate them from colliding with users's applications.

And I started a PR to change a bunch of symbols to static and/or add ```ompi_``` prefixes, based on how this PR pulls many MCA's into libmpi.so
https://github.com/open-mpi/ompi/pull/8262

But there were objections at least to modifying the treematch MCA.  And that MCA in particular is littered with unreasonably generic exported symbol names, like ```choose```, ```create_work```, ```exchange```, ```get_time```, and ```update_val``` for example.

So for this PR I'd at least propose treematch be kept in its own ```mca_topo_treematch.so``` if fixing its generic symbol names is being declared off limits.

jjhursey:
We had a discussion today about this PR, and I just wanted to clarify something for myself. We are not eliminating the DSO mechanism, just changing the default for (most or all) of the components in the tree. So I can still build a DSO an drop it into the DSO lib search path and it will try to include it in the framework component loading. Is that correct?

I say this because it is common for us to ship components that, depending on the client's system configuration, that component may not load because of missing libraries. That's ok and we depend on Open MPI's fallback to the next best component in that case. This is another instance where we never want this component to be linked against libmpi.so since the dependent library may not be present and we still want to run.

rhc54:
@jjhursey that is one of the proposals on the table - but not the only one. 😄

jsquyres:
@rhc54 and I talked about this on the phone today:

1. We still think that this is a good idea.  In current Open MPI master, a typical build on my cluster shows 51 DSOs.  Reducing the corresponding number of `dlopen()`s is probably a Good Idea.
1. Slurping in the components to project libraries without disabling dlopen is not going to completely eliminate filesystem access at process launch and during `MPI_INIT`, of course, but it will definitely significantly decrease it.  This will be mostly felt at mid- to large-scale, but we're assuming that there's a non-zero number of "mid-sized" systems out there that may still have 1GB Ethernet + NFS for their filesystem.  This may be a noticeable help for those users.
1. We think that most users will fall into one of the following categories:
    1. Whatever the default is (i.e., they don't care).  **This is probably the majority of users.**
    1. All-static. Large scale users will want this.  And per above, this will probably even benefit mid-scale users.
    1. All-DSO.  Developers will probably still use this for convenience (e.g., when developing/debugging a single component).  Some small number of users may still want this for historical or religious reasons.
    1. Some static and some DSO.  This can be useful for users that have a single Open MPI install that spans multiple different clusters, and they want different components loaded for each cluster.  We expect this to be a pretty small number of users.
1. Adding another level of defaults (e.g., the `auto` system I proposed, above) would seem to:
    1. Further complicate the already-complicate m4 code
    2. Potentially violate the Law of Least Surprise for end users
    3. Be difficult to explain to the user.
1. Our opinions:
    1. The defaults of this PR will be good for most users (i.e., static).
    1. This PR gives the ability to go all-DSO for those who want it.
    1. This PR gives the ability to do "mostly static, but a few DSOs" for those who want it.

In short:

* I retract my `auto` proposal from above.  I don't think it's worth it.
* This PR has the potential to improve startup filesystem performance for users.
* Static linking of components also (slightly) improves per-MPI-process memory utilization.
* This PR has CLI options to cover all use cases.
* This PR is good to go.  We should merge.

jjhursey:
Is there a timeout by when we want to try to merge this PR? Just to put a time box on the discussion.

jsquyres:
> Is there a timeout by when we want to try to merge this PR? Just to put a time box on the discussion.

No, not really.  The main issue I think is that if we want to do this, we should give our sister projects time to follow our lead (i.e., PMIx / PRRTE).

jsquyres:
After I posted yesterday, I re-read everything more carefully and was reminded of the treematch symbol pollution issues.  That should be addressed; we don't want poorly-named symbols polluting the MPI application namespace.  @jjhursey and I talked through this a bit on the phone today.  Things I'll look into:

* #8262 -- @markalle took a shot at this already.
* Can we use a `rename.h`-style approach to make all the treematch symbols safe?  Yes, this has the drawback of editing treematch, but the impact might well be minimal (i.e., just adding `#include "rename.h"` in each treematch file)
* If we make treematch not be a libtool convenience library, and instead keep it as a separate library that gets installed (e.g., in `$pkglibdir`) and then have the component link against it as a regular shared library, do we still have the symbol pollution problem?  I don't know/remember offhand.

I'll check these things out.

gpaulsen:
We started a good discussion on the 1/12 call, but ran out of time, so we'll tackle this earlier on 1/19 call for those who are interested.

gpaulsen:
Failed to get to this on 1/19 due to release branch discussions.

jsquyres:
The symbol pollution issue is a real one, and we should figure it out.  But keep in mind that the symbol pollution problem _already exists_ and _isn't because of this PR._

This PR makes the effects of it more pronounced, but the problem already exists.

markalle:
On the option @jsquyres listed about what if treematch was its own separate library and libmpi was linked against it, I think that still exposes the end user to all the genericly named symbols in libtreeematch.so.  We'd be able to say "hey, it's not OUR fault, those symbols aren't in our code" but the user experience would still be bad.  If we dlopened libtreematch.so and accessed everything in it via function pointer I believe that would be safe.  In the olden days Platform MPI did that for several external libraries like libibverbs.  I'd still lean toward the rename.h approach though

markalle:
@jsquyres also to put on your radar screen, a proposed testcase for generic symbol names being exported:  https://github.com/open-mpi/ompi-tests-public/pull/3  It currently accepts a pretty broad list of prefixes.

And even though there might be discussions still about whether to fix this with prefixes or by going all in on DECLSPEC and only exporting a strict list, the testcase wouldn't care about that, it just looks at the results.

bwbarrett:
Jumping back into this conversation...

We load DSOs into the global scope today (because reasons) and allow customers to build as static libraries, so it seems like we should just fix treematch, rather than screw with defaults.

I think what George is suggesting with allowing components to set a default might make sense, although it's more work than I can commit to doing right now (particularly chasing through setting a proper default for all the components).

To clarify for Josh, this does not change the ability for Open MPI to search for components built out of tree (or similar).  This only changes whether components are built as static or dso libraries by default.  For at least 10 years, we've split the logic between looking for dsos (the `--disable-dlopen` option) and how components are built (the `--enable-mca-dso` and `--enable-mca-static` options).

gpaulsen:
@bwbarrett - This looks ready to go.  Please PR to v5.0 after merging.

PR #8136, opened 2020-10-26
configury: make build reproducible

bmwiedemann:
If defined, use SOURCE_DATE_EPOCH environment variable make the build
Reproducible by forcing timestamps.  See
https://reproducible-builds.org/docs/source-date-epoch/ for more
information.

Thanks Bernhard M. Wiedemann for bringing this to our attention.

Fixes open-mpi/ompi#3759

This is mostly a rebase of #3779 by @ggouaillardet with some feedback from @jsquyres applied.


awlauria:
ok to test

config/opal_functions.m4
```diff
@@ -96,7 +96,7 @@ EOF
 
 OPAL_CONFIGURE_USER="${USER:-`whoami`}"
 OPAL_CONFIGURE_HOST="${HOSTNAME:-`(hostname || uname -n) 2> /dev/null | sed 1q`}"
-OPAL_CONFIGURE_DATE="`date`"
+OPAL_CONFIGURE_DATE="`date --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}"`"
```
jsquyres, line 99: Is there a reason to not use `config/getdate.sh` here?

bmwiedemann: Would `$(top_srcdir)` be available here?

jsquyres: It looks like `$(srcdir)` is available -- is that sufficient?

jsquyres:
@bmwiedemann I took the liberty of pushing to this branch:

1. I removed some redundant code from `opal_functions.m4` (which affected this PR)
1. I changed to `$srcdir/getdate.sh` in a few places

It seems to work.  Can you confirm?

config/opal_get_version.m4
```diff
@@ -91,7 +91,7 @@ m4_define([OPAL_GET_VERSION],[
                     $2_REPO_REV=`git describe --tags --always`
                 fi
             else
-                $2_REPO_REV="date`date '+%Y-%m-%d'`"
+                $2_REPO_REV=`$srcdir/getdate.sh`
```
bmwiedemann, line 94: Was it intentional to change the `$2_REPO_REV` date format? Who is using this? could this break scripts?

jsquyres: Good catch.  Yes, this change in format will break things.

Fix inbound...

README
```diff
@@ -197,10 +197,11 @@ General notes
...
 - Open MPI has taken some steps towards Reproducible Builds
   (https://reproducible-builds.org/).  Specifically, Open MPI's
-  "configure" and "make" process, by default, records some
-  system-specific information such as the hostname where Open MPI was
-  built and the username who built it.  If you desire a Reproducible
-  Build, set the $USER and $HOSTNAME environment variables before
+  "configure" and "make" process, by default, records the build date
+  and some system-specific information such as the hostname where
+  Open MPI was built and the username who built it.
+  If you desire a Reproducible Build, set the $SOURCE_DATE_EPOCH,
```
jsquyres, line 203: Should we put something here about the expected format of `$SOURCE_DATE_EPOCH` (e.g., a simple/trivial example)?

jsquyres: Actually, it looks like the commit message has a URL that points to how to set it... but that URL is now stale (i.e., there's a new URL).

Maybe it's sufficient to a) update the commit message and b) also put that URL here in the README text.

jsquyres: I updated README and the commit message.  Please review.

bmwiedemann: README LGTM

config/opal_get_version.m4
bmwiedemann: I dislike the added complexity of the `--ymd`
My fix would have been:

```diff
config/getdate.sh
-date --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}"
+date --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" "$@"


config/opal_get_version.m4
-  $2_REPO_REV=`$srcdir/getdate.sh`
+  $2_REPO_REV=date`$srcdir/getdate.sh '+%Y-%m-%d'`
```

In the end, it is you guys maintaining it, so you decide which variant you like.

jsquyres: Sure, I have no real opinion here.  I'll make the change you suggested.

bmwiedemann: I pushed a small fix to re-add the `date` prefix string to keep the previous format. Looks good to me now.

jsquyres (approved):

Issue #8117, opened 2020-10-22
mpirun realloc() invalid next size


zrss:
Thank you for taking the time to submit an issue!

### What version of Open MPI are you using? (e.g., v3.0.5, v4.0.2, git branch name and hash, etc.)

v4.0.0 and compiled with src code

### Describe how Open MPI was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

from a source tarball

### Please describe the system on which you are running

* Operating system/version: ubuntu16.04
* Computer hardware: x86_64
* Network type: socket

-----------------------------

## Details of the problem

mpirun failed at `realloc`

```
*** Error in `mpirun': realloc(): invalid next size: 0x0000000001af4da0 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f59483ff7e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x834aa)[0x7f594840b4aa]
/lib/x86_64-linux-gnu/libc.so.6(realloc+0x179)[0x7f594840c839]
/usr/local/openmpi/lib/libopen-pal.so.40(opal_util_keyval_parse+0x459)[0x7f59489e0ac9]
/usr/local/openmpi/lib/libopen-pal.so.40(+0x5d08a)[0x7f59489cc08a]
/usr/local/openmpi/lib/libopen-pal.so.40(mca_base_var_cache_files+0x455)[0x7f59489cefe5]
/usr/local/openmpi/lib/libopen-pal.so.40(opal_init_util+0x1a2)[0x7f59489a7e32]
/usr/local/openmpi/lib/libopen-rte.so.40(orte_submit_init+0x19d)[0x7f5948d1774d]
mpirun[0x400fcf]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f59483a8830]
mpirun[0x400e39]
======= Memory map: ========
00400000-00402000 r-xp 00000000 fc:14 252069977                          /usr/local/openmpi/bin/orterun
00601000-00602000 r--p 00001000 fc:14 252069977                          /usr/local/openmpi/bin/orterun
00602000-00603000 rw-p 00002000 fc:14 252069977                          /usr/local/openmpi/bin/orterun
01ab7000-01afe000 rw-p 00000000 00:00 0                                  [heap]
7f5940000000-7f5940021000 rw-p 00000000 00:00 0 
[...]

PR #8144, opened 2020-10-27
OPAL: fix string buffer allocation for large env variables

devreal:
In OPAL's keyval parser in `add_to_env_str`, for the first variable a buffer of size 1024 was allocated, regardless of the space needed. For any additional variables, the size might have been doubled at best, even if more space was required. This PR makes sure the allocated buffer is sufficiently large to store the value.

See https://github.com/open-mpi/ompi/issues/8117

*Note:* the issue described in #8117 does not seem to occur on `master`. I will backport this to 4.0.x (where it occurs) and 4.1.x (not tested yet) but the same code exists in PMIX and PRRTE so it should be picked up there as well.

Signed-off-by: Joseph Schuchart <schuchart@icl.utk.edu>

### DIFF
--- a/opal/util/keyval_parse.c
+++ b/opal/util/keyval_parse.c
@@ -276,21 +276,37 @@ static int save_param_name (void)
 
 static int add_to_env_str(char *var, char *val)
 {
-    int sz, varsz, valsz;
+    int sz, varsz, valsz, new_envsize;
     void *tmp;
 
     if (NULL == var) {
         return OPAL_ERR_BAD_PARAM;
     }
 
+    varsz = strlen(var);
+    if (NULL != val) {
+        valsz = strlen(val);
+        /* account for '=' */
+        valsz += 1;
+    }
+    sz = 0;
     if (NULL != env_str) {
-        varsz = strlen(var);
-        valsz = (NULL != val) ? strlen(val) : 0;
-        sz = strlen(env_str)+varsz+valsz+2;
-        if (envsize <= sz) {
-            envsize *=2;
+        sz = strlen(env_str);
+        /* account for ';' */
+        sz += 1;
+    }
+    /* add required new size incl NUL byte */
+    sz += varsz+valsz+1;
 
-            tmp = realloc(env_str, envsize);
+    /* make sure we have sufficient space */
+    new_envsize = envsize;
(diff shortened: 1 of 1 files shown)


bosilca (approved):

opal/util/keyval_parse.c
```diff
@@ -276,33 +276,38 @@ static int save_param_name (void)
...
-    if (NULL != env_str) {
-        varsz = strlen(var);
-        valsz = (NULL != val) ? strlen(val) : 0;
-        sz = strlen(env_str)+varsz+valsz+2;
-        if (envsize <= sz) {
-            envsize *=2;
+    varsz = strlen(var);
+    valsz = (NULL != val) ? strlen(val) : 0;
+    sz = (NULL != env_str) ? strlen(env_str) : 0;
+    sz += varsz+valsz+2;
```
bosilca, line 289: If you want to be picky that's one wasted byte when ```NULL != env_str```.

jsquyres:
FYI @rhc54 @jjhursey for the "we should also do this in PMIX and PRRTE" note.

opal/util/keyval_parse.c
```diff
@@ -276,33 +276,38 @@ static int save_param_name (void)
...
-            tmp = realloc(env_str, envsize);
+    if (NULL != env_str) {
+        if (new_envsize > envsize) {
+            tmp = realloc(env_str, new_envsize);
             if (NULL == tmp) {
                 return OPAL_ERR_OUT_OF_RESOURCE;
             }
             env_str = tmp;
         }
         strcat(env_str, ";");
```
jsquyres, line 303: Is there a `+1` missing somewhere to account for the `;` character?

jsquyres: There's also a `strcat()` of an additional `=` character on old line 309.  Does that need to be counted, too?

devreal: You're right, the original version didn't account for the terminating NUL byte. The chance of hitting that is pretty slim as allocations are made in multiples of 1K but I force-pushed an update  that makes accounting =, ;, and \0 explicit.

jsquyres (approved):

jsquyres:
Please also update #8155 and #8156.  Thanks!

PR #8147, opened 2020-10-28
v4.0.x: reproducible builds + portability fix

jsquyres:
Cherry pick of #8136 and #8152 from master regarding reproducible builds.  Thanks to @bmwiedemann for bringing this back to our attention.

Also bring in a MacOS/`readlink` portability fix.


bmwiedemann:
This and 4.1 needs one more patch:
```diff
orte/tools/orte-info/Makefile.am
-            -DOMPI_BUILD_DATE="\"`date`\"" \
+            -DOMPI_BUILD_DATE="\"`$(top_srcdir)/config/getdate.sh`\"" \
```

jsquyres:
@bmwiedemann I notice on master that in one place, we have:

```sh
${USER:-`whoami`}
```

but in other places we just have

```sh
$$USER
```

Should they all be of the form 

```sh
# With $ or $$, as appropriate
{USER:-`whoami`}
```

?

bmwiedemann:
I'd leave it as is. I have seen good PRs with feature-creep in the past that got stalled too long.

jsquyres:
Ah, I see -- `$USER` is an automatically-defined shell variable, anyway (I thought it was something unique to Reproducible builds, and therefore might not actually be set).

rhc54:
Is it safe to say we need similar changes in PMIx and PRRTE?

bmwiedemann:
@rhc54 I tested with our `openmpi4` package and these changes (with orte-info) were enough to make it reproducible.

rhc54:
Understood - my apologies. My question was more about building PMIx and PRRTE standalone packages (i.e., when they aren't embedded in OMPI)

jsquyres:
@rhc54 It would not be a bad thing to also apply this to PMIx or PRRTE.  I wouldn't say it's critical, but it's definitely in the "nice to have" category.

jsquyres:
This PR will also need the MacOS portability fixes from #8152.

jsquyres:
bot:ompi:retest

rhc54 (approved):

orte/tools/orte-info/Makefile.am
```diff
@@ -26,7 +26,7 @@ AM_CFLAGS = \
             -DOPAL_CONFIGURE_DATE="\"@OPAL_CONFIGURE_DATE@\"" \
             -DOMPI_BUILD_USER="\"$$USER\"" \
             -DOMPI_BUILD_HOST="\"`(hostname || uname -n) | sed 1q`\"" \
```
bmwiedemann, line 28: Tested it and this line still caused variations.
=>
```
\"$${HOSTNAME:-`(hostname || uname -n) 2> /dev/null | sed 1q`}\"" \
```

jsquyres: oshmem_info and mpisync will therefore cause the same issue.

rhc54: So we need that line to change? FWIW: that suggested line is what we have in PRRTE today (someone asked for reproducible build over there earlier)

jsquyres: I just pushed that change here (and #8148) for orte-info and mpisync.  Anywhere you have a `-DBUILD_HOST` kind of construct in PMIX/PRTE Makefile.am's, you'll want to allow the `$HOSTNAME` override.

bmwiedemann (approved):

PR #8167, opened 2020-11-01
.gitignore: ignore sym linked F08 profile bindings

jsquyres:
A recent commit made the use_mpi_f08 bindings sym link into their
profile directory (just like we do for C and other bindings) instead
of having standalone PMPI-ized copies of the bindings.  Make sure to
.gitignore the sym linked files.


ggouaillardet (approved):
Thanks Jeff!

in a near future, we might want to revamp `[p]sizeof_f08.f90` generation to make it consistent with other bindings (e.g. `#include "mpi-f08-rename.h"`)

PR #8168, opened 2020-11-01
README: Provide example of differing linker search paths

jsquyres:
I ran into this exact case on MacOS (the C and Fortran compiler have
different default linker search paths).  Technically, we've always had
this problem, but it has just become a bit more likely for real people
to run into because we're now preferring the system-installed
Libevent.

### DIFF
--- a/README
+++ b/README
@@ -230,6 +230,23 @@ Compiler Notes
   example, such configurations may require additional compiler /
   linker flags to make Open MPI build properly.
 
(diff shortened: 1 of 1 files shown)


jsquyres:
@bosilca @rhc54 I'm guessing you guys haven't run into this because you don't typically build the Fortran bindings on MacOS...?

rhc54 (approved):
Sadly true...

